PR #901, opened 2019-03-18
KEP: add non-preempting option to PriorityClasses

vllry:


keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+    * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+    * [Risks and Mitigations](#risks-and-mitigations)
+* [Graduation Criteria](#graduation-criteria)
+* [Implementation History](#implementation-history)
+
+
+## Summary
+
+[PriorityClasses](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) are a beta feature,
```
k82cn: it's GA now :)

keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+the scheduler will not preempt pods in order to schedule a pod of that priority.
+
+## Motivation
+
+High-priority, non-preempting workloads are a common data science use case.
+Preempting batch workloads is a waste, as the work unit must be repeated.
+
+### Goals
+
+Add a boolean to PriorityClasses,
```
k82cn: Do we also need to add this field into Pod? For now, PriorityClass will be translate into `spec.priority` in pod for scheduling.

keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+
+Add a NonPreempting field to PriorityClasses.
+This field will default to false,
+for backwards compatibility.
+
+If NonPreempting is false,
+the scheduler will preempt lower priority pods to schedule this pod,
+as is current behavior.
+
+If NonPreempting is true,
```
k82cn: Do we support update this?

keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+### Goals
+
+Add a boolean to PriorityClasses,
+to enable or disable preemption for pods of that PriorityClass.
+
+### Non-Goals
+
+## Proposal
+
+Add a NonPreempting field to PriorityClasses.
```
k82cn: That'll be great if there's example code of `types.go` :)

keps/sig-scheduling/20190317-non-preempting-priorityclass
k82cn: BTW, please also highlight whether kubelet should consider this option/configuration :)

keps/sig-scheduling/20190317-non-preempting-priorityclass
vllry: My bad, docs are old and say beta. Filed https://github.com/kubernetes/website/issues/13253 re: that.

keps/sig-scheduling/20190317-non-preempting-priorityclass
vllry: I'll have to double check the code.

keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+editor: Vallery Lancey
+creation-date: 2019-03-17
+last-updated: 2019-03-17
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Allow PriorityClasses To Be Non-Preempting
```
misterikkit: GitHub hasn't figured out this is a markdown file. Could you rename to `20190317-non-preempting-priorityclass.md`?

keps/sig-scheduling/20190317-non-preempting-priorityclass
```diff
@@ -0,0 +1,104 @@
...
+
+High-priority, non-preempting workloads are a common data science use case.
+Preempting batch workloads is a waste, as the work unit must be repeated.
+
+### Goals
+
+Add a boolean to PriorityClasses,
+to enable or disable preemption for pods of that PriorityClass.
+
+### Non-Goals
```
misterikkit: Seems like "protecting pods from preemption" is not a goal of this KEP. It's worth mentioning that PodDisruptionBudget is meant to serve that purpose.

keps/sig-scheduling/20190317-non-preempting-priorityclass
misterikkit: Do we need to finalize the API in the kep, or during implementation?

keps/sig-scheduling/20190317-non-preempting-priorityclass
misterikkit: The nice thing about this KEP is that it only affects the scheduling of new pods and not the "evictability" of running pods. Updating this value seems simple to support.

derekwaynecarr:
If I declare a priority class "foo" and say it is not pre-emptible, is a system-node-critical or system-cluster-critical priority based workload able to pre-empt it?  It seems like we would need to always enable those priorities to force pre-emption to maintain a functional cluster.

misterikkit:
> If I declare a priority class "foo" and say it is not pre-emptible, is a system-node-critical or system-cluster-critical priority based workload able to pre-empt it? It seems like we would need to always enable those priorities to force pre-emption to maintain a functional cluster.

So I had to read the KEP carefully to get this, but I don't think that's how it works. This new setting controls whether a pod will **trigger** eviction of lower priority pods, not whether the pod can itself be evicted.

If that's the behavior we decide to implement, we'll need to be extra careful to document it clearly.

derekwaynecarr:
@misterikkit -- i reread, and i think this is right.  the system-* priorities can pre-empt other workloads, but the "data-science" priority would not pre-empt anything else.  if that was the case, that makes sense.

keps/sig-scheduling/20190317-non-preempting-priorityclass
vllry: Thanks, that's a good suggestion.

vllry:
I'll think about rephrasing the description so that's more apparent.

keps/sig-scheduling/20190317-non-preempting-priorityclass
k82cn: IMO, it's better to have the API there which is easier for other to understand which part of code are going to be update; and it's ok to change it later during implementation.

keps/sig-scheduling/20190317-non-preempting-priorityclass
k82cn: hmm.... if we add a new field in `PodTemplate`, it's hard to update created pods :) If we hold PriorityClass in scheduler, how about kubelet?

vllry: I'll look more into this.

keps/sig-scheduling/20190317-non-preempting-priorityclass
liggitt: numeric priority was denormalized from priority class into the pod at admission time, which:
* avoids issues with mutation of priorityclass objects
* ensures the scheduler only has to pay attention to pods
* allows kubelets to set numeric priority on static pods

is there a reason not to do the same here?

bsalamat: IIUC, the idea that @k82cn and @liggitt are proposing is to add a "Preempting" field to PodSpec instead of PriorityClass. Then a user sets the field at the time of pod creation and its default value will be true. Is my understanding correct?
I like this idea, because as Jordan pointed out, the scheduler does not need to look at PriorityClasses. It also gives us the flexibility to use the same PriorityClass for two sets of pods where one set can preempt others and the other cannot.
The downside of this approach is that we lose the ability to grant preempting quota to users. ResourceQuota supports PriorityClass. An admin can give a certain amount of quota to a user at a specific priority identified by a PriorityClass. When the "Preempting" field is in PriorityClass, an admin can choose to give users preempting or non-preempting priority, but if the field in the PodSpec, the admin will not be able to choose between the two options anymore. The admin will only be able to assign quota at a specific priority and it will be up to users to decide whether they want to create pods as preempting or non-preempting.

An alternative approach is to add the field to both PriorityClass and PodSpec. Similar to the "Priority" field of PodSpec, the "preempting" field is populated from the PriorityClass by the "Priority" admission controller at the time of pod creation. This approach has the benefits that Jordan pointed out without the downside that I mentioned above.

vllry: Thanks, that makes sense.
cc @denkensk

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+
+* Users are reporting that this resolves their workload priority use-cases
+(if not, additional enhancements would be tightly linked to this one).
+* The feature has been stable and reliable in at least 2 releases.
+* Adequate documentation exists for preemption and the optional field.
+* Test coverage includes non-preempting use cases.
+* Conformance requirements for non-preempting PriorityClasses are agreed upon.
+
+## Testing Plan
+Add unit and e2e tests for nonpreempting PriorityClasses to the existing scheduler tests.
```
bsalamat: We need integration tests as well. In fact, most of the logic should be tested in unit and integration tests. e2e should be used to ensure that end-to-end workflow (admission control, API server, scheduler, kubelet) works as expected. e2e tests don't need to have full coverage of the added code.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+### Risks and Mitigations
+
+The new feature may malfuction,
+or preemption may be accidentally impaired.
+New tests (covering both nonpreepting workloads and mixed workloads),
+and the existing preempting PriorityClass tests should be used to prove stability.
+
+## Graduation Criteria
+
+* Users are reporting that this resolves their workload priority use-cases
```
ravisantoshgudimetla: So this feature will be featuregated or since for the most part it will be backward compatible assuming default preempting field value to be true, we'd keep it without a featuregate?

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+* [Implementation History](#implementation-history)
+
+
+## Summary
+
+[PriorityClasses](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/) are a GA feature as on 1.14,
+which impact the scheduling and eviction of pods.
+Pods will be scheduled according to descending priority.
+If a pod cannot be scheduled due to insufficient resources,
+lower-priority pods will be descheduled ("preempted") to make room.
```
ravisantoshgudimetla: Can you remove please `descheduled` here as it might confuse people using descheduler

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+
+The new feature may malfuction,
+or preemption may be accidentally impaired.
+New tests (covering both nonpreepting workloads and mixed workloads),
+and the existing preempting PriorityClass tests should be used to prove stability.
+
+## Graduation Criteria
+
+* Users are reporting that this resolves their workload priority use-cases
+(if not, additional enhancements would be tightly linked to this one).
```
ravisantoshgudimetla: Also adding use-cases section might be helpful.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
vllry: I don't think we'd need a feature gate. The default behavior is the same as existing behavior.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
vllry: I mentioned the broad use case, but I'm not very familiar with with the workloads that require this.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
liggitt: new fields added to stable APIs get feature gated for a release to ensure HA upgrade can succeed without data loss

vllry: Oh, okay.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
bsalamat: One of the more important use-case of this feature is for batch workloads. In batch processing, clusters often have much more work that they can fit simultaneously onto the nodes. So, there are many pods waiting in the queue to be scheduled. If a user has an important workload that needs to be scheduled before other pods in the queue, the user puts a higher priority on the workload to make it go to the head of the scheduling queue. At the same time, they don't want this workload to preempt running pods. Preemption is not desired for batch workloads, because a preempted pod often loses all it had done when it gets preempted and has to start from scratch next time it is scheduled. In order to achieve faster scheduling without preempting existing pods, the user can put a non-preempting high priority on their pods.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+
+This proposal makes the preempting behavior optional for a PriorityClass,
+by adding a new field to PriorityClasses.
+If a PriorityClass does not have preemption enabled,
+a pod of that PriorityClass will not trigger preemption of other pods.
+
+## Motivation
+
+High-priority, non-preempting workloads are a common data science use case.
+Preempting batch workloads is a waste, as the work unit must be repeated.
```
davidopp: I think this could be explained a bit more clearly. This sentence is fine, but after it I would add something like "In general, there may be workloads that are not time-sensitive to schedule (it's OK for them to wait if the cluster is full), but that once running, it is important not to preempt them."

vllry:
@denkensk it would help if you could handle some of the technical questions, as the current owner of https://github.com/kubernetes/kubernetes/issues/67671

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
```diff
@@ -0,0 +1,116 @@
...
+
+```
+type PriorityClass struct {
+	metav1.TypeMeta
+	metav1.ObjectMeta
+	Value int32
+	GlobalDefault bool
+	Description string
+	// New option
+	Preempting bool
```
denkensk: a bool pointer here？

keps/sig-scheduling/20190317-non-preempting-priorityclass
ravisantoshgudimetla: So, are we going to have `Preempting` field in both podSpec and PriorityClass?

vllry:
Will do.

keps/sig-scheduling/20190317-non-preempting-priorityclass.md
vllry: This paste is for the PriorityClass (which doesn't use a pointer)

denkensk: But how can we set the default as true ?

vllry: In defaults.go [(reference)](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#edit-defaultsgo) you can convert nil/unset to true.

liggitt: defaulting an optional field while in alpha doesn't work well (since we don't want to persist or return the field of the alpha feature is disabled). this will need to be a *bool so we can correctly omit it until promoted

denkensk: I think if we want to distinguish between an unset value and an automatic false value of bool, we also need *bool?

vllry: Thanks for the context @liggitt, that makes sense. I haven't worked with the API before.

keps/sig-scheduling/20190317-non-preempting-priorityclass
vllry: Looks like we're going with denormalizing the field into PodSpec... so from my understanding, that will present some upgrade challenges.

bsalamat:
To clarify my comment, you may want to move part of the current "motivation" section to a "user story" section.

vllry:
Thanks, @bsalamat, that makes sense.

Still working on the details about feature gating and pod spec upgrades... but it's starting to look a lot more complete.

vllry:
I've addressed all the outstanding comments, except for compatibility and changes w/r/t the PodSpec. I'll do more reading this weekend on the implications and how this is normally handled, but suggestions would be appreciated.

PR #904, opened 2019-03-19
Add a KEP for graduating PodDisruptionBudget to stable

bsalamat:
/sig apps
/assign @kow3ns @liggitt

### DIFF
--- a/keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
+++ b/keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
@@ -1,0 +1,134 @@
+---
+title: graduate-pod-disruption-budget-to-stable
+authors:
+  - "@bsalamat"
+owning-sig: sig-apps
+participating-sigs:
+  - sig-scheduling
+reviewers:
+  - "@liggitt"
+  - "@kow3ns"
+approvers:
+  - "@kow3ns"
+  - "@liggitt"
+editor: TBD
+creation-date: 2019-03-18
+last-updated: 2019-03-18
+status: implementable
+see-also:
+  - 
+replaces:
+superseded-by:
+---
+
+# Graduate PodDisruptionBudget to stable
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+[Pod Disruption Budget (PDB)](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
+is a Kubernetes API that limits the number of pods of a collection that are down simultaneously from voluntary disruptions.
+[Kubernetes eviction API](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#the-eviction-api)
+takes PDB into account when terminating pods. If PDB is
+violated, the eviction API returns failure and does not delete the requested pod.
+The feature was introduced in Kubernetes 1.4 and promoted to beta in 1.5.
+It has been in beta for a long time. This document lays out the plan to promote
+it to stable.
+
+## Motivation
+
+PDB API has been stable and is an important feature that allows users to improve
+reliability of their critical workloads. This feature has been in beta for a
+long time. We need to promote to stable version given that we plan to support it
+long term.
+
+### Goals
+
+* Plan to promote PDB API to stable version.
+* List action items to mitigate potential risks of supporting mutable PDB.
+
+### Non-Goals
+
+* Making changes to the API fields or their meanings.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+#### Mutable PDBs
+
+A mutable PDB object allows its `MinAvailable`, `MaxUnavailable`, and `Selector`
+fields to be modified by clients. Components that use PDB must watch such
+modifications and use the updated values when making decisions.
+
+This feature is implemented by [this PR](https://github.com/kubernetes/kubernetes/pull/69867).
+
+### Risks and Mitigations
+
+We plan to support mutation of PDB objects that didn't exist in previous versions.
+The following needs to be checked and verified before graduation of the API.
+
+- [ ] Ensure that components do not have any logic that relies on immutability
+of PDBs. For example, if component builds a cache of pods that match various
+PDBs, they must add logic to invalid the cache on updates.
+   - Action Item: sweep for in-tree (including kubernetes/* repos) uses to make
+   sure components are informer driven and aren’t assuming immutability.
+- [ ] Check what watches PDBs and make sure there is no performance concerns.
+- [ ] Updating a PDB could cause the state of a cluster to seem incorrect. For
+example, a PDB states that at least 10 replicas of a collection must be alive.
+Kubernetes control plane evicts some of the existing pods, but keeps at least 10
+around. The PDB is updated and states that at least 20 replicas must be kept
+alive. It may appear to an observer that the evictions happened before the PDB 
+update were incorrect, if they don’t notice the PDB update.
+  - Action Item: Update documents and explain mutability of PDB its side effects.
+
+### Test Plan
+
+#### Existing Tests
+PodDisruptionBudget currently has tests in various components that use the feature:
+
+* Disruption controller
+  - [This test](https://github.com/kubernetes/kubernetes/blob/687d759e362b05dcdf11e336e2799704918e048d/pkg/controller/disruption/disruption_test.go#L140)
+  tests PDB MinAvailable, MaxUnavailable, and selector functionality in Disruption controller.
+* Kubectl
+  - [This test](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/generate/versioned/pdb_test.go)
+  tests generation of a PDB objects out of given parameters
+  - [This test](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/create/create_pdb_test.go)
+  tests creation of PDB objects from cmd parameters
+* Scheduler
+  - [This test](https://github.com/kubernetes/kubernetes/blob/ac56bd502ab96696682c66ebdff94b6e52471aa3/test/integration/scheduler/preemption_test.go#L731)
+  tests effects of PDB on preemption (PDB is honored in a best effort way)
+* Eviction integration tests
+  - [These tests](https://github.com/kubernetes/kubernetes/blob/master/test/integration/evictions/evictions_test.go) test eviction logic and its interactions with PDB.
+* Autoscaler
+  - [These tests](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/autoscaling/cluster_size_autoscaling.go) ensure that Autoscaler respects PDB when draining nodes.
+
+#### Needed Tests
+
+As a non-optional feature, there should be a conformance test for
+PodDisruptionBudget.
+
+
+### Graduation Criteria
+
+- [ ] Implement Mutable PDBs
+- [ ] Needs a conformance test
+- [ ] Update documents to reflect the changes
+
+## Implementation History
+
+- PodDisruptionBudget was introduced in Kubernetes 1.4 as an alpha version.
+- PodDisruptionBudget was graduated to beta in Kubernetes 1.5.
+


keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,142 @@
...
+  - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+[Pod Disruption Budget (PDB)](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
+is a Kubernete API that limits the number of pods of a collection that are down simultaneously from voluntary disruptions.
```
liggitt: typo: Kubernetes

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,142 @@
...
+* Making changes to the API fields or their meanings.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+#### Mutable PDBs
+
+A mutable PDB object allows its `MinAvailable` and `MaxUnavailable` fields to be
+modified by clients. Components that use PDB must watch such modifications and
```
liggitt: what about selector fields? is there a good reason to limit which spec fields can be modified?

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,142 @@
...
+
+This feature is implemented by [this PR](https://github.com/kubernetes/kubernetes/pull/69867).
+
+#### PDB in Deployment controller
+
+[Rolling update](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/rolling.go) of Deployment controller reads `MinAvailable` and `MaxUnavailable` from
+the rolling update spec and currently does not take PDB into account. We need to
+change the implementation and use
+`min(PDB.MinAvailable, RollingUpdate.MinAvailable)` and
+`max(PDB.MaxUnavailable, RollingUpdate.MaxUnavailable)` instead.
```
liggitt: Do we want deployments looking at PDB objects or using the evict subresource? Just looking at the PDB does not take into account PDB status for selectors that match across deployments, right?

Respecting PDB could easily lead to a situation in which a deployment could not make progress. Given that, is respecting PDB still desired? If so, describe how that state will be detected, communicated, and/or resolved (automatically or manually).

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,142 @@
...
+* Scheduler
+  - [This test](https://github.com/kubernetes/kubernetes/blob/ac56bd502ab96696682c66ebdff94b6e52471aa3/test/integration/scheduler/preemption_test.go#L731)
+  tests effects of PDB on preemption (PDB is honored in a best effort way)
+* Eviction integration tests
+  - [These tests](https://github.com/kubernetes/kubernetes/blob/master/test/integration/evictions/evictions_test.go) test eviction logic and its interactions with PDB.
+
+#### Needed Tests
+
+If this is considered a non-optional feature, there should be a conformance test
+for it (this needs to be an e2e test tagged as conformance).
```
liggitt: This seems like a good candidate for conformance testing. With an ack from @kubernetes/cncf-conformance-wg I'd make this a more definitive statement

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: Added the selector as well. I am not closely familiar with the internal logic of Disruption controller to say with confidence whether updating the selector (and/or other fields) could cause any issues there, but I don't see a concern after a quick look at the code.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: What I had in mind was to only look at `MinAvailable` and `MaxUnavailable` values and treat those similar to `MinAvailable` and `MaxUnavailable` of a rolling update, without making further changes and without using "evict" subresource, but given your points and reading [some of the old comments](https://github.com/kubernetes/kubernetes/issues/12611#issuecomment-130506236), I think we should drop this requirement.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
timothysc: +1

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,133 @@
...
+- [Proposal](#proposal)
+  - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+## Summary
+
+[Pod Disruption Budget (PDB)](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)
```
timothysc: @bsalamat @liggitt - I've been thinking a lot about this topic when working on ClusterAPI.  

Because the object is very loosely coupled with Pods, and only via label selector, is there any reason why we couldn't have it be more generic and remove the the 'Pod' from the resource kind?  The link via label denotes the resource that it is tied to.  

As a concrete example, we would like to tie disruption budgets to other resources for cluster management. 
/cc @ncdc @detiber

derekwaynecarr: for the other scenarios, is delete the standard definition of disruption?

smarterclayton: Agree with disruption budgets to machines.  There's two orthogonal dimensions and use cases.

derekwaynecarr:
i have no major issue about graduating pdb in its current form.  limitations that we have encountered recently could be handled via a higher level controller writing to a pdb so making them mutable would help a lot.  an example scenario was sizing a pdb relative to the number of control plane machines in a cluster which is easier to do if we support mutability.  this is useful if you use daemonsets for some types of workloads, but still want to minimize potential disruption to the number of control plane machines.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,133 @@
...
+
+PDB API has been stable and is an important feature that allows users to improve
+reliability of their critical workloads. This feature has been in beta for a
+long time. We need to promote to stable version given that we plan to support it
+long term.
+
+### Goals
+
+* Plan to promote PDB API to stable version.
+* Propose changes to the deployment controller to take PDB into account.
```
smarterclayton: Where is "taking it into account" described?

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
detiber: Would just using label selector cause confusion here? Would there be a case where we would want to match both Machines and Pods with the same selector? Or would it be on the consumer to differentiate the type to associate with the label selector?

As to the specific use case around Machines, would it make sense to have the disruption budget exist for the Machines or the underlying Nodes instead? If what we are striving for is to ensure a certain level of capacity is available during an upgrade, then it seems to me that we are more concerned with the Nodes rather than the Machines themselves.

derekwaynecarr: we have protected a pool of machines by running a pod per node with anti-affinity rules setup, and then you can prevent disruption to more than N number of nodes using that pod as a guard.  its a bit of a hack, but it works.  in general, i would prefer we have pod disruption budget, and machine disruption budgets as separate objects, the use cases may vary slightly in practice.  similar to having machine sets and replica sets.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: I removed that section as it turns out to create [complexities](https://github.com/kubernetes/enhancements/pull/904#discussion_r266709445) which are not easy to address. I'll remove this goal.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: Changed wording to make the test a requirement.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: +1 to having different disruption budget for pods and nodes for a few reasons:
1. PDB is enforced through `evict` subresource of Pods which is obviously not applicable to nodes.
1. If we used the same API for both objects, we would need to add a field to the API to specify the type of object that it applies to (node vs pod).
1. Implementation of Node Disruption Budget shares little with that of PDB. So, we won't save much on the coding side by combining the two APIs either.

liggitt: > I would prefer we have pod disruption budget, and machine disruption budgets as separate objects, the use cases may vary slightly in practice. similar to having machine sets and replica sets

I'd tend to agree. The concepts are similar, but the application, audience, and scope are different.

PDB's are namespaced, and write access can be granted to users with write access in a namespace.

Machine or Node disruption budgets would be cluster-scoped, and should only be writeable by the cluster admin.

timothysc: > Machine or Node disruption budgets would be cluster-scoped, and should only be writeable by the cluster admin.

That's not true, they would be namespace scoped in a mgmt cluster. 

FWIW I'm ok if we decide to not rationalize now, but I do think this is an area in the api where there is overlapping concepts and have (N) of the same type is also not good.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,132 @@
...
+* Kubectl
+  - [This test](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/generate/versioned/pdb_test.go)
+  tests generation of a PDB objects out of given parameters
+  - [This test](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/create/create_pdb_test.go)
+  tests creation of PDB objects from cmd parameters
+* Scheduler
+  - [This test](https://github.com/kubernetes/kubernetes/blob/ac56bd502ab96696682c66ebdff94b6e52471aa3/test/integration/scheduler/preemption_test.go#L731)
+  tests effects of PDB on preemption (PDB is honored in a best effort way)
+* Eviction integration tests
+  - [These tests](https://github.com/kubernetes/kubernetes/blob/master/test/integration/evictions/evictions_test.go) test eviction logic and its interactions with PDB.
```
MaciekPytel: A bunch of autoscaling tests: https://k8s-testgrid.appspot.com/sig-autoscaling-cluster-autoscaler#gci-gce-autoscaling&include-filter-by-regex=pdb, code in: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/autoscaling/cluster_size_autoscaling.go

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Risks and Mitigations
+
+We plan to support mutation of PDB objects that didn't exist in previous versions.
+The following needs to be checked and verified before graduation of the API.
+
+- [ ] Ensure that components do not have any logic that relies on immutability
+of PDBs. For example, if component builds a cache of pods that match various
+PDBs, they must add logic to invalid the cache on updates.
+   - Action Item: sweep for in-tree uses to make sure we’re informer driven and
```
MaciekPytel: This should include all repos in kubernetes org, not just k/k. I know Cluster Autoscaler has logic to deal with PDBs (I don't think it will be impacted by this change, but it's worth double checking), I suspect VPA may also use it. Not sure about other projects under kubernetes org.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: > FWIW I'm ok if we decide to not rationalize now,

If we think we may change the API at some point, discussing it now is probably better than after the API is GA'ed.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: Added the test

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
bsalamat: Added

kow3ns:
@janetkuo and @mortent PTAL as well, I'd like your thoughts about this

bsalamat:
ping @janetkuo and @mortent. Is there anything else that you want to see here?
Since PDB is owned by SIG Apps, I would expect someone from your SIG to take the graduation of PDB from here and perform the remaining actions items.

mortent:
I'm looking into adding support for the scale subresource with PDBs, but I'm not sure if this would be a requirement for GA. But I hope this would be ready for the next release of Kubernetes together with mutable PDBs. And then both changes could soak for one release and we can push PDBs to GA in 1.16.

I am interested in taking on the task of getting this to GA.

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
timothysc: We have work-arounds on our side.

kow3ns:
So we've discussed this as something we want to do. It passes the criteria for approval.
/approve

keps/sig-apps/20190318-PodDisruptionBudget-graduation-to-stable.md
```diff
@@ -0,0 +1,134 @@
...
+PDBs, they must add logic to invalid the cache on updates.
+   - Action Item: sweep for in-tree (including kubernetes/* repos) uses to make
+   sure components are informer driven and aren’t assuming immutability.
+- [ ] Check what watches PDBs and make sure there is no performance concerns.
+- [ ] Updating a PDB could cause the state of a cluster to seem incorrect. For
+example, a PDB states that at least 10 replicas of a collection must be alive.
+Kubernetes control plane evicts some of the existing pods, but keeps at least 10
+around. The PDB is updated and states that at least 20 replicas must be kept
+alive. It may appear to an observer that the evictions happened before the PDB 
+update were incorrect, if they don’t notice the PDB update.
```
janetkuo: We already have this problem before PDBs are mutable. 

Immutable PDBs can be deleted and re-created with new spec (20 replicas in this example).

seh:
See kubernetes/kubernetes#95083 for a potential problem with the debatable treatment of the `PodDisruptionBudget.Selector` field.

PR #906, opened 2019-03-19
Kustomize: Generator and Transformer Plugins

pwittrock:


pwittrock:
cc @kow3ns 
cc @janetkuo

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+Notes:
+
+- Kustomize will not execute any plugins for the output Resources at this time.
+- If multiple generators are specified, they are run in the order they are listed.
+
+#### Transforms
+
+Transformers modify existing non-virtual Resources by modifying their fields or replacing them with new Resources.
+
+Transformer plugins are run immediately *before* built-in transformers:
```
Liujingfang1: I don't think we should restrict the transformer plugins to be *before*. They can also be *after* the build-in transformers. Both are useful in certain scenarios. The kustomization.yaml author need to specify whether it is before or after. 

Some *before* transformer plugins I can think about:
- inject a sidecar container to pod template

Some *after* transformer plugins:
- validate the resource configs(the output yaml) according to some specific policy. Since the resource configs are not completely done until performing all the built-in transformers. These validation should be as a post processing step

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+The `team.example.com` binary is invoked with both the `transform.yaml` contents and the Resources passed on STDIN.
+Kustomize replaces its Resources with the emitted Resources.
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+## Design Details
```
anguslees: Are plugins expected to produce stable/repeatable output?

Do the plugins get executed during `--dry-run`?  If not, is there another mode that does execute them but doesn't apply the results?

anguslees: Also, just for my own clarity: if a kustomization.yaml references several remote bases that also use kustomization - I presume each remote base is "expanded" using the generators/transformers specified in that remote kustomization.yaml (but using local plugin executables).  And then the result of each base is concatenated together before passing to the first locally-specified transformer plugin.

Iow, we're building a tree of plugin pipelines here, not just a single linear pipeline.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+
+
+## Drawbacks [optional]
+
+It allows users to do very complex thing with Kustomize.
+
+## Alternatives [optional]
+
+- Imperatively piping generator executables to kubectl apply
+- Writing scripts to invoke generator executables and piping them to kubectl apply
```
anguslees: - Start a new "meta-kustomize" tool that implements all this plugin machinery - and then make the existing kustomize just one of the plugins within that (rather than having the meta version burst from the stomach of the existing code and consume the host - yay aliens metaphors)

pwittrock: This is listed under non-goals - re-implemented kustomize as plugins.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+  - Kustomize directive that accepts 1 Virtual Resource *and* collection of non-Virtual Resources as input
+    and emits non-Virtual Resources as output.
+  - Examples: `commonAnnotations`, `namespace`
+- *Built-In*
+  - Either a Transformer or Generator that is embedded directly in the `kustomization.yaml` rather than
+    as a separate API.
+
+#### Generators
+
+Generators generate 0 or more Resources for some virtual Resource input.
```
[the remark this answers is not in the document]
pwittrock: This is something @monopole also suggested.

IMHO, it is simpler to provide 2 separate focussed and structured patterns through 2 APIs, rather than 1 API that can implement arbitrary patterns.  Mostly this is to simplify ordering and dependency considerations.

However I am open to being convinced someone can show me a solution that addresses dependency and ordering in a way that bounds the complexity users can introduce into the system.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+1. Kustomize adds the emitted Resources to its set of input Resources (e.g. from `resources`, `bases`, etc).
+
+Notes:
+
+- Kustomize will not execute any plugins for the output Resources at this time.
+- If multiple generators are specified, they are run in the order they are listed.
+
+#### Transforms
+
+Transformers modify existing non-virtual Resources by modifying their fields or replacing them with new Resources.
```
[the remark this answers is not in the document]
pwittrock: LMK if this answers any questions:

The `generators` and `transformers` take lists of *Resources*, just like the `resources` field does.

Another way to implement this would be to put what is listed in `generators` and `transformers` under `resources`, have implicitly identify and invoke plugins based on encountering Resources whose group has a transformer or Generator.  I don't really like this approach because it is much hard to debug what is going on in the Kustomization file.

We want to stay as far away as possible from accidentally creating a FSM.

>  should it be in a dryRun mode, where are the chart repos, etc.

We could forward flags to the plugins.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: > Do the plugins get executed during --dry-run

Yes, `kustomize` doesn't have a notion of dry-run.  Could be proposed in a separate KEP.

> ...Iow, we're building a tree of plugin pipelines here, not just a single linear pipeline.

Yes, this is how Kustomize works today.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,389 @@
...
+from a `kustomize.config.k8s.io/v1beta1/Kustomization` virtual Resource.
+
+Generators have 2 components:
+
+1. Generator virtual Resource API definition
+  - By convention, API has a field `generate`
+1. Generator executable plugin that reads the Resources on STDIN and emits generated Resources on STDOUT.
+
+- A Generator is installed by adding the generator executable to
+  `$XDG_CONFIG_HOME/kustomize/plugins/generators/<API group>`.
```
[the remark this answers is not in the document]
pwittrock: Updated

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
pwittrock: > Can Transformers modify/replace/generate virtual Resources too? Why not?

This is something I considered, and don't think we want for v1, but we would consider.

Some thoughts

- We would want to limit this so it wasn't easy for users to create cyclic graphs
- Plugins like `namePrefix` will be hard to make globally idempotent - e.g. re-running them will keep appending.
- Recursing on Virtual Resources isn't necessary within a single Resource Group - since they are handled by the same exec plugin, that plugin can fully expand Virtual Resources internally
- There is dubious need for recursing on Virtual Resources across Resource Groups.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: Hm, maybe this should be a separate pattern - `validators`.  I think we want to stick to a small number of highly structured patterns, rather than 1 very flexible pattern.

keps/sig-cli/20190318-kustomize-generators-transformers.md
anguslees: yes, and that's very reasonable.  I'm suggesting (for completeness, not a serious proposal) that we implement _this_ KEP in a new tool/layer (that isn't existing-kustomize).

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
anguslees: > show me a solution that addresses dependency and ordering

The unified version is just a single ordered list of "processors" (or whatever we call it) - the same as you already need for transformers.  The only hard bit is how you add in the current built-ins, and you already have this problem too (as raised elsewhere).

So the solutions are the same:
- You can just declare for each builtin whether it always gets executed first (builtin generators) or last (builtin transforms).
- Or alternatively, you can give them some representation in the explicit list and allow users to order the transforms explicitly (eg: `transforms: ["scale-resources.yaml", "builtin:nameSuffix"]`).

The only difference is that generator plugins are now listed in the transformers list too, instead of being in a separate unordered list.

I don't think it introduces any new complexity (and indeed removes some for both implementer and user) - they're just a transform that does "append" instead of pure generation.

keps/sig-cli/20190318-kustomize-generators-transformers.md
anguslees: > Yes, kustomize doesn't have a notion of dry-run

kubectl does though.  So if we're proposing that these features become part of the kustomization.yaml support within kubectl, then we need to decide as part of implementing this KEP ...

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
anguslees: Re earlier comment: agreed, I think it's desirable to have kustomisation.yaml only able to describe a fixed transform pipeline.  That allows visualisation and introspection without needing to actually execute the pipeline (at "only" the cost of removing meta-programming).

Fwiw, this means the "virtual resources" don't need to *actually* be resources (they don't need apiVersion/kind/etc structure, they just need to be YAML), since we've just declared they are never acted on as resources.  (I don't expect that to change anything, but I'm just working through the category implications...)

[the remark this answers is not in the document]
pwittrock: > Fwiw, this means the "virtual resources" don't need to actually be resources

Yes this is true.  FWIW, it is the pattern we've follow for both kubeconfig and kustomization.yaml.  It does provide a solution to things we would probably want to tackle - such as the generators detecting if they support a specific version of a Virtual Resource (and failing if they don't), namespacing generators by group, etc.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: Good point.

Essentially you are asking to clarify that bases can define `generator` and `transformer` *Virtual Resources*, and these would be lazily invoked when kustomize was run.  Is this correct?

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: Ah got it.  Yeah I can add that.

keps/sig-cli/20190318-kustomize-generators-transformers.md
anguslees: I think we now have two useful versions of --dry-run, and I was just looking for a statement which one we're planning to do here:

- Run no plugins at all:
  No plugins are executed, but `kubectl create --dry-run -o yaml` does not show the objects/transformations that will actually exist without `--dry-run`
- Run plugins but don't push to server:
  Plugins still get executed, and are just assumed to not have permanent undesirable side-effects (which is why we should document it here).
  `kubectl create --dry-run -o yaml` shows something similar to the objects that would be created without `--dry-run` (assuming plugins are mostly repeatable).

There are issues with both, so it would be good to discuss/agree on it at this stage rather than be surprised by the implementation.

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
pwittrock: Appreciated.

WDYT:
At a lower conceptual level and in the implementation we can have the notion of *Resource Functions*.  *Resource Functions* are executables that accept an arbitrary number of Resources on STDIN and emit an arbitrary number of Resources on STDOUT.  
TBD: Flags and environment variables are propagated to *Resource Functions*.

At a higher level - *Kustomize* execution may be broken up into 3 distinct phases:

- generation
- transformation
- validation (doesn't exist yet)

Users may extend Kustomize by providing specific types of *Resource Functions* which to implement specific patterns based on the phase they are executed during `generator`, `transformer`, `validator`, etc.

This prevents users from developing a Turing complete pipeline on top of *Resource Functions*, for which they can use other tools specifically targeted at pipelines.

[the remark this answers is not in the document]
anguslees: sgtm.

> This prevents users from developing a Turing complete pipeline on top of Resource Functions

This is where humanity will make its final stand against the robot invasion :stuck_out_tongue:

pwittrock:
@anguslees @monopole 

Thinking about this more, some functions will need to be *both* generators and transformers - e.g. a secretGenerator will generate a new name, and then need to transform objects by modifying the name.

From an implementation perspective, I think the notion of a *Resource Function* is correct.
From a kustomize API perspective, I think having discrete patterns is more clear.

FWIW, I can imagine 2 distinct styles:

- The kustomization.yaml style - a bunch of orthogonal operations packed into a single Resource
- The k8s API style - discrete modular Resources

The former works better for simple Transformations, whereas the ladder works better for complex Generators that may have many fields set.

WDYT of this implementation:

- Plugins go under the dir `$XDG_CONFIG_HOME/kustomize/plugins/functions/` and are named `<group>`
- Plugins may have 3 different sub commands
  - `$ pluggin.example.com generate`
  - `$ pluggin.example.com transform`
  - `$ pluggin.example.com canonicalize`
 
**Generate**: Accepts *Generator Virtual Resources* only.  Emits Concrete Resources.
- e.g. bases, secretGenerator
**Transform**: Accepts all Concrete Resources + *Generator & Transformer Virtual Resources* for the matching group.  Never adds or deletes, only updates.  Generators may also be implicitly transformers.
- e.g. secretGenerator, commonAnnotations
**Canonicalize**: Reorders Resources or Fields.  Never changes values.  May validate.
- e.g. sort (implicit)

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
pwittrock: > This is where humanity will make its final stand against the robot invasion 😛

I have always viewed that as my true calling.

pwittrock:
The drawback of this approach is that things like overlays won't work on virtual resources, so if you want to tweak the virtual resources per-environment or cluster, it won't work...

Kustomize could special case a couple things, such as applying overlays.

pwittrock:
@anguslees Probably worth leaving this provisional or open for a bit.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,395 @@
...
+  - Examples: `commonAnnotations`, `namespace`
+- *Built-In*
+  - Either a Transformer or Generator that is embedded directly in the `kustomization.yaml` rather than
+    as a separate API.
+
+#### Generators
+
+Generators generate 0 or more Resources for some virtual Resource input.
+
+Generator plugins are run immediately after built-in generators:
```
sethpollack: Why not just run them in the order that they are listed?

pwittrock: there shouldn't be an order to fields.

pwittrock:
@anguslees Actually, lets merge this as is "provisional" and iterate on it so that other folks can contribute edits.

keps/sig-cli/20190318-kustomize-generators-transformers.md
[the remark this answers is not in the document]
barney-s: curious how would this be/notbe turing complete

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,395 @@
...
+however pipes are an imperative technique.  This KEP proposes a declarative approach for composing tools built
+in the Kubernetes ecosystem or bespoke tools developed by organizations that are tightly coupled to their
+environments.
+
+[kube-gen]: https://events.linuxfoundation.org/wp-content/uploads/2017/12/Services-for-All-How-To-Empower-Engineers-with-Kubernetes-Melanie-Cebula-Airbnb.pdf
+
+### Goals
+
+- Allow config authoring solutions developed in the ecosystem to be integrated into kubectl as declarative
+  Generators and Transformers
```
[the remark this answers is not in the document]
pwittrock: kubectl plugins aren't declarative and don't plugin to the `-f` or `-k` flags

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,395 @@
...
+
+- Rewrite Kustomize on top of plugins
+  - This should be possible, but isn't a problem we need to solve right now
+- Develop a market place of any sort
+  - The ecosystem solutions themselves will have market places.  This allows those ecosystem tools to
+    plug their market places into kubectl via Kustomize.
+
+## Proposal
+
+Introduce 2 new types of `exec` plugins:
```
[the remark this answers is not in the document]
pwittrock: Containers

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,395 @@
...
+
+- Transformers
+- Generators
+
+#### Terms
+
+- *Virtual Resource*
```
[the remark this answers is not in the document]
pwittrock: > vs Virtual API which is a "domain" of a CRD ?

I am not sure I understand the difference.  This could be considered similar to component config.

pwittrock:
PTAL

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: I changed this to say plugins must behave as if they are always run in dry-run

anguslees:
Re https://github.com/kubernetes/enhancements/pull/906#issuecomment-474896368

I think the 3 stages is a fine way to clarify/restrict the various roles.  Importantly "transform" plugins need to be ordered (not commutative).  "Canonicalize" also needs to be ordered (not commutative), unless we rename it to "validate", keep it strictly read-only, and make sorting an implicit builtin last step.

> e.g. a secretGenerator will generate a new name, and then need to transform objects by modifying the name

I don't understand this bit at all.. Isn't this just an example of exactly why we need to order the secretGenerator before the name transformer(s) (and thus secretGenerator should be strictly a "generator")?

> Transform: ... Never adds or deletes, only updates.

Why limit transforms to only updating?  In particular, how can we remove an object (or replace one with another), if not during the transform stage?

pwittrock:
> I don't understand this bit at all.. Isn't this just an example of exactly why we need to order the secretGenerator before the name transformer(s) (and thus secretGenerator should be strictly a "generator")?

Consider the following case

- Deployment Generator generates Deployment `Foo` which refers to a generated Secret `bar`
- Secret Generator generates Secret for `bar`,  with generated name - `bar-xyz` (like the current SecretGenerator does)
- Secret Generator transforms `Foo` reference to `bar-xyz`
- Name-Suffix Transformer changes the Secret name and reference to `bar-xyz-abc`

Generators that use generated names *must* be able to transform Resource all of the Resources that reference them.

> Why limit transforms to only updating? In particular, how can we remove an object (or replace one with another), if not during the transform stage?

I'd like to know more about the use case before explicitly supporting it.  I suspect this ability will create complex interactions between transformers, and will make it harder to map the input files to what gets created.  e.g. if a Transformer replaces ConfigMaps with Secrets, it will make it hard to look at the Resource Config and know that the ConfigMap is actually a Secret.

pwittrock:
Updated the PR.  PTAL.

Also marked it implementable since I think we are close to something we could go out with.

keps/sig-cli/20190318-kustomize-generators-transformers.md
anguslees: > Plugins should probably accept the dry-run flag

The current KEP doesn't have any notion of "common arguments" provided by kustomize itself (rather than from the user-provided virtual resource).  I'm guessing in the future we will want something like this.

I think we will have to use environment variables for this (which is fine) unless we want to change this KEP now to include an "envelope metadata" wrapper around the virtual resource provided on stdin.  We also don't have a versioned-api at all (other than changing the install location) to allow us to change the plugin communication in incompatible ways in future.

I think we can work around that, so mostly just raising it as a future-cost.

anguslees:
> - Secret Generator generates Secret for bar, with generated name bar-xyz
> - Secret Generator transforms Foo reference to bar-xyz

Aha, I naively assumed the secret generator just generated a secret! :stuck_out_tongue:

Thanks, I understand your original comment now.  I see secretGenerator currently does two roles:
1. (generator) Generate the Secret
2. (transformer) Rename the Secret (_and references to it_) to something unique-ish.

And yes, I agree allowing a single plugin executable to fulfil both roles would be useful in this example.  Fwiw the downside of implementing this example as a single plugin though (rather than two separate plugins), is that we can't re-use (2) for Secrets that came from other sources, like literal YAML files.  Maybe we should change secretGenerator so that it does _not_ do both these roles?

This is in the weeds about how we might implement a hypothetical secretGenerator-equivalent-plugin though, and not directly relevant to this KEP.

I note however: (2) is not allowed under the proposed rules because "transformers can't add or delete objects".  :scream:  I still recommend we drop this artificial restriction.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+
+The order of plugin execution is arbitrary.
+
+##### Restrictions
+
+For the initial iteration, transformers cannot add/remove Resources, or change their names/namespaces.
+The ability doing so would be significantly make complex ordering interactions much more likely.  E.g. transformers
+would need to keep and propagate transformations from other transformers.
+
+We are prioritizing a restrictive but predictable API over a powerful but complex one.
```
anguslees: (I've already mentioned it several times before, so this will be my last:  You can't usefully use multiple transformers without being able to control ordering.  Non-commutative transformations are definitely not limited to just changing name/namespace and the proposed "arbitrary ordering" does not lead to predictable output.

I understand that you really don't want to have any ordered lists for some reason, so I'll just keep this passive-aggressive "I told you so" comment here to point back to in the future :stuck_out_tongue: )

pwittrock: +1.  FWIW, I am not against this, I just don't have enough data to understand the requires - e.g.  concrete expected use cases vs conceptual academic use cases.

When this feature has enough adoption that we have this data, I will look for a proposal that uses them to define requirements and come up with a design.

Since supporting explicit ordering is backwards compatible with arbitrary ordering - it is simpler to add in the future, and I am comfortable deferring the decision the specific API that would support this - e.g. is it:
- an ordered list of transformer plugins only
- an ordered list of transformer plugins and generator plugins only
- an absolute ordering of both built-in and plugin transformers and generators
- an acyclic dependency graph
- an cyclic dependency graph
- additional transformation phases
- automatic ordering based on transformer duck-typing or interfaces
- some combination of the above

pwittrock:
> I note however: (2) is not allowed under the proposed rules because "transformers can't add or delete objects". 😱 I still recommend we drop this artificial restriction.

The Secret is never renamed.  The Generator generates a Secret with a generated name.  The transformer normalizes references to the Secret generated name.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: SGTM.  We will need to figure out which ones will be required for the plugin to support (these probably need to be set as flags so the command fails), vs are optional (these probably need to be env vars so the command doesn't value).  Plugins will probably want to use something like Viper to read configuration from flags and env vars arbitrarily.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+- `bases`
+- `secretGenerators`
+- `configMapGenerators`
+
+Example Built-In Transforms:
+
+- `commonAnnotations`
+- `commonLabels`
+- `images`
+- `namePrefix` # not supported in iteration 1 due to transformation restrictions
```
droot: `namePrefix` is supported.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+- `nameSuffix` # not supported in iteration 1 due to transformation restrictions
+- `patches`
+- `namespace` # not supported in iteration 1 due to transformation restrictions
```
droot: `namespace` is also supported.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+
+
+## Motivation
+
+1. Enable organizations to develop their own authoring solutions bespoke to their requirements, and to integrate
+   those solutions with kustomize and kubectl.
+1. Enable tools and authoring solutions developed by the ecosystem to natively integrate with kustomize and kubectl
+   commands.
+1. Enable Kustomize power users to augment Kustomize with their own *Transformers* and *Generators*.
```
droot: +1

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+- Develop a market place of any sort
+  - The ecosystem solutions themselves will have market places.  This allows those ecosystem tools to
+    plug their market places into kubectl via Kustomize.
+
+## Proposal
+
+Introduce an executable plugin that has 2 subcommands.
+
+- `$ team.example.com generate`
+- `$ team.example.com transform`
```
droot: `team.example.com` is a binary that is suppose to exist `~/kustomize/plugins` ?

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+
+- *Virtual Resource*
+  - Resource that is not a server-side Kubernetes API object, but used to configure tools
+  - Examples: kubeconfig, kustomization
+- *Generator* Kustomize directive
+  - Accepts Virtual Resources as input
+  - Emits generated non-Virtual Resources as output
+  - Examples: `configMapGenerator`, `secretGenerator`
+- *Transformer* Kustomize directive
+  - Accepts Virtual Resources
```
droot: I am a bit unsure about this requirement. Not being able to think of a use-case.

keps/sig-cli/20190318-kustomize-generators-transformers.md
```diff
@@ -0,0 +1,435 @@
...
+
+```yaml
+groupVersion: helm.kustomize.io/v1beta1
+kind: Chart
+generate:
+    chart: ./path/to/chart/from/kustomization.yaml
+    values:
+      k1: v1
+      k2: v2
+```
```
droot: plugin can read/write arbitrary files for ex. values.yaml ? (otherwise it will need some sandboxing mechanism).

pwittrock: Plugins are executables, and can be:

- Malicious
- Exploitable

We might be able to provide some sandboxing by running them in a container...

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: right

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: no it isn't, it changes the name of the resource - e.g. adds and deletes.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: nope, this changes the resource.

keps/sig-cli/20190318-kustomize-generators-transformers.md
pwittrock: Secret Generator-Transformer

monopole:
Continued in #931

PR #912, opened 2019-03-22
KEP for critical container feature

CsatariGergely:
Signed-off-by: Gergely Csatari <gergely.csatari@nokia.com>

### DIFF
--- a/keps/sig-node/20190333-critical-container.md
+++ b/keps/sig-node/20190333-critical-container.md
@@ -14,7 +14,7 @@ approvers:
   - "@oscardoe"
 editor: "@CsatariGergely"
 creation-date: 2019-03-22
-last-updated: 2019-03-22
+last-updated: 2019-04-26
 status: implementable
 see-also:
 replaces:
@@ -95,6 +95,7 @@ This proposal aims to the following:
 
 This proposal defers the following:
  - Restart all of the containers in a Pod without destroying the whole Pod (and preserving the emptyDir volumes)
+ - Basic lifecycle management of Pods, like [Sidecar KEP API implementation](https://github.com/kubernetes/enhancements/pull/912) or [Sidecar Containers](https://github.com/kubernetes/enhancements/pull/912)
 
 ## Proposal
 
@@ -121,7 +122,7 @@ spec:
 ### Implementation Details/Notes/Constraints
 
 - It is an open question if the whole feature can be implemented in Kubelet or other components are also affected
-- The existing terminate and create procedures of the Pod donr by Kubelet should be used to implement the replacement of the Pod
+- The existing terminate and create procedures of the Pod done by Kubelet should be used to implement the replacement of the Pod
 
 ### Risks and Mitigations
 
--- a/keps/sig-node/20190333-critical-container.md
+++ b/keps/sig-node/20190333-critical-container.md
@@ -1,0 +1,272 @@
+---
+title: Replace a pod when a critical container fails
+authors:
+  - "@CsatariGergely"
+  - "@libesz"
+  - "@janosi"
+owning-sig: sig-node
+participating-sigs:
+reviewers:
+  - TBD
+  - "@alicedoe"
+approvers:
+  - TBD
+  - "@oscardoe"
+editor: "@CsatariGergely"
+creation-date: 2019-03-22
+last-updated: 2019-03-22
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Replace a pod when a critical container fails
+## Table of Contents
+
+A table of contents is helpful for quickly jumping to sections of a KEP and for highlighting any additional information provided beyond the standard KEP template.
+[Tools for generating][] a table of contents from markdown are available.
+
+- [Replace a pod when a critical container fails](#replace-a-pod-when-a-critical-container-fails)
+  - [Table of Contents](#table-of-contents)
+  - [Release Signoff Checklist](#release-signoff-checklist)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Design Details](#design-details)
+    - [Test Plan](#test-plan)
+    - [Graduation Criteria](#graduation-criteria)
+      - [Examples](#examples)
+        - [Alpha -> Beta Graduation](#alpha---beta-graduation)
+        - [Beta -> GA Graduation](#beta---ga-graduation)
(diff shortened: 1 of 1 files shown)


keps/sig-node/20190333-critical-container.md
```diff
@@ -0,0 +1,272 @@
...
+
+The motivation section can optionally provide links to [experience reports][] to demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+
+### Goals
+
+This proposal aims to the following:
+ - Possibility to mark one or more containers in a Pod as critical in the PodSpec
+ - Replace the Pod if a critical container`s container restart count reaches a threshold
```
derekwaynecarr: can you elaborate on what is meant by replace?  is the request that the pod is evicted from the node and failed?

keps/sig-node/20190333-critical-container.md
```diff
@@ -0,0 +1,272 @@
...
+
+This proposal defers the following:
+ - Restart all of the containers in a Pod without destroying the whole Pod
+
+## Proposal
+
+### User Stories
+
+1. I as an user of Kubernetes I would like to indicate the criticality of a container in the `escalations` section of the PodSpec/Container
+1. I as an user would like to specify the number of container restarts which triggers a Pod replacement
```
derekwaynecarr: can you clarify what is meant by replacement?  a pod's fate is bound to the node it executes against.

libesz:
@derekwaynecarr You can find some use-cases here: https://github.com/kubernetes/kubernetes/issues/40908
Pod replacement in my mind is the process what is done when all the container including the infra (pause) is terminated and re-created from the PodSpecs.
The idea was that we don't want to introduce a new procedure to "restart" a pod in-place (which i.e. preserves the emptyDir volumes, etc.), instead use the existing terminate and create procedures of the Pod, done by kubelet. Hope our assumptions are correct 🤞 .

CsatariGergely:
@derekwaynecarr Thanks for the comments. To join to the sig-node meeting this week is a bit difficult from CET timezone. Next week I will be closer timezone wise and I will try to join.

CsatariGergely:
Notes from sig-node meeting at 02.04.2019:

- Sidecar [containers KEP](https://github.com/kubernetes/enhancements/pull/919
) is doing similar in a sense that some containers are differently handled from the pods licecyle point of view
- We should check if this provides a solution to our problem

libesz:
> 
> 
> Notes from sig-node meeting at 02.04.2019:
> 
>     * Sidecar containers [KEP1](https://github.com/kubernetes/enhancements/pull/919), [KEP2](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/sidecarcontainers.md) is doing similar in a sense that some containers are differently handled from the pods lifececyle point of view
> 
>     * We should check if this provides a solution to our problem

Just checked the other KEPs. I believe those are covering different use-cases around lifecycle management, while this one wants to cover failure/recovery scenarios. I would consider this one as independent from specification/implementation PoV.

InAnimaTe:
This is a great KEP and I'd love to see it get approved. @CsatariGergely @derekwaynecarr what specifically is being waited for prior to approval?

CsatariGergely:
Thanks @InAnimaTe :) I would be also interested to figure out what is needed to approve this. I'm available in Barcelona to discuss if that helps.

bitva77:
/remove-lifecycle stale

We've recently run across a potential bug in dotnet core containers that an enhancement like this would help us workaround downtime in our service (restarting containers is not enough - the pod needs to be restarted.  I suspect an issue more with Docker and dotnet).

byxorna:
This would be an amazingly useful feature for usecases where the pod health is dependent on the health of all component containers - a failure in one container invalidates the health of all others in the pod (yes, i know that isnt ideal SOA design, but legacy is going to legacy). I have tooling currently emulates this behavior external to k8s (via a k8s aware init replacement), but it is kludgy and would be much nicer if this support was mainlined :)

bitva77:
/remove-lifecycle

still love to see this  :)

InAnimaTe:
I think myself and others would still really like to see this get Merged in. Is there really anything else needed here to get the ball rolling? Can I help in any way like with documentation or design or something? I'd love to chat more about what this looks like and why it should exist.

thockin:
This overlaps a lot with other pod-lifecycle issues.  I think we need  holistic review of the topic.

https://github.com/kubernetes/kubernetes/issues/88886
https://github.com/kubernetes/kubernetes/issues/87801
https://github.com/kubernetes/kubernetes/issues/87016
https://github.com/kubernetes/kubernetes/issues/88257
https://github.com/kubernetes/kubernetes/issues/65502
https://github.com/kubernetes/kubernetes/pull/80744
https://github.com/kubernetes/kubernetes/issues/40908
https://github.com/kubernetes/kubernetes/issues/25908
The pod-restart one which I can't find right now
and so forth

@derekwaynecarr has expressed concerns here and the more I look, the more I agree.

bitva77:
/remove-lifecycle stale

By default, just restart the container as it does today.  But an option to replace the pod would be super.

thockin:
Hi,

I know this KEP has languished.  IMO this needs to be part of a larger pod-lifecycle effort.  I think it is likely to be highly entangled with a bunch of other proposals and ideas, and we should probably not be patching each one individually.

e.g. https://github.com/kubernetes/community/pull/2342  https://github.com/kubernetes/kubernetes/issues/25908  and others that I can't find links for right now.

I do support the idea expressed here, but I don't think we should do it piecemeal.

cnmcavoy:
@thockin I saw your comment about this needing a larger review of the pod-lifecycle. Do you have any suggestions for what the next steps for such an effort would be?

My use-case for this is that we have a service which measures the time it takes an ingress to discover the pod the service is running in and reconfigure itself to serve traffic (e.g nginx reload). After the pod detects the ingress is up to date, it reports the metric and kills itself, so it can be rescheduled with a new pod ip. Unfortunately there is not a clean way to do this behavior currently: if a container exits abnormally, kubelet restarts the container in place and the pod retains its ip/network state. As a terrible workaround, we have the service exit by exceeding it's ephemeral disk quota, which causes kubelet to evict, delete the pod, then reschedule it. This KEP would allow us to exit cleanly and let kubelet restart the entire pod.

I'm interested in this enhancement and want to contribute, but It is unclear to me what the next steps are.

dbsanfte:
It feels like being able to set 'restartPolicy: Never' on a ReplicaSet container (and have it be honoured) is a cleaner solution to the use case presented in the parent issue, and as a bonus it requires less code change, and requires no spec change.

This is already logged here if there's any interest:

https://github.com/kubernetes/kubernetes/issues/24725

ehashman:
Hi all,

This PR has not seen any updates in nearly two years, and I see there are some concerns about the approach. The KEP templates have changed pretty significantly since this PR was opened, so it is not in a state where it is mergeable. Hence, I am going to close this, and if someone has time to pick up this idea and ferry it through the KEP process and propose a new PR, that would be much appreciated.

PR #920, opened 2019-03-26
[RFC] Adding an HA solution to storage migrator KEP

caesarxuchao:
/sig api-machinery
/assign @yliaog
/cc @deads2k @lavalamp

### DIFF
--- a/keps/sig-api-machinery/storage-migration-auto-trigger.md
+++ b/keps/sig-api-machinery/storage-migration-auto-trigger.md
@@ -20,8 +20,12 @@ status: provisional
 * [API design](#api-design)
 * [Storage migration triggering controller](#storage-migration-triggering-controller)
 * [Implications to cluster operators](#implications-to-cluster-operators)
-* [Lifecycle of a StorageState object](#life-cycle-of-a-storagestate-object)
-* [Future work: HA clusters](#future-work-ha-clusters)
+* [Life-cycle of a StorageState object](#life-cycle-of-a-storagestate-object)
+* [Extended API proposal for HA cluster](#extended-api-proposal-for-ha-cluster)
+   * [Extending the storage state API](#extending-the-storage-state-api)
+   * [Extending the work flow of the triggering controller](#extending-the-work-flow-of-the-triggering-controller)
+   * [Case walkthrough](#case-walkthrough)
+   * [Pros and Cons](#pros-and-cons)
 * [Future work: persisted discovery document](#future-work-persisted-discovery-document)
 
 ## Goal
@@ -195,27 +199,6 @@ back to 1.14. The migration triggering controller sets the
 storageState.status.currentStorageVersionHash to batch/v2alpha1 and recreates the
 storageVersionMigration object to migrate to v2alpha1.
 
-## Future work: HA clusters
-
-When the masters of an HA cluster undergoes rolling upgrade, the storage
-versions might be configured differently in different API servers. Storage
-migration shouldn't start until the rolling upgrade is done.
-
-Ideally, for an HA cluster undergoing rolling upgrade, the following should
-happen for each resource: 
-* the discovery document should expose the lack of consensus by listing all
-  storage version hashes supported by different API servers.
-* the storage migration triggering controller should
-  * add all listed storage version hashes to
-    storageState.status.persistedStorageVersionHashes.
-  * set storageState.status.currentStorageVersionHash to `Unknown`.
-  * delete any in progress migration by deleting existing
-    storageVersionMigrations.
-
-After the API server rolling upgrade is done, the discovery document would
-expose the agreed storage version hashes, and the storage migration triggering
-controller will resume the work described in the [previous section][].
-
 [previous section]:storage-migration-triggering-controller
 
 How to develop a convergence mechanism for HA masters is beyond the scope of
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+    triggering controller a chance to report the hash.
+
+### Case walkthrough
+
+Case 1. Assuming the HA cluster consists of 3 apiservers. The cluster admin
+rolling upgrades them. The old storage version is v1, and becomes v2 after
+upgrades. The apiserver 3 fails after upgrades, so the cluster admin rolls it
+back at the 22nd minute.
+
+| time (min) | events                                                                                                                             | apiserver 1 sv<sup>1</sup>| apiserver 2 sv | apiserver 3 sv | CurrentSV<sup>2</sup> | PersistedSV<sup>3</sup> |
```
caesarxuchao: This table is much easier to review after rendering.

keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+3. PersistedSV: persistedStorageVersionHashes.
+
+### Pros and Cons
+Pros:
+1. The API changes are limited to the storage migration API, which is only used
+   by the migrator, so we have more flexibility to experiment.
+
+Cons:
+1. There are two assumptions in this design that might not hold in the future
+    * there is a triggering controller co-locates with every apiserver that
```
lavalamp: I came here to say this :)

You might want to consider mentioning this much earlier in the document.

keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+      consists the HA master.
+    * the triggering controller only talks with the local apiserver.
```
lavalamp: If we ever fix apiserver to coordinate internally (e.g. publish a constant discovery doc), will that break this storage migrator design?

keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+upgrades. The apiserver 3 fails after upgrades, so the cluster admin rolls it
+back at the 22nd minute.
+
+| time (min) | events                                                                                                                             | apiserver 1 sv<sup>1</sup>| apiserver 2 sv | apiserver 3 sv | CurrentSV<sup>2</sup> | PersistedSV<sup>3</sup> |
+|------------|------------------------------------------------------------------------------------------------------------------------------------|---------------------------|----------------|----------------|-----------------------|-------------------------|
+| 0          |                                                                                                                                    | v1                        | v1             | v1             | v1, v1, v1            | v1                      |
+| 3          | The cluster admin starts rolling upgrades from apiserer 1.                                                                         | v2                        | v1             | v1             | v1, v1, v1            | v1                      |
```
yliaog: s/apiserer/apiserver/

keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+  * **AND** the lastHeartbeatTime in all items of the CurrentStorageVersionHashes
+  are no older than 1 polling period (10 mins)
+  * **AND** the hashes in currentStorageVersionHashes have been identical in the
+    past two polls. In other words, the hashes have been held identical for at
+    least one polling period (10 minutes). The purpose is to give newly deployed
+    triggering controller a chance to report the hash.
+
+### Case walkthrough
+
+Case 1. Assuming the HA cluster consists of 3 apiservers. The cluster admin
```
yliaog: one interesting case is: normally the HA cluster has three apiservers, but for a prolonged period of time, one apisever keeps crashing, so essentially only two apiservers are working.

keps/sig-api-machinery/storage-migration-auto-trigger.md
caesarxuchao: Not sure what you meant by "constant discovery doc".

I think it depends on how the apiservers are fixed.
* If HA apiservers can flip storage versions at the same time, then no change is required to the storage migrator.
* If HA apiservers show a consistent discovery doc, and marks the storage version as "undefined" when there is no consensus, then the storage migrator needs to be updated to handle the "undefined" state.

In both cases, this design doesn't add additional difficulty.

keps/sig-api-machinery/storage-migration-auto-trigger.md
```diff
@@ -225,6 +208,137 @@ manually, e.g., deploying the [initializer][] manually.
...
+its local storage version hash. Specifically, every 10 minutes, each instance of
+the triggering controller 
+* fetches the discovery doc from the local apiserver.
+* adds the storage version hash to PersistedStorageVersionHashes if the local
+  storage version hash is not present there yet.
+* updates (or creates) the element in status.storageVersionhashes with the
+  matching reporterID, recording the latest currentStorageVersionHash and
+  lastHeartbeatTime.
+* removes storageVersionHashes whose lastHeartbeatTime are more than 2 polling
+  periods (20 mins) old. This garbage collection is necessary because the
```
wojtek-t: Is 2 polling periods enough? If the controller dies immediately before updating the status (or didn't manage to update it due to some other issues), then that kind of introduces a race between next try and GC.
As an example, we're marking node as NotReady 40s after now heartbeat, whereas heartbeat period is 10s (so 4x heartbeat time).
I would suggest increasing 2x to something larger (at least 3x, if not 4x) to avoid strange situations.

caesarxuchao: Yeah, increasing the GC grace period is reasonable. Increased to 3x.

keps/sig-api-machinery/storage-migration-auto-trigger.md
caesarxuchao: Yeah, that's worth documenting. I'll add it if we decide to pursue this design.

keps/sig-api-machinery/storage-migration-auto-trigger.md
caesarxuchao: Moved to the beginning.

PR #924, opened 2019-03-27
Add KEP for avoid serializing the same object multiple times

wojtek-t:


keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,258 @@
...
+
+- Change overall architecture of the system, by changing what data is being
+read/watched by different components.
+
+## Proposal
+
+This proposal does not introduce any user-visible changes - the proposed changes
+are purely implementational detail of kube-apiserver.
+
+We propose introducing an internal type implementing `runtime.Object` interface,
```
lavalamp: Can you add discussion of alternatives? E.g., what about keeping an explicit LRU serialization cache where the key is type & resource version; and modifying the serialization path logic to optionally take such a cache.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,258 @@
...
+for regular GET/LIST operations - reading the same object N times will result
+in serializing that N times independently.
+
+This proposal presents a solution for that problem.
+
+[#75294]: https://github.com/kubernetes/kubernetes/issues/75294
+
+### Goals
+
+- Reduce load on kube-apiserver, by not serializing the same object multiple
```
lavalamp: This makes a big assumption about our code, which is that our serialization is idempotent. In particular I can imagine some paths getting e.g. a self link wrong or different. What can we do to ensure that all serialization entrypoints produce the exact same serialized bytes?

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: Actually there could be a cache per resource, so the key just needs to be RV.

Keeping the serialization cache and watch cache separate would allow one to enable the serialization cache without turning on the watch cache.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,258 @@
...
+read/watched by different components.
+
+## Proposal
+
+This proposal does not introduce any user-visible changes - the proposed changes
+are purely implementational detail of kube-apiserver.
+
+We propose introducing an internal type implementing `runtime.Object` interface,
+that is encapsulating the original object and is additionally able to accumulate
+its serialized versions. It will look like this:
```
lavalamp: Another idea for an alternative: modify / add to the serialization code such that
* proto objects keep the raw bytes; a metadata accessor is generated that either reads the raw proto bytes or deserializes only metadata.
* Setting self link and RV shuffles bytes around rather than decoding and re-encoding.

This would be more work but I'm willing to bet that the performance gains would be 2x as big. Additionally, this should save memory overall rather than consume additional memory, since the deserialized form never needs to be produced.

I also feel like this would be less invasive--I can probably live with the design currently written in the doc, but I'm not excited about the distributed implicit cache, it adds a lot of non-local state.

keps/sig-api-machinery/20190329-less-object-serializations.md
deads2k: > In particular I can imagine some paths getting e.g. a self link wrong or different

Not to derail your general point, but do we think anyone has actually needed and used the self-link.  They were broken for many release in a row a while back....

deads2k: @wojtek-t how come we made a deserialization cache, but never a serialization cache?

keps/sig-api-machinery/20190329-less-object-serializations.md
deads2k: > Can you add discussion of alternatives? E.g., what about keeping an explicit LRU serialization cache where the key is type & resource version; and modifying the serialization path logic to optionally take such a cache.

Heh.  I had the same question above :)

keps/sig-api-machinery/20190329-less-object-serializations.md
deads2k: > Another idea for an alternative: modify / add to the serialization code such that
> 
> * proto objects keep the raw bytes; a metadata accessor is generated that either reads the raw proto bytes or deserializes only metadata.
> * Setting self link and RV shuffles bytes around rather than decoding and re-encoding.

This sounds similar to the smart objects discussion we've had in the past.  I'm interested in the idea of smart objects that only serialize or deserialize on demand, but the effort is significant.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,258 @@
...
+- first check if such serialization has already been done, and if so just
+  returning the cached result
+- otherwise performing the serialization, caching the result and returning
+  it to the called
+
+With that, we will need to modify `versioning.codec` so that when encoding
+`CachingObject`, it will perform necessary conversion if needed to the
+`CachingVersionedObject`.
+
+Now the question is what is responsible for creating `CachingObject`.
```
deads2k: I think I prefer the smart object idea over trying to externally manage a different cache.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,258 @@
...
+- XX% less memory allocations
+
+[POC]: https://github.com/kubernetes/kubernetes/pull/60067
+[#75294#comment-472728088]: https://github.com/kubernetes/kubernetes/issues/75294#issuecomment-472728088
+
+### Risks and Mitigations
+
+Given this proposal doesn't introduce any user visible change, the only risk
+are bugs in implementation. Fortunately, the serialization code is widely
+used by all end-to-end tests, so any bugs should be catched by those or
```
deads2k: Sorry, I need to disagree here.  I think that our experience with the watch cache is that subtle bugs can exist for a long time across multiple releases in spite of usage in tests and by end users.

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: Good question, I'm open to removing it, but it would be a significant undertaking...

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: smart object meaning the cached object? Or the proto-aware object I described?

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: I have to agree with David here :/

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: > This makes a big assumption about our code, which is that our serialization is idempotent.

I'm not sure I understand this question - object once serialized, can't be serialized again (it's no longer object, but []byte). Can you clarify?

Regarding selflinks - they are quite painful indeed (it's the only think we set further in the stack), so if we could get rid of those, it would make things easier.

wojtek-t: Deserialization cachen was added purely in etcd v2 code in the very early days, when there is no watchcache, which meant that every most of watches (like kubelets watching their pods) was watching huge stream of objects, deserializing all of them independently and rejecting majority of events as non-interested. And it was added when the whole code was much easier - serialization cache is more tricky in my opinion.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: See the bottom of the document ("Alternatives" section) :-)
I'm not convinced LRU cache is better, but potentially may be convinced...

I'm happy to have this discussion below in "Alternatives" section.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: > proto objects keep the raw bytes; a metadata accessor is generated that either reads the raw proto bytes or deserializes only metadata.

Metadata is not enough - as an example for pods, we are filtering by NodeName.

> Setting self link and RV shuffles bytes around rather than decoding and re-encoding.

That's quite tricky and error prone for protobufs, I'm afraid.

One thing that we're keep forgetting are different object versions - from etcd we get object in version X, it doesn't mean that it is being served to watcher also in version X. I think making that assumption may cause us significant issues, when we e.g. will decide to migrate to new version of Endpoints object (which is under discussion).
So I'm quite reluctant to this option.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: Let's leave that discussion for a bit later...

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: No, I'm concerned that objects egressing apiserver via different routes might have been serialized differently, and that would stop being true. It sounds like a bug when I phrase it like that...

wojtek-t: Yeah - I think I'm still confused - if we can get different thing it sounds like a bug to me...

keps/sig-api-machinery/20190329-less-object-serializations.md
lavalamp: Yeah, that's a good point. The proto munging would only be available in one case, although it's likely to be a common case.

keps/sig-api-machinery/20190329-less-object-serializations.md
liggitt: self links *are* different for different endpoints that serve the same group/version/kind:

```
$ kubectl get --raw /api/v1/namespaces/kube-system/pods/kube-dns-5994d786b-rtnpq | jq .metadata.selfLink
"/api/v1/namespaces/kube-system/pods/kube-dns-5994d786b-rtnpq"

$ kubectl get --raw /api/v1/namespaces/kube-system/pods/kube-dns-5994d786b-rtnpq/status | jq .metadata.selfLink
"/api/v1/namespaces/kube-system/pods/kube-dns-5994d786b-rtnpq/status"
```

lavalamp: Ah, of course, thank you, let's pretend that this is what I had in mind rather than just a poorly articulated suspicion :)

wojtek-t: Thanks Jordan - I didn't know about that.
Are those different depending on subresource only, or they can differ also in other situations?

liggitt: subresource is the only one that comes to mind, but it's more a matter of the information only being available at output time (in [transformObject](https://github.com/kubernetes/kubernetes/blob/c28b3b1fdd0d00972461653ac33599ffe5eccb2e/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/response.go#L40-L46))

wojtek-t: I don't think this is problem for my proposal - where serializing is done at the very end (when self-links are already set) [so I'm proposing to do serialization (or get data from cache] when regular serialization is done]. 

What I just realized is that actually, given that SelfLink reflects both group0-version and subresource, maybe we can actually do caching based on SelfLink, which would solve this particular issue (caching is done when all the information is already set).

lavalamp: Using selfLink as the key for different serializations of a given object sounds clever.

wojtek-t: The more I think about it, the more I'm convinced that something like that will be needed in every single approach I can imagine.
I will try to update the proposal in the next couple days.

wojtek-t: I updated the proposal to reflect the "SelfLink"-related discussion.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: Resolving given there is "Alternatives" section in the bottom of the doc.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: I added "smart object" to the "alternatives" section.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+- `versioning.codec` when serializing `CachingObject` chooses appropriate
+`Serialization` object from its `Serializations` field and passes it down;
+if such `Serialization` doesn't exist, it simply deep-copies the wrapped
+`Object` field
+- json and protobuf serializers, serialize and cache it in the object only
+on the first serialization (`once` field) and return the cached result
+
+Now the question is what is responsible for creating `CachingObject`.
+In the first attempty, we will put that responsibility into `watchCache`:
+when receiving an event via watch from etcd it will be opaquing it into
```
lavalamp: opaquing? Maybe "wrapping" is more usual in this context?

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+`CachingObject` and operating on objects of that type later.
+
+Note that given that objects are already opaqued in `watchCache`, we
+also avoid unnecessary serialization when GET/LIST requests are served
+from cache.
+
+One open questions is how do we decide what exactly we want to cache.
+Caching everything stored in `watchCache` seems expensive from memory
+usage point of view. What we suggest, is allow configuring individual
```
lavalamp: This seems like a lot of work, and at the end you leave it on for the most hungry resources anyway. Perhaps modify this to "we'll measure and see if it's a problem"?

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+`registries` with exact versions that we want to cache and propagate
```
lavalamp: Another idea: each caching object, on construction, is registered with a per-resource LRU. The LRU deletes cached serializations when it evicts objects. Objects also remove themselves when they are dropped from the watch cache. Something like this.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+We considered using simple LRU cache to store serialized objects.
+
+Pros:
+- performance gains also for reads served from etcd (though these
+doesn't seem to be huge based on experiments)
+
+Cons:
+- potentially significant point of contention
+- no-control over what is still cached (e.g. for frequently changing
+resources, we still keep them in cache, even if they will never be
```
lavalamp: I can imagine an implementation with this problem but it doesn't seem like we'd have to do it that badly--the LRU I suggested above doesn't have this problem.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+served again)
+
+### Smart objects
+We also considered using `smart objects` - an object that carries the
+serialized format of object from etcd with itself.
+
+Pros:
+- very clear encapsulation
```
lavalamp: Also helps the direct read path.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+
+Cons:
+- We need an ability to in-place add fields to serialized object
+(i.e. SelfLink) - very tricky and error-prone
```
lavalamp: That depends a lot on how you do it, I think. If we e.g. annotate the fields we're interested in, I bet we can make the proto encoding library emit functions which safely do these things. @jennybuckley and @apelisse have just obtained some relevant experience.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+- This doesn't work across different (group, version) pairs. As
```
lavalamp: Again, only if we do it wrong. We could e.g. convert, reserialize, and cache the reserialized forms.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+that information down to `Cacher`. That would give us ability to switch
+it on for most demanding resources (e.g. only Node, Endpoints and Pod
+ifrom core/v1 group and only if empty subresource).
+
+Note that based on a [POC][] (slightly different that above design),
+the gains of implementing it include:
+- eliminating kube-apiserver unresponsiveness in case of write of
+a single huge Endpoints object: [#75294#comment-472728088][]
+- ~7% lower cpu-usage
+- XX% less memory allocations
```
lavalamp: It seems to me that most of these gains are from preventing multiple serializations for separate watchers of the same resource, right?

What if we constrained the serialization caching just to those goroutines, so that the cached serialization can be dumped as soon as the object has been sent to the last watcher?

I feel like that makes more local changes (complexity is contained within the watch cache code, the rest of the stack never runs the risk of getting a CachingObject, etc), doesn't suffer from excessive memory usage, and captures most of the benefit.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+	// serializations is a list of cached serializations for different
+	// media types.
+	//
+	// TODO: Maybe explicitly: jsonSerialization and protobufSerialization?
+	serializations []*serialization
+}
+
+// TODO: Better name is welcome :)
+type serialization struct {
+	once sync.Once
```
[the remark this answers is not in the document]
wojtek-t: Depends how exactly we create it, but conceptually yes it's addition what need to be locked.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: So let me clarify one thing - the "registering" part is used to ensure that we will really serialize object once, right?
The non-obvious part here is that, given our identifier is selfLink (or at cache level, probably (selfLink, RV) pair), we don't know the selfLink at the point of creation. So it seems that we would conceptually register a tree with root being RV, to which we can add selfLink-identified children, and media-type below those.

But at the high-level, I think it's basically variation about LRU option, which I think is a useful variation, though I'm not 100% sure it will actually help that much with complexity of the solution. I will think about it a bit more.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: > It seems to me that most of these gains are from preventing multiple serializations for separate watchers of the same resource, right?

This is an interesting observation. There is also an aspect of restarting watches after timeout, where we potentially send a bunch of "older" cache events. But once we have  watch bookmarks, that seems at least like something worth considering. We probably would need to gather a bit more data about this - I will see if I can find a bit of time to do that.

So let's for now assume that we can get 90%+ savings from it (which sounds like a reasonable assumption to me).

> What if we constrained the serialization caching just to those goroutines, so that the cached serialization can be dumped as soon as the object has been sent to the last watcher?

That sounds reasonable.
But I'm not sure I really understand how that really helps with complexity. Yes - it helps with bounding memory increase etc., but all the mechanics will still be there (we will need this object that is serialized differently, right?)
Or an I missing something?

lavalamp: >  but all the mechanics will still be there (we will need this object that is serialized differently, right?)

I was thinking that this way the CachingObject can be private to the watch cache (or in an internal/ package). But, on further thought, maybe that was true anyway? If so I agree this doesn't help with complexity too much.

I do think it is probably good to either always do it this way, or be able to switch (ideally automatically) to doing it this way for really big collections, to avoid memory problems.

If the complexity here can be contained in the watch cache (sub-) packages and doesn't need to e.g. change anything in the runtime package, I'm much happier overall.

wojtek-t: > I do think it is probably good to either always do it this way, or be able to switch (ideally automatically) to doing it this way for really big collections, to avoid memory problem

I agree.

> If the complexity here can be contained in the watch cache (sub-) packages and doesn't need to e.g. change anything in the runtime package, I'm much happier overall.

In the current proposal it's not - it requires some (small but still) changes in serializers.
While I believe that it should be possible to get rid of potential changes in json/proto serializers, I don't yet know how to solve the versioning problem.
Encode we're doing in apiserver is doing "versioning+serialization", and I don't know how we can make this versioning independently from the whole apimachinery infrastructure purely in watchcache.
But if you have any suggestions, I'm more than happy to hear.

wojtek-t: What I started thinking about (though I'm wondering if it's not overkill) is that we may:
- implement some autogeneration of "WrappedObject" for each object that is worth caching serialization (in every single GV)
- in cacher, just wrap an object into its wrapper
- (if needed) autogenerate custom conversions for these objects
I think that with something like that we would be able to solve the versioning. If we would solve the marshaling similarly, maybe that has some potential to work?

I know it's a lot of hand-waving, but maybe it's something worth exploring.
I can try digging deeper in the next couple weeks, if you think this option is worth exploring further.

lavalamp: > In the current proposal it's not - it requires some (small but still) changes in serializers.

If this is making them more flexible/reusable, then I think I can live with that, as long as existing usage sites don't have to be changed in non-mechanical ways.

wojtek-t: > If this is making them more flexible/reusable, then I think I can live with that, as long as existing usage sites don't have to be changed in non-mechanical ways.

The current proposal doesn't require any changes (neither mechanical nor non-mechanical) for any usage of existing libraries - only minor changes to runtime/serializers.
FTR the POC is here: https://github.com/kubernetes/kubernetes/pull/60067


But I started wondering, if what I described above actually won't be less complicated (or at least, the complication will be much more separated from anything else). I will try to think about it a bit more.

lavalamp: Thanks, I made a couple of concrete suggestions in the PR for if you want to do it that way.

I'm (predictably?) not too excited about solving this with more autogenerated code :)

keps/sig-api-machinery/20190329-less-object-serializations.md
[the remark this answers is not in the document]
wojtek-t: Should be more clear now from the types above.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,283 @@
...
+	//
+	// TODO: Maybe it should be a map?
+	Serializations []*Serialization
+}
+
+// TODO: Better name is welcome :)
+type Serialization struct {
+	// selfLink of the serialized object identifying endpoint
+	// (e.g. group, version, subresource)
+	selfLink string
```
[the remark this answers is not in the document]
wojtek-t: Obsolete.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: This took a long of time, but I finally got back to it. I have created a new version of POC: https://github.com/kubernetes/kubernetes/pull/76942/files that should be addressing those comments.

lavalamp: Thanks, I spent a while reading that code. It's definitely a lot more palatable than the first take. I'll leave some comments/questions there.

Have you considered trying to cache the serialized form of the entire watch event, maybe even post-compression (https://github.com/kubernetes/enhancements/pull/1115)?

wojtek-t: > Have you considered trying to cache the serialized form of the entire watch event, maybe even post-compression (#1115)?

Not sure I understand this - this is what I'm doing right? Or maybe what you're saying is that I'm just caching the "runtime.Object" instead of "watch.Event" (the different is opaquing it into another object and adding type field - so I don't think it will change much).

Re compression - I didn't try it. But IIUC what we're now proceeding with is that we will be only compressing large enough list requests and not do that for watch.

wojtek-t:
Changed names according to suggestions from review of POC.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,305 @@
...
+for regular GET/LIST operations - reading the same object N times will result
+in serializing that N times independently.
+
+This proposal presents a solution for that problem.
+
+[#75294]: https://github.com/kubernetes/kubernetes/issues/75294
+
+### Goals
+
+- Reduce load on kube-apiserver and number of memory allocations, by avoiding
```
smarterclayton: Protobuf encoding should only result in 1 or 2 memory allocations.  Removing that is unlikely to show any significant benefit.

Note that the protobuf encoding changes proposed by @apelisse should have a significant speed up in protobuf encoding, which means that we should really consider the impact of this change AFTER we get those changes in mainline.  It may improve things - or it may have much less impact now that protobuf encoding is significantly faster.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,305 @@
...
+
+This proposal does not introduce any user-visible changes - the proposed changes
+are purely implementation details of kube-apiserver.
+
+The first observation is that a given object may be serialized to multiple
+different formats, based on it's:
+- group and version
+- subresource
+- output media type
+However, group, version and subresource are reflected in the `SelfLink` of
```
smarterclayton: Removing self link completely would actually probably remove several allocations from each object, which should not be discounted.  I don't want to get tied into having to return self link when it is used at most in 1% of client interactions (if that).

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,305 @@
...
+
+### LRU cache
+We considered using simple LRU cache to store serialized objects.
+
+Pros:
+- performance gains also for reads served from etcd (though these
+doesn't seem to be huge based on experiments)
+
+Cons:
+- potentially significant point of contention
```
smarterclayton: Really?  Isn't there a significant time/space locality to serializations?  Couldn't we be exploiting this affinity between multiple watches and the serialization of an object better?

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Smart objects
+We also considered using `smart objects` - an object that carries the
+serialized format of object from etcd with itself.
+
+Pros:
+- very clear encapsulation
+
+Cons:
+- We need an ability to in-place add fields to serialized object
```
smarterclayton: Do we have evidence that we have multiple watchers at different versions?  I assume 99% of watchers are from the same api version, so even a very dumb smart object (last serialization to a version is preserved) could keep that locality.

smarterclayton:
I really would want to see prototype numbers this after the protobuf marshal efficiency stuff is landed.  A significant speed up there may significantly reduce the benefit of this approach.

Note that we have a very high per GET / watch event fixed allocations.  Right now each GET to the API server does about 120 allocations, of which at least a hundred are things besides serialization.  Each watch is ~20 allocations - only 2 of which are serialization.  I would wager we are in the point where we need to reduce fixed cost rather than variable cost (in encoding).

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: There is still a deep-copy issue that we're doing for every single watch event. So that helps, but there is still a lot of place for improvement.

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: Yeah - I wanted to restore that discussion about removing SelfLink. But that will not happen within a quarter even if we make the decision about it quickly due to deprecation policy, right?

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: I did a quick and super dirty poc for that an year ago, and the contention was quite huge actually...

wojtek-t:
> I really would want to see prototype numbers this after the protobuf marshal efficiency stuff is landed. A significant speed up there may significantly reduce the benefit of this approach.

Protobuf is only one part of this problem. Another one is the problem of deep-copying the whole object for every watch-event that is happening (because we're modifying it for conversions, selflinks, etc.)
@apelisse helps here, but it doesn't solve the problem.

> Note that we have a very high per GET / watch event fixed allocations. Right now each GET to the API server does about 120 allocations, of which at least a hundred are things besides serialization. Each watch is ~20 allocations - only 2 of which are serialization. I would wager we are in the point where we need to reduce fixed cost rather than variable cost (in encoding).

Cost of deep-copies is exactly that cost. I wouldn't say it's fixed, but it's huge.

keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -0,0 +1,305 @@
...
+// TODO: Better name is welcome :)
+type CachingVersionedObject struct {
+	// Ojbect is the versioned object for which serializations
+	// should be cached.
+
+	// FIXME: We may want to change that during performance experiments
+	// e.g. to use sync.Map or slice or something different, also to
+	// allow some fast-path.
+	lock sync.Mutex
+	serialization map[cachingKey]*serializationResult
```
wojtek-t: Given that (at least initially) we will use that only for watch (which doesn't support subresources), cachingKey can probably be simply replaced by encoder...

lavalamp:
We discussed at the last sig and no one had specific concerns, just a vague sense of uneasiness over complexity. so...

/approve
/lgtm

wojtek-t:
Thanks Daniel - I will open an issue in this repo and would like to target Alpha (disabled by default) for 1.16. Will be sending some smaller updated to the KEP in the meantime.

PR #926, opened 2019-03-28
Initial KEP for API union/oneOf

apelisse:
Not meant to be a feature, mostly to document what we're doing.

### DIFF
--- a/keps/sig-api-machinery/20190325-unions.md
+++ b/keps/sig-api-machinery/20190325-unions.md
@@ -1,0 +1,272 @@
+---
+title: Union types
+authors:
+  - "@apelisse"
+owning-sig: sig-api-machinery
+participating-sigs:
+reviewers:
+- @sttts
+- @lavalamp
+- @thockin
+- @DirectXMan12
+approvers:
+- @lavalamp
+editor: TBD
+creation-date: 2019-03-25
+last-updated: 2019-03-25
+status: implementable
+see-also:
+  - "/keps/sig-api-machinery/0006-apply.md"
+replaces:
+  - https://docs.google.com/document/d/1lrV-P25ZTWukixE9ZWyvchfFR0NE2eCHlObiCUgNQGQ
+superseded-by:
+---
+
+# Union types
+
+## Summary
+
+Modern data model definitions like OpenAPI v3 and protobuf (versions 2 and 3)
+have a keyword to implement “oneof” or "union". They allow APIs to have a better
+semantics, typically as a way to say “only one of the given fields can be
+set”. We currently have multiple occurrences of this semantics in kubernetes core
+types, at least:
+- VolumeSource is a structure that holds the definition of all the possible
+  volume types, only one of them must be set, it doesn't have a discriminator.
+- DeploymentStrategy is a structure that has a discrminator
+  "DeploymentStrategyType" which decides if "RollingUpate" should be set
+
+The problem with the lack of solution is that:
+- The API is implicit, and people don't know how to use it
+- Clients can't know how to deal with that, especially if they can't parse the
+  OpenAPI
+- Server can't understand the user intent and normalize the object properly
+
+## Motivation
+
+Currently, changing a value in an oneof type is difficult because the semantics
+is implicit, which means that nothing can be built to automatically fix unions,
+leading to many bugs and issues:
+- https://github.com/kubernetes/kubernetes/issues/35345
+- https://github.com/kubernetes/kubernetes/issues/24238
+- https://github.com/kubernetes/kubernetes/issues/34292
+- https://github.com/kubernetes/kubernetes/issues/6979
+- https://github.com/kubernetes/kubernetes/issues/33766
+- https://github.com/kubernetes/kubernetes/issues/24198
+- https://github.com/kubernetes/kubernetes/issues/60340
+
+And then, for other people:
+- https://github.com/rancher/rancher/issues/13584
+- https://github.com/helm/charts/pull/12319
+- https://github.com/EnMasseProject/enmasse/pull/1974
+- https://github.com/helm/charts/pull/11546
+- https://github.com/kubernetes/kubernetes/pull/35343
+
+This is replacing a lot of previous work and long-standing effort:
+- Initially: https://github.com/kubernetes/community/issues/229, then
+- https://github.com/kubernetes/community/pull/278
+- https://github.com/kubernetes/community/pull/620
+- https://github.com/kubernetes/kubernetes/pull/44597
+- https://github.com/kubernetes/kubernetes/pull/50296
+- https://github.com/kubernetes/kubernetes/pull/70436
+
+Server-side [apply](http://features.k8s.io/555) is what enables this proposal to
+become possible.
+
+### Goals
+
+The goal is to enable a union or "oneof" semantics in Kubernetes types, both for
+in-tree types and for CRDs.
+
+### Non-Goals
+
+We're not planning to use this KEP to release the feature, but mostly as a way
+to document what we're doing.
+
+## Proposal
+
+In order to support unions in a backward compatible way in kubernetes, we're
+proposing the following changes.
+
+Note that this proposes unions to be "at most one of". Whether exactly one is
+supported or not should be implemented by the validation logic.
+
+### Go tags
+
+We're proposing a new type of tags for go types (in-tree types, and also
+kubebuilder types):
+
+- `// +union` before a structure means that the structure is a union. All the
+  fields must be optional (beside the discriminator) and will be inlucded as
+  members of the union. That structure CAN be embedded in another structure.
+- `// +unionDeprecated` before a field means that it is part of the
+  union. Multiple fields can have this prefix. These fields MUST BE optional,
+  omitempty and pointers. The field is named deprecated because we don't want
+  people to embed their unions directly in structures, and only exist because of
+  some existing core types (e.g. `Value` and `ValueFrom` in
+  [EnvVar](https://github.com/kubernetes/kubernetes/blob/3ebb8ddd8a21b/staging/src/k8s.io/api/core/v1/types.go#L1817-L1836)).
+- `// +unionDiscriminator` before a field means that this field is the
+  discriminator for the union. Only one field per structure can have this
+  prefix. This field HAS TO be a string, and CAN be optional.
+
+Multiple unions can exist per structure, but unions can't span across multiple
+go structures (all the fields that are part of a union has to be together in the
+same structure), examples of what is allowed:
+
+```
+// This will have one embedded union.
+type TopLevelUnion struct {
+	Name string `json:"name"`
+
+	Union `json:",inline"`
+}
+
+// This will generate one union, with two fields and a discriminator.
+// +union
+type Union struct {
+	// +unionDiscriminator
+	// +optional
+	UnionType string `json:"unionType"`
+
+	// +optional
+	FieldA int `json:"fieldA"`
+    // +optional
+	FieldB int `json:"fieldB"`
+}
+
+// This also generates one union, with two fields and on discriminator.
+type Union2 struct {
+	// +unionDiscriminator
+	Type string `json:"type"`
+	// +unionDeprecated
+    // +optional
+	Alpha int `json:"alpha"`
+	// +unionDeprecated
+    // +optional
+	Beta int `json:"beta"`
+}
+
+// This has 3 embedded unions:
+// One for the fields that are directly embedded, one for Union, and one for Union2.
+type InlinedUnion struct {
+	Name string `json:"name"`
+
+	// +unionDeprecated
+	// +optional
+	Field1 *int `json:"field1,omitempty"`
+	// +unionDeprecated
+	// +optional
+	Field2 *int `json:"field2,omitempty"`
+
+	Union  `json:",inline"`
+	Union2 `json:",inline"`
+}
+```
+
+### OpenAPI
+
+OpenAPI v3 already allows a "oneOf" form, which is accepted by CRD validation
+(and will continue to be accepted in the future). That oneOf form will be used
+for validation, but is "on-top" of this proposal.
+
+A new extension is created in the openapi to describe the behavior:
+`x-kubernetes-unions`.
+
+This is a list of unions that are part of this structure/object. Here is what
+each list item is made of:
+- `discriminator: <discriminator>` is set to the name of the discriminator
+  field, if present,
+- `fields-to-discriminateBy: {"<fieldName>": "<discriminateName>"}` is a map of
+  fields that belong to the union to their discriminated names. The
+  discriminatedValue will typically be set to the name of the Go variable.
+
+Conversion between OpenAPI v2 and OpenAPI v3 will preserve these fields.
+
+### Discriminator
+
+For backward compatibility reasons, discriminators should be added to existing
+union structures as an optional string. This has a nice property that it's going
+to allow conflict detection when the selected union field is changed.
+
+We also do strongly recommend new APIs to be written with a discriminator, and
+tools (kube-builder) should probably enforce the presence of a discriminator in
+CRDs.
+
+The value of the discriminator is going to be set automatically by the apiserver
+when a new field is changed in the union. It will be set to the value of the
+`fields-to-discriminateBy` for that specific field.
+
+When the value of the discriminator is explicitly changed by the client, it
+will be interpreted as an intention to clear all the other fields. See section
+below.
+
+### Normalizing on updates
+
+A "normalization process" will run automatically for all creations and
+modifications (either with update or patch). It will happen automatically in order
+to clear fields and update discriminators. This process will run for both
+core-types and CRDs. It will take place before validation. The sent object
+doesn't have to be valid, but fields will be cleared in order to make it valid.
+This process will also happen before fields tracking (server-side apply), so
+changes in discriminator, even if implicit, will be owned by the client making
+the update (and may result in conflicts).
+
+This process works as follows:
+- If there is a discriminator, and its value has changed, clear all fields but
+  the one specified by the discriminator,
+- If there is no discriminator, or if its value hasn't changed,
+  - if there is exactly one field, set the discriminator when there is one
+    to that value. Otherwise,
+  - compare the fields set before and after. If there is exactly one field
+    added, set the discriminator (if present) to that value, and remove all
+    other fields. if more than one field has been added, leave the process so
+    that validation will fail.
+
+#### "At most one" versus "exactly one"
+
+The goal of this proposal is not to change the validation, but to help clients
+to clear other fields in the union. Validation should be implemented for in-tree
+types as it is today, or through "oneOf" properties in CRDs.
+
+In other word, this is proposing to implement "at most one", and the exactly one
+should be provided through another layer of validation (separated).
+
+#### Clearing all the fields
+
+Since the system is trying to do the right thing, it can be hard to "clear
+everything". In that case, each API could decide to have their own "Nothing"
+value in the discriminator, which will automatically trigger a clearing of all
+fields beside "Nothing".
+
+### Backward compatibilities properties
+
+This normalization process has a few nice properties, especially for dumb
+clients, when it comes to backward compatibility:
+- A dumb client that doesn't know which fields belong to the union can just set
+  a new field and get all the others cleared automatically
+- A dumb client that doesn't know about the discriminator is going to change a
+  field, leave the discriminator as it is, and should still expect the fields to
+  be cleared accordingly
+- A dumb client that knows about the discriminator can change the discriminator
+  without knowing which fields to clear, they will get cleared automatically
+
+
+### Validation
+
+Objects have to be validated AFTER the normalization process, which is going to
+leave multiple fields of the union if it can't normalize. As discussed in
+drawbacks below, it can also be useful to validate apply requests before
+applying them.
+
+## Future Work
+
+Since the proposal only normalizes the object after the patch has been applied,
+it is hard to fail if the patch is invalid. There are scenarios where the patch
+is invalid but it results in an unpredictable object. For example, if a patch
+sends both a discriminator and a field that is not the discriminated field, it
+will either clear the value sent if the discriminator changes, or it will change
+the value of the sent discriminator.
+
+Validating patches is not a problem that we want to tackle now, but we can
+validate "Apply" objects to make sure that they do not define such broken
+semantics.


apelisse:
Note that a lot of the logic is already written here:
- https://github.com/kubernetes-sigs/structured-merge-diff/pull/72/

Steps missing:
- Parse the OpenAPI properly,
- Generate OpenAPI extensions from Go types,
- Validation

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+# Union types
+
+## Summary
+
+Modern data model definitions like OpenAPI v3 and protobuf (versions 2 and 3)
+have a keyword to implement “oneof” or "union". They allow APIs to have a better
+semantic, typically as a way to say “only one of the given fields can be
+set”. We currently have multiple occurrences of this semantic in kubernetes core
+types, at least:
+- DeploymentStrategy is a structure that has a discrminator
```
lavalamp: nit: put this example second--don't include a discriminator in the first example, discriminator with only one choice is perhaps a surprising thing to include for some people.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+- `// +oneof` before a field means that it is part of the union. Multiple fields
+  can have this prefix. Alternatively, the form `// +oneof=<name>` allows to
+  provide the discriminated name for that specific field. These fields MUST BE
+  optional, omitempty and pointers.
+- `// +discriminator` before a field means that this field is the discriminator
+  for the union. Only one field per structure can have this prefix. This field HAS
+  TO be a string, and CAN be optional.
+
+Note that this means that only one "union" type can be defined per structure,
+which is a deliberate decision.
```
lavalamp: Can I effectively get multiple by using embedding?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+
+As for Go types, we only accept one union per structure, which means there can
+be only one discriminator, and all other `oneof` fields are assumed to belong to
+the same union.
+
+### Discriminator
+
+For backward compatibility reasons, discriminators can be added to existing
+union structures as an optional string. This has a nice property that it's going
+to allow conflict detection when a union field is flipped.
```
lavalamp: What does "flipped" mean exactly?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+
+The value of the discriminator is going to be set automatically by the apiserver
```
lavalamp: I think you intend the value of the discriminator to exactly match the names of the discriminated fields. Please say that explicitly?

Is it an option to choose none of the fields, and if so, how is that represented in the discriminator?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+```
+discriminator: b
+a: 1
+b: 1
+```
+
+Since the discriminator hasn't changed, but `a` has been added, the discriminator
+will flip to `a`, and `b` will be dropped.
+
+Validating patches is not a problem that we want to tackle now, but we can
```
lavalamp: Maybe label this section "future work"? I don't think we can leave this problem in forever.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+- https://github.com/kubernetes/kubernetes/issues/33766
+- https://github.com/kubernetes/kubernetes/issues/24198
+- https://github.com/kubernetes/kubernetes/issues/60340
+
+And then, for other people:
+- https://github.com/rancher/rancher/issues/13584
+- https://github.com/helm/charts/pull/12319
+- https://github.com/EnMasseProject/enmasse/pull/1974
+- https://github.com/helm/charts/pull/11546
+- https://github.com/kubernetes/kubernetes/pull/35343
```
lavalamp: The lists are great. To this section I'd add that the serverside apply work is what enables us to finally solve this problem.

tallclair:
I thought we, the kubernetes maintainers, were unionizing.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+We're proposing a new type of tags for go types (in-tree types, and also
+kubebuilder types):
+
+- `// +oneof` before a field means that it is part of the union. Multiple fields
+  can have this prefix. Alternatively, the form `// +oneof=<name>` allows to
+  provide the discriminated name for that specific field. These fields MUST BE
+  optional, omitempty and pointers.
+- `// +discriminator` before a field means that this field is the discriminator
+  for the union. Only one field per structure can have this prefix. This field HAS
+  TO be a string, and CAN be optional.
```
thockin: If optional it must have a documented default behavior including "not set".

keps/sig-api-machinery/20190325-unions.md
thockin: Do we do a lot of embedding?  It has nice properties for sharing "base type" but less nice properties for code that consumes them.  I don't know if our OpenAPI schema generation flattens embedded types properly (it seems so from a quick glance).  If we think it's a good pattern, we might do well to use it more.  I guess we can't refit existing code with embedding (breaks protobuf encoding) but we could use it more in the future (and in future API versions)

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+- `x-kubernetes-oneof-discriminator: true` is set for the discriminator field,
+- `x-kubernetes-oneof: "<discriminatedName>` is set for discriminated fields,
+
+As for Go types, we only accept one union per structure, which means there can
+be only one discriminator, and all other `oneof` fields are assumed to belong to
+the same union.
+
+### Discriminator
+
+For backward compatibility reasons, discriminators can be added to existing
```
thockin: Should we take a stronger stance?  Existing one-of sets might not have discriminators, but in order to add a field to a one-of block in a skew-compatible it's really the only way that the laggier half of the skew can know "this block was set but to something I don't understand", right?

Especially for "exactly one of these must be set" blocks.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+
+The goal is to enable a union or "oneof" semantic in Kubernetes types, both for
+in-tree types and for CRDs.
+
+### Non-Goals
+
+We're not planning to use this KEP to release the feature, but mostly as a way
+to document what we're doing.
+
+## Proposal
```
thockin: We have examples in our code of both "at most one of" and "exactly one of" blocks.  Does that affect this proposal?  Should we write strong rules in one case than the other?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+For backward compatibility reasons, discriminators can be added to existing
+union structures as an optional string. This has a nice property that it's going
+to allow conflict detection when a union field is flipped.
+
+The value of the discriminator is going to be set automatically by the apiserver
+when a new field is changed in the union, and will be interpreted as an
+intention to clear all the other fields when its value is changed. See section
+below.
+
+### Normalization process
```
thockin: "...upon updates" ?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+
+The normalization process (which runs before validating that the union property
```
thockin: Is this going to be handled automatically as part of the REST strategy or something?  E.g. are we generating "normalize" functions that get called before validation?  It would be amazing if this was magical and Just Worked without individual API authors having to do anything.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,201 @@
...
+
+- `// +oneof` before a field means that it is part of the union. Multiple fields
+  can have this prefix. Alternatively, the form `// +oneof=<name>` allows to
+  provide the discriminated name for that specific field. These fields MUST BE
+  optional, omitempty and pointers.
+- `// +discriminator` before a field means that this field is the discriminator
+  for the union. Only one field per structure can have this prefix. This field HAS
+  TO be a string, and CAN be optional.
+
+Note that this means that only one "union" type can be defined per structure,
```
thockin: Does this fail for any existing types?

keps/sig-api-machinery/20190325-unions.md
apelisse: I'll clarify that but basically: since we now have access to the openapi in all the handler path (CREATE/UPDATE/PATCH/"Apply") to track fields manager, we can also automatically process these normalization before validation.

keps/sig-api-machinery/20190325-unions.md
apelisse: Not that I know of. This is sort of old (and the conclusions there are not up-to-date): https://docs.google.com/document/d/1lrV-P25ZTWukixE9ZWyvchfFR0NE2eCHlObiCUgNQGQ

keps/sig-api-machinery/20190325-unions.md
apelisse: Yeah, this is not really replacing the validation logic, just trying to help clear the fields so that there is at most one. If there is 0, I expect the validation will catch that. I'll think about it a little bit more, it'd be nice if we had a way to formalize that.

apelisse: OK, I thought about it, I like the parallel with protobuf: this gives a semantic for "at most one of" and clears other fields. It's up to validation to decide if exactly-one of is accepted.

apelisse:
I've updated *some* things, I'm not exactly happy with everything yet.

apelisse:
cc @sttts @mbohlool

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,185 @@
...
+kubebuilder types):
+
+- `// +oneof` before a field means that it is part of the union. Multiple fields
+  can have this prefix. These fields MUST BE optional, omitempty and pointers.
+- `// +discriminator` before a field means that this field is the discriminator
+  for the union. Only one field per structure can have this prefix. This field HAS
+  TO be a string, and CAN be optional.
+
+Note that this means that only one "union" type can be defined per structure,
+which is a deliberate decision. This also includes go embedding which, once
```
sttts: "This also includes" means what? Embedding is a workaround for having multiple union, or it's forbidden?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,185 @@
...
+### OpenAPI
+
+OpenAPI v3 already allows a "oneOf" form, which is accepted by CRD validation
+(and will continue to be accepted in the future), they will be interpreted
+properly if possible (see CRD GA effort).
+
+Since OpenAPI v2 has no support for oneOf, a couple of kubernetes specific
+extensions will be added to OpenAPI:
+- `x-kubernetes-oneof-discriminator: true` is set for the discriminator field,
+- `x-kubernetes-oneof: "<discriminatedName>` is set for discriminated
```
sttts: for easier interpretation of our v3 output, shouldn't be add these extensions there as well?

At best we come up with some oneOf pattern in OpenAPI which is pretty hard to detect. There will be more things to check than just the existence of the oneOf operator. E.g. oneOf of to patterns is no union.

sttts: Another point: I don't want to see that we turn a concrete oneOf pattern into some API convention that we cannot change ever for compatibilty reasons. It is much easier to leave the pattern open (of course it must comply with OpenAPI semantics, but nobody should depend on it to stay the same), and at the same time have these more abstract extensions which will stay forever and are part of the API server API contract.

sttts: Last but not least, we have to offer these extensions in the CRD validation schemata. Translating some oneOf pattern into those extensions for v2 is unnecessary complex. Also we couple the openapi generator to this pattern (compare previous comment), although both apiserver and generator have different life cycles.

keps/sig-api-machinery/20190325-unions.md
apelisse: it is forbidden, I'll rephrase.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+  discriminated names. The discriminatedValue will typically be set to the name
+  of the Go variable.
+
+These fields belong directly in the object that holds the union.
+
+Conversion between OpenAPI v2 and OpenAPI v3 will preserve these fields.
+
+As for Go types, we only accept one union per structure, which means there can
+be only one discriminator, and all other `oneof` fields are assumed to belong to
+the same union.
```
sttts: what does this paragraph mean here for OpenAPI?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+OpenAPI v3 already allows a "oneOf" form, which is accepted by CRD validation
+(and will continue to be accepted in the future). That oneOf form will be used
+for validation, but is "on-top" of this proposal (validation is considered as
+not included in this proposal).
+
+New extensions are created in the openapi to describe the behavior:
+- `x-kubernetes-union-discriminator: <discriminator>` is set to the name of the
+  discriminator field, if present,
+- `x-kubernetes-oneof-fields-discriminated: {"<fieldName>":
+  "<discriminatedName>"}` is a map of fields that belong to the union to their
```
sttts: if we restrict us to one union per object, let's simplify this tag to a simple list of field names.

Also I would suggest a simpler tag name: x-kubernetes-union-fields

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+will be interpreted as an intention to clear all the other fields. See section
+below.
+
+If the object had exactly one of the union field set, and that field is removed,
+the discriminator will be cleared.
+
+### Normalizing on updates
+
+A "normalization process" will run automatically for all creations and
+modifications (either with put or patch). It will happen automatically in order
```
sttts: we call them update and patch

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+doesn't have to be valid, but fields will be cleared in order to make it valid.
+This process will also happen before fields tracking (server-side apply), so
+changes in discriminator, even if implicit, will be owned by the client making
+the update (and may result in conflicts).
+
+This process works as follows:
+- If there is a discriminator, and its value has changed, clear all fields but
+  the one specified by the discriminator,
+- If there is no discriminator, or if its value hasn't changed,
+  - if there is exactly one field set, set the discriminator (if exists) to that
```
sttts: what does "if exists" mean here exactly? We fall through to the second step? or leave the discriminator missing?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+The value of the discriminator is going to be set automatically by the apiserver
+when a new field is changed in the union. It will be set to the value of the
+`x-kubernetes-union-fields-discriminated` extension for that specific field.
+
+When the value of the discriminator is explicitly changed by the client, it
+will be interpreted as an intention to clear all the other fields. See section
+below.
+
+If the object had exactly one of the union field set, and that field is removed,
+the discriminator will be cleared.
```
sttts: is there a difference between empty and non-existing discriminator? I.e. will it be a `*string` in the Go types? This seems to play role two paragraphs further down.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,215 @@
...
+
+In other word, this is proposing to implement "at most one", and the exactly one
+should be provided through another layer of validation (separated).
+
+#### Clearing all the fields
+
+Since the system is trying to do the right thing, it can be hard to "clear
+everything". In that case, each API could decide to have their own "Nothing"
+value in the discriminator, which will automatically trigger a clearing of all
+fields beside "Nothing".
```
sttts: An explicit empty string would not do?

apelisse: We're assuming that clients that deserialize in golang and are not aware of a newly introduced discriminator would reset that string by accident. We don't want to infer intent from that.

keps/sig-api-machinery/20190325-unions.md
apelisse: will fix

keps/sig-api-machinery/20190325-unions.md
apelisse: I'll clarify, thanks

keps/sig-api-machinery/20190325-unions.md
apelisse: it's not a list on purpose, because we really want to be able to have a discriminated name that is not explicitely the name of the field. For example today, we have an implicit discriminator in `DeploymentStrategyType` whose value must be `RolloutStrategy` (matching the field name `rolloutStrategy`). For volumes, it's not as simple as capitalizing, e.g. nfs -> NFS. Keeping this mapping of field name to discriminated name is useful.

keps/sig-api-machinery/20190325-unions.md
apelisse: yeah, one union per openapi object (allof flattening included). I'll clarify the doc

keps/sig-api-machinery/20190325-unions.md
jpbetz: How about instead of "discriminated", we call this "union member", where the discriminator field value is the name of a union member?

Then could we have `x-kubernetes-union-member-fields: { "<fieldName>": "<unionMemberName>"}` ? Or, if I'm following @sttts, maybe we could have `{ "<unionMemberName>": "<listOfFieldNames>"}`

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+supported or not should be implemented by the validation logic.
+
+### Go tags
+
+We're proposing a new type of tags for go types (in-tree types, and also
+kubebuilder types):
+
+- `// +union` before a field means that it is part of the union. Multiple fields
+  can have this prefix. These fields MUST BE optional, omitempty and pointers.
+- `// +discriminator` before a field means that this field is the discriminator
```
jpbetz: Quite a few union type systems I've seen us "tag" instead of "discriminator". Conveniently, it's much shorter.  How about `// +union-tag` here?

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+
+Note that this proposes unions to be "at most one of". Whether exactly one is
+supported or not should be implemented by the validation logic.
+
+### Go tags
+
+We're proposing a new type of tags for go types (in-tree types, and also
+kubebuilder types):
+
+- `// +union` before a field means that it is part of the union. Multiple fields
```
DirectXMan12: I think we should put this on types:

```go
// +union
type MetricSource struct {
  // +discriminant
  Type MetricSourceType
  Object *ObjectMetricSource
  Pods *PodsMetricSource
  ...
}
```

We should be thinking of unions like enums in Rust, algebraic datatypes in Haskell,  discriminated unions in typescript, etc: as a type-level construct with multiple variants and a discriminator, at both the IDL-ish (Go) level and at an API level.  This is conceptually similar to what people are familiar with, and makes writing things out easier.

This should hold for most cases in Kubernetes.  The few where it doesn't we can either a) special-case, b) fix before they go alpha (*cough* webhookconfig *cough*), or c) fix at a type level and leave the API the same by making an external struct and embedding.

lavalamp: I also prefer a formulation like this one, I think. Embedding can potentially be used to transition existing places where union and non-union fields are mixed.

sttts: What about something in the middle: `+union-fields=Object,Pods` on types.

apelisse: That seems more error-prone and tedious and hard to read? :-/

I'm going with both: you can either specify tags on members individually or entire structs. They'll generate the same OpenAPI, and we'll only document the struct way, which is the one we want people to use (and also is probably going to be the only one available through kubebuilder)

liggitt:
xref https://github.com/kubernetes/kubernetes/issues/35345

keps/sig-api-machinery/20190325-unions.md
sttts: > 'm going with both:

sounds good.

keps/sig-api-machinery/20190325-unions.md
sttts: makes sense.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+As for Go types, we only accept one union per structure (including embeddeing),
+which means there can be only one discriminator, and all other `oneof` fields
+are assumed to belong to the same union.
+
+Only one union is allowed per openapi objects (even after flattening of allOfs).
+
+### Discriminator
+
+For backward compatibility reasons, discriminators can be added to existing
+union structures as an optional string. This has a nice property that it's going
```
sttts: If this value is defaulted, it does not have to be optional anymore.

sttts: which remind me that client side validation does not see defaults yet and hence falls over a non-optional discriminator. I guess that's what you have in mind.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+- A dumb client that knows about the discriminator can change the discriminator
+  without knowing which fields to clear, they will get cleared automatically
+
+
+### Validation
+
+Objects have to be validated AFTER the normalization process, which is going to
+leave multiple fields of the union if it can't normalize. As discussed in
+drawbacks below, it can also be useful to validate apply requests before
+applying them.
```
sttts: would normalization be part of the defaulter?

apelisse: No, for server-side apply it happens BEFORE we default, for all other modifying end-points, it happens after defaulting. It's part of the "field management"

keps/sig-api-machinery/20190325-unions.md
apelisse: Right, if we make it required, then client-side validation will force people to set it, which we don't want.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+
+The problem with the lack of solution is that:
+- The API is implicit, and people don't know how to use it
+- Clients can't know how to deal with that, especially if they can't parse the
+  OpenAPI
+- Server can't understand the user intent and normalize the object properly
+
+## Motivation
+
+Currently, changing a value in an one-of type is difficult because the semantic
```
sttts: semantics, doesn't exist in singular AFAIK.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+- https://github.com/kubernetes/kubernetes/pull/44597
+- https://github.com/kubernetes/kubernetes/pull/50296
+- https://github.com/kubernetes/kubernetes/pull/70436
+
+Server-side [apply](http://features.k8s.io/555) is what enables this proposal to
+become possible.
+
+### Goals
+
+The goal is to enable a union or "oneof" semantic in Kubernetes types, both for
```
sttts: oneof or one-of? both is used in the text

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+  - if there is exactly one field, set the discriminator when there is one
+    to that value. Otherwise,
+  - compare the fields set before and after. If there is exactly one field
+    added, set the discriminator (if present) to that value, and remove all
+    other fields. if more than one field has been added, leave the process so
+    that validation will fail.
+
+#### "At most one" versus "exactly one"
+
+The goal of this proposal is not to change the validation, but to help client
```
sttts: ... help clients to clear ...

keps/sig-api-machinery/20190325-unions.md
sttts: If the "Nothing" value is customizable, how will that happen via tags?

keps/sig-api-machinery/20190325-unions.md
sttts: ok, so the plumbing is mostly in place. good.

sttts:
Lgtm.

What I miss: will we add discriminators and tags to all existing unions in k/k? What is the roadmap for this?

keps/sig-api-machinery/20190325-unions.md
DirectXMan12: `// +discriminator:nothingValue="Nothing"` or some such, presumably.  We've got at least one union in kube today that requires this.

It'd be nice to be able to explicitly say that you don't want this behavior though, for new APIs going forward.

apelisse: We already have one of these and it works very well:  `DeploymentStrategy`.

`Recreate` is the "nothing", `RolloutStrategy` is the one that has an associated value. We should let it up to validation to verify which values are accepted for the discriminator, but I don't think we should build anything more complicated here.

apelisse:
> will we add discriminators and tags to all existing unions in k/k? What is the roadmap for this?

I'm currently working at this, we'll add discriminator as optional before the release.

keps/sig-api-machinery/20190325-unions.md
thockin: On one hand, it should be OK to have clients be aware of defaulting rules (they are part of the API), so clients could make a copy, apply defaults, and validate that.

On the other hand, that fails in case of admission controllers.  But if that is a requirement, then client-side validation can never work.

I know this is a little controversial, but I think it's OK to client-side default (in a temporary copy!!) to validate.

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Discriminator
+
+For backward compatibility reasons, discriminators can be added to existing
+union structures as an optional string. This has a nice property that it's going
+to allow conflict detection when the selected union field is changed.
+
+The value of the discriminator is going to be set automatically by the apiserver
+when a new field is changed in the union. It will be set to the value of the
+`fields-to-discriminatedBy` for that specific field.
```
thockin: s/discriminated/discriminate/  # note the "d" ?

keps/sig-api-machinery/20190325-unions.md
apelisse: I'm not a big fan of "tag", I think it's much less explicit than discriminator, I'm going with `// +union-discriminator`

sttts: tag is overused.

apelisse:
Updated, PTAL

keps/sig-api-machinery/20190325-unions.md
sttts: we have dry-run, don't we?

apelisse: Sure

sttts: We could also just stop validating required fields client-side.

apelisse: I don't see why this is important anyway, you can have a look at the behavior here (which I think is a big improvement): https://github.com/kubernetes-sigs/structured-merge-diff/blob/master/merge/union_test.go

keps/sig-api-machinery/20190325-unions.md
```diff
@@ -0,0 +1,264 @@
...
+- `// +union` before a structure means that the structure is a union. All the
+  fields must be optional (beside the discriminator) and will be inlucded as
+  members of the union. That structure CAN be embedded in another structure.
+- `// +unionDeprecated` before a field means that it is part of the
+  union. Multiple fields can have this prefix. These fields MUST BE optional,
+  omitempty and pointers. The field is named deprecated because we don't want
+  people to embed their unions directly in structures, and only exist because of
+  some existing core types (e.g. `Value` and `ValueFrom` in
+  [EnvVar](https://github.com/kubernetes/kubernetes/blob/3ebb8ddd8a21b/staging/src/k8s.io/api/core/v1/types.go#L1817-L1836)).
+- `// +union-discriminator` before a field means that this field is the
```
thockin: nit:  why the variance in naming?  `unionDeprecated` and `union-discriminator` ?

keps/sig-api-machinery/20190325-unions.md
thockin: Still worth thinking about?

apelisse:
Updated reviewer's list and squashed.

deads2k:
Nice to see our existing behavior documented.  I've fielded questions about this in the past.

/approve

deads2k:
reapplying @thockin 's lgtm

/lgtm

PR #930, opened 2019-04-02
Added proposal with priority but assured concurrency and fairness

MikeSpreitzer:


MikeSpreitzer:
CC @yue9944882 
CC @wojtek-t

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -233,7 +233,606 @@ yet but we think may be interesting to consider in the future.
...
+  name: workload-low
+spec:
+  priorityLevel: 10000
+  assuredConcurrencyShares: 10
+  queuesPerWidth: 128
+  handSize: 6
+  queueLengthLimit: 100
+```
+
+Some flow schemata.
```
yue9944882: I believe the api model of flow schema here can be simplified by:
- removing support for `patternMatch` and `nonPatternMatch`. additionally, all controllers are sharing the same user identity as `system:kube-controller-manager` but differs in user-agent http header.
- make the field `matchingPriority` optional and only filled to resolve conflicts among multiple flow schemas. A default value will be working for most cases I suppose.
- instead of defining explicit boolean tests in the api model. consider make it a bit simpler by costing a bit ambiguity. as the yaml examples in the following shows:

```yaml
kind: FlowSchema
meta:
  name: system-top
spec:
  matchingPriority: 500 // default to 500, override if necessary
  requestPriority:
    name: system-top
  matchRules:
  - field: groups
     valueSet: [ "system:masters" ]
---
kind: FlowSchema
meta:
  name: system-high-node-heartbeat
spec:
  matchingPriority: 500 // default to 500, override if necessary
  requestPriority:
    name: system-high
  flowDistinguisher:
    source: user
    # no transformation in this case
    matchRules:
  - field: groups
     valueSet: [ "system:nodes" ]
  - field: resource
     valueSet: [ "nodes" ]
---
kind: FlowSchema
meta:
  name: system-high-system-objects
spec:
  matchingPriority: 500 // default to 500, override if necessary
  requestPriority:
    name: system-high
  flowDistinguisher:
    source: user
    # no transformation in this case
    matchRules:
  - field: groups
     valueSet: [ "system:nodes" ]
  - field: namespace
     valueSet: [ "kube-system" ]
---
kind: FlowSchema
meta:
  name: system-high-system-objects
spec:
  matchingPriority: 500 // default to 500, override if necessary
  requestPriority:
    name: system-high
  flowDistinguisher:
    source: user
    # no transformation in this case
    matchRules:
  - field: groups
     valueSet: [ "system:nodes" ]
  - field: namespace
     valueSet: [ "kube-system" ]
---
kind: FlowSchema
meta:
  name: workload-low-serviceaccounts
spec:
  matchingPriority: 500 // default to 500, override if necessary
  requestPriority:
    name: workload-high
  flowDistinguisher:
    source: namespace
    # no transformation in this case
  matchRules:
  - field: groups
     valueSet: [ "system:serviceaccounts" ]
....
```

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -233,7 +233,606 @@ yet but we think may be interesting to consider in the future.
...
+
+Whenever the width-weighted sum of running non-zero priority requests
+is below the concurrency limit, it is time to consider dispatching
+another request for service.  This is done in two stages.  First, a
+priority level with non-empty queues is chosen. The logic for doing
+that makes use of each priority level's _assured concurrency value_
+(ACV), which is calculated as follows.
+
+```
+ACV(l) = ceil( SCL * ACS(l) / (100 + sum[priority levels k] ACS(k) ) )
```
yue9944882: why is it 100? i have a complex feeling about introducing more magic numbers..

keps/sig-api-machinery/20190228-priority-and-fairness.md
liggitt: > * all controllers are sharing the same user identity as `system:kube-controller-manager` but differs in user-agent http header.

that's not generally accurate... typically, each controller loop gets a distinct identity

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: This number is not particularly magic.  It just establishes a magnitude standard to compare the ACS values with.  All that matters is their ratios.  We _could_ create another configuration parameter to define the share for shared concurrency.  But, without loss of generality, we can fix it at a particular number.

MikeSpreitzer:
In the meeting today we agreed to start simpler, and #933 is an attempt to capture that agreement.
I expect we will soon want to restore borrowing between priority levels.
Nobody was confident that we will want to restore the distinction between mutating and read-only.

wojtek-t:
@MikeSpreitzer - should this one be closed now?

MikeSpreitzer:
It is not something we are trying to merge now.  OTOH, it has a more advanced design that I think we will want to look at second.

PR #953, opened 2019-04-12
Add kubelet support lxcfs

fanux:


keps/sig-node/20190412-lxcfs.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Non-Goals
+
+This feature only mount lxcfs files, the implemention of resource view is depend on [lxcfs](https://github.com/lxc/lxcfs)
+
+## Proposal
+
+### Implementation Details/Notes/Constraints [optional]
+
+Add lxcfs volumes when mount pod volumes:
```
derekwaynecarr: Can you discuss the pro/cons of having this be a feature of a CRI implementation versus the kubelet itself?

keps/sig-node/20190412-lxcfs.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Kubelet support lxcfs
+
+The following criteria applies to `SupportKubeletLxcfs` feature gate:
+
+Alpha
+- basic support integrated in kubelet
+
+Beta
+- ensure proper node e2e test coverage is integrated verifying cgroup settings
```
derekwaynecarr: Does this mean on by default and used everywhere?  Typically feature gates once enabled are on everywhere, and it doesn’t seem like this is universally portable.  It’s partly why I am wondering if this belongs more in the CRI provider.

keps/sig-node/20190412-lxcfs.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Summary
+
+A proposal to enable pod container resource view, we get resource info in pod
+is the host resource when using resource quota, with lxcfs we can will have a
+right reource view in pod.
+
+## Motivation
+
+Java apps always abnormal run in pods with resource quota, because JVM allocate
```
yujuhong: Would this JDK improvement handle your use case? https://bugs.openjdk.java.net/browse/JDK-8146115

keps/sig-node/20190412-lxcfs.md
resouer: +1, as I mentioned in one of previous comments, we can not assume lxcfs is installed in every cluster. Defer to a CRI provider would make more sense. AFAIK, there're runtimes planning to implement lxcfs support while have to workaround CRI.

keps/sig-node/20190412-lxcfs.md
fanux: Yes, it can solve this problem, but the platform can't control user behavior

keps/sig-node/20190412-lxcfs.md
fanux: Yes, some runtime like pouch already supports this feature, but our platform using multiple runtimes we feel tired to implement it on all runtimes. Now I have an other idea to solve this problem, call it kubelet premout, we can appoint some volumes in the config file and mount then when pod creating, so we needn't care about lxcfs, and it also can use in other use case.s.

salanki:
Sad this got closed. Really useful.

PR #959, opened 2019-04-13
Add KEP for Behavior-driven Conformance Testing

johnbelamaric:


neolit123:
thank you for this KEP @johnbelamaric !

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,268 @@
...
+    description: Two containers can mount the same volume and see each others' files.
+```
+
+In order to allow incremental adoption of this framework, rather than a new
+conformance framework function, test authors will indicate the behaviors covered
+by their tests with a label. So, a change may look something like:
+
+```
+-       framework.ConformanceIt("should get a host IP [NodeConformance]", func() {
++       framework.ConformanceIt("should get a host IP [NodeConformance] [Behavior:pods/lifecycle/hostip]", func() {
```
timothysc: At the end of the day you're basically outlining a better description...  Should we just update the current ones to spell out the behavior using a clear defined pattern?

timothysc:
/hold 

I'm not opposed to adding more meaning data for folks to inspect, but I'm not sold on the approach atm.

keps/sig-architecture/20190412-conformance-behaviors.md
johnbelamaric: One goal is to separate the people reviewing the behaviors from those that review the tests. I think we could build out the entire list of behaviors for, say, the Pod API pretty quickly. Then, we would backfill with tests / annotate existing tests with these IDs. We get a measure of coverage of the behaviors, plus a clear consensus on what "conforming behaviors" are.

Just updating current tests means figuring out what each test does, then documenting it. Which doesn't leave us with any notion of how much of the real behavior is covered.

johnbelamaric:
> @johnbelamaric can we talk about this on our next conformance call?

Alas I won't be at the meeting next week, but we can discuss it on the following one.

timothysc:
From the conversation on 20190507 we discussed that it makes the most sense for us to generate the meta-data content using api-machinery from there augment the existing tests with a [tag] reference and then prioritize and expand.  I like this plan alot.  

We should update the KEP to outline the conversation.

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,268 @@
...
+
+type Behavior struct {
+	Id          string `json:"id,omitempty"`
+	Description string `json:"description,omitempty"`
+}
+```
+
+An `Area` is an arbitrary grouping used to organize tests. The `Api` field is
+optional and may specify the API group to which the tests apply, if any. Each
+behavior has a unique identifier and a description. An example is shown below.
```
bowei: Do you want to be more prescriptive about structure of `Id`? Seems like there is a common pattern being used in the example below...

timothysc:
@johnbelamaric - Would love to walk through this proposal in detail during next weeks meeting, to try and get this rolling.

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
+---
+title: Behavior-driven Conformance Testing
+authors:
+  - "@johnbelamaric"
+owning-sig: sig-architecture
+participating-sigs:
+  - sig-testing
+reviewers:
+  - TBD
```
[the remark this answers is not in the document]
johnbelamaric: Ok. Regarding additional meta-data, we can actually get pretty far without it. Things like https://github.com/kubernetes/enhancements/pull/1099 will allow us to improve this tooling.

I am hesitant to add additional meta-data as part of this KEP, just because I want to keep the scope contained and have it not touch too much. We can add in more later. That said, I am open to the discussion.

johnbelamaric:
@timothysc unfortunately due to my travel schedule I am not able to make the meeting tomorrow. Please ask any questions you may have and we'll try to move it along as much as possible, or I am up for a one-off meeting to go through it with anyone who is interested

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+### Representation of Behaviors
+
+Behaviors must be captured in the repository and agreed upon as required for
+conformance. Behaviors are broken into feature areas, and there are multiple
+suites for each feature area. Some of these suites may be machine-generated
+based upon the API schema, whereas others are handwritten. Keeping the
+generated and handwritten suites in separate files allows regeneration of the
+auto-discovered behavior suites.
+
+Validation and conformance designations are made on a per-suite basis,
```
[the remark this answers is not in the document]
alejandrox1: actually, that comment would make more sense in the previous paragraph.

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+* Tooling to generate lists of behaviors from the API schemas.
+* Tooling to generate tests and test scaffoloding to evaulate those behaviors.
+* Tooling to compare the implemented tests to the list of behaviors and
+  calculate coverage.
+
+### Representation of Behaviors
+
+Behaviors must be captured in the repository and agreed upon as required for
+conformance. Behaviors are broken into feature areas, and there are multiple
+suites for each feature area. Some of these suites may be machine-generated
```
alejandrox1: Could you give an example of how a feature area may be broken into suites?

hh:
Would be good to take a look at existing BDD frameworks and see how well they implement the clear separation of behaviour vs testing, generation/reuse of skaffolding/steps, and behaviour coverage. Not suggesting rewriting everything, but I do think it's worth exploring some existing BDD tooling.

I suspect we could one of these in combination with our current approach, as underlying our current approach (and nearly every alternative) is https://golang.org/pkg/testing/.

We currently use a slightly customized version of Gingko, which combines test definition+behavior in go files where an `It` block is placed within a `Describe` or `Context` block.
 
- Gingko/Gomega https://github.com/onsi/ginkgo
  - http://onsi.github.io/ginkgo/
  - http://onsi.github.io/gomega/
  - https://content.pivotal.io/blog/announcing-ginkgo-and-gomega-bdd-style-testing-for-golang

Here are some other active and healthy BDD libraries / communities with more explicit separation of behaviors and tests using some form of [Gherkin](https://cucumber.io/docs/gherkin/reference/) where [Steps (like Given, When, Then)](https://cucumber.io/docs/gherkin/reference/#steps) define (usually in files separate from tests) what we expect to happen, and regex matchers are used to tie (and reuse) behaviour steps to actual test code.

- GoDog https://github.com/DATA-DOG/godog
  - https://gist.github.com/zer0tonin/c5637ed53691ec3245518035b9799afc
  - https://alicegg.tech/2019/03/09/gobdd.html
  - https://github.com/DATA-DOG/godog#running-godog-with-go-test
  - https://codereview.stackexchange.com/questions/203496/hello-world-bdd-in-golang

- Goerkin https://github.com/bunniesandbeatings/goerkin
  - http://www.bunniesandbeatings.com/goerkin/
  - https://github.com/bunniesandbeatings/goerkin#shared-steps

- GoConvey https://github.com/smartystreets/goconvey
  - http://goconvey.co/
  - https://godoc.org/github.com/smartystreets/goconvey/convey#FocusConvey
  - https://github.com/smartystreets/goconvey/wiki/Composition
  - https://medium.com/@durgaprasadbudhwani/golang-best-practices-behavior-driven-development-and-continuous-integration-53ed72c915a5

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+
+Some sets of behaviors may be tested in a similar, mechanical way. Basic CRUD
+operations, including updates to specific fields and constraints on immutable
+fields, operate in a similar manner across all API resources. Given this, it is
+feasible to automate the creation of simple tests for these behaviors, along
+with the behavior descriptions in the `spec.yaml`. In some cases a complete test
+may not be easy to generate, but a skeleton may be created that can be be
+converted into a valid test with minimal effort.
+
+For these tests, the input is a set of manifests that are applied to the
```
johnSchnake: Doesn't this put is back into a similar position as we are today? How will we know how much "coverage" we have when it comes to these, more complex, behavioral tests? Arent those nearly just as manual as the tests today?

I do like the simplicity of having a standard approach (apply manifest, wait for condition, end of story) but what does that mean for all the other complex stuff that we currently do in tests?
 - non-manifest actions, like tearing down a node or checking resources before/after an action
 - larger cycles of apply changes/testX/more changes/testY/etc/etc that can't be reduced

It just seems like this only applies to a small subset of tests of the form (createX testY); is that the intent?

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Implementation History
+
+- 2019-04-12: Created
+- 2019-06-11: Updated to include behavior and test generating from APIs.
+
+## Drawbacks
+
+* Separating behaviors into a file that is not directly part of the test suite
+  creates an additional step for developers and could lead to divergence.
```
johnSchnake: Since the conformance-wg exists I guess I am not too worried about this. If it was just left up to individual dev self-control I would, but these things would only have to be monitored closely when tests become conformance or not conformance behaviors decided upon, right? In which case I think that is a low risk that we'll suddenly forget about it.

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+  * API functions as expected: Resource is created and may be read, and defaults
+    are set. This is mechanical and can be completely generated.
+  * Cluster behaves as expected. This cannot be generated, but a skeleton can be
+    generated that allows test authors to evaluate the condition of the cluster
+    to make sure it meets the expectations.
+* Deletion of the resource. This may be mostly mechanical but if there are side-
+  effects, such as garbage collection of related resources, we may want to have
+  manually written evaluation here as well.
+* Creation of resource with each field set, and update of each mutable field.
+  * For each mutable field, apply a patch to the based resource definition
```
johnSchnake: Maybe too early an optimization, but do you think each field needs separately patched? Or if you just patched them all with unique values and ensured they got applied to the right fields that would be sufficient?

I ask because the `Behavior` type seems to specify that a `Behavior` (which maps to a test, right?) would only be testing a single field at a time. It just seems like that could get really verbose really quickly. I know it will be autogenerated, but it would be preferred IMO to be able to condense those things when possible.

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+field updates can be reused.
+
+### Coverage Tooling
+
+In order to tie behaviors back to the tests that are generated, including
+existing e2e tests that already cover behaviors, new tags with behavior IDs will
+be added to conformance tests. Using the existing conformance framework
+mechanism allows incremental adoption of this proposal. Thus, rather than a new
+conformance framework function, test authors will indicate the behaviors covered
+by their tests with a tag in the `framework.ConformanceIt` call.
```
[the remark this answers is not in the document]
spiffxp: ideas:
- `Testname: behavior id`
- `ConformanceIt("behavior id", "should function for node-pod communication:", func...`
- Change Conformance tag to be like our Feature tag, `[Conformance:behavior id]` (but this would break back compat)

keps/sig-architecture/20190412-conformance-behaviors.md
```diff
@@ -0,0 +1,342 @@
...
+cluster’s operation.
+
+Additionally, progress in writing and promoting tests has been slow and too much
+manual effort is involved. As a starting point, this proposal includes new
+tooling that uses the API schemas to identify expected behaviors and produce
+tests and test scaffolding to quickly cover those behaviors.
+
+### Goals
+
+* Enable separate review of behaviors and tests that evaluate those behaviors.
```
spiffxp: +1 this is the main thing I wish to accomplish

- One group of humans (behavior-approvers) to ensure:
  - does this behavior describe the way kubernetes is supposed to operate
  - is this behavior something that is reasonably expected to work across all kubernetes clusters
- Another group of humans (test-approvers) to ensure:
  - does this test actually exercise this behavior
  - is this test written well according to our standards for conformance (in an ideal world, this would be lintable)

Today, promoting a test involves updating a file that is blocked on behavior-approvers.  Our only options to improve our velocity today are:
- train our test-approvers to reduce doublecheck review burden on behavior-approvers
- grow our pool of behavior-approvers

(I still think these approaches are healthy for the project, but we seem to have been unable to dedicate time or build sufficient trust here)

spiffxp: To decouple things, promoting a test should not block on conformance behavior-approvers, but should block on conformance test-approvers.
- promotion should not involve updating behavior files
- promotion should involve update files under test-approver control

My best guess at the moment:
```
test/conformance
├── behaviors
│   ├── OWNERS # no-parent: true, approvers: behavior-approvers
│   └── {area}
│       ├── OWNERS # optional: reviewers: area-experts
│       └── {suite}.yaml
├── OWNERS # approvers: test-approvers
└── tests.yaml # promotion updates this file; tests MUST map to a behavior
```

hh:
@johnbelamaric I've create a PR for your KEP via your repo https://github.com/johnbelamaric/enhancements/pull/1

It includes much of this comment reworked as a possible implementation of your proposal.
Would love feedback from you and others.

- [How can we declaratively define behaviors in a standard fashion?](#sec-1)
- [Behaviors as Gherkin .feature files](#sec-2)
- [Complex Storytelling made Possible](#sec-3)
- [Metadata to match existing tests](#sec-4)

# How can we declaratively define behaviors in a standard fashion?<a id="sec-1"></a>

Requirement: `A MAP` that enumerates the State Space - Definition of Done

Our current tests are not super easy to write, read, or review. BDD in go was in it's early days when k8s started integration testing with a tightly coupled testing approach. The e2e Ginko framework evolved upon those tightly coupled assumptions. This approach unfortunately lacks a description of the desired behavior other than in comments in the code. It also lacks metadata and usable tagging for efficient organization into test areas and suites.

Documenting and discovering of all our behaviors will require a combination of automated introspection and some old fashioned human storytelling.

We need to standardize the business language that our bottle-necked people can use to write these stories in a way can be powertool assisted with some automation. This would reduce complexity for articulating concrete requirements for execution in editors, humans, and automation workflows.

I'm suggesting a short 4 week exploration into Gherkin as the behavior file format and godog as a BDD library to address our current painpoints and lack of automation tooling and processes for behavior testing.

# Behaviors as Gherkin .feature files<a id="sec-2"></a>

We could go as far as replace **feature** with **behavior**, as the mapping is pretty spot on.
### behaviourX.feature
```feature
Feature: Structured Metadata allowing Behavior Driven tooling automation
  In order to auto-generate testing scaffolding
  As a sig-X member
  I want to describe the behavior of X

  @sig-X
  Scenario: Behavior X
    Given a well formed file describing the behavior X
    When I run the automation
    Then I am provided with the basic structure for a corresponding test
    And this is fine
  @sig-Y
  Scenario: Behavior Y
    Given a well formed file describing the behavior Y
    When I run the automation
    Then I am provided with the basic structure for a corresponding test
    And this is fine
  @sig-Y @sig-X
  Scenario: Behavior X+Y
    Given a well formed file describing the behavior X
    And a well formed file describing the behavior Y
    When I run the automation
    Then I can reuse existing step definitions on multiple tests
    And this is fine
```
### godog test scaffolding generated from .feature file

```shell
~/go/bin/godog --no-colors
```

Generated output:

```feature
Feature: Structured Metadata allowing Behavior Driven tooling automation
  In order to auto-generate testing scaffolding
  As a sig-X member
  I want to describe the behavior of X

  Scenario: Behavior X                                                  # features/behavior.feature:7
    Given a well formed file describing the behavior X
    When I run the automation
    Then I am provided with the basic structure for a corresponding test
    And this is fine

  Scenario: Behavior Y                                                  # features/behavior.feature:13
    Given a well formed file describing the behavior Y
    When I run the automation
    Then I am provided with the basic structure for a corresponding test
    And this is fine

  Scenario: Behavior X+Y                                       # features/behavior.feature:19
    Given a well formed file describing the behavior X
    And a well formed file describing the behavior Y
    When I run the automation
    Then I can reuse existing step definitions on multiple tests
    And this is fine

3 scenarios (3 undefined)
13 steps (13 undefined)
1.253405ms

You can implement step definitions for undefined steps with these snippets:
```

```go
func aWellFormedFileDescribingTheBehaviorX() error {
  return godog.ErrPending
}

func iRunTheAutomation() error {
  return godog.ErrPending
}

func iAmProvidedWithTheBasicStructureForACorrespondingTest() error {
  return godog.ErrPending
}

func thisIsFine() error {
  return godog.ErrPending
}

func aWellFormedFileDescribingTheBehaviorY() error {
  return godog.ErrPending
}

func iCanReuseExistingStepDefinitonsOnMultipleTests() error {
  return godog.ErrPending
}

func FeatureContext(s *godog.Suite) {
  s.Step(`^a well formed file describing the behaviour X$`, aWellFormedFileDescribingTheBehaviorX)
  s.Step(`^I run the automation$`, iRunTheAutomation)
  s.Step(`^I am provided with the basic structure for a corresponding test$`, iAmProvidedWithTheBasicStructureForACorrespondingTest)
  s.Step(`^this is fine$`, thisIsFine)
  s.Step(`^a well formed file describing the behavior Y$`, aWellFormedFileDescribingTheBehaviorY)
  s.Step(`^I can reuse existing step definitions on multiple tests$`, iCanReuseExistingStepDefinitonsOnMultipleTests)
}
```

# Complex Storytelling made Possible<a id="sec-3"></a>

We could use inline json or yaml as CRUD input or verification, as well as reuse step definitions from pervious scenarios as needed.

```feature
Feature: Intrapod Communication
  Pods need to be able to talk to each other, as well as the node talking to the Pod.
  @sig-node @sig-pod
  Scenario: Nodes can communicate to each other
    Given a pods A and B
    When pod A says hello to pod B
    Then pod B says hello to pod A
  @wip
  Scenario: Pods can can communicate to Nodes
    Given a pod A on a node 
    When the node says hello to pod A
    Then pod A says hello to the node
  @tags-that-are-no-longer-part-of-the-test-name
  Scenario: Pods can can communicate to Nodes
    Given I create pod A with this yaml spec
      """
      yaml: [
         values
      ]
      """
    And I create pod B with this json spec
      """
      {
        json: values
      }
      """
    When I request pod A and pod B talk to each other
    Then I can observe a conversations between them
    And this is fine
```

# Metadata to match existing tests<a id="sec-4"></a>

```feature
Scenario: Use existing ginkgo framework
  As a test contributor
  I want to not throw away all our old tests
  In order to retain the value generated in them
  @sig-node @sig-pod @conformance @release-1.15
  Feature: Map behaviors to existing ginkgo tests
    Given existing test It('should do the right thing')
    When I run the test
    Then we utilize our existing test via our new .feature framework
    And this is fine
```

keps/sig-architecture/20190412-conformance-behaviors.md
johnbelamaric: The idea is to get tooling to handle the simple cases that we can in an automated-as-much-as-possible way. We'll never be able to build tooling to handle the complex cases. We can have tooling that helps scaffold the complex cases, which should help speed development.

The goal here is not to automate away the entire process. It's just to boost the velocity. In v1.15, we have 221 tests IIRC; in v1.9 is was something like 159. By my estimates we need > 2500 tests to cover the basic functions we need (even without new features). That means we should finish around in about 40 releases at the current pace. Ten years is a bit long to wait :).

So, general approach (this should be clarified in the KEP.
* For generating the list of behaviors:
  * Simple behaviors are generated through tooling.
  * Complex behaviors are defined by the SIGs.
* For test creation:
  * Simple behaviors use a basic set of mechanical test scaffolding with fill-in-the-blank validation as needed. 
  * Complex behaviors are hand-written, possibly (hopefully) using some validation primitives from the simple behaviors.

keps/sig-architecture/20190412-conformance-behaviors.md
johnbelamaric: Possibly, but changing multiple things at once makes isolation harder. There may be subsets of fields that can be done this way because they are independent...but not sure how to automate that.

johnbelamaric:
Thanks all for the feedback, let's discuss more in the call today. Since @timothysc and @spiffxp will not be there I am open to another call later this week or next week if needed.

Regarding the meeting discussions, I got the following takeaways, let me know if I missed something:

1) The KEP needs more clarity on how the API schema is used to generate lists of behaviors. Specifically, are there are additional annotations needed in `types.go`, and how do we actually translate that into something useful.
   * Response: Ok, we should update the KEP with a concrete example and more details on this. As I mentioned in a comment above, there are real limits on what can be done here - the intent is to bootstrap the process and get some large X% of behaviors enumerated from this, and scaffolding built. But we'll still need to put in a lot of work to deal with complex cases.

2) We need room for humans in the process.
   * Response: Yes, absolutely agree. The tooling is for handling the tedious items and the toil as much as possible. But humans will need to do all the complex pieces. The tooling needs to be built such that the automation and human contributions live side-by-side, and the automation pieces can be re-executed without messing with the human contributions.

3) Critical to separate behavior and test reviews. We need to be sure approvals for tests are gated only on test-approvers, not behavior-approvers.
   * Response: +1

4) Don't lean on tags.
    * Response: Hmm. Not sure how to do that and keep backwards compatibility. But point taken and let's experiment with a few options.

5) Maybe we should start over.
    * Response: I don't think so. I do think it's a good idea to consider the "if we started from scratch" case, but I don't want to gate on that, we need to keep this moving forward. So we can look at that in parallel (I believe @timothysc took that AI).

spiffxp:
/lgtm
Agree with @timothysc please squash and we can merge as provisional

johnbelamaric:
Commits squashed. (Why don't we just have the bot do that on merge?)

jdumars:
/approve
/hold 

I'm adding the hold, and it would be nice to remove it by next Thursday if possible. I think this is a very positive way to evolve our conformance testing strategy, and with very little downside. Let's just make sure we're extra supportive of our contributor community as they acclimate to the new methodology.

spiffxp:
/hold cancel
discussed during triage that we will merge and iterate from here, as this gives us enough of a framework to experiment and prototype

we will update the kep as we figure out what would work well for the community's development processes

PR #966, opened 2019-04-16
Fix typo in 20190316-rebase-images-to-distroless.md

ad-m:


ad-m:
> please reply here with a new comment and we'll verify

Done.

PR #971, opened 2019-04-17
Updates to Topology Manager KEP

lmdaly:
Added Feature Gate and Kubelet Flags
Updated targeted release versions


derekwaynecarr:
thanks for adding the gate.

/lgtm
/approve

PR #976, opened 2019-04-19
Pod overhead implementable

egernst:
Adding sig-windows and marking the KEP implementable.


tallclair:
lgtm.

@derekwaynecarr to approve

Please hold for @yujuhong 's feedback too.

PR #982, opened 2019-04-23
[WIP] KEP: Switch kubeadm to generate ECDSA keys

rojkov:


neolit123:
thanks @rojkov will attempt to have a closer look later today or tomorrow.
deadline for KEPs, as far as i know is 30th of April.

keps/sig-cluster-lifecycle/kubeadm/20190423-switch-to-ecdsa-keys.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Switching kubeadm to generate ECDSA keys
```
timothysc: Why is this a KEP?  You can just open an issue for this in the kubeadm repo...

https://github.com/kubernetes/kubeadm

timothysc:
/hold 

IMO we should close this KEP, it's not a feature or major enhancement, we can just treat it like an ordinary issue.

rojkov:
Sure, if there's a consensus I'm fine with closing it. The only reason for the KEP is that it can span several releases.

fabriziopandini:
ok for me in moving this to an issue

PR #994, opened 2019-04-24
Kubeadm for Windows KEP

ksubrmnn:
KEP PR for Issue #995

### DIFF
--- a/keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
+++ b/keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
@@ -1,0 +1,292 @@
+---
+title: kubeadm-for-windows
+authors:
+  - "@ksubrmnn"
+  - "@neolit123"
+  - "@patricklang"
+owning-sig: sig-windows
+Participating-sigs:
+  - sig-windows
+  - sig-cluster-lifecycle
+reviewers:
+  - "@timothysc"
+  - "@michmike"
+  - "@fabriziopandini"
+  - "@rosti"
+approvers:
+  - "@timothysc"
+editor: "@ksubrmnn"
+creation-date: 2019-04-24
+last-updated: 2019-04-24
+status: implementable
+---
+
+# Kubeadm for Windows
+
+## Table of Contents
+
+* [Kubeadm for Windows](#kubeadm-for-windows)
+      * [Table of Contents](#table-of-contents)
+      * [Release Signoff Checklist](#release-signoff-checklist)
+      * [Summary](#summary)
+      * [Motivation](#motivation)
+         * [Goals](#goals)
+         * [Non-Goals](#non-goals)
+      * [Proposal](#proposal)
+         * [User Stories](#user-stories)
+            * [Story 1](#story-1)
+         * [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+            * [Kubeadm manages the kubelet start / stop as a service](#kubeadm-manages-the-kubelet-start--stop-as-a-service)
+            * [Kubeadm makes assumptions about systemd and Linux](#kubeadm-makes-assumptions-about-systemd-and-linux)
+            * [Windows vs Linux host paths](#windows-vs-linux-host-paths)
+            * [Kube-proxy deployment](#kube-proxy-deployment)
+            * [CNI plugin deployment](#cni-plugin-deployment)
+         * [Risks and Mitigations](#risks-and-mitigations)
+      * [Design Details](#design-details)
+         * [Test Plan](#test-plan)
+         * [Graduation Criteria](#graduation-criteria)
+               * [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+               * [Beta -&gt; GA Graduation](#beta---ga-graduation)
+         * [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+         * [Version Skew Strategy](#version-skew-strategy)
+      * [Implementation History](#implementation-history)
+      * [Drawbacks](#drawbacks)
+         * [Non-identical flows for Linux and Windows](#non-identical-flows-for-linux-and-windows)
+      * [Infrastructure Needed](#infrastructure-needed)
+
+## Release Signoff Checklist
+
+**ACTION REQUIRED:** In order to merge code into a release, there must be an issue in [kubernetes/enhancements] referencing this KEP and targeting a release milestone **before [Enhancement Freeze](https://github.com/kubernetes/sig-release/tree/master/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core Kubernetes i.e., [kubernetes/kubernetes], we require the following Release Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These checklist items _must_ be updated for the enhancement to be released.
+
+- [ ] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [ ] KEP approvers have set the KEP status to `implementable`
+- [ ] Design details are appropriately documented
+- [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] Graduation criteria is in place
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+**Note:** Any PRs to move a KEP to `implementable` or significant changes once it is marked `implementable` should be approved by each of the KEP approvers. If any of those approvers is no longer appropriate than changes to that list should be approved by the remaining approvers and/or the owning SIG (or SIG-arch for cross cutting KEPs).
+
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+In Kubernetes 1.14, official support was added for Windows Containers. However, there is no official tool available to users for joining a Windows node to a cluster. The current solution is to create a set of scripts to download Kubernetes binaries and write their config files, but this has proven to be cumbersome and is a huge pain point for Windows adoption. On Linux, Kubeadm is available to quickly join nodes to a cluster, and the intent of this KEP is to propose a design to implement the same functionality for Windows.
+
+It should be noted that at this time, this document only proposes enablement of support for Windows worker nodes using kubeadm.
+
+## Motivation
+
+The motivation for this KEP is to provide a tool to allow users to take a Windows machine and join it to an existing Kubernetes cluster with a single command. The user should also be able to reset the node with a single command. We will also create a tool to install prerequisites and enable kubeadm to bring the Windows node to a Ready state.
+
+### Goals
+
+* Create and maintain a Powershell script to install and run Kubernetes prerequisites on Windows
+* Support kubeadm join for Windows
+* Support kubeadm reset for Windows
+* Create and maintain a Powershell script to deploy Windows CNIs
+
+### Non-Goals
+
+* Installing the Container Runtime (e.g. Docker or containerd)
+* Implement kubeadm init for Windows (at this time)
+* Implement kubeadm join --control-plane for Windows (at this time)
+* Supporting upgrades using kubeadm upgrade for Windows (to be revisited for Beta)
+* Running kube-proxy as a DaemonSet on Windows (to be revisited for Beta)
+* Running Flannel as a DaemonSet on Windows (to be revisited for Beta)
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+A user will have a Windows machine that they want to join to an existing Kubernetes cluster.
+
+1. The user will download a set of required binaries (kubelet, kube-proxy, kubeadm, kubectl, Flannel) using a script. The same script will also wrap kubeadm execution.
+
+2. The script will register kubelet and kube-proxy as Windows services.
+
+3. The script will upload a default ConfigMap to the cluster for the default Windows KubeletConfiguration if one does not already exist.
+
+4. The script will run "kubeadm join ..." to join the node to the cluster. In this step kubeadm will run preflight checks and proceed with the regular join process.
+
+5. kubeadm will restart the kubelet service using flags that kubeadm fed to the Windows service. kubeadm will proceed to bootstrap the node. After this process is finished the node should show with the status of NotReady.
+
+6. The same script will then configure FlannelD. The script will (re-)register FlannelD as a service and start it with the correct configuration, optionally using parameters from kubeadm.
+
+7. kube-proxy will do the same steps as FlannelD shortly after.
+
+8. The node status should be Ready.
+
+### Implementation Details/Notes/Constraints
+
+#### Kubeadm manages the kubelet start / stop as a service
+
+Kubeadm has a “kubelet” phase that can start/stop the [kubelet service](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/kubelet/kubelet.go).
+
+Kubeadm already supports running the kubelet as a Windows service. This support was added by these two PRs: [60144](https://github.com/kubernetes/kubernetes/pull/60144), [53553](https://github.com/kubernetes/kubernetes/pull/53553).
+
+This existing support will be leveraged by this proposal.
+
+Currently kubeadm join downloads a KubeletConfiguration "component config" from a ConfigMap (that is created during kubeadm init) and writes it on all joining nodes. Then each node proceeds to add it's overrides via kubelet flags (that take precedence over config). This process has to be modified so that this KubeletConfiguration is defaulted properly for Windows nodes. One proposed solution is to upload the defaulted KubeletConfiguration for Windows in a separate ConfigMap, but this solution is a subject to change.
+
+#### Kubeadm makes assumptions about systemd and Linux
+
+Kubeadm is quite coupled to systemd and Linux at the moment.
+This file shows some of the kubelet-flags that are Linux bound:
+[kubelet flags](https://github.com/kubernetes/kubernetes/blob/b8b689aae0d6879de5192b203590330a11c7b9e3/cmd/kubeadm/app/phases/kubelet/flags.go)
+
+The following changes are required::
+* Implement proper cri-socket detection in kubeadm for Windows.
+* Omit flags such as “cgroup-driver”
+* Specify the correct “pod-infra-container-image”
+* Also omit writing systemd drop-in files on Windows in general.
+* Populate the Windows service dynamically.
+
+To support Windows specific flags for the kubelet, there is a requirement to split kubeadm’s app/phases/kubelet/flags.go files into two:
+* app/phases/kubelet/flags_windows.go
+* app/phases/kubelet/flags_linux.go
+
+In the case of Windows, we will be updating the registration of the kubelet as a Windows service, by changing its flags based on the flags kubeadm wants to use, complying with Windows specific defaults in kubeadm.
+
+The existing process of passing the following files to the kubelet, will be leveraged:
+* /etc/kubernetes/bootstrap-kubelet.conf
+* /etc/kubernetes/kubelet.conf
+* /var/lib/kubelet/config.yaml
+
+Windows related adjustments to default paths might be required. 
+
+#### Windows vs Linux host paths
+
+Kubeadm makes a number of non-portable assumptions about paths. E.g. “/etc/kubernetes” is a hardcoded path in kubeadm.
+
+We will use "C:\kubernetes" to hold the paths that are normally created for Linux.
+
+We need to evaluate the kubeadm codebase for such instances of non-portable paths - CRI sockets, Cert paths, etc. Such paths need to be defaulted properly in the kubeadm configuration API.
+
+A consolidated list of paths where kubeadm installs files to a set of paths needs to be created and updated to comply with the Windows OS model. At least a single PR against kubeadm will be required to modify the Windows defaults. 
+
+Last, as new paths are created, restrictive Access Control Lists (ACL) for Windows should be applied. Golang does not convert Posix permissions to an appropriate Windows ACL, so an additional step is needed. See [mkdirWithACL from moby/moby](https://github.com/moby/moby/blob/e4cc3adf81cc0810a416e2b8ce8eb4971e17a3a3/pkg/system/filesys_windows.go#L103)) for an example. This step will be performed by kubeadm.
+
+#### Kube-proxy deployment
+
+On Linux, kube-proxy is deployed as a DaemonSet in the kubeadm init phase. However, kube-proxy cannot run as a container in Windows since Windows does not support privileged containers. Kube-proxy should therefore be run as a Windows service so that it is restarted by windows control manager automatically and has lifecycle control.
+
+We need to modify the Linux kube-proxy DaemonSet to not deploy on Windows nodes. A PR is already in flight for that [76327](https://github.com/kubernetes/kubernetes/pull/76327). *Merging this PR is mandatory for this proposal*.
+
+Running kube-proxy as a Windows service from kubeadm is out of scope for this proposal. This is due to the fact that we don’t want the changes in kubeadm to be intrusive to the existing method of running kube-proxy as a DaemonSet on Linux. This can end up requiring an abstraction layer that is far from ideal.
+
+The proposed Windows wrapper script that executes kubeadm will also manage the restart of the kube-proxy Windows service.
+
+Long term and ideally, kube-proxy should be run as a DaemonSet on Windows.
+
+#### CNI plugin deployment
+
+On Linux, CNI plugins are deployed via kubectl and run as a DaemonSet. However, on Windows, CNI plugins need to run on the node, and cannot run in containers (again because Windows does not currently support privileged containers). Azure-CNI, win-bridge (compatible with kubenet), and Flannel all need the binary and config stored on the node.
+
+This proposal plans for FlannelD as the default option. Currently, FlannelD has to be started before the kube-proxy Windows service is started. FlannelD creates an HNS network on the Windows host, and kube-proxy will crash if it cannot find the network. This should be fixed in the scope of this project so that kube-proxy will wait until the network comes up. Therefore, kube proxy can be started at any time. 
+
+However, if FlannelD is deployed in VXLAN (Overlay) mode, then we need to rewrite the KubeProxyConfiguration with the correct Overlay specific values, and kube-proxy will need to read this config again. This is not true for Host-Gateway (L2Bridge) mode. The script will have a flag that allows users to choose between the two networking modes.
+
+If the users wish to use a different plugin they will technically opt-out of the supported setup for kubeadm based Windows worker nodes in the Alpha release.
+
+Long term, any CNI plugin should be supported for kubeadm based Windows worker nodes.
+
+### Risks and Mitigations
+
+**Risk**: Versioning of the wrapper script can become complicated
+
+Versioning of the script per-Kubernetes version can become a problem if a certain new version diverges in terms of download URLs, flags and configuration.
+
+*Mitigation*: Use git branches to version the script in the repository where it is hosted.
+
+**Risk**: The wrapper script is planned to act as both a downloader and runner of the downloader binaries, which might cause scope and maintenance issues.
+
+*Mitigation*: Use separate scripts, the first one downloads the binaries and the wrapper/runner script then setups the environment. The user then executes the wrapper script.
+
+The initial plan is to give the single script method a shot with different arguments that will execute the different stages (downloading, setting up the environment, deploying the CNI).
+
+**Risk**: Flannel or kube-proxy require special configuration that kubeadm does not handle.
+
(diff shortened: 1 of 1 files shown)


neolit123:
@kubernetes/sig-cluster-lifecycle-pr-reviews

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+The proposed Windows wrapper script that executes kubeadm will also manage the restart of the kube-proxy Windows service.
+
+Long term and ideally, kube-proxy should be run as a DaemonSet on Windows.
+
+#### CNI plugin deployment
+
+On Linux, CNI plugins are deployed via kubectl and run as a DaemonSet. However, on Windows, CNI plugins need to run on the node, and cannot run in containers (again because Windows does not currently support privileged containers). Azure-CNI, win-bridge (compatible with kubenet), and Flannel all need the binary and config stored on the node.
+
+This proposal plans for FlannelD as the default option. Currently, FlannelD has to be started before the kube-proxy Windows service is started. FlannelD creates an HNS network on the Windows host, and kube-proxy will crash if it cannot find the network. This should be fixed in the scope of this project so that kube-proxy will wait until the network comes up. Therefore, kube proxy can be started at any time.
```
rosti: Is this how Flannel operates on Windows, or how all CNI plugins should operate on Windows? I am getting the impression that we are trying to solve a Flannel problem here, rather than trying to get the generic CNI operations picture.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+*Mitigation*: The existing kubeadm codebase already has good abstraction in this regard. Still, a PR that makes some non-intrusive adjustments in _windows.go files might be required.
+
+**Risk**: Permissions on Windows paths that kubeadm generates can pose a security hole.
+
+kubeadm creates directories using MakeAll() and such directories are strictly Linux originated for the time being - such as /etc.
+
+On Windows, the creation of such a path can result in sensitive files to be exposed without the right permissions.
+
+*Mitigation*: Manage the preemptive creation of sensitive paths from the wrapper script. Make such paths require Administrator level access.
```
rosti: I am not convinced that we should outsource that to the script. Providing a `SecureMakeAll()` func in kubeadm, that ensures secure enough permissions on both Windows & Linux, and replacing usage of `MakeAll()` is one simple solution. I am sure, that there are others as well.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Design Details
+
+### Test Plan
```
rosti: Whatever the case, we need a strong CI signal for this. Most of us won't have the ability to test this manually and the moment we loose attention we'll get things broken.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+E2e tests might not be complete but provide good signal.
+Documentation is in a good state.
+
+##### Beta -> GA Graduation
+
+The feature is well tested and adapted by the community.
+E2e test provide sufficient coverage.
+Documentation is complete.
+
+##### Removing a deprecated flag
```
rosti: We probably don't need this section.

benmoss:
> have to employ privileged containers on Windows

I have a questionably acceptable idea to workaround this, but basically a RPC server running on the host that gets mounted into the container via a named pipe. It obviously is still less ideal than if privileged containers could be added to Windows, but assuming we can get a general-enough RPC API, we could then be in the position of distributing/maintaining code mostly via Docker images.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+#### Kubeadm makes assumptions about systemd and Linux
+
+Kubeadm is quite coupled to systemd and Linux at the moment.
+This file shows some of the kubelet-flags that are Linux bound:
+[kubelet flags](https://github.com/kubernetes/kubernetes/blob/b8b689aae0d6879de5192b203590330a11c7b9e3/cmd/kubeadm/app/phases/kubelet/flags.go)
+
+The following changes are required::
+* Implement proper cri-socket detection in kubeadm for Windows.
+* Omit flags such as “cgroup-driver”
+* Specify the correct “pod-infra-container-image”
```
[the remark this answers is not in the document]
benmoss: It is today. I have an [open PR](https://github.com/kubernetes/kubernetes/pull/75618) to allow the default k8s pause image to run on both platforms, but for the time being we need a different image.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: i pointed this out to SIG Windows. the signal should be owned and maintained by them.
also SIG CL doesn't have:
- the infrastructure to test Windows nodes.
- the developers with bandwidth to maintain the feature.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
benmoss: This is true for all CNI, since the job of CNI is to modify the host networking configuration, and this can't be done without privileged containers.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: the abstraction of secure paths would require kubeadm in the case of Windows to import some golang/x or make calls to the tool that changes permissions on Windows.
i don't like having this in kubeadm for the Alpha.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: AFAIK, all CNI
@ksubrmnn can confirm.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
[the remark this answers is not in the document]
neolit123: yes, the pause image is not multi-os at GCR.

neolit123:
@benmoss 

>> Of course, this rosy dream of mine relies on getting privileged Windows containers.

> I have a questionably acceptable idea to workaround this, but basically a RPC server running on the host that gets mounted into the container via a named pipe. It obviously is still less ideal than if privileged containers could be added to Windows, but assuming we can get a general-enough RPC API, we could then be in the position of distributing/maintaining the provisioning code via pods.

we need to evaluate the roadmap of privilege containers on Windows.
the question here is when are they coming, and if hey are not coming anytime soon we can start thinking about alternatives.

the RPC idea seems like something that can be proposed in a document, but workarounds may still block the Beta. for this KEP it seems that we might want to hold onto the Windows services idea.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+The existing process of passing the following files to the kubelet, will be leveraged:
+* /etc/kubernetes/bootstrap-kubelet.conf
+* /etc/kubernetes/kubelet.conf
+* /var/lib/kubelet/config.yaml
+
+Windows related adjustments to default paths might be required.
+
+#### Windows vs Linux host paths
+
+Kubeadm makes a number of non-portable assumptions about paths. E.g. “/etc/kubernetes” is a hardcoded path in kubeadm.
```
benmoss: Is the only thing non-portable about this that it lacks a drive specification?

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+We need to evaluate the kubeadm codebase for such instances of non-portable paths - CRI sockets, Cert paths, etc. Such paths need to be defaulted properly in the kubeadm configuration API.
+
+A consolidated list of paths where kubeadm installs files to a set of paths needs to be created and updated to comply with the Windows OS model. At least a single PR against kubeadm will be required to modify the Windows defaults. 
+
+Last, as new paths are created, restrictive Access Control Lists (ACL) for Windows should be applied. Golang does not convert Posix permissions to an appropriate Windows ACL, so an additional step is needed. See [mkdirWithACL from moby/moby](https://github.com/moby/moby/blob/e4cc3adf81cc0810a416e2b8ce8eb4971e17a3a3/pkg/system/filesys_windows.go#L103)) for an example. This step will be performed by the wrapper script.
```
benmoss: This can be done by kubeadm itself, I don't think there's a need to do it from the wrapper

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: the KEP is peculating in terms of drives and paths for Windows.
we need to settle on defaults once the works begins.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: this part is conflating
cc @ksubrmnn 

we want to minimize the changes in kubeadm, because the maintainers have expressed that the lack of kube-proxy and CNI as DaemonSets is quite "alpha".
if it's simply not feasible to do it outside in the wrapper, a change in kubeadm will be considered.

benmoss: My point is that even if we got privileged containers on Windows these concerns about file permissions would remain valid, so we should be putting them into kubeadm. The wrapper should only be for the things that are working around the lack of privileged support.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
ksubrmnn: @benmoss said it correctly.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+In Kubernetes 1.14, official support was added for Windows Containers. However, there is no official tool available to users for joining a Windows node to a cluster. The current solution is to create a set of scripts to download Kubernetes binaries and write their config files, but this has proven to be cumbersome and is a huge pain point for Windows adoption. On Linux, Kubeadm is available to quickly join nodes to a cluster, and the intent of this project is to implement the same functionality for Windows.
```
timothysc: intent of this KEP is to ...

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+It should be noted that this PR only proposes enablement of support for Windows worker nodes using kubeadm.
+
+## Motivation
+
+The motivation for this KEP is to provide a tool to allow users to take a Windows machine and join it to an existing Kubernetes cluster with a single command. The user should also be able to reset the node with a single command. We will also provide the tools necessary to install prerequisites and to bring the Windows node to a Ready state.
+
+### Goals
+
+* Create and maintain a script to install and run Kubernetes prerequisites on Windows
```
timothysc: Is there any reason why this is not build as a phase for join vs. a script?

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+* Support kubeadm join for Windows
+* Support kubeadm reset for Windows
+* Create and maintain a script to deploy WIndows CNIs
```
timothysc: s/WI/Wi

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Non-Goals
+
+* Installing the Container Runtime (e.g. Docker or containerd)
+* Implement kubeadm init for Windows
+* Implement kubeadm join --control-plane
+* Supporting upgrades using kubeadm upgrade
```
timothysc: That's weird, won't we have to support this?

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+A user will have a Windows machine that they want to join to an existing Kubernetes cluster.
+
+To start, the user will download a set of required binaries (kubelet, kube-proxy, kubeadm, kubectl, Flannel) using a script. The same script will also wrap kubeadm execution.
```
timothysc: I still think this can be a phase for join wrapped with if(windows)

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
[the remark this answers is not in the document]
timothysc: Are the ComponentConfig folks thinking about platform wrapping for knobs. 
cc @mtaufen @luxas @stealthybox

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+The following changes are required::
+* Implement proper cri-socket detection in kubeadm for Windows.
+* Omit flags such as “cgroup-driver”
+* Specify the correct “pod-infra-container-image”
+* Also omit writing systemd drop-in files on Windows in general.
+* Populate the Windows service dynamically.
+
+To support Windows specific flags for the kubelet, there is a requirement to split kubeadm’s app/phases/kubelet/flags.go files into two:
+* app/phases/kubelet/flags_windows.go
+* app/phases/kubelet/flags_linux.go
```
timothysc: I'd like work from the componentconfig folks.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,293 @@
...
+* /etc/kubernetes/kubelet.conf
+* /var/lib/kubelet/config.yaml
+
+Windows related adjustments to default paths might be required.
+
+#### Windows vs Linux host paths
+
+Kubeadm makes a number of non-portable assumptions about paths. E.g. “/etc/kubernetes” is a hardcoded path in kubeadm.
+
+We need to evaluate the kubeadm codebase for such instances of non-portable paths - CRI sockets, Cert paths, etc. Such paths need to be defaulted properly in the kubeadm configuration API.
```
timothysc: IMO this should span several releases and be done judiciously, I want us todo this wisely and not in a rush.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
ksubrmnn: cc @neolit123 

I believe for Linux, getting Kubernetes binaries is a prerequisite for kubeadm

ksubrmnn:
> Of course, this rosy dream of mine relies on getting privileged Windows containers.
> 
> I have a questionably acceptable idea to workaround this, but basically a RPC server running on the host that gets mounted into the container via a named pipe. It obviously is still less ideal than if privileged containers could be added to Windows, but assuming we can get a general-enough RPC API, we could then be in the position of distributing/maintaining the provisioning code via pods.


@benmoss @neolit123 

This is a feasible idea, and I completed a POC for this a month ago. It works. I think we can propose both ideas in the doc, and I can advocate for both. This is a more immediate option than waiting on privileged containers, but I think the community should discuss.

PatrickLang:
There's a lot of circling around "privileged containers on Windows". There are some Windows engineers looking into this, but I don't have a clear answer on whether some form of privileged containers could be made to work on Windows Server 2019, or if it would take a new OS version. I'm trying to get them to write up a public proposal but don't expect anything for at least a few weeks.

rosti:
Thanks for the clarification @PatrickLang !
With or without privileged containers on Windows, there is a bunch of stuff that needs to be done either way to get kubeadm running on Windows - CRI detection, kubelet drop-in params, correct handling of Windows paths, etc.
Therefore, I think, that the focus of the initial effort should be on those grounds.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
timothysc: powershell script.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
timothysc: So is phases work for upgrade a precondition https://github.com/kubernetes/kubeadm/issues/1318

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
ksubrmnn: Leaving binary downloads to the script per our office hours discussion

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
ksubrmnn: Included note to reconsider this for future iterations and added this to the Alpha->Beta criteria

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,287 @@
...
+* Create and maintain a Powershell script to deploy Windows CNIs
+
+### Non-Goals
+
+* Installing the Container Runtime (e.g. Docker or containerd)
+* Implement kubeadm init for Windows
+* Implement kubeadm join --control-plane
+* Supporting upgrades using kubeadm upgrade (to be revisited for Beta)
+* Running kube-proxy as a DaemonSet on Windows
+* Running Flannel as a DaemonSet on Windows
```
neolit123: please add `(to be revisited for Beta)` after `Running kube-proxy as a DaemonSet on Windows` and `Running Flannel as a DaemonSet on Windows`.

whether we end up working on that depends on external factors - e.g. Windows core.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: could you elaborate on your ideas?
this could end up being something quite isolated to kubeadm -> kubelet. 
`if os == "windows" { use flagset A } else { use flag set B }`

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,287 @@
...
+### Implementation Details/Notes/Constraints
+
+#### Kubeadm manages the kubelet start / stop as a service
+
+Kubeadm has a “kubelet” phase that can start/stop the [kubelet service](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/kubelet/kubelet.go).
+
+Kubeadm already supports running the kubelet as a Windows service. This support was added by these two PRs: [60144](https://github.com/kubernetes/kubernetes/pull/60144), [53553](https://github.com/kubernetes/kubernetes/pull/53553).
+
+This existing support will be leveraged by this proposal.
+
```
neolit123: Please add the following in a new paragraph here:

> Currently `kubeadm join` downloads a KubeletConfiguration "component config" from a ConfigMap (that is created during `kubeadm init`) and writes it on all joining nodes. Then each node proceeds to add it's overrides via kubelet flags (that take precedence over config). This process has to be modified so that this KubeletConfiguration is defaulted properly for Windows nodes. One proposed solution is to upload the defaulted KubeletConfiguration for Windows in a separate ConfigMap, but this solution is a subject to change.

timothysc:
poke me on slack after it's been updated to include the content changes from yesterdays review.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
ksubrmnn: What is the resolution on this?

neolit123: "from the wrapper script" is the initial idea, but subject of change.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
timothysc: You could build meta-data in apis ~= +omitempty

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
timothysc: This still needs to be decided.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+Long term and ideally, kube-proxy should be run as a DaemonSet on Windows.
+
+#### CNI plugin deployment
+
+On Linux, CNI plugins are deployed via kubectl and run as a DaemonSet. However, on Windows, CNI plugins need to run on the node, and cannot run in containers (again because Windows does not currently support privileged containers). Azure-CNI, win-bridge (compatible with kubenet), and Flannel all need the binary and config stored on the node.
+
+This proposal plans for FlannelD as the default option. Currently, FlannelD has to be started before the kube-proxy Windows service is started. FlannelD creates an HNS network on the Windows host, and kube-proxy will crash if it cannot find the network. This should be fixed in the scope of this project so that kube-proxy will wait until the network comes up. Therefore, kube proxy can be started at any time. 
+
+However, if FlannelD is deployed in VXLAN (Overlay) mode, then we need to rewrite the KubeProxyConfiguration with the correct Overlay specific values, and kube-proxy will need to read this config again.
```
timothysc: You're going to need to specify the default path settings that flow through the kubelet config as well here.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+
+* Create and maintain a Powershell script to install and run Kubernetes prerequisites on Windows
+* Support kubeadm join for Windows
+* Support kubeadm reset for Windows
+* Create and maintain a Powershell script to deploy Windows CNIs
+
+### Non-Goals
+
+* Installing the Container Runtime (e.g. Docker or containerd)
+* Implement kubeadm init for Windows
```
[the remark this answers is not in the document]
ksubrmnn: Does kubeadm do this for Linux? I thought kubeadm required Docker and the Kubernetes binaries to be installed prior to kubeadm join/init

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+
+##### Alpha -> Beta Graduation
+
+Kube-proxy and CNI plugins are run as Kubernetes pods.
+The feature is maintained by active contributors.
+The feature is tested by the community and feedback is adapted with changes.
+Kubeadm join performs complete preflight checks on the host node
+E2e tests might not be complete but provide good signal.
+Documentation is in a good state. Kubeadm documentation is edited to point to documentation provided by SIG Windows.
+Kubeadm join is implemented.
```
fabriziopandini: support upgrades?

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+
+## Summary
+
+In Kubernetes 1.14, official support was added for Windows Containers. However, there is no official tool available to users for joining a Windows node to a cluster. The current solution is to create a set of scripts to download Kubernetes binaries and write their config files, but this has proven to be cumbersome and is a huge pain point for Windows adoption. On Linux, Kubeadm is available to quickly join nodes to a cluster, and the intent of this KEP is to propose a design to implement the same functionality for Windows.
+
+It should be noted that this PR only proposes enablement of support for Windows worker nodes using kubeadm.
+
+## Motivation
+
+The motivation for this KEP is to provide a tool to allow users to take a Windows machine and join it to an existing Kubernetes cluster with a single command. The user should also be able to reset the node with a single command. We will also provide the tools necessary to install prerequisites and to bring the Windows node to a Ready state.
```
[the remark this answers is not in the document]
ksubrmnn: @fabriziopandini I have the same question. I thought since the wrapper would get prereqs, we would be creating a tool AND enabling kubeadm

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+
+### User Stories
+
+#### Story 1
+
+A user will have a Windows machine that they want to join to an existing Kubernetes cluster.
+
+To start, the user will download a set of required binaries (kubelet, kube-proxy, kubeadm, kubectl, Flannel) using a script. The same script will also wrap kubeadm execution.
+
+After executing the script, the kubelet and kube-proxy will be registered as Windows services.
```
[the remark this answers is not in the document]
ksubrmnn: Updated

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
neolit123: i proposed to @ksubrmnn and @PatrickLang the path `c:\kubernetes\` as the root path of the Windows/kubeadm setup. it would hold `\etc\...`, `\var\lib...` and other paths that normally reside in the `/` on *Nix, this is not an uncommon pattern for tools ported from *Nix.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
rosti: At the current state of kubelet's component config (which is at v1beta1), there are a few command line flags that don't have a corresponding field in the config. Those are:

- dockershim related flags
- `container-runtime` && `container-runtime-endpoint` flags
- `register-with-taints` flag
- `hostname-override`

Only the `resolv-conf` flag has a representation in the component config. However, we would have to _patch_ it, after fetching the config from the config map, for the local machine setting to take place.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
rosti: I still share @benmoss 's opinion that this needs to be done by kubeadm itself.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
```diff
@@ -0,0 +1,289 @@
...
+* Support kubeadm join for Windows
+* Support kubeadm reset for Windows
+* Create and maintain a Powershell script to deploy Windows CNIs
+
+### Non-Goals
+
+* Installing the Container Runtime (e.g. Docker or containerd)
+* Implement kubeadm init for Windows
+* Implement kubeadm join --control-plane for Windows (at this time)
+* Supporting upgrades using kubeadm upgrade for Windows (to be revisited for Beta)
```
rosti: Going through the code, I think, that this would be quite easy to do for a worker node. The only kubeadm portion here that need revising is `kubeadm upgrade node config`. It has a tiny bit of OS specific code, that is shared with join and, thus, it would have been taken care of with the `kubeadm join` porting effort.

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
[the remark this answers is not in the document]
timothysc: I meant the these bullet points 

> * Implement kubeadm init for Windows (at this time) 
> * Implement kubeadm join --control-plane for Windows (at this time)

keps/sig-cluster-lifecycle/kubeadm/20190424-kubeadm-for-windows.md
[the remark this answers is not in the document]
benmoss: I think we should keep the prereqs out of anything that is part of kubeadm itself. Kubeadm is only concerned that there *exists* a service named kubelet right now, but doesn't have any opinion on how it works or how it's installed.

PR #996, opened 2019-04-24
update of KEP even pods spreading

Huang-Wei:
update of KEP even pods spreading
    
- update TopologySpreadConstraints to use a slice
- change status to `implementable`
- update "Test Plan" and "Graduation Criteria" sections

/assign @bsalamat 
/kind design
/sig scheduling


Huang-Wei:
@bsalamat PTAL.

PR #1007, opened 2019-04-26
Adding draft gc-whitelist-kep

CsatariGergely:
Signed-off-by: Gergely Csatari <gergely.csatari@nokia.com>

### DIFF
--- a/keps/sig-node/20190322-image-gc-whitelist.md
+++ b/keps/sig-node/20190322-image-gc-whitelist.md
@@ -1,0 +1,303 @@
+---
+title: Image garbage collection whitelist
+authors:
+  - "@libesz"
+  - "@CsatariGergely"
+owning-sig: sig-node
+participating-sigs:
+reviewers:
+  - "@zhangmingld"
+  - "@Monkeyanator"
+  - "@sjenning"
+  - "@Random-Liu"
+  - "@timothysc"
+approvers:
+  - sig-node-leads
+editor: "@CsatariGergely"
+creation-date: 2019-04-26
+last-updated: 2019-04-26
+status: implemented
+see-also:
+replaces:
+superseded-by:
+---
+
+# image-garbage-collector-whitelist
+
+To get started with this template:
+1. **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking up.
+  KEPs should not be checked in without a sponsoring SIG.
+1. **Make a copy of this template.**
+  Copy this template into the owning SIG's directory (or KEP root directory, as appropriate) and name it `YYYYMMDD-my-title.md`, where `YYYYMMDD` is the date the KEP was first drafted.
+1. **Fill out the "overview" sections.**
+  This includes the Summary and Motivation sections.
+  These should be easy if you've preflighted the idea of the KEP with the appropriate SIG.
+1. **Create a PR.**
+  Assign it to folks in the SIG that are sponsoring this process.
+1. **Create an issue in kubernetes/enhancements, if the enhancement will be targeting changes to kubernetes/kubernetes**
+  When filing an enhancement tracking issue, please ensure to complete all fields in the template.
+1. **Merge early.**
+  Avoid getting hung up on specific details and instead aim to get the goal of the KEP merged quickly.
+  The best way to do this is to just start with the "Overview" sections and fill out details incrementally in follow on PRs.
+  View anything marked as a `provisional` as a working document and subject to change.
+  Aim for single topic PRs to keep discussions focused.
+  If you disagree with what is already in a document, open a new PR with suggested changes.
+
+The canonical place for the latest set of instructions (and the likely source of this file) is [here](/keps/YYYYMMDD-kep-template.md).
+
+The `Metadata` section above is intended to support the creation of tooling around the KEP process.
+This will be a YAML section that is fenced as a code block.
+See the KEP process for details on each of these items.
+
+## Table of Contents
+
+- [Title](#image-garbage-collector-whitelist)
+  - [Table of Contents](#table-of-contents)
+  - [Release Signoff Checklist](#release-signoff-checklist)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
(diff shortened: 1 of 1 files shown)


keps/sig-node/20190322-image-gc-whitelist.md
```diff
@@ -0,0 +1,305 @@
...
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+Enhancing Kubelet image garbage collector with a whitelist of container images. Whitelisted images will be ignored during garbage collection.
+
+## Motivation
+
+There are some use cases when not all the container images are available from a remote repository, therefore container images are stored locally in the image cache of the Nodes. These container images can be preloaded to the Nodes instead of pulled on demand.
```
mattjmcnaughton: Could you share more about the use cases in which all the container images are not available from a remote repository? I certainly understand not wanting to host certain images on a public repository, but I'm not sure I see a time when it would make sense to avoid image repositories entirely, especially when we would need to add extra complexity to the Kubelet to compensate.

libesz: @mattjmcnaughton let me explain our use-case. We deliver our VNF products to on-prem, isolated, private customer environments, where the K8S cluster is embedded in the product. Naturally, the business logic application container images are hosted in an embedded container registry to have proper management for it (scaling, upgrade, etc.).
The container infra however, needs different approach to manage in our scenario. The nodes are 100% standalone from deployment PoV, there is no central source of content (like container images, RPMs, etc.). Everything is packaged inside a single deliverable (i.e. qcow disk image) which is able to spawn up every node according to it's role on IaaS clouds or BM hosts. This is convenient from integration and from user PoV. It is also very fast, because for example we don't have network/disk shortage during deployment due to 70 nodes are pulling down the container infra images from the same private registry. The upgrade is also performed on node level (from new qcow) for similar reasons. This way, there is no need to have special, grand preparation before, or garbage collection after all the nodes are upgraded.
When the nodes are such self-containing, there is need to exclude the containerized parts of the infrastructure from accidental garbage collection. So here we are :smile:

mattjmcnaughton: Interesting interesting :) Thanks for sharing more! Very interested to hear from the community on this, re if other people have a similar use case and how they are currently solving it (if at all).

alisondy: this would be awesome in situations where you can't afford to lose time between pulling an image again and getting it up and running if your pods die.
I've been in situations where I've had to disable the cleanup , so that some of the images don't get wiped. 🔥

CsatariGergely:
Can someone advice me what else is needed to get this KEP merged and have a better status than provisional?

keps/sig-node/20190322-image-gc-whitelist.md
alisondy: @dashpole i just max the vars out which is silly, and not actually disabling it

szuecs: @dashpole for example with Core OS Container Linux, you have toolchain, which loads a container and emergency cases you need this to load, because CoreOS just have not the tools required. I would liked to pull my toolchain container and have it whitelisted.

szuecs:
@dashpole can we get some progress here? Thanks for your time

dashpole:
@szuecs There don't appear to have been any updates since my comments, so there isn't anything new for me to review.

Random-Liu:
Just brought up an alternative idea that we can think about.

What about we assume that users can mark images as "not managed by kubelet" in the container runtime, e.g. `ctr images label` for containerd? And when Kubelet list images from the container runtime, it can see the mark, and knows that the image should be preserved?

This seems more scalable and intuitive to me.

An `image-gc-whitelist` option is not quite scalable, and doesn't quite fit into the Kubernetes' desired-actual-state reconciliation model.

keps/sig-node/20190322-image-gc-whitelist.md
BenTheElder: [kind](https://github.com/kubernetes-sigs/kind) additionally works this way, when you build & test kubernetes locally the core kubernetes images are never pushed to a registry (you wouldn't have access) they're just side-loaded into the container runtime on each of the nodes.

currently we turn off image GC and disk eviction because of this, these are short lived non-production clusters, but it is at least one use case for having images that should never be evicted :-)

szuecs:
@Random-Liu I am also fine with that. For me it does not really matter, it could also be just a string prefix search without regexp at all.
On the other hand we migrate now from CoreOS to Ubuntu and have not the problem anymore.

zhangmingld:
@Random-Liu I think the whitelist is a part of image manager , does that need to fit the Kubernetes' desired-actual-state reconciliation model ?

mtaufen:
Any updates on this? It did seem like it would be a useful idea. Even in GKE there are preloaded images (kube-proxy is one) that can't be pulled from anywhere, and if deleted it could break a node.

keps/sig-node/20190322-image-gc-whitelist.md
```diff
@@ -0,0 +1,305 @@
...
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+Enhancing Kubelet image garbage collector with a whitelist of container images. Whitelisted images will be ignored during garbage collection.
```
BenTheElder: this may also be relevant to the pause image for non-docker CRI cc @Random-Liu

BenTheElder: (currently kubelet hard codes not GCing the image it will use with dockershim, but this may not match cri-o/containerd/....)

BenTheElder:
I don't have a lot of cycles but I'd also like to see this happen FWIW, @CsatariGergely  are you still pursuing this?
/remove-lifecycle stale

moonek:
This function is useful in the operating environment.

BenTheElder:
Attempting to revive this in https://github.com/kubernetes/enhancements/pull/1717

PR #1010, opened 2019-04-27
update maxUnavailable for statefulsets

krmayankk:
/sig apps


krmayankk:
FYI @furykerry @resouer @FillZpp

FillZpp:
Great, the description of choices implementation is very clear. Also, i still recommend choice 3, it keeps balance between faster updating and ordering guarantee. 

The problem of choice 1 is that **it looks like sample, but actually does not**. Let me take an example:
1. First the statefulset with replicas 4 has pods [0-3], then update template and set maxUnavailable to 2.
2. StatefulSet controller starts to update pod 2&3. With choise 1,  controller will update remaining pods only if pod 2&3 are ready. But if pod 3 becomes ready first, currently unavailable pod is 2, how does controller know it should wait for 2 ready before updating remaining pods? 
3. You may say, controller calculates the pod updating groups with maxUavailable number, like [{0, 1}, {2, 3}]. But the question is, if replicas been changed to 5 during updating pods 2&3, then controller will split pods into [{0}, {1, 2}, {3, 4}].
4. Then, you will find pod 2 and 3, which are in different groups, are updating together. Surprise!!

krmayankk:
> You may say, controller calculates the pod updating groups with maxUavailable number, like [{0, 1}, {2, 3}]. But the question is, if replicas been changed to 5 during updating pods 2&3, then controller will split pods into [{0}, {1, 2}, {3, 4}].

Is this even possible ? The replicas cannot be scaled to 5 before the update finishes  i believe. The update knows to start from highest ordinal and count maxUnavailablePods at a time.

Note the goal is faster rolling updates. In faster rolling updates, we could potentially provide two modes, one with ordering and one which says ordering in the maxUnavailable group is not guaranteed. I dont think we have enough data to say we need the modes.  We can start with a option which makes most sense for faster updates. The consumer of faster updates would know the ordering is not guaranteed when using this feature and that is a conscious choice the user will make.

FillZpp:
> Is this even possible ? The replicas cannot be scaled to 5 before the update finishes i believe.

Actually, it can. You may test this case in your Kubernetes cluster or minikube e.g. 

So as i said before, controller with choice 1 will be confused when user modify replicas during updating.

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -125,21 +126,44 @@ type RollingUpdateStatefulSetStrategy struct {
...
-  -  pod with ordinal 4 and 3 will go down at the same time(because of maxUnavailable). When 4 is running and ready, 2 will go down. At
-     this time both 2 and 3 are terminating. If 3 is running and ready before 4, 2 wont go down to preserve ordering semantics. So at 
-     this time, only 1 is unavailable although we requested 2.
+  are the possible behavior choices we have:-
+
+  1: Pods with ordinal 4 and 3 will start Terminating at the same time(because of maxUnavailable). Once they are both running and ready, pods 
+     with ordinal 2 will start Terminating. Pods with ordinal 0 and 1 will remain untouched due the partition. In this choice, the number of pods
+     terminating is not always maxUnavailable, but sometimes less than that. For e.g. if pod with ordinal 3 is running and ready but 4 is not, we
+     still wait for 4 to be running and ready before moving on to 2. This implementation avoids out of order Terminations of pods.
+  2: Pods with ordinal 4 and 3 will start Terminating at the same time(because of maxUnavailable). When any of 4 or 3 are running and ready, pods
```
RobertKrawitz: The three alternatives are getting merged into one paragraph; needs an extra newline between each alternative.

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -125,21 +126,44 @@ type RollingUpdateStatefulSetStrategy struct {
...
+#### Recommendation
+
+I recommend Choice 1, simply because its easy to think and reason about.
+
+Other choices are introduce a flag, which controls whether we want maxUnavailable with ordering or without ordering. I dont think we need that
+complication.
+
 #### Implementation
 
+TBD: Will be updated after we have agreed on the semantics beign discussed above.
```
janetkuo: being

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -125,21 +126,44 @@ type RollingUpdateStatefulSetStrategy struct {
...
+always lead to the fastest rollouts. The maximum difference in the ordinals which are Terminating out of Order, can be more than maxUnavailable.
+
+Choice 3 always guarantees than no two pods are ever Terminating out of order. It sometimes does that, at the cost of not being able to Terminate
+maxUnavailable pods. The implementationg for this might be complicated.
+
+#### Recommendation
+
+I recommend Choice 1, simply because its easy to think and reason about.
+
+Other choices are introduce a flag, which controls whether we want maxUnavailable with ordering or without ordering. I dont think we need that
```
janetkuo: List this as Choice 4?

krmayankk:
> > Is this even possible ? The replicas cannot be scaled to 5 before the update finishes i believe.
> 
> Actually, it can. You may test this case in your Kubernetes cluster or minikube e.g.
> 
> So as i said before, controller with choice 1 will be confused when user modify replicas during updating.

We can specify that in documentation, what the behavior will be. Note that changing the grouping is not a big deal. The main goal is faster updates, whether the pod gets updated in one grouping or another doesnt matter, unless you have a good reason it matters.
Happy to hear others opinion.

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
krmayankk: add a newline

krmayankk:
Adding more people to see if they have opinions on the choices presented here 
@smarterclayton @thockin @brendandburns @kubernetes/sig-apps-feature-requests

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -48,46 +48,47 @@ The purpose of this enhancement is to implement maxUnavailable for StatefulSet d
...
 down at the same time. I want to be able to do faster rollouts by bringing down 2 or more followers at the same time. This is only possible if StatefulSet
 supports maxUnavailable in Rolling Updates.
-4. Sometimes i just want easier tracking of revisions of a rolling update. Deployment does it through ReplicaSets and has its own nuances. Understanding 
-that requires diving into the complicacy of hashing and how replicasets are named. Over and above that, there are some issues with hash collisions which 
-further complicate the situation(I know they were solved). StatefulSet introduced ControllerRevisions in 1.7 which I believe are easier to think and reason 
-about. They are used by DaemonSet and StatefulSet for tracking revisions. It would be so much nicer if all the use cases of Deployments can be met and we 
-could track the revisions by ControllerRevisions.
+4. Sometimes I just want easier tracking of revisions of a rolling update. Deployment does it through ReplicaSets and has its own nuances. Understanding 
+that requires diving into the complicacy of hashing and how ReplicaSets are named. Over and above that, there are some issues with hash collisions which
```
smarterclayton: I feel like this point should be made clear in feedback to @kubernetes/sig-apps-api-reviews - are there any supporting statements we should have here about changes that *should* happen in deployments?

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -48,46 +48,47 @@ The purpose of this enhancement is to implement maxUnavailable for StatefulSet d
...
-further complicate the situation(I know they were solved). StatefulSet introduced ControllerRevisions in 1.7 which I believe are easier to think and reason 
-about. They are used by DaemonSet and StatefulSet for tracking revisions. It would be so much nicer if all the use cases of Deployments can be met and we 
-could track the revisions by ControllerRevisions.
+4. Sometimes I just want easier tracking of revisions of a rolling update. Deployment does it through ReplicaSets and has its own nuances. Understanding 
+that requires diving into the complicacy of hashing and how ReplicaSets are named. Over and above that, there are some issues with hash collisions which 
+further complicate the situation(I know they were solved). StatefulSet introduced ControllerRevisions in 1.7 which are much easier to think and reason 
+about. They are used by DaemonSet and StatefulSet for tracking revisions. It would be so much nicer if all the use cases of Deployments can be met in 
+StatefulSet's and additionally we could track the revisions by ControllerRevisions.
```
smarterclayton: Hrm - that sounds like a dangerous statement.  Are you saying stateless deployments should be implemented by StatefulSets?

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -48,46 +48,47 @@ The purpose of this enhancement is to implement maxUnavailable for StatefulSet d
...
+same time during rolling update, would not cause issues with their Stateful Application which have per pod state and identity. Other Stateful Applications
+which cannot tolerate more than one pod going down, will resort to the current behavior of one pod at a time Rolling Updates.
 
 ### Goals
 StatefulSet RollingUpdate strategy will contain an additional parameter called `maxUnavailable` to control how many Pods will be brought down at a time,
 during Rolling Update.
 
 ### Non-Goals
-maxUnavailable is only implemeted to affect the Rolling Update of StatefulSet. Considering maxUnavailable for Pod Management Policy of Parallel is beyond 
+maxUnavailable is only implemented to affect the Rolling Update of StatefulSet. Considering maxUnavailable for Pod Management Policy of Parallel is beyond
```
smarterclayton: Why?

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
krmayankk: @smarterclayton I will reword this. What i am trying to say is that sometimes , people just want this little extra thing which statefulsets provide which is stable identity, without the storage part , but everything else is exactly similar to deployments. They stick with deployments, because statefulset doesnt have maxUnavailable . If statefulset had maxUnavailable they could move to it and additionally take advantage of better tracking of versions.

In general I do feel that the distinction between statefulset and deployment is blurring. The only advantage at this point i can think of using deployments vs statefulset is that deployments can maxSurge but statefulset's cannot because of their identity. But that is a different discussion of making them as one object.

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
krmayankk: @smarterclayton because  podManagementPolicy is only applicable during StatefulSet creation, deletion and scaling. The main use case is faster rolling updates. We could make it applicable to create/delete/scale, but that is secondary to the goal.  I thought we could just limit the scope of the initial implementation to make it less confusing, but if you think we should apply this to both, that is fine as well.

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
```diff
@@ -125,21 +126,51 @@ type RollingUpdateStatefulSetStrategy struct {
...
-     ordinal 2 will go down. Pods with ordinal 0 and 1 will remain untouched due the partition.
-  -  pods with ordinal 4 and 3 will go down at the same time(because of maxUnavailable). When any of 4 or 3 are running and ready, pods
-     with ordinal 2 will start going down. This could violate ordering guarantees, since if 3 is running and ready, then both 4 and 2
-     are terminating at the same time out of order.
-  -  pod with ordinal 4 and 3 will go down at the same time(because of maxUnavailable). When 4 is running and ready, 2 will go down. At
-     this time both 2 and 3 are terminating. If 3 is running and ready before 4, 2 wont go down to preserve ordering semantics. So at 
-     this time, only 1 is unavailable although we requested 2.
+  are the possible behavior choices we have:-
+
+  1. Pods with ordinal 4 and 3 will start Terminating at the same time(because of maxUnavailable). Once they are both running and ready, pods
```
krmayankk: @smarterclayton do you have any suggestions/opinions on which of these semantics makes more sense when implementing maxUnavailable ?

keps/sig-apps/20190226-maxunavailable-for-statefulsets.md
smarterclayton: So during upgrade, if 3 of 5 nodes are evacuated, PMP is ignored?  Since this is mostly established behavior of replicasets and deployments, you should mention *why* it's out of scope.  I am more concerned with creating something that differs between RS/Deployment and StatefulSet right now that I am about implementation ordering, so I just want the KEP to reflect that it is a) consistent with user expectations and b) justifies statements about timing with reasons

krmayankk:
@smarterclayton @kow3ns based on discussion in sig-apps meeting and offline discussion with Clayton, I have updated the proposal. Please take and look and provide feedback.

krmayankk:
@kow3ns can you approve the changes to mark it `implementable` so that I can get a Alpha in 1.16 ?

praseodym:
For reference: the Kruise third-party project [provides an advanced StatefulSet implementation with `maxUnavailable` and more](https://github.com/openkruise/kruise/blob/master/docs/concepts/astatefulset/README.md).

FillZpp:
I'm glad to see OpenKruise been noted in this KEP.  @praseodym 

[Kruise](https://github.com/openkruise/kruise) is a set of controllers which extends and complements Kubernetes core controllers on workload management, and the [AdvancedStatefulSet](https://github.com/openkruise/kruise/blob/master/docs/concepts/astatefulset/README.md) provides features like **maxUnavailable**, **inPlaceUpdate** and so on.

krmayankk:
@janetkuo @kow3ns can one of you also lgtm this ?

kerthcet:
Kindly ping @krmayankk Are we going to graduate this feature to beta in v1.26, or do we have any plans?

PR #1018, opened 2019-04-29
sig-cluster-lifecycle: update OWNER_ALIASES

neolit123:
/kind cleanup
/sig cluster-lifecycle


neolit123:
/assign @timothysc @justaugustus 
cc @justinsb

PR #1021, opened 2019-04-30
Add Test plan for watch bookmarks KEP

wojtek-t:
Ref https://github.com/kubernetes/enhancements/issues/956


lavalamp:
/lgtm
/approve

Maybe for beta / in a followup, you can mention looking for evidence that it's working as intended in the performance suite.

wojtek-t:
> Maybe for beta / in a followup, you can mention looking for evidence that it's working as intended in the performance suite.

Absolutely - it's already in graduation criteria, but I will send out a follow up once I'm back from vacation to make it clear in KEP too.

PR #1036, opened 2019-05-02
Ingress GA, updated KEP, merged review comments

bowei:
Merging comment feedback.

### DIFF
--- a/keps/sig-network/20190125-ingress-api-group.md
+++ b/keps/sig-network/20190125-ingress-api-group.md
@@ -99,13 +99,11 @@ This section describes the API fixes proposed for GA.
 1. Specify healthcheck behavior and be able to configure the healthcheck path
    and timeout.
 1. Improve the Ingress status field to be able to include additional
-   information. The current status currently only contains provisions the IP
-   address of the load balancer.
+   information. The current status currently only contains the
+   provisioned IP address(es) of the load balancer.
 1. Formalize the Ingress class annotation into a field and an associated
    `IngressClass` resource.
 1. Add support for non-Service Backend types.
-
-Note: this is a proposed list. Please discuss in the comments.
 
 ### Path as a prefix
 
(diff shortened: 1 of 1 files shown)


bowei:
I will ping when I've merged all of the comments. (I'm about halfway there)

@thockin @liggitt @M00nF1sh @aledbf

ElvinEfendi:
This PR might not the best place for this, but does the following imply every implementation has to implement healthchecking?

> If Healthcheck is nil, then the implementation default healthcheck will be configured, healthchecking the root / path. 

It's great you're introducing a separate healthchecking but given not all existing controllers (i.e ingress-nginx) do healthchecking and relies on Readiness probe and this is accepted by users would it not be better to default to whatever the implementation decides to default to?

ElvinEfendi:
I've few proposals, please let me know if you want me create a separate issue for this.

1. I love the idea of "Alternative backend types", but can we also detach backend definition from rules? That would let us to dynamically decide what backend to use for a given frontend. And would let us implement things like canaring, blue/green deploy easier.

Here's an example manifest:

```
kind: Ingress
metadata:
  name: example-ingress
spec:
  backends:
    - name: example-backend
      serviceName: service1
      servicePort: 80
    - name: example-backend-canary
      serviceName: service1-canary
      servicePort: 80
  tls:
  - hosts:
    - example.foo.com
    secretName: example-tls
  rules:
    - host: example.foo.com
      http:
        paths:
        - path: /
          backend: example-backend
```

In the world default backend would be defined using `default: true` in a backend.

2. For `rules` introduce another field to explicitly refer to the K8s secret holding TLS certificate. Otherwise currently controllers have to deduce what certificate to use based on host matching which is an extra indirection and limits in certain cases. This field would also enable HTTPs frontend provisioning.

3. `rules.http` is misleading - people use it to configure i.e a GRPC front-end. Also we have rule inside a rule (referring to the fact that you can define multiple paths in a rule). What if we introduce a new field called `protocol` in `Rule` struct and remove `http` and `paths`. So example two rules would look like:

```
- host:  example.foo.com
  path: /
  protocol: http
  backend: example-backend
  secretName: example-tls
- host:  example.foo.com
  path: /foo
  protocol: http
  backend: example-backend-foo
  secretName: example-tls
```

Next we can extend this further with HTTP verb/method:

```
- path: /
  protocol: http
  backend: example-backend-foo
  secretName: example-tls
  method: GET
```

This would configure a front-end that proxies _only_ GET requests to `example-backend-foo` backend.

When no rule is configured then the Ingress would configure a front-end accepting all HTTP requests and proxying to default backend.

M00nF1sh:
Adding to @ElvinEfendi 
We need to support for verb/header/query based too. It can be supported in Nginx and AWS LB easily.

Also,  Since we are adding support for multiple backend(service/external), can we make backend a list instead of object? We can add a weight in each entry to achieve weight based routing, which will benefit canary deployment 

BTW, I think these parts of istio's API is well designed: https://github.com/istio/api/blob/master/networking/v1alpha3/virtual_service.pb.go#L494

ElvinEfendi:
> We need to support for verb/header/query based too. It can be supported in Nginx and AWS LB easily.

Yes, this would be very useful and I've seen users requesting this in ingress-nginx.

> We can add a weight in each entry to achieve weight based routing, which will benefit canary deployment

IMHO `backend` is a wrong level of abstraction to configure weight. The knowledge of weight is needed by front-end to make routing decision. Therefore I think it'd be more appropriate to configure weight in front-end when configuring backends, for example:

```
- host:  example.foo.com
  path: /foo
  protocol: http
  backends:
    - name: example-backend-foo
      weight: 30
    - name: example-backend-foo-canary
      weight: 70
  secretName: example-tls
```

where `example-backend-foo` and `example-backend-foo-canary` are the names of backends defined in `spec.backends:`.

If we decide to support HTTP method/verb, header, request params on top of path and host in a rule we have to define their priorities as well. Maybe we can introduce an optional `priority` in Rule struct.

M00nF1sh:
@ElvinEfendi 
Yes, that's was what I'm intend to express 😄. when i mentioned `backends`, it's not reference the `spec.backends` you proposed.

To be honest, i'm not a fan to predefine backends in `spec.backends`, it introduced a another name customers need to decide/specify/match. If we do need to do name matching like this, we should add a separate resource like `ServiceBackend`, but it's more complicated than needed, I prefer to 
```
path: /
   backends:
      -  weight: 50
          backend:
              serviceName: my-svc
              servicePort: 443
      -  weight: 50
          backend:
               resource: my-gke-bucket-crd
```

ElvinEfendi:
👍  I guess it would not be necessary to predefine backends if we implement the other suggested proposals.

keps/sig-network/20190125-ingress-api-group.md
```diff
@@ -128,138 +126,127 @@ Ingress providers) treats the path match as a prefix match. For
...
+  // PathType determines the interpretation of the Path
+  // matching. PathType can be one of the following values:
+  //
+  // Prefix - matches based on a URL path prefix split
+  //   by '/'. This is not a substring match. For example:
+  //   "/foo/bar" will match "/foo/bar", "/foo/bar/abc", but
+  //   will not match "/foo/barzzz". A Prefix Path of "/" will
+  //   match any paths.
+  //
+  // ImplementationSpecific - interpretation of the Path
```
cmluciano: I assume they are here since we're adding this but could you add an example here?

keps/sig-network/20190125-ingress-api-group.md
```diff
@@ -128,138 +126,127 @@ Ingress providers) treats the path match as a prefix match. For
...
-| `x*`       | 0 or more x's           |
-| `x+`       | 1 or more x's           |
-| `xy`       | x followed by y         |
-| `x|y`      | x or y (prefer x)       |
-| `(abc)`    | grouping                |
+Interoperability between versions will be as follows:
 
-Maintaining a regular expression subset is likely not worth the complexity and
-is likely impossible across the [many implementations][regex-survey].
+1. v1beta1 will default to `ImplementationSpecific`.
```
cmluciano: Why default this to ImplementationSpecific in v1beta1 but not in v1?

keps/sig-network/20190125-ingress-api-group.md
```diff
@@ -128,138 +126,127 @@ Ingress providers) treats the path match as a prefix match. For
...
 ### API field naming
 
 These are straightforwarding one-to-one renames to for consistency.
 
-| v1beta1 field  | ga field              | rationale                   |
-|----------------|-----------------------|-----------------------------|
-| `spec.backend` | `spec.defaultBackend` | Explicitly mentions default |
+| v1beta1        | v1                    | Rationale                  |
+|----------------|-----------------------|----------------------------|
+| `spec.backend` | `spec.defaultBackend` | Explicitly mention default |
```
cmluciano: How do we rename this safely? IIRC we may need to create a new field at this point in the API process.

liggitt: fields can be renamed across version boundaries by conversion

keps/sig-network/20190125-ingress-api-group.md
cmluciano: Mentioned in sig-net meeting- this is strictly for backwards-compat for now

bowei:
arghh.. accidentally trashed this PR

PR #1061, opened 2019-05-10
apiextensions: clarify XPreserveUnknownFields type in structural schema

sttts:
This is needed so we can catch/forbid explicit false values from the user to avoid confusion


liggitt:
/lgtm

needed so we can catch/forbid explicit false values from the user to avoid confusion

deads2k:
> /lgtm
> 
> needed so we can catch/forbid explicit false values from the user to avoid confusion

fwiw, I had to ask to get a handle on this change.  If I understand correctly, this means that you can switch a subtree to true (from a top level default of false), but you cannot switch anything in any subtree back to false.  I think I now understand the change, but I may come back to, "nothing is a bool".

/approve

PR #1069, opened 2019-05-20
Runtimeclass scheduling updates

tallclair:
Follow ups from the RuntimeClass scheduling API / architecture review. Concerns raised in the review:

- The term `Topology` is too ambiguous and overloaded
- @thockin was concerned about adding new special cased scheduler predicates.

This update addresses those concerns by:

- Renaming `Topology` to `Scheduling`
- Replacing the `corev1.NodeSelector` type with `map[string]string`. This makes the merging logic much simpler, so we can mix in the node selector in admission and leave the scheduler untouched. If we find a need for the more complicated node selector logic in the future, we can revisit this decision with an additional field.

/assign @thockin @dchen1107 @bsalamat @yastij 
/sig node
/milestone v1.15
/priority important-soon

### DIFF
--- a/keps/sig-node/runtime-class-scheduling.md
+++ b/keps/sig-node/runtime-class-scheduling.md
@@ -90,11 +90,11 @@ date.
 
 ## Proposal
 
-A new optional `Topology` structure will be added to the RuntimeClass API. The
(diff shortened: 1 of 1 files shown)


thockin:
Thanks!

/approve

PR #1072, opened 2019-05-27
implement job allocation completions index enhancement

fudali113:


keps/sig-apps/20190523-support-completions-index-for-job.md
```diff
@@ -0,0 +1,78 @@
...
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+
+
+## Summary
+
+[Job](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/)
+is a Kubernetes CRD that help us run task, Job allows a user to specify the task or parallel task through either `podTemplate` and `completions`; but i use parallel task i must ensure task unrepeat in myself application, this is diffcult; if job support sahrds, i can use shards index to implement task unrepeat is so easy;
```
k82cn: Can you give an example on what's kind of workload is running in the Pod, e.g. MPI, Spark? If it's pure customized application, why not crd + operator? If

keps/sig-apps/20190523-support-completions-index-for-job.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Motivation
+
+I have a case need shards to job, i want to use kubernetes implement my case; i read the [document](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#parallel-jobs), this is a not implement function, so i want implement it;
```
k82cn: hm... it's better to have a general motivation, e.g. resolve a problem that raised by several dev/user.

keps/sig-apps/20190523-support-completions-index-for-job.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Goals
+
+- Implement each Pod is passed a different index in the range 1 to .spec.completions.
+
+## Proposal
+
+### Implementation Details/Notes/Constraints
+
+In the current implementation, the Job controller not allocation index on `.completions != nil`; in my implementation, when `.completions != nil` i will through completions value and succeeded pods get the available indexes, i will record index to the job create pods, and set env to the pod （user application get index is easy）; next i will allocation index to creating pod function, i will lock this step ensure not repeat index not more than one, if one index repeat, i will kill one on next this job created pod changes time;
```
k82cn: How do we persist the index, e.g. annotation, new field of Pod? If not persist, how to recover it when kube-controller-manager failover or change leader? 

BTW, please also highlight the interface in design doc for review, e.g. env name ?

keps/sig-apps/20190523-support-completions-index-for-job.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Risks and Mitigations
+
+The major risk with this change is the additional load on the apiserver since we need frequent each pod for get succeeded index and error index.
```
k82cn: How many "additional load" are there?

keps/sig-apps/20190523-support-completions-index-for-job.md
k82cn: btw, we implemented similar feature by CRD+Operator at https://github.com/volcano-sh/volcano/blob/master/docs/design/job-api.md#plugins-for-job which maybe helpful for your case.

fudali113: such as we have million user, each user has different user results, i need push one article to user where has some result。  

we expected run these task in one platform，that platform must easy to use and in common use，so we choice k8s；and we don`t want to use more platform(because we need lower maintenance cost)

solution is more， but i expected i need be for maintenance do something is less and i see this feature is not implement in document. i think this feature can help more people easy to use k8s.(we company use [argo](https://github.com/argoproj/argo) to implement we case, but [argo](https://github.com/argoproj/argo) more feature not to use)

keps/sig-apps/20190523-support-completions-index-for-job.md
fudali113: ok, i will addition it;

my motivation is i think shared index is a general feature in job platform, i want to easy to use job when user have relate demand and   maintenance platform less

keps/sig-apps/20190523-support-completions-index-for-job.md
fudali113: ok, i will fix it;

i persist the index to labels, i choice label because i think easy to use on `kubectl get po -l`, label name is `job-completions-index` and i set `JOB_COMPLETIONS_INDEX ` env in pod (user get this index in application easier to use )

keps/sig-apps/20190523-support-completions-index-for-job.md
fudali113: 1. group with index on activePods
2. remove duplicate index activePods
3. allocation index need an lock

The above three points is additional operation，i think that additional load is acceptable. and i don`t know how to get how many additional load is better, can you help me ?

this is my first pr in k8s, maybe there's something wrong with it， look forward to your comments @k82cn

keps/sig-apps/20190523-support-completions-index-for-job.md
k82cn: is there others have similar case with yours?

fudali113: we company has this case and i think this is a common question. because this implement can make job application easy to scale.

but my issue not more people comment support this implement.

@k82cn

fudali113:
@k82cn @prydonius @kow3ns PTAL

keps/sig-apps/20190523-support-completions-index-for-job.md
k82cn: @fudali113 , IMO, that's a specific case, and prefer to handle that by CRD :)

fengzixu:
Is this enhancement still being handling? It seems like there is no response from the author for several months.  If it is possible, I can re-create the enhancement like this and implement it.

@k82cn @palnabarun @kow3ns @janetkuo

mattfarina:
@fudali113 adding to the agenda for the 2/24 SIG Apps meeting to discuss if you can make it.

fudali113:
> @fudali113 adding to the agenda for the 2/24 SIG Apps meeting to discuss if you can make it.

sorry，i just saw this comment now；What can i do to continue this PR ?

PR #1077, opened 2019-05-30
PV health monitor KEP

NickrenREN:
Moving https://github.com/kubernetes/community/pull/1484 here

### DIFF
--- a/keps/sig-storage/20190530-pv-health-monitor.md
+++ b/keps/sig-storage/20190530-pv-health-monitor.md
@@ -1,0 +1,575 @@
+---
+title: PV Health Monitor
+authors:
+  - "@NickrenREN"
+  - "@xing-yang"
+owning-sig: sig-storage
+reviewers:
+  - "@msau42"
+  - "@jingxu97"
+  - "@liggitt"
+  - "@gnufied"
+approvers:
+  - "@msau42"
+  - "@saad-ali"
+  - "@thockin"
+  - "@liggitt"
+editor: TBD
+creation-date: 2019-05-30
+last-updated: 2020-01-23
+status: implementable
+
+---
+
+# PV Health Monitor
+
+## Table of Contents
+
+<!-- toc -->
+- [Motivation](#motivation)
+- [User Experience](#user-experience)
+  - [Use Cases](#use-cases)
+- [Proposal](#proposal)
+- [Implementation](#implementation)
+  - [CSI changes](#csi-changes)
+    - [Add GetVolume RPC](#add-getvolume-rpc)
+    - [Add Node Volume Health Function](#add-node-volume-health-function)
+  - [External controller](#external-controller)
+    - [CSI interface](#csi-interface)
+    - [Node down event](#node-down-event)
+  - [External node agent](#external-node-agent)
+    - [CSI interface](#csi-interface-1)
+  - [Alternatives](#alternatives)
+    - [Alternative option 1](#alternative-option-1)
+    - [Alternative option 2](#alternative-option-2)
+    - [Alternative option 3](#alternative-option-3)
+    - [Optional HTTP(RPC) service](#optional-httprpc-service)
+- [Graduation Criteria](#graduation-criteria)
+  - [Alpha -&gt; Beta](#alpha---beta)
+  - [Beta -&gt; GA](#beta---ga)
+- [Test Plan](#test-plan)
+  - [Unit tests](#unit-tests)
+  - [E2E tests](#e2e-tests)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Motivation
+
+Currently there is no way to monitor Persistent Volumes after they are provisioned in Kubernetes. This makes it very hard to debug and detect root causes when something happens to the volumes. An application may detect that it can no longer write to volumes. However, the root cause happened at the underlying storage system level. This requires application and infrastructure team jointly debug and find out what has triggered the problem. It could be that the volume runs out of capacity and needs an expansion. It could be that the volume was deleted by accident outside of Kubernetes. In the case of local storage, the local PVs may not be accessed any more due to the nodes break down. In either case, it will take lots of effort to find out what has happened at the infrastructure layer. It will also take complicated process to recover. Admin usually needs to intervene and fix the problem.
+
+With health monitoring, unhealthy volumes can be detected and fixed early and therefore could prevent more serious problems to occur. While some problems may be corrected automatically by a Kubernetes controller, most problems may involve manual admin intervention or need to be fixed with specific application knowledge. In any case, volume health monitoring is a very important tool that would be beneficial to Kubernetes users.
+
+## User Experience
+
+### Use Cases
+
+Many things could happen to the underlying storage system after a volume is provisioned in Kubernetes.
+
+* Volume could be deleted by accident outside of Kubernetes.
+* The disk that the volume resides on could be removed temporarily for maintenance.
+* The disk that the volume resides on could fail.
+* There may be configuration issues with the underlying storage system that affect the volume’s health.
+* Volume could be out of capacity.
+* The disk may be degrading which affects its performance.
+
+If the volume is mounted on a pod and used by an application, the following problems could also happen.
+* There may be read/write I/O errors.
+* The file system on the volume may be corrupted.
+* Filesystem may be out of capacity.
+* Volume may be unmounted by accident outside of Kubernetes.
+
+If the CSI driver has implemented the CSI volume health function proposed in this design document, Kubernetes could communicate with the CSI driver to retrieve any errors detected by the underlying storage system. Kubernetes can report an event and log an error about this PVC so that user can inquire this information and decide how to handle them. For example, if the volume is out of capacity, user can request a volume expansion to get more space. In the first phase, volume health monitoring is informational only as it is only reported in the events and logs. In the future, we will also look into how Kubernetes may use volume health information to automatically reconcile.
+
+There could be conditions that cannot be reported by a CSI driver. One or more nodes where the volume is attached to may be down. This can be monitored and detected by the volume health controller so that user knows what has happened.
+
+The Kubernetes components that monitor the volumes and report events with volume health information include the following:
+
+* An external monitoring node agent on each kubernetes worker node.
+* An external monitoring controller on the master node.
+
+Details will be described in the following proposal section.
+
+## Proposal
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+The following area will be the focus of this proposal at first:
+
+* Provide a mechanism for CSI drivers to report volume health problems at the controller and node levels.
+
+Two main parts are involved here in the architecture.
+
+- External Controller:
+  - The external controller will be deployed as a sidecar together with the CSI controller driver, similar to how the external-provisioner sidecar is deployed.
+  - Trigger controller RPC to check the health condition of the CSI volumes.
+  - The external controller sidecar will also watch for node failure events.
+
+- External Node Agent:
+  - The external node agent will be deployed as a sidecar together with the CSI node driver on every Kubernetes worker node.
+  - Trigger node RPC to check volume's mounting conditions.
+  - Note that currently we do not have CSI support for local storage. When the support is available, we will implement relavant CSI monitoring interfaces as well.
+
+## Implementation
+
+### CSI changes
+
+Container Storage Interface (CSI) specification will be modified to provide volume health check leveraging existing RPCs and adding new ones.
+
+- Add new GetVolume controller RPC
+  - External controller calls a new `GetVolume` RPC to check health condition of a particular PV instead of calling `ListVolumes` with a `volume_id` filter. Some CSI drivers chose not to implement `ListVolumes` as it is an expensive operation. Therefore it is better to introduce a new RPC `GetVolume` which has a new capability.
+- Extend the existing NodeGetVolumeStats RPC
+  - For any PV that is mounted, the external node agent calls `NodeGetVolumeStats` to see if volume is still mounted;
+  - Calls `NodeGetVolumeStats` to check if volume is usable, e.g., filesystem corruption, bad blocks, etc.
+
+Two new controller capabilities are added. If a CSI driver supports `GET_VOLUME` capability, it MUST support calling `GetVolume` to provide general volume information in `GetVolumeResponse`. If a driver supports `GET_VOLUME_HEALTH`, it MUST provide additional health information in `GetVolumeResponse`.
+
+A new node capability is added. If CSI driver supports the `GET_VOLUME_STATS_HEALTH` capability, it MUST provide health information in `NodeGetVolumeStats`.
+
+Detailed changes needed in the CSI Spec are described in the following.
+
+#### Add GetVolume RPC
+
+Add a new `GetVolume` RPC. `GetVolumeResponse` should contain current information of a volume if it exists. If the volume does not exist any more, `GetVolume` should return gRPC error code `NOT_FOUND`.
+
+Add `VolumeHealth` message in `VolumeStatus` field in `GetVolumeResponse`. This can indicate additional health information for the volume.
+
+Whether the volume is still attached to a node is already handled by `LIST_VOLUMES_PUBLISHED_NODES` controller capability. This capability will be used by Kubernetes to reconcile - re-attach the volume if the volume is not attached any more.
+
+```
+message GetVolumeRequest {
+  // Identity information for a specific volume. This field is
+  // REQUIRED. GetVolume will return with current volume information.
+  string volume_id = 1;
+}
+```
+
+```
+message GetVolumeResponse {
+  message VolumeStatus{
+    // health shows error conditions reported by the SP.
+    // This field MUST be specified if the
+    // GET_VOLUME_HEALTH controller capability is supported.
+    repeated VolumeHealth health = 1;
+  }
+
+  // This field is REQUIRED
+  Volume volume = 1;
+
+  // This field is OPTIONAL. This field MUST be specified if the
+  // GET_VOLUME_HEALTH controller capability is supported.
+  VolumeStatus status = 2;
+}
+```
+
+```
+message VolumeHealth {
+  // The error code describing the health condition of the volume.
+  // This field is REQUIRED.
+  string error_code = 1;
+
+  // The error message associated with the above error_code. This field is OPTIONAL.
+  string message = 2;
+}
+```
+
+The following common error codes are proposed for volume health:
+* VolumeNotFound
+* OutOfCapacity
+* DiskFailure
+* DiskRemoved
+* ConfigError
+* NetworkError
+* DiskDegrading
+* VolumeUnmounted
+* RWIOError
+* FilesystemCorruption
+
+Add `GET_VOLUME` and `GET_VOLUME_HEALTH` controller capabilities.  If a driver supports `GET_VOLUME` capability, it MUST implement the `GetVolume` RPC.  If a driver supports `GET_VOLUME_HEALTH`, it MUST supports the `health` field in the `status` field in `GetVolumeResponse`.
+
+```
+// Specifies a capability of the controller service.
+message ControllerServiceCapability {
+  message RPC {
+    enum Type {
+      UNKNOWN = 0;
+      CREATE_DELETE_VOLUME = 1;
+      PUBLISH_UNPUBLISH_VOLUME = 2;
+      LIST_VOLUMES = 3;
+      GET_CAPACITY = 4;
+      // Currently the only way to consume a snapshot is to create
+      // a volume from it. Therefore plugins supporting
+      // CREATE_DELETE_SNAPSHOT MUST support creating volume from
+      // snapshot.
+      CREATE_DELETE_SNAPSHOT = 5;
+      LIST_SNAPSHOTS = 6;
+
+      // Plugins supporting volume cloning at the storage level MAY
+      // report this capability. The source volume MUST be managed by
+      // the same plugin. Not all volume sources and parameters
+      // combinations MAY work.
+      CLONE_VOLUME = 7;
+
+      // Indicates the SP supports ControllerPublishVolume.readonly
+      // field.
+      PUBLISH_READONLY = 8;
+
+      // See VolumeExpansion for details.
+      EXPAND_VOLUME = 9;
+
+      // Indicates the SP supports the
+      // ListVolumesResponse.entry.published_nodes field
+      LIST_VOLUMES_PUBLISHED_NODES = 10;
+
+      // Indicates the SP supports the GetVolume RPC
+      GET_VOLUME = 11;
+
+      // Indicates the SP supports the
+      // GetVolumeResponse.volume_health field
+      GET_VOLUME_HEALTH = 12;
+    }
+
+    Type type = 1;
+  }
+
+  oneof type {
+    // RPC that the controller supports.
+    RPC rpc = 1;
+  }
+}
+```
+
+#### Add Node Volume Health Function
+
+Node Volume Health checks if a volume is still mounted and usable. To check whether a volume is usable, the CSI driver shall check if filesystem is corrupted, whether there are bad blocks, etc. in this RPC.
+
+Instead of adding a new RPC, we can leverage the existing NodeGetVolumeStats RPC.
+
+```
+rpc NodeGetVolumeStats (NodeGetVolumeStatsRequest)
+    returns (NodeGetVolumeStatsResponse) {}
+```
+
+In the NodeGetVolumeStatsRequest, there are `volume_id`, `volume_path`, and `staging_target_path`. CSI driver can figure out whether a volume is still mounted based on these parameters.
+
+In NodeGetVolumeStatsResponse, there is already a `VolumeUsage` which includes information (available, used, and total) of the file system on the volume.
+
+A new message `VolumeHealth` will be added to `NodeGetVolumeStatsResponse`. A new Node capability `VOLUME_STATS_HEALTH` will be added to indicate whether a CSI driver has implemented this function. In a `VolumeHeath` message, there is an error code indicating the type of code and a message that describes the details of the error.
+
+```
+message NodeGetVolumeStatsRequest {
+  // The ID of the volume. This field is REQUIRED.
+  string volume_id = 1;
+
+  // It can be any valid path where volume was previously
+  // staged or published.
+  // It MUST be an absolute path in the root filesystem of
+  // the process serving this request.
+  // This is a REQUIRED field.
+  string volume_path = 2;
+
+  // The path where the volume is staged, if the plugin has the
+  // STAGE_UNSTAGE_VOLUME capability, otherwise empty.
+  // If not empty, it MUST be an absolute path in the root
+  // filesystem of the process serving this request.
+  // This field is OPTIONAL.
+  string staging_target_path = 3;
+}
+```
+
+```
+message NodeGetVolumeStatsResponse {
+  // This field is OPTIONAL.
+  repeated VolumeUsage usage = 1;
+  // This field is OPTIONAL.
+  repeated VolumeHealth volume_health = 1;
+}
+```
+
+```
+message VolumeUsage {
+  enum Unit {
+    UNKNOWN = 0;
+    BYTES = 1;
+    INODES = 2;
+  }
+  // The available capacity in specified Unit. This field is OPTIONAL.
+  // The value of this field MUST NOT be negative.
+  int64 available = 1;
+
+  // The total capacity in specified Unit. This field is REQUIRED.
+  // The value of this field MUST NOT be negative.
+  int64 total = 2;
+
+  // The used capacity in specified Unit. This field is OPTIONAL.
+  // The value of this field MUST NOT be negative.
+  int64 used = 3;
+
+  // Units by which values are measured. This field is REQUIRED.
+  Unit unit = 4;
+}
+```
+
+```
+message VolumeHealth {
+  // The error code describing the health condition of the volume.
+  // This field is REQUIRED.
+  string error_code = 1;
+
+  // The error message associated with the above error_code. This field is OPTIONAL.
+  string message = 2;
+}
+```
+
+Add a `GET_VOLUME_STATS_HEALTH` node capability. If driver supports this capability, it MUST fetch `volume_health` information in `NodeGetVolumeStats`.
+
+```
+message NodeGetCapabilitiesRequest {
+  // Intentionally empty.
+}
+```
+
+```
+message NodeGetCapabilitiesResponse {
+  // All the capabilities that the node service supports. This field
+  // is OPTIONAL.
+  repeated NodeServiceCapability capabilities = 1;
+}
+```
+
+```
+// Specifies a capability of the node service.
+message NodeServiceCapability {
+  message RPC {
+    enum Type {
+      UNKNOWN = 0;
+      STAGE_UNSTAGE_VOLUME = 1;
+      // If Plugin implements GET_VOLUME_STATS capability
+      // then it MUST implement NodeGetVolumeStats RPC
+      // call for fetching volume statistics.
+      GET_VOLUME_STATS = 2;
+      // See VolumeExpansion for details.
+      EXPAND_VOLUME = 3;
+      // If Plugin implements GET_VOLUME_STATS_HEALTH capability
+      // then it MUST implement NodeGetVolumeStats RPC
+      // call for fetching volume health information.
+      GET_VOLUME_STATS_HEALTH = 4;
+    }
+
+    Type type = 1;
+  }
+
+  oneof type {
+    // RPC that the controller supports.
+    RPC rpc = 1;
+  }
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves. The frequency of the check should be tunable. A configure option will be available in the external controller to adjust this value.
+
+#### Node down event
+* External monitoring controller will check if node is marked as unresponsive by the node controller.
+* The external monitoring controller will track which pods are using which PVCs and what nodes they got scheduled to.
+* In the case that a node goes down, the controller will report an event for all PVCs on that node.
+
+### External node agent
+
+#### CSI interface
+Call NodeGetVolumeStats() RPC to check the mounting and other health conditions. The frequency of the check should be tunable. A configure option will be available in the external node agent to adjust this value.
+
+The external node agent will go through all the pods on that node and find out all volumes used by those pods. It will watch all those volumes and call NodeGetVolumeStatus() to find out their health status.
+
+### Alternatives
+
+#### Alternative option 1
+
+The Status field in the PVC will be used to mark volumes if they are unhealthy. The external monitoring controller sidecar will be responsible of monitoring the volumes and updating the PVC status field when needed.
+
+```
+// PersistentVolumeClaimHealthConditionType defines the health condition of PV claim.
+// Valid values are "HealthFailure", "HealthWarning", "HealthUnknown".
+type PersistentVolumeClaimHealthConditionType string
+
+// These are valid health conditions of PVC
+const (
+        // PersistentVolumeClaimHealthFailure - Volume health failure indicates a severe problem that makes the volume unusable
+        PersistentVolumeClaimHealthFailure PersistentVolumeClaimHealthConditionType = "HealthFailure"
+        // PersistentVolumeClaimHealthWarning - Volume health warning indicates there is a problem but volume is still usable
+        PersistentVolumeClaimHealthWarning PersistentVolumeClaimHealthConditionType = "HealthWarning"
+        // PersistentVolumeClaimHealthUnknown - Volume health unknown indicates the health condition of the volume is unknown
+        PersistentVolumeClaimHealthUnknown PersistentVolumeClaimHealthConditionType = "HealthUnknown"
+)
+
+// PersistentVolumeClaimHealthCondition represents the current health condition of PV claim
+type PersistentVolumeClaimHealthCondition struct {
+        Type   PersistentVolumeClaimHealthConditionType
+        Status ConditionStatus
+        ErrorCode string
+        // +optional
+        LastProbeTime metav1.Time
+        // +optional
+        LastTransitionTime metav1.Time
+        // +optional
+        Reason string
+        // +optional
+        Message string
+}
+
+// PersistentVolumeClaimStatus represents the status of PV claim
+type PersistentVolumeClaimStatus struct {
+        // Phase represents the current phase of PersistentVolumeClaim
+        // +optional
+        Phase PersistentVolumeClaimPhase
+        // AccessModes contains all ways the volume backing the PVC can be mounted
+        // +optional
+        AccessModes []PersistentVolumeAccessMode
+        // Represents the actual resources of the underlying volume
+        // +optional
+        Capacity ResourceList
+        // +optional
+        Conditions []PersistentVolumeClaimCondition
+        // +optional
+        HealthConditions []PersistentVolumeClaimHealthCondition
+}
+
+// ConditionStatus defines conditions of resources
+type ConditionStatus string
+
+// These are valid condition statuses. "ConditionTrue" means a resource is in the condition;
+// "ConditionFalse" means a resource is not in the condition; "ConditionUnknown" means kubernetes
+// can't decide if a resource is in the condition or not. In the future, we could add other
+// intermediate conditions, e.g. ConditionDegraded.
+const (
+        ConditionTrue    ConditionStatus = "True"
+        ConditionFalse   ConditionStatus = "False"
+        ConditionUnknown ConditionStatus = "Unknown"
+)
+```
+
+To avoid stale information being stored on a PVC, each periodic update will mark the PVC with the latest health information, replacing previous health information added by the external controller. If the PVC becomes healthy after being marked as unhealthy previously, the controller should remove the previous information.
+
+As mentioned earlier, reaction is not in the scope of this proposal but will be considered in the future. Before reacting to any negative health condition, the controller responsible for the reaction should call GetVolume() again to ensure the information is update to date.
+
+Alternative option 1 is not in the main proposal because it requires giving all nodes access to a credential that has the ability to modify all PVCs. This is not desirable as it allows a malicious node to put an unhealthy status on a PVC.
+
+#### Alternative option 2
+
+If the agent on the node cannot be used to modify the PVC status and the monitoring logic cannot be added to Kubelet directly, we can introduce a CRD to represent the volume health. This volume health CRD is in the same namespace as the PVC that it is monitoring. It contains the PVC name and health conditions as defined in the main option. It needs to have a one on one mapping with the PVC. In the PVC status, there will be a `volumeHealthName` field pointing back to the volume health CRD.
+
+Both the controller and the agent can create the volume health CRD for a PVC if it does not exist yet. Only one volume health CRD should be created for a PVC. Only the controller can set the `volumeHealthName` field in the PVC status.
+
+```
+type VolumeHealth struct {
+        metav1.TypeMeta
+        metav1.ObjectMeta
+	Spec VolumeHealthSpec
+	// +optional
+	Status *VolumeHealthStatus
+}
+
+type VolumeHealthSpec struct {
+	// The PVC name that this VolumeHealth object is monitoring
+	PersistentVolumeClaimName string
+}
+
+// Note: PersistentVolumeClaimHealthCondition is already defined in the
+// main section.
+type VolumeHealthStatus struct {
+        HealthConditions []PersistentVolumeClaimHealthCondition
+}
+
+// PersistentVolumeClaimStatus represents the status of PV claim
+type PersistentVolumeClaimStatus struct {
+        // Phase represents the current phase of PersistentVolumeClaim
+        // +optional
+        Phase PersistentVolumeClaimPhase
+        // AccessModes contains all ways the volume backing the PVC can be mounted
+        // +optional
+        AccessModes []PersistentVolumeAccessMode
+        // Represents the actual resources of the underlying volume
+        // +optional
+        Capacity ResourceList
+        // +optional
+        Conditions []PersistentVolumeClaimCondition
+        // +optional
+        VolumeHealthName *string
+}
+```
+
+Alternative option 2 does not address the concerns from option 1. Assuming the node-level sidecar is running in a pod as a service account, we need to grant that service account permission to modify any instance of the volume health CR, and then the monitoring controller just copies that into the PVC status.
+
+#### Alternative option 3
+
+We can also reuse the PV Taints and introduce a new Taint called PVUnhealthMessage for PV health condition whose key is specific (PVUnhealthMessage) and value can be set differently. 
+
+For example, if the PV is not attached now, we can mark the PV using the PVUnhealthMessage taint like this:
+```
+Key: “PVUnhealthMessage”
+Value: “AttachError - the pv is not attached to node1 now”
+VolumeTaintEffect: NoEffect
+```
+
+If the volume is deleted, we can mark the PV using the PVUnhealthMessage taint like this:
+```
+Key: “PVUnhealthMessage” 
+Value: “VolumeError - the volume is deleted from backend”
+VolumeTaintEffect: NoEffect
+```
+
+Note that:
+
+- all the VolumeTaintEffects are NoEffect now at first, we may talk about the reactions later in another proposal;
+- the taint Value is string now, it is theoretically possible that several errors are detected for one PV, we may extend the string to cover this situation: combine the errors together and splited by semicolon or other symbols.
+
+This was initially brought up to address the external data populator use case, but there were lots of concerns about this proposal. We do not want our main proposal to depend on something that is not approved.
+
+#### Optional HTTP(RPC) service
+
+Create a HTTP(RPC) service to receive volume health condition reports from other components.
+
+Users can extend the volume health condition monitoring by setting up their own detector and report the result to the service.
+
+This optional service is out of scope for this KEP but may be introduced in a future KEP if needed.
+
+This option is not in the main proposal because it is a push-based method while CSI uses pull-based method.
+
+## Graduation Criteria
+### Alpha -> Beta
+* Feature complete, including:
+  * External controller volume health monitoring.
+  * External node agent volume health monitoring.
+* Tests outlined in design proposal implemented.
+
+### Beta -> GA
+* Volume health support is added to multiple CSI drivers.
+* Volume health feature deployed in production and have gone through at least one K8s upgrade.
+
+## Test Plan
+### Unit tests
+* Unit tests for external controller and external node agent volume health monitoring. The following error codes will be simulated in the unit tests:
+  * VolumeNotFound
+  * OutOfCapacity
+  * VolumeUnmounted
+  * NodeDown
+
+### E2E tests
+* e2e tests for external controller and external node agent volume health monitoring. Hostpath CSI driver and GCE PD driver will be used for e2e tests. The following error codes will be tested in the e2e tests:
+  * VolumeNotFound
+  * OutOfCapacity
+  * VolumeUnmounted
+* Add stress and scale tests before moving from beta to GA.
+
+## Implementation History
+
+- 20191021: KEP updated
+
+- 20190730: KEP updated
+
+- 20190530: KEP submitted
+
+- Demo implementation (using annotations): 
+https://github.com/NickrenREN/kubemonitor/tree/master/build/kube_storage_monitor/local_monitor
+https://github.com/NickrenREN/kubemonitor/tree/master/build/kube_storage_monitor/node_watcher


cdickmann:
This KEP is quite interesting to me, and I think this topic is quite important, so I am curious about your answers.

Some questions about the goals:
Q1: The goals mention Local PVs several times, but I felt the same health monitoring needs apply to any PV. Remote PVs can also suffer network partitions, temporary failure, permanent failure, IO issues leading to mount issues, etc. What do you think? Would it make sense to update the KEP to be less Local PV centric?
Q2: Do I understand correctly that the purpose of the health reporting is to allow for automated actions to be taken, e.g. by a database operator or Statefulset controller? In contrast to providing insights aimed at human users. it would be great to clarify this in the goals.

I also have some questions/suggestions on the concrete proposal:
Q3: The API change calls for 3 specific taints. Would this be extensible just like Node taints, i.e. would the use of custom taint keys be supported?
Q4: I was curious if you had considered a few more conditions, which I couldn’t map to the currently proposed conditions. 
+ One would be “evacuation suggested”, which could either be because the underlying provider wishes to perform maintenance and will take back the volume, or because it was detected that the physical media is degrading, and so evacuation may still be possible, but performance may already be impacted. 
+ Another would be to differentiate “Absent” and “Failed” underlying media when NotUsable it hit. My background is more bare metal software defined storage, and there it is a frequent situation that a physical drive is physically removed, as opposed to having failed. In network attached there is similarly the condition of “network partition” versus media failure. In my experience the distinction matters because the likelihood that a volume comes back from network disconnect is high, while coming back from “failure” is low (turns out it is not 0 though, because IO failures may be related to a bad firmware which a host reset can fix). 
+ Finally, it may be valuable to expose that a volume is experiencing conditions of “reduced redundancy” or “temporarily reduced performance” due to physical media failure that is being automatically remediated. Applications could choose to direct less traffic to these volumes when possible.

xing-yang:
Hi @cdickmann, thanks for your comments!  In addition to what @NickrenREN has replied to you on Slack, I want to add that I have an action item to add the CSI support in this KEP, which should cover the remote PV part.  In general new features will only be added to support CSI drivers.

NickrenREN:
Hi, @cdickmann thanks for copying your questions/suggestions here.
Below is my reply on slack.
```
Hi, Christian,
Thanks for your questions and suggestions on the monitor PR. I am so happy that you are interested in it.

Q1: yes, local PVs as well as remote PVs need health monitor.

Q2: The health reportings aim to provide info for reaction, but reaction is more complex, it is hard for us to find a suitable solution for all workloads. So, reaction needs further discussion, i am not sure if reactions should be taken automatically or manually . We can discuss offline, and it is not in the scope of the monitor proposal.

Q3: Regarding the taints, we try to integrate failure conditions into as few taints as possible. So that CSI support becomes more easily. And of course, we can and need to expand the taints if needed.

And finally, regarding the conditions you mentioned, if the PV is “unhealthy” now, the causes may be different and hard to list all the causes. So we choose to integrate them all together into one which is `NotUsable`

And I agree that extension mechanism is important, if we do want to differentiate the causes, we need to find a way to achieve that. I will rethink this, thanks a lot.
```

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,200 @@
...
+
+For networked storage, we just need to create a **controller** to check the volumes’ status periodically, and each volume plugin should implement its own health checking methods. 
+Since we are migrating in-tree volume plugins out of tree, the health checking methods should be supported in CSI spec. Detailed CSI spec info is [here](https://docs.google.com/document/d/1mjYlgXflAayLMPVFNT-qf3iJTVJYJBD-2RQh1jZTJ6s/edit#heading=h.3uxdz6kyha0s).
+
+The whole architecture is as below:
+![pv health monitor architecture](./pv-health-monitor.png)
+
+
+## Implementation
+
```
richardelling: in the case of a reserved volume: shared between multiple nodes, but RWO and this node does not own the reservation, do we need such a state? The volume could be functioning perfectly well, but is not available to this node at the moment.

NickrenREN: Good question, if the volume itself is problematical, we should definitely taint it.
If the volume is shared between multiple nodes and RWX, and because of network problem or sth like that, it is not available to one specific node, i think we need to taint it, it is an attaching problem.
But if it is RWO, and the node does not own the reservation, IIUC, the volume is healthy and will not lead to data loss, as you mentioned above, it may lead to reduced performance.I prefer not tainting it at first.
It seems that we need to list the specific causes. I will rethink the extension mechanism, and validate if Taint is suitable here.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,362 @@
...
+When we finish discussing the reaction for unhealthy PVs, we need to revisit all the VolumeTaintEffects.
+
+### Networked storage
+
+For networked storage, the following volume health conditions will be checked:
+* Checks if volume still exists and/or is attached
+* Checks if volume is still mounted and usable
+
+Container Storage Interface (CSI) specification will be modified to add two RPCs ControllerCheckVolume and NodeCheckVolume. A monitor controller (external-monitor) that is responsible for watching the PersistentVolumeClaim, PersistentVolume, and VolumeAttachment API objects will be added. The external-monitor can be deployed as a sidecar container together with the CSI driver.  For PVs, the external-monitor calls CSI driver’s monitor interface to check volume health status.
```
cdickmann: Why is it necessary to deploy a new daemon and further add to complexity? Intuitively, it feels to me that this should be able to fit into existing daemons, as it just extends CSI slightly to cover more state. CSI already communicates multiple pieces of state in both directions, and this is a natural extension to me.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,362 @@
...
+
+* ControllerCheckVolume RPC
```
cdickmann: This seems to define a poll model. Would it also be possible for a push mechanism? I would assume that many storage systems underlying the CSI would actively notice and react to issues, and be able to notify upper layers in an expedient way. Thus, it should be possible for the CSI to communicate this. This is similar to the SCSI device specs, where a physical drive or physical SCSI controller can raise a notification when it detects an issue, so that upper layers can take immediate action. Being expedient can be important for applications that depend on application level replication and needing to know when to kick in rebuilds to reduce downtime.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Currently for every new feature added in kubernetes-csi, a new sidecar container will be added accordingly.  For example, we have external-provisioner, external-attacher, external-snapshotter, and external-resizer, etc.  This way it is easier to maintain each component, but I do see your point that this adds more complexity as well.  I'll think more about this.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: That's a good point.  Let me think about this.

NickrenREN:
Hi, @xing-yang @cdickmann @richardelling 
FYI, i updated the KEP, please take a look when you get a chance, thanks

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -34,235 +34,172 @@ status: provisional
...
 ```
 
-This can be used for both device mount and volume mount checking. If the volume is not mounted, we need to taint it.
-
-For in use volumes, if the volumes are not mounted, reactions should be taken. 
+Note that:
 
-The reactions needs further discussion, and is not in the scope of this doc. 
+- all the VolumeTaintEffects are NoEffect now at first, we may talk about the reactions later in another proposal;
+- the taint Value is string now, it is theoretically possible that several errors are detected for one PV, we may extend the string to cover this situation: combine the errors together and splited by semicolon or other symbols.
```
xing-yang: If there is only one key and values are in free text form, it will be difficult to add reactions later based on those information.  Should we keep the keys such as "VolumeNotAttached" or "VolumeNotMounted" and define a few constant string values, i.e., VOLUME_DELETED, FILESYSTEM_CORRUPTED?  Maybe allow user to define their own custom values too but those will always have "NoEffect".

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -34,235 +34,172 @@ status: provisional
...
-  * If volume is attached, it calls ControllerCheckVolume() to see if volume is still attached; Taints PV with VolumeNotAttached if not attached any more.
-
-* NodeCheckVolume RPC
-  * For any PVC that is mounted, the external-monitor calls NodeCheckVolume() to see if volume is still mounted; Taints PV with VolumeNotMounted if not mounted any more.
-  * Calls NodeCheckVolume to check if volume is usable, e.g., filesystem corruption, bad blocks, etc; Taint PV with VolumeNotUsable if volume is not usable.
+- ControllerCheckVolume RPC
+  - External controller calls ControllerCheckVolume() to check the health condition of PVs themselves, for example, if the PVs are deleted ...
+  - Move Attach checking here ?
+- NodeCheckVolume RPC
+  - If volume is attached to this node, the external agent calls NodeCheckVolume() to see if volume is still attached; Move this check to  ControllerCheckVolume ?
```
xing-yang: We can keep this here.  In the case of multi-attach, it is easier to check on each node as discussed earlier.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,484 @@
...
+  // A token to specify where to start paginating. Set this field to
+  // `next_token` returned by a previous `ListVolumes` call to get the
+  // next page of entries. This field is OPTIONAL.
+  // An empty string is equal to an unspecified field value.
+  string starting_token = 2;
+
+  // Identity information for a specific volume. This field is
+  // OPTIONAL. It can be used to list only a specific volume.
+  // ListVolumes will return with current volume information.
+  string volume_id = 3;
```
xing-yang: new RPC GetVolume()

yoesmat:
Adding a link to this since this looks like a good use-case for volume health: https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/pull/126#issuecomment-574522023

My feeling is that where-ever possible we should try to deal with the failures automatically in the PV controller/attach-detach controller/CSI driver/Kubelet. But it would definitely be useful to know the current state of the volume by looking at the PV/PVC.

NickrenREN:
Thanks for the link @yoesmat 

>But it would definitely be useful to know the current state of the volume by looking at the PV/PVC.

yes, application developers are always the first ones to discover the volume problem(application io error or sth like that) and they usually have no access to debug directly in infrastructure(k8s cluster and storage cluster behind). Application developers need to ask infra admin to help. If we have the monitor already reported the problem, infra admin will find the cause more easily.

>My feeling is that where-ever possible we should try to deal with the failures automatically in the PV controller/attach-detach controller/CSI driver/Kubelet.

It definitely would be great if we **can** auto correct the volume states. 
But reaction is hard to define, different use cases may need different reactions. Auto correcting is cool ,but hard to cover different scenarios. Monitor metrics provide another option to choose(dealing with the problem manually or developing external handler depending on the metrics).

xing-yang:
@liggitt Hi Jordon, can you please take a look of this KEP?  We are proposing to have an agent on each k8s node to monitor volume health and store the volume health information in the PVC status.  Any comments?

xing-yang:
CC @gnufied 
Hi Hemant, can you share your comments on the API changes and whether the agent should update PVC status field?

gnufied:
@xing-yang yeah agent updating PVC's status from the node should be fine.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+}
+```
+
+The following common error codes are proposed for volume health:
+* VolumeNotFound
+* VolumeUnmounted
+* OutOfCapacity
+* RWIOError
+
+There could be many other error conditions, i.e. DiskFailure, DiskRemoved, ConfigError, NetworkError, DiskDegrading, FilesystemCorruption, etc. Because different storage systems may report health status differently, we leave it to the CSI drivers to determine what other error codes to report.
```
gnufied: Are we planning to introduce these other error codes in a follow up PR?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+
+#### CSI interface
+Call GetVolume() RPC for network storage PVCs periodically to check the health condition of volumes themselves. To avoid stale information being stored on a PVC, each periodic update will mark the PVC with the latest health information, replacing previous health information added by the external controller. If the PVC becomes healthy after being marked as unhealthy previously, the controller should remove the previous information.
+
+As mentioned earlier, reaction is not in the scope of this proposal but will be considered in the future. Before reacting to any negative health condition, the controller responsible for the reaction should call GetVolume() again to ensure the information is update to date.
+
+#### Node failure event
+Watch node failure event for both network and local storage. 
+In the case of a node failure, the controller will mark all local PVCs on that node.
+For network storage in the case of a node failure, the controller will log an event.
```
gnufied: Can you clarify which condition will be added to PVC when a node fails?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+- The health condition checking of volumes themselves. For example, whether the volume is deleted, whether the usage is reaching the threshold, and so on.
+- Mounting conditions checking.
+- Other errors that could affect the usability of the volume.
+
+In addition, a service will be created to receive volume health condition reports from other compoments implemented and deployed by users.
+
+Three main parts are involved here in the architecture.
+
+- API changes: The Status field in the PVC will be used to mark volumes if they are unhealthy.
+- External Controller:
```
liggitt: will this controller run centrally or on each node?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+  - Trigger controller RPC to check the health condition of network volumes themselves for network storage.
+  - Watch for node failure events for both network and local storage.
+  - Create HTTP(RPC) service to receive PVs health condition reports.
```
liggitt: how are those reports authenticated/authorized?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+* Volume could be out of capacity.
+* The disk may be degrading which affects its performance.
+
+If the volume is mounted on a pod and used by an application, the following problems could also happen.
+* There may be read/write I/O errors.
+* The file system on the volume may be corrupted.
+* Filesystem may be out of capacity.
+* Volume may be unmounted by accident outside of Kubernetes.
+
+If the CSI driver has implemented the CSI volume health function proposed in this design document, Kubernetes could communicate with the CSI driver to retrieve any errors detected by the underlying storage system. Kubernetes can store this data on the associated PVC so that user can inquire this information and decide how to handle them. For example, if the volume is out of capacity, user can request a volume expansion to get more space. Kubernetes may also use volume health information stored on the PV to automatically reconcile.
```
liggitt: which kubernetes component would modify the PVC?

xing-yang: There will be two new components that may modify the PVS status with volume health info:
1) An external monitoring agent on each kubernetes worker node
2) An external monitoring controller on the master node

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: The external controller will be running centrally.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: That depends on whether sig-storage wants to add them in the future.  So we can start with a minimal set that we can agree upon, and add more when needed.

keps/sig-storage/20190530-pv-health-monitor.md
liggitt: how is this service accessed? is it a service routed to via kube-proxy, or does this assume network visibility from agents on nodes to the controller service?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Good question. This is not covered by the proposed 4 error codes.  We should add another error code for this one, i.e., NodeFailure.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: We really have not thought through about this HTTP(RPC) service.  This was brought up in a review comment, but it won't go through the CSI driver so it is an alternative approach.  I'm going to drop this from the current KEP.  In the future we can add it through a new KEP if needed.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+// PersistentVolumeClaimHealthConditionType defines the health condition of PV claim.
+// Valid values are "VolumeNotFound", "VolumeUnMounted", "VolumeOutOfCapacity", or "VolumeIOError".
+type PersistentVolumeClaimHealthConditionType string
+
+// These are valid health conditions of PVC
+const (
+        // PersistentVolumeClaimVolumeNotFound - Volume not found
+        PersistentVolumeClaimVolumeNotFound PersistentVolumeClaimHealthConditionType = "VolumeNotFound"
+        // PersistentVolumeClaimVolumeUnMounted - Volume not mounted
+        PersistentVolumeClaimVolumeUnMounted PersistentVolumeClaimHealthConditionType = "VolumeUnMounted"
```
msau42: Does Kubernetes actually need explicit error conditions, or can it just pass through the CSI error codes?

Adding new error codes will be hard if it requires a K8s API change every time. And there's no way to surface any of the "custom" error codes that drivers may have.

keps/sig-storage/20190530-pv-health-monitor.md
msau42: Is this running as a sidecar to the CSI driver?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+Three main parts are involved here in the architecture.
+
+- API changes: The Status field in the PVC will be used to mark volumes if they are unhealthy.
+- External Controller:
+  - Trigger controller RPC to check the health condition of network volumes themselves for network storage.
+  - Watch for node failure events for both network and local storage.
+  - Create HTTP(RPC) service to receive PVs health condition reports.
+
+- External Agent:
+  - Trigger node RPC to check PVs’ mounting conditions for network storage.
```
msau42: It's important to note here that this has to run on the node. Is it deployed as a sidecar to the CSI node driver?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
...
+
+There could be conditions that cannot be reported by a CSI driver. There could be network failure where Kubernetes may not be able to get response from the CSI driver. In this case, a call to the CSI driver may time out. There could be network congestion which causes slow response. One or more nodes where the volume is attached to may be down. This can be monitored and detected by the volume health controller so that user knows what has happened.
+
+## Proposal
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+- If a volume provisioned by the CSI driver is deleted, we need to mark the corresponding PVC to inform users.
```
msau42: Do we need to make every error condition a first class error? Can we just surface any error the driver reports?

keps/sig-storage/20190530-pv-health-monitor.md
msau42: These sound like useful and common error conditions to surface too, even if we don't use it in Kubernetes. Can we include all these other error codes as well to help avoid fragmentation?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,539 @@
+---
+title: PV Health Monitor
+authors:
+  - "@NickrenREN"
+  - "@xing-yang"
+owning-sig: sig-storage
```
msau42: We should also add @kubernetes/sig-instrumentation-proposals to review and also provide recommendations.

keps/sig-storage/20190530-pv-health-monitor.md
msau42: If it runs on the master node, how does it make the csi controller calls?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: sure

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes.  Will update.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: It will be deployed as a sidecar like external-provisioner.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, will add that description.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Sure

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: We need to at least show that the volume in unhealthy in the status field.  Maybe we can just have simplified health conditions like these:
- HealthFailure
- HeathWarning
- HealthUnknown
All the detailed error codes can be reported by the driver and will be included in the message field?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Sure. How about having HealthFailure, HeathWarning, HealthUnknown condition types and including actual error codes from the driver in the message field?

xing-yang: If we use Condition, we can add HealthFailure, HeathWarning, HealthUnknown condition types. ConditionStatus would be true/false/unknown. Maybe we can store error code in Reason and details messages in Message field.

```
type PersistentVolumeClaimHealthCondition struct {
        Type   PersistentVolumeClaimHealthConditionType
        Status ConditionStatus
        // +optional
        LastProbeTime metav1.Time
        // +optional
        LastTransitionTime metav1.Time
        // +optional
        Reason string
        // +optional
        Message string
}
```

msau42: I think that approach would be more extensible. Do we want to include error code in the struct to so that we could potentially add reconciliation based on certain errors in the future?

xing-yang: Sure, we can add an ErrorCode field in the struct to be clear.

xing-yang:
@msau42 @liggitt @gnufied addressed your comments.  PTAL.  Thanks.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,577 @@
...
+- Other errors that could affect the usability of the volume.
+
+Three main parts are involved here in the architecture.
+
+- API changes: The Status field in the PVC will be used to mark volumes if they are unhealthy.
+
+- External Controller:
+  - The external controller will be deployed as a sidecar together with the CSI controller driver, similar to how the external-provisioner sidecar is deployed.
+  - Trigger controller RPC to check the health condition of network volumes themselves for network storage.
+  - Watch for node failure events for both network and local storage.
```
msau42: Is this part going to be part of the sidecar too? Or is that an independent controller?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,577 @@
...
+- Mounting conditions checking.
+- Other errors that could affect the usability of the volume.
+
+Three main parts are involved here in the architecture.
+
+- API changes: The Status field in the PVC will be used to mark volumes if they are unhealthy.
+
+- External Controller:
+  - The external controller will be deployed as a sidecar together with the CSI controller driver, similar to how the external-provisioner sidecar is deployed.
+  - Trigger controller RPC to check the health condition of network volumes themselves for network storage.
```
msau42: Why only network volumes? It should just be any CSI volume for that driver?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,577 @@
...
+  - Watch for node failure events for both network and local storage.
+
+- External Agent:
+  - The external agent will be deployed as a sidecar together with the CSI node driver on every Kubernetes worker node.
+  - Trigger node RPC to check PVs’ mounting conditions for network storage.
+  - Trigger node RPC to check local volumes health condition for local storage;
```
msau42: Do we need to call out local storage vs network storage explicitly? It's the same for both as long as it's a CSI driver right?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,577 @@
...
+Three main parts are involved here in the architecture.
+
+- API changes: The Status field in the PVC will be used to mark volumes if they are unhealthy.
+
+- External Controller:
+  - The external controller will be deployed as a sidecar together with the CSI controller driver, similar to how the external-provisioner sidecar is deployed.
+  - Trigger controller RPC to check the health condition of network volumes themselves for network storage.
+  - Watch for node failure events for both network and local storage.
+
+- External Agent:
```
msau42: So I think the main thing we need to resolve here is:

Should this be a node-level sidecar, or kubelet itself? There's been some effort to push monitoring out of kubelet, but for our purposes, we need to work with [node authorizer](https://github.com/kubernetes/kubernetes/blob/bbc035a66cacc345dd54676e6c4e1f0389dbcb5a/plugin/pkg/auth/authorizer/node/node_authorizer.go#L102) since we want to update the status.

xing-yang: Discussed this with @gnufied .  So the current proposal is to have a node-level sidecar update the PVC status. If we do it in Kubelet, as you said we may not be allowed to add more monitoring logic there.  So there is a tradeoff.

@liggitt Hi Jordan, can you help answer the question?  Thanks!

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, it can be combined.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, that makes sense.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: So right now the idea is to just do this check inside the external controller sidecar.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Feedback from Jordan: He is concerned about giving a node agent permissions to modify PVC Status.  He asked “what the impact is if a malicious node put an unhealthy status on a PVC”.  His suggestion is to “start with something like reporting events”.

xing-yang: Discussed with Michelle.  Removed API changes from the main proposal.  Both controller and node agent will only report events now so that they are consistent.

xing-yang:
@msau42 @liggitt Your comments are addressed.  PTAL.  Thanks!

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+* An external monitoring agent on each kubernetes worker node.
+* An external monitoring controller on the master node.
+
+Details will be described in the following proposal section.
+
+## Proposal
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+- If a volume provisioned by the CSI driver is deleted, we need to report an event and log a message to inform users.
```
msau42: Do we need to explicitly call out these errors conditions since they're already mentioned in the use case? Will the monitoring controller/agent actually be looking for these conditions, or is it the responsibility of the CSI driver to detect and report them?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+
+- External Controller:
+  - The external controller will be deployed as a sidecar together with the CSI controller driver, similar to how the external-provisioner sidecar is deployed.
+  - Trigger controller RPC to check the health condition of the CSI volumes.
+  - The external controller sidecar will also watch for node failure events.
+
+- External Agent:
+  - The external agent will be deployed as a sidecar together with the CSI node driver on every Kubernetes worker node.
+  - Trigger node RPC to check volume's mounting conditions.
+  - Note that currently we do not have CSI support for local storage. As a workaround, we may check the local volumes directly by the agent at first, and then move the checks to RPC interfaces when it is ready.
```
msau42: Is this actually going to be part of the plan? Where would this go?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+
+
+## Implementation
+
+### CSI changes
+
+Container Storage Interface (CSI) specification will be modified to provide volume health check leveraging existing RPCs and adding new ones.
+
+- GetVolume RPC
```
msau42: Is this RPC part of the Controller or Node service?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+  - External controller calls a new `GetVolume` RPC to check health condition of a particular PV instead of calling `ListVolumes` with a `volume_id` filter. Some CSI drivers chose not to implement `ListVolumes` as it is an expensive operation. Therefore it is better to introduce a new RPC `GetVolume` which has a new capability.
+- NodeGetVolumeStats RPC
```
msau42: clarify this is an existing RPC that will be extended

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+}
+```
+
+```
+message GetVolumeResponse {
+  message VolumeStatus{
+    // health shows error conditions reported by the SP.
+    // This field MUST be specified if the
+    // GET_VOLUME_HEALTH controller capability is supported.
+    VolumeHealth health = 1;
```
msau42: should this be repeated, so a driver can report multiple errors?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+    // RPC that the controller supports.
+    RPC rpc = 1;
+  }
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves.
```
msau42: How frequent? Will it be tunable?

keps/sig-storage/20190530-pv-health-monitor.md
msau42: Is network storage being handled differently than local volumes for this case?

How do you detect that a PVC for network storage is on that node?

keps/sig-storage/20190530-pv-health-monitor.md
msau42: Thanks!

Regarding naming, it may be more clear to call it "external node agent" to easily distinguish it from the external controller

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### Node failure event
+Watch node failure events.
+In the case of a node failure, the controller will report an event for all PVCs on that node.
+For network storage in the case of a node failure, the controller will log an event.
+
+### External agent
+
+#### CSI interface
+Call NodeGetVolumeStats() RPC to check the mounting and other health conditions.
```
msau42: How often? How does it know which volumes to call it for?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+
+Call both GetVolume() and NodeGetVolumeStats() RPCs for local storage when local storage CSI support is enabled.
```
msau42: Or should local storage CSI driver also deploy the controller piece as a daemonset? Then we don't need to duplicate the responsibilities.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+For now, check local volumes directly by the agent.
+
+### Alternatives
+
+#### Alternative option 1
```
msau42: For each alternative, can you add a short summary explaining why we are not pursuing this option?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+  // This field is REQUIRED.
+  string error_code = 1;
+
+  // The error message associated with the above error_code. This field is OPTIONAL.
+  string message = 2;
+}
+```
+
+The following common error codes are proposed for volume health:
+* VolumeNotFound
```
msau42: Do you see a reason to separate out specific error codes for Controller vs Node?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+```
+message VolumeUsage {
+  enum Unit {
+    UNKNOWN = 0;
+    BYTES = 1;
+    INODES = 2;
+  }
+  // The available capacity in specified Unit. This field is OPTIONAL.
+  // The value of this field MUST NOT be negative.
+  int64 available = 1;
```
serathius: Is available redundant? How it differs from total - used?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Controller RPC.  @saad-ali suggested to have a GetVolume() instead of add a volume_id to the existing ListVolumes().

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: This field is already in the CSI spec, not added by this KEP.  
available = total - used

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+Instead of adding a new RPC, we can leverage the existing NodeGetVolumeStats RPC.
+
+```
+rpc NodeGetVolumeStats (NodeGetVolumeStatsRequest)
+    returns (NodeGetVolumeStatsResponse) {}
+```
+
+In the NodeGetVolumeStatsRequest, there are `volume_id`, `volume_path`, and `staging_target_path`. CSI driver can figure out whether a volume is still mounted based on these parameters.
+
+In NodeGetVolumeStatsResponse, there is already a `VolumeUsage` which includes information (available, used, and total) of the file system on the volume.
```
serathius: Have you though how `VolumeUsage` will be exposed/consumed within cluster? I think this is interesting topic that also applies to container runtime and network. It's ok if not.

There was one idea on my mind to introduce USE (utilization, saturation, error) metrics from interfaces like (CRI, CNI, CSI) and expose them in cluster using [kubernetes metrics framework](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/component-base/metrics) (wrapper for prometheus).
I think this case shows that there is a need to expand scope of monitoring in k8s (currently limited to autoscaling)
/cc @brancz @davidashpole
I've been drafting a proposal on this and I would be interested in discussing it more in SIG-instrumentation.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+message NodeGetVolumeStatsResponse {
+  // This field is OPTIONAL.
+  repeated VolumeUsage usage = 1;
+  // This field is OPTIONAL.
+  repeated VolumeHealth volume_health = 1;
+}
+```
+
+```
+message VolumeUsage {
```
serathius: How fresh Volume usage needs to be? can it be cashed? if so could we include timestamp when measurement was done?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+```
+
+#### Add Node Volume Health Function
+
+Node Volume Health checks if a volume is still mounted and usable. To check whether a volume is usable, the CSI driver shall check if filesystem is corrupted, whether there are bad blocks, etc. in this RPC.
+
+Instead of adding a new RPC, we can leverage the existing NodeGetVolumeStats RPC.
+
+```
+rpc NodeGetVolumeStats (NodeGetVolumeStatsRequest)
```
serathius: Pull based method for fetching utilization makes sense, but for status changes (health) maybe we should consider push?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,576 @@
...
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves.
+
+#### Node failure event
+Watch node failure events.
```
serathius: What is node failure event here? Can you give example?
Assuming that this is some k8s event on node object, does it mean that it will be coping events from Node to PVC?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yeah, it is a little redundant here as it is already covered by the use cases.  I'll remove them.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Nick said he will do this, but there's no detailed design on how to do it.  I'm going to remove this for now.  It can always be added later when we have a concrete plan.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Sure

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, it should be tunable.  We can have a configure option in the monitoring controller.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: The responsibilities should not be duplicate.  The monitoring controller will get information directly from the storage system so we don't need this type of information to be retrieved on every node. However, mount related information can only be retrieved on every node.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Ok

keps/sig-storage/20190530-pv-health-monitor.md
msau42: By duplicate responsibilities I mean that 1) external controller is already calling GetVolume() and 2) node agent also calls GetVolume only in the local volume case.  Can we just have external controller call GetVolume() in all cases?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Push based method is brought up in the section "HTTP(RPC) Service".  We didn't adopt that approach as the main proposal because CSI uses pull based method.  I'll add some clarification there.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: This refers to a node down event.  So the monitoring controller will be checking if the node is still up and only report event when the node is down.  I'll reword this section to clarify.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Got it.  Will change.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, that makes sense.

xing-yang:
> Is network storage being handled differently than local volumes for this case?
> How do you detect that a PVC for network storage is on that node?

Yes, it is handled differently.  For local volumes, we can figure out what local volumes are on a particular node.  So an event can be reported for every PVC affected.  For network storage, we will only report a general node down event, not specific to any volume.  Added clarifications.

msau42:
> > Is network storage being handled differently than local volumes for this case?
> > How do you detect that a PVC for network storage is on that node?
> 
> Yes, it is handled differently. For local volumes, we can figure out what local volumes are on a particular node. So an event can be reported for every PVC affected. For network storage, we will only report a general node down event, not specific to any volume. Added clarifications.

What object is the node down event reported on? Also how will you tell the difference between a local volume vs network volume?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Added at the end of each alternative.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Will add a config option to adjust the frequency.

The agent can go through all pods on that node and find out all volumes used by those pods.  I couldn't find a more straight-forward way.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: I think some can be separated but others may be in both category.  Can we sort this out later during the CSI spec review?

xing-yang: I removed "NodeFailure" as that will be detected by the external monitoring controller without issuing a CSI call.

xing-yang:
> What object is the node down event reported on? Also how will you tell the difference between a local volume vs network volume?

For local storage, @NickrenREN said there is a way to tell what volumes are on that node.  So this will be reported on the PVC objects.  
For all other storage, we don't really have an object to report the event. So maybe we can only log a message.  I'll drop the event part for this one.
In the external controller sidecar, we should know the driver name.  Based on that we can tell whether it is local storage or not.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+The main architecture is shown below:
+
+![pv health monitor architecture](./pv-health-monitor.png)
+
+The following areas will be the focus of this proposal at first:
+
+- The health condition checking of volumes themselves. For example, whether the volume is deleted, whether the usage is reaching the threshold, and so on.
```
msau42: The proposal is not actually doing the health or mount checks right? The proposal is about providing a mechanism for plugins to report volume health/errors.

keps/sig-storage/20190530-pv-health-monitor.md
msau42: Yeah we can discuss the error codes in more detail during CSI spec review

keps/sig-storage/20190530-pv-health-monitor.md
msau42: For a push model, can the node agent define some RPC service that CSI drivers can call out to?

@serathius can you point to any existing design that uses a push based model?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves. The frequency of the check should be tunalbe. A configure option will be available in the external controller to adjust this value.
+
+#### Node down event
+Watch node down events.
+In the case that a node goes down, the controller will report an event for all local PVCs on that node.
+For network storage in the case of a node failure, the controller will just log a general message, not specific to individual PVCs because the controller has no knowledge of what PVCs are on the affected node.
```
msau42: When there's a local volume CSI driver, I'm not sure how the controller will be able to tell the difference between network and local.

One possibility is to track which pods are using which PVCs and what nodes they got scheduled to?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+    // RPC that the controller supports.
+    RPC rpc = 1;
+  }
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves. The frequency of the check should be tunalbe. A configure option will be available in the external controller to adjust this value.
```
msau42: typo: tunable

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+* Volume may be unmounted by accident outside of Kubernetes.
+
+If the CSI driver has implemented the CSI volume health function proposed in this design document, Kubernetes could communicate with the CSI driver to retrieve any errors detected by the underlying storage system. Kubernetes can report an event and log an error about this PVC so that user can inquire this information and decide how to handle them. For example, if the volume is out of capacity, user can request a volume expansion to get more space. In the first phase, volume health monitoring is informational only as it is only reported in the events and logs. In the future, we will also look into how Kubernetes may use volume health information to automatically reconcile.
+
+There could be conditions that cannot be reported by a CSI driver. There could be network failure where Kubernetes may not be able to get response from the CSI driver. In this case, a call to the CSI driver may time out. There could be network congestion which causes slow response. One or more nodes where the volume is attached to may be down. This can be monitored and detected by the volume health controller so that user knows what has happened.
```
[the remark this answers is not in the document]
xing-yang: I said "conditions that cannot be reported by a CSI driver" on line 83 so I meant CSI driver cannot report a network failure.  Maybe I should just remove it to avoid confusion.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+
+Details will be described in the following proposal section.
+
+## Proposal
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+The main architecture is shown below:
+
+![pv health monitor architecture](./pv-health-monitor.png)
```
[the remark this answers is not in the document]
xing-yang: Sure

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, will modify this.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Sure.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Fixed.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: In the current CSI spec, I don't see a push model.  Therefore the design is based on a pull model.  We can ask about it at the CSI community meeting during the CSI spec review.  If a push model is possible, we can come back to update the KEP later.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,571 @@
...
+Details will be described in the following proposal section.
+
+## Proposal
+
+Volume monitoring is the main focus of this proposal. Reactions are not in the scope of this proposal.
+
+The following areas will be the focus of this proposal at first:
+
+- Provide a mechanism for CSI drivers to report volume health problems. For example, whether the volume is deleted, whether the usage is reaching the threshold, and so on.
+- Mounting conditions checking.
```
msau42: Does the node agent actual check for mounts? Or is it the responsibility of the driver?

Maybe we should just say to provide a way for CSI drivers to report volume health problems at the controller and node levels.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,571 @@
...
+- Other errors that could affect the usability of the volume.
```
msau42: Are these "other errors" things that a CSI driver would not be able to detect?

keps/sig-storage/20190530-pv-health-monitor.md
msau42: still see typo

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,571 @@
...
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves. The frequency of the check should be tunalbe. A configure option will be available in the external controller to adjust this value.
+
+#### Node down event
+* Watch node down events.
```
msau42: Would this be watching Node status?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,571 @@
...
+  * External controller volume health monitoring.
+  * External node agent volume health monitoring.
+* Tests outlined in design proposal implemented.
+
+### Beta -> GA
+* Volume health support is added to multiple CSI drivers.
+* Volume health feature deployed in production and have gone through at least one K8s upgrade.
+
+## Test Plan
+### Unit tests
```
msau42: It would be nice to be more specific about the test cases, what error scenarios will we be testing? What drivers are we going to use to test?

Also, any plan for stress/scale tests?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yeah, driver will check it.  I'll reword this.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: This refers to errors the CSI drivers can detect.  I'll reword.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Yes, it will watch node status.  In addition I think we also need to ping the node as it could be an unplanned shut down.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Updated.

xing-yang:
@msau42 comments addressed.  PTAL.  Thanks.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+}
+```
+
+### External controller
+
+#### CSI interface
+Call GetVolume() RPC for volumes periodically to check the health condition of volumes themselves. The frequency of the check should be tunable. A configure option will be available in the external controller to adjust this value.
+
+#### Node down event
+* Watch node down events by checking node status and also pinging the node.
```
msau42: There's already NodeLease objects that kubelet periodically updates (like a heartbeat). We could possibly use that, but even then, I'm still not sure why we cannot just rely on the NodeController to mark the Node as unhealthy when the NodeLease is stale. It seems like we're crossing over responsibilities here.

xing-yang: Fixed

xing-yang:
Squashed commits into 1.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -0,0 +1,575 @@
...
+* VolumeNotFound
+* OutOfCapacity
+* DiskFailure
+* DiskRemoved
+* ConfigError
+* NetworkError
+* DiskDegrading
+* VolumeUnmounted
+* RWIOError
+* FilesystemCorruption
```
saad-ali: Enumerating the failure types as error codes in CSI makes me worry. Are we going to be able to capture all possible use cases for all possible storage systems? How will we handle backwards compat?

The main reason to have error codes (as opposed to opaque strings) is to enable programmatic response. Like the existing CSI error codes we should detail the expected recovery behavior for each error code that may help us reduce to a minimum set.

xing-yang: Sure.  Let's discuss more in the CSI API review.

blackgold:
will the volume health event be attached to the pod using the volume?

NickrenREN:
@blackgold yes, we are planning to do this, please refer to this issue for progress: https://github.com/kubernetes-csi/external-health-monitor/issues/10

PR #1081, opened 2019-06-03
Server-side apply: Create flag to get managedFields

apelisse:
cc @kwiesmueller @jennybuckley @liggitt

### DIFF
--- a/keps/sig-api-machinery/0006-apply.md
+++ b/keps/sig-api-machinery/0006-apply.md
@@ -157,6 +157,10 @@ object.
 
 A flag will be implemented in kubectl to trigger that behavior.
 
+For validating and mutating webhooks, the field is also going to be stripped,
+unless explicitly requested by the webhook through a new `IncludeManagedFields`
+boolean field in the ValidatingWebhook and MutatingWebhook configuration.
+
 One of the downside of having this field hidden by default is that people can't
 access it unless they know the triggering flag, which might lead to confusing
 situations.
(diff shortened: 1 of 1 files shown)


kwiesmueller:
I like it!
Does this address the scalability issues we saw?

liggitt:
partially... for every round trip, there are two serializations:

Read path:
* read from etcd, decode into in-memory in apiserver
* encode for API client

Write path:
* read from incoming API request, decode into in-memory in apiserver
* read from etcd, decode into in-memory in apiserver
* compare resourceVersion, etc
* encode new object for etcd

This change would help the read path, but not the write path

kwiesmueller:
Okay, but the write path is no longer so much different from before we had managedFields, is it?

apelisse:
Yeah, this is mostly helping reads, not writes. The apiserver is the only component that pays the write cost of SSA, all components are paying the read (if we don't add this flag). Especially on big clusters, we do tend to have insanely big reads (thousands of pods/nodes etc ...).

apelisse:
New thought that I have: We need to decide whether admission chain webhook would get the fields or not (or provide an option for them to decide). I suppose we could add a new field in the WebhookConfiguration that would default to no.

apelisse:
I've added a paragraph to talk about webhooks

apelisse:
We also need to consider serialization cache. I'll add a paragraph about that.

apelisse:
@wojtek-t Any opinions on how this affect serialization cache and how we can address it?

kwiesmueller:
@apelisse is this still a thing?
/remove-lifecycle rotten

apelisse:
No that's irrelevant now.

PR #1088, opened 2019-06-04
Add beta milestone criteria

RobertKrawitz:
Add beta milestone criteria

### DIFF
--- a/keps/sig-node/20180906-quotas-for-ephemeral-storage.md
+++ b/keps/sig-node/20180906-quotas-for-ephemeral-storage.md
@@ -13,7 +13,7 @@ approvers:
   - "@derekwaynecarr"
 editor: TBD
 creation-date: 2018-09-06
-last-updated: 2019-04-30
+last-updated: 2019-06-04
 status: implementable
 ---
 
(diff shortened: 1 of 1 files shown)


keps/sig-node/20180906-quotas-for-ephemeral-storage.md
```diff
@@ -622,15 +625,21 @@ appropriate end to end tests.
 The following criteria applies to
 `LocalStorageCapacityIsolationFSMonitoring`:
 
-Alpha
+## Phase 1: Alpha (1.15)
 
 - Support integrated in kubelet
+- Alpha-level documentation
+- Feature gate disabled by default
```
dashpole: The feature gate is probably not worth mentioning here or below.  It is a result of, not a requirement for graduation.

keps/sig-node/20180906-quotas-for-ephemeral-storage.md
```diff
@@ -622,15 +625,21 @@ appropriate end to end tests.
...
+- Unit test coverage
+- Node e2e test
 
-Beta
+## Phase 2: Beta (target 1.16)
```
dashpole: Since some of the objectives for the feature are a reduction in metric latency and overhead from metrics collection.  I think we need benchmarks or other validation that the feature accomplishes those goals to move to beta.

RobertKrawitz: I'm trying to work up a benchmark/stress test for this.  The basic method I'm trying to use is many pods using emptydirs with many directories and files within them.

dashpole: Sounds good.  The statement here doesn't have to be specific at all.  Something to the effect of "benchmark to determine latency, resource cost, and syscall overhead (?) of using quotas relative to existing monitoring solution"

RobertKrawitz: OK, done.

RobertKrawitz: Added microbenchmark results.

RobertKrawitz:
ping @derekwaynecarr @dchen1107

RobertKrawitz:
@dashpole just rebased needs new lgtm -- thanks!

keps/sig-node/20180906-quotas-for-ephemeral-storage.md
```diff
@@ -613,19 +627,111 @@ appropriate end to end tests.
...
-Alpha
+### Phase 1: Alpha (1.15)
 
 - Support integrated in kubelet
+- Alpha-level documentation
+- Unit test coverage
+- Node e2e test
 
-Beta
+### Phase 2: Beta (target 1.16)
```
dchen1107: nit: We need the user feedbacks from alpha feature, I doubt we can get that to promote it to the beta in 1.16. But I am ok to target here.

PR #1089, opened 2019-06-05
[WIP] Metrics Validation and Verification

solodov:


keps/sig-instrumentation/20190605-metrics-validation-and-verification.md
```diff
@@ -0,0 +1,94 @@
...
+## Motivation
+
+### Goals
+
+* Given a stable metric, validate that we cannot remove or modify it (other
+  than adding deprecation information).
+* Given a deprecated but stable metric, validate that we cannot remove or
+  modify it until the deprecation period has elapsed.
+* Given an alpha metric which is promoted to be 'stable', automatically include
+  proper instrumentation reviewers (for schema validation and conformance to
```
brancz: How can a reviewer know whether labels are in an appropriate state? Even the metrics we're extremely careful about today already have changed very slightly over time already. All of that would not be permitted once stable, and even harder for a reviewer how hasn't been involved the entire time. I have a feeling that we need to play through an example of this.

logicalhan: You only have to validate it once (i.e. when the metric is getting promoted to GA).

brancz: Right. I'm saying, if we had had the framework in place, I'm certain we would have had a need to break the apiserver request counter and duration metrics multiple times already, but couldn't have under these circumstances. What would happen? We would have to stay in a state that would be "broken" for the sake of compatibility. I feel like there needs to be a step in-between, like beta, where we are saying, we would want this thing to become a stable metric and we think it's a good candidate, but it has to stay in beta for x releases to validate this.

logicalhan: I think it is a reasonable concern, though probably not super related to this particular KEP. In general, I think we should probably have a metrics API review process, akin to the API review process in api-machinery. We should discuss this in the sig meeting today.

mrbobbytables:
@solodov @brancz Has this WIP KEP been superseded by https://github.com/kubernetes/enhancements/pull/1169 ?

smarterclayton:
Looks like it

/close

PR #1092, opened 2019-06-06
Add KEP for presubmit config in tested repo

alvaroaleman:
/assign @cjwagner @fejta @stevekuznetsov 

/cc @matthyx @kdomanski

### DIFF
--- a/keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
+++ b/keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
@@ -1,0 +1,277 @@
+---
+title: Presubmit config inside the tested repo
+authors:
+  - "@alvaroaleman"
+owning-sig: sig-testing
+participating-sigs:
+  - sig-testing
+reviewers:
+  - TBD
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2019-06-04
+last-updated: 2019-06-05
+status: provisional
+---
+
+# Presubmit config inside the tested repo
+
+
+## Table of Contents
+
+
+* [Presubmit config inside the tested repo](#presubmit-config-inside-the-tested-repo)
+* [Table of Contents](#table-of-contents)
+* [Release Signoff Checklist](#release-signoff-checklist)
+* [Summary](#summary)
+* [Motivation](#motivation)
+   * [Goals](#goals)
+   * [Non-Goals](#non-goals)
+* [Proposal](#proposal)
+   * [prow/config adjustments](#prowconfig-adjustments)
+   * [New prow/inrepoconfig/api package](#new-prowinrepoconfigapi-package)
+   * [New prow/inrepoconfig package](#new-prowinrepoconfig-package)
+   * [Extend prow/config with configuration to opt-in to this feature](#extend-prowconfig-with-configuration-to-opt-in-to-this-feature)
+   * [Add code to tide and the trigger plugin so they will use presubmits from a prow.yaml if the feature is activated](#add-code-to-tide-and-the-trigger-plugin-so-they-will-use-presubmits-from-a-prowyaml-if-the-feature-is-activated)
+* [Risks and Mitigations](#risks-and-mitigations)
+   * [Plugins may have an inconsitent view on what presubmits exist for a repo](#plugins-may-have-an-inconsitent-view-on-what-presubmits-exist-for-a-repo)
+   * [Someone could accidentally delete all presubmits in prow.yaml and have their PR instantly merged](#someone-could-accidentally-delete-all-presubmits-in-prowyaml-and-have-their-pr-instantly-merged)
+* [Implementation History](#implementation-history)
+
+## Release Signoff Checklist
+
+**ACTION REQUIRED:** In order to merge code into a release, there must be an issue in [kubernetes/enhancements] referencing this KEP and targeting a release milestone **before [Enhancement Freeze](https://github.com/kubernetes/sig-release/tree/master/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core Kubernetes i.e., [kubernetes/kubernetes], we require the following Release Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These checklist items _must_ be updated for the enhancement to be released.
+
+- [ ] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [ ] KEP approvers have set the KEP status to `implementable`
+- [ ] Design details are appropriately documented
+- [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] Graduation criteria is in place
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+**Note:** Any PRs to move a KEP to `implementable` or significant changes once it is marked `implementable` should be approved by each of the KEP approvers. If any of those approvers is no longer appropriate than changes to that list should be approved by the remaining approvers and/or the owning SIG (or SIG-arch for cross cutting KEPs).
+
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+This document proposes to change the Prow presubmit handling to optionally version some or
+all presubmits in the same repository that also contains the code that is being tested.
+
+## Motivation
+
+Currently, all jobs for Prow are configured in a central `infra-config` repository that is
+in most cases distinct from the repositories whose code is being tested in presubmits. This
+poses severall challenges:
+
+* It is not possible to see in a Pull Request that introduces a new presubmit if that presubmit will
+  actually pass. The two workarounds for this are to either have both the job author and the reviewer
+  use two CLIs to create a Pod from the job or to make the Job initially optional, verify it with a
+  test Pull Request to the code repository, then make it mandatory. Both of these workarounds are
+  cumbersome.
+* The same issues apply when doing changes to the config of a presubmit
+* When a project maintains multiple branches, e.G. because there are release branches, the
+  maintainers must remember to create a copy of the presubmit in the `infra-config` repository.
+  Otherwise it is easely possible that the presubmit becomes incompatible with one of the branches
+  it is used in, because somone makes a change to its config and forgets to test against all branches.
+  This is an additional step maintainers have to remember when branching off.
+
+
+### Goals
+
+* It is possible to version some or all presubmits of a given repository inside that repository in a
+  `yaml` file
+* This feature is optional and opt-in
+* The triggering of presubmits on pull request creation or updates continues to work and includes the
+  jobs that are managed inside the code repository
+* Re-Running tests via the `/retest` command continues to work and includes the jobs that are
+  managed inside the code repository
+* Explicitly executing optional tests via the `/test <<myjob>>` command continues to work and includes
+  all jobs that are managed inside the code repository
+* All the existing defaulting and validation for Presubmit jobs is being used to default and validate
+  jobs that are managed inside the code repository
+* If there is an error parsing, defaulting or validating the presubmits that are managed inside the
+  code repository, a comment will be posted to GitHub stating the error
+* Pull Requests on which an error occurred during parsing, defaulting or validation the presubmits that
+  are managed in the code repository are not consideres as merge candidates by Tide
+* Tide executes the presubmits that are defined inside the code repository when it re-tests
+* Renamed blocking presubmits added via pull request trigger a migration on all in-flight PRs
+* Removed blocking presubmits via pull request trigger a status retire on all in-flight PRs
+
+### Non-Goals
+
+* The option to configure Postsubmits or Periodics inside the tested repository. This may be
+  done in a future iteration.
+
+## Proposal
+
+It is proposed to introduce a the option to configure additional presubmits inside the code
+repositories that are tested by prow via a file named `prow.yaml`
+
+The minimum part of Prow this has to be integrated with is:
+
+* The `trigger` plugin as it manages the triggering of Jobs upon Pull Request Events or commands
+* `Tide`, as it re-excutes jobs when the base branch for a pull request changed after it was
+  tested
+
+The following changes will be required:
+
+#### `prow/config` adjustments
+
+The `prow/config` package needs to unite its various defaulting and
+validation funcs for presubmits so that in the end a func can be
+exported that will be used both in case of config reload and
+when additional presubmits are loaded from a `prow.yaml`.
+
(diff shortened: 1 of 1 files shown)


keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+
+*Open question:* What happens if the merged `prow.yaml` becomes invalid?
+Is tide able to recover from that?
+
+In the `trigger` plugin, we have to check if the `prow.yaml` feature is
+enabled and if yes, use the functionality described from the new `prow/inrepoconfig`
+package to fetch additional presubmits from it.
+
+
+### Risks and Mitigations
```
alvaroaleman: @cjwagner Re your [comment:](https://github.com/kubernetes/test-infra/pull/12836#issuecomment-498373618)

> I'm excited to see this functionality, but I'm worried that this is going to cause problems by breaking assumptions.
> * It only works for job config not core prow config or plugin config.

Yes, that is true. What assumption exactly is broken by that?

> * Jobs defined in prow.yaml will be handled differently than jobs in the --job-config directory (prow.yaml jobs are parsed with the inclusion of PR changes).

By that you mean plugins/components that have a `ref` will or may get a different view than those that don't, like e.G. `branchprotector`?

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+This document proposes to change the Prow presubmit handling to optionally version some or
+all presubmits in the same repository that also contains the code that is being tested.
+
+## Motivation
+
+Currently, all jobs for Prow are configured in a central `infra-config` repository that is
+in most cases distinct from the repositories whose code is being tested in presubmits. This
+poses severall challenges:
+
+* It is not possible to see in a Pull Request that introduces a new presubmit if that presubmit will
```
stevekuznetsov: For OpenShift, we are currently creating "rehearsal" jobs using the changed version of the job in question, so there are other useful workarounds that report job status to the PR changing the job.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+This document proposes to change the Prow presubmit handling to optionally version some or
+all presubmits in the same repository that also contains the code that is being tested.
+
+## Motivation
+
+Currently, all jobs for Prow are configured in a central `infra-config` repository that is
```
stevekuznetsov: Needs to clearly note that the reason for this is a decision on trust -- deployments that are fully OSS like for prow.k8s.io cannot trust the users that are checking in job config. 

Having jobs in a repo allows a PR to get itself to merge by removing all the blocking presubmits.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+in most cases distinct from the repositories whose code is being tested in presubmits. This
+poses severall challenges:
+
+* It is not possible to see in a Pull Request that introduces a new presubmit if that presubmit will
+  actually pass. The two workarounds for this are to either have both the job author and the reviewer
+  use two CLIs to create a Pod from the job or to make the Job initially optional, verify it with a
+  test Pull Request to the code repository, then make it mandatory. Both of these workarounds are
+  cumbersome.
+* The same issues apply when doing changes to the config of a presubmit
+* When a project maintains multiple branches, e.G. because there are release branches, the
```
stevekuznetsov: Please explain why this is a downside?

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+  maintainers must remember to create a copy of the presubmit in the `infra-config` repository.
+  Otherwise it is easely possible that the presubmit becomes incompatible with one of the branches
+  it is used in, because somone makes a change to its config and forgets to test against all branches.
+  This is an additional step maintainers have to remember when branching off.
+
+
+### Goals
```
stevekuznetsov: New blocking presubmits added via pull request trigger a run on all in-flight PRs 
Renamed blocking presubmits added via pull request trigger a migration on all in-flight PRs
Removed blocking presubmits via pull request trigger a status retire on all in-flight PRs

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+
+The `prow/config` package needs to unite its various defaulting and
+validation funcs for presubmits so that in the end a func can be
+exported that will be used both in case of config reload and
+when additional presubmits are loaded from a `prow.yaml`.
+
+*Open question:* What would be the best signature for such a function?
+It must be able to optionally accept additional presubmits for one repo
+
+#### New `prow/inrepoconfig/api` package
```
stevekuznetsov: This is pretty low-level for a KEP IMO

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+  `prow.yaml` parsing is done
+* Has functionality to remove status contexts that existed in an older
+  `baseRef` and got removed in a newer one. This could be done by:
+    * Recognizing status contexts created by Prow  using a combination
+      of a static list (e.G. `tide`) and for the remaining contexts by
+      comparing their `targetURL` with what is configured in Prow as
+      `jobURLPrefix`. If the `targetURL` starts with `jobURlPrefix`
+      and the context does not belong to one of the known presubmits,
+      the status context will be retired.
+    * Leveraging the `status-reconciler`. The `status-reconciler`
```
stevekuznetsov: When? On merge?

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+Currently, the presubmit config is exported as a `map[string][]Presubmit`. With the approach outlined above, plugins have to opt-in
+to see presubmits defined in `prow.yaml`.
+
+A possible mitigation for this could be to unexport the `map[string][]presubmit`
+and replace it with a `func (c* Config) Presubmits(org, repo, baseRef string, headRefs []string) ([]Presubmit, error)`.
+This would require adjusments in all plugins that use the presubmit config.
+
+*Open question with this approach:* What do we do about plugins that
+require the presubmit config but do not have a revision they work on,
+like e.G `branchprotector`?
```
stevekuznetsov: nit: `branchprotector` is not a plugin

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,259 @@
...
+
+#### Someone could accidentally delete all presubmits in `prow.yaml` and have their PR instantly merged
```
stevekuznetsov: Someone could intentionally change prow.yaml to run their bitcoin miner in the Prow cluster as a presubmit on their PR
Someone could intentionally change prow.yaml to run N jobs eating all possible disk space to DDOS the prow cluster

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: Could you point to some code I can look at? Sounds like a useful workaround

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: > Someone could intentionally change prow.yaml to run their bitcoin miner in the Prow cluster as a presubmit on their PR

That can already be done today by just changing the scripts that are executed by the job.

> Someone could intentionally change prow.yaml to run N jobs eating all possible disk space to DDOS the prow cluster

For a single node, that can already be done today.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: Can you elaborate what attack vector this opens that doesn't exist today?

If I can get my pull to be tested on k/k today, I can change the scripts that are executed by the job to get access to credentials.

The differences I can think off is that additional presets can be added and that its possible to set the cluster on the job.

I will add a section about this in the `Risks and mitigation` section.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: It is an additional step that has to be remembered, otherwise the job may be changed and become silently incompatible.

With a `prow.yaml` file this happens implicitly, as changes to the `prow.yaml` in master wont affect `prow.yaml` in other branches

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: I agree in that this is pretty low level for a kep but I was asked for a design so I made this a bit more detailled. It may not matter so much where new packages exactly are, but we should agree on the question if we actually create a new package or put the new functionality in existing ones

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: As mentioned further down the status reconciler would have to get notified on Push Events to any branch that is the base of any pull request, then reconcile all pull requests whose base changed.

IMHO the first approach is simpler thought.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
alvaroaleman: > Having jobs in a repo allows a PR to get itself to merge by removing all the blocking presubmits.

As noted in the `risks and mitigation` section, it is expected that PRs need a review so the author can not just get it merged without having someone look at it. Also, Prow doesn't even test PRs of untrusted ppl

alvaroaleman: IMHO this doesn't have too much to do with OSS or not. We will definitely activate this on our OSS repos, the fact that external contributions won't get tested without a member using `/test all` is enough safety here

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
stevekuznetsov: https://github.com/openshift/ci-operator-prowgen/blob/master/cmd/pj-rehearse/main.go

We've thought about how to upstream it but in our impl we mixed information about the ProwJob with our executor (ci-operator) so it would be messy, and the security questions still loomed.

keps/sig-testing/20190605-presubmitconfig-in-tested-repo.md
```diff
@@ -0,0 +1,277 @@
...
+
+* The option to configure Postsubmits or Periodics inside the tested repository. This may be
+  done in a future iteration.
+
+## Proposal
+
+It is proposed to introduce a the option to configure additional presubmits inside the code
+repositories that are tested by prow via a file named `prow.yaml`
+
+The minimum part of Prow this has to be integrated with is:
```
cjwagner: There is a lot more than `tide` and `trigger` that needs to be updated. Any plugin or component that consumes job config will need to be updated otherwise they may be looking at a truncated list of jobs. This is problematic if a specific job is being looked up and cannot be found, but it is also problematic when doing operations across all jobs e.g. enforcing branch protection, determining if a context can be skipped, or trying to migrating a job's status context. If we don't know that a job is configured we may remove its branch protection or improperly allow it to be skipped.
The following are the components/plugins I can think of off the top of my head that handle job config and will most likely need to be updated. There may be more.
- update-config
- config-bootstrapper
- branchprotector
- status-reconciler
- tide
- trigger
- skip
- override
- refresh
- horologium
- gerrit
- crier
- sub
- mkpj
- check-config

Another thing to consider is that this will make loading the job config require fetching `prow.yaml` from every repo that configures ProwJobs. A simple `git clone` would make that extremely expensive so we'll probably need to do a shallow clone or download the `prow.yaml` file individually.  Allowing config to live outside the cluster will also mean that config needs to be reloaded whenever we want to check for changes and that config loading will now be susceptible to network problems.

I'm worried that making the current job configuration dynamic and dependent on the PR(s) in question is going to add a lot of complexity and inefficiency to components that need to iterate over the currently configured jobs.

alvaroaleman: > There is a lot more than tide and trigger that needs to be updated. Any plugin or component that consumes job config will need to be updated otherwise they may be looking at a truncated list of jobs. This is problematic if a specific job is being looked up and cannot be found, but it is also problematic when doing operations across all jobs e.g. enforcing branch protection, determining if a context can be skipped, or trying to migrating a job's status context. If we don't know that a job is configured we may remove its branch protection or improperly allow it to be skipped.

Generally speaking I think the simplest approach for this problem would be to extend the `Config` type with to funcs for accessing `Presubmits` from it:

* `func (c *Config) GetPresubmits(gc *git.Client, org, repo, base string, heads ...string)([]Presubmit, error)` for components that have a revision they work on like `trigger`
* `func (c *Config) GetStaticPresubmits(org, repo string) []Presubmit` for components that do not have a revision they work on, e.G. the branchprotector

We could then move all components piece by piece over to use that and eventually unexport the `Presubmit` field from the `Config` type. WDYT?

> Another thing to consider is that this will make loading the job config require fetching prow.yaml from every repo that configures ProwJobs

There is the `prow/git` package which has caching, so this is quite efficient.

Re the components:

> * update-config

This doesn't access the `Presubmit` field and basically just copies the content of some files into a configmap. What do you think needs to be changed there?

> config-bootstrapper

Basically the same question as for update-confog

> branchprotector

Yeah, this will be a bit special. I don't think its feasible to apply branchprotection for jobs that are versioned in the `prow.yaml`, as a PR could remove them so for that PR, they must not be protected.

Instead,  I think it makes more sense to have one protected context `prow-config-parsing` to make sure that happens and do not have branchprotection for the jobs in `prow.yaml`

> status-reconciler

I don't think this needs any change. The only thing we need is handling for the situation that a base had a job that got removed. I've outlined an approach to solve that without the status-reconciler in this doc, because making the status-reconciler cope with a `prow.yaml` will be rather tricky as its edge-driven

> skip

This uses `trigger.FilterPresubmits` under the hood, so adjusting trigger will be sufficient

> override

This works on github status contexts only and doesn't access the Presubmit config

> refresh

This gets its data from Deck I guess? Ultimatively a `prowjobs.js` file

> horologium

This is just for periodics or not?

> gerrit

Thats a bit tricky as it doesn't use the `org/repo` format. What about simply saying this feature is not supported with Gerrit?

> crier

Crier works on the Prowjob objects only or not? From a quick glance  I can't see it accessing the ProwJobs from the config

> sub

Doesn't access any the Prowjobs from the config either

> mkpj

More or less the whole point of this is not to need mkpj, so I'd say we can skip that :)

> check-config

Sure we need to validate the config for this

I guess my point is: Lets just migrate everything that accesses the `ProwJobs` field from the config to instead use a getter and decide on a case-by-case basis what makes sense. We can do that before we actually add the feature, that way we will see if there are bigger problems with any components. WDYT?

alvaroaleman:
@stevekuznetsov @cjwagner I've updated the PR here to include what we discussed. I also removed all specifics around the actual implementation, as that is most likely easier discussed based on the sketch PR that outlines how I imagine the whole thing to look like: https://github.com/kubernetes/test-infra/pull/13342

LMK what you think and if I missed anything.

alvaroaleman:
@stevekuznetsov from my POV this is good to go

PR #1098, opened 2019-06-11
Addressed config changes in priority-and-fairness

MikeSpreitzer:


keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -906,6 +906,41 @@ spec:
...
+
+When the assured concurrency value of a priority level increases,
+additional requests are dispatched if possible.  When the assured
+concurrency value decreases, there is no immediate reaction --- this
+filter does not abort or suspend requests that are currently being
+served.
+
+When the queue length limit of a priority level increases, no
+immediate reaction is required.  When the queue length limit
+decreases, the queues are trimmed to the new length limit by rejecting
```
yue9944882: naturally draining the queues till the new limit seems better than trimming the queue? considering client-side retrying, rejecting the young end can also run out the client-side QPS quota or even lead to thundering requests herds.

MikeSpreitzer: Good idea.  I will make this change.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -902,6 +903,61 @@ spec:
       set: [ "system:serviceaccount:example-com:network-apiserver" ]
 ```
 
+### Reaction to Configuration Changes
+
+We do not seek to make our life easy by making any configuration
+object fields immutable.  Recall that objects can be deleted and
+replacements (with the same name) created, so server-enforced
+immutability of field values provides no useful guarantee to
```
lavalamp: However, it's perhaps easier to implement if we force people to delete and recreate. On delete of the api object, we wait for the queues to drain, then delete the in-memory construct.

This works universally without us needing to implement all the changes written below.

MikeSpreitzer: I am not sure I understand the thinking here.  This PR says that when the filter is informed of deletion of the API object the filter waits for the queues to drain then deletes the in-memory construct.  The concern in this paragraph is that we can not stop admins from effecting a change by deleting an object and then creating a replacement with the same name and different spec values.

lavalamp: The UID always changes if something is deleted & recreated--that can be detected reliably.

lavalamp: I guess we can use my mechanism to implement any change; we can detect changes by either the UID or the generation changing. I don't really object to what is in the text here, it just seems more labor intensive to write bug-free code for implementing those updates than always deleting and remaking.

MikeSpreitzer: Yes, the filter can detect that the UID changed.  But the point is that the filter has to deal with the possible changes.  Since the reaction to delete is _not_ immediately tear it all down, we can not reduce the problem of any change to {delete; create}.

yue9944882: the draining can work asynchronously, we don't have to wait until completely drained. note taht we're already using  UID-based approach in managing CR storage nowadays.

MikeSpreitzer: I hope there is no confusion between API objects and implementation objects in the apiservers.

Once a PriorityLevelConfiguration API object is deleted, the corresponding implementation objects linger in the apiservers until their queues have all drained.

If, in that interim, a replacement PriorityLevelConfiguration API object of the same name is created then this effectively becomes a change because those lingering implementation objects have to be updated to match the new API object.

lavalamp: > If, in that interim, a replacement PriorityLevelConfiguration API object of the same name is created then this effectively becomes a change 

My claim is that if we reference internal objects e.g. by UID, it is possible for both the old, draining object and the new object to co-exist.

MikeSpreitzer: Yes, we coud do that, but I do not think we want to.  It would be strange to have two (or even potentially more) sets of queues for what is logically the same priority level.  Administrators reference these things by name, as do FlowSchema objects.  It would be edge based rather than level based to say that the number of sets of queues for a given priority level depends on the history of the configuration object(s) for that priority level.

lavalamp: The flip side is that this condition will be transient (it really should not take long to drain the queues?), and it is easier to demonstrate that the system is correct when it never does any mutations--we only have to show that delete/replace works correctly, not that each individual modification does.

Using a {name, UID, generation} triplet for identification to admins should be pretty clear.

MikeSpreitzer: Yes, I understand that not implementing any changes makes for a simpler implementation.  But I think it is counter-intuitive, positively counter-productive in some cases, requires exceptional precision in references, and makes for a rather small difference in implementation.

Is the proposal here that when the AssuredConcurrencyShares of a PriorityLevelConfiguration object is decreased the filter will create a replacement priority level object with a new set of queues, in addition to retaining the old priority level object while its queues drain?  That makes for a temporary bump in the priority level's allowed concurrency, when the admin was just trying to get a decrease.  And the admin did that because something is going wrong with that priority level, this approach will make it even worse.

We can implement what this KEP proposes with one mutex per priority level (see the FairQueuingSystem abstraction in https://github.com/kubernetes/kubernetes/pull/78966), locked/unlocked once per reaction to request arrival or completion and held only briefly.

yue9944882: > That makes for a temporary bump in the priority level's allowed concurrency

yes that's indeed an issue. but it can also be applied to CR storage today. if the admin frequently update the CRD, there can be a bump in etcd connections, duplicated memory usage due to watch-cache, etc. both the flow-control system and CRD stuff should be considered system-critical configuration objects and only those cluster-admin should be privileged to operate the system, carefully and patiently.. what's more, mapping using a trio of `<name, uid, generation>` can also help simplifying the FQ abstraction i suppose? the interface looks a bit complex in the sketch pull

MikeSpreitzer: I updated https://github.com/kubernetes/kubernetes/pull/78966 , and simplified the FairQueuingSystem interface.  Now it has only three methods.  The latest rev of that PR implements the behavior I am suggesting here.  I think the code complexity is not unreasonable, and the runtime costs are almost always as low as I think we can reasonably expect.

PR #1099, opened 2019-06-12
Initial KEP for immutable fields

apelisse:


keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Proposal
+
+The behavior on mutation of immutable field is orthogonal to selection of
+immutable fields. If we accepted these two semantics at the same time, we would
+have to either have two different markers (one for selection and one for
+behavior) or have a single field with multiple options.
+
+For the moment, we understand that "Recursive" is the most important and simple
+selection mechanism, and "error" is the most natural way to handle mutations.
```
kwiesmueller: Do I understand this correctly that for now the other options (recursive with add/del, ignore) are not proposed as part of this kep (yet) and therefore there is no Go marker for them in the examples below?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+feature through this issue for example:
+https://github.com/kubernetes/kubernetes/issues/65973
+
+### Goals
+
+The goals are mostly twofold:
+- Provide a mechanism to document APIs by having that additional marker provide
+  information to users,
+- Automatically provide logic for developers to guarantee immutability of
+  specific fields.
```
sttts: - publish as a `x-kubernetes-immutable: true` extension in OpenAPI for external tooling.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Proposal
+
+This proposal attempts to create a concept of immutable and to visit some of
+what this can mean, but doesn't intend to propose all the possible
+semantics. The proposal should not close the door to further improvements.
+
+### Semantics
+
+We'll define immutable as "writeOnly", which means that the fields can only be
```
sttts: writeOnce

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+
+There are different semantics for a field to be immutable, especially if we
+consider non-leaf fields like lists and maps.
+
+#### Field selection
+
+- Recursive:
+  - Scalar fields are immutable (they can't change). Changing the
+    "type" will also typically be considered an error: `14 -> 14.0`. The field
+    should be purely round-tripped and will be compared as-is.
```
sttts: A different type leads to a different value. But we also assume (at least in structural schemas of CRDs, and in most native types) that types are not polymorphic. So type changes won't validate. But an important point here is that immutability will be orthogonal to validation, right?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+could potentially add that semantics later.
+
+It's also important to note that this concept is orthogonal to field ownership
+and list associativity. Ideally we would manage to keep the concepts entirely
+separate. e.g. An atomic list is a different concept from an immutable list.
+
+There are different semantics for a field to be immutable, especially if we
+consider non-leaf fields like lists and maps.
+
+#### Field selection
```
sttts: have problem to follow the meaning of the text below. Please explain what those items mean? Is it a grammer, an inductive definition of what can be immutable and what is always immutable?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+    and creating a new item in the list, which would be permitted by these
+    rules.
+- Mutable: A field set as mutable could cancel the recursive immutability.
+
+#### On mutation of selected fields
+
+- Error: mutation of an immutable field returns an error and the request is
+  rejected. No mutation is performed at all.
+- Ignored: mutation of the field (and/or sub-fields) are completely ignored, and
+  not persisted. They do not trigger an error.
```
sttts: Describe where this immutable checks will be in the request pipeline. I assume we do that before update validation, but after admission?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+Containers []Containers
+```
+
+### OpenAPI Extension
+
+Since the semantics of the "readOnly" tag in OpenAPI is not the one we're trying
+to have here.  We propose a new Kubernetes specific extensions, which has to
+allow for further changes in the semantics of our immutable marker:
+```
+X-Kubernetes-immutable: recursive,error
```
sttts: lower case

sttts: Which other values do you envision in the future? Why is `true` not enough? At least as a standard value. If we need more for compound types like maps and slices, we can use other values there. But I would prefer to have `true` at least for scalars.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+### Mutating admission chain
+Mutating admission chain would have the exact same effects as user changes,
+meaning that they wouldn't be able to change an object after creation. This is
+very similar to what is done today since validation for update is run AFTER all
+mutations. On creation, mutation are permitted.
+
+### Risks and Mitigations
+
+The risks are considered low. For CRDs, it would be possible for someone to set
+immutable fields on metadata, which could result in objects that can't be
```
sttts: no. We forbid specs for anything but `name` and `generateName`. Those are immutable anyway. So there is no risk.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,175 @@
...
+### Test Plan
+
+Not ready yet
+
+### Graduation Criteria
+
+This feature is mostly considered as a sub-feature of [Server-side
+apply](https://github.com/kubernetes/enhancement/pull/555) and will probably
+follow a very similar graduation line. It is expected to start directly as a
+beta feature and will graduate to GA along with Server-side apply.
```
sttts: why is it a subfeature? Other than shared code, I don't see a strong connection to server side apply. I wouldn't connect them.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: That's not really a goal? that's a mean to a goal?

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Thanks

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Only true for CRDs, not native type.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Once we define this as a boolean, we're kind of stuck to it. Some other types of values can be inferred (but not proposed yet) from the semantics section: "error", "ignore", "recursive", "non-recursive", "recursive-new-delete" (allow new and deleted items maybe?)

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Well, we need the OpenAPI schema for both CRDs and native type. We already have the logic in place in SSA to handle both.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Was thinking about `kubectl explain` as the usecase. Publishing the info there should be a goal IMO.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: It's better now. Thanks.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Am not speaking against using the SSA code, just that this is a different feature.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,177 @@
...
+    names are "immutables" for objects). Changing the keys would mean deleting
+    and creating a new item in the list, which would be permitted by these
+    rules.
+- Mutable: A field set as mutable could cancel the recursive immutability.
+
+#### On mutation of selected fields
+
+- Error: mutation of an immutable field returns an error and the request is
+  rejected. No mutation is performed at all.
+- Ignored: mutation of the field (and/or sub-fields) are completely ignored, and
```
sttts: Where do we have "ignored` semantics today? Are we sure we want that?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,177 @@
...
+set at creation time, and can never be updated after. Attempts to update an
+immutable field will result in an error (as opposed to being ignored), though we
+could potentially add that semantics later.
+
+It's also important to note that this concept is orthogonal to field ownership
+and list associativity. Ideally we would manage to keep the concepts entirely
+separate. e.g. An atomic list is a different concept from an immutable list.
+
+There are different semantics for a field to be immutable, especially if we
+consider non-leaf fields like lists and maps.
```
sttts: I miss one super important detail: what is the equality? Do we identify null and empty lists/maps? Is undefined and null the same?

Note: CRDs have a defaulting semantics which is strict in the sense that only undefined values are defaulted. There is no identification of null, empty and undefined. I would expect that immutability semantics match that for CRDs.

sttts: I miss one super important detail: what is the equality? Do we identify null and empty lists/maps? Is undefined and null the same?

Note: CRDs have a defaulting semantics which is strict in the sense that only undefined values are defaulted. There is no identification of null, empty and undefined. I would expect that immutability semantics match that for CRDs.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,177 @@
...
+validating webhooks for CRDs.
+
+Providing a new `// +immutable` marker would make the API more descriptive to
+users and help API developers by verifying these assertions automatically.
+
+## Motivation
+
+The kubernetes model tend to promote "immutable" objects rather than "delete and
+recreate" on changes approach, and users have expressed the need for such a
+feature through this issue for example:
```
matte21: I find the first sentence very confusing, what is the conveyed message? That API objects fields can currently be modified without having to delete and re-create the object?

apelisse: I'll rephrase :-)

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: I've seen API expose something like "initialCount". I suspect that the only behavior that makes sense when someone changes the intial count is to ignore that value.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: I'll rephrase then

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Yeah that makes sense

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Thought that's a bad API admittedly.

sttts: Adding this semantics encourages to create or keep those APIs. We should probably not do that.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: I think I rephrased that,  thanks

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Updated

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: tentatively rephrased

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Yeah, I don't think I disagree. Do we agree that we may have multiple "field selection" semantics in the future?

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: Added new "equality" sub-section

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: For maps and slices we will have multiple for sure.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,202 @@
...
+has to rely on either built-in validation for core types, or have to build a
+validating webhooks for CRDs.
+
+Providing a new `// +immutable` marker would make the API more descriptive to
+users and help API developers by verifying these assertions automatically.
+
+## Motivation
+
+Kubernetes tend to encourage "immutable" objects rather than "delete and
+recreate" on changes approach. Users have expressed the need for automatically
```
matte21: Better, but I'm still confused by `"immutable" objects`. What does it mean? That currently object fields can be updated with the object staying the same?

sttts: "There are resources in Kubernetes which have immutable fields by design, i.e. after creation of an object, those fields cannot be mutated anymore. E.g. a pod's specification is (mostly?) unchangeable once it is created. To change the pod, it must be deleted, recreated and rescheduled. Users want to implement the same kind of read-only semantics for CustomResources."

apelisse: Yeah, that looks great

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,202 @@
...
+  not persisted. They do not trigger an error.
+
+#### Equality
+
+Equality of two fields is very important:
+- Scalar fields are compared directly: fields and values must exactly match from old
+  object to new object.
+- Lists fields are compared using the same logic currently used for native
+  kubernetes types: Empty and null lists are equal. A missing list is NOT equal
+  to a null or empty list.
```
sttts: I would really like to have a way in CRD OpenAPI to specify a normalization behaviour for lists and maps:

```
x-kubernetes-normalize-null: empty
x-kubernetes-normalize-null: undefined
x-kubernetes-normalize-empty: null
x-kubernetes-normalize-empty: undefined
```

And then we should be strict here in the mutation test.

Also note that in Proto we have neither null nor empty by default. I still think we need a more broad discussion of null, undefined, empty values.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,202 @@
...
+- Same rule for list applies to maps and struct: An empty map and a null map are
+  considered equal. The set of fields in the map must be the same. Values are
```
sttts: The keys in the map must be the same.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: I don't disagree, I don't know if this should be blocking?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,204 @@
...
+or not. Things get more complicated for lists and maps, since we need to know if
+the flag applies to the list/map itself, to its members, or both.
+
+A few possible semantics are described here:
+
+- Non-recursive: For a list or a map, one can not remove or add new items, but
+  existing items can be modified.
+- Recursive: means that none of their field can change, and new fields are not
+  accepted.
+- Recursive with addition and/or deletion:
```
apelisse: Actually, this can be done by making the list/map mutable, but the values of the list/map immutable.

keps/sig-api-machinery/20190603-immutable-fields.md
matte21: That is extremely clear, thanks. Only thing IMO could be changed: `s/the same kind of read-only semantics/the same kind of semantics/`. "read-only" makes perfectly sense in this paragraph, but it could be confusing when reading the doc as a whole because elsewhere there are references to OpenAPI readOnly semantics and it's explicitly stated that it's not what is needed.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: At least we have to leave it open here and have a plan before marking it implementable. But I like to move forward with implementation before being very clear about this topic and where it is heading.

sttts: Or in other words: whatever we decide here will drive the API design of the whole CRD ecosystem. We should take the time it needs to understand the topic.

erictune:
Is it a goal that this could, in theory, be used to annotate the Pod type?
If so, it looks like recursive is insufficient.

I think we'd want to do:
```
type PodSpec struct {
...
// No new items can be added, but mutable fields of items can be modified.
// +immutable=keys 
Containers []Container
...
}

type Container struct {
	// +immutable
	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`

         // Can be updated.  
	Image string `json:"image,omitempty" protobuf:"bytes,2,opt,name=image"`
        ...
}
```

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,204 @@
...
+This proposal attempts to create a concept of immutable and to visit some of
+what this can mean, but doesn't intend to propose all the possible
+semantics. The proposal should not close the door to further improvements.
+
+### Semantics
+
+We'll define immutable as "writeOnce", which means that the fields can only be
+set at creation time, and can never be updated after. Attempts to update an
+immutable field will result in an error (as opposed to being ignored), though we
+could potentially add that semantics later.
```
erictune: Perhaps add the following, to make it explicit:

---
Clients implement  [idempotent creation](https://kubernetes.io/docs/reference/using-api/api-concepts/#standard-api-terminology) by retrying creation, and they currently do not get errors if the same creation is tried twice.  This KEP will not change that.   Specifically, two consecutive POSTs of the same object, which are mutated the same way by mutating admission controller (which is the typical case), will both succeed.  If an error were returned the second time, and the client did not see the first success response, the client would be confused.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,204 @@
...
+#### Field selection
+
+Scalar fields have an obvious "selection" mechanism: they are either immutable
+or not. Things get more complicated for lists and maps, since we need to know if
+the flag applies to the list/map itself, to its members, or both.
+
+A few possible semantics are described here:
+
+- Non-recursive: For a list or a map, one can not remove or add new items, but
+  existing items can be modified.
```
erictune: This is needed to implement an object with the same semantics as Pod.

keps/sig-api-machinery/20190603-immutable-fields.md
erictune: This is a great goal.

erictune:
I think we can get by with just one "selection" behavior, if we want to keep it simple.  That behavior is:
- `+immutable` tag on a scalar FieldDecl means that field is immutable.
- `+immutable` tag on a map of list FieldDecl means that the set of map keys cannot change. Each item is validated separately, comparing same keys of old/new.  This tag does not affect item validation.   
- `+immutable` tag on a list FieldDecl means that that the list length cannot change.  Each item is validated separately, comparing same keys of old/new.  This tag does not affect item validation.

apelisse:
Agreed, that makes a lot of sense to me.

erictune:
The above semantics supports the Pod/Containers use case, where you want to allow some fields to be updated on list items, but don't allow adding list items.
It also supports where you want to allow new list items.

The Deployment use case is this: you want to have a PodTemplate where everything is mutable.  Today this is handled by having a separate copy of the type definition, which has lacks `+immutable` specifiers.  I think this is a good-enough pattern for other Deployment-like use cases, and so there is no need for special support for this case (like "Recursive with addition and/or deletion").

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: I think we should make a decision and enforce it rather than offer too many complicated options. We currently consider empty and null to be equal for existing type, and I think that we should just tell people that's how things are going to work.

sttts: We decided against that for defaulting. Also in validation null is distinguishable from undefined. It would be very odd to have immutable checks to differ.

michaelgugino:
I'm interested in immutable fields in CRDs as well.

Would it be simpler to do something like:
```
spec:
  immutableFields:
  otherField
  ...
```
instead?  That seems simpler to implement to me, less to document, anything that's immutable, put in there.  Seems like this could be implemented rather uniformly.

keps/sig-api-machinery/20190603-immutable-fields.md
jpbetz: I'd like to reboot this discussion.

Please let me know if I'm missing something important, but when reviewing defaulting, validation and equality, it looks like we have:

**`Semantic.DeepEqual`:**

| ==               | missing field | null | empty collection |
|------------------|---------------|------|------------------|
| missing field    | yes           | no   | no               |
| null             | no            | yes  | yes              |
| empty collection | no            | yes  | yes              |


**Defaulting:**

Applied only to missing fields. It does not default null or empty collections or have any opinion about null vs. empty.

**Validation:**

Has "nullable" rule.

  - non-nullable fields have strict equality.
  - nullable fields have "null == empty collecion" equality. 

So...

If immutability had "null == empty collecion" equality when checking for mutations, then we are allowing for field values that readers can distinguish between (any client can choose, if they want to, to handle null differently than an empty collection) to be considered equal throughout the system for the long term (forever?).

If immutability had strict equality for mutations, than we really need to update `Semantic.DeepEqual` and whatever other equality checks we have to match. Is this a API breaking change?

Is the logical next step to review `Semantic.DeepEqual` to see if that is something that is eligible for being changed?

sttts: Have to remember where we left off with the disucssion. Compare https://github.com/apelisse/enhancements/pull/1 where we went into detail more already.

sttts: As markdown tables are so tedious, I started https://docs.google.com/spreadsheets/d/1KpX_ZAgw8LKQMsgOV-UMhD2THzjljAGN31_8lCRquLA/edit?usp=sharing.

sttts: My main problem with any non-strict behaviour for CRDs is that it feels very odd to have different objects in etcd, but consider them equal. 

For native resource with proto encoding this is not the because proto unifies undefined, null and empty all to one value und unmarshal those to null for comparison.

I wonder whether we should talk about normalization for CRDs before talking about equality:

- we have one normalization already: undefined => default value
- we don't have normalization for empty
- we don't have normalization for null

With normalization happening as part of our deserialization stack, we can separate both concerns and use strict equality in immutability checks.

Today we use semantic equality mostly because JSON and Proto have different normalization behaviours (partly driven by omitempty as well).

| normalization JSON omitempty | |
| - | - |
| undefined | undefined |
| null | undefined |
| empty | undefined |

| normalization JSON | |
| - | - |
| undefined | null |
| null | null |
| empty | empty |

| normalization Proto omitempty/non-omitempty | |
| - | - |
| undefined | undefined |
| ~~null~~ | ~~undefined~~ |
| ~~empty~~ | ~~undefined~~ |

| normalization CRD omitempty/non-omitempty | |
| - | - |
| undefined | undefined |
| null | null |
| empty | empty |

| normalization CRD default=null | |
| - | - |
| undefined | null |
| null | null |
| empty | empty |

| normalization CRD default=empty | |
| - | - |
| undefined | empty |
| null | null |
| empty | empty |

keps/sig-api-machinery/20190603-immutable-fields.md
jpbetz: xref https://github.com/kubernetes/enhancements/pull/1099#issuecomment-504537706 and https://github.com/kubernetes/enhancements/pull/1099#issuecomment-504557970 for more detail on how Pod immutability can be supported

keps/sig-api-machinery/20190603-immutable-fields.md
jpbetz: If I'm understanding all the normalization rules right, then for native types we have something like this?

| normalization native, client speaking JSON | non-omitempty | omitempty |
|--------------------------------|-----------|-----------|
| undefined                      | empty | undefined |
| null                           | empty | undefined |
| empty                          | empty | undefined |

| normalization native omitempty/non-omitempty, client speaking Proto | |
|--------------------------------|-----------|
| undefined                      | undefined |
| null                           | undefined |
| empty                          | undefined |

And we don't have omitempty empty for CRDs?

Would it make sense to normalize the same way for CRDs as we do for native types (using proto)? I.e. normalize null and empty to undefined. We'd need to normalize all data received from clients and all data read from etcd (since we have non-normalized data at rest). If we did that, we should be able to do strict equality in the api-server code.

I'm guesses that to be consistent we should also normalize native types stored as json like we do for CRDs?

jpbetz:
@apelisse Would it be okay to add the two main open issues (equality, recursive) to this PR and merge it as provisional? We can open followups to resolve those, hopefully this week.

apelisse:
Sure, I'll update the KEP

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: not sure about the first table, what are the values?

> And we don't have the concept of omitempty empty for CRDs?

There is no Golang type that could have omitempty in the request handler logic. JSON unmarshalling for CRD is faithfully, i.e. does not swallow or normalize anything.

I am not convinced we want to start putting up fuzzy glasses for CRDs by adding normalization by default. Moreover, we cannot even do that without breaking compatibility, at least not in general (we could do it for equality only).

I don't like the idea of having the normal REST logic with strict empty-null-undefined logic, but immutability to act differently. That's pretty counter-intuitive.

I tend to think about normalization as a property of the fields, i.e. the schema properties. For example we can express nullable fields (which preserve null in all cases, and empty and undefined) using special Golang types (like extra pointers) and at worst custom marshallers. What we don't do today is publishing of this as part of OpenAPI, and in general we never formalized that. But my point is that there are native types which do not normalize prior to persisting values and before comparing, it is just not what we get by default with naive or lazy Golang type definitions and our serializer stack.

In other words, we might want to formalize the normalization of any slice and map in the schemas, for native types and for CRDs. CRD schemas could get a `x-kubernetes-normalize-empty: undefined` vendor extensions (not necessarily for 1.17, could be added later). Equality should be strict, relative to the defined normalization (which is not the indenty normalization for native types then).

Also for later introduction of protobuf support for a CRDs, we can require a certain normalization that is compatible with proto. I.e. the CRD validation of a CRD instance with proto enabled, would validate the proto numerals are all specified, and it would validate that normalization is compatible.

sttts:
```
properties:
  foo:
    x-kubernetes-proto: 6
    x-kubernetes-normalize-empty: empty
    x-kubernetes-normalize-null: empty # if nullable
    x-kubernetes-normalize-undefined: undefined
    nullable: true
```

`validation error: properties.foo.x-kubernetes-normalize-undefined must be empty if x-kubernetes-proto is set.`

jpbetz: Thanks for explaining. I don't think I appreciated the extend our existing golang type definitions play into the situation. That's good to know.

From a usability perspective, it wish we could provide a sane default normalization rule that users could get by default and not have to specify any normalization settings explicitly, but I agree we can't just start doing that for CRDs since it breaks backward compatibility.

I really like the idea of specifying the normalization we expect then validating our encodings and go types are all aligned with that.

apelisse:
Replaced by #1099 1265

PR #1110, opened 2019-06-21
[WIP] Apply Scalability KEP

jennybuckley:


keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+Based on initial numbers, this solution has the most advantages:
+etcd objects (based on pods above) only grow by 30%
+listing objects has nearly zero cost since we don’t send unless requested,
+patches and updates can deserialize the data very quickly directly in protobuf, and process the data
+API is still “usable”
+
+This could be implemented as a subresource similar to scale. (something like “/api/v1/pods/managedfields”) The subresource would decode the stored form of managedFields and exposes an object of type meta/v1.ManagedFields which would be highly structured. The form would be similar to or the same as how managedFields is stored today. This part could be implemented at a future date.
+
+### Add a default specifier syntax to managed fields
```
apelisse: This solution is not ideal because we have fields that are owned by no-one (defaulted fields) and should never conflict.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+Some structured-merge-diff operations (like compare) also create many small objects that are not needed. Most “field set” operations are not done in place.
+
+### Experiments
+
+The following experiments convinced us that the problem is mostly the allocations of small objects. Latency in etcd because of read/write with bigger objects, and/or data transfer was not a significant issue.
+
+For all these scenarios, we’ve created a deployment with 5,000 pods. Pods that have the managedFields (when server-side apply is enabled) are obviously larger. Here are some numbers for the comparison, with the objects in protobuf format:
+
+- Server-side apply disabled:
+  - A pod is about 8,137 bytes
```
smarterclayton: bytes in memory or serialized?  historically pods in memory are 2-10x the size of serialized bytes due mostly to golang maps

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+The internal format could be stored in protobuf, which would make it much smaller than json (json has a lot of repeated fields that make it extremely large).
+
+### Only update managedFields if previously applied to
+
+Most objects created with post or put are never applied to, so tracking managedFields on them is a waste of resources. If an object without managedFields does get applied to, we can still generate a single managedFields entry for all the existing fields. The generated managedFields conflicts set will be the same, but we won’t have any way of knowing know who is being conflicted with. CPU and memory cost on the apiserver will be dependent on the number of objects managed by server-side apply. It would increase more gradually as controllers decided to adopt SSA. Adoption could be decided on a controller by controller basis, weighing the cost vs. benefits each time. If someone applies to an object created by post or put, the conflicts they see would not contain information about the timestamp or field manager, so conflicts become less useful.
+
+### Storing in one format, serving in another
```
apelisse: Whether we do it as a subresource or not, this is the favorite solution because it significantly increases the speed of all operations, keeps a "kubernetes"-style API and reduces the etcd impact. Main drawback is that we probably need two fields: 1 for the storage, 1 that is served.

keps/sig-api-machinery/20190621-apply-scalability.md
apelisse: This is serialized as protobuf

apelisse: (I made pretty big pods)

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+TODO
+
+## Proposal
+
+We have considered a few overlapping solutions:
+
+### Only send managedFields if requested:
+
+Add a query parameter (or some other mechanism) to allow users to request to see the managedFields field in metadata. If the parameter is missing,  strip the field on reads and writes (in responses). The benefit here is we can skip serializing the field most of the time. Most clients won’t and shouldn’t have to understand how to read this field. So making them receive it is a waste of resources, and confusing.
```
smarterclayton: Since we've tried to avoid sending back parts of objects (instead preferring the transform use case), this has a pretty high bar to clear for design complexity (even if it might be the simplest).

Also, not returning all objects is confusing because PUT semantics should be resetting that field, so PUT would not be correct if it didn't clear managed fields.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Store managedFields as a blob
+
+Instead of a highly structured nested map, with lots of small fields, we could store it as a blob. The blob would save on allocation/GC of small objects when deserializing. We could also gzip the entire thing to save space. The downside is that it would make the field harder/impossible for a user to understand. This is mitigated by combining with solution 1, or by client-go/kubectl unzipping the field automatically.
```
smarterclayton: Alternatively revisit the serialization and choose a slightly more complicated serialization but with less size.  Maps are very expensive in memory, strings are less expensive, integers and slices are basically free.

For instance, an array with index references to protobuf fields, or internal index references.  Slices should be mostly free in deserialization, although field names will each require an allocation.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Convert managedFields directly to internal fieldpath
+
+On updates and applies, the apiserver needs to convert managedFields into an internal representation. We originally decoupled the managedFields API from the internal representation of a fieldset, to allow arbitrary changes to the internal representation without breaking the API. However converting from a serialized config to a gostruct and then into a different struct causes a lot of allocations of small objects and stress on the garbage collector. Instead, we can directly serialize as the internal representation.
```
smarterclayton: Or you could use protobuf field ids which are part of the external API for a given version.  That still requires tree behavior though.

keps/sig-api-machinery/20190621-apply-scalability.md
smarterclayton: Note that you could store it as a json blob fairly easily in etcd and it wouldn't be that difficult for clients that care.  You'd pay the cost to the deserializer to check the syntax (in json) but protobuf would be free.

Basically a typed RawExtension that was less generic (uses json.Message / equiv to avoid deserialization).  You'd still have to check for valid structure before saving to etcd, but you could avoid the check on read.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+The internal format could be stored in protobuf, which would make it much smaller than json (json has a lot of repeated fields that make it extremely large).
+
+### Only update managedFields if previously applied to
+
+Most objects created with post or put are never applied to, so tracking managedFields on them is a waste of resources. If an object without managedFields does get applied to, we can still generate a single managedFields entry for all the existing fields. The generated managedFields conflicts set will be the same, but we won’t have any way of knowing know who is being conflicted with. CPU and memory cost on the apiserver will be dependent on the number of objects managed by server-side apply. It would increase more gradually as controllers decided to adopt SSA. Adoption could be decided on a controller by controller basis, weighing the cost vs. benefits each time. If someone applies to an object created by post or put, the conflicts they see would not contain information about the timestamp or field manager, so conflicts become less useful.
```
smarterclayton: This feels like the cheapest up front but most expensive long term.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Storing in one format, serving in another
+
+Another possible solution would be to combine many of these solutions:
+
+Some combination of storing the field managers as one big blob in etcd, using the internal fieldpath representation, as gzipped protobuf, would give us a much lower overhead in etcd and very fast serialization/deserialization. This is especially true if deserialization doesn’t need to know about the specifics of managedfields (hence just deserializing as a blob).
+
```
smarterclayton: Just as an off the cuff, I think you could get 90% of the benefit by declaring the field a raw extension and adding validation logic to it that does late deserialization.  Use JSON maps (same format) but simply lazily decode it only on clients that care about it.

keps/sig-api-machinery/20190621-apply-scalability.md
smarterclayton: From previous tests with Go 1.6 or 1.7, a single Golang map[string]string was ~360 bytes or so.  So every labels map, annotations, nested map, etc was roughly this expensive.  It might be interesting to know what your API server / controller manager memory usage looked like between scenarios 1 and 2

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+2. Server-side apply enabled, 
+3. Server-side apply disabled, but with an extra annotation. The annotation is about 5,000 bytes to compensate for the difference in object size with server-side apply.
+4. Server-side apply disabled, with 320 small annotations. Those annotations total about 5,000 bytes to simulate the small object allocation.
+
+The average latency are as follows:
+
+| Scenario | LIST | PATCH | LIST while PATCH |
+|---|---|---|---|
+| 1. |360.4 ms | 8.0 ms | 361.0 ms |
+| 2. | 1294.5 ms | 38.5 ms | 2529.6 ms |
```
deads2k: This is interesting. "LIST while PATCH" shows a marked difference between this and case 4.  Did you investigate the reason or will that come separately?

keps/sig-api-machinery/20190621-apply-scalability.md
deads2k: When the gostructs are identical, we do an unsafe pointer assignment today.  How much slower does the implementation get working against external?  We feared it before with iterating slices instead of querying maps, but practically the change was negligible.

keps/sig-api-machinery/20190621-apply-scalability.md
deads2k: I'm a -1 on this approach.  In addition to not knowing the creator, don't we also lose the ability to distinguish between defaulted or mutated (admission) fields and what the user set?

keps/sig-api-machinery/20190621-apply-scalability.md
deads2k: We've done this in a couple spots in the openshift API.  As long as you have good helpers for digesting it, using a runtime.Unknown isn't a bad way to pass the bits around.

keps/sig-api-machinery/20190621-apply-scalability.md
deads2k: > Just as an off the cuff, I think you could get 90% of the benefit these approaches have by declaring the field a raw extension a

We did this for image metadata in OpenShift.  Client have to know they want it to get it an manipulate it, but it's fairly manageable.  When they mutate, they can place the converted bit into RawExtension.Object and clear RawExtension.Raw and it *should* serialize back properly.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+Another possible solution would be to combine many of these solutions:
+
+Some combination of storing the field managers as one big blob in etcd, using the internal fieldpath representation, as gzipped protobuf, would give us a much lower overhead in etcd and very fast serialization/deserialization. This is especially true if deserialization doesn’t need to know about the specifics of managedfields (hence just deserializing as a blob).
+
+Serving the API as it currently is (or maybe changing it if we believe we need to change what we serve), but only on demand so that the price of deserializing would only be paid by clients that specifically request it.
+
+Based on initial numbers, this solution has the most advantages:
+etcd objects (based on pods above) only grow by 30%
+listing objects has nearly zero cost since we don’t send unless requested,
+patches and updates can deserialize the data very quickly directly in protobuf, and process the data
```
deads2k: If you did this as proto, how does it help CustomResources?

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,154 @@
...
+API is still “usable”
+
+This could be implemented as a subresource similar to scale. (something like “/api/v1/pods/managedfields”) The subresource would decode the stored form of managedFields and exposes an object of type meta/v1.ManagedFields which would be highly structured. The form would be similar to or the same as how managedFields is stored today. This part could be implemented at a future date.
```
deads2k: This seems like it would have atomicity issues if a  user wanted to claim ownership. Wasn't there a use-case where we manually twiddled the field ownership?

keps/sig-api-machinery/20190621-apply-scalability.md
jennybuckley: >  Client have to know they want it to get it an manipulate it

Does this mean that it is only served when specifically requested? or just that it is more complicated to manipulate because the client has to know how to do the (de)?serialization, so only clients which already know about it can actually change it.

keps/sig-api-machinery/20190621-apply-scalability.md
jennybuckley: Storing it as a blob, we can still use the proto form of managedfields, even if the object containing the blob is in json/yaml or whatever format, if we published the proto for the type.

keps/sig-api-machinery/20190621-apply-scalability.md
jennybuckley: That's true, this would remove the ability to easily tie a specific managed fields to a specific version of the object, unless resource versions were used, or the managedfields sub resource  could serve the entire object + the managedfields

keps/sig-api-machinery/20190621-apply-scalability.md
apelisse: Both LIST and PATCH do perform lots of allocations: LIST because of deserialization, PATCH because of deserialization and conversion from external/internal/external type

keps/sig-api-machinery/20190621-apply-scalability.md
apelisse: > Also, not returning all objects is confusing because PUT semantics should be resetting that field, so PUT would not be correct if it didn't clear managed fields.

Because the managed fields can already be dropped by typed client that are not aware of it, we already ignore updates that just completely clear it. This wouldn't really be a new use-case.

keps/sig-api-machinery/20190621-apply-scalability.md
wojtek-t: 1. Storing json blob helps only one way (i.e. read path) where we don't need to recompute it. On patches, we still need to recompute it, so deserialize, modify and serialize, so the overhead there will still be the same.

2. Forgive my ignorance (I didn't look into exact representation of managed fields), but I actually agree with Clayton`s suggestion that we should consider more complicated form of storing that data (possibly with good helpers with digestion).
Assuming that managed-fields is actually a nested map, we should definitely consider flattening it, e.g.:
["metadata.name": "name", "spec.nodeName": "nodeName", ...]
While maps are expensive, a single map is actually not that bad, given we already have maps in our objects.
If we would combine it with protobuf tags (e.g. "metadata.name" == "1.1") we could save a lot.

keps/sig-api-machinery/20190621-apply-scalability.md
wojtek-t: I'm also -1 on this. What is the most fightening for me is: "Adoption could be decided on a controller by controller basis" - we don't want to have infrastructure that is unusable at scale and only controllers that don't do much can use it.

keps/sig-api-machinery/20190621-apply-scalability.md
apelisse: > Is the managedField conceptually map[string]string or map[string]interface{} with some multilayer deserialization? - that's one thing that can affect it.

Yes, it is `map[string]interface{}` with multilayer deserialization

> Was your master machine (apiserver) cpu-saturated in these experiments?

I haven't checked that ..

keps/sig-api-machinery/20190621-apply-scalability.md
apelisse: > 1. Storing json blob helps only one way (i.e. read path) where we don't need to recompute it. On patches, we still need to recompute it, so deserialize, modify and serialize, so the overhead there will still be the same.

Yeah, I think we all agree on that :-/ That's why we also wanted to store it in a format that was convenient for patch, that is, in internal representation in protobuf format.

> Assuming that managed-fields is actually a nested map, we should definitely consider flattening it

We're storing the "set" of fields that are owned by each actor of the system. Today, we're using a trie, which I think is "complicated with less size". We could try using a list of such fields with their full path, which would be less complicated and bigger, and possibly faster.

wojtek-t: > We could try using a list of such fields with their full path, which would be less complicated and bigger, and possibly faster.

I'm pretty sure it will be faster. (we both agree it will be less complicated too).
And I actually would like to see data about the impact on size - I'm not sure it's that significant. And I think we should also consider what I mentioned about using protobuf tags instead of names, and that should significantly help (I actually kind of hope that it may be enough).

apelisse: > And I think we should also consider what I mentioned about using protobuf tags instead of names

Quick notes on that.

Pros:
- Shorter than strings, leading to smaller representation both in memory and serialized,
- Less allocations than strings,

Cons:
- API makes no sense unless you know what each protobuf number means, absolutely NOT human readable,
- Protobuf numbers are not in the API at all, there is no coupling between the two: I don't know how we would implement it, and certainly not how people would use it,
- No way to get this to work for CRs (I've heard rumors that we might get protobuf for CRs, but we don't have anything yet).

I'm fairly against it. I think it would really hurt the API for an uncertain performance boost.

> I'm pretty sure it will be faster. (we both agree it will be less complicated too).

Are you suggesting we do that on top of the "RawMessage", or we just serialize as lists of strings directly?
We would still have to parse these and generate the trie that we use internally to run the set operations that we have.

wojtek-t: Re cons:
1) I agree it's much less human-readable, but I don't actually agree it's "absolutely NOT human readable". It's possible for a human to get those numbers and parse it. And it's really easy to create helpers to those conversions automatically for humans.
2) I actually would say that these are kind of API. Auto-generated protobufs are part of the "api" repo and the tags are all there, e.g. for a Pod:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/generated.proto#L2624
3) CRs are more tricky, but I actually can imagine doing first a simple check to see if protobuf exists for a given type and if not using names instead of tags.

So the cons are relatively weak in my opinion.

> Are you suggesting we do that on top of the "RawMessage", or we just serialize as lists of strings directly?
We would still have to parse these and generate the trie that we use internally to run the set operations that we have.

I wouldn't do RawMessage, but I'm also not against it. Simply list of pairs of strings, or maybe even map[string]string should make that much better.

> We would still have to parse these and generate the trie that we use internally to run the set operations that we have.

That makes me thing that slice of pairs of strings is better. If we would enforce that keys are sorted in trie order (which is what we can easily do and introduce validation), then building a trie from that representation (and vice versa) is trivial.

apelisse: I'm not sure why we need "pairs". We're simply tracking a list of fields, if we have a single string representing the path to a specific field, it's good enough.

wojtek-t: Hmm - we basically need mapping from field to its owner, right? That's why I say we need pairs.

Also to copy from our offline conversation about protos:
well - we need more numbers, but basically conceptually tags save us the biggest amount of data (in json vs proto) and there will be much less data to serialize/deserialize
but if you show me data of experiments that shows that this is not needed, I would be more than happy with it (I just think that simple flattening may not be enough)

apelisse:
cc @kwiesmueller

keps/sig-api-machinery/20190621-apply-scalability.md
deads2k: I'm biased towards trying to be embed something recognizeable to most clients and I think JSON meets that need better than proto.  If we have many, small, nested maps perhaps it makes more sense to keep a hierarchical structure and simply use short slices instead.  The gains in conversion being able to do a simple pointer assignment may offset the o(1) behavior of maps if your slices are short.

keps/sig-api-machinery/20190621-apply-scalability.md
```diff
@@ -0,0 +1,158 @@
...
+The average latency are as follows:
+
+| Scenario | Object Size | LIST | PATCH | LIST while PATCH |
+|---|---|---|---|---|
+| 1. | 8.1 KB | 360.4 ms | 8.0 ms | 361.0 ms |
+| 2. | 13.2 KB | 1294.5 ms | 38.5 ms | 2529.6 ms |
+| 3. | 13.2 KB | 480.8 ms | 11.3 ms | 500.2 ms |
+| 4. | 13.2 KB | 1260.1 ms | 9.7 ms | 1503.2 ms |
+| 5. | 13.7 KB | 544.0 ms | 26.6 ms | 735.1 ms |
+| 6. | 27.2 KB | 1791.3 ms | 39.7 ms | 3803.4 ms |
```
apelisse: We did update this table with new numbers, which I think are pretty encouraging.

keps/sig-api-machinery/20190621-apply-scalability.md
wojtek-t: As I mentioned above, I completely buy the argument of human-readability, if we can get enough gains from json-like format.

That said, I don't really understand why you're pushing for NOT flattening that - flattening it would visibly reduce amount of allocations, right? And it's similarly understandable for users.

keps/sig-api-machinery/20190621-apply-scalability.md
wojtek-t: So apparently that is much worse - is it because of a significant nesting in pod?

So going back to point 5 (simple serialization as list), is there any way to reduce the size of this map further?
I'm assuming that in 99% of cases, majority of spec will be written by a single actor, whole status written by kubelet, nodeName by scheduler, and that's mostly it, right?
Having additional 5kB of data to pass that information seems like quite huge.

The results look promissing enough that I'm not that afraid about some corner cases where all fields are written by different parties, but in majority of cases I would expect it to be like I described above.

@deads2k - for thoughts too

deads2k: Looking at the result of 5 compared with 3, I think I'd look at trying to close the 50% bump in "list while patch" or the 2.5x bump in the patch performance before trying to shrink the size.  Can the difference be reproduced in a benchmark where you have good shot at seeing where it comes from?

keps/sig-api-machinery/20190621-apply-scalability.md
smarterclayton: > Because the managed fields can already be dropped by typed client that are not aware of it, we already ignore updates that just completely clear it. This wouldn't really be a new use-case.

That is very surprising to me.  That is not what PUT means.

keps/sig-api-machinery/20190621-apply-scalability.md
smarterclayton: No, it's always served.   But deserialization is expensive for two reasons - structure (json boundaries) and memory allocation, and memory allocation dwarfs structure by far (microbenchmark encoding/json.Unmarshal(complex, &json.RawMessage{}) which just verifies it's valid JSON vs decoding to a map or struct and it will be 90% of work).  So we get most of the benefit avoid deserialization except when the client needs it, and otherwise pass it around unmodified.

Since only a limited number of clients actually care, you can define rules that are reasonable for our client types.

Now, other clients will be impacted (hey, why does my python client slow down) and that's definitely a risk, but I think we can justify that if the python client code base gets simpler.  And other languages could if they cared do something simpler.

keps/sig-api-machinery/20190621-apply-scalability.md
lavalamp: PUT should replace the object. But, it doesn't make sense to change both managed fields and other things with the same request, since all updates imply changes to managed fields. There definitely needs to be a way to clear the managed fields.

keps/sig-api-machinery/20190621-apply-scalability.md
lavalamp: Let's take this thread to https://github.com/kubernetes-sigs/structured-merge-diff/issues/99.

Proto numbers would make the objects non-self describing and don't currently work for CRDs.

keps/sig-api-machinery/20190621-apply-scalability.md
lavalamp: We have basically done this. It seems like increased object size is still a problem.

keps/sig-api-machinery/20190621-apply-scalability.md
lavalamp: We need to update this with the latest numbers.

keps/sig-api-machinery/20190621-apply-scalability.md
lavalamp: @jennybuckley can we update this with the latest numbers?

PR #1124, opened 2019-07-01
Adds metadata verification

chuckha:
Validate KEP metadata

`verify-all-metadata` is omitted from `verify` because `verify-all-metadata` is failing as there are many errors in metadata validation across old KEPs.

A couple of next few steps:

* fix the validations on every existing KEP
* add this as a non-blocking prow pull job to raise attention that this is being checked
* we should add another verify step that only verifies the modified files in a PR, `kepval` is set up nicely to do that, but requires some understanding of k8s-infra

cc @justaugustus

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -10,6 +10,19 @@
   version = "v0.3.1"
 
 [[projects]]
+  branch = "master"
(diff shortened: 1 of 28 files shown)


justaugustus:
Thank you again for working on this, @chuckha!
/lgtm
/approve

PR #1153, opened 2019-07-19
Update "less object serializations KEP" and move it to implementable state

wojtek-t:


wojtek-t:
@prow - why it isn't merging?

BenTheElder:
this was https://github.com/kubernetes/test-infra/issues/13551

PR #1163, opened 2019-07-23
Tools hack

chuckha:
/assign @justaugustus 

an alternative to #1162 

This is the tools-hack that would make go.mod work with a vendor.

I'm really curious about the preference for this over the original solution. We've been talking about this for a while and folks seem split on the issue. If you feel like sharing your take, I'd be very interested to hear it.


chuckha:
/close
discussed offline, we're going to try the no-vendor approach and see how that goes.

PR #1165, opened 2019-07-23
Remove SelfLink KEP implementable

wojtek-t:
As this this required to happen before feature freeze


deads2k:
/lgtm

Yeah, sorry, I didn't realize the original wasn't in this state.

PR #1170, opened 2019-07-26
KEP: Graduate kube-scheduler component config to v1beta1.

jfbai:
#785


justinsb:
Something that came up when I tried to use these types, which I raised with wg-component-standard, is that we probably want all fields to be nilable: https://groups.google.com/forum/#!topic/kubernetes-wg-component-standard/hO7Lmi0cwQU

I owe a KEP on that, but we might not want to push componentconfig out of alpha given that.

justinsb:
I sent (a draft of) the KEP I owe :-) https://github.com/kubernetes/enhancements/pull/1173  I figure we need to decide whether that impacts beta, and if so how to prioritize/schedule.

bsalamat:
/hold

I just saw @justinsb comments.

PR #1189, opened 2019-07-30
crd-defaulting: add details about v1 API

sttts:


deads2k:
This is consistent with https://github.com/kubernetes/kubernetes/pull/79604

/lgtm

PR #1200, opened 2019-08-05
Addressed both default behavior problems in priority-and-fairness

MikeSpreitzer:


MikeSpreitzer:
/cc @yue9944882 
/cc @aaron-prindle 
@mars1024

MikeSpreitzer:
@MikeSpreitzer

PR #1212, opened 2019-08-10
KEP: add ReadOnlyBindAddress to KubeletConfiguration.

jfbai:
The KEP for #1208

### DIFF
--- a/keps/sig-node/20190809-add-readonly-bind-address.md
+++ b/keps/sig-node/20190809-add-readonly-bind-address.md
@@ -1,0 +1,110 @@
+---
+title: KEP Template
+authors:
+  - "@jfbai"
+owning-sig: sig-node
+participating-sigs:
+  - sig-cluster-lifecycle
+reviewers:
+  - "@mtaufen"
+  - "@mattjmcnaughton"
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2019-08-09
+last-updated: 2019-08-09
+status: implementable
+---
+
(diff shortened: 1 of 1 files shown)


jfbai:
/cc @mtaufen 
/cc @mattjmcnaughton
/cc @yujuhong
/cc @tallclair

PTAL, thanks a lot. :)

mtaufen:
@jfbai sorry for the rather long delay in addressing this KEP.

I asked @tallclair about this. The long term direction is still to get rid of the insecure server altogether, but this could be a reasonable in-between option as long as it doesn't distract from eventually removing the insecure server.

@tallclair also mentioned that localhost doesn't work for Windows containers, so this isn't necessarily a solution for Windows. 

I'm ok with this moving forward if we update the KEP to note that it isn't an alternative to someday removing the insecure server.

jfbai:
> @jfbai sorry for the rather long delay in addressing this KEP.
> 
> I asked @tallclair about this. The long term direction is still to get rid of the insecure server altogether, but this could be a reasonable in-between option as long as it doesn't distract from eventually removing the insecure server.
> 
> @tallclair also mentioned that localhost doesn't work for Windows containers, so this isn't necessarily a solution for Windows.
> 
> I'm ok with this moving forward if we update the KEP to note that it isn't an alternative to someday removing the insecure server.

@mtaufen Thanks a lot, I add a note. PTAL.

mtaufen:
Thanks. This LGTM. 

/lgtm

mtaufen:
@derekwaynecarr @dchen1107 WDYT? Can you approve?

PR #1214, opened 2019-08-19
Updates default behavior for the request-management KEP to sync up with implementation

yue9944882:
/sig api-machinery

### DIFF
--- a/keps/sig-api-machinery/20190228-priority-and-fairness.md
+++ b/keps/sig-api-machinery/20190228-priority-and-fairness.md
@@ -710,11 +710,13 @@ queue’s virtual start time is decremented by G - S.
 ### Example Configuration
 
 
-For kube-apiserver self-maintaining, solving priority-inversion, including requests from: 
+For kube-apiserver self-maintaining, solving priority-inversion, including requests 
+from: 
 
-1. kubectl requests from admin
-2. apiserver loopback requests
-3. delegated auth/admission requests
+1. requests from system-privileged group, including kubectl requests from admin and 
+apiserver loopback requests.
+2. delegated authentication/authorization API groups, including delegated delegated 
+authentication/authorization requests.
 
 ```yaml
 kind: RequestPriority
@@ -733,7 +735,7 @@ For kubernetes cluster self-maintaining, including requests from:
 # 
 kind: RequestPriority
 meta:
-  name: system-high
+  name: system
 spec:
   assuredConcurrencyShares: 100
   queues: 128
@@ -749,7 +751,7 @@ For the other essential system in the kubernetes cluster. covering requests from
 ```yaml
 kind: RequestPriority
 meta:
-  name: system-low
+  name: workload-high
 spec:
   assuredConcurrencyShares: 30
   queues: 1
@@ -761,7 +763,7 @@ For normal workloads:
 ```yaml
 kind: RequestPriority
 meta:
-  name: workload
+  name: workload-low
 spec:
   assuredConcurrencyShares: 30
   queues: 128
@@ -786,20 +788,20 @@ Some flow schemata.
 ```yaml
 kind: FlowSchema
 meta:
-  name: system-top
+  name: exempt
 spec:
   matchingPrecedence: 1500
   priorityLevelConfiguration:
     name: exempt
   rules:
-  - rule:
+  - rule: # All resource requests from system-privileged group
       verbs: ['*']
       apiGroups: ['*']
       resources: ['*']
     subjects:
     - kind: Group
       name: "system:masters"
-  - rule:
+  - rule: # Delegated auth[n/z] requests
       verbs: ['*']
       apiGroups: ['authentication.k8s.io', 'authorization.k8s.io']
       resources: ['*']
@@ -813,28 +815,41 @@ spec:
 ```yaml
 kind: FlowSchema
 meta:
-  name: system-high
+  name: nodes
 spec:
   matchingPrecedence: 2500
   priorityLevelConfiguration:
-    name: system-high
+    name: system
   distinguisherMethod:
     type: ByUser
   rules:  
-  - rule:
+  - rule: # All resource requests from node group
       verbs: ['*']
       apiGroups: ['*']
       resources: ['*']
     subjects:
     - kind: Group
       name: "system:nodes"
-  - rule:
+  - rule: # All non-resource requests from node group
       verbs: ['*']
       nonResourceURLs: ['*']
     subjects:
     - kind: Group
       name: "system:nodes"    
-  - rule:
+```
+
+```yaml
+kind: FlowSchema
+meta:
+  name: leader-election
+spec:
+  matchingPrecedence: 2500
+  priorityLevelConfiguration:
+    name: system
+  distinguisherMethod:
+    type: ByUser
+  rules:  
+  - rule: # Leader-election requests from kube-controller-manager/kube-scheduler
       verbs: ['*']
       apiGroups: ['coordination.k8s.io']
       resources: ['leases']
@@ -848,29 +863,207 @@ spec:
 ```yaml
 kind: FlowSchema
 meta:
-  name: system-low
+  name: kube-manager-controller
 spec:
   matchingPrecedence: 3500
   priorityLevelConfiguration:
-    name: system-low
+    name: workload-high
   distinguisherMethod:
-    type: ByUser
+    type: ByNamespace
   rules:  
-  - rule:
+  - rule: # All resource requests from kube-controller-manager
       verbs: ['*']
       apiGroups: ['*']
       resources: ['*']
     subjects:
     - kind: User
       name: "system:kube-controller-manager"    
-    - kind: User
-      name: "system:kube-scheduler"    
-  - rule:
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: attachdetach-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: certificate-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: clusterrole-aggregation-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: cronjob-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: daemon-set-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: deployment-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: disruption-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: endpoint-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: expand-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: generic-garbage-collector
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: horizontal-pod-autoscaler
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: job-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: namespace-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: node-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: persistent-volume-binder
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pod-garbage-collector
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pv-protection-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pvc-protection-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: replicaset-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: replication-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: resourcequota-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: route-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: service-account-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: service-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: statefulset-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: ttl-controller
+  - rule: # All non-resource requests from kube-controller-manager
       verbs: ['*']
       nonResourceURLs: ['*']
     subjects:
     - kind: User
       name: "system:kube-controller-manager"    
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: attachdetach-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: certificate-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: clusterrole-aggregation-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: cronjob-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: daemon-set-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: deployment-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: disruption-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: endpoint-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: expand-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: generic-garbage-collector
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: horizontal-pod-autoscaler
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: job-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: namespace-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: node-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: persistent-volume-binder
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pod-garbage-collector
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pv-protection-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: pvc-protection-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: replicaset-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: replication-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: resourcequota-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: route-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: service-account-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: service-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: statefulset-controller
+    - kind: ServiceAccount 
+      namespace: kube-system
+      name: ttl-controller      
+```
+
+```yaml
+kind: FlowSchema
+meta:
+  name: kube-scheduler
+spec:
+  matchingPrecedence: 3500
+  priorityLevelConfiguration:
+    name: workload-high
+  distinguisherMethod:
+    type: ByNamespace
+  rules:  
+  - rule: # All resource requests from kube-scheduler
+      verbs: ['*']
+      apiGroups: ['*']
+      resources: ['*']
+    subjects:
+    - kind: User
+      name: "system:kube-scheduler"    
+  - rule: # All non-resource requests from kube-scheduler
+      verbs: ['*']
+      nonResourceURLs: ['*']
+    subjects:
     - kind: User
       name: "system:kube-scheduler"         
 ```
@@ -878,22 +1071,22 @@ spec:
 ```yaml
 kind: FlowSchema
 meta:
-  name: workload
+  name: serviceaccounts
 spec:
   matchingPrecedence: 7500
   priorityLevelConfiguration:
-    name: workload
+    name: workload-low
   distinguisherMethod:
     type: ByNamespace
   rules:  
-  - rule:
+  - rule: # All in-cluster resource requests from service-account tokens
       verbs: ['*']
       apiGroups: ['*']
       resources: ['*']
     subjects:
     - kind: Group
       name: "system:serviceaccounts"
-  - rule:
+  - rule: # All in-cluster non-requests from service-account tokens
       verbs: ['*']
       nonResourceURLs: ['*']
     subjects:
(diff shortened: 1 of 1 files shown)


yue9944882:
+@MikeSpreitzer @aaron-prindle

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,49 +994,18 @@ of this subsystem.  To accomplish these things there are two levels of
...
-objects exist, and it goes as follows.  Whenever there is no actual
-RequestPriority object with `exempt == true`, the RequestPriority
-objects exhibited above are created --- except those with a name
-already in use by an existing RequestPriority object.  Whenever there
-is no actual FlowSchema object that refers to an exempt
-RequestPriority object, the schema objects shown above as examples are
-generated --- except those with a name already in use.
+The effective configuration is the union of (a) preserving configuration 
+objects and (b) all the exhibited default configurations above. The 
+system uses the preserving configuration as its initial state on
```
MikeSpreitzer: "The system" is a bit vague and probably not technically correct.  I suspect what you intend here is "the filter in an apiserver".

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,49 +994,18 @@ of this subsystem.  To accomplish these things there are two levels of
...
-RequestPriority object with `exempt == true`, the RequestPriority
-objects exhibited above are created --- except those with a name
-already in use by an existing RequestPriority object.  Whenever there
-is no actual FlowSchema object that refers to an exempt
-RequestPriority object, the schema objects shown above as examples are
-generated --- except those with a name already in use.
+The effective configuration is the union of (a) preserving configuration 
+objects and (b) all the exhibited default configurations above. The 
+system uses the preserving configuration as its initial state on 
+launching when not all the default configuration present in the system
```
MikeSpreitzer: Here I think "the system" is intended to refer to persistently stored API objects (i.e., today, in etcd).  If that is right then there is a gap here, we also need to say what happens between the time an apiserver starts handling requests and when it has had a chance to read the persistently stored API objects.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,49 +994,18 @@ of this subsystem.  To accomplish these things there are two levels of
...
-objects exhibited above are created --- except those with a name
-already in use by an existing RequestPriority object.  Whenever there
-is no actual FlowSchema object that refers to an exempt
-RequestPriority object, the schema objects shown above as examples are
-generated --- except those with a name already in use.
+The effective configuration is the union of (a) preserving configuration 
+objects and (b) all the exhibited default configurations above. The 
+system uses the preserving configuration as its initial state on 
+launching when not all the default configuration present in the system 
+and won't reload until ensuring/observing the preserving objects exists.
```
MikeSpreitzer: "reload" refers to an implementation detail, I think instead we should speak here in terms of the delivered behavior.  See my example in Slack of language that is in terms of the delivered behavior rather than the implementation.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,49 +994,18 @@ of this subsystem.  To accomplish these things there are two levels of
...
-generated --- except those with a name already in use.
+The effective configuration is the union of (a) preserving configuration 
+objects and (b) all the exhibited default configurations above. The 
+system uses the preserving configuration as its initial state on 
+launching when not all the default configuration present in the system 
+and won't reload until ensuring/observing the preserving objects exists. 
+The default configurations will always be created everytime an 
+kube-apiserver restarts.
+
+Additionally, if an administrator disagrees with the default settings and
```
MikeSpreitzer: This sentence is grammatically incomplete.  It says "if X" but there is no "then" clause.
I suggest rewriting to be something like "Administrators who wish to operate with a different configuration can create, update, and delete additional configuration objects and edit the predefined ones; the predefined ones can not be deleted but they can be modified to minimize their effects.  In particular, a FlowSchema can be modified to match no requests and a PriorityLevelConfiguration can be given AssuredConcurrencyShares=1 ".  You might try the exercise of writing a FlowSchema that is guaranteed to match no requests.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,18 +710,27 @@ queue’s virtual start time is decremented by G - S.
...
 kind: RequestPriority
 meta:
-  name: system-top
+  name: exempt
 spec:
   exempt: true
 ```
 
-For system self-maintenance requests.
+For kubernetes cluster self-maintaince, including requests from:
```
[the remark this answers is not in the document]
yue9944882: updating to node group, i.e kubelet, kube-proxy.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,18 +710,27 @@ queue’s virtual start time is decremented by G - S.
...
+
+1. every requests from node group, i.e kubelet, kube-proxy. 
+2. controller/scheduler leases
+
```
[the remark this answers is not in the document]
yue9944882: the leader-election is all moving leases api, https://github.com/kubernetes/kubernetes/issues/80289 it doesn't conflict with that. i can put my bandwidth to help moving this into 1.16 milestone

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: > I suspect what you intend here is "the filter in an apiserver".

yes, how about the "the request-management system"

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: updating

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,18 +710,27 @@ queue’s virtual start time is decremented by G - S.
 ### Example Configuration
 
 
-For requests from admins and requests in service of other, potentially
-system, requests.
+For kube-apiserver self-maintaince, solving priority-inversion, including requests from: 
+
+1. kubectl admins and requests in service of other
```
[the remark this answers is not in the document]
yue9944882: updating to ""kubectl requests from admins"

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: updated and added example for an unreachable flow-schema

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: > "the filter in an apiserver".

users are not suppose to know about the inner implementation of apiserver, so the term "filter" should not present in the doc i suppose..

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: updated

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,18 +710,27 @@ queue’s virtual start time is decremented by G - S.
...
-  name: system-top
+  name: exempt
 spec:
   exempt: true
 ```
 
-For system self-maintenance requests.
+For kubernetes cluster self-maintaince, including requests from:
+
+1. every requests from node group, i.e kubelet, kube-proxy.
```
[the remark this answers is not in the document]
yue9944882: removed

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -745,11 +756,12 @@ spec:
   queueLengthLimit: 1000
 ```
 
-For user requests from kubectl.
+For normal workloads:
```
MikeSpreitzer: The original intent here was to distinguish user requests issued using kubectl from all other normal workload requests.  This revision replaces (A) "user requests from kubectl" vs "requests from controllers processing workload" with (B) "normal workloads" vs "requests from controllers processing workload".  I think that "requests from controllers processing workload" is part of "normal workloads", so that makes the wording for (B) problematic.  Also, this revision replaces two names that include "workload" with one name that is simply "workload" and another that does not include "workload" but is associated with a meaning that involves "workload".  If we are going to insist on one name that does not include the word "workload" but has a meaning that *does* involve workload then I suggest the other name be one that identifies its distinction (e.g., "workload-controllers").  This revision also moves much of the workload to "system-low"; in the original, "system-low" was deliberately very restricted, to just the garbage collector, and the other controllers were assigned to "workload-low".

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -777,127 +788,147 @@ kind: FlowSchema
...
+      name: "system:kube-controller-manager"    
+    - kind: User
+      name: "system:kube-scheduler"         
 ```
 
 ```yaml
 kind: FlowSchema
 meta:
-  name: workload-high
+  name: workload
```
MikeSpreitzer: The original design was that the workload consists of two parts: kubectl requests from users and requests from controller.  The original design prioritized kubectl requests higher than controller requests.  The revised design prioritizes the controller requests higher (in "system-low"); I think this is not an improvement.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -777,127 +788,147 @@ kind: FlowSchema
...
-      field: resource
-      set: [ "tokenreviews", "subjectaccessreviews" ]
-    - superSet:
-      field: user
-      set: [ "system:serviceaccount:example-com:network-apiserver" ]
+  matchingPrecedence: 10000
+  priorityLevelConfiguration:
+    name: default
+  distinguisherMethod:
+    type: ByUser
```
MikeSpreitzer: In the original this was by object namespace rather than username.  Why the change?

MikeSpreitzer:
@MikeSpreitzer

MikeSpreitzer:
The KEP should clearly identify that there are **two** "default" behavior problems, or perhaps better described as initial behavior or initialization problems.  One is initializing a cluster.  The other is how the filter behaves before the persisted configuration objects are read.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -777,127 +788,147 @@ kind: FlowSchema
 meta:
   name: system-top
 spec:
-  requestPriority:
-    name: system-top
-  match:
-  - and: # writes by admins (does this cover loopback too?)
```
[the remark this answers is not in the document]
yue9944882: added comments

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,18 +710,27 @@ queue’s virtual start time is decremented by G - S.
 ### Example Configuration
 
 
-For requests from admins and requests in service of other, potentially
-system, requests.
+For kube-apiserver self-maintaining, solving priority-inversion, including requests from:
```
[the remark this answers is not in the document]
yue9944882: updated

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: oversight. fixed

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: i think it's better we keep it simple in the alpha stage. users can customize their own settings theirself?

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: i renamed all default objects. now they're:

1.  FS `exempt` => PL `exempt`
2. FS `nodes`, `leader-election` => PL `system-high`
3. FS `kube-controller-manager`, `kube-scheduler` => PL `system-low`
4. FS `serviceaccounts` => PL `workload`
5. FS `default` => PL `default`

is it cleaner ?

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,50 +994,43 @@ of this subsystem.  To accomplish these things there are two levels of
...
+  priorityLevelConfiguration:
+    name: default
+  rules:  
+  - rule:
+      verbs: ['*']
+      apiGroups: ['*']
+      resources: ['*']
+    subjects:
+    - kind: User
+      name: "" # unreachable username
```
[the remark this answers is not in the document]
yue9944882: indeed, i updated example matching-nothing flow-schema

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
+    name: exempt
+  rules:
+  - rule: # All resource requests from system-privileged group
+      verbs: ['*']
+      apiGroups: ['*']
+      resources: ['*']
+    subjects:
+    - kind: Group
+      name: "system:masters"
+  - rule: # Delegated auth[n/z] requests
```
MikeSpreitzer: I think the matching condition here could be tightened up.  Have a look at the example data at https://github.com/kubernetes/kubernetes/pull/81224#issuecomment-521835562

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
-    - inSet:
-      field: resource
-      set: [ "endpoints", "configmaps", "leases" ]
-    - equals:
-      field: namespace
-      value: kube-system
+  distinguisherMethod:
+    type: ByUser
+  rules:  
+  - rule: # All resource requests from node group
```
MikeSpreitzer: If you look at what the KEP says right now, it calls for "heartbeats by nodes" and "kubelet and kube-proxy ops _on system objects_".  That qualifier "on system objects" was deliberate.  The idea was to give higher priority to maintenance of system things (e.g., pods, services) than for regular things.  But since our current models can not discriminate on object namespace, that qualifier can not be expressed.  That is one of the reasons for https://github.com/kubernetes/kubernetes/pull/81637 .

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
+  matchingPrecedence: 2500
+  priorityLevelConfiguration:
+    name: system-high
+  distinguisherMethod:
+    type: ByUser
+  rules:  
+  - rule: # Leader-election requests from kube-controller-manager/kube-scheduler
+      verbs: ['*']
+      apiGroups: ['coordination.k8s.io']
+      resources: ['leases']
```
MikeSpreitzer: Leader elections are not done this way yet.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
+      name: "system:kube-controller-manager"    
+    - kind: User
+      name: "system:kube-scheduler" 
 ```
 
 ```yaml
 kind: FlowSchema
 meta:
-  name: workload-high
+  name: kube-manager-controller
```
MikeSpreitzer: This name covers only part of what is matched.  Did you mean to leave the scheduler in here?  It is also has a flow schema dedicated to it.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
+  distinguisherMethod:
+    type: ByUser
+  rules:  
+  - rule: # All resource requests from kube-controller-manager
+      verbs: ['*']
+      apiGroups: ['*']
+      resources: ['*']
+    subjects:
+    - kind: User
+      name: "system:kube-controller-manager"
```
MikeSpreitzer: The kube-controller-manager has two modes.  In one mode, all the controllers appear as the same user.  In the other mode, each controller appears as a distinct user.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -775,129 +788,188 @@ Some flow schemata.
...
-    - notPatternMatch:
-      field: user
-      pattern: system:serviceaccount:.*
+  matchingPrecedence: 3500
+  priorityLevelConfiguration:
+    name: system-low
+  distinguisherMethod:
+    type: ByUser
+  rules:  
+  - rule: # All resource requests from kube-controller-manager
```
MikeSpreitzer: In the KEP currently, almost all of these are given much lower priority.  They are part of "workload-low", while kubectl requests from users are prioritized more highly at "workload-high".  I think moving controller requests above user kubectl requests is not an improvement.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: Did another review to respond.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: This is not really a matter of simplicity.  We already made an estimate of what we think will be a good configuration.  I do not think it is an improvement to switch to a less good configuration just because it can be overridden.

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: ouch removed

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: i think it's difficult to wirte a rule for "kubectl requests".. kubectl is just a tooling, whether use kubectl or not isn't related with the user identity..

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: added service-account rules for matching controller-manager

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: and i renamed `system-low` to `workload-high`

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: no, ` ['authentication.k8s.io', 'authorization.k8s.io']` is sufficient

MikeSpreitzer: Do you mean that no other requests involve those API groups?  I doubt it.  I am suggesting that we be more discriminating here, giving this exceptionally high priority only to requests that really need it, as best we can distinguish.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -712,44 +710,60 @@ queue’s virtual start time is decremented by G - S.
...
 
 
-For requests from admins and requests in service of other, potentially
-system, requests.
+For kube-apiserver self-maintaining, solving priority-inversion, including requests 
+from: 
+
+1. requests from system-privileged group, including kubectl requests from admin and 
+apiserver loopback requests.
+2. delegated authentication/authorization API groups, including delegated delegated
```
MikeSpreitzer: typo here: "delegated" is repeated.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: It is true that kubectl is not directly identified in the request attributes but the complement is nearly true.  After logically higher precedence levels have identified other stuff, what is left is workload.  That mainly consists of requests from controllers --- running in-cluster using a service account --- and requests from users using kubectl.  So the latter can be identified by subtracting out the former.  That is the approach you have been seeing from me since I had to start coping with the lack of negation in the matching language.

keps/sig-api-machinery/20190228-priority-and-fairness.md
yue9944882: yes, the group is dedicated for delegated authn/authz

MikeSpreitzer: Oh, I see.  OK then.  We could still tighten up a bit, by requiring that the user be a service account --- but that still leaves a lot of access to this high priority.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: Good, that was an improvement.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,50 +994,43 @@ of this subsystem.  To accomplish these things there are two levels of
...
+Additionally, if an administrator disagrees with the default settings and
+feel like replacing with a complete new set. To avoid re-creation of the  
+system defaults above, the administrator is expected to update the objects
+to mute those configurations doesn't meet expectation.
+Administrators who wish to operate with a different configuration can create, 
+update, and delete additional configuration objects and edit the predefined 
+ones; the predefined ones can not be deleted but they can be modified to 
+minimize their effects. In particular, to mute a FlowSchema, you can modify 
+to match no requests. To mute a PriorityLevelConfiguration modify 
+AssuredConcurrencyShares=1. For example:
```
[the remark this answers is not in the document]
MikeSpreitzer: Here is a reworded replacement for the paragraph that starts with the incomplete sentence, not referring to "the filter".

>  Administrators can choose to operate with a partially or completely different set of persistent configuration API objects than the default one; note that this does not affect the behavior of the apiserver before it has had a chance to read the persisted configuration API objects. Additional objects can be created, updated, and deleted. The predefined objects can be updated. The predefined objects can not be deleted; administrators can get almost the same effects by updating the predefined objects to minimize their effects. A PriorityLevelConfiguration can be set to have just 1 assured concurrency share; care should be taken to do this only after its queues have drained, otherwise this setting may greatly slow down the rate at which the queues drain. A FlowSchema can be updated to match no requests; following is an example.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -963,50 +1187,44 @@ of this subsystem.  To accomplish these things there are two levels of
...
+  priorityLevelConfiguration:
+    name: default
+  rules:  
+  - rule:
+      verbs: ['*']
+      apiGroups: ['*']
+      resources: ['*']
+    subjects:
+    - kind: User
+      name: "null" # or assign it any unreachable username
```
MikeSpreitzer: Are there any invalid usernames?
Looking at the validation now, I see gaps in validation of AIPGroup, Resources, and NonResourceURLs.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: Sorry, that was not quite right.  Leader elections are not all necessarily done that way now.  https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/interface.go::New still supports configmaps and endpoints.

MikeSpreitzer:
I think we have agreed to split this into two PRs; see #1221 for my attempt to draft one of those two parts.

PR #1221, opened 2019-08-23
Updated API priority and fairness KEP wrt default behaviors

MikeSpreitzer:
.. to match the latest implementation thinking.

### DIFF
--- a/keps/sig-api-machinery/20190228-priority-and-fairness.md
+++ b/keps/sig-api-machinery/20190228-priority-and-fairness.md
@@ -47,6 +47,9 @@ superseded-by:
   - [Example Configuration](#example-configuration)
   - [Reaction to Configuration Changes](#reaction-to-configuration-changes)
   - [Default Behavior](#default-behavior)
+    - [Default Cluster Configuration](#default-cluster-configuration)
+    - [API Server Initial Behavior](#api-server-initial-behavior)
+    - [Default Request Handling](#default-request-handling)
   - [Prometheus Metrics](#prometheus-metrics)
   - [Testing](#testing)
   - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
@@ -717,7 +720,7 @@ system, requests.
 ```yaml
 kind: RequestPriority
 meta:
-  name: system-top
+  name: exempt
 spec:
   exempt: true
 ```
@@ -757,11 +760,12 @@ spec:
   queueLengthLimit: 100
 ```
 
-For requests from controllers processing workload.
+For requests from controllers processing workload, and anything else
+that does not go in a logically higher priority.
 ```yaml
 kind: RequestPriority
 meta:
-  name: workload-low
+  name: default
 spec:
   catchAll: true
   assuredConcurrencyShares: 100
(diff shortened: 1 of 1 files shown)


MikeSpreitzer:
@yue9944882 and I have agreed to split #1214 in two; this is my attempt to create first first half.

MikeSpreitzer:
To be clear, I do not think the behavior described here is our best choice.  I think what is described in the KEP as updated by #1200 is better, but have not been able to get agreement on that.  I wrote this PR in hopes of getting some agreement.

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -957,55 +961,89 @@ complete their service.
...
+is enough; the apiserver does not insist that the spec matches its
+predefined content.
+
+Administrators can choose to persist a partially or almost completely
+different set of configuration API objects than the predefined set
+(but remember that the behavior of the apiserver before it has had a
+chance to read the persisted configuration API objects is a different
+matter). Additional objects can be created, updated, and deleted. The
+predefined objects can be updated and even deleted --- but they will
+re-appear at the next kube-apiserver startup.  Thus, administrators
```
yliaog: what is the reason to allow them to re-appear? that is 1) counter intuitive 2) requires more work from admin/operator

MikeSpreitzer:
@MikeSpreitzer

keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -957,55 +961,89 @@ complete their service.
...
+had a chance to read all the persisted configuration API objects.
+
+After the initial behavior period is over, regular behavior begins.
+
+#### Default Request Handling
+
+It is possible that the administrators can modify the configuration
+API objects to create a situation where some requests will not match
+any FlowSchema.  Once the initial behavior period is over, any request
+that does not match a FlowSchema object is rejected.
```
yliaog: what about having a default/catch-all for all those requests that don't match any flow schema, instead of have them all 'rejected', put them all into a low priority queue.

the reasoning is there could be new request / new apis added over time, and the current behavior (max-in-flight limit) is not rejecting them

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: I agree.  The KEP specifies that predefined objects get created only if there are not already objects that look adequate.

keps/sig-api-machinery/20190228-priority-and-fairness.md
MikeSpreitzer: I agree that we should do better.  The KEP stipulates similar handling for this case.  #81003 implements it.

MikeSpreitzer: .. and it is implemented in https://github.com/kubernetes/kubernetes/pull/82196

keps/sig-api-machinery/20190228-priority-and-fairness.md
lavalamp: I commented there instead of here.

PR #1224, opened 2019-08-29
Add exec as user KEP

max88991:
Signed-off-by: Lei Gong <lgong@alauda.io>

### DIFF
--- a/keps/sig-node/20190825-exec-as-specified-user.md
+++ b/keps/sig-node/20190825-exec-as-specified-user.md
@@ -89,7 +89,7 @@ By providing the `user` option for `exec`, users can use their own ID to enter t
 Since the container does not have a uniform `exec` format, the format of `user` should be just a string.
 
 #### Story 2
-By providing the `user` option for `exec`, user can do the auit inside the container.
+By providing the `user` option for `exec`, user can do the audit inside the container.
 
 ## Design Details
 
--- a/keps/sig-node/20190825-exec-as-specified-user.md
+++ b/keps/sig-node/20190825-exec-as-specified-user.md
@@ -1,0 +1,129 @@
+---
+title: Exec as specified user
+authors:
+  - "@max8899"
+owning-sig: sig-node
+participating-sigs:
+  - sig-cli
+  - sig-auth
+reviewers:
+  - "@smarterclayton"
+  - "@dchen1107"
+approvers:
+  - TBD
+editor: TBD
+creation-date: 2019-08-25
+last-updated: 2019-08-25
+status: implemented
+see-also:
+  - "https://github.com/kubernetes/kubernetes/issues/30656"
+  - "https://github.com/kubernetes/kubernetes/pull/81883"
+replaces:
+  - "n/a"
+superseded-by:
+  - "n/a"
+---
+
+# Exec as specified user
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+- [Design Details](#design-details)
+  - [API Specification](#api-specification)
+  - [Library Specification](#library-specification)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+As a Kubernetes User, we should be able to specify username for the containers when doing exec, similar to how docker allows that using docker exec options `-u`, 
+```
+-u, --user="" Username or UID (format: <name|uid>[:<group|gid>]) format
+```
+
+## Motivation
+
+The root problem is lacking of standard description for the `exec` operation.
+
+OCI didn't introduce `exec` as a regular operation for container runtime[2],
+
+so when Kubernetes announce [Container Runtime Interface(CRI)](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) in release 1.5, it became the de-facto standard of `exec`, without a `user` option.
+
+
+>  In the Kubernetes 1.5 release, we are proud to introduce the Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes, without the need to recompile. CRI consists of a protocol buffers and gRPC API, and libraries, with additional specifications and tools under active development. CRI is being released as Alpha in Kubernetes 1.5.
+
+But There are many container runtime implementation, e.g Docker, Kata, gVisior, most of which provide a way to enter the container as a special `user`.
+
+It is because the `exec` is basically to start a process inside the container runtime, and the [`process`](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L39) has a user definition, which is required.
+
+> User specifies specific user (and group) information for the container process
+
+It brings convenience to debugging, and users are used to it. Adding a `user` option with `exec`  in Kubernetes seems reasonable.
+
+### Goals
+
+1. Provide the ability to specify the username for a exec operation inside a container
+
+## Proposal
+
+### User Stories [optional]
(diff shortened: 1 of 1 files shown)


max88991:
@mattjmcnaughton @smarterclayton please take a look at this

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Exec as specified user
+authors:
+  - "@max8899"
+owning-sig: sig-node
+participating-sigs:
+  - sig-cli
+  - sig-auth
+reviewers:
+  - "@smarterclayton"
```
mattjmcnaughton: Could we please add a reviewer from @kubernetes/sig-auth-api-reviews ?

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+The root problem is lacking of standard description for the `exec` operation.
+
+OCI didn't introduce `exec` as a regular operation for container runtime[2],
+
+so when Kubernetes announce [Container Runtime Interface(CRI)](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) in release 1.5, it became the de-facto standard of `exec`, without a `user` option.
+
+
+>  In the Kubernetes 1.5 release, we are proud to introduce the Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes, without the need to recompile. CRI consists of a protocol buffers and gRPC API, and libraries, with additional specifications and tools under active development. CRI is being released as Alpha in Kubernetes 1.5.
+
+But There are many container runtime implementation, e.g Docker, Kata, gVisior, most of which provide a way to enter the container as a special `user`.
```
mattjmcnaughton: To be sure I'm understanding, are you saying that this `exec as user` feature will only work for certain container runtime implementations (i.e. Docker, Kata...)?

That's interesting to me - my sense is that k8s is trying to move towards less special casing of certain container runtime implementations.

I'd be interested to hear from others in the community about the risk of increasing featuring between different container runtime implementations.

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+
+It is because the `exec` is basically to start a process inside the container runtime, and the [`process`](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L39) has a user definition, which is required.
+
+> User specifies specific user (and group) information for the container process
+
+It brings convenience to debugging, and users are used to it. Adding a `user` option with `exec`  in Kubernetes seems reasonable.
```
mattjmcnaughton: Can you share a little bit more about the debugging experience?

What are the specific types of problems this would allow you to debug that you couldn't debug before?

I'm also curious about if some of the ideas proposed in the [Ephemeral Containers KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190212-ephemeral-containers.md) could be helpful w.r.t. these types of debugging (cc @verb ).

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+By providing the `user` option for `exec`, users can use their own ID to enter the container and exercise their respective rights.
+
+```
+-u, --user="" Username or UID (format: <string>]) format
+```
+
+Since the container does not have a uniform `exec` format, the format of `user` should be just a string.
+
+#### Story 2
+By providing the `user` option for `exec`, user can do the auit inside the container.
```
mattjmcnaughton: Apologies for not knowing this :) but what is `auit`?

keps/sig-node/20190825-exec-as-specified-user.md
verb: Currently we don't allow changing user in an ephemeral container, so what you get is what is in the dockerfile of the ephemeral container image. Likely some of the use cases here could be solved with an ephemeral debugging container (for example troubleshooting as root when a container image specifies a non-root user), but I see value in both approaches. They would be complementary.

I, too, would like to hear more about the debugging experience and agree that adding examples of things one cannot currently accomplish would help quite a bit.

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Design Details
```
smarterclayton: There’s a lot of additional security protection which is required to make arbitrary exec safe, especially by username (which is only mappable inside a container).

Anyone using PSP or similar systems would need a way to control which users are allowed to exec.

The use of username is contrary to how kubernetes pods run so it needs substantially more justification.

The design would also have to take into account how windows exec would support exec, and how runtimes that don’t support exec changes would communicate that to users.

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+OCI didn't introduce `exec` as a regular operation for container runtime[2],
+
+so when Kubernetes announce [Container Runtime Interface(CRI)](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) in release 1.5, it became the de-facto standard of `exec`, without a `user` option.
+
+
+>  In the Kubernetes 1.5 release, we are proud to introduce the Container Runtime Interface (CRI) – a plugin interface which enables kubelet to use a wide variety of container runtimes, without the need to recompile. CRI consists of a protocol buffers and gRPC API, and libraries, with additional specifications and tools under active development. CRI is being released as Alpha in Kubernetes 1.5.
+
+But There are many container runtime implementation, e.g Docker, Kata, gVisior, most of which provide a way to enter the container as a special `user`.
+
+It is because the `exec` is basically to start a process inside the container runtime, and the [`process`](https://github.com/opencontainers/runtime-spec/blob/master/specs-go/config.go#L39) has a user definition, which is required.
```
php-coder: I'd suggest to use permanent link instead:
```suggestion
It is because the `exec` is basically to start a process inside the container runtime, and the [`process`](https://github.com/opencontainers/runtime-spec/blob/52e2591aa9f7211d64c49c4fed8691a183189284/specs-go/config.go#L39) has a user definition, which is required.
```

keps/sig-node/20190825-exec-as-specified-user.md
php-coder:
```suggestion
By providing the `user` option for `exec`, user can do the audit inside the container.
```

php-coder: It's a typo :)

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Goals
+
+1. Provide the ability to specify the username for a exec operation inside a container
+
+## Proposal
+
+### User Stories [optional]
+
+As a Kubernetes User, I should be able to control how people can enter the container, by default kubernetes use the default user in docker image.
```
yujuhong: This doesn't control how people access the containers. They'll have to specify user name when exec'ing, or are you suggesting using PSP to further restrict users?

keps/sig-node/20190825-exec-as-specified-user.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Story 1
+By providing the `user` option for `exec`, users can use their own ID to enter the container and exercise their respective rights.
+
+```
+-u, --user="" Username or UID (format: <string>]) format
+```
+
+Since the container does not have a uniform `exec` format, the format of `user` should be just a string.
```
yujuhong: Windows does not support uid. For Linux,  kubelet uses uid when starting the containers. 
The current design may be confusing for cross-platform support.

zqzten:
@mikebrow There's a new on-going KEP: #3372. I've been watching on that but it also seems to be inactive for some while.

PR #1225, opened 2019-08-29
Add rationale for priority and fairness example config

MikeSpreitzer:


keps/sig-api-machinery/20190228-priority-and-fairness.md
```diff
@@ -726,6 +728,60 @@ queue’s virtual start time is decremented by G - S.
...
+- Next comes a priority level for the generic object garbage
+  collector.  This serves both system self-maintenance and ordinary
+  workload --- in an integrated way, so it does not make sense to try
+  to differentiate its traffic.  Instead we prioritize it between the
+  two.  We expect the higher priority levels can tolerate garbage
+  collector latency, since there is no promise about that.
+
+At this point we have covered the "system" traffic, and what is left
+is ordinary workload.  We divide it into two priority levels as
+follows.
```
yliaog: are requests from scheduler "system" traffic? it seems to belong to the 2nd system self-maintenance group, but it is explicitly listed (only three are listed there)

MikeSpreitzer: I think you mean it is _not_ listed; I see it absent in both the exhibited example objects and the design rationale discussion.  And yes, that looks like an oversight.  Requests from the scheduler _about system objects_ should be part of the system self-maintenance priority level.

PR #1231, opened 2019-09-03
kustomize reorder plugin

jbrette:


keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+reviewers:
+ - "@monopole"
+ - "@Liujingfang1"
+approvers:
+ - "@monopole"
+editors:
+ - "@jbrette"
+creation-date: 2019-09-01
+last-updated: 2019-09-01
+status: proof of concept
```
Liujingfang1: Change the status to `provisional`. You can take a look at the KEP template [here](https://github.com/kubernetes/enhancements/blob/master/keps/YYYYMMDD-kep-template.md).

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+- Certificate
+- ClusterIssuer
+- Issuer
+```
+
+A new transformer could also be implemented to filter resources during `kubectl apply` for instance:
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: KubectlApplyFilterTransformer
```
Liujingfang1: Can you elaborate this transformer? Does it only apply the resources kind being listed?
What's special for `filter`?

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+- Ensure dependencies between Workloads
+
+## Proposal
+
+Provide a simple mechanism allowing users to override the order that
+Resource operations are applied or deleted. Leverage the plugins feature of kustomize
+in order to let the user override the default sorting plugin.  
+
+```bash
+kustomize build <xxx> --reorder=kubectlapply | kubectl apply -f -
```
Liujingfang1: How do user customize the kind order when using the `--reorder` flag?

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+### kubectl delete example:
+
+The `--reorder=kubectctldelete` option indicates to kustomize to load a `kubectlapplyordertransformer.yaml transformer config`,
+itself loading a kustomize external plugin in charge of sorting and filtering kustomize build output.
+
+An example of configuration of such a transformer.
+
+```yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: KubectlDeleteOrderTransformer
```
Liujingfang1: The `KubectlDeleteOrderTransformer` looks quite similar to `KubectlApplyOrderTransformer`. Why do you need both? Can you combine them into one?

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,217 @@
...
+### Risks and Mitigations
+
+Risk: Current external plugins require the presence of the module in the XDG_CONFIG_HOME.
+Mitigation: Good documentation and recommendations about how to keep things simple.
+
+## Graduation Criteria
+
+Use customer feedback to determine if we should support:
+
+- Let the user specify the transformers configuration file in the "top" kustomization.yaml.
```
Liujingfang1: Assume there are base and overly. Both the base and overlay contain its own order transformers. Do the resources from the base follow the order customized in base? Does the overlay order overwrite all the base order?

keps/sig-cli/kustomize-reorder-plugins.md
Liujingfang1: Are you proposing to use the custom transformer (the order tansformer) with a builtin flag?

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: done

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: done.

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: done

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: done

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: done

keps/sig-cli/kustomize-reorder-plugins.md
jbrette: @Liujingfang1 This is more like a hook. In the current POC, only the names "kubectlapplyordertransformer.yaml" and "kubectldeleteordertransformer.yaml" are hardcoded. Just before outputing the file, kustomize loads the plugin/transformer sort and filter the resmap.

jbrette:
@Liujingfang1 Thank you for reviewing. I think I integrated your remarks.

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,261 @@
...
+kustomize build <xxx>
+```
+
+The user can also choose to preserve the resources in the order in which they were loaded:
+
+```bash
+kustomize build <xxx> --reorder=none
+```
+
+The new --reorder=kubectlappy and --reorder=kubectldelete options extend what was done
```
EronWright: `kubectlappy` -> `kubectlapply`

keps/sig-cli/kustomize-reorder-plugins.md
```diff
@@ -0,0 +1,261 @@
...
+
+```bash
+kubectl apply -k <xxx>
+kubectl delete -k <xxx>
+```
+
+
+### kubectl apply example:
+
+The `--reorder=kubectctlapply` option indicates to kustomize to load the `kubectlapplyordertransformer.yaml transformer config` ,
```
EronWright: The backticks seem misplaced here.

EronWright:
The existence both `kubectctlapply` and `kubectctldelete` is off-putting to be sure.   Frankly I'm surprised that `kubectl delete -f` doesn't naturally use a bottom-to-top ordering to better compliment `kubectl apply -f` (considering that the traditional approach is to bake the ordering into your filenames `00_`, `01_`).   Are we sure that the ordering cannot be invariant to the action?

monopole:
thanks @jbrette 

As you note, the "legacy" sorting we do is already performed by a plugin (the  [legacyordertransformer plugin](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/LegacyOrderTransformer.go)) but there's no way to change that plugin's config, since it gets it's sort order from constant defined in the `gvk` package.

We should change that.  Continue to use this plugin, but rename it to `OrderTransformer`, and allow it to get its configuration from a file that [replicates what the hard-coding is doing](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/gvk/gvk.go#L82), e.g.

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: OrderTransformer
metadata:
  name: orderTransformer
first:
- Namespace
- ResourceQuota
- StorageClass
- CustomResourceDefinition
- MutatingWebhookConfiguration
- ServiceAccount
- PodSecurityPolicy
- Role
- ClusterRole
- RoleBinding
- ClusterRoleBinding
- ConfigMap
- Secret
- Service
- LimitRange
- PriorityClass
- Deployment
- StatefulSet
- CronJob
- PodDisruptionBudget
last:
- ValidatingWebhookConfiguration
```
This approach means we don't need to exhaustively list all resources - just those things we want first and last.  The rest  pops out in the middle (in alphabetical order).

The core logic can be adapted from the [Gvk code](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/gvk/gvk.go#L107).

A user that wants a different sort (than what's specified in `Gvk`) just creates a config file as desired and specifies it as their last transformer in the `transformers:` field.

Unfortunately, they'll also have to specify

```
kustomize build --reorder none
```

to suppress the legacy sort (until we change kustomize itself).

Sorting is just a transform and should not get special status on the command line.  It should be declared in the kustomization file (per _declarative application management_).  

BTW, i hope everyone knows this, but you don't _need_ a sorter if your config is self-contained.   Just specify `--reorder none` and specify items in your `resources:` field in the desired order, e.g. specify files with namespace objects first, then files with resourcequotas, etc.

I'm not sure that it should be kustomize's problem to reverse the YAML stream for the benefit of `kubectl delete`.

jbrette:

PR #1232, opened 2019-09-04
Kustomize FieldSpec skip

jbrette:


yujunz:
Could you please clarify a bit the precedence of the configurations?  @jbrette 

- default configurations
- custom field specs (without `skip`)
- custom field specs (with `skip: true`)

What is the expected behavior?

- order matters?
- `skip` always take the precedence?
- error when conflict is detected?

keps/sig-cli/kustomize-default-config.md
```diff
@@ -0,0 +1,314 @@
...
+The remaining case is linked to the handling of a none wild card entry. By default kustomize adds commonLabel of the selectors
+of development. Depending on the project, this can be an issue. The `behavior: remove` or `behavior: replace` can be used to
+changed the handlig of the field by the transformer.
+
+```yaml
+- path: spec/template/spec/affinity/podAntiAffinity/requiredDuringSchedulingIgnoredDuringExecution/labelSelector/matchLabels
+  create: false
+  group: apps
+  kind: Deployment
+  behavior: remove
```
Liujingfang1: I think `skip: true` can do the same thing.

keps/sig-cli/kustomize-default-config.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+3. Wrap FieldSpec into a FieldSpecConfig, to let the user remove and replace configuration entries embedded in the default configuration
+   provided with kustomize
+   
+```go
+type FieldSpecConfig struct {
+	FieldSpec `json:",inline,omitempty" yaml:",inline,omitempty"`
+ // Behavior legal values are "", "add", "replace", "remove"
+	Behavior  string `json:"behavior,omitempty" yaml:"behavior,omitempty"`
```
Liujingfang1: Why you need this. The `skip` field from 1 can already overwrite the default behavior.

keps/sig-cli/kustomize-default-config.md
```diff
@@ -0,0 +1,314 @@
...
+
+The propsal has three parts:
+
+1. Update the FieldSpec definition to add a "skip" field as follow (It is also useful external transformers needing wild card):
+
+```go
+type FieldSpec struct {
+	gvk.Gvk            `json:",inline,omitempty" yaml:",inline,omitempty"`
+	Path               string `json:"path,omitempty" yaml:"path,omitempty"`
+	CreateIfNotPresent bool   `json:"create,omitempty" yaml:"create,omitempty"`
```
Liujingfang1: What's the relationship between `CreateIfNotPresent` and `SkipTransformation`? Which one has higher priority?

keps/sig-cli/kustomize-default-config.md
```diff
@@ -0,0 +1,314 @@
...
+
+- Provide the ability for the user to skip transformation for specific group/version/kind especially usefull if the default
+  configuration is using a wild card (see namespace and prefixsuffix transformer configuration).
+- Provide the ability of the user to replace or remove an entry in the default configuration which is not matching his needs
+  (see commonLabel transformer configuration)
+
+  
+### Non-Goals
+
+- Provide backward compatibility with the transformer default configurations.
```
Liujingfang1: We should keep backward compatibility otherwise who customize the builtin transformer configuration will experience breaking changes.

Keep backward compatibility should be listed as a Goal

monopole:
Not a fan of adding a skip field to FieldSpec.

First, the purpose of the type FieldSpec is to specify a _field_ (not a transformation), and is used in other places besides transformer configs.  Adding a 'skip' bit there to describe how a transformation should behave is hacky.  Having `CreateIfNotPresent` in this type, rather than up in some encapsulation along with FieldSpec, was a poor choice, which we should not make it worse by adding more transformer controlling data to the type.  The KEP is right to think about putting such fields up in a wider type that includes FieldSpec and other data as fields.


Second, we can already 'skip' Kinds with no code change, by invoking a `PrefixSuffixTransformer` as a transformer plugin explicitly, and listing all the kinds it should touch, omitting the kinds it should not touch. For a given user of 20 or so k8s kinds, this isn't particularly hard.  If it were, one might consider using namespace isolation rather than the namePrefix approach.

Third, we need to think carefully about the new plugin architecture and how changes to kustomization.yaml influence it.  The kustomization file is now a means to launch plugins, and in the new world those plugins are in control of their own config.   There's an adaptive layer to map old fields (e.g. namePrefix) to plugins, but that's all it is - an adaptive layer to shove config from the old world to the new.  

It's possible that kustomization fields (other than `transformers:`, `generators:` and `resources:`) could be used to apply common configuration to all (or some) plugins, but lets first establish that that is actually a problem that cannot be solved by adding new fields to the plugin configuration directly, or by using other transformers to inject data into plugin config as a means to reduce boilerplate.

monopole:
Have you considered using a plugin to modify another plugin config as a way to build up either an inclusion or exclusion list?

jbrette:
1. I actually did consider the option of a two phases build as described [here](https://github.com/kubernetes-sigs/kustomize/pull/1485#issuecomment-526341865). Since the kustomization.yaml and the transformers.yaml can be considered as CRDs, we could apply kustomize to the kustomization.yaml and transformers.yaml first to transform the kustomization.yaml and transformers.yaml before applying them to the resources themselves. 

Projects like kubeadm and kubeflow seems to end up having to write go code to create those kustomization.yaml, basically transforming the kustomization.yaml before invoking kustomize on the actual resources.

 
2. I tried to see how a user using kustomize binary and the prefixsuffix transformer to create the k8s files for a canary upgrade:
   -  Using your example and method the [transfomer complete override](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_b) 
   - Creating a canary upgrade using composition  [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_c/using-composition)
   - Creating a canary upgrade using "the skip feature" [here](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_c/using-skip)

None of them were actual simple to use (especially dealing with the ingress object).

3. The current proposal is aiming at keeping thing simple and not having to juggle too much with kustomization folders.

4. The "transformers in the kustomization.yaml" option is the most powerful and flexible but currently comes at the cost of complexity. None of the following issues are really hard to solve, they are more  an obstacle to somebody trying to use kustomize out of the box:

    - Using "transformers:" seems to currently require to have compiles the builtin transformer modules as  .so files.
    - When you compile the transformer you also have to recompile the kustomize to ensure that all your sub-libraries have the same version. Upgrading Makefile provide a "build-plugins" target along with the "install" target could be useful
    - you still need to put the --enable-alpha-plugin option.
    => Looks like we package kustomize with the modules in it when we release instead of just the executable (docker image ?/ debian or brew package ?)

    - Where do you find the proper syntax for those configuration file for those transformers:
    => Seems to be mainly inside the _test.go.
    => We need real plain examples of configuration for each plugin : [like here for instance](https://github.com/keleustes/kustomize/tree/plugin/examples/plugin)
    - We currently have to copy/paste all the "transformer.yaml" files to each base, and overlay directory.
       nameprefix: field in the kustomization.yaml is local to that folder. Going the "transformer.yaml" path would mean copy/pasting I think the transformer.yaml in each kustomization folder/context. The transformers are not [accumulated](https://github.com/keleustes/kustomize/blob/plugin/pkg/target/kusttarget.go#L318) and I don't they could without loosing the local concept coming with the folder and the kustomization.yaml.

All those thing render the kustomization.yaml and folders really much harder to build. 

_I think that we could probably have a real balance between ease of use and flexibility provided by the plugins if `kustomize build .` was under the hood converting the "nameprefix:" entry in the kustomization.yaml into an in memory "prefixsuffixtransforrmer.yaml" , loading the transformer like any external transformer (without the need for installing it in the XDG_CONFIG) and executing without using the --enable-alpha. That procedure would be followed unless the "prefixsuffixtransformer.yaml" is provided by the user, letting him change whatever he needs too_.

5. Assuming we go for something like 4:
    - to keep the configuration of each transformer simple, we still need the "exclude|except|skip" concept of a wild card selection of a FieldSpec. 
    - we can drop the "behavior:" xxx concept.
   - we still need higher level operations to deal with collections of FieldSpec. The same way the order of the transformers have and impact on the final resources, the order of the FieldSpec has an impact too [for instance here](https://github.com/keleustes/kustomize/blob/plugin/pkg/transformers/namereference.go#L84)

monopole:
thanks, well considered.  lots to unpack here.

 * re: transforming a kustomization file
   
   Both kustomization.yaml files and plugin config files can be
   transformed by kustomize (since they are yaml files).  The former
   is a bit wild, and i've not seen anyone do it, but the latter is
   simple to incorporate into a model if you've already bought into
   the idea of how kustomize works.

   We've got one example of [kustomizing a plugin config](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/transformerplugin_test.go#L189).

   It's a bit contrived.  It uses a `DatePrefixer` transformer
   (plugin) to modify the value being applied by a `StringPrefix`
   transformer (plugin).

   If there's value to the pattern of transfomers modifying the
   configs of other transformers, i suspect it will surface in
   patching in one (common) set of `FieldSpec`s or `Gvk`s into the
   configs of different transformers, so they operate on (or exclude)
   the same set.

 * Using transformers seems to require .so files?

   Not if one uses _builtin plugins_ (which are truly compiled into
   the kustomize binary, although the source code is expressed as a
   plugin).

   [Your example](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_b) is the same as the [unit test](
https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/customconfigofbuiltinplugin_test.go); both use a plugin, but don't require compilation.
 
   An .so file only appears in the flow if one chooses to write your
   own plugin, and write it as a Go plugin instead of an Exec plugin.
   Tradeoffs discussed [here](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/plugins/goPluginCaveats.md), a doc referenced by the main plugin doc authoring doc [here](https://github.com/kubernetes-sigs/kustomize/tree/master/docs/plugins).

 * One must use `--enable-alpha-plugin` to configure a builtin plugin.

   Yes.  That shouldn't be necessary.  Filed [issue](https://github.com/kubernetes-sigs/kustomize/issues/1504).

 * Where do you find the syntax for configuring a builtin
   plugin?

   That's certainly a problem! There are no docs detailing the
   configuration of the builtin plugins, besides, as you say, the unit
   tests. These tests, however undiscoverable, have the benefit of
   actually being known to be executionably correct because they run
   on every PR.
   
   Documentation PRs are welcome.

   Your [work](https://github.com/keleustes/kustomize/tree/plugin/examples/plugin]) at looks like a good start (wow).  But i fear we need a way to  _generate_ the documentation so that it doesn't get out of date as the plugins change.  Documentation at this level is hard to create unless it's automated.  Something like godoc would be nice.

 * We currently have to copy/paste all the "transformer.yaml" files to each base...  The transformers are not accumulated...
   
   Don't agree.  Plugin configs are just normal k8s objects.

   So one may collect them in a base, and reuse them in any number of other kustomizations, by referring to that kustomization in the `transformers:` field.  E.g. this [unit test](https://github.com/kubernetes-sigs/kustomize/blob/master/pkg/target/customconfigofbuiltinplugin_test.go#L105).

   That's how you get _write-once/use-many_ behavior from plugins.

 * we still need higher level operations / wildcards to go with FieldSpec.

   Maybe, but in the context of indivdiual transformer (plugin) configuration, not in the context of the global `configuration:` field which is now redundant, was never easy to use, and should be deprecated.

   All the discussion I've seen about excluding things from wildcarded transforms was excluding at the `Kind` level, which means we're talking about an exclusion field of type `gvk.Gvk`, not of type `config.FieldSpec`.

jbrette:
**.so issues**;  I added a bug because kustomize neither 3.1.0 nor master does seems to be able to locate the builtin module [here](https://github.com/kubernetes-sigs/kustomize/issues/1508) Let's hope this is just a question of documenting what to set in the environment.

jbrette:
In order to  be able compare the pros and cons of a the solution,  I updated the "canary/production" deployment environment:
- assuming the .so as in the .config folder, using the [transformer]( https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_b)
- using the [skip field spec](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_c)
- using separation using [kustomize context](https://github.com/keleustes/kustomize/tree/allinone/examples/issues/issue_0519_d)

All those three examples are building the same output, and pass mdrip --mode=test README.md

jbrette:
Regarding the "transfomers:" solution, [lease tell me what is missing, but using the tranformers method seems to currently constraint the user to copy paste the transformer configuration. 

Unlike the "configurations:" entry which is only accumulating fieldspec accross kustomize contexts and allow reuses, the yaml for instance of the prefixsuffix transformer contains the "suffix" value and the "field spec list". To be able to use two suffices in two different context, you end up have to copy paste the fieldspec.

Will add additional tests to address that point.

jbrette:

PR #1248, opened 2019-09-20
add KEP future of kubectl cp

sallyom:
**future-of-kubectl-cp**

### DIFF
--- a/keps/sig-cli/20190920-future-of-kubectl-cp.md
+++ b/keps/sig-cli/20190920-future-of-kubectl-cp.md
@@ -1,0 +1,165 @@
+---
+title: future-of-kubectl-cp
+authors:
+  - "@sallyom"
+owning-sig: sig-cli
+participating-sigs:
+  - sig-Usability
+reviewers:
+  - "@liggitt"
+  - "@brendandburns"
+approvers:
+  - "@pwittrock"
+  - "@soltysh"
+editor: TBD
+creation-date: 2019-09-20
+last-updated: 2019-09-20
+status: provisional
+---
+
+# future-of-kubectl-cp
+
+## Table of Contents
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals for kubectl cp](#goals-for-kubectl-cp)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [UserA wants to copy a file or a directory from a local host to a pod (or a container in a pod)](#usera-wants-to-copy-a-file-or-a-directory-from-a-local-host-to-a-pod-or-a-container-in-a-pod)
+    - [UserB wants to copy a file or a directory from a pod to a local host](#userb-wants-to-copy-a-file-or-a-directory-from-a-pod-to-a-local-host)
+  - [Implementation Details/Notes/Constraints [optional]](#implementation-detailsnotesconstraints-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+- [ ] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [ ] KEP approvers have set the KEP status to `implementable`
+- [ ] Design details are appropriately documented
+- [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] Graduation criteria is in place
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+This document summarizes and originates from this email thread, 
+[Proposal to drop kubectl cp](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-cli/_zUy67lK49k/aE6vncYiAgAJ).   
+
+This document aims to solidify the future of `kubectl cp` as a tool that provides basic function of copying files between local environments and pods.  Any advanced use cases
+such as those involving symlinks or modifying file permissions should be performed outside of `kubectl cp` through `kubectl exec`, addons, or shell commands.    
+
+Over the past few releases, there have been numerous security issues with `kubectl cp` that have resulted in release updates in all supported versions of kubectl.
+At the same time,any new PR that extends `kubectl cp` must undergo extra reviews to evaluate security threats that may arise [1][2].  Over the past few months,
+security fixes have required dropping edge cases and function of the command.  It is increasingly difficult to maintain a cp command that is both
+useful and secure.  There are alternative approaches that provide the same function as `kubectl cp` [3].  Using `kubectl exec ...| tar`
+provides transparency when copying files as well as mitigations for path traversals, symlink directory escapes, tar bombs, and other exploits.
+Use of tar is more featureful, in that it can preserve file permissions and copy pod-to-pod.  Also, `kubectl cp` is dependent on the tar binary
+in a container.  A malicious tar binary is outside of what `kubectl cp` can control.    
+
+With all of this in mind the cost and risk of maintaining the cp command should be weighed against what is considered crucial functionality in kubectl. 
+It's better to address 80% of use cases with a simple tool than trying to address the remaining 20% at the cost of risking those 80%.     
+
+[1] https://github.com/kubernetes/kubernetes/pull/78622   
+[2] https://github.com/kubernetes/kubernetes/pull/73053   
+[3] https://gist.github.com/tallclair/9217e2694b5fdf27b55d6bd1fda01b53   
+
+## Motivation
+
+- The `kubectl cp` command has been the subject of multiple security vulnerability reports.
+    * [CVE-2018-1002100](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-1002100)
+    * [CVE-2019-1002101](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-1002101)
+    * [CVE-2019-11246](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2019-11246)
+    * [CVE-2019-11249](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-11249)
+- To use `kubectl cp`, container images are required to have the tar binary. `kubectl cp` is not available when running containers from the minimal [scratch image](https://hub.docker.com/_/scratch/).    
+  Running from scratch is by itself a tactic to securing containers, as it encourages the best practice of limiting the tools packaged in an image to only what's required by a workload.   
+
+This proposal is that `kubectl cp` should perform only basic copying of files.  Advanced features of file copying should be out of scope for `kubectl cp`.  
+
+### Goals for kubectl cp
+
+- Provide simple function to copy a single file or directory, without advanced features such as symlinks or file permission changes
+- Ensure, through extra review process tbd, that features added to `kubectl cp` are well understood and easy to secure.
+- Offer users example `kubectl exec`/ shell commands to address advanced options of copying files. (There are a few already in --help)
+  - As an alternative, problematic `kubectl cp` code (handling symlinks) can be replaced with code to shell out to `kubectl exec ... | tar`
+
+### Non-Goals
+
+For either of these, a separate proposal weighing the cost/benefit would be required.  These are out of scope of this proposal to simplify `kubectl cp`:
+- Rewrite `kubectl cp` to not use tar, by modifying CRI as outlined partially [here](https://github.com/kubernetes/kubernetes/issues/58512). 
+- Rewrite `kubectl cp` to be functional in scratch based containers through use of ephemeral containers as outlined [here](https://github.com/kubernetes/kubernetes/issues/58512#issuecomment-528384746)
+
+## Proposal
+
+- `kubectl cp` should provide simple function of copying single file or directory between local environments and pods.
+- Identify and document `kubectl exec` commands to address more advanced options for copying files.  
+- Provide users attempting to use `kubectl cp + symlinks/etc` with output showing comparable `kubectl exec ...| tar` cmds.
+- It is up for a decision in this proposal whether the community prefers to implement the `shelling out to tar from within kubectl cp` 
+or leave as suggestions in error output. 
+- Barring decision of the above, only the user stories listed below should be supported by `kubectl cp`.  If additional user stories are added via shelling out to tar from kubectl, 
+  those will be outlined below.    
+
+### User Stories
+
+#### UserA wants to copy a file or a directory from a local host to a pod (or a container in a pod)
+
+```console
+ $ kubectl cp localdir somens/somepod:/poddir (-c acontainer)
+ $ kubectl cp localdir somens/somepod:/poddir (-c acontainer)
+ $ kubectl cp localdir/filea somens/somepod:/poddir (-c acontainer)
+```
+
+#### UserB wants to copy a file or a directory from a pod to a local host
+
+```console
+ $ kubectl cp somens/somepod:/poddir localdir
+ $ kubectl cp somens/somepod:/poddir/filea localdir/filea
+```
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
+
+Any scripts or automation that currently rely on advanced features of `kubectl cp` will be broken.
+To mitigate, detailed information about why the command now fails as well as example `kubectl exec ...| tar` alternatives will be output. 
+
+## Design Details
+
+### Test Plan
+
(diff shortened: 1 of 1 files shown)


keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+This document summarizes and originates from this email thread, 
+[Proposal to drop kubectl cp](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-sig-cli/_zUy67lK49k/aE6vncYiAgAJ).   
+
+This document aims to solidify the future of `kubectl cp` as a tool that provides basic function of copying files between local environments and pods.  Any advanced use cases
+such as those involving symlinks or modifying file permissions should be performed outside of `kubectl cp` through `kubectl exec`, addons, or shell commands.    
+
+Over the past few releases, there have been numerous security issues with `kubectl cp` that have resulted in release updates in all supported versions of kubectl.
+At the same time,any new PR that extends `kubectl cp` must undergo extra reviews to evaluate security threats that may arise [1][2].  Over the past few months,
+security fixes have required dropping edge cases and function of the command.  It is increasingly difficult to maintain a cp command that is both
+useful and secure.  There are alternative approaches that provide the same function as `kubectl cp` that are proven more secure [3].  Using `standalone pipe + tar`
```
mattfarina: What about scratch containers?

sallyom: right, i mentioned below that an issue is the container image has to have the tar binary.  it's in there, I could add a note about that there, too.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+provides transparency when copying files as well as more proven mitigations for path traversals, symlink directory escapes, tar bombs, and other exploits.
+Use of tar is also more more featureful, in that it can preserve file permissions and copy pod-to-pod.
```
brendandburns: You could easily implement the existing functionality by shelling out to 'tar' from within the 'kubectl' executable while maintaining the same ease of use for users of the tool. I'd like to see that considered as an alternative.

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: What is the evidence for the statement "proven more secure"? The foot note is to a script but there's no evidence that it is more secure.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+[1] https://github.com/kubernetes/kubernetes/pull/78622   
+[2] https://github.com/kubernetes/kubernetes/pull/73053   
+[3] https://gist.github.com/tallclair/9217e2694b5fdf27b55d6bd1fda01b53   
+
+## Motivation
+
+- The `kubectl cp` command has been the subject of multiple security vulnerability reports.
+- Even with basic use cases of `kubectl cp`, there are long-standing issues that have not been addressed.   
+     1) container images are required to have the tar binary in order to use `kubectl cp`.    
+     2) cp command is not well maintained for Windows, if it works at all.
```
brendandburns: What is the evidence for this statement?

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+With all of this in mind the cost and risk of maintaining the cp command should be weighed against what is considered crucial functionality in kubectl. 
+It's better to address 80% of use cases with a simple tool than trying to address the remaining 20% at the cost of risking those 80%.     
+
+[1] https://github.com/kubernetes/kubernetes/pull/78622   
+[2] https://github.com/kubernetes/kubernetes/pull/73053   
+[3] https://gist.github.com/tallclair/9217e2694b5fdf27b55d6bd1fda01b53   
+
+## Motivation
+
+- The `kubectl cp` command has been the subject of multiple security vulnerability reports.
```
brendandburns: RBAC and containers themselves have also been subject to multiple security vulnerability reports and there's no proposal to remove them from Kubernetes. The existence of security vulnerabilities is by itself not a motivation to remove features.

The motivation to remove features must be that we can not fix/improve the existing implementation, not that it is easier to remove it. Or it could be that the cost of fixing it is too costly relative to removing things.

Either those could be sufficient motivations (I'm not saying they are) but the simple presence of security issues is not a motivation in and of itself imho.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+- Even with basic use cases of `kubectl cp`, there are long-standing issues that have not been addressed.   
+     1) container images are required to have the tar binary in order to use `kubectl cp`.    
+     2) cp command is not well maintained for Windows, if it works at all.    
+
+Given its history, `kubectl cp` should perform only basic copying of files.  Advanced features of file copying should be out of scope for `kubectl cp`.
```
brendandburns: I don't think that this conclusion follows from the motivation above. Specifically there is no consideration for ways in which the existing implementation could be improved. The determination is made to simply remove functionality.

There needs to be a more clear discussion explaining why this is the right decision and what other alternatives were considered.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+- Even with basic use cases of `kubectl cp`, there are long-standing issues that have not been addressed.   
+     1) container images are required to have the tar binary in order to use `kubectl cp`.    
+     2) cp command is not well maintained for Windows, if it works at all.    
+
+Given its history, `kubectl cp` should perform only basic copying of files.  Advanced features of file copying should be out of scope for `kubectl cp`.  
+
+### Goals for kubectl cp
+
+- Provide simple function to copy a single file or directory, without advanced features such as symlinks or file permission changes
+- Don't accept new features or flags to cp beyond the use cases listed in the user stories below.
```
brendandburns: I don't think it is a good precedent to explicitly block future features without justification.

"Ensure that all features added to kubectl cp are well understood and easy to secure" 

I think better captures what you are trying to say.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+#### UserB wants to copy a file or a directory from a pod to a local host
+
+```console
+ $ kubectl cp somens/somepod:/poddir localdir
+ $ kubectl cp somens/somepod:/poddir/filea localdir/filea
+```
+
+### Implementation Details/Notes/Constraints [optional]
+
+### Risks and Mitigations
```
brendandburns: This needs to be expanded (e.g. this will definitely break people's scripts)

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Design Details
+
+### Test Plan
```
brendandburns: This needs to be expanded

How will you make sure you don't break people?

How will you demonstrate that you've improved the security?

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Graduation Criteria
+
+### Upgrade / Downgrade Strategy
```
brendandburns: This needs to be expanded to describe how you will deprecate the existing functionality (if indeed that is the right decision)

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Version Skew Strategy
+
+## Implementation History
+
+## Drawbacks
```
brendandburns: This needs to be expanded to discuss how this impacts existing users (and their automation)

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: good point, b4 i make a statement like that i should do my research (i took that point from the email thread w/out verifying) will do..

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: again, good point, i'll rethink/change wording there

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: that footnote is for the example tar cmds, I've moved it.  the 'proven more secure' was meant to be explained by the following statements: `provides transparency when copying files as well as more proven mitigations for path traversals, symlink directory escapes, tar bombs, and other exploits.  Use of tar is also more more featureful, in that it can preserve file permissions and copy pod-to-pod`  however, the `more proven mitigations` can use some proof, as I haven't listed any.   I've moved the footnote, removed the `proven more secure`.

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: true, up for discussion

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: hm, this is still open, although there is an open PR: https://github.com/kubernetes/kubernetes/issues/77310

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: :+1: , but that's true for any feature of any command.  I'm trying to express that `kubectl cp` has a history of security issues, and because of this extra consideration/scrutiny should be taken when accepting new features.

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: I'm not proposing we remove anything (more than what's already been removed recently - such as symlink support) but more suggesting we accept  `kubectl cp` as it is currently - a tool for copying files without advanced features like file permission control, pod-to-pod copies, symlink support.  This is rather than:
a) remove `kubectl cp`
b) rewrite it   

Yes, that discussion can happen here : ).  If during this KEP review we modify it to a) or b) that would be fine with me.  That's why I've titled it `future-of-kubectl-cp` to keep the discussion from the thread going.

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: ^^ I'm confused now about the `preserving file permissions` in kubectl cp, doing local testing to sort out what's currently supported.  there isn't any useful info on that in the help menu. hm..

keps/sig-cli/20190920-future-of-kubectl-cp.md
soltysh: The process will simple, we have to remove that functionality and inform users of alternatives, which is what we've been doing on every CVE fix. My proposal is: user is getting information that file is a symlink and we will skip this file, along with detailed information how to use `kubectl exec` as a replacement. 
@sallyom mind adding something like that here

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Goals for kubectl cp
+
+- Provide simple function to copy a single file or directory, without advanced features such as symlinks or file permission changes
+- Ensure that features added to `kubectl cp` are well understood and easy to secure.
+- Offer users example `kubectl exec`/ shell commands to address advanced options of copying files. (There are a few already in --help)
+
+### Non-Goals
+
+- Rewrite `kubectl cp` to not use tar, by modifying CRI as outlined partially [here](https://github.com/kubernetes/kubernetes/issues/58512).
```
BenTheElder: what about using an ephemeral container? https://github.com/kubernetes/kubernetes/issues/58512#issuecomment-533640463
- still uses tar, but one kubernetes controls
- doesn't involve modifying CRI
- may not require dropping this feature from kubectl

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: Since this is a general comment about `kubectl`, this is unrelated to `kubectl cp` the topic of this KEP, so this line should be removed.

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: I think "extra care and attention or somesuch" is fine, it's the explicit block that I don't like...

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: I don't disagree, but the section still needs to be filled in, b/c without it we're taking on faith that there is a test plan...

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: As above, agree it's easy, but it needs to be filled in.

keps/sig-cli/20190920-future-of-kubectl-cp.md
brendandburns: Seems like it could be good, but it will require a bunch of kubelet work, so we should weigh cost/benefit....

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: i've added info there

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: added info

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: added as a non-goal for this proposal, but agreed it could solve the issue of requiring tar in image

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: added info there

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: sure, done (the ones I know about/could find)

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: I've removed that.

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: I've added information to discuss there.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Design Details
+
+### Test Plan
+
+### Graduation Criteria
+
+### Upgrade / Downgrade Strategy
+
+Any function that is removed as a result of a CVE fix or other will be documented clearly.  Any function that is removed will result in the user getting
```
brendandburns: There needs to be a deprecation period where we print a warning but don't actually remove the functionality. Please document what you plan to do to warn people prior to removing functionality.

keps/sig-cli/20190920-future-of-kubectl-cp.md
```diff
@@ -0,0 +1,158 @@
...
+Any function that is removed as a result of a CVE fix or other will be documented clearly.  Any function that is removed will result in the user getting 
+information about why this subcommand/option is no longer supported and also will be given alternative `kubectl exec` commands as a replacement in the failed command output. 
+
+### Version Skew Strategy
+
+## Implementation History
+
+## Drawbacks
+
+Users relying on automation with `kubectl cp` may be inconvenienced if options and features are removed from the command.  The motivation of improving security is weighed
```
brendandburns: I think `inconvenienced` is pretty light. Users will be forced to rewrite continuous delivery scripts. Applications and installers that they have written will fail and need to be rewritten. It's possible that source for images that rely on this will be lost, etc etc.

This is a breaking change and we need to take any such change very seriously for the impact that it has on our users.

brendandburns:
Thanks for the additional detail, I added a few more comments, I'm still looking for details on the test plan.

Additionally, this is missing any explanation as to why `kubectl exec < foo> | tar` is better than performing equivalent functionality inside of `kubectl cp`.

Why not simply replace the code that is problematic with shelling out.  You could even literally shell out to `bash -c "kubectl exec ... | tar"`. That would allow you to preserve the existing UX and still obtain the perceived benefits of using a pipe and tar instead of go code...

keps/sig-cli/20190920-future-of-kubectl-cp.md
BenTheElder: Potentially it also makes the tar binary then relatively trusted?
What kuebelet work does this require? Didn't we already add ephemeral containers?

keps/sig-cli/20190920-future-of-kubectl-cp.md
soltysh: We most probably be printing that warning always, given the limited capability I'd like to point users to what they need to do to workaround the problem. @sallyom can you add that here.

keps/sig-cli/20190920-future-of-kubectl-cp.md
soltysh: I do agree the wording is not the best, I don't think we can do more than inform users :disappointed:

sallyom: updated wording

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: added info

keps/sig-cli/20190920-future-of-kubectl-cp.md
sallyom: added the info will always be output, not just for deprecation period.

sallyom:
> Additionally, this is missing any explanation as to why `kubectl exec < foo> | tar` is better than performing equivalent functionality inside of `kubectl cp`.
> 
There is info in the Summary, it could be moved or re-worded though. 
If the tar binary in the container is corrupt, it could output unexpected, malicious results and `kubectl cp` doesn't control the tar binary.

jayunit100:
is the future of kubectl cp mostly that it will "stay the same" based on this issue  ? I dont see alot of new KEPS about it when i tried to search. 

cc @keganhollern

PR #1263, opened 2019-10-01
Kubernetes's OpenAPI v3 Specification

p0lyn0mial:
I still need to fill in a few sections but would like to gather your feedback before.

### DIFF
--- a/keps/sig-api-machinery/20191001-oasv3.md
+++ b/keps/sig-api-machinery/20191001-oasv3.md
@@ -1,0 +1,949 @@
+---
+title: Kubernetes's OpenAPI v3 Specification
+authors:
+  - @p0lyn0mial
+owning-sig: sig-api-machinery
+participating-sigs:
+  - sig-api-machinery
+reviewers:
+  - @sttts
+  - @yue9944882
+approvers:
+  - @sttts
+  - @yue9944882
+editor: TBD
+creation-date: 2019-10-01
+last-updated: 2019-10-01
+status: provisional
+---
+
+# Kubernetes's OpenAPI v3 Specification
+
+## Table of Contents
+
+TBD
+
+## Release Signoff Checklist
+
+**ACTION REQUIRED:** In order to merge code into a release, there must be an issue in [kubernetes/enhancements] referencing this KEP and targeting a release milestone **before [Enhancement Freeze](https://github.com/kubernetes/sig-release/tree/master/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core Kubernetes i.e., [kubernetes/kubernetes], we require the following Release Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These checklist items _must_ be updated for the enhancement to be released.
+
+- [ ] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [ ] KEP approvers have set the KEP status to `implementable`
+- [ ] Design details are appropriately documented
+- [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] Graduation criteria is in place
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+**Note:** Any PRs to move a KEP to `implementable` or significant changes once it is marked `implementable` should be approved by each of the KEP approvers. If any of those approvers is no longer appropriate than changes to that list should be approved by the remaining approvers and/or the owning SIG (or SIG-arch for cross cutting KEPs).
+
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+TDB
+
+## Motivation
+
+TBD
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+
+### Goals
+- Define OpenAPI v3 Specification that will be published by the API servers
+- Improve the definition names
+   - do not leak package paths
+   - be prepared for $ref support (e.g. referencing a PodSpec from a CRD)
+- Reuse parameters
+- Reduce CPU load due to spec merging
+- Align aggregated API server OpenAPI specs with per-group APIServices
+- Allow aggregated API servers to publish v2 and v3 to kube-aggregator, without doubling the CPU load
+
+### Non-Goals
+ - Define a spec that uses every field from OASv3 
+
+## Proposal
+### The spec
+The following example demonstrates the specification structure that will be published by the API servers. It defines two endpoints(`paths`), a set of objects(`components/schemas`) and reusable parameters(`components/parameters`) for operations e.g. `GET`.
+
+It was created from [spec3](https://github.com/kubernetes/kube-openapi/pull/169) package that was created based on [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md).
+
+TODO: I added the following sniped so that comments can be added more easily, once reviewed I'm going to move it to a file and paste a link instead.
+```json
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Kubernetes",
+    "license": {
+      "name": "Apache 2.0"
+    },
+    "version": "v1.16.0"
+  },
+  "paths": {
+    "/api/v1/componentstatuses/{name}": {
+      "get": {
+        "tags": [
+          "core_v1"
+        ],
+        "description": "read the specified ComponentStatus",
+        "operationId": "readCoreV1ComponentStatus",
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ComponentStatus"
+                }
+              },
+              "application/vnd.kubernetes.protobuf": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ComponentStatus"
+                }
+              },
+              "application/yaml": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ComponentStatus"
+                }
+              }
+            }
+          },
+          "401": {
+            "description": "Unauthorized"
+          }
+        }
+      },
+      "parameters": [
+        {
+          "name": "name",
+          "in": "path",
+          "description": "name of the ComponentStatus",
+          "required": true,
+          "schema": {
+            "type": "string"
+          }
+        },
+        {
+          "$ref": "#/components/parameters/query.pretty.5ca316"
+        }
+      ]
+    },
+    "/api/v1/namespaces/{namespace}/configmaps/{name}": {
+      "get": {
+        "tags": [
+          "core_v1"
+        ],
+        "description": "read the specified ConfigMap",
+        "operationId": "readCoreV1NamespacedConfigMap",
+        "parameters": [
+          {
+            "name": "exact",
+            "in": "query",
+            "description": "Should the export be exact.  Exact export maintains cluster-specific fields like 'Namespace'. Deprecated. Planned for removal in 1.18.",
+            "schema": {
+              "type": "boolean"
+            }
+          },
+          {
+            "name": "export",
+            "in": "query",
+            "description": "Should this value be exported.  Export strips fields that a user can not specify. Deprecated. Planned for removal in 1.18.",
+            "schema": {
+              "type": "boolean"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/vnd.kubernetes.protobuf": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/yaml": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              }
+            }
+          },
+          "401": {
+            "description": "Unauthorized"
+          }
+        }
+      },
+      "put": {
+        "tags": [
+          "core_v1"
+        ],
+        "description": "replace the specified ConfigMap",
+        "operationId": "replaceCoreV1NamespacedConfigMap",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/query.pretty.5ca316"
+          },
+          {
+            "name": "fieldManager",
+            "in": "query",
+            "description": "fieldManager is a name associated with the actor or entity that is making these changes. The value must be less than or 128 characters long, and only contain printable characters, as defined by https://golang.org/pkg/unicode/#IsPrint.",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
+        "requestBody": {
+          "content": {
+            "*/*": {
+              "schema": {
+                "$ref": "#/components/schemas/core.v1.ConfigMap"
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/vnd.kubernetes.protobuf": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/yaml": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              }
+            }
+          },
+          "201": {
+            "description": "Created",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/vnd.kubernetes.protobuf": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              },
+              "application/yaml": {
+                "schema": {
+                  "$ref": "#/components/schemas/core.v1.ConfigMap"
+                }
+              }
+            }
+          },
+          "401": {
+            "description": "Unauthorized"
+          }
+        }
+      },
+      "delete": {
+        "tags": [
+          "core_v1"
+        ],
+        "description": "delete a ConfigMap",
+        "operationId": "deleteCoreV1NamespacedConfigMap",
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/query.pretty.5ca316"
+          },
+          {
+            "name": "gracePeriodSeconds",
+            "in": "query",
+            "description": "The duration in seconds before the object should be deleted. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period for the specified type will be used. Defaults to a per object value if not specified. zero means delete immediately.",
+            "schema": {
+              "type": "integer"
+            }
+          },
+          {
+            "name": "orphanDependents",
+            "in": "query",
+            "description": "Deprecated: please use the PropagationPolicy, this field will be deprecated in 1.7. Should the dependent objects be orphaned. If true/false, the \"orphan\" finalizer will be added to/removed from the object's finalizers list. Either this field or PropagationPolicy may be set, but not both.",
+            "schema": {
+              "type": "boolean"
+            }
+          },
+          {
+            "name": "propagationPolicy",
+            "in": "query",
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+This not only doesn't add any value but also reveals the internal path to a type. 
+
+Instead the new names will contain information about a group, version and kind(`GVK`). `ConfigMap` in the new format would become `core.v1.ConfigMap`.
+
+### Reusable parameters
+
+In order to make the spec smaller and more clearer common parameters for operations will be kept under `components/parameters`. 
+The [OASv3 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object) says that a unique parameter is defined by a combination of a name and location. 
+Thus to make a parameter unique across all operations its name will be derived as `name.location.SHA1(spec3.Parameter.MarshalJSON()) | hex.EncodeToString() | hexEncodedString[1:5]`.
+The last part takes the first 6 characters from a hex-encoded `SHA1` value that was calculated from a JSON encoded representation of a parameter e.g. `query.pretty.5ca316`.
```
jpbetz: Won't we be at risk of collisions when using only the first 6 characters of a encoded `SHA1`?

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+The old spec uses the full path to a package as a name for an object, for example `ConfigMap` is defined as `io.k8s.api.core.v1.ConfigMap`.
+This not only doesn't add any value but also reveals the internal path to a type. 
+
+Instead the new names will contain information about a group, version and kind(`GVK`). `ConfigMap` in the new format would become `core.v1.ConfigMap`.
+
+### Reusable parameters
+
+In order to make the spec smaller and more clearer common parameters for operations will be kept under `components/parameters`. 
+The [OASv3 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object) says that a unique parameter is defined by a combination of a name and location. 
+Thus to make a parameter unique across all operations its name will be derived as `name.location.SHA1(spec3.Parameter.MarshalJSON()) | hex.EncodeToString() | hexEncodedString[1:5]`.
```
jpbetz: Would this be for parameter types like `CreateOptions`? For that I was expecting an identifier like `meta.v1.CreateOptions`.

It would seem better to share parameter definitions because we know they're the same (e.g. they share the same go type), not because they happen to have the same same json representation. E.g. have a single parameter definition for `CreateOptions` because we know they're all the same type, but not for two paramters that happen to look the same but have different purposes.

keps/sig-api-machinery/20191001-oasv3.md
p0lyn0mial: IIRC this is related to [the birthday problem](https://en.wikipedia.org/wiki/Birthday_problem). The chances of a collision can be calculated from `k^2/2N`. In this case, we could expect the first collision after generating `~5700` values. Which is enough for our needs.

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+Since an aggregated API server can claim more that one group, the OpenAPI specifications will be published per group for exactly the same reasons as in the previous section.
+
+### Implementation Details
+
+- There is no need to maintain and generate separate schema definitions (`GetOpenAPIDefinitions`) for v2 and v3 specs. 
+  [A new extension](https://github.com/kubernetes/kube-openapi/pull/166) `x-kubernetes-v2-spec` will be used for those types have that differences in v2 and v3 schema. 
+  For now, it will be meant for `IntOrString`. In v2 spec, a workaround was made which will be changed by the use of `anyOf` for v3.
+  
+- In order to generate the new names for the definitions `openapi-gen` will be changed.
+  For types that don't have kind e.g. `PodSpec` the golang type will be used.
```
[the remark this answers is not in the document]
p0lyn0mial: correct, I will update the doc.

keps/sig-api-machinery/20191001-oasv3.md
p0lyn0mial: this will be primarily for `query nad path` parameters for operations, that contain simple types like `string, boolean`. 

types like `CreateOptions` are now defined in a [RequestBody](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#requestBodyObject) field and will contain `$ref` identifier pointing to a definition, for example:

https://github.com/kubernetes/enhancements/pull/1263/files#diff-e3ca43e3e2f48d58ccbbb9f49671c017R712

https://github.com/kubernetes/enhancements/pull/1263/files#diff-e3ca43e3e2f48d58ccbbb9f49671c017R306

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+TBD
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+
+### Goals
+- Define OpenAPI v3 Specification that will be published by the API servers
+- Improve the definition names
+   - do not leak package paths
+   - be prepared for $ref support (e.g. referencing a PodSpec from a CRD)
+- Reuse parameters
```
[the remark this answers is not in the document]
p0lyn0mial: sure, thx, I'll add it to the doc.

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+- Reduce CPU load due to spec merging
+- Align aggregated API server OpenAPI specs with per-group APIServices
+- Allow aggregated API servers to publish v2 and v3 to kube-aggregator, without doubling the CPU load
+
+### Non-Goals
+ - Define a spec that uses every field from OASv3 
+
+## Proposal
```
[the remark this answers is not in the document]
p0lyn0mial: sure, thx.

p0lyn0mial:
@jpbetz thanks for the feedback.

keps/sig-api-machinery/20191001-oasv3.md
jpbetz: The options are both on the body and in query params. For example, `dryRun`, `gracePeriodSeconds`, and all the `DeleteOptions` are also listed as query params: https://github.com/kubernetes/enhancements/pull/1263/files#diff-e3ca43e3e2f48d58ccbbb9f49671c017R720.

Since we know the type these parameters are defined with, would it be reasonable to have those represented in OpenAPI with something like this?

```
...
      "parameters": [
          {
            "name": "dryRun",
            "in": "query"
            "schema": {
              "$ref": "#/components/parameters/meta.v1.DeleteOptions/dryRun"
            }
          },
       ]
...
```

keps/sig-api-machinery/20191001-oasv3.md
jpbetz: I was curious, so I checked and it looks like we only have ~35 distinct parameters. Probibly a couple more if we distinguish between dryRun from the different options types.

I'd be inclined to give the common ones clear names like `meta.v1.DeleteOptions/dryRun` to all the common ones and the just inline the less common ones and avoid the hashes altogether.

```
curl -s localhost:8085/openapi/v2 | jq '..|. | select(type=="object") | select(has("parameters")) | .parameters | select(type=="array")| .[].name' | sort | uniq -c | sort -rn
    452 "dryRun"
    452 "body"
    432 "pretty"
    365 "watch"
    365 "timeoutSeconds"
    365 "resourceVersion"
    365 "limit"
    365 "labelSelector"
    365 "fieldSelector"
    365 "continue"
    365 "allowWatchBookmarks"
    316 "fieldManager"
    233 "namespace"
    201 "name"
    136 "propagationPolicy"
    136 "orphanDependents"
    136 "gracePeriodSeconds"
    115 "force"
     61 "export"
     61 "exact"
      9 "path"
      3 "container"
      2 "tty"
      2 "stdout"
      2 "stdin"
      2 "stderr"
      1 "timestamps"
      1 "tailLines"
      1 "sinceSeconds"
      1 "previous"
      1 "ports"
      1 "logpath"
      1 "limitBytes"
      1 "follow"
      1 "command"
```

keps/sig-api-machinery/20191001-oasv3.md
p0lyn0mial: not all `dryRun` parameters are equal, here they have different types:

`dryRun` from an operation:
```
{
  "description": "When present, indicates that modifications should not be persisted. An invalid or unrecognized dryRun directive will result in an error response and no further processing of the request. Valid values are: - All: all dry run stages will be processed",
  "in": "query",
  "name": "dryRun",
  "type": "string",
  "uniqueItems": true
}
```

`dryRun` from `meta.v1.DeleteOptions`:
```
{
  "description": "When present, indicates that modifications should not be persisted. An invalid or unrecognized dryRun directive will result in an error response and no further processing of the request. Valid values are: - All: all dry run stages will be processed",
  "type": "array",
  "items": {
    "type": "string"
  }
}
```

sometimes `gracePeriodSeconds` differ in `format`:

`gracePeriodSeconds` from `meta.v1.DeleteOptions`:
```
"gracePeriodSeconds": {
  "description": "The duration in seconds before the object should be deleted. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period for the specified type will be used. Defaults to a per object value if not specified. zero means delete immediately.",
  "type": "integer",
  "format": "int64"
}
```

`gracePeriodSeconds` from  an operation:
```
{
  "description": "The duration in seconds before the object should be deleted. Value must be non-negative integer. The value zero indicates delete immediately. If this value is nil, the default grace period for the specified type will be used. Defaults to a per object value if not specified. zero means delete immediately.",
  "in": "query",
  "name": "gracePeriodSeconds",
  "type": "integer",
  "uniqueItems": true
}
```

Sometimes the parameters differ only in one field, like in this example, only `description` is different:
```
{
  "name": "name",
  "in": "path",
  "description": "name of the ComponentStatus",
  "required": true,
  "schema": {
    "type": "string"
  }
}
```

```
{
  "name": "name",
  "in": "path",
  "description": "name of the ConfigMap",
  "required": true,
  "schema": {
    "type": "string"
  }
}
```

p0lyn0mial: even if we could tell they are the same then I think we would like to change `meta.v1.DeleteOptions` to reference the common parameter:
```
"meta.v1.DeleteOptions": {
  ...
  "properties": {
    "dryRun": {
       "$ref": "#/components/parameters/meta.v1.DeleteOptions/dryRun"
    }
    …
}
```

but it looks strange to reference `components/parameters` from `components/schemas`.

if this idea makes the spec hard to read then I am fine with abandoning it.

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+
+TBD
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+
+### Goals
+- Define OpenAPI v3 Specification that will be published by the API servers
+- Improve the definition names
+   - do not leak package paths
+   - be prepared for $ref support (e.g. referencing a PodSpec from a CRD)
```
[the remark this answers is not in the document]
p0lyn0mial: Thanks for linking this issue, I didn't now that the names can change from release to release. One way of making sure that the names are stable would be to add an additional tag close to the type that would denote Kind for that type.

keps/sig-api-machinery/20191001-oasv3.md
jpbetz: Interesting. I hadn't considered restructuring `meta.v1.DeleteOptions` to reference the params. It might make more sense to just put the options params in `components/schemas` (instead of `components/parameters`) if we did that. But when I first saw it, my initial thought was also that it might be inconvenient to read. Even if we don't take it that far, having the param schemas consolidated could be an improvement over what we have today.

Thanks for digging in on this.

keps/sig-api-machinery/20191001-oasv3.md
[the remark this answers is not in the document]
p0lyn0mial: another way would be to take the name from a JSON tag. The tags have the advantage of being stable as they appear in the responses so changing a tag would be considered backward-incompatible.

https://play.golang.org/p/gWtpyGkuRA0

WDYT?

keps/sig-api-machinery/20191001-oasv3.md
roycaihw: currently `metav1.DeleteOptions` is supported both in HTTP body and url query parameter, and the openapi spec is correctly reflecting that. A couple of bugs:

- `dryRun` from an operation query parameter shouldn't be of `string` type. The in memory structure is a string slice and I believe our query parser supports array. 
- `gracePeriodSeconds` from an operation lacks format: we can [add format detail](https://github.com/kubernetes/kubernetes/blob/3317805652d145c4003f5c534c7d87fb5cd87d50/staging/src/k8s.io/apiserver/pkg/endpoints/installer.go#L996-L1003) when we generate the spec

The query parameter and the body parameter should be consistent since they are both generated from the go type. I don't know if OAS v3 supports referring query parameters like https://github.com/kubernetes/enhancements/pull/1263/files#r330702453 suggested though

p0lyn0mial: I think that they point to the same members of the same struct.

It seems like both [the typed client](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go#L465) and [the dynamic client](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/dynamic/simple.go#L218) always use `metav1.DeleteOptions` as a body parameter. Moreover, all the other parameters are passed as query parameters, why `metav1.DeleteOptions` are special?

`metav1.DeleteOptions` as a query parameter doesn't have [preconditions](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L474) field, this might be due to https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/installer.go#L970

p0lyn0mial: Knowing that they should point to the same structure I think we could just use the scheme. For example:

```
{
  "name": "DeleteOptions",
  "in": “query”,
  "schema": {
    "$ref": "#/components/schemas/meta.v1.DeleteOptions
  }
 …
}
```

we could use `style` and `explode` keywords to control the serialization of the parameter - https://swagger.io/docs/specification/serialization/

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+  }
+}
+```
+
+### New names for objects
+
+The old spec uses the full path to a package as a name for an object, for example `ConfigMap` is defined as `io.k8s.api.core.v1.ConfigMap`.
+This not only doesn't add any value but also reveals the internal path to a type. 
+
+Instead the new names will contain information about a group, version and kind(`GVK`). `ConfigMap` in the new format would become `core.v1.ConfigMap`.
```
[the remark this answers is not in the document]
yue9944882: > Instead the new names will contain information about a group, version and kind(`GVK`). `ConfigMap` in the new format would become `core.v1.ConfigMap`.

for the record, there's no such group as "core", the group name is actually an empty string "" for so-called "core" group.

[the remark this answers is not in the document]
yue9944882: > https://github.com/kubernetes-client/gen/blob/a4d67bd881b30c5f2b1579d38b3d5dfb86e9d086/openapi/preprocess_spec.py#L213-L228. I will double check

it can also be used by third-party libraries i suppose.. we have to preserve backward-compatibilities when refactoring the script.

keps/sig-api-machinery/20191001-oasv3.md
yue9944882: > It seems like both the typed client and the dynamic client always use metav1.DeleteOptions as a body parameter. Moreover, all the other parameters are passed as query parameters, why metav1.DeleteOptions are special?

it's because we can't express nested structs in the query parameters. all fields are flattened in queries.

yue9944882: > we could use style and explode keywords to control the serialization of the parameter - https://swagger.io/docs/specification/serialization/

👍🏻the default setting i.e `style=form,explode=true` is working for us. but the specification doesnt seem to define how to deal with nested structures from the schema object.

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+The [OASv3 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object) says that a unique parameter is defined by a combination of a name and location. 
+Thus to make a parameter unique across all operations its name will be derived as `name.location.SHA1(spec3.Parameter.MarshalJSON()) | hex.EncodeToString() | hexEncodedString[1:5]`.
+The last part takes the first 6 characters from a hex-encoded `SHA1` value that was calculated from a JSON encoded representation of a parameter e.g. `query.pretty.5ca316`.
+
+### Publish specifications per API group
+The API groups were primarily implemented to make it easier to extend the Kubernetes API. 
+This is reflected in the REST paths where each request must be in the form of `/apis/$GROUP_NAME/...`.
+
+Thus exposing OpenAPI specification should follow exactly the same pattern.
+This proposal proposes to publish per-group spec under `apis/$GROUP_NAME/openapi/v3` as opposed to the current approach which exposes one document for all groups under `openapi/v2`
```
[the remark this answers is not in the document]
yue9944882: > If we remove the aggregated one, how would a client (e.g. kubectl) discover and download all the per-group specs?

probably looping over the paths returned from discovery `/apis`?

another issue for per-group openapi is the AA servers. for now the AA servers can be compiled with older kubernetes libraries which misses per-group openapi, a 404 will be returned if {kube-aggregator#openapi-controller, or kubectl} tries fetching per-group specs from those groups. am leaning on keep the aggregated path for backward-compatibility..

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/meta.v1.Status"
```
yue9944882: for delete APIs, in v3 openapi, we should return an `anyOf` wrapping `meta.v1.Status` and `<actual GVK types>`, isn't it?

keps/sig-api-machinery/20191001-oasv3.md
roycaihw: > why metav1.DeleteOptions are special?

also see: https://github.com/kubernetes/kubernetes/issues/34856

keps/sig-api-machinery/20191001-oasv3.md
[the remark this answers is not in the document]
roycaihw: what we did when we introduced the v2 endpoint was that
- client tries the new endpoint and falls back to the old one if the new one doesn't exist
- server keeps the old endpoint for 4 releases following the deprecation policy: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#openapi-and-swagger-definitions

the case for v3 may be different

yue9944882:
for the record, what's the targeting milestone this time do you think?

p0lyn0mial:
> for the record, what's the targeting milestone this time do you think?

the next release `1.18` - I need to rethink those parameters - thanks for shedding light on it.

jpbetz:
@p0lyn0mial I realized that we are not handling enums correctly in OpenAPI v2.

For example, for matchType, I was expecting:

```
"matchPolicy": {
  "description": "...",
  "type": "string",
  "enum": ["Exact", "Equivalent"]
},
```

But in OpenAPI v2 we have:

```
"matchPolicy": {
  "description": "...",
  "type": "string",
  "enum": ["Exact", "Equivalent"]
},
```

Let's make this a requirement for v3 since that would be the easiest way to add support without potentially breaking generated client bindings by introducing it in v2. I already have working code and opened https://github.com/kubernetes/gengo/pull/157 to start getting i merged.

mitar:
So OpenAPI v3 supports [callbacks](https://swagger.io/docs/specification/callbacks/) ([spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.0.md#callback-object)). Would it be possible to define for all resources which can be watched, the responses of watching using those callbacks?

Similarly OpenAPI v3 supports [links](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.0.md#link-object). I think adding those would also be useful.

Why I am thinking about this is because I am looking into how to express Kubernetes API as GraphQL. There are [tools](https://github.com/IBM/openapi-to-graphql) which know how to convert between Swagger and GraphQL. And GraphQL supports subscriptions, which would be a natural way to have a standard interface to watching. So by describing the Kubernetes API and providing callbacks and links in the description, we would be much closer to GraphQL API as well.

keps/sig-api-machinery/20191001-oasv3.md
```diff
@@ -0,0 +1,949 @@
...
+- Allow aggregated API servers to publish v2 and v3 to kube-aggregator, without doubling the CPU load
+
+### Non-Goals
+ - Define a spec that uses every field from OASv3 
+
+## Proposal
+### The spec
+The following example demonstrates the specification structure that will be published by the API servers. It defines two endpoints(`paths`), a set of objects(`components/schemas`) and reusable parameters(`components/parameters`) for operations e.g. `GET`.
+
+It was created from [spec3](https://github.com/kubernetes/kube-openapi/pull/169) package that was created based on [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md).
```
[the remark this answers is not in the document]
kumarshantanu: +1 to @roycaihw since [OpenAPI 3.1.0](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.1.0.md) is released as well.

jkone27:
would be really nice to have this 👍 , any updates in 2021?

ehashman:
/close

This KEP has not been updated in over a year. Please reopen when you've had a chance to update to the new template/address comments.

@mitar please don't remove the stale label from things that are not being actively worked on. It creates more work for maintainers to have to manually close stale PRs.

roycaihw:
cc @jefftree

PR #1265, opened 2019-10-01
api-machinery: KEP for immutable fields

sttts:
Continuous https://github.com/kubernetes/enhancements/pull/1099


keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -46,100 +56,112 @@ i.e. after creation of an object, those fields cannot be mutated anymore. E.g. a
...
+  foo:
+    type: object
+    x-kubernetes-immutable: true
+```
 
-- Error: mutation of an immutable field returns an error and the request is
-  rejected. No mutation is performed at all.
-- Ignored: mutation of the field (and/or sub-fields) are completely ignored, and
-  not persisted. They do not trigger an error.
+means that the whole `.foo` object can be removed if set, or set if it is unset, but
```
deads2k: allowing a delete and a recreate with different values defies my expectations.  This is consistent with how it's treated for objects in some way? Seems weird to allow removing entire stanzas to remove values.  Is this to make it work nicely with unions?

munnerz: I *think* this is what @deads2k is already alluding to, but how is a delete and insert identified differently from an update of an entry in-place? It seems like this sort of thing would only be possible if using a list that is keyed on, e.g. a `name` field. Even then, I agree it seems like an odd thing to permit as from a users perspective, they now just have to perform two API operations instead of one in order to update certain fields.

sttts: I get your point. tl/dr: you question the sense of pure removal/addition without identifier. I don't object.

What about string slices? Is the identifier implicitly the value and we would allow a deletion and an insertion in a special mode, but no order changes?

Note, we are very flexible what to allow. The big question is what we really want to offer in Kubernetes-like APIs defined through CRDs.

sttts: Not sure your two question are actually the same. I fixed what I think @deads2k pointed out. My example was wrong. Specifying `x-kubernetes-immutable` applies to the field itself, not only the sub-fields in case of an object.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,308 @@
...
+For native resources we add a pre-immutability-check normalization step for objects 
+decoded from JSON which have normalizations defined:
+
+1. versioned JSON blob comes in a request
+2. unmarshalled into versioned Golang struct
+3. defaulting
+4. conversion to internal
+5. if immutability is enabled for the resource:
+   1. marshalling into JSON in-memory
+   2. normalizaion creating a copy with shared data-structures
```
marun: nit: normalization

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,308 @@
...
+   3. strict immutability check against the old object, coming from proto assuming it is normalized   
+
+### Future outline: field selection
+
+By default the equality is a strict deep-equal equality on a field. This can be generalized
+to a immutability mode. Here are examples:
```
marun: nit: s/a/an/

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,308 @@
...
+- **Non-recursive**: For a list or a map, one can not remove or add new items, but existing items can be modified.
+- **Recursive with allowed addition and deletion**:
+  - New/deleted keys in maps are tolerated.
+  - Appended items or removal of the last item in arrays are tolerated.
+- **Mutable**: A field set as mutable could cancel the recursive immutability.
+
+In CRDs these modes could be set via `x-kubernetes-immutable-mode: non-recursive/recursive/...`. Inside
+of Golang types the `+immutable` marker could be extended to `// +immutable=non-recursive/recursive/...`.
+
+More complex modes like allowed addition and removal with a custom identifies can be expressed with a
```
marun: s/identifies/identifier/ ?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,308 @@
...
+Containers []Container
+```
+
+### Future outline: protobuf
+
+Protobuf encoding identifies unset, empty and null values for slices and maps. In native types
+we use a semantic equality to factor those differences out. In CRs decoding and encoding
+is always faithful with the traditional JSON wire format.
+
+When adding protobuf encoding to CRDs in the future, we have to (without major, unnaturally
```
marun: nit: s/unnaturally/unnatural/

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,308 @@
...
+modes which replicate the Golang serialization behaviour. Tooling like openapi-gen
+and the CRD validation could verify that the normalization specification matches 
+that of Golang.
+
+### Mutating admission chain
+
+Mutating admission chain would have the exact same effects as user changes,
+meaning that they wouldn't be able to change an object after creation. This is
+very similar to what is done today since validation for update is run AFTER all
+mutations. On creation, mutation are permitted.
```
marun: nit: 'mutations are' or 'mutation is'

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+  foo:
+    type: array
+    items:
+      x-kubernetes-immutable: true
+      type: string
+```
+
+means that the `foo` array itself is mutable, but the items are not. This means that
+items can be deleted and added, but at each existing index the string values are
+immutable.
```
jpbetz: Might be worth mentioning that since maps are OpenAPI arrays (with `x-kubernetes-list-type`: `map`) that this also covers the "the restriction of the equality to only map keys, but not their values" goal listed above.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+// No single field in existing containers can be changed, added or deleted.
+// New containers can be added, existing containers can be removed. Containers
+// are identified by its name.
+// +immutable=addition=allowed,removal=allowed,identifier=.name
+Containers []Container
+```
+
+### Future outline: protobuf
+
+Protobuf encoding identifies unset, empty and null values for slices and maps. In native types
```
jpbetz: The phrasing in this paragraph seems incompatible (or I'm just parsing it wrong) with what we say above with "the addition of protobuf or other encodings which unify values like empty, null and undefined.". Should this be "Protobuf encoding unifies..." too?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+deep-equal equality even for protobuf, and potentially native types (if we decide to implement 
+immutability for those).
+
+With the container example we get this, with a sketch of a `+normalize` marker which
+normalizes empty and null to undefined:
+
+```
+// No single field in existing containers can be changed, added or deleted.
+// New containers can be added, existing containers can be removed. Containers
+// are identified by its name.
```
jpbetz: s/its name/their names ?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+Hence, it looks sensible to split normalization from immutability equality, and keep a strict 
+deep-equal equality even for protobuf, and potentially native types (if we decide to implement 
+immutability for those).
+
+With the container example we get this, with a sketch of a `+normalize` marker which
+normalizes empty and null to undefined:
+
+```
+// No single field in existing containers can be changed, added or deleted.
+// New containers can be added, existing containers can be removed. Containers
```
jpbetz: The "existing containers can be removed" case is interesting. We're not supporting that one with this proposal are we?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+// are identified by its name.
+// +immutable
+// +normalize=undefined
+Containers []Container `json:containers,omitempty` `protobuf:"bytes,2,opt,name=containers"
+```
+
+For fields which carry no `omitempty`, we could allow more advanced normalization
+modes which replicate the Golang serialization behaviour. Tooling like openapi-gen
+and the CRD validation could verify that the normalization specification matches 
+that of Golang.
```
jpbetz: By matches, we mean that when `+normalize=undefined` is set then `json:omitempty` must also be set?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,313 @@
...
+- the semantics of immutability must be driven by:
+  - we do not break/change old CRD persistence semantics.
+  - the user-observed equality used for immutability checks must match the equality on
+    persisted objects. I.e. if `StorageRoundtrip(object)` is the object returned by a
+    create or update call, then we want that `StorageRoundtrip(a) == StorageRoundtrip(b)`
+    is the equality used for comparing `a` and `b`. If that check fails, a request
+    is rejected because of immutability conflict.
+- the mechanism must extend to
+  - the addition of protobuf or other encodings which unify values like empty, null and undefined.
+  - the use for existing native types in order to replace complex validation code with a simple declarative marker on the types.
```
jpbetz: @erictune mentioned a requirement that this proposal supports with but might be worth calling in the goals here: create requests that set immutable fields must be idempotent. It highlights the importance of handling equality correctly.

sttts: do you have an example?

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: There are also real maps, e.g. annotations and labels. We avoid them most of the times.

It is probably worth to add a section about `x-kubernetes-list-type` and `x-kubernetes-list-map-keys` and how these interact with immutability.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: identifies and unifies are the same in my understanding and this context. They are the same thing actually, i.e. cannot be differentiated. I can rephrase with unifies if you think that's clearer.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: I think I have to update this. It's a bit tricky with kubebuilder's semantics. `+immutable` should recursively apply to the array, i.e. removal and addition are forbidden.

sttts: fixed. Had copied a wrong example here.

keps/sig-api-machinery/20190603-immutable-fields.md
jpbetz: It threw me, not sure how others will read it. But might switch to unify to be safe.

sttts: fixed

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: yes, that omitempty and the normalization match. Also it might be a pointer type which preserves null vs. omitempty swallows nulls into undefined.

keps/sig-api-machinery/20190603-immutable-fields.md
jpbetz: Here's the original comment: https://github.com/kubernetes/enhancements/pull/1099#discussion_r296360894

It's got an example in it.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+- to help API developers by verifying these assertions automatically
+- and to publish this information via OpenAPI.
+
+## Motivation
+
+There are resources in Kubernetes which have immutable fields by design,
+i.e. after creation of an object, those fields cannot be mutated anymore. E.g. a
+pod's specification is mostly unchangeable once it is created. To change the
+pod, it must be deleted, recreated and rescheduled. Users want to implement the
+same kind of read-only semantics for CustomResources, for example:
```
enxebre: We have this use case in our CRD driven machine API implementation, this will definitely be handy.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+        type: object
+```
+
+At the root-level and inside `.metadata` of an object, `x-kubernetes-immutable` is forbidden.
+
+### OpenAPI extension `x-kubernetes-immutable-keys`
+
+The topology of arrays is specified via `x-kubernetes-list-type` and `x-kubernetes-list-map-keys`. We currently support the `atomic`, `set` and `map` types. These interact with immutability in the following way:
+
+- **Non-recursive** immutability: For an array or a map, one can not remove or add new items/values, but existing items/values can be modified.
```
apelisse: Maybe explain what this means for sets?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+      someMap:
+        type: object
+        x-kubernetes-immutable-keys: true # allowed with additionalProperties
+        additionalProperties:
+          type: string
+    ```
+    with `x-kubernetes-immutable-keys` allowed due to `additionalProperties` and mutually
+    exclusive with `x-kubernetes-immutable` on the same node.
+    
+  - For arrays of list-type `map` we get the same behaviour, but enforce that the map-keys
```
apelisse: note: we typically call these "associative lists"

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+    properties:
+      someArray:
+        type: array
+        x-kubernetes-list-type: map
+        x-kubernetes-list-map-keys: ["name"]
+        items:
+          properties:
+            name:
+              type: string
+              x-kubernetes-immutable: true
```
apelisse: I'm not sure I understand what this means

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+Name string
+
+// The list of containers can not change AT ALL after creation.
+// No single field in existing containers can be changed, added or deleted,
+// no new containers can be added, no existing container can be removed.
+// +immutable
+Containers []Containers
+```
+
+The `x-kubernetes-immutable-keys: true` vendor extension is expressed via `// +immutable=keys` in Golang types:
```
apelisse: My understanding was that you can have both immutable and immutable keys at the same time,

would you have to do:
```
// +immutable
// +immutable=keys
```

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: done

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: done

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: We can move these up to the `items` level. Both are equivalent.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: We choose to not allow both at the same time. The former is stronger as the latter.

Note: an `// +marker` on a slice **does not** apply to the items, but on the slice itself. E.g. `// +nullable` allows to set `null` for a slice, but not `[null, null]` as value. Here `// +immutable`'s semantics is recursive, so `// +immutable=keys` is subsumed by `// +immutable` for the slice.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: actually, there is a slight difference between both:
```
type: object
nullable: true
properties:
  x:
    type: string
    x-kubernetes-immutable: true
```
allows a change from `{x: something}` to `null` or undefined (when used as a field value), but

```
type: object
x-kubernetes-immutable: true
nullable: true
properties:
  x:
    type: string
```

does not allow that.

Without `nullable` and as array items, both coincide.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+        type: object
+        x-kubernetes-immutable-keys: true # allowed with additionalProperties
+        additionalProperties:
+          type: string
+    ```
+    with `x-kubernetes-immutable-keys` allowed due to `additionalProperties` and mutually
+    exclusive with `x-kubernetes-immutable` on the same node.
+    
+  - For arrays of list-type `map` (associative lists) we get the same behaviour, but enforce that the map-keys
+    are immutable as well (via CRD validation):
```
enxebre: > but enforce that the map-keys are immutable as well

isn't this true for the case above as well?

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+- the mechanism must extend to
+  - the addition of protobuf or other encodings which unify values like empty, null and undefined.
+  - the use for existing native types in order to replace complex validation code with a simple declarative marker on the types.
+  - the restriction of the equality to only map keys, but not their values.
+  - the allowance of addition and/or deletion of map keys.
+  - the allowance of addition and/or deletion of keys in array of list-type `map`.  
+
+### Non-Goals
+
+- The mechanism must be extensible to native types, but its implementation is optional.
```
apelisse: Not sure I agree with this.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,462 @@
...
+    properties:
+      bar:
+        type: object
+```
+
+At the root-level and inside `.metadata` of an object, `x-kubernetes-immutable` is forbidden.
+
+### OpenAPI extension `x-kubernetes-immutable-keys`
+
+The topology of arrays is specified via `x-kubernetes-list-type` and `x-kubernetes-list-map-keys`. We currently support the `atomic`, `set` and `map` types. These interact with immutability in the following way:
```
apelisse: technically, we can also have "atomic" maps so you might want to consider that.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: The implementation for CRDs can come first, and native types later. This statement says that. For native types we have no immediate need, for CRDs we have.

apelisse: OK

sttts: And the whole idea around normalization exists in order to conceptionally cover native types and CRDs with protobuf as well in a coherent way. This I think is most important. Implementations we can change and extend any time.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: No, in the upper case only the keys are immutable, i.e. the set of keys. But values can change.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: done. In theory we could allow key-immutability for those atomic maps, but as that would contradict the meaning that an atomic map is not a set, but one value. Changing one bit changes the whole value.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,539 @@
...
+// The name can not be changed after creation.
+// +immutable
+Name string
+
+// The list of containers can not change AT ALL after creation.
+// No single field in existing containers can be changed, added or deleted,
+// no new containers can be added, no existing container can be removed.
+// +immutable
+Containers []Containers
+```
```
erictune: Suggestion:  Recommend a marker syntax for maps and map-slices in Go.  
Possible syntax:

```
// Okay to change any way
A map[string]string

// Can't touch this.
// +immutable
B map[string]string 

// Allowed addition and deletion.
// +immutable-map-values  <--- not sure how to express this in Go.
C map[string]string 

// Allowed addition and deletion.
// +listType=map
// +listMapKey=name
// +immutable-map-values
D []SomeType

// Allowed existing value modification.
// +immutable-keys
E map[string]string 

// Allowed existing value modification.
// +listType=map
// +listMapKey=name
// +immutable-keys
F []SomeType
```
// Can't add ports to running pod but can  existing options but can add new ones.
// +immutable=values  <--- not sure how to express this in Go.
Options map[string]string

sttts: cc @DirectXMan12

sttts: @erictune there is a discussion (for quite some time) about how to express properties of items and map values. Am not sure this is so different here, at least it is overlapping.

I'd like to have general `+kubebuilder:validation:items:<some-marker>` and `+kubebuilder:validation:values:<some-marker>` syntax.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,539 @@
...
+### Future outline: protobuf
+
+Protobuf encoding unifies unset, empty and null values for slices and maps. Those three cases 
+cannot be differentiated an lead to the same on-the-wire value. In native types
+we use a semantic equality to factor those differences out if JSON is used on the wire, but
+protobuf is used for storage. In CRs decoding and encoding is always faithful with the 
+traditional JSON format used on-the-wire and for storage.
+
+When adding protobuf encoding to CRDs in the future, we have to (without major, non-standard
+efforts) identify unset, empty and null for CRs as well. This leads to the idea to use a less
```
erictune: There is an alternative.  

We could  generate proto files and proto descriptors from the `types.go`.  A `map[string]string` could turn into a `map<StringOrEmptyOrNull, StringOrEmptyOrNull>`.

sttts: How unusual would such an encoding be (outside of kube-proto) ?

We could derive the right proto schema from the specified normalization behaviour. But then we come to the question of compatibility if you change normalization at a later point.

We should start early to flesh out proto targetting 1.18. There seem to be so many details to get right.

erictune: Proto's standard libraries define something like this which is used for proto's JSON support:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/struct.proto

We'd basically be defining something similar.

deads2k:
I got my comments from slack and in the PR addressed.

/approve
/hold

holding until monday PM for @lavalamp to have a pass if he wishes.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Ic, just another sum type (oneof). The question will be whether we then model something in Proto that was just an accident by our use of JSON initially. With JSON we have no choice to get null and undefined as separate values, with proto we could fix our accident.

apelisse: > With JSON we have no choice to get null and undefined as separate values, with proto we could fix our accident.

We need to find a good balance between the "API semantics make sense", "APIs have too many knobs" and "consistency between CRD and built-in types" trade-off.

sttts: Some further thoughts for proto and normalization:

- We already have `nullable` to explicitly allow nulls.
- With native types we have `omitempty` which maps null and empty to undefined.

1. In the case of non-nullable, we only have two values. Without normalization we get:
   ```
   undefined -> undefined
   empty -> empty
   ```
   By using a default, we can get:
   ```
   undefined -> empty
   empty -> empty
   ```
   Omitempty would give us:
   ```
   undefined -> undefined
   empty -> undefined
   ```
   Hence, if we had `x-kubernetes-omitempty: true` we would have covered all variants we know 
   today with native types.

2. With nullable and no normalization we get
   ```
   undefined -> undefined
   null -> null
   empty -> empty
   ```
   With defaults, we are able to express:
   ```
   undefined -> null
   null -> null
   empty -> empty
   ```
   With omitempty, we would get:
   ```
   undefined -> undefined
   null -> undefined
   empty -> undefined
   ```
   What we miss is:
   ```
   undefined -> empty
   null -> empty
   empty -> empty
   ```
   I question that we really need this at all. If we identity empty in proto with undefined in JSON, `x-kubernetes-omitempty: true` would be enough.

In other words, we maybe only need `x-kubernetes-omitempty: true` and cover all interesting cases.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+    schemas are followed.
+  - rejection of `x-kubernetes-immutable: true` and `x-kubernetes-immutable-keys: true` for non-v1 CRDs
+- e2e and conformance tests that
+  - immutability is followed during updates, patches and server-side-apply. 
+   
+### Graduation Criteria
+
+Because of the very limited risk of the additional immutability check and our
+experience with defaulting getting no additional soak time by the community
+we propose to start directly as beta.
```
liggitt: This results in potential data loss during HA API server upgrade or rollback (a 1.16 API server will drop the proposed fields). We need the n-1 API server to be capable of preserving the data in the field before we enable/allow it by default, which implies a 1.17 alpha and a 1.18 beta.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+
+## Motivation
+
+There are resources in Kubernetes which have immutable fields by design,
+i.e. after creation of an object, those fields cannot be mutated anymore. E.g. a
+pod's specification is mostly unchangeable once it is created. To change the
+pod, it must be deleted, recreated and rescheduled. Users want to implement the
+same kind of read-only semantics for CustomResources, for example:
+https://github.com/kubernetes/kubernetes/issues/65973. Today this is only possible
+with the unreasonable development overhead of a webhook.
```
liggitt: I have a vague memory of an immutable-once-set semantic we wanted at one point (would enable an async allocator to set a field once via the API, after which it would be locked). Was that discarded or deferred? Not having that means any immutable field must be set synchronously on create by the object originator or a webhook.

liggitt:
One question about immutable-once-set, and one comment about the direct-to-beta proposal. LGTM otherwise.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Am not aware of that old discussion, but it definitly makes sense to flesh that out. 

It sounds like you want a way to declare a struct not completely immutable, but stop it from being deleted, and certain fields to be set-once-only:
- This is more than marking fields immutable, because you could still set them to undefined. 
- It is less than marking the whole struct's keys immutable because not all fields might be set only.
In other words, this is some weaker form of `x-kubernetes-keys` on the hosting struct, let's call it `x-kubernetes-keys-set-once`:

Example:

```yaml
apiVersion: setonce/v1
kind: Once
spec:
   once: true
   name: foo
```

with

```yaml
type: object
x-kubernetes-immutable-keys: true
properties:
  spec:
    type: object
    x-kubernetes-keys-set-once: ["once","name"]
    properties:
      once:
        type: boolean
      name:
        type: string
        x-kubernetes-immutable: true
```

Here, `name` can only be set to a fixed value once, and never changed anymore. `once` can be set once (i.e. from undefined to some value), but the actual value can be changed any time.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+With the container example we get this, with a sketch of a `+normalize` marker which
+normalizes empty and null to undefined:
+
+```
+// The list of containers can not change AT ALL after creation, modulo
+// empty, null and undefined.
+// No single field in existing containers can be changed, added or deleted,
+// no new containers can be added, no existing container can be removed.
+// +immutable
+// +normalize=undefined
```
lavalamp: I really strongly object to letting our deep-equal & proto round trip behavior leak into our api design specification. API authors should not have to understand go's peculiar deep equal behavior.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+3. defaulting
+4. conversion to internal
+5. if immutability is enabled for the resource:
+   1. marshalling into JSON in-memory
+   2. normalization creating a copy with shared data-structures
+   3. strict immutability check against the old object, coming from proto assuming it is normalized.
+
+### Future outline: protobuf
+
+Protobuf encoding unifies unset, empty and null values for slices and maps. Those three cases
```
lavalamp: We should never use the difference between unset, empty, null to drive a semantic behavior difference. We have in the past, and it's been a mistake every time.

In addition to users not expecting that, it breaks apply, because in apply, we need to interpret unset/empty/null as meaning "I have no opinion".

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+strict equality in the immutabiltiy checks with that case in mind. But protobuf encoding in
+native types also has a normalization effect, namely posted JSON object are normalized through
+the encoding to protobuf when writing to etcd. 
+
+Hence, it looks sensible to split normalization from immutability equality, and keep a strict 
+deep-equal equality even for protobuf, and potentially native types (if we decide to implement 
+immutability for those).
+
+With the container example we get this, with a sketch of a `+normalize` marker which
+normalizes empty and null to undefined:
```
lavalamp: I think we should always treat empty/null/undefined as having the same meaning. So, I don't agree with letting API authors choose one.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+This process is meant to happen right before the update validation and after
+mutating, but before validating webhooks, and only run on updates. This will allow us to
+keep the exact same behavior while removing the validation code that checks the
+immutability of fields.
+
+![Decoding steps which must apply defaults](20191001-crd-immutability-pipeline.png)
+
+### Risks and Mitigations
+
+- immutable metadata would break API machinery. We forbid the `x-kubernetes-immutable`
```
lavalamp: We could copy metadata from the built-in one. At first I thought maybe it'd be OK for users to add fields to it, but actually it's not safe--it would make it possible for a collision in the future if we add to metadata a field of the same name.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+  required to support protobuf for CRs. But this KEP does not aim at defining these
+  and hence defining a custom equality which is compatible with normalization.
+- The mechanism is not supposed to allow different orders in lists to be considered equal.
+
+## Proposal
+
+We propose
+
+1. adding boolean vendor extensions to CRD OpenAPI schemas named `x-kubernetes-immutable` and `x-kubernetes-immutable-keys` with `true` as the only valid value.
+2. do **strict deep-equal** comparison of those fields marked as immutable during
```
lavalamp: Now that I have read the bottom, I suggest you revise this to consider unset, empty, and 0-length items to be the same.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+    type: object
+    x-kubernetes-immutable: true
+    properties:
+      bar:
+        type: object
+```
+
+At the root-level and inside `.metadata` of an object, `x-kubernetes-immutable` is forbidden.
+
+### OpenAPI extension `x-kubernetes-immutable-keys`
```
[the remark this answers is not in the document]
lavalamp: Another (more intuitive?) way to support this behavior is to have an override tag (e.g. +always-mutable, +permit-mutations-anyway, +transient-values, +ephemeral-- something like that) on the fields that are actually mutable. Something similar is mentioned in the alternatives at the bottom but I don't really buy the reasoning.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+N/A
+
+## Alternative Considered
+
+- OpenAPI has a notion of `readOnly`. This is meant to restrict fields to be set
+  only in responses, not in a request payload. This does not match our 
+  `never-change-after-creation` semantics.
+- Allowing `false` as value for `x-kubernetes-immutable: false` was considered to
+  disable immutability imposed by a parent node. This complicates the semantics
+  considerably and can be expressed with a combination of `x-kubernetes-immutable-keys`
```
lavalamp: This sentence contradicts itself. If you can already express it, the system can mechanically convert one expression to the other, the semantics must already be the same.

keps/sig-api-machinery/20190603-immutable-fields.md
apelisse: @liggitt Did you have an actual use-case in mind? I can't remember talking about that either, and I'd really like to keep the number of knobs low if we can.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+## Implementation History
+
+N/A
+
+## Alternative Considered
+
+- OpenAPI has a notion of `readOnly`. This is meant to restrict fields to be set
+  only in responses, not in a request payload. This does not match our 
+  `never-change-after-creation` semantics.
+- Allowing `false` as value for `x-kubernetes-immutable: false` was considered to
```
lavalamp: I think this would be much less surprising for humans making specs.

keps/sig-api-machinery/20190603-immutable-fields.md
liggitt: > Did you have an actual use-case in mind?

We use this capability in built-in resources today:
* scheduler setting spec.nodeName (and it being immutable after that point)
* populating clusterIP in services
* append-only lists (like ephemeralContainers)
* remove-only lists (like finalizers that may only shrink after deletion)

Talked through a lot of scenarios with @sttts, and I think the API guideline to [think twice about boolean use](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types) probably applies here. Modeling restrictions on key/value immutability as an enum rather than a boolean allows more use cases to avoid requiring a validating webhook. Something like:

```
x-kubernetes-mutability: Immutable | AddOnly | RemoveOnly
x-kubernetes-key-mutability: Immutable | AddOnly | RemoveOnly
```

lavalamp: > scheduler setting spec.nodeName (and it being immutable after that point)

This is write-once, I agree (but not sure we need a general mechanism for this).

> populating clusterIP in services

Is this actually write-once?

> append-only lists (like ephemeralContainers)

append-only isn't quite the same as write-once.

> remove-only lists (like finalizers that may only shrink after deletion)

IMO the "after deletion" qualifier definitely means that there's not going to be a general mechanism for this.

liggitt: >> scheduler setting spec.nodeName (and it being immutable after that point)
>
> This is write-once, I agree (but not sure we need a general mechanism for this).
>
>> populating clusterIP in services
>
> Is this actually write-once?

We set clusterIPs in-process in the service registry implementation, and do not allow it to change on update. To replicate these interactions without write-once capabilities, a custom resource would either have to:
* use a mutating webhook on create (which would require allocation to be really fast) and built-in immutability on the CR schema
* use a controller to allocate asynchronously and a validating webhook to prevent changing/unsetting a value

If we can allow custom resources to express invariants without hooking up dynamic logic, that seems valuable.

> append-only isn't quite the same as write-once

You can accomplish append-only with a combination of `key-mutability: AddOnly` on the container, and `mutability: Immutable` on the individual items.

> IMO the "after deletion" qualifier definitely means that there's not going to be a general mechanism for this.

I agree the AddOnly case is stronger than RemoveOnly.

lavalamp: > If we can allow custom resources to express invariants without hooking up dynamic logic, that seems valuable.

Yes, the question isn't *if* it is valuable, the question is how valuable, to how many users, and at what cost, where cost is initial implementation cost (amortized) + ongoing maintenance cost + ongoing API complexity cost to users and us

keps/sig-api-machinery/20190603-immutable-fields.md
lavalamp: I object to going straight to beta, but on the grounds that I object to adding the immutable-keys tag *at all* :)

sttts: changed to go to alpha first.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+same kind of read-only semantics for CustomResources, for example:
+https://github.com/kubernetes/kubernetes/issues/65973. Today this is only possible
+with the unreasonable development overhead of a webhook.
+
+### Goals
+
+- extend the CRD API to be able to specify immutability for fields.
+- publish the immutability field of CRDs via OpenAPI as vendor extension.
+- verify immutability on CR update and patch requests.
+- propose a source code marker to be consumed by kubebuilder and openapi-gen.
```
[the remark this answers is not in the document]
sttts: yes, openapi-gen will produce the immutable tags as vendor extensions in the go-openapi definition we compile in. SSA will get them through the schema passed into the handlers.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+- the semantics of immutability must be driven by:
+  - we do not break/change old CRD persistence semantics.
+  - the user-observed equality used for immutability checks must match the equality on
+    persisted objects. I.e. if `StorageRoundtrip(object)` is the object returned by a
+    create or update call, then we want that `StorageRoundtrip(a) == StorageRoundtrip(b)`
+    is the equality used for comparing `a` and `b` **only modulo the order in `x-kubernetes-list-type: set` arrays**. If that check fails, a request
```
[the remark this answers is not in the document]
sttts: clarified.

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+    is rejected because of immutability conflict.
+- the mechanism must extend to
+  - the addition of protobuf or other encodings which unify values like empty, null and undefined.
+  - the use for existing native types in order to replace complex validation code with a simple declarative marker on the types.
+  - the restriction of the equality to only map keys, but not their values.
+  - the allowance of addition and/or deletion of map keys.
```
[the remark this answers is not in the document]
sttts: added @liggitt examples

keps/sig-api-machinery/20190603-immutable-fields.md
```diff
@@ -0,0 +1,464 @@
...
+  - the allowance of addition and/or deletion of keys in array of list-type `map`.
```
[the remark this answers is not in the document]
sttts: added @liggitt's examples

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Referring to the Sig-API-Machinery meeting of Oct 23. Added a summary in the paragraph about the equality choice.

keps/sig-api-machinery/20190603-immutable-fields.md
sttts: Today we forbid any restriction to anything else than name and generated name. This is just a clarification that we don't change this with immutability.

sttts: clarified.

PR #1273, opened 2019-10-03
update CSI topology kep for GA

msau42:
Added more details around upgrade and version skew considerations + testing.

### DIFF
--- a/keps/sig-storage/20190124-csi-topology.md
+++ b/keps/sig-storage/20190124-csi-topology.md
@@ -13,7 +13,7 @@ approvers:
   - "@saad-ali"
 editor: TBD
 creation-date: 2019-01-24
-last-updated: 2019-01-24
+last-updated: 2019-10-14
 status: implementable
 see-also:
   - n/a
@@ -31,10 +31,15 @@ CSI Volume Topology
 
 <!-- toc -->
 - [Summary](#summary)
+- [Upgrade/Downgrade Strategy](#upgradedowngrade-strategy)
+  - [Deprecations](#deprecations)
+- [Version Skew Strategy](#version-skew-strategy)
 - [Test Plan](#test-plan)
+  - [GA testing](#ga-testing)
 - [Graduation Criteria](#graduation-criteria)
   - [Alpha-&gt;Beta](#alpha-beta)
   - [Beta-&gt;GA](#beta-ga)
+- [Implementation History](#implementation-history)
 <!-- /toc -->
 
 ## Summary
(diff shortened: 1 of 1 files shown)


keps/sig-storage/20190124-csi-topology.md
```diff
@@ -53,6 +92,15 @@ The rest of the document includes required information missing from the original
   * (Negative) Volume provisioning with delayed volume binding and pod zone missing from AllowedTopologies.
 Initially topology tests are run against a single CSI driver. As the CSI test suites become modularized they will run against arbitrary CSI drivers.
 
+### GA testing
+* Manual e2e testing for upgrade and version skew scenarios.
+  * An older driver that only understands v1beta1 should continue to work when the cluster is
```
jsafrane: `driver` -> `sidecar`? Drivers do not know anything about CSINodes.

keps/sig-storage/20190124-csi-topology.md
```diff
@@ -43,6 +48,40 @@ This KEP is written after the original design doc has been approved and implemen
 
 The rest of the document includes required information missing from the original design document: test plan and graduation criteria.
 
+## Upgrade/Downgrade Strategy
+When moving to GA in K8s 1.17, the storage/v1 CSINode object will be introduced. 1.17
+kubelets will immediately start creating v1 CSINode objects. However, the master
+controllers (scheduler and attach-detach controller) will continue to use the
+v1beta1 CSINode object for one more release. This is to accommodate HA master
```
jsafrane: https://kubernetes.io/docs/setup/release/version-skew-policy/ says that controller-manager cannot be ever newer than API server. So controller-manager 1.17 can expect that it has API server 1.17 or newer.

Do I miss anything?

msau42: That's a good point. In the past, this was how beta->v1 changes were handled because the idea is that kube-controller-manager could be talking to the API server through a load balancer instead of localhost.  But the version skew policy seems to not support that kind of config. cc @liggitt for clarification.

keps/sig-storage/20190124-csi-topology.md
msau42: fixed

jsafrane:
lgtm except the version skew

keps/sig-storage/20190124-csi-topology.md
liggitt: > So controller-manager 1.17 can expect that it has API server 1.17 or newer.

You are correct. When switching to use a GA API, that is reasonable for a controller to do.

Note that if a controller is already using a GA API (e.g. `v1` Events), it should not switch to require a less stable API (e.g. `events.k8s.io/v1beta1` Events). That has less to do with skew and more to do with enabling production clusters to run with only GA APIs enabled.

msau42: Thanks for the clarification. Then in that case, I should be able to change the controller logic to start using v1 (instead of v1beta) in the same release where v1 is introduced.

msau42: updated to clarify that in-tree controllers can be switched to use v1 objects.  However, sidecars still need to support both for a deprecation period since the sidecars can run on older releases.

PR #1274, opened 2019-10-03
Update node self-labeling timeline

liggitt:
Updating timeline since we didn't turn off node labeling on create until 1.16

/cc @mikedanese


liggitt:
ping @mikedanese

PR #1285, opened 2019-10-07
Mark Default Even Pods Spreading as Implementable

alculquicondor:
Also:

- Check fulfilled items
- Specify relationship to "Even Pods Spreading" graduation

### DIFF
--- a/keps/sig-scheduling/20190926-default-even-pod-spreading.md
+++ b/keps/sig-scheduling/20190926-default-even-pod-spreading.md
@@ -10,8 +10,8 @@ approvers:
   - "@ahg-g"
   - "@k82cn"
 creation-date: 2019-09-26
-last-updated: 2010-09-26
-status: provisional
+last-updated: 2010-10-07
+status: implementable
 see-also:
   - "/keps/sig-aaa/20190221-even-pods-spreading.md"
 ---
@@ -48,9 +48,9 @@ see-also:
 
 ## Release Signoff Checklist
 
-- [ ] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
-- [ ] KEP approvers have set the KEP status to `implementable`
-- [ ] Design details are appropriately documented
+- [x] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [x] KEP approvers have set the KEP status to `implementable`
+- [x] Design details are appropriately documented
 - [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
 - [ ] Graduation criteria is in place
 - [ ] "Implementation History" section is up-to-date for milestone
@@ -280,6 +280,9 @@ Alpha (v1.17):
 - [ ] Priority Implementation.
 - [ ] Predicate implementation.
 - [ ] Test cases mentioned in the [Test Plan](#test-plan).
+
+Default `topologySpreadConstaints` will graduate from Alpha along with
+[Even Pods Spreading](/keps/sig-scheduling/20190221-even-pods-spreading.md).
 
 ## Implementation History
 


keps/sig-scheduling/20190926-default-even-pod-spreading.md
```diff
@@ -281,6 +281,9 @@ Alpha (v1.17):
 - [ ] Predicate implementation.
 - [ ] Test cases mentioned in the [Test Plan](#test-plan).
 
+Default `topologySpreadConstaints` will graduate from Alpha along with
```
Huang-Wei: It's a little confusing to me. Are you trying to say (1) it will become Alpha in 1.17, and (2) it won't be promoted to Beta earlier than `EvenPodsSpread` feature?

alculquicondor: They would be promoted to beta together.

Huang-Wei: Then let's reword it to:

```
Default `topologySpreadConstaints` (this feature) will graduate from Alpha along with or later than `EvenPodsSpread` feature.
```

alculquicondor: I think we actually want to graduate them together. Having "Even Pods Spreading" graduate to beta without the Default implies that we have to maintain from then the support of `EvenPodSpreadPriority` and `SelectorSpreadingPriority` running together.
And we want to avoid that.

Huang-Wei: We _can_ promote them together. But stating "no early than" is _technically_ correct, and give us a buffer.

alculquicondor: What I'm saying is that they have to graduate together so that we can remove `SelectorSpreadingPriority`. WDYT @ahg-g ?

ahg-g: yes, they should graduate in lockstep, they should use the same feature flag in fact.

ahg-g: @Huang-Wei do you have any concerns with using the same flag for both?

Huang-Wei: @alculquicondor I rethink this a bit. Even if we promote them altogether to beta in some release, users can still choose to disable one and enable the other, right? In that case, it's still in conflict with `SelectorSpreadingPriority`. So in other words, `SelectorSpreadingPriority` still needs to be kept before the `EvenPodsSpread` and `DefaultTopologySpreadConstraints` both promote to GA. We can't remove `SelectorSpreadingPriority` in Beta phase.

@ahg-g Using the same flag for both seems to be a good idea. So that we don't need to consider the complex cases that `EvenPodsSpread` is enabled while `DefaultTopologySpreadConstraints` is disabled, or vice verse; we need to only consider they're both enabled or disabled. @alculquicondor to make our logic simpler, I think we should use one featuregate, WDYT?

ahg-g: > @alculquicondor I rethink this a bit. Even if we promote them altogether to beta in some release, users can still choose to disable one and enable the other, right? In that case, it's still in conflict with `SelectorSpreadingPriority`. So in other words, `SelectorSpreadingPriority` still needs to be kept before the `EvenPodsSpread` and `DefaultTopologySpreadConstraints` both promote to GA. We can't remove `SelectorSpreadingPriority` in Beta phase.

Yeah, I think it is implied that if the feature gate is disabled, then  SelectorSpreadingPriority will be executed.

alculquicondor: Sharing the feature gate was implied, but I just added that explicitly.

For the nuances of enabling the flag or having a policy that uses `SelectorSpreadingPriority`, see https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/20190926-default-even-pod-spreading.md#implementation-detailsnotesconstraints
Let me know if you want to suggest any changes.

Huang-Wei:
/lgtm

Thanks, @alculquicondor !

alculquicondor:
Thanks @Huang-Wei 

@k82cn any concerns?

k82cn:
> @k82cn any concerns?

Just went through the history, LGTM overall, thanks :)

alculquicondor:
Thanks Klaus.

/hold cancel

PR #1292, opened 2019-10-09
Update cheaper node heartbeats KEP

wojtek-t:
Ref https://github.com/kubernetes/enhancements/issues/589

### DIFF
--- a/keps/sig-node/0009-node-heartbeat.md
+++ b/keps/sig-node/0009-node-heartbeat.md
@@ -189,27 +189,29 @@ Once all the code changes are done, we will:
    but it’s fine to reduce its frequency (though we should continue writing it at
    least once per eviction period).
 
+Based on experiments, unsuccessful attempt to enable `Server Side Apply` due to
+performance issues it was causing, in v1.17 release, we will:
+- reduce frequency of NodeStatus update to once every 5 minutes instead of
+  1 minute (note that if any condition changes, the update will happen within
+  next 10 seconds anyway, so we are only talking about periodic heartbeats)
+- reduce frequency of updating NodeStatuses in NodeProblemDetector also to
+  5 minutes (to keep those two frequencies in sync) (similarly as Kubelet,
+  if any condition changes, update will happen within next 30s as it is now)
 
-To be considered:
+Obviously, in both cases, we are only talking about default values, which may
+be changed by cluster operators if needed.
+We don't want to increase it further, as we would like to keep at least one
+NodeStatus update at least every eviction period (which is defaulted to 5 minutes).
 
-1. We may consider reducing frequency of NodeStatus updates to once every 5 minutes
-   (instead of 1 minute). That would help with performance/scalability even more.
-   Caveats:
-   - NodeProblemDetector is currently updating (some) node conditions every 1 minute
-     (unconditionally, because lastHeartbeatTime always changes). To make reduction
-     of NodeStatus updates frequency really useful, we should also change NPD to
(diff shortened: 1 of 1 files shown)


keps/sig-node/0009-node-heartbeat.md
```diff
@@ -189,27 +189,23 @@ Once all the code changes are done, we will:
...
-     significantly increase the size of object Node.
-     Note that if we keep adding node conditions owned by other components, the
-     number of writes of Node object will go up. But that issue is separate from
-     that proposal.
+Based on experiments, unsuccessful attempt to enable `Server Side Apply` due to
+performance issues it was causing, in v1.17 release, we will:
+- reduce frequency of NodeStatus update to once every 5 minutes instead of
+  1 minute (note that if any condition changes, the update will happen within
+  next 10 seconds anyway, so we are only talking about periodic heartbeats)
+- reduce frequency of updating NodeStatuses in NodeProblemDetector also to
```
bgrant0607: I assume that if the status changes, NPD would also make the change promptly

keps/sig-node/0009-node-heartbeat.md
```diff
@@ -228,6 +224,20 @@ relying on frequent Node object updates. However, in non-managed solutions, cust
 will still be able to restore previous behavior by setting appropriate flag values.
 Thus, changing defaults to what we recommend is the path to go with.
 
+To evaluate the risk of further reduction of NodeStatus updates, I went through
+couple thousands of files mentioning `LastHeartbeatTime` and `k8s` on github to
+understand the potential of breaking some system. There are couple places that
+are incompatible with node leases, but they would already be broken by current
+1m frequency, e.g.:
```
bgrant0607: The Mesos Framework is deprecated, but would be nice to file issues on the other repos, if possible, to let them know about this KEP.

keps/sig-node/0009-node-heartbeat.md
wangzhen127: How do you plan to change NPD configuration? Use the new flag added in https://github.com/kubernetes/node-problem-detector/pull/356?

wojtek-t: Yes - clarified in the text.

wojtek-t: Kind-of implementation detail, but yes. Though we need to double check if it always works as we want (but that's doable, I'm focusing on having KEP approved first).

keps/sig-node/0009-node-heartbeat.md
wojtek-t: Filed (couldn't file the one for mesos, as the repo is already in read-only mode).

keps/sig-node/0009-node-heartbeat.md
```diff
@@ -189,27 +189,29 @@ Once all the code changes are done, we will:
...
-     that should be reflected in [NPD documentation][].
-   - Note that reducing frequency to 1 minute already gives us almost 6x improvement.
-     It seems more than enough for any foreseeable future assuming we won’t
-     significantly increase the size of object Node.
-     Note that if we keep adding node conditions owned by other components, the
-     number of writes of Node object will go up. But that issue is separate from
-     that proposal.
+Based on experiments, unsuccessful attempt to enable `Server Side Apply` due to
+performance issues it was causing, in v1.17 release, we will:
+- reduce frequency of NodeStatus update to once every 5 minutes instead of
```
derekwaynecarr: This is fine with me.

PR #1297, opened 2019-10-11
Update less serializations KEP based on final implementation

wojtek-t:
/assign @liggitt

### DIFF
--- a/keps/sig-api-machinery/20190329-less-object-serializations.md
+++ b/keps/sig-api-machinery/20190329-less-object-serializations.md
@@ -13,8 +13,8 @@ approvers:
   - "@deads2k"
   - "@lavalamp"
 creation-date: 2019-03-27
-last-updated: 2019-07-19
-status: implementable
+last-updated: 2019-10-10
+status: implemented
 see-also:
   - TODO
 replaces:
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/20190329-less-object-serializations.md
```diff
@@ -120,91 +120,106 @@ read/watched by different components.
...
+// SelfLink (which is set lately in the path) are ignored.
+type cachingObject struct {
+	lock sync.RWMutex
+
+	// Object for which serializations are cached.
+	object metaRuntimeInterface
+
+	// serializations is a cache containing object`s serializations.
+	// The value stored in atomic.Value is of type serializationsCache.
+	// The atomic.Value type is used to allow fast-path.
```
lavalamp: What type is stored in the atomic.Value?

lavalamp:
/approve

I won't hold this up for my question.

I know it's for a good cause but this really did not improve the understandability of our encoding stack, I'm kind of sad about that, it was already very difficult to understand :(

keps/sig-api-machinery/20190329-less-object-serializations.md
wojtek-t: The line above is saying "The value stored in atomic.Value is of type serializationsCache."

Apparently I forgot to copy the serializationsCache type from the PR:
```
// serializationsCache is a type for caching serialization results.
type serializationsCache map[runtime.Identifier]*serializationResult
```

PR #1312, opened 2019-10-15
Update TopologyManager KEP to bring inline with current implementation and update with 1.17 plans

klueska:


klueska:
Detailed tracking issue for the new beta features mentioned can be found here: https://github.com/kubernetes/kubernetes/issues/83479

PR #1329, opened 2019-10-21
[WIP] [provisional] Default-deny pod traffic

vllry:


kfox1111:
The current NetworkPolicy behavior is tied to the namespace. IE, if there is no NetworkPolicy object in the namespace, then it has X behavior. Bumping a version of the NetworkPolilcy object to a new api version wouldn't help with a nonexistent object. Its kind of a default behavior of a namespace.
Its like an implicit field in the Namespace object of: defaultNetworkPolicy: open.

So, one other solution to this would be to make a namespace v2 where the implicit defaultNetworkPolicy: closed? Then the user can pick what behavior they want by creating a v1 namespace or v2 namespace. Existing network plugins only need to look at what version namespace was used.

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+## Summary
+
+Make pod-to-pod communication in Kubernetes secure by default.
+Pods should require an explicit NetworkPolicy ingress configuration,
+in order to receive traffic.
+
+This proposal explicitly targets ingress traffic.
+Egress traffic is not _currently_ being targeted,
+not is not yet deemed out of scope.
```
rikatz: Typo here?

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+NetworkPolicy is one example.
+
+TODO pull examples
+
+### Goals
+
+* Restrict pod ingress by default on a cluster.
+
+### Non-Goals
+
```
rikatz: Maybe put also here that Egress is a non-goal by now?

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+What is out of scope for this KEP?
+Listing non-goals helps to focus discussion and make progress.
+
+## Proposal
+
+There are multiple options for design/implementation.
+
+Restricting network access requires a CNI driver that supports NetworkPolict.
```
rikatz: Typo: s/NetworkPolict/NetworkPolicy/

rikatz:
I’ve left some small reviews/typos there :)

About the behavior, I think the approach of a controller creating default-deny rules less disruptive than changing NetworkPolicy API. 

As I think the API needs some more changes (as maybe multiple actions to Log, Deny) this could really break a lot of stuffs. 

One suggestion: instead of dealing this with a “DefaultDeny” rule, this could be a new object as “NetworkProfile” (probably same idea as Calico) with the default rules of each newly created Namespace. 

As an example, the NetworkProfile could be something as “allow dns queries, allow kubernetes master queries, default deny” so this way the KEP could cover also Egress cases (just an idea, don’t wan’t to make this really complex). 

If this is still not the desired objective, this KEP could start at least with the idea of NetworkProfile dealing firstly only with Ingress :)

And then specify some sort of Namespace association with NetworkProfile. 

WDYT?

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+The most conceptually simple option is to change default NetworkPolicy behavior.
+Currently,
+NetworkPolicy allows all traffic to pods that are not selected by a NetworkPolicy. [1]
+
+If all pods are treated as isolated,
+users must explicitly create NetworkPolicy objects
+(either as needed, or by allowing all ingress and _explicitly_ enabling the existing behavior).
+This would potentially require a v2 API, due to changes in behavior (if not the API itself).
+
+### Option 2: Inject explicit deny-all NetworkPolicy objects
```
johscheuer: This would be more or less similar to the `ServiceAccount` AdmissionController -> https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#serviceaccount which creates the default `ServiceAccount`. The same behaviour could be used to create a default deny rule for a new Namespace.

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+If making changes to NetworkPolicy API is too undesirable,
+we could automatically apply NetworkPolicy objects to achieve the same effect.
+
+There is no precedent(?) or mechanism(?) for performing pre-creation actions
+(IE the opposite of finalizers).
+
+A simple controller could monitor for new namespaces,
+but it would be subject to race conditions,
+where new pods would be subject to the default-open behavior before the NetworkPolicy is created.
```
johscheuer: In the `ServiceAccount` AdmissionController there is currently also a race condition. Another race condition is the create of the `NetworkPolicy` object and the actual implementation in the CNI plugin (which can take more or less time depending on the size of your cluster).

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+Make pod-to-pod communication in Kubernetes secure by default.
+Pods should require an explicit NetworkPolicy ingress configuration,
```
szuecs: What about clusters that do not support network policy?

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+If making changes to NetworkPolicy API is too undesirable,
+we could automatically apply NetworkPolicy objects to achieve the same effect.
+
+There is no precedent(?) or mechanism(?) for performing pre-creation actions
+(IE the opposite of finalizers).
+
+A simple controller could monitor for new namespaces,
+but it would be subject to race conditions,
+where new pods would be subject to the default-open behavior before the NetworkPolicy is created.
+
```
johscheuer: IMHO Option 3 would be to implement `GlobalNetworkPolicy` similar to what Calico and Clilum already supports and would leave the freedom to operators to decide if they want the default deny rule or not (globally).

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+### Risks and Mitigations
+
+Changing network defaults will existing break user setups (EG apps with no NetworkPolicies),
```
johscheuer:
```suggestion
Changing network defaults will break existing user setups (EG apps with no NetworkPolicies),
```

keps/sig-network/20191020-default-deny-pod-traffic
johscheuer: Or don't require them

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+and may break existing clusters if designed or applied incorrectly.
+If this behavior is introduced to an existing cluster,
+it should be done in a permissive fashion
+(EG marking the cluster or specific namespaces as using "legacy mode").
+
+This change may frustrate users,
```
szuecs: yes!

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+as it introduces another step (configure NetworkPolicy) that was previously optional.
```
szuecs: No, because networkpolicy is having race condition since forever and there is no way of fixing it in the current state.
Pod ready++ was more a joke, than a fix and tear down is not even thought about

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+### Option 1: Change default NetworkPolicy behavior
+
+The most conceptually simple option is to change default NetworkPolicy behavior.
+Currently,
+NetworkPolicy allows all traffic to pods that are not selected by a NetworkPolicy. [1]
+
+If all pods are treated as isolated,
+users must explicitly create NetworkPolicy objects
+(either as needed, or by allowing all ingress and _explicitly_ enabling the existing behavior).
+This would potentially require a v2 API, due to changes in behavior (if not the API itself).
```
danwinship: This isn't a change in the behavior of NetworkPolicy though, it's a change in the behavior of lack-of-NetworkPolicies. Changing the API version of NetworkPolicy would not make that non-breaking.

The original design docs seem to have gotten lost in some website/repo reorg a few years back, but we originally specified that in the absence of any NetworkPolicies, the default behavior was undefined. It is definitely allowable for a network plugin to say that pods are isolated by default. (eg, see the pseudo-code [here](https://github.com/kubernetes/kubernetes/pull/24154/files#diff-d07dd9aa199f30808c8d5d44f2f6135fR218); this is in terms of the pre-v1 API where there was an explicit per-Namespace default-deny annotation, but you can see that the expectation was that the default might not always be allow.)

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+### Option 2: Inject explicit deny-all NetworkPolicy objects
+
+If making changes to NetworkPolicy API is too undesirable,
+we could automatically apply NetworkPolicy objects to achieve the same effect.
+
+There is no precedent(?) or mechanism(?) for performing pre-creation actions
+(IE the opposite of finalizers).
```
danwinship: OpenShift has the "[default project template](https://docs.openshift.com/container-platform/4.2/applications/projects/configuring-project-creation.html#modifying-template-for-new-projects_configuring-project-creation)", although it only applies to user-provisioned namespaces, not administrative namespaces.

The nice thing about this sort of approach is that (given the right design) the administrator could use this for more than just "default-deny". Eg, they could automatically add an "allow-from-monitoring-namespace" policy to every namespace, etc. And they could choose to add default egress policies as well.

keps/sig-network/20191020-default-deny-pod-traffic
danwinship: It also preserves the current confusing insecure behavior that deleting NetworkPolicies makes your namespace more secure... until you delete the last one, then it flips to totally unsecure.

keps/sig-network/20191020-default-deny-pod-traffic
danwinship: In general, Kubernetes currently leans towards making things easy for users who don't need to worry about security and assuming that people who do need to worry about security will... worry about security, and end up doing the right thing...

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+A simple controller could monitor for new namespaces,
+but it would be subject to race conditions,
+where new pods would be subject to the default-open behavior before the NetworkPolicy is created.
+
+### Risks and Mitigations
+
+Changing network defaults will existing break user setups (EG apps with no NetworkPolicies),
+and may break existing clusters if designed or applied incorrectly.
```
danwinship: and LOTS of tutorials

danwinship:
> IMHO Option 3 would be to implement `GlobalNetworkPolicy` similar to what Calico and Clilum already supports and would leave the freedom to operators to decide if they want the default deny rule or not (globally).

Something like this seems like the best combination of compatibility and simplicity to me; have some standardized and *easy* way for the cluster admin to opt in to isolated-by-default.

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+This proposal explicitly targets ingress traffic.
+Egress traffic is not _currently_ being targeted,
+not is not yet deemed out of scope.
+
+## Motivation
+
+Kubernetes and Kubernetes workloads are notoriously easy to leave insecure.
+Many users are not aware of relevant settings (like NetworkPolicy),
+or choose not to configure them due to friction and learning curves.
```
thockin: The fact that we have no default implementation also inhibits adoption.

Also because most examples on the web come with Deployment and Service YAML, but do not come with NetworkPolicy

thockin: The fact that we have no default implementation also inhibits adoptions.

Also because most examples on the web come with Deployment and Service YAML, but do not come with NetworkPolicy

keps/sig-network/20191020-default-deny-pod-traffic
thockin: This is an interesting statement.  CNI itself doesn't say anything about network policy.  There are policy engines that work with a variety of CNI drivers.

It is my position that we would do well to rethink network drivers top-to-bottom...

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+
+There are multiple options for design/implementation.
+
+Restricting network access requires a CNI driver that supports NetworkPolict.
+All options will require codifying NetworkPolicy support as a requirement,
+or creating a clear user experience when using a CNI driver that does not support NetworkPolicy.
+
+### Option 1: Change default NetworkPolicy behavior
+
+The most conceptually simple option is to change default NetworkPolicy behavior.
```
thockin: Note that an admin can already "fake" this by putting an "empty" policy in every Namespace

keps/sig-network/20191020-default-deny-pod-traffic
thockin: I'm all in favor of making things more correct by default, but @szuecs this is not a helpful comment.

kfox1111:
> My feeling is that, rather than tack this onto NetworkPolicy, we want a different API that sets the defaults and maybe even bounds for NetworkPolicy. NetworkPolicyPolicy or ClusterNetworkPolicy or something.

*SNIP*

> We've heard from many customers that they want default-deny but also "guard rails" on network policies, so users can't accidentally say "allow from all".

Yes, I have some multitenant clusters that would really benefit from this. I've been using NetworkPolicy to restrict one tenant from talking with another. But NetworkPolicy was never meant to be a black list, just a white list. Being able to enforce those rules per namespace, but allow users to still define NetworkPolicy themselves would be great.

keps/sig-network/20191020-default-deny-pod-traffic
vllry: Agreed.

keps/sig-network/20191020-default-deny-pod-traffic
vllry: Clusters that don't support NetworkPolicy would inherently "fail open" as is the case now. Longer term, I would personally like to see NetworkPolicy (or some alternative/successor) become required, but I should list that as an explicit non-goal.

Having an off-switch to expose old behaviour would be necessary, due to compatibly/migration concerns.

keps/sig-network/20191020-default-deny-pod-traffic
vllry: Exactly! I can't believe I blanked on the use of AdmissionControllers/AdmissionWebhooks for this kind of thing.

keps/sig-network/20191020-default-deny-pod-traffic
vllry: Thanks, that's some interesting context.

Given that, the option of creating policies automatically makes more sense than actually changing the default behaviour.

vllry:
Thanks everyone for the preliminary input! I'm going to think more about option 2 (whoops, I somehow forgot that admission controllers etc exist...) and auto-creating deny-all policies.

Some of this discussion slots in with the idea of the Tenant CRD, from the Multitenancy WG. The Tenant CRD is meant to act as a cluster-superadmin override for some functionality, which can't be contradicted by individual namespace admins. https://docs.google.com/document/d/1hpJX5O_siMmNGMvIHvz8Pm7XOjJLz5g57XWrgwWarFw/

keps/sig-network/20191020-default-deny-pod-traffic
danwinship: > E.g. user can edit any NetworkPolicy except the "default" instance?

oh, I wasn't suggesting that the "allow from monitoring namespace" would be not removable by the project admin. It was more that, if you know every namespace is going to *want* to have that policy, then the admin could cause it to be auto-created for them.

thockin:
Yeah, but they should not be able to remove ALL policies, because then they
flip back to default-pen mode.  In this particular model, anyway...

keps/sig-network/20191020-default-deny-pod-traffic
bjhaid: As someone who has had to write a custom controller (to create allow to dns, allow to web-proxy, allow to monitoring rules and deny everything in prod but allow all in other environments rules for every namespace) because we did not want to use a "third party" globalNetworkPolicy implementation, this is my personal preference.

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+## Summary
+
+Make pod-to-pod communication in Kubernetes secure by default.
```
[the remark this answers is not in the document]
lzang: This rule doesn't make the communication to be more secure as it simply blocks communication. Now user has to think about security and define the rules before they want to try anything. It is a pain to users who don't want to consider that. Also it means you have to give user the privilege to configure network policy for them to deploy their application. User can simply add a rule to allow all ingress and bypass the pre-defined rule. So why this would make pod-to-pod communication more secure?

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+Pods should require an explicit NetworkPolicy ingress configuration,
+in order to receive traffic.
+
+This proposal explicitly targets ingress traffic.
+Egress traffic is not _currently_ being targeted,
```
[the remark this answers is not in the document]
lzang: Why egress is a less concern than ingress here? A malicious pod can still do harm with egress allowed.

keps/sig-network/20191020-default-deny-pod-traffic
```diff
@@ -0,0 +1,270 @@
...
+What is out of scope for this KEP?
+Listing non-goals helps to focus discussion and make progress.
+
+## Proposal
+
+There are multiple options for design/implementation.
+
+Restricting network access requires a CNI driver that supports NetworkPolict.
+All options will require codifying NetworkPolicy support as a requirement,
+or creating a clear user experience when using a CNI driver that does not support NetworkPolicy.
```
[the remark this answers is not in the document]
lzang: It looks to me that this information should be put into some cluster object, to indicate whether there is a network policy controller registered or ingress controller registered, etc.

keps/sig-network/20191020-default-deny-pod-traffic
lzang: IMHO current network policy is written from application owner view because only they know what traffic is expected for their application and what is not. If we want to enforce anything else from a cluster administrator view, we will want to have some new policy that is cluster-wide and serve as a guardrail that is not easily broken through by app owners. Delegating the policy completely to app owners doesn't seem to be very secure for the whole system.

thockin:
Thinking through some use-cases, an admin-centric APY is almost certainly
going to be significantly more complicated than NetworkPolicy.  And I think
I am OK with that, since it is admin-focused, whereas NetworkPolicy is
primarily app-dev focused.

keps/sig-network/20191020-default-deny-pod-traffic
szuecs: @thockin , sorry, let's ignore the mentioned problem.

keps/sig-network/20191020-default-deny-pod-traffic
szuecs: I am fine, if the KEP would show what @thockin wrote as a comment:

> E.g. imagine a ClusterNetworkPolicy that selects namespaces by label and can express things like:
> *  default allow/deny
> * whitelist for all policies
> * blacklist for all policies

So the choice is up to the owner of the cluster, if the cluster supports networkPolicy or not and what the default should be (blacklist vs. whitelist).

keps/sig-network/20191020-default-deny-pod-traffic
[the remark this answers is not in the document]
lzang: If we want to add a rule at the beginning to secure the communication, another choice could be to add a rule saying that a pod only accepts traffic from pods from the same namespace (except for kube-system namespace where every pod needs to access dns pod so we may pre-install different rules there). This way, everything still works until they start to cross the namespace boundary.

abhiraut:
looks like introducing the ClusterNetworkPolicy object for admins has more yays than nays? @vllry do you plan to update the KEP or start a doc to document/design the CRD[s]?

pierrchen:
@vllry Do you think if there will be some benefits for others if you can leave some comments about the why the KEP is closed and what is the conclusion come out of this or path forward?

PR #1335, opened 2019-10-24
Keep CSI Raw block feature beta in 1.17

jsafrane:
The refactoring needed for separating NodeStage / NodePublish is quite large and we'd like to make sure it's well tested before declared GA.

It was discussed on today's SIG storage meeting.
cc @mkimuram @wongma7 @msau42


wongma7:
/lgtm

Yes, not much time left for testing. Need different kinds of drivers to help test: one that implements both stage/publish, or only implements publish, and/or MULTI_NODE, etc.

PR #1359, opened 2019-11-06
Fix invalid list type: associative->map

apelisse:


apelisse:
OpenAPI already uses map, I think we have to stick with it.

PR #1370, opened 2019-11-18
add KEP for cgroups v2 support

giuseppe:

### DIFF
--- a/keps/sig-node/20191118-cgroups-v2.md
+++ b/keps/sig-node/20191118-cgroups-v2.md
@@ -1,0 +1,241 @@
+---
+title: cgroups v2
+authors:
+  - "@giuseppe"
+owning-sig: sig-node
+participating-sigs:
+      - sig-architecture
+reviewers:
+  - "@yujuhong"
+  - "@dchen1107"
+  - "@derekwaynecarr"
+approvers:
+  - "@yujuhong"
+  - "@dchen1107"
+  - "@derekwaynecarr"
+editor: Giuseppe Scrivano
+creation-date: 2019-11-18
+last-updated: 2019-11-18
+status: implementable
+see-also:
+replaces:
+superseded-by:
+---
+
+# Cgroups v2
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+- [Goals](#goals)
+- [Non-Goals](#non-goals)
+- [User Stories](#user-stories)
+- [Implementation Details](#implementation-details)
+- [Proposal](#proposal)
+  - [Dependencies on OCI and container runtimes](#dependencies-on-oci-and-container-runtimes)
+  - [Current status of dependencies](#current-status-of-dependencies)
+- [Current cgroups usage and the equivalent in cgroups v2](#current-cgroups-usage-and-the-equivalent-in-cgroups-v2)
+  - [cgroup namespace](#cgroup-namespace)
+  - [Phase 1: Convert from cgroups v1 settings to v2](#phase-1-convert-from-cgroups-v1-settings-to-v2)
+  - [Phase 2: Use cgroups v2 throughout the stack](#phase-2-use-cgroups-v2-throughout-the-stack)
+- [Risk and Mitigations](#risk-and-mitigations)
+- [Graduation Criteria](#graduation-criteria)
+<!-- /toc -->
+
+## Summary
+
+A proposal to add support for cgroups v2 to kubernetes.
+
+## Motivation
+
+The new kernel cgroups v2 API was declared stable more than two years
+ago. Newer features in the kernel such as PSI depend upon cgroups
+v2. groups v1 will eventually become obsolete in favor of cgroups v2.
+Some distros are already using cgroups v2 by default, and that
+prevents Kubernetes from working as it is required to run with cgroups
+v1.
+
+## Goals
+
+This proposal aims to:
+
+*   Add support for cgroups v2 to the Kubelet
+
+## Non-Goals
+
+*	Expose new cgroup2-only features
+*	Dockershim
+*	Plugins support
+
+## User Stories
+
+*   The Kubelet can run on a host using either cgroups v1 or v2.
+*   Have features parity between cgroup v2 and v1.
+
+## Implementation Details
+
+## Proposal
+
+The proposal is to implement cgroups v2 in two different phases.
+
+The first phase ensures that any configuration file designed for
+cgroups v1 will continue to work on cgroups v2.
+
+The second phase requires changes through the entire stack, including
+the OCI runtime specifications.
+
+At startup the Kubelet detects what hierarchy the system is using.  It
+checks the file system type for `/sys/fs/cgroup` (the equivalent of
+`stat -f --format '%T' /sys/fs/cgroup`).  If the type is `cgroup2fs`
+then the Kubelet will use only cgroups v2 during all its execution.
+
+The current proposal doesn't aim at deprecating cgroup v1, that must
+still be supported through the stack.
+
+Device plugins that require v2 enablement are out of the scope for
+this proposal.
+
+### Dependencies on OCI and container runtimes
+
+In order to support features only available in cgroups v2 but not in
+cgroups v1, the OCI runtime specs must be changed.
+
+New features that are not present in cgroup v1 are out of the scope
+for this proposal.
+
+The dockershim implementation embedded in the Kubelet won't be
+supported on cgroup v2.
+
+### Current status of dependencies
+
+- CRI-O+crun: support cgroups v2
+
+- runc support for cgroups v2 is work in progress [current status](#current-cgroups-usage-and-the-equivalent-in-cgroups-v2)
+
+- containerd: [https://github.com/containerd/containerd/issues/3726](https://github.com/containerd/containerd/issues/3726)
+
+- Moby: [https://github.com/moby/moby/pull/40174](https://github.com/moby/moby/pull/40174)
+
+- OCI runtime spec: TODO
+
+- cAdvisor already supports cgroups v2 ([https://github.com/google/cadvisor/pull/2309](https://github.com/google/cadvisor/pull/2309))
+
+## Current cgroups usage and the equivalent in cgroups v2
+
+|Kubernetes cgroups v1|Kubernetes cgroups v2 behavior|
+|---|---|
+|CPU stats for Horizontal Pod Autoscaling|No .percpu cpuacct stats.|
+|CPU pinning based on integral cores|Cpuset controller available|
+|Memory limits|Not changed, different naming|
+|PIDs limits|Not changed, same naming|
+|hugetlb|Added to linux-next, targeting Linux 5.6|
+
+### cgroup namespace
+
+A cgroup namespace restricts the view on the cgroups.  When
+unshare(CLONE_NEWCGROUP) is done, the current cgroup the process
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
+Privileged pods will still use the host cgroup namespace so to have
+visibility on all the other cgroups.
+
+### Phase 1: Convert from cgroups v1 settings to v2
+
+We can convert the values passed by the k8s in cgroups v1 from to
+cgroups v2 so Kubernetes users don’t have to change what they specify
+in their manifests.
+
+crun has implemented the conversion as follows:
+
+**Memory controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| limit | memory.max | y = x ||
+| swap | memory.swap.max | y = x ||
+| reservation | memory.low | y = x ||
+
+**PIDs controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| limit | pids.max | y = x ||
+
+**CPU controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |  comment |
+|---|---|---|---|
+| shares | cpu.weight | y = (1 + ((x - 2) * 9999) / 262142) | convert from [2-262144] to [1-10000]|
+| period | cpu.max | y = x| period and quota are written together|
+| quota | cpu.max | y = x| period and quota are written together|
+
+**blkio controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| weight | io.bfq.weight | y = (1 + (x - 10) * 9999 / 990) | convert linearly from [10-1000] to [1-10000]|
+| weight_device | io.bfq.weight | y = (1 + (x - 10) * 9999 / 990) | convert linearly from [10-1000] to [1-10000]|
+|rbps|io.max|y=x||
+|wbps|io.max|y=x||
+|riops|io.max|y=x||
+|wiops|io.max|y=x||
+
+**cpuset controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| cpus | cpuset.cpus | y = x ||
+| mems | cpuset.mems | y = x ||
+
+**hugetlb controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| <PAGE_SIZE>.limit_in_bytes | hugetlb.<PAGE_SIZE>.max | y = x ||
+
+With this approach cAdvisor would have to read back values from
+cgroups v2 files (already done).
+
+Kubelet PR: [https://github.com/kubernetes/kubernetes/pull/85218](https://github.com/kubernetes/kubernetes/pull/85218)
+
+### Phase 2: Use cgroups v2 throughout the stack
+
+This option means that the values are written directly to cgroups v2
+by the runtime. The Kubelet doesn’t do any conversion when setting
+these values over the CRI.  We will need to add a cgroups v2 specific
+LinuxContainerResources to the CRI.
+
+This depends upon the container runtimes like runc and crun to be able
+to write cgroups v2 values directly.
+
+OCI will need support for cgroups v2 and CRI implementations will
+write to the cgroups v2 section of the new OCI runtime config.json.
+
+## Risk and Mitigations
+
+Some cgroups v1 features are not available with cgroups v2:
+
+- _cpuacct.usage_percpu_
+- network stats from cgroup
+
+Some cgroups v1 controllers such as _device_ and _net_cls_,
+_net_prio_ are not available with the new version.  The alternative to
+these controllers is to use eBPF.
+
+## Graduation Criteria
+
+- Alpha: Phase 1 completed and basic support for running Kubernetes on
+  a cgroups v2 host,  e2e tests coverage or have a plan for the
+  failing tests.
+  A good candidate for running cgroup v2 test is Fedora 31 that has
+  already switched to default to cgroup v2.
+
+- Beta: e2e tests coverage and performance testing.
+
+- GA: Assuming no negative user feedback based on production
+  experience, promote after 2 releases in beta.
+  *TBD* whether phase 2 must be implemented for GA.


keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+v1.
+
+## Goals
+
+This proposal aims to:
+
+*   Add support for cgroups v2 to cAdvisor ([https://github.com/google/cadvisor/pull/2309](https://github.com/google/cadvisor/pull/2309))
+*   Add support for cgroups v2 to the Kubelet
+
+## User Stories
```
AkihiroSuda: rootless?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+## Graduation Criteria
+
+- Alpha: Basic support for running Kubernetes on a cgroups v2 host.
+
+- Beta: e2e tests coverage, or have a plan for the failing tests.
+
+- GA: Assuming no negative user feedback based on production
+  experience, promote after 2 releases in beta.
+
+## Testing
```
AkihiroSuda: Minimum kernel version?
I think 5.2 (introduced freezer) is fine.

AkihiroSuda:
Do you think we should open a separate KEP for rootless?

giuseppe:
> Do you think we should open a separate KEP for rootless?

yes let's keep it separate from cgroups v2, which is already touching enough components :-)

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
+---
+title: cgroups v2
+authors:
+  - "@giuseppe"
+owning-sig: sig-node
+participating-sigs:
+      - sig-architecture
+reviewers:
+  - TBD
+approvers:
```
derekwaynecarr: Add yuju, Dawn, and myself.

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: I would keep that separate

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+## Goals
+
+This proposal aims to:
+
+*   Add support for cgroups v2 to cAdvisor ([https://github.com/google/cadvisor/pull/2309](https://github.com/google/cadvisor/pull/2309))
+*   Add support for cgroups v2 to the Kubelet
+
+## User Stories
+
+*   The Kubelet can run on a cgroups v2 host
```
derekwaynecarr: Clarify that kubelet will support v1 or v2

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+- Moby: [https://github.com/moby/moby/pull/40174](https://github.com/moby/moby/pull/40174)
+
+- OCI runtime spec: TODO
+
+## Current cgroups usage and the equivalent in cgroups v2
+
+|Kubernetes cgroups v1|Kubernetes cgroups v2 behavior|
+|---|---|
+|Killing processes using runc uses the freezer|There will be freezer for cgroups v2.|
```
derekwaynecarr: Kube doesn’t use freezer yet

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+|CPU stats for Horizontal Pod Autoscaling|No .percpu cpuacct stats.  Check cadvisor.|
+|CPU pinning based on integral cores|Cpuset controller available|
+|Memory limits|Not changed, different naming|
```
derekwaynecarr: Can you add detail on hugetlb and pids (both used today in kube)

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Proposal
```
derekwaynecarr: Can you add some detail on how kubelet startup may change?  For example, if kubelet detects it’s on a cgroup v2 host, it needs to update how it manages QOS and pod level cgroups.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+| cpus | cpuset.cpus | y = x ||
+| mems | cpuset.mems | y = x ||
+
+
+With this approach cAdvisor would have to read back values from
+cgroups v2 files (already done).
+
+Kubelet PR: [https://github.com/kubernetes/kubernetes/pull/85218](https://github.com/kubernetes/kubernetes/pull/85218)
+
+### Option 2: Use cgroups v2 throughout the stack
```
derekwaynecarr: For option 1, kubelet still has to write cgroups v2 for its interactions with cgroup.  Please add some detail around kubelet / cm package and eviction manager.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+The lack of the network controllers probably affects some CNI
+networking plugins.
+
+Investigate if any other components beside the Kubelet interact with
+cgroups and must be aware of it.
+
+## Graduation Criteria
+
+- Alpha: Basic support for running Kubernetes on a cgroups v2 host.
```
derekwaynecarr: Worth discussing if we should add test coverage on fedora as a good candidate upstream.

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: rootless KEP: https://github.com/kubernetes/enhancements/pull/1371

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,207 @@
...
+
+OCI will need support for cgroups v2 and CRI implementations will
+write to the cgroups v2 section of the new OCI runtime config.json.
+
+## Risk and Mitigations
+
+Some cgroups v1 features are not available with cgroups v2,
+e.g. _cpuacct.usage_percpu_.
+
+Some cgroups v1 controllers such as _device_ and _net_cls_, _net_prio
```
AkihiroSuda: nit: markdown format broken

keps/sig-node/20191118-cgroups-v2.md
giuseppe: I think I can drop this line at this point.  When I've started working on the KEP there was no freezer yet for cgroups v2 so the OCI runtime could not use it.

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: Unshare cgroupns by default?
https://github.com/containers/libpod/issues/4363

keps/sig-node/20191118-cgroups-v2.md
giuseppe: sadly hugetlb is still missing in the kernel for cgroups v2.

I've sent a patch to add support for hugetlb on cgroups v2: https://marc.info/?l=linux-cgroups&m=157437087608728&w=2

let's see how that goes...

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+## Risk and Mitigations
+
+Some cgroups v1 features are not available with cgroups v2,
+e.g. _cpuacct.usage_percpu_.
+
+Some cgroups v1 controllers such as _device_ and _net_cls_,
+_net_prio_ are not available with the new version.  The alternative to
+these controllers is to use eBPF.
+
+The lack of the network controllers probably affects some CNI
```
squeed: I don't currently know of any CNI plugins that use cgroups for network management. (My list is not exhaustive, of course...). Bandwidth limiting, if implemented, is done via a tc filter / qdisc for all of the plugins I'm aware of.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Moby: [https://github.com/moby/moby/pull/40174](https://github.com/moby/moby/pull/40174)
+
+- OCI runtime spec: TODO
+
+## Current cgroups usage and the equivalent in cgroups v2
+
+|Kubernetes cgroups v1|Kubernetes cgroups v2 behavior|
+|---|---|
+|CPU stats for Horizontal Pod Autoscaling|No .percpu cpuacct stats.  Check cadvisor.|
```
PatrickLang: Does this have to depend on CAdvisor? I thought we were trying to depend on the stats API instead of requiring cadvisor.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+v1.
+
+## Goals
+
+This proposal aims to:
+
+*   Add support for cgroups v2 to cAdvisor ([https://github.com/google/cadvisor/pull/2309](https://github.com/google/cadvisor/pull/2309))
+*   Add support for cgroups v2 to the Kubelet
+
+## User Stories
```
yujuhong: The user stories don't seem too user-focused. How are users going to use cgroup v2 and why? Is there additional benefit,  or is it just that users have nodes using v2 and they don't want to be aware of this?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+*   The Kubelet can run on a host using either cgroups v1 or v2
+*   Scheduler should be able to use cgroups v2
+
+## Implementation Details
+
+### Dependencies on OCI and container runtimes
+
+Cgroups v2 requires changes to the OCI runtime specs. Changes to the
+OCI specs are out of the scope for this proposal.
```
yujuhong: Does this mean OCI runtime spec already supports v2? If not, why don't we start by changing the OCI spec?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+This proposal aims to:
+
+*   Add support for cgroups v2 to cAdvisor ([https://github.com/google/cadvisor/pull/2309](https://github.com/google/cadvisor/pull/2309))
+*   Add support for cgroups v2 to the Kubelet
+
+## User Stories
+
+*   The Kubelet can run on a host using either cgroups v1 or v2
+*   Scheduler should be able to use cgroups v2
```
yujuhong: Why calling out scheduler? Does cgroup v2 require changing the resource representation in pod specs?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+This depends upon the container runtimes like runc and crun to be able
+to write cgroups v2 values directly.
+
+OCI will need support for cgroups v2 and CRI implementations will
+write to the cgroups v2 section of the new OCI runtime config.json.
+
+## Risk and Mitigations
+
+Some cgroups v1 features are not available with cgroups v2,
```
yujuhong: Would be good to have a list of unsupported features (that are used by kubernetes) in cgroups v2.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+The lack of the network controllers probably affects some CNI
+networking plugins.
+
+Investigate if any other components beside the Kubelet interact with
+cgroups and must be aware of it.
+
+## Graduation Criteria
+
+- Alpha: Basic support for running Kubernetes on a cgroups v2 host.
```
yujuhong: There should be some test coverage even for alpha, whether it's CRI validation suite, or node e2e tests.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Beta: e2e tests coverage, or have a plan for the failing tests.
+
+- GA: Assuming no negative user feedback based on production
+  experience, promote after 2 releases in beta.
+
+## Testing
+
```
yujuhong: Suggestion: add testing for performance

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+**cpuset controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| cpus | cpuset.cpus | y = x ||
+| mems | cpuset.mems | y = x ||
+
+
+With this approach cAdvisor would have to read back values from
+cgroups v2 files (already done).
```
yujuhong: So, with this option, CRI implementations will need to map the cgroup v1 fields to cgroup v2 values, right? 

Could you list what each layer needs to do for each option to make it clear?
 - user (pod spec)
 - kubelet
 - CRI implementation
 - OCI runtime

keps/sig-node/20191118-cgroups-v2.md
giuseppe: that will be done in the second phase where we will also expose/use cgroup v2 only features.
  
At this point I am not sure the KEP should account for that, as it is a much broader scope, and each functionality requires a new discussion.

In the first part we need to have feature parity with what we have now, so that Kubernetes can run on a cgroup v2 host.

keps/sig-node/20191118-cgroups-v2.md
giuseppe: is dropping CAdvisor tracked somewhere?  I've added only because I saw that it is still used.

Also the changes requires for CAdvisor were already merged upstream.  Another iteration might be required once hugetlb is supported though.

keps/sig-node/20191118-cgroups-v2.md
giuseppe: from the discussion we had, it seems there is no need to change the scheduler.  I am going to drop it

keps/sig-node/20191118-cgroups-v2.md
giuseppe: I'll move the tests coverage to the Alpha phase

keps/sig-node/20191118-cgroups-v2.md
giuseppe: the changes in the CRI implementation and OCI runtime might be implemented in a different way, deciding where to draw the line.
For CRI-O+crun, most of the logic is in the OCI runtime itself, but that is not the only way to achieve it.
For the Kubelet, the changes in this patch should be enough (at least until we have not hugetlb available): https://github.com/kubernetes/kubernetes/pull/85218

In the second phase though, when cgroup v2 is fully supported through the stack there is need to change both pod specs+CRI.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,204 @@
...
+in their manifests.
+
+crun has implemented the conversion as follows:
+
+**Memory controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |   comment |
+|---|---|---|---|
+| limit | memory.max | y = x ||
+| swap | memory.swap_max | y = x ||
```
AkihiroSuda: memory.swap.max

keps/sig-node/20191118-cgroups-v2.md
giuseppe: the hugetlb controller patch was accepted and it will hit linux-next in few days.

giuseppe:
> Do we want to unshare cgroupns for pod sandboxes as well as for containers? ([containerd/cri#1371](https://github.com/containerd/cri/pull/1371))

I think yes.  I'll update the KEP to specify it

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+then the Kubelet will use only cgroups v2 during all its execution.
+
+### cgroup namespace
+
+A cgroup namespace restricts the view on the cgroups.  When
+unshare(CLONE_NEWCGROUP) is done, the current cgroup the process
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
```
AkihiroSuda: Cgroupns is created for pod sandboxes as well as for each of the containers, and the pod sandbox cgroupns is NOT shared by the containers, right?

giuseppe: yes right, each container must have their own cgroupns

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: can you clarify why you feel the pod spec needs to change?  i see no major reason to change the pod spec or resource representation.

giuseppe: I was assuming we'd like to expose all/most of the cgroup v2 features.

e.g. the memory controller on cgroup v1 allows to configure:
- `memory.soft_limit_in_bytes`
- `memory.limit_in_bytes`

while on cgroup v2 we have:
- `memory.high`
- `memory.low`
- `memory.max`
- `memory.min`

but that is probably out of the scope as each new feature (if needed) must go through its own KEP?

giuseppe: @derekwaynecarr are future improvements based on what cgroup 2 offers out of scope for the current KEP?

giuseppe:
I've addressed the comments.  Is there any other issue left?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+### Dependencies on OCI and container runtimes
+
+Cgroups v2 requires changes to the OCI runtime specs. Changes to the
+OCI specs are out of the scope for this proposal.
+
+### Current status of dependencies
+
+- CRI-O+crun: support cgroups v2
+
+- runc support for cgroups v2 is work in progress
```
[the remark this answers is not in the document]
derekwaynecarr: ignore above comment, maybe just provide a link to the "current cgroups usage..." sectionl

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: very cool, i think we can resolve this section now.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+|CPU pinning based on integral cores|Cpuset controller available|
+|Memory limits|Not changed, different naming|
+|PIDs limits|Not changed, same naming|
+|hugetlb|Added to linux-next, targetting Linux 5.6|
+
+## Proposal
+
+The proposal is to implement cgroups v2 in two different phases.
+
+The first phase ensures that any configuration file designed for
```
derekwaynecarr: i think having the proposal text before the dependencies section would make this document read clearer.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+the OCI runtime specifications.
+
+At startup the Kubelet detects what hierarchy the system is using.  It
+checks the file system type for `/sys/fs/cgroup` (the equivalent of
+`stat -f --format '%T' /sys/fs/cgroup`).  If the type is `cgroup2fs`
+then the Kubelet will use only cgroups v2 during all its execution.
+
+### cgroup namespace
+
+A cgroup namespace restricts the view on the cgroups.  When
```
derekwaynecarr: it is not clear how this section maps to the kep.  it feels like it is referencing rootless requirements, which in 1/28 sig-node meeting we discussed as not being an initial target.  if you do not intend to use the cgroup namespace because the kubelets view on cgroups should not be restricted, please omit it.  cadvisor metrics endpoints can still be enabled in the kubelet so i feel the kubelet still needs to have full visibility of cgroups on host.

/cc @dashpole

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: @giuseppe future improvements for what cgroup v2 offers should be out of scope of this kep.  i would keep this kep focused on kubelet is tolerant of cgroup v2 host.  adding new cgroup v2 specific features to resource model would be a separate enhancement.

To address @yujuhong question, I think we are saying the following:

- no change required to pod spec
- kubelet cgroup manager uses v1 or v2 mode based on its introspection of sys/fs on startup
- kubelet qos and pod level cgroup creation on a v2 host uses the mapping table specified
- cri implementers that support cgroup v2 hosts must use a similar mapping to ensure that pod bounding cgroup values are consistent with the container cgroup values.
- oci runtime spec is not required to change as cri implementer provides mapping in the transitional period.

@giuseppe agree with above?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+OCI will need support for cgroups v2 and CRI implementations will
+write to the cgroups v2 section of the new OCI runtime config.json.
+
+## Risk and Mitigations
+
+Some cgroups v1 features are not available with cgroups v2:
+
+- _cpuacct.usage_percpu_
+- network stats from cgroup
```
derekwaynecarr: it is worth calling this out to sig-instrumentation.

/cc @brancz for awareness.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,216 @@
...
+
+Some cgroups v1 controllers such as _device_ and _net_cls_,
+_net_prio_ are not available with the new version.  The alternative to
+these controllers is to use eBPF.
+
+## Graduation Criteria
```
derekwaynecarr: I think this criteria is fine as outlined, but we will need to evolve this a little bit or call out things that are not in scope.

For example, device plugin implementations that require v2 enablement are explicitly out of scope of kubernetes.

In addition, I think we need to make it very clear in this proposal that there is no identified timeline for deprecating v1 support.  I suspect kubelet and cri implementations will have a desire to support either mode for an extended period of time.

The last thing we should call out is that the dockershim implementation embedded in kubelet is explicitly not supported on a v2 host.  I would assert that kubelet when configured to use dockershim on a v2 host should immediately error.

keps/sig-node/20191118-cgroups-v2.md
brancz: I think a lot of people rely on these metrics today. Can we provide possible other source(s) we can point people to for this. Any ideas @giuseppe?

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: No, unrelated to rootless

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: >  I would assert that kubelet when configured to use dockershim on a v2 host should immediately error.

`dockerd` itself already has basic support for v2.
Will [cri-dockerd](https://github.com/kubernetes/kubernetes/issues/73933) support v2?

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: More context is available here: https://github.com/containers/libpod/issues/4363

giuseppe: as Akihiro pointed out, this is not related to rootless.  It is useful to limit the container's view on cgroups.  Differently than cgroup v1, every kernel supporting cgroups v2 has support for the cgroup namespace so it makes sense to enable it by default:

```
## Without a cgroup namespace
# podman run --cgroupns=host fedora cat /proc/self/cgroup
0::/machine.slice/libpod-c8aabfba22cae68f60063d63cd6a9c17d91cf7bbf3cf9052b13d78d6af7d37b0.scope

## With a cgroup namespace
# podman run --cgroupns=private fedora cat /proc/self/cgroup
0::/
```

keps/sig-node/20191118-cgroups-v2.md
giuseppe: @brancz I think the way for reading network stats is through eBPF

Are the percpu stats used anywhere?

brancz: Well cadvisor practically exposes all stats but as long as cpu seconds used total is available I think we’re ok.

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: @giuseppe @AkihiroSuda thanks for the clarification.

we need to be able to preserve the ability to run monitoring agents that have full visibility.  is usage of this namespace only enabled if the container is not privileged?  for example, we will soon require the need to run cAdvisor as a pod rather than via the kubelet, and that pod will need full host visibility.

keps/sig-node/20191118-cgroups-v2.md
derekwaynecarr: @AkhiroSuda that would be a choice for cri-dockerd, but I think its reasonable to not continue to iterate on dockershim further as it is moving out of tree.

keps/sig-node/20191118-cgroups-v2.md
AkihiroSuda: Even privileged pod should unshare cgrouns, as well as netns and pidns.

Adding new podspec field seems consistent with netns and pidns

keps/sig-node/20191118-cgroups-v2.md
dchen1107: Agreed with @derekwaynecarr on focusing cgroup v2 support as offering parity with v1. Can we document this as another goal for this KEP?

keps/sig-node/20191118-cgroups-v2.md
dchen1107: Can you summarize what we discussed above, what we agreed and not-agreed and unclear in the doc? For example, we are not clear on how to support privileged pod which should have the full view of the cgroups on the host.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,229 @@
...
+## Motivation
+
+The new kernel cgroups v2 API was declared stable more than two years
+ago. Newer features in the kernel such as PSI depend upon cgroups
+v2. groups v1 will eventually become obsolete in favor of cgroups v2.
+Some distros are already using cgroups v2 by default, and that
+prevents Kubernetes from working as it is required to run with cgroups
+v1.
+
+## Goals
```
dchen1107: May I suggest to explicitly document non-goal at the beginning of doc too for a better communication? For example, podspec changes to enable some advanced features, dockershim, etc.

dchen1107:
We discussed this KEP at today's SIG Node and agreed to move the KEP to implementable to unblock cgroup v2 progress assuming:

1) This KEP is only focusing on cgroup v2 support as offering parity with v1. 
2) The missing features should have an alternative approach being proposed and documented.
3) The advanced feature enablement for cgroup v2 are out-of-scope in this KEP, and should be discussed and followed up through separate KEPs.
4) We need fix, evolve and modify node e2e tests for cgroup v2, especially adding more stress tests & failure tests to the resource management related node e2e. Some of the tests are not included in K8s conformance tests and regular release qualification process given the nature of the tests themselves, but very valuable for  A/B testing for cgroup v1 and v2.

cc/ @derekwaynecarr

keps/sig-node/20191118-cgroups-v2.md
giuseppe: yes sure, I'll add it to the user stories.

keps/sig-node/20191118-cgroups-v2.md
giuseppe: for privileged pods we can default to use the host cgroupns, so that we won't break existing privileged containers.

I'll specify it in the KEP

AkihiroSuda: Privileged pods are already going to be broken because of incompatibility across v1 and v2.

Why not treat cgroupns just same as pidns and netns?

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,233 @@
...
+
+### cgroup namespace
+
+A cgroup namespace restricts the view on the cgroups.  When
+unshare(CLONE_NEWCGROUP) is done, the current cgroup the process
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
+Privileged pods will still use the host cgroup namespace so to have
```
AkihiroSuda: This seems inconsistent with other namespaces. On cgroup v2 systems, the host cgroup namespace should be enabled only if `PodSpec.HostCgroup` is set to true.

keps/sig-node/20191118-cgroups-v2.md
giuseppe: Privileged pods will be broken if the container payload doesn't handle cgroup v2, but that doesn't depend from the container configuration.

Changing the default means the configuration must be changed to maintain the current behaviour.

keps/sig-node/20191118-cgroups-v2.md
giuseppe: sure, I'll add it

giuseppe:
@derekwaynecarr @yujuhong I took care of the others comments and changed `status: implementable`

keps/sig-node/20191118-cgroups-v2.md
bg-chun: wow, hugetlb on cgroup v2...
one of my concerns just resolved by @giuseppe .
thank you.

I just hope it will be available on  major linux distributions.

giuseppe: > I just hope it will be available on major linux distributions.

It should be part of Linux 5.6, that is the next release.

keps/sig-node/20191118-cgroups-v2.md
bg-chun: If we create a Pod1 which has container1 and container2.
cgroupns hierarchy will be:
Pod1/container1
Pod1/container2

Then values in Pod1 will be merged values of containers.
It means if each container in Pod1 requests 1GB memory.
memory.limit_in_bytes in Pod1 cgroup will be 2GB as bytes.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,240 @@
...
+- _cpuacct.usage_percpu_
+- network stats from cgroup
+
+Some cgroups v1 controllers such as _device_ and _net_cls_,
+_net_prio_ are not available with the new version.  The alternative to
+these controllers is to use eBPF.
+
+## Graduation Criteria
+
+- Alpha: Basic support for running Kubernetes on a cgroups v2 host,
```
bg-chun: @giuseppe 
all phases(1 and 2), will it be done in alpha?

giuseppe: I think we should not block on phase 2 for Alpha, that is going to take much longer.  I'll clarify that

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,242 @@
...
+unshare(CLONE_NEWCGROUP) is done, the current cgroup the process
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
+Privileged pods will still use the host cgroup namespace so to have
+visibility on all the other cgroups.
+
+### Phase 1: Convert from cgroups v1 settings to v2
+
```
bg-chun: when(and where) values will be converted?
1) before values set to CRI(LinuxContainerResources, etc)?
2) after values set to CRI and before it sends to CRI runtimes?
    (at the [remote_runtime](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/remote/remote_runtime.go))

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,242 @@
...
+With this approach cAdvisor would have to read back values from
+cgroups v2 files (already done).
+
+Kubelet PR: [https://github.com/kubernetes/kubernetes/pull/85218](https://github.com/kubernetes/kubernetes/pull/85218)
+
+### Phase 2: Use cgroups v2 throughout the stack
+
+This option means that the values are written directly to cgroups v2
+by the runtime. The Kubelet doesn’t do any conversion when setting
+these values over the CRI.  We will need to add a cgroups v2 specific
```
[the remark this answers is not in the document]
giuseppe: most likely 1.  It also depends how these will be modelled in the OCI runtime specs, so we can follow a similar schema

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,242 @@
...
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
+Privileged pods will still use the host cgroup namespace so to have
+visibility on all the other cgroups.
+
+### Phase 1: Convert from cgroups v1 settings to v2
+
+We can convert the values passed by the k8s in cgroups v1 from to
```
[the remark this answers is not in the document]
giuseppe: Phase 1 doesn't require changes to CRI.  The existing settings will be re-used.

` Node Allocatable` is part of this KEP.  Feature parity with cgroup v1 is an explicit goal

keps/sig-node/20191118-cgroups-v2.md
giuseppe: the OCI runtime will change them.

These can be converted early when the Kubelet/CRI runtime accesses directly the cgroup, e.g. https://github.com/kubernetes/kubernetes/pull/85218/files#diff-fa76bb6ae2d9b4bb3d023737fe5e6029R381

bg-chun: oh, you will change cgroup manager.
ok I think it will work for pod cgroup and node allocatable feature.
Then CRI/OCI runtime take care of container cgroup.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,242 @@
...
+The dockershim implementation embedded in the Kubelet won't be
+supported on cgroup v2.
+
+### Current status of dependencies
+
+- CRI-O+crun: support cgroups v2
+
+- runc support for cgroups v2 is work in progress [current status](#current-cgroups-usage-and-the-equivalent-in-cgroups-v2)
+
+- containerd: [https://github.com/containerd/containerd/pull/3799](https://github.com/containerd/containerd/pull/3799)
```
tedyu: This has been merged

AkihiroSuda: Better to add a link to https://github.com/containerd/containerd/issues/3726

giuseppe: thanks, changed to containerd/containerd#3726

giuseppe:
@yujuhong are you fine with the last version of the KEP?

AkihiroSuda:
@derekwaynecarr @yujuhong @dchen1107 PTAL?

derekwaynecarr:
per dawn's comment and discussion on sig-node, we can merge this kep, and update as we find issues with implementation in 1.19.  there is no disagreement on need to iterate here.

see: https://github.com/kubernetes/enhancements/pull/1370#issuecomment-582102634

/approve
/lgtm

hectorj2f:
awesome work!

crosbymichael:
LGTM :)

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,241 @@
...
+
+**CPU controller**
+
+| OCI (x) | cgroup 2 value (y) | conversion  |  comment |
+|---|---|---|---|
+| shares | cpu.weight | y = (1 + ((x - 2) * 9999) / 262142) | convert from [2-262144] to [1-10000]|
+| period | cpu.max | y = x| period and quota are written together|
+| quota | cpu.max | y = x| period and quota are written together|
+
+**blkio controller**
```
smileusd, line 176: Any kuberente versions and plans to support blkio controller ?

warmchang: > Any kuberente versions and plans to support blkio controller ?

[kubernetes/enhancements#1907 ](https://github.com/kubernetes/enhancements/pull/1907) the KEP is WIP.

keps/sig-node/20191118-cgroups-v2.md
```diff
@@ -0,0 +1,241 @@
...
+
+### cgroup namespace
+
+A cgroup namespace restricts the view on the cgroups.  When
+unshare(CLONE_NEWCGROUP) is done, the current cgroup the process
+resides in becomes the root.  Other cgroups won't be visible from the
+new namespace.  It was not enabled by default on a cgroup v1 system as
+older kernel lacked support for it.
+
+Privileged pods will still use the host cgroup namespace so to have
```
AkihiroSuda, line 143: Let's make this customizable, especially for nested containers:
- https://github.com/kubernetes/enhancements/pull/5715

PR #1408, opened 2019-12-12
Update PRR KEP to link to current questionnaire

johnbelamaric:
Duplication of the existing questions causes confusion.

/sig architecture


wojtek-t:
/lgtm

Thanks!

PR #1418, opened 2019-12-20
[WIP] Apiserver discovery URL env variable

anguslees:


anguslees:
/cc thockin

I'll flesh out the rest in the next few days, but I think this captures our earlier slack discussion.

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+## Proposal
+
+1. Unconditionally inject a new environment variable into pods called
+   `KUBERNETES_SERVICE_URL`.  The value would be the correctly-escaped
+   equivalent of
+   `https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT`. Note
+   that this would continue to use a (single) IPv4 or v6 literal
+   address in most cases.
+
+   An open question is whether the URL could ever include a path
```
thockin: TTBOMK it is allowed, but nobody does it (yet)

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+
+The biggest drawbacks of this proposal are that several
+[alternatives](#alternatives) are also appealing.  It is not clear
+that _this_ proposal is the right tradeoff since it is neither the
+most minimal, nor the "best".
+
+## Alternatives
+
+There are several fine alternatives to this proposal:
+- Continue to use KUBERNETES_SERVICE_HOST, but change it to
```
thockin: I don't think we can assume DNS, even if it is almost always present.

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+  Cons: This has numerous issues for instances. We should not repeat
+  these mistakes given the alternatives available.
+
+- Write a file to a well-known path in the pod containing the
+  apiserver endpoint information (perhaps in
+  `/var/run/secrets/kubernetes.io/serviceaccount`).  If this was in
+  kube/config format, and we set `KUBECONFIG` to the path then all the
+  client in-cluster vs out-of-cluster bootstrap code goes away.
+
+  Cons: Larger change.
```
thockin: And not mutually exclusive

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+  - "@anguslees"
+owning-sig: sig-network
+participating-sigs:
+  - sig-apimachinery
+  - sig-architecture
+reviewers:
+  - "@thockin"
+  - "@liggitt"
+approvers:
+  - TBD
```
thockin: thockin

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+editor: TBD
+creation-date: 2019-12-20
+last-updated: 2019-12-20
+status: provisional
```
thockin: We can let this gather some feedback, but I think it's effectively implementable

aojea:
+1
nice 👏

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+- Declare the endpoint is always `_kubernetes._tcp` DNS `SRV`.
+  Cool.  Unfortunately SRV is not trivial to use in a URL.
+
+- Declare a well-known (anycast) IPv4/IPv6 address and port.
+  Like some cloud providers did for 169.254.169.254 instance metadata.
+
+  Cons: This has numerous issues for instances. We should not repeat
+  these mistakes given the alternatives available.
+
+- Write a file to a well-known path in the pod containing the
```
mvladev: A file has the benefit that it can be changed at runtime. Environment variables require restart of the containers to reload.

thockin: To do this as a file we would need to use downward API into a volume.  Users would have to opt-in to it, which makes it less than ideal for being the backstop for clients to find the master

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Non-Goals
+
+- Actually remove `FOO_SERVICE_HOST`/`FOO_SERVICE_PORT` variables.
+  This is covered by [other KEPs][servicevarskep].
+
+## Proposal
+
+1. Unconditionally inject a new environment variable into pods called
+   `KUBERNETES_SERVICE_URL`.  The value would be the correctly-escaped
```
danwinship: Do we want to change the name at all while we're changing things? eg, `KUBERNETES_APISERVER_URL`? (Or just `KUBERNETES_APISERVER`).

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+   component.  This KEP makes no statement about that, but notes the
+   choice of URL format leaves this and alternate schemes as possible
+   extension points in future (unlike just "host:port" format).
+
+2. Announce the deprecation of the existing `KUBERNETES_SERVICE_HOST`
+   and `KUBERNETES_SERVICE_PORT`.  Given the widespread client impact
+   this KEP does not propose a timeline on actual removal, just that
+   in-cluster clients SHOULD use the new `KUBERNETES_SERVICE_URL`
+   mechanism.
+
```
danwinship: I assume also:

    3. Update `k8s.io/client-go/tools/rest.go:InClusterConfig()` to use
       `KUBERNETES_SERVICE_URL`

?

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+without having to think about IPv6 syntax details.
+
+#### Story 2
+
+As a cluster admin, I would like to enable an apiserver IPv6 endpoint
+without encountering widespread URL bugs.
+
+### Implementation Details/Notes/Constraints [optional]
+
+What are the caveats to the implementation?
```
danwinship: What if a pod was previously defining an environment variable named `KUBERNETES_SERVICE_URL` but with different semantics? (eg, it points to something other than the apiserver)

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: Longer is less likely to collide accidentally.

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: +1

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: We can only set it if we don't find any existing `env` of the same name.

keps/sig-network/20191220-apiserver-url-envvar.md
asauber: Currently `KUBERNETES_SERVICE_HOST` is not populated if the [`kubernetes` service does not have a ClusterIP](https://github.com/kubernetes/kubernetes/blob/2a7225ff0e90e76f634b29d3a35a1ae3f0631bb7/pkg/kubelet/envvars/envvars.go#L42). Although it does not necessarily need to be spelled out in this KEP, it would be very useful to allow the `KUBERNETES_SERVICE_URL` hostname to be retrieved from an ExternalName type service. Which is [very close to being implemented here](https://github.com/kubernetes/kubernetes/blob/2a7225ff0e90e76f634b29d3a35a1ae3f0631bb7/pkg/kubelet/envvars/envvars.go#L66), but short-circuited in the logic inside `FromServices` above.

asauber: Currently `KUBERNETES_SERVICE_HOST` is only populated if the `kubernetes` service [has a ClusterIP](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/envvars/envvars.go#L46). Although it may not need to be spelled out in this KEP, `KUBERNETES_SERVICE_URL` should also populate its hostname component by looking at the hostname of `kubernetes` if it is an `ExternalName` type service.

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+The biggest drawbacks of this proposal are that several
+[alternatives](#alternatives) are also appealing.  It is not clear
+that _this_ proposal is the right tradeoff since it is neither the
+most minimal, nor the "best".
+
+## Alternatives
+
+There are several fine alternatives to this proposal:
+- Continue to use KUBERNETES_SERVICE_HOST, but change it to
+  `kubernetes.default` or other hostname instead of an IP (v4 or v6)
```
asauber: The typical CoreDNS deployment relies on the `KUBERNETES_SERVICE_HOST` environment variable, so using an in-cluster DNS name for the `kubernetes` service appears to be blocked by this bootstrapping problem. However, a publicly-resolvable DNS does work in the case of CoreDNS and other controllers, if provided as a forced override to `KUBERNETES_SERVICE_HOST`. Given this, supporting `kubernetes` of type `ExternalName` for this KEP matches well with the proposal, but might be considered an implementation detail.

keps/sig-network/20191220-apiserver-url-envvar.md
```diff
@@ -0,0 +1,294 @@
...
+  This is covered by [other KEPs][servicevarskep].
+
+## Proposal
+
+1. Unconditionally inject a new environment variable into pods called
+   `KUBERNETES_SERVICE_URL`.  The value would be the correctly-escaped
+   equivalent of
+   `https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT`. Note
+   that this would continue to use a (single) IPv4 or v6 literal
+   address in most cases.
```
asauber: Currently KUBERNETES_SERVICE_HOST is only populated if the kubernetes service has a ClusterIP. Although it may not need to be spelled out in this KEP, KUBERNETES_SERVICE_URL should also populate its hostname component by looking at the hostname of kubernetes if it is an ExternalName type service, resulting in KUBERNETES_SERVICE_URL of the form https://[clusteripv6]:6443 or https://publicly.resolvable.example.net:6443

keps/sig-network/20191220-apiserver-url-envvar.md
anguslees: Well, we _could_ mandate it as a new CRI requirement (hand wave) - and not use downward API.

To the original point, I think many clients will only read it once at startup so they can (eg) cache/reuse a persistent HTTPS connection.  So I'm guessing changing it at runtime won't work anyway .. unless we _force_ frequent changes on everyone by being deliberately perverse (cycle ClusterIP addresses just because, etc).

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: The binary name doesn't matter - it is the kubernetes API server.

I could live with  KUBERNETES_API_URL - good enough?

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: As of now, ExternalName is bot really supported for the master Service, but I do not object to declaring support here.

keps/sig-network/20191220-apiserver-url-envvar.md
thockin: Agree with all points.  I don't see much value in making this mutable -- if that becomes an issue, we can discuss downward API support for it.

keps/sig-network/20191220-apiserver-url-envvar.md
danwinship: That doesn't fully solve the problem though. If we think there actually might be pods defining a `KUBERNETES_SERVICE_URL` variable (or whatever we call it) and using it for something other than pointing to the apiserver, and we're not going to override its value in that case, then that means that generic code running in a pod can't assume that `KUBERNETES_SERVICE_URL` necessarily contains a URL pointing to the apiserver.

The easiest thing to do would be to assume/pretend this isn't actually going to happen (possibly after picking an even-less-likely-to-conflict name for the variable), and if it later turns out that someone was using that name and things break, then apologize and tell them they have to rename their variable.

Alternatively, we could say that the new variable really only exists for use directly from the `command` of a `PodSpec`, where the person using the variable can see that the pod isn't overriding it to mean something else, and that generic code like `rest.InClusterConfig` that may end up running in arbitrary pods needs to keep using `KUBERNETES_SERVICE_HOST` / `KUBERNETES_SERVICE_PORT`. This implies we continue supporting the old variables forever, and `KUBERNETES_SERVICE_URL` can never be anything besides just the correctly-concatenated value of the two old variables (eg, no `ExternalName`).

Or, if we take a bit longer to roll things out, we could declare that in the future, all environment variables starting with `KUBERNETES_` will be reserved for kubernetes-internal usage and can't be used by pods for unrelated purposes. To distinguish "I'm coincidentally using a conflicting variable name" from "I'm intentionally overriding a standard variable", we'd add an `override` field to `corev1.EnvVar`, and make kubelet throw warnings for the next 2 releases any time it launches a pod that sets a variable with a name starting with `KUBERNETES_` without setting `override: true` (and warn anyway if they're "overriding" a variable that kubelet doesn't know about). Then after 2 releases, make it an error rather than a warning to omit `override` (but still just a warning to override an unknown variable), and introduce the new apiserver URL variable.

thockin: Dan makes a good point.  There's literally no name we can assume is not in use, and WAY back, we opted not to reserve names with a prefix.  I am still not comfortable reserving a prefix in a space that is really the end-users' to own.

I think "set-unless-already-set" is the best we can do, maybe with an event or something to indicate a warning.  That does mean that the old variables are there for a very long time, and that is also part of why ExternalName is a hard transition to make.  We can fix all the places internally that do dumb catenation of host and port to be v6 safe.  We can maybe even use the URL if it exists AND "tastes right" (for some definition).

thockin (commented):
@anguslees Are you hoping to revisit this?

anguslees:
> @anguslees Are you hoping to revisit this?

Yes .. and "hope" is the right word :/

PR #1443, opened 2020-01-10
Modify wrong word of sig-network nodelocal dns

lemonli:
Fix wrong word: 'repond' to 'respond'

/kind bug
/sig network


lemonli:
retry

PR #1448, opened 2020-01-14
Add runtime handler as annotations to ImageSpec

kkmsft:


Random-Liu:
/lgtm

This looks good to me.

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -333,6 +336,46 @@ message PullImageRequest {
...
+```
+type ImageSpec struct {
+  Image string
+  Annotations []Annotation
+}
+```
+ 
+The runtimeHandler annotation will be based on the Runtime Class specified by the user:
+```
+“runtimehandler”: “<corresponding values>”
```
Random-Liu: Can we also add `kubernetes.io` domain?

PatrickLang: as in `kubernetes.io/runtimehandler` ?

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -333,6 +336,46 @@ message PullImageRequest {
...
+ ```
+
+We could potentially also add the kubernetes specification annotations for consideration of the runtime:
+```
+“kubernetes.io/arch”: “amd64”
+“kubernetes.io/os”: ”linux”
+```
+
+Note that these are currently arrived from GOARCH and GOOS at runtime, which does not reflect the image which needs to be pulled but instead corresponds to the system on which kubelet is running. The runtime handler specified in the annotation will provide more accurate indication of the user intend to the runtime.
```
jterry75: `derived`

jterry75:
@Random-Liu - Is there any expectation at all that on `ImageSpec` return values from `*Stats*` and `List*` style functions that there be annotations? IE: Are the annotations only for `*Request` input to various functions and not used for `*Response`

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -333,6 +336,46 @@ message PullImageRequest {
...
+ 
+The runtimeHandler annotation will be based on the Runtime Class specified by the user:
+```
+“runtimehandler”: “<corresponding values>”
+ ```
+
+We could potentially also add the kubernetes specification annotations for consideration of the runtime:
+```
+“kubernetes.io/arch”: “amd64”
+“kubernetes.io/os”: ”linux”
```
jterry75: Should we define these in the `kubernetes.io/image/*` namespace?

jterry75: And have:
`.../arch`
`.../os`
`.../variant`
as known values?

Random-Liu:
> Is there any expectation at all that on ImageSpec return values from *Stats* and List* style functions that there be annotations? IE: Are the annotations only for *Request input to various functions and not used for *Response

I think we should return the annotation in all Responses.

We may want to make that clear in the doc.

keps/sig-windows/windows-runtimeclass-support.md
PatrickLang: Ah good point. I see why this is confusing. Today we already have well-known labels applied to the node by the kubelet [doc](https://github.com/kubernetes/enhancements/blob/f1a799d5f4658ed29797c1fb9ceb7a4d0f538e93/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md#proposal)

* `kubernetes.io/os`
* `kubernetes.io/arch`

Those are intended to be used for scheduling as nodeSelectors, not for determining what image to pull.

In this KEP I propose to encode a user's _desired_ os/arch as part of the RuntimeClass. This may be different from the node labels above, so it would be good to use different names. RuntimeClass is the only thing passed in the pod sandbox config. I think for image/* to be useful, we'd need matching ones in the pod sandbox config too.

jterry75: Ah I see. Ok then yes we will just have one known annotation `kubernetes.io/runtimehandler` that is passed in the `ImageSpec.Annotations`. Thanks

keps/sig-windows/windows-runtimeclass-support.md
PatrickLang: those annotations are bit different scope though - they're stored in the container manifest. We're talking about changing CRI, not OCI

mikebrow: right just a mention regarding another reserved annotation domain for these annotations.

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -297,6 +298,8 @@ To make this easier to consume in the Kubelet and APIs, it will be updated in mu
 
 #### Adding handler to CRI pull API
 
+> As per discussions in sig-node meetings based on the PR https://github.com/kubernetes/kubernetes/pull/84486, this approach will no longer be considered. An alternate approach to add annotations to the ImageSpec will be path taken.
```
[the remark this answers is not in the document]
kkmsft: Sure, will remove

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -343,6 +386,9 @@ The names of aren't part of a versioned API today, so there's no risk to upgrade
 
 This will be reviewed with contributors from both Kubernetes and ContainerD and tested together while Windows support in CRI-ContainerD is still in development. As this is an optional field today in `RunPodSandboxRequest`, there's no risk if someone wants to not specify it in `RuntimeClass`. In that case ContainerD will use the default configuration for pull, create sandbox, and so on.
 
+#### Adding annotations to ImageSpec
+These annotation will be an optional parameter and default configuration is used by `ContainerD`(or any other runtime involved) when no annotations are specified.
```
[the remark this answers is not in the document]
kkmsft: Have reframed the words to reflect this.

kkmsft:
@Random-Liu @PatrickLang - Have update the ImageSpec usage in List. PTAL.

PatrickLang:
/LGTM
/hold
Can remove hold once @Random-Liu has reviewed

keps/sig-windows/windows-runtimeclass-support.md
```diff
@@ -297,53 +300,98 @@ To make this easier to consume in the Kubelet and APIs, it will be updated in mu
...
+```
+type ImageService interface {
+    // PullImage pulls an image from the network to local storage using the supplied
+    // secrets if necessary. It returns a reference (digest or ID) to the pulled image.
+    PullImage(image ImageSpec, pullSecrets []v1.Secret, podSandboxConfig *runtimeapi.PodSandboxConfig) (string, error)
+    // GetImageRef gets the reference (digest or ID) of the image which has already been in
+    // the local storage. It returns ("", nil) if the image isn't in the local storage.
+    GetImageRef(image ImageSpec) (string, error)
+    // Gets all images currently on the machine.
+    ListImages() ([]ImageSpec, error)
```
Random-Liu: `ListImages` should continue return Images, we just need to add ImageSpec information into `Image`.

PatrickLang: ok - can we work out the specifics of that during phase 2 PR? We don't have the background on why `Image` and `ImageSpec` are different

kkmsft: Thanks you @Random-Liu . I have modified the struct Image to include the ImageSpec in the updated PR. Please have a look - @Random-Liu @PatrickLang

PatrickLang: oh ok - that makes sense

PatrickLang: /lgtm
/hold
(remove once @Random-Liu ok with final update)

PR #1467, opened 2020-01-18
kepify: New command to generate KEP json for Netlify

dims:
`go run cmd/kepify/main.go --dir keps --output out.json`

### DIFF
--- a/cmd/kepify/main.go
+++ b/cmd/kepify/main.go
@@ -1,0 +1,184 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"crypto/md5"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"k8s.io/enhancements/pkg/kepval/keps"
+)
+
+func Usage() {
+	fmt.Fprintf(os.Stderr, `
+Usage: %s [-dir <kep-directory>] [-output <path-to-json-file>]
+Command line flags override config values.
+`, os.Args[0])
+	flag.PrintDefaults()
+}
+
+func main() {
+	dirPath := flag.String("dir", "keps", "root directory for the KEPs")
+	filePath := flag.String("output", "keps.json", "output json file")
+
+	flag.Usage = Usage
+	flag.Parse()
+
+	if len(*dirPath) == 0 {
+		fmt.Fprintf(os.Stderr, "please specify the root directory for KEPs using '--dir'\n")
+		os.Exit(1)
+	}
+	if _, err := os.Stat(*dirPath); os.IsNotExist(err) {
+		fmt.Printf("directory does not exist : %s", *dirPath)
+		os.Exit(1)
+	}
+
+	if len(*filePath) == 0 {
+		fmt.Fprintf(os.Stderr, "please specify the file path for the output json using '--output'\n")
+		os.Exit(1)
+	}
+
+	// Find all the keps
+	files, err := findMarkdownFiles(dirPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "unable to find markdown files: %v\n", err)
+		os.Exit(1)
+	}
+	if len(files) == 0 {
+		fmt.Fprintf(os.Stderr, "did not find any KEPs\n")
+		os.Exit(1)
+	}
+
+	// Parse the files
+	proposals, err := parseFiles(files)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error parsing files: %q\n", err)
+		os.Exit(1)
+	}
+
+	// Generate the json output
+	err = printJSONOutput(*filePath, proposals)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "could not open file: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func findMarkdownFiles(dirPath *string) ([]string, error) {
+	files := []string{}
+	err := filepath.Walk(
+		*dirPath,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+			if info.IsDir() {
+				return nil
+			}
+			if ignore(info.Name()) {
+				return nil
+			}
+			files = append(files, path)
+			return nil
+		},
+	)
+	return files, err
+}
+
+func parseFiles(files []string) (keps.Proposals, error) {
+	var proposals keps.Proposals
+	for _, filename := range files {
+		parser := &keps.Parser{}
+		file, err := os.Open(filename)
+		if err != nil {
+			return nil, fmt.Errorf("could not open file: %v\n", err)
+		}
+		defer file.Close()
+		kep := parser.Parse(file)
+		// if error is nil we can move on
+		if kep.Error != nil {
+			return nil, fmt.Errorf("%v has an error: %q\n", filename, kep.Error.Error())
+		}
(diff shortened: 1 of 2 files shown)


dims:
cc @saschagrunert @jeefy 

here's a crack at what we talked about on friday. running the command generates a json file which hopefully can be massaged to work with netlify. PTAL

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "authors", strings.Join(kep.Authors, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "editor", kep.Editor)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "creation-date", kep.CreationDate)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "last-updated", kep.LastUpdated)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "status", kep.Status)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "see-also", strings.Join(kep.SeeAlso, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "replaces", strings.Join(kep.Replaces, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "superseded-by", strings.Join(kep.SupersededBy, ","))
+		contents, _ := json.Marshal(kep.Contents)
+		fmt.Fprintf(file, "\t\t\"%s\": %s\n", "markdown", contents)
```
saschagrunert: Hm, can we use the build tags to serialize the JSON like we already do for the YAML part?
https://github.com/kubernetes/enhancements/blob/502b4a7ace29104780658b41527a10ee9806b3c7/pkg/kepval/keps/proposals.go#L37-L49

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		return err
+	}
+	defer file.Close()
+
+	total := len(proposals)
+	fmt.Printf("Total KEPs: %d\n", total)
+
+	fmt.Fprintln(file, "{")
+	for i, kep := range proposals {
+		fmt.Fprintf(file, "\t\"%s\": {\n", hash(kep.OwningSIG+":"+kep.Title))
```
saschagrunert: I think this is okay, but makes the key unusable for consumers. We probably want to switch to a plain array and remove the key entirely. Or we use the KEPs path in the repository as key, this way we could easily link it back into this repository.

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "title", kep.Title)
```
saschagrunert: From the example JSON I see that the titles differ in its style: like `aws-k8s-tester` vs `Custom endpoints support for AWS Cloud Provider`. Do you think we can avoid that somehow?

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "owning-sig", kep.OwningSIG)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "participating-sigs", strings.Join(kep.ParticipatingSIGs, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "reviewers", strings.Join(kep.Reviewers, ","))
```
saschagrunert: It would be good if this will be a plain JSON array.

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+	defer file.Close()
+
+	total := len(proposals)
+	fmt.Printf("Total KEPs: %d\n", total)
+
+	fmt.Fprintln(file, "{")
+	for i, kep := range proposals {
+		fmt.Fprintf(file, "\t\"%s\": {\n", hash(kep.OwningSIG+":"+kep.Title))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "title", kep.Title)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "owning-sig", kep.OwningSIG)
```
saschagrunert: Stripping the `sig-` prefix would probably help. The same applies for the participating SIGs.

cmd/kepify/main.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "participating-sigs", strings.Join(kep.ParticipatingSIGs, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "reviewers", strings.Join(kep.Reviewers, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "authors", strings.Join(kep.Authors, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "editor", kep.Editor)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "creation-date", kep.CreationDate)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "last-updated", kep.LastUpdated)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "status", kep.Status)
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "see-also", strings.Join(kep.SeeAlso, ","))
+		fmt.Fprintf(file, "\t\t\"%s\": \"%s\",\n", "replaces", strings.Join(kep.Replaces, ","))
```
saschagrunert: If the fields are empty, then I suggest in removing them or setting them to `null`.

cmd/kepify/main.go
saschagrunert: I think we could somehow split the table of contents from the markdown. This way we easily could preview the KEP only by its TOC.

saschagrunert:
Thank you for working on this :pray: I just have a couple of comments.

cmd/kepify/main.go
dims: @saschagrunert this gives us more control over the output if we run into any netlify idiosynchracies. we can always do this in a follow up PR.

cmd/kepify/main.go
dims: the hash of sig + title is unique, there was discussion about moving the keps around (separate directories for keps in different status etc), so i needed something that will stay across any such moves. note that we have the `Filename` in the data structure as well, So worst case we just ignore this hash.

saschagrunert: Makes sense :+1:

cmd/kepify/main.go
dims: not in this PR, we have to raise this as a problem in https://github.com/kubernetes/enhancements/issues/822

cmd/kepify/main.go
dims: DONE

cmd/kepify/main.go
dims: Can we do this in netlify? i'd like the markdown to have the "official" names

cmd/kepify/main.go
dims: let's do this in a follow up please

cmd/kepify/main.go
saschagrunert: Yes, sure.

cmd/kepify/main.go
dims: DONE (see sample output in https://gist.github.com/dims/97ca19d97d74c5ec697f7d01c5dc1247)

cmd/kepify/main.go
saschagrunert: Okay, good :+1:

dims:
updated generated json in https://gist.github.com/dims/97ca19d97d74c5ec697f7d01c5dc1247

dims:
/assign @justaugustus @jdumars @calebamiles 

Folks, let's please merge and iterate, this is just the first cut. Hoping @saschagrunert @jeefy and others can help tweak things as we try this with netlify

dims:
thanks @justaugustus

saschagrunert:
@dims a first implementation proposal can be found here: https://github.com/kubernetes-sigs/release-notes/pull/140

We probably want to setup a new k-sigs project for this.

dims:
@saschagrunert new repo sounds good! please request one at https://github.com/kubernetes/org/issues

PR #1479, opened 2020-01-21
Update OIDC KEP and mark implementable

mtaufen:
/assign @liggitt @mikedanese

### DIFF
--- a/keps/sig-auth/20190730-oidc-discovery.md
+++ b/keps/sig-auth/20190730-oidc-discovery.md
@@ -3,6 +3,7 @@ title: Service Account signing key retrieval
 authors:
   - "@mikedanese"
   - "@cceckman"
+  - "@mtaufen"
 owning-sig: sig-auth
 participating-sigs:
   - sig-auth
@@ -18,8 +19,8 @@ approvers:
   - "@ericchiang"
 editor: TBD
 creation-date: 2018-06-26
-last-updated: 2019-07-30
-status: provisional
+last-updated: 2020-01-25
+status: implementable
 replaces:
   - "https://github.com/kubernetes/community/pull/2314/"
 ---
@@ -29,26 +30,58 @@ replaces:
 ## Table of Contents
 
 <!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
+
+## Release Signoff Checklist
+
+**ACTION REQUIRED:** In order to merge code into a release, there must be an issue in [kubernetes/enhancements] referencing this KEP and targeting a release milestone **before [Enhancement Freeze](https://github.com/kubernetes/sig-release/tree/master/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core Kubernetes i.e., [kubernetes/kubernetes], we require the following Release Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These checklist items _must_ be updated for the enhancement to be released.
+
+- [x] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [x] KEP approvers have set the KEP status to `implementable`
+- [x] Design details are appropriately documented
+- [x] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] Graduation criteria is in place
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+**Note:** Any PRs to move a KEP to `implementable` or significant changes once it is marked `implementable` should be approved by each of the KEP approvers. If any of those approvers is no longer appropriate than changes to that list should be approved by the remaining approvers and/or the owning SIG (or SIG-arch for cross cutting KEPs).
+
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
 
 ## Summary
 
 The Kubernetes API server generates (signs) JSON Web Tokens that are meant to
-authenticate Kubernetes service accounts. The API server can indeed authenticate
+authenticate Kubernetes service accounts (KSA). The API server can indeed authenticate
 these tokens, but in general, no other system can: there is no standard way to
 get the public portion of the keypair used to sign KSA tokens. Systems ("relying
-parties") that want authenticate KSA tokens must either send them back to the
+parties") that want to authenticate KSA tokens must either send them back to the
 API server (via `TokenReview`) or use some provider-specific method to get the
 authentication key.
 
 If a relying party could obtain trusted metadata about the service account token
-provider - in particular, the issuer (`iss`) value and the public key(s) used -
+provider, in particular the issuer (`iss`) value and the public key(s) used,
 then the relying party could authenticate tokens without putting proportionate
 load on the API server. This would allow KSA tokens to be used as a general
 authentication mechanism, including to services outside the cluster or in other
@@ -84,11 +117,11 @@ credentials; and consider:
     compromise a credential with an extended validity period?
 -   How often do the keys expire? How often do I rotate them?
 
-If services (other than the API server could authenticate KSA tokens directly:
+If services (other than the API server) could authenticate KSA tokens directly:
 
--   the API server wouldn't have to scale with the data-plane (i.e.
-    authentication) load; and
--   workloads could use native credentials to authenticate to services outside
+-   The API server wouldn't have to scale with the data-plane (i.e.
+    authentication) load.
+-   Workloads could use native credentials to authenticate to services outside
     of the cluster.
 
 ### Goals
@@ -97,34 +130,38 @@ If services (other than the API server could authenticate KSA tokens directly:
     authenticate KSA tokens.
 -   Attempt compatibility with OIDC: common libraries that authenticate OIDC
     tokens should be able to authenticate KSA tokens.
-
-As a stretch goal / consideration:
-
 -   Support authentication when the API server is not directly reachable by the
     relying party.
     -   e.g.: a cloud-based service authenticating an API server that doesn't
         have a public Internet address.
 
 Note that the API server has a very different flow from OIDC with respect to
-generating tokens. As such, our goal is OIDC *compatibility*...
+generating tokens. As such, our goal is OIDC *compatibility*.
 
 ### Non-Goals
 
-...but not OIDC *compliance*.
+Our goal is *not* OIDC *compliance*.
 
-We aren't trying to: - Make the KSA token process fully compliant with OIDC
-specifications. - OIDC includes flows for token acquisition, token exchange,
-getting user data out of tokens, etc. - But we're only interested in the parts
-relevant to *relying parties*, i.e. token authentication. - We don't need to do
+We aren't trying to make the KSA token process fully compliant with OIDC
+specifications. OIDC includes flows for token acquisition, token exchange,
+getting user data out of tokens, etc. But we are primarily interested in the parts
+relevant to *relying parties*, i.e. token authentication. We don't need to do
 something `REQUIRED` just because the spec says `REQUIRED`; but we may need to
-do it if relying parties expect that field. - This may mean that our
-implementation doesn't comply with the spec; e.g. we might skip `token_endpoint`
-in the
+do it if relying parties expect that field. This may mean that our
+implementation doesn't fully comply with the spec; e.g. we might skip
+`token_endpoint` in the
 [discovery document](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata),
 even without supporting the
-[Implicit Flow](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth). -
-Define new cryptographic or authentication protocols. - Change the format of KSA
-tokens, how they're generated, or how they're issued.
+[Implicit Flow](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth).
+Ensuring compatibility with existing implementations, for example that libraries
+can validate tokens without erroring out during the discovery fetch/parse,
+is critical, and is part of the Alpha to Beta graduation criteria. If we
+determine that it is impossible to broadly meet our compatibility goals with the
+current design, then we will revisit serving a complete discovery doc with
+placeholder values.
+
+We will not define new cryptographic or authentication protocols, or change the
+format of KSA tokens, how they're generated, or how they're issued.
 
 ## Proposal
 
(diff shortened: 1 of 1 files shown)


keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -200,10 +236,68 @@ ClusterRole to grant access to the new paths; some other roles may have
...
 
+### Risks and Mitigations
+
+- Security is being reviewed by @mikedanese and @liggitt.
+- This feature exposes public key information that is derived from sensitive
+  private keys. It is important that code reviewers pay careful attention
+  to the construction of the keysets so that sensitive private keys are not
+  exposed.
+- This proposal allows Kubernetes identities to be federated into other systems,
+  and creates a dependency on the Kubernetes API server to verfiy these identities.
```
liggitt:
```suggestion
  and creates a dependency on the Kubernetes API server to verify these identities.
```

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -29,14 +30,49 @@ replaces:
...
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core Kubernetes i.e., [kubernetes/kubernetes], we require the following Release Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These checklist items _must_ be updated for the enhancement to be released.
+
+- [x] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [x] KEP approvers have set the KEP status to `implementable`
+- [x] Design details are appropriately documented
+- [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
```
liggitt: this seems like it could be checked based on the current content

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -29,14 +30,49 @@ replaces:
...
+**Note:** Any PRs to move a KEP to `implementable` or significant changes once it is marked `implementable` should be approved by each of the KEP approvers. If any of those approvers is no longer appropriate than changes to that list should be approved by the remaining approvers and/or the owning SIG (or SIG-arch for cross cutting KEPs).
+
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
+TODO(mtaufen): Ensure this matches the PR, as implemented, and includes a
```
liggitt: resolve before merge

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Done.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Done.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -146,7 +182,7 @@ configuration:
 > GET /.well-known/openid-configuration
 {
   "issuer": "https://dev.cluster.internal",
-  "jwks_uri": "https://dev.cluster.internal/serviceaccountkeys/v1",
+  "jwks_uri": "https://dev.cluster.internal/openid/v1/jwks",
   "authorization_endpoint": "urn:kubernetes:programmatic_authorization",
```
liggitt: the current PR does not include this field, even though it is required by spec; xref https://github.com/kubernetes/kubernetes/pull/80724/files#r369248302 and resolve the approach there, then update here if needed

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Updated.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Removed from this KEP since we decided to omit it for now.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -97,34 +130,32 @@ If services (other than the API server could authenticate KSA tokens directly:
...
-relevant to *relying parties*, i.e. token authentication. - We don't need to do
+We aren't trying to make the KSA token process fully compliant with OIDC
+specifications. OIDC includes flows for token acquisition, token exchange,
+getting user data out of tokens, etc. But we're only interested in the parts
+relevant to *relying parties*, i.e. token authentication. We don't need to do
 something `REQUIRED` just because the spec says `REQUIRED`; but we may need to
-do it if relying parties expect that field. - This may mean that our
-implementation doesn't comply with the spec; e.g. we might skip `token_endpoint`
-in the
+do it if relying parties expect that field. This may mean that our
```
liggitt: This assumes the libraries that relying parties use partition their validation of the discovery document based on whether it is being used for token verification or token acquisition. I don't know whether that is universally true. If we intend to serve an incomplete discovery document, verifying the subset we are providing allows token validation (e.g. doesn't error out during the discovery doc fetch/validation step) in the N top OIDC libraries in the M top languages would build confidence. That seems like a reasonable thing to do before promoting to beta. The beta graduation criteria already describe verifying compatibility with multiple relying parties, but describing what will be verified more specifically in terms of libraries/languages would be helpful.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -138,16 +169,33 @@ In addition, the API server will serve a JWKS consisting of the public keys from
...
+value `https://cluster.example.com`, which is also the API server root, and
+`--service-account-jwks-uri` value `https://cluster.example.com/openid/v1/jwks`
+(the cluster serves the JWKS at the `/openid/v1/jwks` path by default), the
+API server could expose the following configuration. Note that if the JWKS
+URI overide is not provided, the API server will report it as relative to its
+external address. For example: `https://192.168.10.4:443/openid/v1/jwks`. Note
+also this example intentionally omits the `authorization_endpoint`
+field, as it is not necessary for token verification flows. An older version
+of this KEP set it to `urn:kubernetes:programmatic_authorization`, but the OIDC
+spec requires an `https` URL, so this value was invalid.
```
liggitt: drop the "An older version" sentence

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -195,10 +239,66 @@ configuration:
 ```
 
 The API server would treat these as `nonResourceURLs`, and restrict access
-appropriately. We will consider expanding `system:public-info-viewer` RBAC
-ClusterRole to grant access to the new paths; some other roles may have
-permission already via a pattern match on `nonResourceURLs` (e.g.
-`cluster-admin`).
+appropriately. We will not initially write default RBAC rules for these
```
liggitt: A default clusterrole that includes the discovery and jwk nonResourceURLs would make sense to create, but leave it up to the cluster administrator to bind that role to specific users, or all authenticated users, or all unauthenticated users.

mtaufen: That's true. Updated the PR and this KEP.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Updated to clarify that we'll test against top clouds and top libraries (how to determine top libraries TBD...) and if we determine that it's impossible to be broadly compatible without serving a complete discovery doc, we'll explore what good placeholder values will be.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -29,26 +30,58 @@ replaces:
...
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://github.com/kubernetes/enhancements/issues
+[kubernetes/kubernetes]: https://github.com/kubernetes/kubernetes
+[kubernetes/website]: https://github.com/kubernetes/website
+
 ## Summary
 
 The Kubernetes API server generates (signs) JSON Web Tokens that are meant to
 authenticate Kubernetes service accounts. The API server can indeed authenticate
```
enj: Pre-existing but this should say `Kubernetes service accounts (KSA).` so that the `KSA` acronym is clear.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -97,34 +130,38 @@ If services (other than the API server could authenticate KSA tokens directly:
     authenticate KSA tokens.
 -   Attempt compatibility with OIDC: common libraries that authenticate OIDC
     tokens should be able to authenticate KSA tokens.
-
-As a stretch goal / consideration:
-
 -   Support authentication when the API server is not directly reachable by the
     relying party.
     -   e.g.: a cloud-based service authenticating an API server that doesn't
```
enj: Is this saying that the relying party will be able to reach the KAS OIDC discovery document but not reach the rest of the API?

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -138,16 +175,31 @@ In addition, the API server will serve a JWKS consisting of the public keys from
...
+API server (the user may choose an alternative issuer URL if their cluster does
+not serve on the public internet, as long as the cluster is configured to use
+that URL for issuer claims).
+
 For example, if the API server is configured with the `--service-account-issuer`
-value `https://dev.cluster.internal`, the API server could expose the following
-configuration:
+value `https://cluster.example.com`, which is also the API server root, and
+`--service-account-jwks-uri` value `https://cluster.example.com/openid/v1/jwks`
+(the cluster serves the JWKS at the `/openid/v1/jwks` path by default), the
```
enj: How did we arrive at `/openid/v1/jwks`?

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -138,16 +175,31 @@ In addition, the API server will serve a JWKS consisting of the public keys from
 public keys used for valid KSA tokens. The OIDC discovery document will point to
 the JWKS path as the `jwks_uri`.
 
+Finally, it will be possible for users to override the `jwks_uri` via a new
+`--service-account-jwks-uri` flag. We do attempt to construct a default JWKS URI
```
enj: If I understand this correctly, this is purely to support having a publicly exposed JWKS that is different from the API server external address?  IIRC GKE does this today?

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -195,10 +243,70 @@ configuration:
...
 The API server would treat these as `nonResourceURLs`, and restrict access
-appropriately. We will consider expanding `system:public-info-viewer` RBAC
-ClusterRole to grant access to the new paths; some other roles may have
-permission already via a pattern match on `nonResourceURLs` (e.g.
-`cluster-admin`).
+appropriately. We will provide a default RBAC `ClusterRole` called
+`service-account-issuer-discovery` that provides `GET` access to these
+`nonResourceURLs`, but will *not* provide a default `ClusterRoleBinding`. This
+leaves the decision of who is allowed to access these endpoints up to cluster
+admins. A default binding requires further discussion, including ongoing efforts
```
enj: While I am fine with this as a starting point, I am skeptical that this poses a security risk.  OIDC providers innately expose this type of configuration, and many do so on the public internet (Google, GitLab, Azure, etc).

To me, by not exposing this by default, we just make it hard for relying parties to consume this functionality agnostically of how the cluster was provisioned.  I think that misses the whole point of this KEP.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -195,10 +243,70 @@ configuration:
...
+
+### Test Plan
+
+This feature includes unit, integration, and E2E tests:
+- Unit tests in `pkg/serviceaccount` to test that the server code constructs
+  OIDC responses in the correct format.
+- Integration tests in `test/integration/auth/svcacct_test.go` to attempt a
+  token verification based on the OIDC flow against an API server, and also
+  to verify the expected headers on the responses.
+- An E2E test that exercises the full stack by mounting a new-style Kubernetes
```
enj: An e2e test that confirms a pod running in cluster A can validate a KSA from cluster B would also be interesting.  Maybe too much of a hassle to set up, but I think it is a valid scenario to consider.

keps/sig-auth/20190730-oidc-discovery.md
liggitt: * /openid to group any future endpoints related to this
* /v1 to version it so if format or structure or spec changes, we can publish a second version (also makes it clearer where this falls under the deprecation policy)
* /jwks to identify particular content

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -195,10 +243,70 @@ configuration:
...
+`service-account-issuer-discovery` that provides `GET` access to these
+`nonResourceURLs`, but will *not* provide a default `ClusterRoleBinding`. This
+leaves the decision of who is allowed to access these endpoints up to cluster
+admins. A default binding requires further discussion, including ongoing efforts
+to harden the unauthenticated API surface area.
+
+### Risks and Mitigations
+
+- Security is being reviewed by @mikedanese and @liggitt.
+- This feature exposes public key information that is derived from sensitive
```
enj: Can we have a test to confirm no private keys are emitted via the JWKS URI?

keps/sig-auth/20190730-oidc-discovery.md
enj: SGTM

keps/sig-auth/20190730-oidc-discovery.md
mikedanese: While this endpoint appears to be fairly benign, we've thought the same thing about other permissions in public-info-viewer and been burned. Now benchmarks like CIS recommend blanket turning off anonymous auth which is an outcome but not a good one. In the future, we should maintain a very high bar when granting permissions to anonymous users.

keps/sig-auth/20190730-oidc-discovery.md
```diff
@@ -97,34 +130,38 @@ If services (other than the API server could authenticate KSA tokens directly:
...
-...but not OIDC *compliance*.
+Our goal is *not* OIDC *compliance*.
 
-We aren't trying to: - Make the KSA token process fully compliant with OIDC
-specifications. - OIDC includes flows for token acquisition, token exchange,
-getting user data out of tokens, etc. - But we're only interested in the parts
-relevant to *relying parties*, i.e. token authentication. - We don't need to do
+We aren't trying to make the KSA token process fully compliant with OIDC
+specifications. OIDC includes flows for token acquisition, token exchange,
+getting user data out of tokens, etc. But we're only interested in the parts
```
mikedanese: Grammar nit: But at the beginning of a sentence.

```suggestion
getting user data out of tokens, etc. We are primarily interested in the parts
```

keps/sig-auth/20190730-oidc-discovery.md
enj: > While this endpoint appears to be fairly benign, we've thought the same thing about other permissions in public-info-viewer and been burned. Now benchmarks like CIS recommend blanket turning off anonymous auth which is an outcome but not a good one. In the future, we should maintain a very high bar when granting permissions to anonymous users.

I understand and respect the desire to limit what unauthenticated users can do.  But OIDC discovery endpoints are explicitly meant to be pre-authentication.  If we are careful in implementation (i.e. keep the endpoint "dumb" and have it just emit bytes with no user input), I think this is far safer than discovery endpoints that proxy for aggregated API servers.  Ideally we are talking about an endpoint only slightly more complex than `/healthz`.  Even if we include a default cluster role binding, the cluster admin can still reconcile protect it to tighten access.

If we want this functionality to be adopted by relying parties, it really needs to be present in all clusters by default.

Again, I am fine leaving this as-is and revisiting once the code matures.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Done

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Yes. For example, something with access to (or responsible for the deployment of) the cluster can cache the responses from the API server in a storage bucket, as long as either (a) the issuer URL is set to the bucket location or (b) the /.well-known path at the issuer URL redirects to the bucket.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Done.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Correct.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Yes, that's a good idea. Added to 0->alpha graduation criteria.

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: Yeah that could be interesting. I imagine if relying parties outside the cluster can validate it, then pods in some other cluster should also be able to as well (as long as they have a way to make a request outside their host cluster).

keps/sig-auth/20190730-oidc-discovery.md
mtaufen: SG

PR #1492, opened 2020-01-22
Update 20190125-out-of-tree-openstack.md

ramineni:
This PR marks the KEP as implementable.
related to : #669

### DIFF
--- a/keps/sig-cloud-provider/20190125-out-of-tree-openstack.md
+++ b/keps/sig-cloud-provider/20190125-out-of-tree-openstack.md
@@ -14,7 +14,7 @@ approvers:
 editor: TBD
 creation-date: 2019-01-25
 last-updated: 2019-01-25
-status: implementable
+status: implemented
 ---
 
 # Supporting Out-of-Tree OpenStack Cloud Provider
--- a/keps/sig-cloud-provider/20190125-out-of-tree-openstack.md
+++ b/keps/sig-cloud-provider/20190125-out-of-tree-openstack.md
@@ -14,7 +14,7 @@ approvers:
 editor: TBD
 creation-date: 2019-01-25
 last-updated: 2019-01-25
-status: provisional
+status: implementable
 ---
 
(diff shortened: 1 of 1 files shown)


ramineni:
cc @adisky

johnbelamaric:
Is this complete or not? If it's complete, it should be `implemented`. If it's still in progress, then `implementable` makes sense.

ramineni:
> Is this complete or not? If it's complete, it should be `implemented`. If it's still in progress, then `implementable` makes sense.

Thanks. Marked as implemented.

keps/sig-cloud-provider/20190125-out-of-tree-openstack.md
```diff
@@ -83,7 +83,7 @@ OpenStack Cloud Provider is well tested, CI running at [OpenLab](https://github.
 
 ### Graduation Criteria
 
-This feature is complete, well tested, in parity with intree openstack provider. Documents needs to be updated as per sig-cloud-provider guideline. 
+This feature is complete, well tested, in parity with intree openstack provider. Documention regarding migration from in-tree cloud provider to openstack-cloud-controller-manager can be found [here](https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/migrate-to-ccm-with-csimigration.md)
```
andrewsykim: Are there also general docs on installing the out-of-tree openstack provider?

andrewsykim: Are there also general docs on installing the out-of-tree openstack provider?

ramineni: @andrewsykim yes, we have documentation regarding installing ccm using kubeadm and also getting started guide on installing the provider on configuration needed for using external provider. https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/using-controller-manager-with-kubeadm.md , https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/getting-started-provider-dev.md

PR #1494, opened 2020-01-22
Add KEP for generic data populators

bswartz:


keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,309 @@
...
+
+### User Stories
+
+There are a long list of possible use cases around generic data populators, and I won't
+try to detail them all here. I will detail a few that illustrate the challenges faced by
+users and developers, but it's important to see these as a few examples among many.
+
+#### VM Images
+
+One use case was relevant to KubeVirt, where the Kubernetes `Pod` object is actually a
```
xing-yang: Add a link to KubeVirt project.

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,309 @@
...
+compatibility, and also perfect for the purposes of a data populator. This allows
+developers to add new types of data sources that the dynamic provisioners will
+simply ignore, enabling a different controller to see these objects and respond
+to them.
+
+The way I expect most data populators to work in practice is that they will see
+a request for a PVC `foo` with a data source of `bar`. The populator will create
+a shadow PVC `foo'` in another namespace with all of the same spec except the
+data source will be empty. The dynamic provisioner will ignore `foo` but will
+respond to `foo'` by creating an empty PV. The populator can create a pod
```
xing-yang: Without taints, how do you prevent PVC and the empty PV from being bound to each other and being used by a pod?

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,309 @@
...
+sidecar. If the external-provisioner sidecar sees a data source it doesn't
+understand, it simply ignores the request, which is both important for forward
+compatibility, and also perfect for the purposes of a data populator. This allows
+developers to add new types of data sources that the dynamic provisioners will
+simply ignore, enabling a different controller to see these objects and respond
+to them.
+
+The way I expect most data populators to work in practice is that they will see
+a request for a PVC `foo` with a data source of `bar`. The populator will create
+a shadow PVC `foo'` in another namespace with all of the same spec except the
```
xing-yang: If another namespace is involved, should this be using the method provided by the namespace transfer KEP?

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,309 @@
...
+
+Since the original design of the `DataSource` API field, we have been aware of the desire to
+populate volumes with data from other sources, both CSI-specific and generic (compatible with
+any CSI plugin). For new CSI-specific data sources, the path forward is clear, but for other
+sources of data, which I call "Generic Data Populators" we don't have a mechanism. The main
+problem is that current API validation logic uses a white list of object types, driven by
+feature gates for each object type. This approach won't scale to generic populators, which
+will by their nature be too numerous and varied.
+
+This proposal recommends that we relax validation on the `DataSource` field to allow
```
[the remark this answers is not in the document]
bswartz: Fortunately, the external-provisioner already ignores PVCs with unknown sources. This is why the approach works today, if you relax the validation in kube-apiserver.

The populator can be implemented entirely outside of the existing sidecars, although it's likely that for some data sources we will want CSI support as well as generic populator support. I have samples of that approach as well. The main point I'm making with this KEP and my sample code is that data population outside of CSI is possible, and generally useful.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: If it's not clear, there are two PVCs -- foo and foo-prime. The PV initially binds to foo-prime. After the PV is populated, we delete foo-prime and rebind the PV to foo.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: Namespace transfer is already easy to do if you have a controller and the appropriate cluster role/RBAC. The example controller demonstrates this.

The purpose of the namespace-transfer feature (as I understand it) is to provide the same functionality to low-privilege users through some kind of standardized interface.

Conceivably, if the namespace transfer feature is added, then we could use that capability to simplify the design of populators, and require less privileges. However it's not required.

bswartz:
> Feedback from @thockin:
> 
>     * Must be feature gated and off by default (alpha).

This is mentioned in the KEP.

>     * Criteria: before moving to beta, must have a implementation of data populators (that addresses the question of how do we allow for a PV/PVC to be provisioned and only made available to a populator pod for "filling" before being made generally available) and API agreed on with API Approvers.

Most of the detail in the KEP focuses on this specific problem. TLDR: we create a second (hidden) PVC to do the populating, and then we rebind the populated PV to the original PVC. There are a few strategies for PV rebinding, but I think I could convince you that there is at least one safe way.

> Feedback from @msau42:
> 
>     * Comment on this field what happens if a type is not supported (provisioning fails).

Currently my proposal is that the PVC would sit there and never bind, similar to what would happen if you used a StorageClass with an invalid provisioner field. I'm open to better ideas in this area. Obviously it would be good to provide feedback to users it they commit some error in specifying their data source. Unfortunately it's hard to tell the difference between "user specified a bad data source" and "the CRD and populator controller for this data source haven't been installed yet".

keps/sig-storage/20200120-generic-data-populators.md
j-griffith: Would you be able to include an example workflow/manifests, you've called out the multiple PVCs but I'm not sure how you're intending that to work from a users perspective, or from the ns transfer perspective.

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,307 @@
...
+to be pre-populated with various OS images. OS images tend to be stored in external
+repositories dedicated to that purpose, often with various mechanisms for retrieving
+them efficiently that are external to Kubernetes.
+
+One way to achieve this is to represent disk images as custom resources that point to
+the image repository, and to allow creation of PVCs from these custom resources such
+that the volumes come pre-populated with the correct data. Efficient population of the
+data could be left up to a purpose-built controller that knows how to get the bits
+where they need to be with minimal I/O.
+
```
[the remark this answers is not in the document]
thockin: The "sync from git" example is really not good for this because it is a perfect application of sidecars or init containers.

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,307 @@
...
+more sense to enable developers to try different approaches by making the API allow
+restoring from various kinds of things. 
+
+### Implementation Details/Notes/Constraints
+
+As noted above, the proposal is extremely simple -- just remove the validation on
+the `DataSource` field. This raises the question of WHAT will happen when users
+put new things in that field, and HOW populators will actually work with so small
+a change.
+
```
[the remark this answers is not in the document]
thockin: Agree

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,307 @@
...
+It's first important to note that only consumers of the `DataSource` field today
+are the various dynamic provisioners, most notably the external-provisioner CSI
+sidecar. If the external-provisioner sidecar sees a data source it doesn't
+understand, it simply ignores the request, which is both important for forward
```
thockin: It seems like ignoring it is wrong - the user expected SOMETHING to happen and instead they get nothing?  It seems like a case of "operation failed successfully".

keps/sig-storage/20200120-generic-data-populators.md
bswartz: Yes the idea is that we don't want end users to see weird PVCs and pods appearing in their namespace, but we need to create at least a pod to do the population work. My proposal also involves a second PVC to work around the problem that we don't want and end-user's pod to race with the populator pod to attach to the PVC first. By doing the population in a different PVC, we avoid that problem. The second namespace serves to just hide what's going on from the end user (assuming the end user can't see all namespaces). The re-binding of the PV is that hardest part of the scheme, but once you have a way to rebind PVs, then moving them across namespaces is trivial because the PV itself is not-namespaced. I will do a separate writeup of PV-rebinding techniques since it seems interesting for a variety of reasons.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: The example I provided is in the hello-populator implementation. I could try to replicate that example in slides to better illustrate how it works, but KEPs don't include slides...

keps/sig-storage/20200120-generic-data-populators.md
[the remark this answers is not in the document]
bswartz: Yes! I thought this was weird too, but I didn't think to mention it.

keps/sig-storage/20200120-generic-data-populators.md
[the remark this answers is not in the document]
bswartz: I can add a 3rd story if there's need to revise this again before the KEP merge deadline.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: I'm happy to add a link, but not sure which link is best. To the main project page? To the github page?

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,307 @@
...
+compatibility, and also perfect for the purposes of a data populator. This allows
+developers to add new types of data sources that the dynamic provisioners will
+simply ignore, enabling a different controller to see these objects and respond
+to them.
+
+The way I expect most data populators to work in practice is that they will see
+a request for a PVC `foo` with a data source of `bar`. The populator will create
+a shadow PVC `foo'` in another namespace with all of the same spec except the
+data source will be empty. The dynamic provisioner will ignore `foo` but will
```
thockin: What prevents the real underlying provisioner from proceeding with both in parallel?  What signal do provisioners need to learn about to "wait for completion" ?

keps/sig-storage/20200120-generic-data-populators.md
thockin: Please - excruciating detail is needed

keps/sig-storage/20200120-generic-data-populators.md
```diff
@@ -0,0 +1,307 @@
...
+respond to `foo'` by creating an empty PV. The populator can create a pod
+attached to `foo'` and fill it with data pointed to by `bar` using whatever
+mechanism the developer chooses. After the volume is populated, the pod can be
+deleted, and the PV can be rebound to `foo`.
```
thockin: We don't have a general model of a PV being re-bound, do we?

keps/sig-storage/20200120-generic-data-populators.md
bswartz: Currently the external-provisioner sidecar (the main dynamic provisioner controller) looks at the `DataSource` field and ignores anything it doesn't understand. This is exactly what we want, because a different controller will be creating the PV for that PVC. We only want controllers to respond to PVCs that have data sources they know how to supply. This allows multiple provisioners/populators to cooperate with eachother.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: Well the funny thing about Kubernetes is that you can do things out of order. You can create a pod that refers to a PVC that doesn't exist yet. Then later you can create the PVC, but refer to a non-existent storage class. Then later you can create the storage class that refers to a provisioner that hasn't been installed yet. Then you can install the provisioner last. Eventually the volume finally gets created and the pod starts.

I agree that generally speaking people expect the system to be installed and configured such that most requests will be satisfied immediately. Given the fail-until-success model used in Kubernetes the best thing approach I can imagine to solve this problem would be for there to be a registry of "valid" data sources types, and if a user specifies a data source not on the list, we could generate an event on the PVC warning them that the data source isn't valid yet. However it wouldn't be a good idea to fail at that point because after installing a controller, the data source could become valid and the PVC could get created. A warning would be desirable though.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: I can make a slide deck on this subject... Let me start here with a description of what the hello-populator example code does.

The hello-populator creates a second PV with all of the same details as the first PV, but bound to the other PVC. This takes advantage to the fact that two different PV objects can point to exactly the same storage. Because the original PV is bound to a PVC that the populator created -- we can ensure that there are definitely no pods attached to the PVC at the time this happens, avoiding any dangers related to multiple attachments when that shouldn't be allowed. And of course we change the retentionPolicy on the original PV to Retain before we delete it so nothing happens to the actual storage while we clean up the temporary k8s objects.

There are other flavors of this -- you can delete the PV first and then re-create it bound to the other PVC. This make recovering from errors a little harder because you have to persist the state needed to clone the PV object somewhere.
You can delete the shadow PVC before cloning its PV, but leave a finalizer on it to keep it around, to make extra sure no new pods can attach to it while there are two PVs pointing to the same storage. I didn't implement this extra safety-step, but it's conceptually not hard. I also did some experiments with simply modifying the PV object directly to force it to rebind. If you set its rentionPolicy to Retain and delete the PVC, you can get the bind controller to think it's okay to bind it again. I forget why, but this method did cause some problems which is why I used the two PV approach for my sample.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: The only case I know of this being done is for non-dynamic PVs when they get recycled. It's not a common thing, but I think it should be for the populator case.

keps/sig-storage/20200120-generic-data-populators.md
bswartz: Actually I just re-created my old experiment where I directly modify the PV to rebind it, and nothing bad happened, so that approach may be preferable.

keps/sig-storage/20200120-generic-data-populators.md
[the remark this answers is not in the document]
j-griffith: @thockin say it's not a good example, so it's not a good example.  Don't waste your time.

keps/sig-storage/20200120-generic-data-populators.md
adityadani: I feel that the PV rebinding logic will be used by most (probably all) data populators. Does it make sense to make this logic common and part of the CSI external-provisioner? Though if we do it, we will need some kind of handshaking/work offloading between CSI and the data populators.

bswartz: While I do think the rebinding logic will be used by most (if not all) populators, I don't see the relation to the external provisioner. The external provisioner did it's job when it created the empty volume, and we don't need it to do anything more at that point. The populator can use ordinary k8s functionality to fill the volume with data, and then rebind the PV to the original PVC. Rebinding isn't a complex operation -- it can be as simple as editing some fields in the PV. If it does turn out that the rebind logic needs to be more complex, I would recommend creating a package to share that code, but that package would still be unrelated to the external-provisioner (IMO).

The one exception to this is for cases when there's a data source type that we want to sometimes have handled by the CSI driver and sometimes be handled by another controller. In those cases, there will need to be agreement on which controller should handle the PVC. Those mechanisms will depend on the exact nature of the data source. I did do a proof-of-concept of this kind of data source, but the code isn't public yet.

saad-ali:
I've asked Ben to simplify this KEP. The purpose of this KEP is to add a alpha gate which permits the loosening of the validation on the DataPopulator field. This is a necessary first step to allow experimentation for DataPopulators. The details of how DataPopulators will work is a huge open question and should not block this KEP right now. This KEP calls out that moving from alpha to beta of this feature will be gated on a full API and design of DataPopulators.

Based on that this LGTM. Will let @thockin take a look before adding LGTM/Approve.

PR #1509, opened 2020-01-25
Update Snapshot KEP for Beta

xing-yang:
This PR updates the snapshot KEP following the Beta release.

### DIFF
--- a/keps/sig-storage/20190709-csi-snapshot.md
+++ b/keps/sig-storage/20190709-csi-snapshot.md
@@ -1,8 +1,9 @@
 ---
-title: CSI Snapshot 
+title: CSI Snapshot
 authors:
   - "@jingxu97"
   - "@xing-yang"
+  - "@yuxiangqian"
 owning-sig: sig-storage
 participating-sigs:
   - sig-storage
@@ -40,8 +41,10 @@ CSI Snapshot
 - [Graduation Criteria](#graduation-criteria)
   - [Alpha-&gt;Beta](#alpha-beta)
   - [Beta-&gt;GA](#beta-ga)
-- [Changes](#changes)
-  - [Changes Implemented](#changes-implemented)
+- [Snapshot Beta](#snapshot-beta)
+  - [API Changes](#api-changes)
+  - [Controller Split](#controller-split)
+  - [Other Changes Implemented](#other-changes-implemented)
   - [Work in Progress](#work-in-progress)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
@@ -80,7 +83,7 @@ The rest of the document includes required information missing from the original
   * Snapshot Object in Use Protection
   * Separate the common controller from the sidecar controller
   * Add secrets field to list-snapshots RPC in the CSI spec. Add “data-source-secret” in create-volume intended for accessing the data source. Implement them in external-snapshotter and external-provisioner.
-  * Add metrics support
+  * Add metrics support to the external-snapshotter sidecar.
 * Unit and e2e tests implemented
 * Update snapshot CRDs to v1beta1 and enable VolumeSnapshotDataSource feature gate by default.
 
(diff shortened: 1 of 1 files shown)


keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -119,12 +244,14 @@ There are other things we are working on for Beta:
     labels:
     * operation_name, i.e., creation-snapshot, delete-snapshot
     * csi-driver-name
-* Update snapshot CRDs to v1beta1 and enable VolumeSnapshotDataSource feature gate by default.
 
 ## Implementation History
 
-Volume snapshot is implemented as alpha feature in this repo in Kubernetes v1.12:
+Volume snapshot is implemented as alpha feature in this repo in Kubernetes v1.12 and is promoted to beta in Kubernetes 1.17:
```
msau42: I would probably just structure this section as:

```
K8s 1.12: alpha
K8s 1.17: beta
```

I'm not sure we need the actual feature gate PRs here

xing-yang: Updated.

msau42:
lgtm, will let @yuxiangqian review as well

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
 * Feature deployed in production and have gone through at least one K8s upgrade.
 
-## Changes
-
-### Changes Implemented
+## Snapshot Beta
+
+### API Changes
+
+A number of changes were made to the Kubernetes volume snapshot API between alpha to beta. These changes are not backward compatible and the alpha API is no longer supported. The purpose of these changes was to make API definitions clear and easier to use.
```
yuxiangqian: "make API more clear and easier to use"

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+The following changes are made:
```
yuxiangqian: "The following changes have been made from the Alpha API:"

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+* DeletionPolicy is now a required field rather than optional in both VolumeSnapshotClass and VolumeSnapshotContent. This way the user has to explicitly specify it, leaving no room for confusion.
+* VolumeSnapshotSpec has a new required Source field. Source may be either a PersistentVolumeClaimName (if dynamically provisioning a snapshot) or VolumeSnapshotContentName (if pre-provisioning a snapshot).
+* VolumeSnapshotContentSpec also has a new required Source field. This Source may be either a VolumeHandle (if dynamically provisioning a snapshot) or a SnapshotHandle (if pre-provisioning volume snapshots).
+* VolumeSnapshotStatus now contains a BoundVolumeSnapshotContentName to indicate the VolumeSnapshot object is bound to a VolumeSnapshotContent.
+* VolumeSnapshotContentnow contains a Status to indicate the current state of the content. It has a field SnapshotHandle to indicate that the VolumeSnapshotContent represents a snapshot on the storage system.
```
yuxiangqian: "It has a field SnapshotHandle which is the unique identifier of a snapshot on the storage system"

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+        Driver string
+        Parameters map[string]string
+        DeletionPolicy DeletionPolicy
+}
+```
+
+### Controller Split
+
+When Volume Snapshot is promoted to Beta in Kubernetes 1.17, the CSI external-snapshotter sidecar controller is split into two controllers: a snapshot-controller and a CSI external-snapshotter sidecar.
```
yuxiangqian: "Along with VolumeSnapshot being promoted to Beta in Kubernetes 1.17, the CSI external-snapshotter sidecar controller has been split into two controllers:"

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+The snapshot controller will be watching the Kubernetes API server for `VolumeSnapshot` and `VolumeSnapshotContent` CRD objects. The CSI `external-snapshotter` sidecar only watches the Kubernetes API server for `VolumeSnapshotContent` CRD objects.
```
yuxiangqian: also watch for VolumeSnapshotClass objects

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+The creation of a new `VolumeSnapshot` object referencing a `SnapshotClass` CRD object corresponding to this driver causes the snapshot controller to trigger the creation of a Kubernetes `VolumeSnapshotContent` object to represent the to-be-created new snapshot.
```
yuxiangqian: this is only for dynamic provisioning, please add the clarification

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+The creation of a new `VolumeSnapshotContent` object causes the sidecar container to trigger a `CreateSnapshot` operation against the specified CSI endpoint to provision a new snapshot. When a new snapshot is successfully provisioned, the sidecar container updates the status field of the `VolumeSnapshotContent` object to represent the new snapshot.
+
+The snapshot controller will be updating the status field of the `VolumeSnapshot` object accordingly based on the status field of the `VolumeSnapshotContent` object to indicate the new snapshot is ready to be used.
```
yuxiangqian: or failed

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
+
+The deletion event of a `VolumeSnapshot` object bound to a `VolumeSnapshotContent` corresponding to this driver with a `delete` deletion policy causes the snapshot controller to start deleting the `VolumeSnapshotContent` object and add an annotation to the object to indicate it is being deleted. Note that both the `VolumeSnapshot` object and the `VolumeSnapshotContent` object will not be deleted immediately due to the finalizers. When the sidecar container detects this update on the `VolumeSnapshotContent` object, it triggers a `DeleteSnapshot` operation against the specified CSI endpoint to delete the snapshot. Once the snapshot is successfully deleted, the sidecar container removes the finalizer on the `VolumeSnapshotContent` object which leads to the deletion of the object from Kubernetes. The snapshot controller then removes the finalizer on the `VolumeSnapshot` object and as a result the object will be deleted from Kubernetes.
```
yuxiangqian: I am not sure how much details we need to expose in this type of documents, do we need to cover all major cases? for example, what if the deletion policy is retain?
@msau42 @xing-yang

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,150 @@ The rest of the document includes required information missing from the original
...
 
 Here are the changes since the original design proposal:
 
 * Renamed `Ready` to `ReadyToUse` in the `Status` field of `VolumeSnapshot` API object.
 * Changed type of `RestoreSize` in `CSIVolumeSnapshotSource` from `*resource.Quantity`  to `*int64`.
 * Lease based Leader Election support is added.
 * Added `VolumeSnapshotContent` deletion policy which is also specified in `VolumeSnapshotClass`.
-* Added `VolumeSnapshot` and `VolumeSnapshotContent` in Use Protection using Finalizers.
 * Added Finalizer on the snapshot source PVC to prevent it from being deleted when a snapshot is being created from it.
+* Added Finalizer on the `VolumeSnapshotContent` object to prevent it from being deleted when it is bound to the `VolumeSnapshot` object.
```
yuxiangqian: "from being deleted directly from API server"

keps/sig-storage/20190709-csi-snapshot.md
xing-yang: I added a sentence to briefly describe what will happen if the deletion policy is retain.

yuxiangqian: I think it should be fine as long as we have detailed documentation in the external-snapshot repo. In this KEP probably its OK to keep it on high level.
I'd suggest a nit to add a small piece to elaborate what to expect if a user deletes a bound "VolumeSnapshotContent" object directly, it will have a deletion timestamp set however will persist in API server until its corresponding "VolumeSnapshot" object gets deleted.

xing-yang: Done

xing-yang:
@yuxiangqian addressed your comment.  PTAL.

keps/sig-storage/20190709-csi-snapshot.md
```diff
@@ -89,28 +92,152 @@ The rest of the document includes required information missing from the original
...
+
+The deletion event of a `VolumeSnapshot` object bound to a `VolumeSnapshotContent` corresponding to this driver with a `delete` deletion policy causes the snapshot controller to start deleting the `VolumeSnapshotContent` object and add an annotation to the object to indicate it is being deleted. Note that both the `VolumeSnapshot` object and the `VolumeSnapshotContent` object will not be deleted immediately due to the finalizers. When the sidecar container detects this update on the `VolumeSnapshotContent` object, it triggers a `DeleteSnapshot` operation against the specified CSI endpoint to delete the snapshot. Once the snapshot is successfully deleted, the sidecar container removes the finalizer on the `VolumeSnapshotContent` object which leads to the deletion of the object from Kubernetes. The snapshot controller then removes the finalizer on the `VolumeSnapshot` object and as a result the object will be deleted from Kubernetes. If a user deletes a bound `VolumeSnapshotContent` object directly, it will have a deletion timestamp set however will persist in API server until its corresponding `VolumeSnapshot` object also gets a deletion timestamp set from a deletion request.
```
msau42: Why is deletion timestamp not sufficient to determine if the object is being deleted? Why do you also need an annotation?

xing-yang: Because the sidecar does not have access to the VolumeSnapshot API object so it can't check if VolumeSnapshot has a deletion timestamp.  The sidecar checks the VolumeSnapshotBeingDeleted annotation to make sure that it is a deletion request on the VolumeSnapshot.

msau42: How come deletion timestamp on the VolumeSnapshotContent object is not sufficient?

xing-yang: We changed the logic in the beta version. 
 
In the Alpha version, we followed PV/PVC logic.  If a user tries to delete VolumeSnapshot, the VolumeSnapshot API object will be deleted immediately if it is not used by a PVC.  For VolumeSnapshotContent, if there's a deletion timestamp and VolumeSnapshot no longer exists, the controller will try to delete the physical snapshot and VolumeSnapshotContent.  However this could potentially delete the physical snapshot by accident when user copies the content yaml to another environment and cause data loss.

In Beta version, VolumeSnapshot will not be deleted before VolumeSnapshotContent and the physical snapshot resource.  We want to make sure that the deletion request is coming from a request to delete the VolumeSnapshot, not directly on VolumeSnapshotContent.  If we are only checking deletion timestamp on VolumeSnapshotContent, then it could leave VolumeSnapshot object behind.

msau42: Isn't there a finalizer on the Content object that only gets removed when the Snapshot object is deleted? (PV has a finalizer that only gets removed when PVC is gone)

msau42:
Merging this for now. But I think we should revisit the finalizers + annotations and see if we can simplify anything. Right now it seems a bit brittle and difficult to get correct.

/lgtm
/approve

PR #1512, opened 2020-01-27
Move raw block feature to GA in 1.18

jsafrane:
/assign @msau42

### DIFF
--- a/keps/sig-storage/20190130-csi-raw-block-support.md
+++ b/keps/sig-storage/20190130-csi-raw-block-support.md
@@ -41,6 +41,7 @@ see-also:
   - [K8s 1.13: Alpha](#k8s-113-alpha)
   - [K8s 1.14: Beta](#k8s-114-beta)
   - [K8s 1.17: Beta](#k8s-117-beta)
+  - [K8s 1.18: Targeting GA](#k8s-118-targeting-ga)
 <!-- /toc -->
 
 ## Summary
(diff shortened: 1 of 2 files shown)


keps/sig-storage/20190130-csi-raw-block-support.md
```diff
@@ -162,3 +163,5 @@ and we can test with them
 * Separated NodeStage / NodePublish calls.
 * Fixed volume reconstruction after kubelet restart.
 * Stress tests as noted above.
+
+### K8s 1.18: GA
```
msau42: Add a note that 1.18 is the target? Same below

jeremyrickard: Hey @msau42 @jsafrane can you link to which KEP issue#s this relates to so we can better track? 

Thanks!

jsafrane: There is https://github.com/orgs/kubernetes-csi/projects/4, not sure it should be part of the KEP though.

jsafrane: @msau42, not sure I understood, changed to `K8s 1.18: Targeting GA`

msau42: I guess there's a title for this section, but no content under it. Is there anything we should be adding to the section? More test cases/corner cases to fix/test?

jsafrane: Added this:
> * Disruptive testing with block devices in /dev reordered after reboot.

PR #1536, opened 2020-01-31
Update CSR test plan

liggitt:
Adds a test plan to the CSR KEP

/cc @mikedanese 
fyi @jeremyrickard

### DIFF
--- a/keps/sig-auth/20190607-certificates-api.md
+++ b/keps/sig-auth/20190607-certificates-api.md
@@ -32,8 +32,9 @@ status: implementable
   - [Manual CSR Approval With Kubectl](#manual-csr-approval-with-kubectl)
   - [Automatic CSR Approval Implementations](#automatic-csr-approval-implementations)
   - [Automatic Signer Implementations](#automatic-signer-implementations)
+  - [Test Plan](#test-plan)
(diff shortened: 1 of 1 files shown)


keps/sig-auth/20190607-certificates-api.md
```diff
@@ -379,9 +380,31 @@ that uses a webhook to sign all approved CSRs. This allows the root certificate
...
+- integration tests of:
+  - auto-approve and signing of kubelet client CSR
+  - manual-approve and auto-signing of kubelet serving CSR
+  - manual-approve and auto-signing of API client CSR
+  - manual-approve and auto-signing of legacy CSR
+
+- e2e (and conformance, once v1) tests of:
+  - CSR API presence, basic crud behavior
+  - status subresource
+  - approval flow
```
deads2k: Approval as general concept isn't owned by kubernetes.  Neither is signing.  An e2e test for what we ship makes a lot of sense.  I'm less convinced that non-kube owned code should have conformance tests.

liggitt: scoped to a custom signerName, to ensure the mechanics work as designed and are usable by other approvers/signers

PR #1542, opened 2020-02-07
CHPA: Until merged, the state should be implementable

gliush:
We set incorrect state of the KEP. This PR fixes that.


gliush:
bump @mwielgus

PR #1549, opened 2020-02-11
kubeadm remove clusterstatus KEP: Make minor clarifications

dlipovetsky:
/assign @fabriziopandini


dlipovetsky:
/hold

My apologies, I confused myself. The KubeadmControlPlane controller doesn't use ClusterStatus as the source of truth--it checks the etcd health by talking to each static Pod. However, the controller will have to update ClusterStatus when removing a node, because it will not run (or, if the node is offline, cannot run)  `kubeadm reset`.

I will update my comments to reflect this.

neolit123:
/remove-lifecycle stale
@fabriziopandini @dlipovetsky hi, is this PR still needed?

dlipovetsky:
Thanks for reminding me. I removed the part about Cluster API. Should be correct now.

PR #1575, opened 2020-02-29
modify the error url of cloud-controller-manager.md

timyinshi:
/kind bug
/sig docs

the  error url is: https://github.com/kubernetes/community/blob/master/keps/sig-cloud-provider/0002-cloud-controller-manager.md

the right url is: https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/20180530-cloud-controller-manager.md


timyinshi:
cla/linuxfoundation

PR #1638, opened 2020-03-28
the ref.go has been deleted

timyinshi:
the ref.go file has been deleted

/kind bug
/sig docs

### DIFF
--- a/keps/sig-api-machinery/20190711-remove-selflink.md
+++ b/keps/sig-api-machinery/20190711-remove-selflink.md
@@ -134,8 +134,6 @@ I went through a k/k repo (including its staging repos) and all repos under [kub
 and this is the list of places that reference `SelfLink` fields (I excluded tests and all places
 in apiserver responsible for setting it):
 
-- [ ] https://github.com/kubernetes/kubernetes/blob/master/pkg/api/ref/ref.go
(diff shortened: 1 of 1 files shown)


lavalamp:
Did the contents get moved, though?

keps/sig-api-machinery/20190711-remove-selflink.md
```diff
@@ -134,8 +134,6 @@ I went through a k/k repo (including its staging repos) and all repos under [kub
 and this is the list of places that reference `SelfLink` fields (I excluded tests and all places
 in apiserver responsible for setting it):
 
-- [ ] https://github.com/kubernetes/kubernetes/blob/master/pkg/api/ref/ref.go
```
wojtek-t: There were two copies of that file in the codebase - this one and staging/src/k8s.io/client-go/tools/reference/ref.go

I migrated stuff to the latter and removed this one.

wojtek-t: I wouldn't removed it from the KEP though - it was there when the KEP was created.

timyinshi: ok,this is fine.

PR #1650, opened 2020-03-30
fix broken url:https://github.com/kubernetes/kubernetes/blob/master/p…

ttonline6:
…kg/kubectl/cmd/get/get.go


thockin:
The problem with embedding links to code in a KEP is that when that code inevitably moves, that KEP becomes broken. There are a bunch of PRs open to fix links. Some change to "master", some change to branches, and none with any particular rhyme or reason that I can discern.

I think we should just leave old KEPs alone. My inclination is to close all of these PRs - tell me why I should not?

PR #1689, opened 2020-04-15
Dynamic Authentication Config

enj:


enj:
/hold

To prevent accidental merge.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+prefix config to be explicitly opted out of by the cluster admin.
+
+#### Ordering of authentication modes
+
+To prevent interference with other authentication modes configured on the API
+server, all `AuthenticationConfig` objects will be considered after the other
+authenticators have had a chance to run.  This prevents issues such as a
+misbehaving authenticator attempting to authenticate a Kubernetes service
+account bearer token.  It also prevents performance issues caused by invoking
+a token webhook for the built-in authentication methods.
```
micahhausler, line 302:
```suggestion
a token webhook before the built-in authentication methods.
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+### Non-Goals
+
+<!--
+What is out of scope for this KEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+1. Changing the internal behavior of the authentication modes (ex: using
+different fields in x509 certificates to build the `user.Info` object)
+2. Restricting the REST API to the exact structure of the CLI flags
```
micahhausler, line 220: What about preservation of the existing authn flags? I can see paranoid admins who would be afraid of allowing dynamic authn and prefer not to use the API.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+a token webhook for the built-in authentication methods.
+
+The ordering of the `AuthenticationConfig` objects in relation to each other is
+(this is based on the performance of each type):
+
+1. `x509` type
+2. `oidc` type
+3. `webhook` type
+
+`AuthenticationConfig` objects of the same type will be ordered lexicographically.
```
micahhausler, line 311: So multiple OIDC and multiple webhooks can be configured simultaneously?

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above.
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+TBD.
+
+### Risks and Mitigations
```
micahhausler, line 267: This new API is obviously admin-level. Are there any gaps in RBAC to managing this API? Ex: "I only want to allow Alice to configure x509 and Bob to configure OIDC" Is that kind of case just delegated to admission?

What about persistant connections? If I remove an x509 config, and there is a client using a cert from that CA with an open connection, what happens?

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+notes and implementing the feature itself.  KEP editors, SIG Docs, and SIG PM
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+Authentication is Kubernetes is quite flexible and unopinionated.  There is no
+requirement for the API server to understand where an identity originates from.
+The API server supports a variety of command line flags that enable authentication
```
micahhausler, line 152: For my benefit, has this idea been discussed prior? What were the objections? It seems like webhook AuthN would have been a natural fit to start with an API instead of an API flag, but wasn't.

enj: @liggitt and @mikedanese may be able to provide some history.

Looking at OIDC as an example, it was added in 2015 via https://github.com/kubernetes/kubernetes/pull/10957

My guess is that it was probably just easier to start with in-tree functionality that was wired through a CLI flag.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: Yes.  The current limitation of one OIDC and one token webhook is a by product of the CLI flag based config.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: > What about preservation of the existing authn flags?

Not sure what you mean.  I am not proposing we change the CLI flags in any way.

> I can see paranoid admins who would be afraid of allowing dynamic authn and prefer not to use the API.

I would assume the paranoid admin would not give users cluster-admin access?

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: > This new API is obviously admin-level.

Yeah the assumption is cluster-admin level access.

> Are there any gaps in RBAC to managing this API? Ex: "I only want to allow Alice to configure x509 and Bob to configure OIDC"

While we could certainly build custom permission checks per `spec.type`, I am not sure that would be meaningful since you can assert basically any identity from x509 or OIDC.  You would just manufacture an identity that had enough authz to access the other `spec.type`.

> Is that kind of case just delegated to admission?

For super custom stuff, yes.  But we have plenty of flexibility via the REST storage layer to add restrictions if we deem them to be valuable.

> Is our only answer to preventing privilege escalation through the API "use RBAC and Admission"?

I think that is generally true for such APIs, though if there are ways we can add meaningful protection directly built-in to k/k, then I am all for that instead of asking people to write custom admission webhooks.

> I just want to be conscious that we're introducing a new privilege escalation vector.

I am not against following a similar pattern to RBAC's cluster role aggregation, i.e. requiring full cluster-admin access:

```go
var fullAuthority = []rbac.PolicyRule{
	rbac.NewRule("*").Groups("*").Resources("*").RuleOrDie(),
	rbac.NewRule("*").URLs("*").RuleOrDie(),
}

func (s *Storage) Create(ctx context.Context, obj runtime.Object, createValidatingAdmission rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
	if rbacregistry.EscalationAllowed(ctx) || rbacregistry.RoleEscalationAuthorized(ctx, s.authorizer) {
		return s.StandardStorage.Create(ctx, obj, createValidatingAdmission, options)
	}

	clusterRole := obj.(*rbac.ClusterRole)

	// ...

	// to set the aggregation rule, since it can gather anything, requires * on *.*
	if hasAggregationRule(clusterRole) {
		if err := rbacregistryvalidation.ConfirmNoEscalationInternal(ctx, s.ruleResolver, fullAuthority); err != nil {
			return nil, apierrors.NewForbidden(groupResource, clusterRole.Name, errors.New("must have cluster-admin privileges to use the aggregationRule"))
		}
	}

	return s.StandardStorage.Create(ctx, obj, createValidatingAdmission, options)
}
```

> What about persistant connections? If I remove an x509 config, and there is a client using a cert from that CA with an open connection, what happens?

I am not 100% sure on this one but this problem already exists today since we dynamically reload CA bundles.  I _think_ we close connections.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: Something to consider, if we gave this API its own group, then we could use `--runtime-config` to allow disabling of that single API.  That might be overkill / an undesired approach.

A better approach may be to allow configuring of admission webhooks via CLI flags so that an infrastructure provider could guard some subset of admission plugins from the cluster-admin (presumably that cluster-admin would not have direct access to the API server host).

Something we could do today is include an admission plugin that is disabled by default that, once enabled, would reject any `create` and `delete` requests (unless `deletionTimestamp` was set) to this API and only allow `updates` to via `IsOnlyMutatingGCFields`.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: Note also that if you have cluster-admin level access, you can just run an impersonation proxy / use impersonation to create any identity, similar to this API.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: > I just mean I wouldn't want to see existing CLI flags deprecated in favor of this.

Yeah, I do not see us changing the CLI functionality in any way since the CLI flags allow for static, unchangeable config.  IMO there is enough value in the CLI flags that even with this new functionality, we would not attempt to deprecate the flags.

mikedanese:
Could this functionality be implemented entirely out of tree using a front-proxy configured by CRDs?

enj:
> Could this functionality be implemented entirely out of tree using a front-proxy configured by CRDs?

Meaningfully?  No.

By front-proxy I assume you mean impersonating proxy (and not like an aggregated API proxy).

The goal here is to allow end users that have `cluster-admin` access to Kubernetes API endpoints (possibly across different infrastructure providers) but not direct host access to the API server to have meaningful control over the authentication that the system uses (as a purely additive extension on whatever authn is already configured).

Some thoughts on the proxy approach:

1. You must abuse the impersonation API
1. You must run the proxy as cluster-admin
1. Nothing is aware of this proxy because it is unrelated to the Kubernetes API.  Thus you have to build some discovery mechanism.  Ingress implementations vary across providers so have to handle the dynamic hostname or use some DNS configuration elsewhere.
1. You must configure and provision TLS certs for this proxy
1. Network traffic for some subset of requests must now flow through the proxy
1. It is unclear how to handle requests incoming to the proxy that attempt to use impersonation
1. You have to somehow tell some subset of your users to use the proxy (instead of the Kube API directly)
1. Proxies exist [[1]](https://github.com/jetstack/kube-oidc-proxy)[[2]](https://github.com/ericchiang/kube-oidc#kube-oidc-proxy) but they are not meaningful solutions (anything that uses impersonation is just a hack IMO)

With a built-in API:

1. Network traffic continues to go to a single place (the Kubernetes API)
1. No extra components to run unless you need a custom webhook or want to run your own OIDC server (instead of using something like `gitlab.com`)
1. For `x509` and `oidc`, the verification is 100% local to the API server (high performance, high trust)
1. It always works across any Kubernetes cluster regardless of infrastructure provider
1. It encourages the implementation of opinionated Kube authn stacks that can be used across infrastructure providers (i.e. folks could build apps that rely on some assumptions about the underlying authn stack because that authn stack is actually portable)

While I am all for implementations outside of core, this functionality is akin to RBAC and admission webhooks.  RBAC is built-in and fully configurable via the Kube API.  This has lead to its adoption and common usage.  It is also a relatively limited API.  It is effectively the equivalent of `x509` and `oidc` in this proposal (i.e. super easy to use and rely upon but not very flexible).  Admission webhooks allow the creation of arbitrary policy extension points.  They are super flexible but put a high burden on the end user.  They are the equivalent of `webhook` in this proposal.

Just as RBAC and admission webhooks would not be meaningful if built out of core, the same is true for this proposal.  If one could not rely on RBAC and admission webhooks being consistently present across all infrastructure providers, they would be useless.

ritazh:
cc @weinong for review

seh:
It's mentioned with some subtlety in the "Motivation" section, but this capability would unlock use of many hosted Kubernetes offerings by organizations that need their own authentication system. As it stands today, operators must choose between owning the full cluster provisioning and operation in order to retain control of the API server command-line flags, or giving up on their custom authentication system to buy into a hosted offering. Azure Kubernetes Engine is one counterexample, where it looks at least _possible_ to jam in the right API server command-line flags, perhaps to the surprise of the maintainers.

I had heard in the past (from @liggitt, if I recall correctly) that we couldn't offer a subsystem like this, because it's not possible to govern who can configure it, because authentication is a prerequisite to determine who is even asking to configure it. Does the design mandate that any user with "cluster admin" permission can manipulate these _AuthenticationConfig_ objects?

enj:
> this capability would unlock use of many hosted Kubernetes offerings by organizations that need their own authentication system. As it stands today, operators must choose between owning the full cluster provisioning and operation in order to retain control of the API server command-line flags, or giving up on their custom authentication system to buy into a hosted offering.

I will try to distill this down into a sentence and add as an explicit goal.

> it's not possible to govern who can configure it, because authentication is a prerequisite to determine who is even asking to configure it.

There is an inherent assumption that you have an existing identity that has `cluster-admin` access.  This is trivial to create with a cert based user in the `system:masters` group (this handles both authn and authz).  Every Kube cluster has to provide you with some baseline way to authenticate to it, otherwise you would never be able to do anything.  Note that RBAC is the same way - something has to give you `cluster-admin` level access so that you can, for example, create new roles for custom resources.

> Does the design mandate that any user with "cluster admin" permission can manipulate these _AuthenticationConfig_ objects?

Any user with `cluster-admin` access can basically do anything.  This API is not materially different from the `cluster-admin` "creating" identities using the impersonation API.  Unlike impersonation, we can create stronger restrictions similar to the RBAC API.  For example, `write` access to this API would require:

1. A user in the `system:masters` group, or
1. `escalate` verb on `authenticationconfigs.authentication.k8s.io`, or
1. `*` verb on `*.*`

A `cluster-admin` can always run the following command:

`kubectl create clusterrolebinding --group=system:unauthenticated --clusterrole=cluster-admin turn-off-security`

Just as there is no meaningful way for the RBAC API to guard against a `cluster-admin`, the same is true for the `AuthenticationConfig` API.

seh:
Thank you for clarifying. I now see that @micahhausler had asked a similar question earlier (https://github.com/kubernetes/enhancements/pull/1689#discussion_r408968321). I missed that yesterday, and should have read more carefully to avoid repeating the same point.

I hope the Webhook support makes it in before too long. (It's commented out for now in the `AuthenticationConfigSpec` struct.)

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+  AuthenticationConfigTypeX509    AuthenticationConfigType = "x509"
+  AuthenticationConfigTypeOIDC    AuthenticationConfigType = "oidc"
+  AuthenticationConfigTypeWebhook AuthenticationConfigType = "webhook"
+)
+
+type X509Config struct {
+  // caBundle is a PEM encoded CA bundle used for client auth (x509.ExtKeyUsageClientAuth).
+  // +listType=atomic
+  // Required
+  CABundle []byte `json:"caBundle" protobuf:"bytes,1,opt,name=caBundle"`
```
anguslees, line 386: Please, *please* avoid the mistake of the aggregated apiserver api and make this `caBundle` a pointer to a separate Secret (or ConfigMap) and not inline.  The rest of the AuthenticationConfig is likely to be static (eg: helm) manifest, but the caBundle contents will be unique for every install.  This is _way_ easier to generate (and rotate) if it is a separate object (eg: you could just use cert-manager to maintain a self-signed CA).

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+will be stored in an `atomic.Value`.
+
+A new `authenticatorfactory.CAContentProvider` will be required to support
+dynamic certificate based authentication on the API server (via
+`BuiltInAuthenticationOptions`) and the kubelet (via `DelegatingAuthenticatorConfig`).
+This `CAContentProvider` will look similar to the dynamic authenticator except
+that it will build and maintain a concatenation of all of the CA bundles from
+the `AuthenticationConfig` objects with `spec.type` equal to `x509`.
+
+Question: should we add white listing support to only allow certain identities?
```
anguslees, line 422: Question: What happens with cached tokens when config changes?  (probably just keep them cached and let them expire naturally)

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+Most of the functionality required to create the dynamic authenticator already
+exists in the API server (to read and process the CLI flags).  It simply needs
+to be wired up to an informer that triggers a rebuild of an authenticator that
+will be stored in an `atomic.Value`.
+
+A new `authenticatorfactory.CAContentProvider` will be required to support
+dynamic certificate based authentication on the API server (via
+`BuiltInAuthenticationOptions`) and the kubelet (via `DelegatingAuthenticatorConfig`).
+This `CAContentProvider` will look similar to the dynamic authenticator except
+that it will build and maintain a concatenation of all of the CA bundles from
```
anguslees, line 419: Erm. How do we provide ordering here if they're all just smooshed together?  Particularly wrt flags vs dynamic config?

In particular, a broken dynamic x509 provider should not be able to affect the cli-configured x509 provider that you need to use to _fix_ the broken authconfig.

keps/sig-auth/1688-dynamic-authentication-config/README.md
seh: I agree: This turned out to be the most annoying part of deploying an admission Webhook.

I vaguely recall hearing the reason why that CA certificates was defined statically, but I can't find the discussion now in Slack. I think it had something to do with not wanting the API server to have to watch the _Secrets_ to reconfigure its Webhook client, though of course the API server already has to watch the Webhook configuration itself.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+Consider including folks that also work outside the SIG or subproject.
+-->
+
+#### Restrictions on asserted identities
+
+To prevent confusion with identities that are controlled by Kubernetes, the
+`system:` prefix will be disallowed in the username and groups contained in the
+`user.Info` object.  A disallowed username will cause authentication to fail.
+All disallowed groups will be filtered out.
```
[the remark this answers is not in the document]
enj: The assumption is that you would create your own `my-admins` group and assign it `cluster-admin` via RBAC.  There is no need to abuse `system:` identites.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+Since authorization modes such as RBAC are unaware of the origin of an identity,
+care must be taken to avoid ambiguity between identities asserted by different
+authenticators.  The `AuthenticationConfig` API will require username and groups
+prefix config to be explicitly opted out of by the cluster admin.
+
+#### Ordering of authentication modes
+
+To prevent interference with other authentication modes configured on the API
+server, all `AuthenticationConfig` objects will be considered after the other
```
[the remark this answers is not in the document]
enj: > how we recover a cluster from a bad AuthenticationConfig

What does it mean to be bad?  Typo in the issuer URL?

> now we need an explicit break-glass mechanism to escape the circular dependency.

Not really.  You have to have a `cluster-admin` identity before you can use this API in the first place.  That identity must come from the CLI flags.  So you would just use that to fix anything you messed up.  This is no different than if you delete all RBAC data and need to use a `system:masters` identity to fix the cluster.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+TBD.
+
+## Alternatives
```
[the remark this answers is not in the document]
enj: >     1. Define a standard (json/yaml) config, but require cloud-providers to expose it as an apiserver flag/config-file via cloud-specific out-of-band mechanisms (ie: cloud-provider's own api).  Avoids any concerns about in-band auth escalation or circular dependency - but also requires a cloud-specific entry point.  I guess this is basically the current situation, except with a more explicit statement saying "we expect providers to expose this part of the apiserver config".

The Kubernetes API is how we get cloud providers to agree on standard config.

>     2. Just mandate token webhook only and remove x509+oidc flags.

`x509` and `oidc` have clear benefits for the simple use cases and they do not leave the API server for verification.  I do not see us removing the OIDC flags from the API server in favor of webhook only.  I do not know why I would make that decision here.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: I touched upon it slightly in the testing section:

```yaml
1. Token caching semantics work (ex: cache is not thrown away on informer resync)
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: > Erm. How do we provide ordering here if they're all just smooshed together? Particularly wrt flags vs dynamic config?

All the dynamic stuff is always after the static stuff.  Thus the flags would always go first.

> In particular, a broken dynamic x509 provider should not be able to affect the cli-configured x509 provider that you need to use to _fix_ the broken authconfig.

You cannot have a broken `x509` provider because the config can be statically validated.  Even without that, invalid PEM bundles would just be ignored.  There is no ordering requirements for CA bundles - any bundle that verifies the cert is fine.  The identity comes from the cert so the CA bundles themselves do not matter.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: I do not plan to diverge from the approach used by existing APIs here.  If they expand to allow references, I will mirror that here.  Overall I am not too worried about the `x509` case.  It is not really an extension point I expect to see used in helm charts.

For `webhook` I agree that this adds pain, but again, I do not plan on diverging from the existing APIs here.  Long term I would like Kube to provide first class support for a short-lived, auto rotated CA for the service network (and mint serving certs for all services that ask for one).  Then the common case of "I just want to run this webhook on the cluster please make the TLS nonsense go away" would be painless.

cc @munnerz since cert-manager is mentioned.

enj:
cc @JoshVanL @simonswine @mhrabovcin @mattbates @phanama @wallrj I see that you all contributed to `jetstack/kube-oidc-proxy` and thus may be interested in this KEP.

keps/sig-auth/1688-dynamic-authentication-config/README.md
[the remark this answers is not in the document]
mvladev: You can already achieve (abuse) this via the `CertificateSigningRequest` API and there is no such restriction build in it. Why would the Dynamic Authentication Config API be different in this regard?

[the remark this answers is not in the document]
enj: > You can already achieve (abuse) this via the `CertificateSigningRequest` API and there is no such restriction build in it. Why would the Dynamic Authentication Config API be different in this regard?

The latest version of the CSR API actively tries to [prevent](https://github.com/kubernetes/kubernetes/blob/d7e10f98697644367f689be79189583b9163dcab/plugin/pkg/admission/certificates/subjectrestriction/admission.go#L83-L90) this kind of abuse.

In general, I believe it was a mistake for us to not tightly control what can assert a `system:` identity.  Some things need to be reserved for Kube.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+times [#71162].  These flag based configs also require an API server restart to
+take effect.
+
+Since there is no way to configure authentication other than via flags, hosted
+environments have no consistent way to expose this functionality [#166].  To get
+around this limitation, users install authentication proxies that run as
+cluster-admin and impersonate the desired user [kube-oidc-proxy] [teleport].
+Other than being a gross abuse of the impersonation API, this opens the API
+server to escalation bugs caused by the proxy (such as improper handling of
+incoming request headers).  This proxy also intercepts network traffic which
```
tallclair, line 182: > this opens the API server to escalation bugs caused by the proxy (such as improper handling of incoming request headers)

Isn't this true with any authentication handler? I suppose the proposed `Prefix` requirement mitigates it to some degree.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+authenticators.  The `AuthenticationConfig` API will require username and groups
+prefix config to be explicitly opted out of by the cluster admin.
+
+#### Ordering of authentication modes
+
+To prevent interference with other authentication modes configured on the API
+server, all `AuthenticationConfig` objects will be considered after the other
+authenticators have had a chance to run.  This prevents issues such as a
+misbehaving authenticator attempting to authenticate a Kubernetes service
+account bearer token.  It also prevents performance issues caused by invoking
```
tallclair, line 301: Are authenticators always executed serially?

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+How will UX be reviewed and by whom?
+
+Consider including folks that also work outside the SIG or subproject.
+-->
+
+#### Restrictions on asserted identities
+
+To prevent confusion with identities that are controlled by Kubernetes, the
+`system:` prefix will be disallowed in the username and groups contained in the
+`user.Info` object.  A disallowed username will cause authentication to fail.
```
tallclair, line 285: It would be nice if the credentials for a disallowed username weren't even forwarded to the dynamic authenticator. The only way of making this work that I can think of is if a static authenticator was able to recognize the format of the credential, but it was the wrong credential, then it could shortcircuit-deny the request. WDYT?

EDIT: alternatively, would it be possible to require clients authenticating with a dynamic authenticator to provide a username hint? (e.g. as a request header, only relevant for OIDC & webhook types). This way, the authentication could be skipped if the requested username doesn't match the prefix, and the request would be auto-denied if the authenticated username didn't match.

keps/sig-auth/1688-dynamic-authentication-config/README.md
tallclair: To clarify, this is the config for the x509 authentication type (i.e. authenticating the user request), not for authenticating the apiserver request to the authenticator, right?

I believe the concerns raised here are with regards to the latter situation.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+  Issuer admissionregistrationv1.WebhookClientConfig `json:"issuer" protobuf:"bytes,1,opt,name=issuer"`
+
+  ClientID string `json:"clientID" protobuf:"bytes,2,opt,name=clientID"`
+
+  UsernameClaim string `json:"usernameClaim" protobuf:"bytes,3,opt,name=usernameClaim"`
+
+  GroupsClaim string `json:"groupsClaim" protobuf:"bytes,4,opt,name=groupsClaim"`
+}
+
+type PrefixConfig struct {
```
tallclair, line 400: Rather than a prefix, maybe use a glob pattern (please not a regex)? E.g. if I want to authenticate `*@example.com` or `*@*.example.com`

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: > To clarify, this is the config for the x509 authentication type (i.e. authenticating the user request)

Correct.  This CA bundle allows validation of client certs.

> I believe the concerns raised here are with regards to the latter situation.

IIUC, the concern was "please let me separate my static config from dynamic config."  For example, when configuring an OPA admission webhook that enforces policy across all resources, all `ValidatingWebhookConfiguration` across all clusters will basically look the same minus the CA bundle.

One could imagine a similar situation where most of the `AuthenticationConfig` object is static minus the CA bundle.  I think it is far less likely and I must prefer being able to validate the CA bundle inline.  I think there could be a case for allowing an empty CA bundle which would be ignored until filled in by some controller.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: Yes because they have a defined order they run in (i.e. SA token validation is always before a token webook).  The token cache has logic to single flight requests for the same token but this process is still serial from the perspective of a single request.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: This sounds like a different use case than `PrefixConfig` which is meant to disambiguate the same username/group from different authenticators.

I think the use case is valid because it makes it significantly easier to use something like `gitlab.com` as your `oidc` IDP (i.e. maybe you do not want every user with a GitLab account to be able to authenticate to your cluster even if they are not authorized to do anything).

The approach I have been thinking around such a use case is:

```go
// the asserted user must be a member of at least one of these groups to authenticate.
// set to ["*"] to disable this check.
// required (cannot be empty)
requiredGroups []string
```

Basically, a simple gate on group membership for the `x509` and `oidc` authenticators.  `webhook` can internally do anything to limit authentication based on other parameters (i.e. some form of authz check) so it matters a lot less in that case.

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: I believe this is the point that @mikedanese brought up in the sig-auth call around "I do not want to forward credentials that were meant for my static webhook to these dynamic webhooks when my static webhook has a transient failure and authentication fell through to the later authenticators."  This is a valid concern and we should try to address it.

> It would be nice if the credentials for a disallowed username weren't even forwarded to the dynamic authenticator.

This does assume that the API server knows in advance what the username will be.  I think we can only know this for sure with certs and SA tokens.

> The only way of making this work that I can think of is if a static authenticator was able to recognize the format of the credential, but it was the wrong credential, then it could shortcircuit-deny the request. WDYT?

The idea of adding short circuit deny logic into authentication makes me very nervous.  I think the authn stack is much easier to reason about because we do not have to worry about "which authenticator owns a credential."  I could also imagine cases where a user has a webhook today that validates tokens that look like service account JWTs.  I do not want to accidentally break them.

> would it be possible to require clients authenticating with a dynamic authenticator to provide a username hint? (e.g. as a request header, only relevant for OIDC & webhook types). This way, the authentication could be skipped if the requested username doesn't match the prefix, and the request would be auto-denied if the authenticated username didn't match.

I think there are some problems with such an approach:

1. The client has to know the username it is trying to authenticate as.  Some environments use UUIDs which make this type of flow very painful.
1. How would the user tell `kubectl` to pass a such a header?  Would we extend the `exec` credential plugin mechanism?
1. If you have more than one webhook configured and they both use opaque tokens, how do you know which webhook to send the token to?

Here is my current thoughts on how we could make this safer:

1. We require every `AuthenticationConfig` object to have a name with the format of `domain:path_segment`.  Something like `company.com:v1`.  We reserve `[*.]k8s:*` for any future Kube usage we come up with.
1. We make no other changes to `x509` or `oidc` because the validation for those authenticators is in-memory of the API server.
1. For `webhook`, we only pass a token to a webhook for validation if it has the prefix `<name_of_authentication_config>::`.  For example, if we see a token `company:v1::D2tIR6OugyAi70do2K90TRL5A` and we have a webhook configured with the name `company:v1`, we would strip the prefix (so that the webhook does not need to know its name) and pass `D2tIR6OugyAi70do2K90TRL5A` to the webhook.  Note that this guarantees that we never pass a standard OIDC token to a webhook.

The above flow requires that a client be given the name of the webhook out of band in some manner and that it concatenate the prefix to the token (this could be done by an `exec` credential provider though we could add direct support to `kubectl` to help facilitate this).

I cannot think of any approach that does not involve the client providing us extra information to make the `webhook` case not leak tokens meant for one webhook to another webhook.  Encoding it directly into the token as a prefix seems like the simplest approach that could be done today via `exec` plugins.  Having to know the (probably static, maybe even well-known like `github.com:kubernetes`) name of the webhook authenticator seems okay to me.

Thoughts?

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: The concern I was raising here is that with an impersonation proxy your network diagram looks like:

```
[ user agent ] -- user creds over TLS --> [ impersonation reverse proxy ]  -- proxy credentials over TLS with impersonation headers --> [ API server ]
```

The reverse proxy has to handle the incoming request, authenticate it, sanitize it (ex: fail if incoming request has impersonation headers), and then pass it through to the API server with its own credentials.  What happens if Kube ever adds new headers that have special meaning?  Some headers may be benign from a security perspective and the proxy should pass them through.  Others could be like impersonation and change the security parameters of the request.  Having a proxy that is `cluster-admin` act a confused deputy is something I would like to avoid.

This network is far easier to protect IMO:

```
[ user agent ] -- user creds over TLS --> [ API server ]  -- possible webhook over TLS with user creds --> [ webhook ]
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: We could also support a simple form of wildcard by allowing the client to send the token as `company.com:*/D2tIR6OugyAi70do2K90TRL5A` as a way of indicating "I am okay with sending this token to all `company.com` authenticators."

Not sure if that is a good idea, but it would be easy to implement.

tallclair: This proposal makes sense to me.

enj: > It requires tokens to be provisioned with the prefix though (unless the client adds them). Is that acceptable?

Overall, I think this is acceptable because it could be made transparent to the end-user.

Authentication webhooks and `exec` plugins work together quite nicely to give a seamless experience.  We also do not need to concern ourselves with `oidc` tokens (they just continue to work as they always have and are never sent to the dynamic webhooks).  I see a few ways the webhooks could handle the prefix requirement:

1. The webhook is aware of the prefix it needs to add (i.e. the prefix is static or the webhook knows how it was configured on a given cluster) and does so automatically; no client side changes are required
1. The webhook is unaware of the prefix (it could be an "old" webhook or the name of the webhook is not stable across clusters) but the client adds it automatically via:
   1. Automatic config by the process that creates the kubeconfig file for the end-user
   1. Probing the cluster somehow for the information (i.e. the `exec` plugin knows how to access some public datastore that tells it what prefix to use on a given cluster)
   1. Manual config by the end-user (the user is informed out of band of the prefix information)

Only the last option leaks this implementation detail to end-users.

mikedanese:
Incomplete list of concerns:

* Dynamic authentication webhook configuration builtin to kube-apiserver means that operators of traditional statically configured TokenReview webhooks need to consider that an outage may cause credentials presented to the kube-apiserver to redirect to $anywhere. This includes both credentials of end users and service accounts used by the operator for cluster management. This API means that no hosting provider can make a guarantee around the confidentiality of tokens presented to the kube-apiservers that they operate.
* The extensibility of dynamic authentication webhook if built into the kube-apiserver is very limited. Any experimentation beyond the basic x509/bearer token/oidc mechanisms will require modification of the kube-apiserver or implementation in a front proxy. We have no path to support SAML, Kerberos, token binding, request signing, SPIFFE, v4, ALTS. We really need to focus our investment on extensibility in core.
* Dynamic authentication built into the kube-apiserver substantially increases the bar for any pre-authorization DoS protection layer. The layer must now keep pace with changes to dynamic authentication configuration in the cluster, and keep pace with feature additions and API changes between versions of k8s.

If this API were used to configure an authenticating proxy, these issues would largely be avoided (in that they would be the responsibility of the proxy operator, not the cluster operator). The front proxy architecture is what we've been recommending to users for the past N years. If there are challenges with developing and operating an authentication proxy, we should identify and fix them.

enj:
> Incomplete list of concerns

Can I get the complete list 😝

On a more serious note, please take the time to enumerate all concerns that you have.  I cannot address them if I do not know what they are. 

> an outage may cause credentials presented to the kube-apiserver to redirect to $anywhere.

This is handled via the proposed change in https://github.com/kubernetes/enhancements/pull/1689#discussion_r419738032.  PTAL.
 
> The extensibility of dynamic authentication webhook if built into the kube-apiserver is very limited

I disagree.  Authentication's extra field (set by a webhook) combined with webhook authorization and webhook admission give you a lot of control over the auth stack if you can set the CLI flags on the API server.  All this KEP does it make it easier to use some of that functionality generically without CLI flag access.

> Any experimentation beyond the basic x509/bearer token/oidc mechanisms will require modification of the kube-apiserver or implementation in a front proxy.

This is simply not true.  The only requirement is that the final credential used for authentication be x509 / oidc / token.  How you get that credential is completely open ended and well supported via `exec` plugins.  For example, an `exec` plugin could perform a Kerberos challenge flow with some component and upon success be issued a short lived token for use against the Kube API.  I think this is a very reasonable and desired approach.  This is similar to how OpenShift supports a [broad range of IDPs](https://github.com/openshift/api/blob/944c57cb1477eaee4a94e3f82c8aab639fd0b94c/config/v1/types_oauth.go#L146-L173) - no matter what IDPs you configure, the credential presented to the API server is always a bearer token.  What this design avoids is forcing that "Kerberos component" to be a proxy for all requests.

> We have no path to support

I would argue that these are orthogonal concerns to this KEP.  If we did decide to support more authentication methods, we would have to configure them somehow.  I am willing to bet that the structure of this API is far better suited to handle new authentication methods than a CLI flags based approach.

> SAML, Kerberos

OpenShift, PKS, Dex, etc support these types of IDPs by giving the user a token upon successful authentication.

> token binding, request signing, SPIFFE, v4, ALTS.

This KEP does not hinder any efforts in that space.

> We really need to focus our investment on extensibility in core.

I would argue that this KEP does exactly that.  If you could `kubectl apply` a complete auth stack, people would be far more willing to invest in this space, IMO.  I am also personally signing up for all implementation, testing and maintenance of this feature for as long as I work on Kube, which is the best any single actor can do.

> increases the bar for any pre-authorization DoS protection layer.

I do not see how this is any different from a front proxy approach.  DoS via the proxy is just as serious as DoS directly against the API.  Saying that this kind of DoS is "not your problem because its not your proxy" is hardly helpful.  I would argue this is the realm of the on-going priority and fairness efforts.

> The layer must now keep pace with changes to dynamic authentication configuration in the cluster, and keep pace with feature additions and API changes between versions of k8s.

I do not see this as a real problem.  This KEP covers every "good" authn API we have today that has been developed over 5+ years.  No one is actively trying to add net new authentication protocols to Kube.  It is simply easier to issue short lived certs / tokens that are obtained using a user's desired mode of authentication.  In terms of API velocity, auth in Kube is incredibly slow moving.

> If this API were used to configure an authenticating proxy, these issues would largely be avoided (in that they would be the responsibility of the proxy operator, not the cluster operator).

This is not a solution.  All that does is push the responsibility elsewhere and it introduces a whole set of new problems.

> The front proxy architecture is what we've been recommending to users for the past N years.

And this was a mistake.  It was a convenient response that allowed us to do nothing and push the burden onto end-users.  This is a real problem that end-users face.  The response to this proposed API change from the community has been positive, even from managed Kubernetes offerings.  Dynamic control over authn is the third most requested feature in EKS per @micahhausler.

All of authentication, authorization, admission, etc could be implemented with a proxy.  But no one would be satisfied with that approach.  Why do we think that is okay for authentication?  Would projects like OPA be nearly as successful in Kube if admission plugins could only be configured via the CLI?  Note that I mentioned "an ecosystem" on the last SIG Auth call.  I was not referring to an ecosystem of auth proxies.  I was referring to an ecosystem of apps built on top of an opinionated Kuberentes native auth stack.  If auth stacks were generically portable across Kube in the way admission plugins are, it would actually make sense to develop on top of them.  Auth is just another extension point.

> If there are challenges with developing and operating an authentication proxy, we should identify and fix them.

I have listed quite a few in https://github.com/kubernetes/enhancements/pull/1689#issuecomment-617402067.  They cannot be fixed because they are a by product of a proxy based architecture.  It is also simply not possible to protect CRD based API with the same level of control as a built-in API.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+possible to collect this information before implementation begins in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself.  KEP editors, SIG Docs, and SIG PM
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+Authentication is Kubernetes is quite flexible and unopinionated.  There is no
```
ncdc, line 150:
```suggestion
Authentication in Kubernetes is quite flexible and unopinionated.  There is no
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
ncdc: @enj to resolve this question, do you want to add a new non-goal of deprecating or removing the existing flags?

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation.  The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+This change aims to add a new API Kubernetes REST API called
```
ncdc, line 232:
```suggestion
This change aims to add a new Kubernetes REST API called
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+`AuthenticationConfig`.  It is similar to the `ValidatingWebhookConfiguration`
```
ncdc, line 233: Bike-shedding: AuthenticationConfiguration for consistency?

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,620 @@
...
+
+`AuthenticationConfig` objects of the same type will be ordered lexicographically.
+No explicit control over ordering will be provided to the cluster admin.  All
+`AuthenticationConfig` objects are equally trusted to assert arbitrary identities
+and there is no concept of priority among them.  Token webhook implementations
+are responsible for understanding what tokens are meant for them.
+
+#### Source of identity
+
+One important characteristic of the an impersonation based proxy is that it is
```
ncdc, line 319:
```suggestion
One important characteristic of an impersonation based proxy is that it is
```

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: ¯\_(ツ)_/¯

enj:
@mikedanese @tallclair the KEP is up to date with all comments now.

dims:
cc @liggitt

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,807 @@
...
+webhook for verification.
+
+This guarantees that a JWT like the ones used for service account tokens and
+with `oidc` are never sent to a `webhook`, increasing both performance and
+preventing any leakage of these credentials.  Furthermore, only one `webhook`
+will be invoked for a given token, thus cross webhook credential leakage cannot
+occur.  Performance will also be increased by limiting the number of webhooks
+that can be invoked on a given request.
+
+It is the responsibility of the token issuer, the webhook, and the client to
```
deads2k, line 445: From a purely practical point of view, how would the token issuer even know the name of the AuthenticationConfig driving it?  This model seems like it might work if you deployed a per-cluster authenticator and then wired the configuration across, but that seems antithetical to previous notion of authenticator webhooks as things that can exist outside of cluster.

deads2k:
I'm generally opposed to this KEP.

1. Authentication is already extensible via command line flags without recompiling a kube-apiserver binary.  This gives cluster providers (the actor setting the flags) the ability to extend the authentication external to the kube-apiserver binary.
2. This configuration is rarely changed, so there is no need to introduce additional risk by refactoring easy to reason about static initialization to dynamically reloaded configuration.  This also means it doesn’t warrant introducing risk by allowing remote configuration of request interception.
3. This configuration is highly privileged, which is another reason to avoid introducing risk by allowing remote configuration of request interception.
4. To directly reply to "Dynamic control over authn is the third most requested feature in EKS” - there is no reason that EKS cannot expose an API to do exactly that.  My understanding is that EKS (all other major hosted services) already curate the list of configuration options that they want to support.

seh:
> there is no reason that EKS cannot expose an API to do exactly that.  My understanding is that EKS (all other major hosted services) already curate the list of configuration options that they want to support.

They _can_, but so far they won't. I know this from having tried for a long while.

What I expected you to say is that one could possibly build out this feature on top of the existing configuration approach, by implementing a Webhook server that in turn delegates to any number of other configured Webhook servers. Is that what @mikedanese was suggesting with the front proxy (https://github.com/kubernetes/enhancements/pull/1689#issuecomment-616830634)?

keps/sig-auth/1688-dynamic-authentication-config/README.md
enj: Even if a webhook exists outside of the cluster, it still needs to know what cluster is calling it.  Otherwise you risk getting into scenarios where the token meant for one cluster can be replayed against another (and if the webhook and cluster have a 1:1 mapping they could just be hard coded with the name).

A simple approach would be to encode the name and cluster info into the request path.  An `exec` plugin could also perform this by discovering what name it needs to use based on some public pre-auth metadata hosted on the cluster itself.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,807 @@
...
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Configuring authentication via command line flags has some limitations.  Some
+forms of config are simply easier to specify and understand in a structured
+manner such as via a Kubernetes API or file.  The existing authentication flags
+for OIDC and token webhook limit the user to only one of each of these types of
+authentication modes as there is no way to specify the set of flags multiple
+times [#71162].  These flag based configs also require an API server restart to
+take effect.
```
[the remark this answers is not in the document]
timothysc: @liggitt could the config file be loaded via a configmap + restart/sigup to yield similar dynamic behavior.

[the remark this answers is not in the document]
liggitt: the configmap bit would depend on whether the API server process was running in a pod that could have configmaps mounted to it (static pods can't), but the reloadable bit seems plausible... that's the alternative I described in https://github.com/kubernetes/enhancements/pull/1689#discussion_r427028922

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,807 @@
...
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+TBD.
+
+## Alternatives
```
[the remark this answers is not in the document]
enj: Responding to Jordan's overall comment about "make OIDC better."

Overall, I do not understand how a REST API (I discuss two below) ends up being different than the impersonation API or the CSR API from a security perspective.  If we give the cluster operator the same level of control, why does it matter?

As a thought experiment, if I was to propose a new `UserTokenRequestAPI` that worked like the CSR API's `kubernetes.io/kube-apiserver-client` signer, would that be okay?  This would be a way for some privileged component to ask for a _token_ for a particular user info that the Kube API would honor.  The token itself would be opaque to the client.  What ever component issued the token could choose not to based on whatever requirements the cluster-operator specifies.  This handles the "the identity is rooted in a trust owned the API server / cluster operator."  Such an API adds one layer of indirection but that could be hidden from the end user.

---

> My recommendation would be to narrow this to improving the OIDC provider config, and for clusters that wanted to allow control of that config via a REST API, to use a CRD-based API paired with a config file writer.
 
>     1. allow the kube-apiserver to configure multiple oidc providers via a structured, reloadable config file
> 
>     2. Separately, create a CRD for custom resources that map to the items in the oidc config in a straightforward way
> 
>     3. Providers that want to expose the config via a REST API can install the CRD and the custom-resource -> OIDC config file process
> 
>     4. Providers that do not want to expose the config still benefit from multiple oidc provider capability and have no additional exposure and no new API surface to lock down

So we would need to in `k/k`:

1. Make a new file based API
1. Wire that file based config through
1. Reload that config

Outside of `k/k`:

1. Create a new CRD
1. Create new controllers to handle the CRD
1. Create new admission plugins to restrict access to the CRD, but with far weaker security (they can be removed if you can delete the webhook)

For every single provider:

1. Ask them to run all of those components co-located with their API servers
1. Configure credentials and the correct authz for these components
1. Set the right flags on the API server

This sounds an incredible amount of work in-tree, out-of-tree, and for every single provider.  I understand the desire to build things out of core, but I do not think this moves us forward in any meaningful way.

Why would we instead not:

In `k/k`:

1. Create a new dynamic OIDC config API, in a new group, disabled by default
1. Wire the new API through
1. Reload the API config

Outside of `k/k`:

No work

For every single provider:

1. Ask them to pass a single flag to enable the API if they wish to support it

This is the same amount of work and changes in `k/k`.  There is no work outside of `k/k` and the burden on the providers is as small as we can make it.  Since its off by default, no security concerns are added.  The provider would opt-in in the same way they would for the CRD based flow.  They just would not have to do a large amount of provider specific work.

---

> Is there evidence this is a thing most hosted Kubernetes offerings want and plan to enable? If this would not be enabled by default, or made part of conformance, or opted into broadly, I question whether it should exist as a built-in API at all.

From what I can tell, EKS and AKS would.  GKE and OpenShift would not (though OpenShift already lets you control IDPs via a Kube REST API).  Not sure about others.  I suspect Linode, Digital Ocean, Rancher might.  It would be enabled in VMware's products (duh).

If every hosted offering had a Kube API that let you configure auth in some way that they were comfortable with, then we would be fine as-is.  The reality is that even when providers want to allow this, it is a lot of work to build all the tooling to wire this stuff through.  They also tend to use non Kube APIs, which makes it hard to integrate with.

lavalamp:
We make extensibility features dynamic when it is expected that they will change over the life of a cluster.

I personally can't see ever wanting to change an authentication system once the cluster was doing something. (Also, an unwanted authn system change sounds like the one of the most thorough pwnings I can imagine.)

So, I think this is a candidate for static config file configuration, not dynamic API based configuration.

Access to the static config files is a higher privilege than "cluster admin". I think requiring the highest privilege to change this is good.

If the problem is just getting this right once, at cluster initialization, as I suspect, I'd look at making a config file + support in Cluster API.

If the problem is not that, and people want to change this dynamically, during the life of their cluster, more than once... I guess I'd want to see a long list of people who want that, it's really hard for me to believe.

keps/sig-auth/1688-dynamic-authentication-config/README.md
```diff
@@ -0,0 +1,807 @@
...
+know that this has succeeded?
+-->
+
+1. Create a new Kubernetes REST API that allows configuration of authentication
+  - x509
+  - OIDC
+  - Token Webhook
+2. Changes made via the REST API should be active in the order of minutes without
+requiring a restart of the API server
+3. Allow the use of a custom authentication stack in hosted Kubernetes offerings
```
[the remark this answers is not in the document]
dcberg: I’m not in agreement with this proposal and don’t agree that such a capability should be exposed as a built-in API within the community Kubernetes project. We would not enable this feature and we would desire that, if accepted, the feature is opt-in.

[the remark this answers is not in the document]
jbeda: Many many customers I talk to are looking to use k8s clusters from multiple providers and have consistent auth among them. This is something that comes up again and again with enterprise customers.

I think we should be looking at this from the point of view of the end users of the clusters (both cluster admins and app teams) and not hosting providers.  After all, they are the real drivers of the success of Kubernetes.  

@dcberg -- if this is something that users want and you don't enable it then you'll be at a competitive disadvantage.  That is how this should work.  Create the capability and then let the real users vote with their feet.

jbeda:
@lavalamp I think you are conflating things.  There is the need to have end users configure it in a standard way regardless of hosting provider that is different for when and how often it changes.

This is something that end users have hit their head against.  

Allowing the apis/mechanisms for doing this in a provider-neutral way is a first step.  It gives end users something to tool against and allows them to work with their providers to enable the capability.

Cluster API is great but it doesn't get to that need to work against managed clusters.  In fact, cluster API is a great option for those customers (and there are many) that opt to run their own clusters so that they *can* get access to configure auth.

keps/sig-auth/1688-dynamic-authentication-config/README.md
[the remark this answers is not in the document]
seh: Yes, Joe, thank you for putting it so well. In several organizations I've seen so far, the cloud provider's authentication system integrated into their Kubernetes offerings are of no use to us. They are merely an impediment, used only begrudgingly and thereafter ignored.

The cost is that these clusters wind up using very few principals with coarsely assigned permissions ("Fine, then every developer is a cluster administrator!"), for lack of finer-grained control that could be employed consistently across all the clusters hosted with different providers.

Audit logging is no longer accurate. RBAC is less useful. In all, the system becomes less secure and less trustworthy for important workload.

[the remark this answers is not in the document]
dcberg: @jbeda I don’t disagree that there is a need to have common auth across clusters. While such an approach would not break a hosted solution, it may limit the use of other integrations with the provider.
My ask is that the feature has an “opt-in” mechanism to allow providers flexibility to expose the feature in such as a way to limit impact and/or to point users to an alternative approach.

lavalamp:
@jbeda 

> There is the need to have end users configure it in a standard way regardless of hosting provider that is different for when and how often it changes. This is something that end users have hit their head against.

I've asked the people in my org who would know these things if they agree with this assertion or not. If you can provide clearer evidence for this assertion that would be helpful.

Note that static config is also a "standard way". The debate is actually about whether the config must be end-user accessible or not.

> Allowing the apis/mechanisms for doing this in a provider-neutral way is a first step.

It is pointless if the providers wouldn't enable it. If this is to be an end user extensibility feature, it needs to universal. And therefore non-optional, required by conformance. Otherwise you won't actually get more portability, you get less. So, I wouldn't want to start down this road unless there's very clear demand from users. Right now it's not clear that there is (maybe it's clear to you but that doesn't really help the rest of us).

seh:
> I wouldn't want to start down this road unless there's very clear demand from users. Right now it's not clear that there is (maybe it's clear to you but that doesn't really help the rest of us).

The audience here is the administrator or operator deploying Kubernetes clusters. It may be a small group with a big impact. In a given organization, there may be one or two such people setting up dozens of clusters to be used by hundreds of people to run applications serving millions of people. Does that count as one user (the organization), or two (the administrators), or hundreds (the developers)?

When I've talked to others interested in Kubernetes about authentication, I've heard several times, "Oh, I didn't think you were allowed to configure things like that." In other words, increasingly, this configuration being inaccessible turns people off from thinking about it. They just assume that "Kubernetes does not have a way to add users."

These are all anecdotes. What counts as proof? What counts as enough?

sftim:
Mostly an aside, could be something to bear in mind.

Can I create an AuthenticationConfig that lets me implement authorisation with the client identity masked / erased?

For example, using SAML it's relatively easy for an IdP to make an assertion about the bearer: they're a staff member, they have a role in Project 23 and another role in project 42, and they have the following 7 entitlements [entitlements go here].

Current ways of working with Kubernetes seem to me to infer that all authorization decisions are based around a known user identity, whereas that's not always the case. Sometimes you can, eg, take the ARN of an AWS IAM role, and put that in place of the username, or have a mapping.

It's good to retain the ability to authorized something even if you don't have a unique identifier for the caller, and never will.

liggitt:
My comment at https://github.com/kubernetes/enhancements/pull/1689#pullrequestreview-414073983 still reflects my current thinking, and I confirmed with David https://github.com/kubernetes/enhancements/pull/1689#issuecomment-630416457 still reflects his. 

Of the authentication mechanisms this KEP discusses, supporting multiple OIDC providers seems the most reasonable goal. However, I would characterize what this KEP proposes as cluster configuration API, not something that should be built into the kube-apiserver.

I considered the following questions:

1. Are there gaps in Kubernetes components that prevent this from being built on top? In this case, there are, and I would welcome proposals addressing these gaps:
    * expand the kube-apiserver OIDC support to allow multiple OIDC providers (useful for a variety of reasons)
    * add support for updating the kube-apiserver OIDC config file without restarting the server (we already do similar things for files containing things like CA bundles and TLS certificates)

2. With those gaps addressed, could this proposal to expose OIDC config as a REST API be built on top of the kube-apiserver outside kubernetes/kubernetes?
    * An `OIDCProvider` (strawman) CRD could be defined that maps easily to items in the kube-apiserver OIDC config file
    * `OIDCProvider` instances could be created via the REST API and transformed to the kube-apiserver OIDC config file

Some benefits of that approach:
  * the OIDCProvider API can be iterated on quickly
  * cluster operators that want to expose an API are not limited to exposing it via the kube-apiserver being configured (they could expose it in an underlay/config API)
  * cluster operators that don't want to expose an API can still benefit from multi-oidc-provider support

Since we are in agreement that this proposal is not acceptable, we will close this PR and look forward to reviewing proposals to fill in the capability gaps identified in the kube-apiserver.

@liggitt, @deads2k, and @mikedanese

PR #1691, opened 2020-04-16
CSR v1 details, Approved/Denied/Failed conditions

liggitt:
/sig auth
/assign @mikedanese @deads2k 
/cc @munnerz

### DIFF
--- a/keps/sig-auth/20190607-certificates-api.md
+++ b/keps/sig-auth/20190607-certificates-api.md
@@ -34,6 +34,8 @@ status: implementable
   - [Automatic CSR Approval Implementations](#automatic-csr-approval-implementations)
   - [Automatic Signer Implementations](#automatic-signer-implementations)
   - [Test Plan](#test-plan)
+- [Graduation Criteria](#graduation-criteria)
+  - [Beta to GA Graduation](#beta-to-ga-graduation)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
 
@@ -111,10 +113,8 @@ certificate field of the status can only be updated via the `/status`
 subresource allowing signing permission to be authorized independently of other
 operations on the CertificateSigningRequest.
 
-Terminal failures can be indicated with a `Failed` condition added by a controller,
-such as a signer indicating a refusal to sign the CSR.
-A `Failed` condition cannot be removed, cannot have a `False` or `Unknown` status,
-and is treated as `True` if present.
+Terminal signer failures (such as a refusal to sign the CSR) can be indicated with a `Failed` condition.
+A `Failed` condition cannot be removed, cannot have a `False` or `Unknown` status, and is treated as `True` if present.
 
 The API is designed to support the standard asynchronous controller model of
 Kubernetes where the approver and signer act as independent controllers of the
@@ -286,6 +286,9 @@ Cluster admins can either:
 `/status` subresource:
 
 * allows modifying `status.conditions` other than `Approved`/`Denied`
+  (`Approved`/`Denied` conditions are still limited to the `/approval` subresource
+  to ensure version-independent authorization policy grants consistent permissions
+  across v1beta1 and v1 versions)
 
 ### CertificateSigningRequest API Definition
 
@@ -380,14 +383,26 @@ type CertificateSigningRequestCondition struct {
   // conditions, including ones that indicate request approval ("Approved", "Denied") and failure ("Failed").
   // Approved/Denied conditions are mutually exclusive.
   // Approved, Denied, and Failed conditions cannot be removed once added.
+  // Only one condition of a given type is allowed.
   Type RequestConditionType
   // status of the condition. Valid values are True, False, and Unknown.
   // Approved, Denied, and Failed conditions are limited to a value of True.
+  // In v1beta1, this is optional and defaults to True.
+  // In v1, this is required and not defaulted.
   Status v1.ConditionStatus
   // brief reason for the request state
   Reason string
   // human readable message with details about the request state
   Message string
+  // lastUpdateTime timestamp for the last update to this condition.
+  // If unset, when a CSR is written the server defaults this to the current time.
+  // +optional
+  LastUpdateTime metav1.Time
+  // lastTransitionTime is the time the condition last transitioned from one status to another.
+  // If unset, when a new condition type is added or an existing condition's status is changed,
+  // the server defaults this to the current time.
+  // +optional
+  LastTransitionTime metav1.Time
 }
 
 type RequestConditionType string
@@ -448,6 +463,16 @@ control plane.
   - Admission plugin limiting system:master client CSRs
   - Auto-approving controller
   - Auto-signing controller
+  - v1beta1 behavior validation
+    - `Approved`/`Denied` conditions can be added via `/approval` subresource, modifications are ignored via `/status` subresource
+    - `Failed` condition can be added via `/approval` or `/status` subresource
+    - `status.certificate` can be set via `/status` subresource, modification is ignored via `/approval` subresource
+    - `status.certificate` can be set to arbitrary content
+  - v1 behavior validation
+    - `Approved`/`Denied` conditions can be added via `/approval` subresource, modifications are rejected via `/status` subresource
+    - `Failed` condition can be added via `/approval` or `/status` subresource
+    - `status.certificate` can be set via `/status` subresource, modification is rejected via `/approval` subresource
+    - `status.certificate` content is validated when modified
 
 - integration tests of:
   - auto-approve and signing of kubelet client CSR
@@ -460,10 +485,18 @@ control plane.
   - status subresource get/update behavior
   - creating, approving, and signing a CSR for a custom signerName (e.g. `k8s.example.com/e2e`)
 
+## Graduation Criteria	
+
+### Beta to GA Graduation
+
+- Multi-signer API, admission plugins, and controller-handling for `kubernetes.io/*` signers is implemented and tested
+- Test plan is completed
+- e2e tests acceptable as conformance tests exist for all v1 endpoints+verbs
+
 ## Implementation History
 
 - 1.4: The Certificates API was merged as Alpha
 - 1.6: The Certificates API was promoted to Beta
 - 1.18: 2020-01-15: Multi-signer design added
 - 1.18: 2020-01-21: status.certificate field format validation added
-- 1.19: 2020-04-15: v1 details added
+- 1.19: 2020-05-07: v1 details added
--- a/keps/sig-auth/20190607-certificates-api.md
+++ b/keps/sig-auth/20190607-certificates-api.md
@@ -28,13 +28,12 @@ status: implementable
   - [Sequence of an Issuance](#sequence-of-an-issuance)
   - [Signers](#signers)
     - [Limiting approval and signer powers for certain signers.](#limiting-approval-and-signer-powers-for-certain-signers)
+  - [API changes between v1beta1 and v1](#api-changes-between-v1beta1-and-v1)
   - [CertificateSigningRequest API Definition](#certificatesigningrequest-api-definition)
   - [Manual CSR Approval With Kubectl](#manual-csr-approval-with-kubectl)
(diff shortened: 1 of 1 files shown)


keps/sig-auth/20190607-certificates-api.md
```diff
@@ -101,15 +100,22 @@ committing it to storage so that they can be used later during CSR approval. The
 information contained in the spec is immutable after the request is created.
 
 An approver updates approval status of the CertificateSigningRequest via the
-CertificateSigningRequestStatus. The approval condition can only be updated via
-the `/approval` subresource allowing approval permission to be authorized
+CertificateSigningRequestStatus. `Approved`/`Denied` conditions can only be set via
+the `/approval` subresource, allowing approval permission to be authorized
```
deads2k, line 104: Are we happy with this?  I think I may prefer setting status in a v1.

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -101,15 +100,22 @@ committing it to storage so that they can be used later during CSR approval. The
...
+cannot have a `False` or `Unknown` status, and are treated as `True` if present.
 
-Contingent on approval, a signer posts a signed certificate to the status. The
+Once an `Approved` condition is present, a signer posts a signed certificate to the status. The
 certificate field of the status can only be updated via the `/status`
 subresource allowing signing permission to be authorized independently of other
 operations on the CertificateSigningRequest.
 
+Terminal failures can be indicated with a `Failed` condition added by a controller,
+such as a signer indicating a refusal to sign the CSR.
```
deads2k, line 115: why would this not be denied?

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -101,15 +100,22 @@ committing it to storage so that they can be used later during CSR approval. The
...
+`Approved`/`Denied` conditions are mutually exclusive, cannot be removed,
+cannot have a `False` or `Unknown` status, and are treated as `True` if present.
 
-Contingent on approval, a signer posts a signed certificate to the status. The
+Once an `Approved` condition is present, a signer posts a signed certificate to the status. The
 certificate field of the status can only be updated via the `/status`
 subresource allowing signing permission to be authorized independently of other
 operations on the CertificateSigningRequest.
 
+Terminal failures can be indicated with a `Failed` condition added by a controller,
```
deads2k, line 114: I'm not really clear on what terminal failure would not result in a Denied condition.

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -122,19 +128,39 @@ This is typical of many PKI architectures.
...
+1. The requestor observes the `Denied` condition.
+
+Signer failure sequence:
+
+1. The requestor generates a private key, builds a certificate signing request,
+   and submits the `CertificateSigningRequest` to the Kubernetes certificates
+   API.
+1. The approver controller observes the newly submitted request, validates and
+   authorizes the request and adds an `Approved` condition via the `/approval` subresource.
+1. The signer observes the approval, but encounters a terminal failure during the signing process,
```
deads2k, line 160: it's not obvious to me why this isn't denied==true.  Is this pre-existing?

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -234,7 +261,31 @@ Cluster admins can either:
...
+  to allow for explanatory text as described in section 5.2 of RFC7468.
+
+`status.conditions` validation:
+
+* type is required
+* status is required and must be `True`, `False`, or `Unknown`
+
+`/status` subresource:
+
+* allows modifying `status.conditions` other than `Approved`/`Denied`
```
deads2k, line 288: why wouldn't we use the general status subresource for setting approved and denied conditions too?  Someone already had to opt-in for v1 and they already have to understand /status to set the signed certificate.  it doesn't seem like too much to ask. In fact, it makes it atomic.

liggitt: authz policy isn't version-aware. `certificatesigningrequests/status` would mean different things for v1 and v1beta1.

keps/sig-auth/20190607-certificates-api.md
liggitt: A signing failure is not the same as a denial. We don't currently have a way for a signer to indicate permanent failure signing an approved CSR.

keps/sig-auth/20190607-certificates-api.md
mikedanese: Agree that the subpresource here is obsolete now that we have the admission control. We should definitely consider simplifying the strategy here for v1. I recommend that we revisit this separately once we are sketching the GA surface. Maybe record it here: https://github.com/kubernetes/kubernetes/issues/69836

keps/sig-auth/20190607-certificates-api.md
mikedanese: We need to distinguish between signing failure do to sidning policy violation and approver denial, right?

keps/sig-auth/20190607-certificates-api.md
deads2k: > A signing failure is not the same as a denial. We don't currently have a way for a signer to indicate permanent failure signing an approved CSR.

Yeah, got my actors messed up

keps/sig-auth/20190607-certificates-api.md
deads2k: > authz policy isn't version-aware. `certificatesigningrequests/status` would mean different things for v1 and v1beta1.

So it would.  Explain why it isn't an option so the future reader like me knows why it looks weird, but needs to stay?

keps/sig-auth/20190607-certificates-api.md
deads2k: > We need to distinguish between signing failure do to sidning policy violation and approver denial, right?

yeah.  I'm there now.

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -325,8 +377,13 @@ type CertificateSigningRequestStatus struct {
 }
 
 type CertificateSigningRequestCondition struct {
```
deads2k, line 379: I'd like to see a LastTransitionTime in this type.

keps/sig-auth/20190607-certificates-api.md
liggitt: I considered that, but thought it was dangerous for subresource-level policies to [mean different things against different versions of the API](https://github.com/kubernetes/enhancements/pull/1691/files#r411658249).

keps/sig-auth/20190607-certificates-api.md
liggitt: sure

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -234,7 +261,31 @@ Cluster admins can either:
...
- 
+
+### API changes between v1beta1 and v1
+
+`spec.signerName` validation:
+
+* required, not defaulted
+* `kubernetes.io/legacy-unknown` is disallowed when creating new CSR objects
+
+`status.certificate` validation:
```
mikedanese, line 272: Should we validate that no certificates are self signed (i.e. rootless chain)? You mentioned that you want to avoid using this as a trust distribution mechanism.

liggitt: I wasn't planning to add tight signer chain validation at the API level, since multiple certificate blocks in this field can be used for purposes other than intermediates by other signerNames.

liggitt: I could see enforcing that in kube-controller-manager when determining what intermediate chains to accept/append

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -335,18 +392,19 @@ type CertificateSigningRequestCondition struct {
 
 type RequestConditionType string
 
-// These are the possible conditions for a certificate request.
+// These are well-known conditions types for a certificate request.
```
enj, line 395: I assume this is it to allow arbitrary conditions for use by custom signers?

keps/sig-auth/20190607-certificates-api.md
enj: Presumably this would be immutable once set for the `Approved`, `Denied` and `Failed` conditions.

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -325,8 +377,13 @@ type CertificateSigningRequestStatus struct {
 }
 
 type CertificateSigningRequestCondition struct {
-  // request approval state, currently Approved or Denied.
+  // conditions, including ones that indicate request approval ("Approved", "Denied") and failure ("Failed").
+  // Approved/Denied conditions are mutually exclusive.
+  // Approved, Denied, and Failed conditions cannot be removed once added.
   Type RequestConditionType
```
enj, line 383: All conditions must have a unique `Type` right?  i.e. no double `Approve`, `Denied`, etc.

keps/sig-auth/20190607-certificates-api.md
munnerz, line 115: > such as a signer indicating a refusal to sign the CSR.

Should we provide some clear guidance as to which actor we expect to 'own' this condition/set it? `such as` implies it *may* be some other actor, and this presumably isn't true unless signer controllers are built to check for this condition's status *before* deciding whether Deny it once again itself

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -325,8 +377,13 @@ type CertificateSigningRequestStatus struct {
...
+  // status of the condition. Valid values are True, False, and Unknown.
+  // Approved, Denied, and Failed conditions are limited to a value of True.
+  Status v1.ConditionStatus
```
munnerz, line 386: This is a new field so should have a default value and be optional right? From our previous discussions, I think this also needs to default to *True* in order to maintain backward compatibility with clients who already use their own condition types.

liggitt: > This is a new field so should have a default value and be optional right?

In v1beta1, yes. In v1, no.

> From our previous discussions, I think this also needs to default to True in order to maintain backward compatibility with clients who already use their own condition types.

In v1beta1, yes.

keps/sig-auth/20190607-certificates-api.md
liggitt: clarified this is for signers

keps/sig-auth/20190607-certificates-api.md
liggitt: yes, clarified

keps/sig-auth/20190607-certificates-api.md
liggitt: > Presumably this would be immutable once set for the Approved, Denied and Failed conditions.

I'm not sure I'd add that level of server-side validation

keps/sig-auth/20190607-certificates-api.md
```diff
@@ -402,18 +460,10 @@ control plane.
...
 
-## Graduation Criteria
-
-### Beta to GA Graduation
-
-Things to resolve for v1.
-1. .spec.signerName should be non-defaulted and required
-2. Should we disallow the legacy .spec.signerName in v1?
-3. Define how signers indicate terminal failure in signing on a CSR. Fix status conditions perhaps?
-
```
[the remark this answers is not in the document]
liggitt: good point, restored the section with normal "implement and test all the things" text

keps/sig-auth/20190607-certificates-api.md
liggitt: right

keps/sig-auth/20190607-certificates-api.md
liggitt: I can add LastTransitionTime and backfill on change server-side.

liggitt:
@deads2k @mikedanese @enj @munnerz - addressed comments in standalone commit, PTAL

PR #1695, opened 2020-04-16
Update NodeLocalDNS KEP status to implemented.

prameshj:


keps/sig-network/20190424-NodeLocalDNS-beta-proposal.md
```diff
@@ -12,7 +12,7 @@ approvers:
   - "@thockin"
 creation-date: 2019-04-24
 last-updated: 2019-04-25
```
palnabarun, line 14: A small nit: Can you also update the `last-updated` field?

PR #1706, opened 2020-04-21
Add support for symlinks and KEP metadata files

enj:
@kubernetes/enhancements-admins

### DIFF
--- a/cmd/kepval/main_test.go
+++ b/cmd/kepval/main_test.go
@@ -24,7 +24,8 @@ import (
 )
 
 const (
-	kepsDir = "keps"
+	kepsDir     = "keps"
+	kepMetadata = "kep.yaml"
 )
 
 // This is the actual validation check of all keps in this repo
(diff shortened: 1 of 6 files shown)


cmd/kepval/main_test.go
```diff
@@ -40,7 +41,19 @@ func TestValidation(t *testing.T) {
...
+
+			dir := filepath.Dir(path)
+			// true if the file is a symlink
+			if info.Mode()&os.ModeSymlink != 0 {
+				// assume symlink from old KEP location to new
+				newLocation, err := os.Readlink(path)
+				if err != nil {
+					return err
+				}
+				files = append(files, filepath.Join(dir, filepath.Dir(newLocation), kepMetadata))
```
johnbelamaric, line 53: Won't this result in running this multiple times on any `kep.yaml` that has a symlink for it's associated `README.md`? If we are only checking the metadata, I would think instead we would want to:

1) ignore the symlink
2) legacy KEP format, we need to check the file
3) new KEP format, we need to verify a `kep.yaml` exists, and validate that file

enj: I did this on purpose to make sure that when you create a symlink, you actually split the old file up correctly for the new format.  Processing the `kep.yaml` files multiple times seems okay to me, the whole process takes a few seconds at most.

Presumably at some point we can remove the old symlinks and drop all of this code.

johnbelamaric: Yeah, when there is no error it's fine, when there is an error it's slightly annoying but that's not really a big deal.

We're going to be tweaking all of this code as we evolve the KEP management process, so I am OK with this for now.

justaugustus:
Thanks for this, @enj!
/lgtm
/approve

PR #1712, opened 2020-04-24
Manifest based webhook configuration

vivekbagade:


vivekbagade:
cc @wojtek-t @JeremyOT

vivekbagade:
Bumping this up as a gentle reminder :) @deads2k @liggitt

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+- name: MutatingAdmissionWebhook
+  configuration:
+    ....
+    staticConfigFile: "<path-to-static-config-file>"
+```
+
+### Reconfiguring static webhook configuration
+
+The static configuration file is watched so that the static webhook
+configuration can be changed by editing the contents of the file.
```
deads2k, line 134: The idea of a dynamic, static webhook configuration seems somewhat unusual.  The forces that drew you to this solution (per the motivation) were related to avoiding a gap in enforcement and preventing users from mutating the values.  This prevents mutation, but doesn't solve the gap in enforcement.  If you need to change the file (hopefully infrequent), why wouldn't you simply restart the process at the same time.  The degree of access seems comparable.

deads2k:
I don't object in principle, but I haven't read it in detail.  I don't know that I'm prepared to sign up to guide it through reviews this release though.

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
vivekbagade: Watching the file for changes allows for updating other parts of the webhook configuration like eg. the matching criteria for performance reasons without causing unavailability of the API server. This is the motivation of allowing this.

liggitt: It's easy to imagine multiple categories of policy webhooks which would be placed in this file:

1. ones which are always present with a fixed config (would be present before the server is started, are guaranteed effective)
2. ones which have a variable config (typically related to intercepting as few resources as possible for performance reasons), but are not intended to be modifiable via the API

Handling updates to this file seems similar to reacting to CA bundle updates without requiring a server restart.

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Drawbacks
+
+- Reduced visibility for users wanting to list all active admission webhooks.
+  This KEP has similar visibility characteristics as compiled in admission
+  controllers.
+- Webhooks hosted in-cluster can still be disabled by cluster users with enough
+  privilege.
```
liggitt, line 142: this doesn't seem to be a drawback with this proposal, just a note that would inform decisions about where/how to host the backing webhook server

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+In short, this proposal is about augmenting `AdmissionConfig`'s plugin
+`configuration` [specification](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers)
+to include a path to a configuration file containing one of the following:
+- `admissionregistration.k8s.io/v1.ValidatingWebhookConfigurationList` 
+- `admissionregistration.k8s.io/v1.MutatingWebhookConfigurationList` 
+- v1.List with items of admissionregistration.k8s.io/v1.ValidatingWebhookConfiguration 
+- v1.List  with items of admissionregistration.k8s.io/v1.MutatingWebhookConfiguration.
+This configuration is loaded  as static webhooks in api servers. These
+webhooks are called by validating and mutating admission plugins along with
+dynamically loaded webhooks for relevant admission requests. This also means
```
liggitt, line 108: I'd like to see more discussion about the implications of this bit:

1. are webhooks in the manifest file required to have names? unique names (within the file)? (I would expect yes to both)
2. what happens if a webhook in the manifest file and an API webhook object have the same name? (I would expect both to be invoked, as they are effectively in different domains)
3. would metrics and audit annotations from these webhooks be surfaced identically or differently (see admissionmetrics.Metrics calls)? it is likely that config-file-based webhooks would be monitored differently, since they are fully under the control of the kube-apiserver admin and any issues with them can only be resolved by the admin

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+that these webhooks are not API visible objects and hence cannot be modified by
+an API user.
```
tallclair, line 110: Are the static webhooks ordered differently from the dynamic webhooks? Or just mixed in? This is mostly only relevant to mutating hooks.

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+
+- A robust admission webhook registration process where there is no period of
+  time between the kube-apiserver coming up and the registration of an admission
+  webhook.
+
+- The registration of webhooks registered through this process should be
+  protected from alteration by API requests made by cluster users.
+
+- The new registration process should also have a highly privileged means of
+  being able to alter webhook registrations.
```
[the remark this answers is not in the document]
vivekbagade: Yes. Changed it to be more clear hopefully

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
```diff
@@ -0,0 +1,149 @@
...
+`AdmissionConfig`.
+
+```yaml
+apiVersion: apiserver.config.k8s.io/v1
+kind: AdmissionConfiguration
+plugins:
+- name: ValidatingAdmissionWebhook
+  configuration:
+    ....
+    staticConfigFile: "<path-to-static-config-file>"
```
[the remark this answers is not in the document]
vivekbagade: changed it to webhooksFile and pivoted from static to 'manifest-based' throughout the proposal

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
vivekbagade: Makes sense. Removed

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
vivekbagade: 1. Yes to both. Added more info.
2. Both will be invokes. Added more info.
3. Good point. added more info.

keps/sig-api-machinery/20200421-static-admission-webhook-configuration.md
vivekbagade: Manifest based webhooks won't provide any guarantees about ordering. I feel that it would be out of the scope of the KEP.

Please let me know if you feel differently

vivekbagade:
@liggitt Please give this another review. I went through the comments and addressed them.

keps/sig-api-machinery/1872-manifest-based-admission-webhooks.md
```diff
@@ -0,0 +1,216 @@
...
+Metrics and audit annotations for webhooks registered through the manifest file
+will be surfaced differently. This allows the administrator to unique monitor
+activity of these webhooks as they would be the only party able to take action
+against issues with these webhooks. 
+
+All admission metrics for webhooks will have an additional label `manifest_based`
+with a value `true` for webhooks registered via manifest.
+
+Audit annotations for mutation webhooks will include another field `manifestBased`
+with a value `true` for webhooks registered via manifest.
```
liggitt, line 147: add :

```
<<[UNRESOLVED sig-instrumentation ]>>
get feedback on how to indicate this in metrics
<<[/UNRESOLVED]>>
```

keps/sig-api-machinery/1872-manifest-based-admission-webhooks.md
```diff
@@ -0,0 +1,216 @@
...
+  configuration:
+    ....
+    webhooksFile: "<path-to-manifest-file>"
+```
+
+### Reconfiguring manifest file
+
+The manifest file is watched so that the webhook configuration can be
+dynamically changed by editing the contents of the file.
+
```
liggitt, line 172: add

```
<<[UNRESOLVED]>>
need to define behavior in the following cases:
1. file goes missing
2. file cannot be read
3. parse error reading file contents
4. some webhook configurations listed in the file do not pass validation
<<[/UNRESOLVED]>>
```

keps/sig-api-machinery/1872-manifest-based-admission-webhooks.md
```diff
@@ -0,0 +1,216 @@
...
+#### Beta -> GA Graduation
+
+#### Removing a deprecated flag
+
+### Upgrade / Downgrade Strategy
+
+### Version Skew Strategy
+
+## Production Readiness Review Questionnaire
```
liggitt, line 189: update with latest from https://raw.githubusercontent.com/kubernetes/enhancements/master/keps/NNNN-kep-template/README.md

liggitt:
thanks, I noted a few places that need to add an UNRESOLVED comment to make it clear the section is still in progress

go ahead and move the file to `keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md` and populate `keps/sig-api-machinery/1872-manifest-based-admission-webhooks/kep.yaml` from https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template/kep.yaml

then this looks reasonable to merge and iterate on, filling out the remaining sections and answering the unresolved questions.

liggitt:
/lgtm
/approve

thanks, this is at a reasonable spot to merge and iterate on

vivekbagade:
@deads2k Could you have a look too and approve if it looks good

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -0,0 +1,382 @@
...
+
+All admission metrics for webhooks will have an additional label `manifest_based`
+with a value `true` for webhooks registered via manifest.
+
+Audit annotations for mutation webhooks will include another field `manifestBased`
+with a value `true` for webhooks registered via manifest.
+
+```
+<<[UNRESOLVED sig-instrumentation ]>>
+get feedback on how to indicate this in metrics
```
deads2k, line 132: thanks

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -0,0 +1,382 @@
...
+```
+
+### Reconfiguring manifest file
+
+The manifest file is watched so that the webhook configuration can be
+dynamically changed by editing the contents of the file.
+
+```
+<<[UNRESOLVED]>>
+need to define behavior in the following cases:
```
deads2k, line 162: one advantage to running from a file on disk is that, "this configuration is used".  For certificates, you can externally observe when the change is picked up.

I'd like to have a mechanism for indicating when this change has taken affect.

deads2k:
/approve

I think this is good to merge as provisional, but I'd like to see some more exploration on a few points before going to implementable.

PR #1720, opened 2020-04-28
Conformance Profiles additions 

Jefftree:
/assign @johnbelamaric

### DIFF
--- a/keps/sig-architecture/1618-conformance-profiles/README.md
+++ b/keps/sig-architecture/1618-conformance-profiles/README.md
@@ -88,6 +88,9 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Role: Plugin API &amp; Plugin Developers](#role-plugin-api--plugin-developers)
     - [Role: Application Packager](#role-application-packager)
     - [Role: End User](#role-end-user)
+    - [Role: SIG](#role-sig)
+    - [Role: Vendor](#role-vendor)
+    - [Role: CNCF Conformance Program](#role-cncf-conformance-program)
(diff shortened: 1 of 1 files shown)


johnbelamaric:
/approve
/lgtm

/hold
for @spiffxp 
/assign @spiffxp

keps/sig-architecture/1618-conformance-profiles/README.md
```diff
@@ -267,10 +272,36 @@ deploy.
...
+
+As a reviewer part of the CNCF Conformance Program, I must be able to evaluate a
+vendor submission for key criteria:
+
+- Must confirm that the version of the tests being run matches the version being
+  certified
+- Must confirm the set of tests being run matches the set of tests for the
+  version (+ profile(s)) being certified
+- Must confirm that all behaviors are covered by a test that executes, and that
+  no tests fail
```
spiffxp, line 303: I'm less certain whether behaviors need to bubble up to the level of the CNCF Conformance Program.  I think it's up to the Kubernetes project itself to ensure the set of tests for a profile cover all the behaviors for a profile.  It's up to the CNCF to verify that all tests passed.

spiffxp:
/lgtm
/hold cancel
thanks

keps/sig-architecture/1618-conformance-profiles/README.md
Jefftree: I agree that behaviors don't need to bubble up, but it'll probably be inadvertently included since a profile consists of a set of behaviors which map to the tests. To verify that tests belong to a profile, including information about behaviors seems unavoidable.

Anyway, this isn't something that we need to explicitly define so removed the point on behavior coverage.

PR #1724, opened 2020-04-29
fix error url

ttonline6:


mattfarina:
@ttonline6 thanks for the PR. All contributions to Kubernetes require a signed Contributor License Agreement. Can you please sign the agreement and follow the directions here. We cannot accept the PR until that happens.

PR #1740, opened 2020-05-04
Add alpha criteria

xing-yang:
This PR adds alpha criteria to the volume health monitoring KEP.

### DIFF
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -1,38 +1,16 @@
----
-title: PV Health Monitor
-authors:
-  - "@NickrenREN"
-  - "@xing-yang"
-owning-sig: sig-storage
-reviewers:
-  - "@msau42"
-  - "@jingxu97"
-  - "@liggitt"
-  - "@gnufied"
-approvers:
-  - "@msau42"
-  - "@saad-ali"
-  - "@thockin"
-  - "@liggitt"
-editor: TBD
-creation-date: 2019-05-30
-last-updated: 2020-01-23
-status: implementable
-
----
-
-# PV Health Monitor
+# KEP-1432: Volume Health Monitor
 
 ## Table of Contents
 
 <!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
 - [Motivation](#motivation)
 - [User Experience](#user-experience)
   - [Use Cases](#use-cases)
 - [Proposal](#proposal)
 - [Implementation](#implementation)
   - [CSI changes](#csi-changes)
-    - [Add GetVolume RPC](#add-getvolume-rpc)
+    - [Add ControllerGetVolume RPC](#add-controllergetvolume-rpc)
     - [Add Node Volume Health Function](#add-node-volume-health-function)
   - [External controller](#external-controller)
     - [CSI interface](#csi-interface)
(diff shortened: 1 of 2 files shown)


keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -537,6 +538,13 @@ This optional service is out of scope for this KEP but may be introduced in a fu
 This option is not in the main proposal because it is a push-based method while CSI uses pull-based method.
 
 ## Graduation Criteria
+### Alpha
+* Initial feature implementation, including:
+  * External controller volume health monitoring.
```
msau42, line 543: Is there a difference between the features implemented in alpha vs beta?

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -537,6 +538,13 @@ This optional service is out of scope for this KEP but may be introduced in a fu
...
+  * External node agent volume health monitoring.
+* Implementation in the csi-mock driver.
+* Add corresponding unit tests.
```
msau42, line 546: The tests listed there are specifying error codes, which I believe we removed from the CSI spec proposal. Do we need to update the KEP with the csi spec updates?

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Feature should be the same.  Just add more tests (especially E2E) for beta.

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Will do.

xing-yang: Done

xing-yang:
Added a section for "Production Readiness Review Questionnaire".

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -85,7 +88,9 @@ There could be conditions that cannot be reported by a CSI driver. One or more n
 The Kubernetes components that monitor the volumes and report events with volume health information include the following:
 
 * An external monitoring node agent on each kubernetes worker node.
+  * Node agent reports events on the pod object. If multiple pods are using the same volume, report events on all pods.
```
msau42, line 91: Does this mean every node agent has to watch every pod? I think this could be quite expensive. I wonder if we need to consider direct kubelet support instead.

keps/sig-storage/20190530-pv-health-monitor.md
```diff
@@ -32,7 +32,7 @@ status: implementable
 - [Proposal](#proposal)
 - [Implementation](#implementation)
   - [CSI changes](#csi-changes)
-    - [Add GetVolume RPC](#add-getvolume-rpc)
+    - [Add ControllerGetVolume RPC](#add-controllergetvolume-rpc)
```
msau42, line 35: I think the metadata section above can be removed with the new kep format.

keps/sig-storage/20190530-pv-health-monitor.yaml
```diff
@@ -0,0 +1,38 @@
+title: PV Health Monitor
```
msau42, line 1: It looks like the new organization for keps is to put them in a directory, and the enhancement issue is the kep number? See https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture for examples

xing-yang: Any idea how to generate the KEP number?  I somehow remember the KEP number was there in the past but got removed sometime ago.  None of my KEPs have a number at the moment.

msau42: I think the kep number is the enhancement issue number

xing-yang: Got it.  That makes sense:).

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Done

keps/sig-storage/20190530-pv-health-monitor.md
xing-yang: Node agent watches PVC health.  If a PVC has an abnormal health condition, an event will to reported to the Pod that is using the PVC.

We can consider direct Kubelet support in Beta?

msau42: That's fine but we should start the conversation now. I think it may be a complex topic as I think sig-node wanted to move monitoring out of kubelet.

PR #1741, opened 2020-05-04
Update completed tasks and alpha target to 1.19

alculquicondor:
/sig scheduling


alculquicondor:
@assign @Huang-Wei @ahg-g

PR #1745, opened 2020-05-06
Mark the storage version API and mechanism as implementable

caesarxuchao:
/assign @lavalamp @deads2k


lavalamp:
This is just an oversight from https://github.com/kubernetes/enhancements/pull/1361, right? Were there any open discussion items?

caesarxuchao:
Right, it was an oversight. No new discussion.

PR #1754, opened 2020-05-08
KEP-1753: Kubernetes system components logs sanitization

44past4:


44past4:
cc @wojtek-t @immutableT @PurelyApplied

@kubernetes/sig-instrumentation-feature-requests

44past4:
@cc dims

dashpole:
Some feedback based on the discussion at Sig-Instrumentation today:

Static analysis seems like an alternative to sanitation at runtime, although as you point out it can be complementary.  Can you still add that to the alternatives section?  In particular, provide example scenarios where runtime enforcement would prevent logging of sensitive data that static analysis would not catch.

If you can demonstrate that static analysis alone is not sufficient to meet the requirements of the audit report, I think that is sufficient justification for the enhancement.

44past4:
I have updated the KEP proposal with the alternatives section discussing Static code analysis @dashpole, @ehashman, @brancz please take a look.

keps/sig-instrumentation/1753-logs-sanitization/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+##### Practical 
+
+A major practical limitation of static analysis concerns array references and evaluation of pointer variables. Array subscripts and pointers are mechanisms for selecting a data item at runtime, based on previous computations performed by the program. Static analysis cannot evaluate subscripts pointers and, thus, is unable to distinguish between elements of an array or members of a list. Although it might be possible to analyze subscripts and pointers using techniques of symbolic execution, it is generally simpler and more efficient to use dynamic testing.
```
dashpole, line 235: Just to make sure I understand, you are saying that static analysis wouldn't be able to catch me logging (from your example above):

```golang
cm := ConfigMap{Data: map[string]string{"password": "123456"}}
klog.Errorf("Configmap with password: %+v", &cm)
```
https://play.golang.org/p/T-jcUceYQpb

PurelyApplied: **tl;dr: pretty much**

I think `unable to distinguish` might be stronger than is strictly, categorically true, but it's not far from the mark.  I'd say it breaks down further into two possible points of failure: (a) failure in analysis configuration, and (b) difficulty in analysis for identification.  This snippet touches both.

(a) In the snippet above, we are storing sensitive information in `field Data map[string]string`.  From a static analysis standpoint, it's difficult to say "some key-value pairs might be sensitive."  If we said the entire field should never reach a sink, that is easier, but obviously more limiting than necessary.  But without some other prior knowledge, there's no reason to think an arbitrary `map[string]string` is containing something sensitive.  We might make some progress saying "any `map[string]string` entry with key `"password"` is sensitive."  Does that include `map[string]interface{}` or `map[interface{}]string`?  Do we need to do fuzzy-matching to include the keys `"passwd"` or `"PWORD"`?  Are we likely to produce false-positives if we match `"pass"`?

(b) Suppose we solve the issue of configuration and know `ConfigMap.Data` could contain a sensitive key-value pair.  This explicit reference in the above snippet would correspond to [ssa](https://pkg.go.dev/golang.org/x/tools/go/ssa) method `ssa.MapUpdate`.  We identify the `MapUpdate` using the key `"password"` and a non-empty value.  We could possibly mark the register as containing a secret and track it during analysis to ensure it doesn't reach a sink.

However, that requires an explicit `MapUpdate` with the key `"password"`.  It would miss any `ConfigMap` that arrived on the wire.  Even something as simple as [obfuscating the underlying type with an interface](https://play.golang.org/p/p1HVWGYEUtx) would require some level of pointer analysis to identify the argument reaching the sink.

So, while we might catch that two-line snippet, we're likely to miss much more in practice.

dashpole: Does that mean we can't use static analysis at all with klog?  All of the Infof/Warningf/Errorf functions take `args ...interface{}`.

PurelyApplied: It would be an issue of the static type of the argument at the callsite, not in the call signature.

For instance, it would be pretty easy with `ssa` to identify the following:
```golang
func foo(data SuperSecretType) {
  klog.Infof("%v", data)
}
```
Here, the type of `data` at the `klog.Infof` call is concrete at compile time

Conversely, like the playground link in my previous comment:
```golang
func bar(data SuperSecretType) {
  leak(data)
}

func leak(data interface{}) {
  klog.Infof("%v", data)
}
```

Here, the type of `data` at the `klog.Infof` call is `interface{}` and its concrete type is not known to `ssa` at the callsite.

There could be a path forward if the function `leak` were identified as a data sink to avoid, either by manual configuration or a separate analysis step.  Pointer analysis could help here, but that requires full-program analysis and can get computationally expensive.

keps/sig-instrumentation/1753-logs-sanitization/README.md
```diff
@@ -0,0 +1,243 @@
...
+Currently the klog library used by Kubernetes does not provide any extension point which could be used for filtering logs.
+
+We propose to add a new interface to klog library:
+
+```go
+type LogFilter interface {
+  Filter(args []interface{}) (args []interface{})
+  FilterF(format string, args []interface{}) (string,[]interface{})
+  FilterS(msg string, keysAndValues []interface{}) (string, []interface{})
+}
```
logicalhan, line 173: Why are we exposing this interface?

keps/sig-instrumentation/1753-logs-sanitization/README.md
```diff
@@ -0,0 +1,243 @@
...
+
+For external types which are not part of the Kubernetes source code such as go lang standard libraries or third party vendored libraries for which we cannot change source code we will define a global function which will map those into relevant datapolicy tag:
+
+```go
+func GlobalDatapolicyMapping(val interface{}) []string
+```
+
+### datapolicy verification library
+
+datapolicy verification library will implement verification if the provided value does not contain any sensitive data identified by the datapolicy tag. Verification will be performed using reflection. Recursion will stop on pointers as values for those are usually not logged. Validation will depend only on the datapol tags values. Values of individual fields of primitive types like string will not be analyzed other than checking if a given field is not empty.
```
logicalhan, line 151: "datapolicy verification library will implement verification if the provided value does not contain any sensitive data identified by the datapolicy tag." doesn't make sense to me. 

Don't you have to perform verification in order to figure out if it may contain sensitive data to begin with? Am I misunderstanding this?

logicalhan: Would you mind explicitly drawing out the difference between verification and validation (if there is one)?

keps/sig-instrumentation/1753-logs-sanitization/README.md
immutableT: We absolutely **can** use static analysis with functions that accept `args ...interface{}`. We can do this because Go's needs to perform conversions to `interface{}` and we can detect such conversions and thus determine the original type. 
Even with the second example provided by @PurelyApplied, the analyzer could infer the fact that the `SuperSecretType` was sent to the `leak` function and consequently treat the `data` argument as a source (when the callee is `bar`). I agree that with  @PurelyApplied that this requires full program analysis (most Static Analysis tools have such capabilities).

To summarize the answer provided by @PurelyApplied:
Both Static and Dynamic analysis would have a lot of problems with the original example provided by @dashpole . This is due to the fact that both Static and Dynamic analysis rely on the input supplied by users (domain experts) that identify potential sources of PII (ex. `Secret`) and sinks (ex. `klog.Infof`). 
Clearly, we can't simply tell an analyzer to consider and `map[string]string` as a source of PII - this will lead to a large number of false positives.

So how would one handle @dashpole's scenario? There are a couple of ways:
1. Discourage (through code review) placement of PII into generic types such as maps. Instead,  require use of specialized types (ex. `Secret`) to hold security sensitive data.
2. Detect data with high levels of entropy (tokens and keys) in the logging pipeline. Since this type of analysis could be handled outside of k8s it was not mentioned in this KEP.

To answer @dashpole's original question, regarding an example where Dynamic Analysis would have advantages over static Analysis, below is (a somewhat contrived) example.
At runtime, it will be easy to see that a `Secret` is being sent to `stdout`.
In principle, a static analysis tool could also catch this, but IMO it would not be the right tool for this sort of scenarios. Therefore, we should consider a combination of both analysis: static and dynamic.

```
type value interface {
 value() string
}

type Secret struct {
 v string
}

func (s *Secret) value() string {
 return s.v
}

type NonSecret struct {
 v string
}

func (s *NonSecret) value() string {
 return s.v
}

func main() {
 values := []value{
  &Secret{v: "pwd"},
  &NonSecret{v: "hello"},
 }

 for _, v := range values {
  if len(v.value()) < 4{
   fmt.Printf("%v", v)
  }
 }
}


```

keps/sig-instrumentation/1753-logs-sanitization/README.md
44past4: We do not want to implement sanitization logic as an integral part of the klog but instead we want to inject it whenever the Kubernetes component is configured to do so. Therefore we need to expose this interface so that it could be implemented in Kubernetes.

keps/sig-instrumentation/1753-logs-sanitization/README.md
44past4: There is no difference between verification and validation in this context. I have updated a description a bit to make it clearer and aligned it to use verification term.

dashpole:
I'm sharing my feedback here, since we ran out of time at the meeting this morning.  I agree with the motivation and direction of the KEP.  Given that static analysis is already being done (just not as a presubmit), and that the authors plan to implement static analysis soon after this KEP is accepted, I don't have any issues with doing dynamic before static.

One idea: While this feature is in alpha and is disabled by default, we could accept the advice given by the presenters, and enable it during some oss e2e tests.  As long as the overhead isn't so bad that tests fail, interested parties can search for the message you get when something is filtered out (e.g. https://github.com/dashpole/gubernator_search).  This would be a way for us to get value from this feature while it is still in alpha.

I agree it would be good to get reviews of the new API from @thockin, @DirectXMan12, and @serathius.

keps/sig-instrumentation/1753-logs-sanitization/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+For external types which are not part of the Kubernetes source code such as go lang standard libraries or third party vendored libraries for which we cannot change source code we will define a global function which will map those into relevant datapolicy tag:
+
+```go
+func GlobalDatapolicyMapping(val interface{}) []string
+```
+
+### datapolicy verification library
+
+datapolicy verification library will implement logic for checking if the provided value contains any sensitive data identified by the datapolicy tag. Verification will be performed using reflection. Recursion will stop on pointers as values for those are usually not logged. Verification will depend only on the datapol tags values. Values of individual fields of primitive types like string will not be chec ked other than checking if a given field is not empty.
```
serathius, line 150:
```suggestion
datapolicy verification library will implement logic for checking if the provided value contains any sensitive data identified by the datapolicy tag. Verification will be performed using reflection. Recursion will stop on pointers as values for those are usually not logged. Verification will depend only on the datapol tags values. Values of individual fields of primitive types like string will not be checked other than checking if a given field is not empty.
```

serathius: `will not be checked other than checking if a given field is not empty.` Why we are planning to check if field is empty?

serathius:
Design and motivation looks good to me. Changes to klog make sense and align with current direction moving logging configuration into component-base.

I agree with David note about running e2e tests with this mechanism enabled. I think it makes sense to add it to Test Plan to provide immediate benefit. Due performance problems we could either do it in Beta or as a separate e2e periodic.

keps/sig-instrumentation/1753-logs-sanitization/README.md
44past4: The assumption is that empty string cannot leak any sensitive data.

44past4:
@dashpole Taking into account Marek's comments on the klog apart from the suggestion to enable this feature for some of the e2e tests which I will include in the Test Plan section for this feature, is there anything more we need to address for this enhancement to be approved?

PR #1758, opened 2020-05-08
Update x509 extension handling to match signerName implementation

liggitt:
Update the signerName definitions to clarify x509 extension handling (matches docs released with 1.18 at https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/#kubernetes-signers)

* Enumerate specific SAN types (Email, URI, IP, DNS)
* `kubernetes.io/kube-apiserver-client` ensures the request is for [client key usage](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/signer/signer.go#L148-L149), and [passes through requested extensions](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/controller/certificates/authority/authority.go#L70-L82), which get truncated to [just  URI/Email/IP/DNS SANs by the signer](https://github.com/kubernetes/kubernetes/blob/release-1.18/pkg/controller/certificates/authority/policies.go#L37-L69)


enj:
/lgtm
/hold

To let others review.

liggitt:
Thanks. 

/hold cancel

PR #1766, opened 2020-05-11
Update kep validations to support new kep.yaml fields

wojtek-t:
We can add more validations later - I just want to unblock presubmits for correct KEPs now.


johnbelamaric:
Thanks Wojtek.

/lgtm
/approve

PR #1796, opened 2020-05-18
Add Pod Topology Spread KEP to new template

Huang-Wei:
ref #895

/sig scheduling

### DIFF
--- a/keps/sig-scheduling/895-pod-topology-spread/README.md
+++ b/keps/sig-scheduling/895-pod-topology-spread/README.md
@@ -1,25 +1,4 @@
----
-title: Even Pods Spreading
-authors:
-  - "@Huang-Wei"
-owning-sig: sig-scheduling
-reviewers:
-  - "@bsalamat"
-  - "@lavalamp"
-  - "@krmayankk"
-  - "@ahg-g"
-  - "@alculquicondor"
-approvers:
-  - "@ahg-g"
-  - "@alculquicondor"
-creation-date: 2019-02-21
-last-updated: 2020-01-21
-status: implementable
----
-
-# Pod Topology Spread
-
-## Table of Contents
+# KEP-895: Pod Topology Spread
 
 <!-- toc -->
 - [Release Signoff Checklist](#release-signoff-checklist)
@@ -44,19 +23,28 @@ status: implementable
   - [Graduation Criteria](#graduation-criteria)
 - [Alternatives](#alternatives)
 - [Impact to Other Features](#impact-to-other-features)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
 
 ## Release Signoff Checklist
 
-- [x] kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
-- [x] KEP approvers have set the KEP status to `implementable`
-- [x] Design details are appropriately documented
-- [x] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
-- [x] Graduation criteria is in place
+- [x] (R kubernetes/enhancements issue in release milestone, which links to KEP (this should be a link to the KEP location in kubernetes/enhancements, not the initial KEP PR)
+- [x] (R) KEP approvers have set the KEP status to `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
 - [x] "Implementation History" section is up-to-date for milestone
 - [x] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [x] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 ## Terms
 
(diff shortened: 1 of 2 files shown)


keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Even Pods Spreading
+title: Pod Topology Spread
```
alculquicondor, line 2: This can be removed from here now.

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+### Rollout, Upgrade and Rollback Planning
+
+* **How can a rollout fail? Can it impact already running workloads?**
+
+  Since this feature requires users to opt-in by setting new field in pod spec,
+  it should not impact already running workloads.
+
+* **What specific metrics should inform a rollback?**
+
+  Metric "schedule_attempts_total" remaining at zero when new pods are added.
```
alculquicondor, line 570: Pods would still be attempted. I would add that the "result" sub-field can be used to check for a spike on errors or unschedulable results.
Also, it's worth looking at latency metrics.

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+* **Is the rollout accompanied by any deprecations and/or removals of features,
+  APIs, fields of API types, flags, etc.?**
+
+  No.
+
+### Monitoring requirements
+
+* **How can an operator determine if the feature is in use by workloads?**
+
+  Operator can query `pod.spec.topologySpreadConstraints` field and identify if
```
alculquicondor, line 585: Also the metric "plugin_execution_duration_seconds" which has a "plugin" sub-field

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+  this is being set to non-default values.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to
+  determine the health of the service?**
+
+  N/A.
```
alculquicondor, line 591: The metric "plugin_execution_duration_seconds" is the perfect candidate.

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+
+  N/A.
```
alculquicondor, line 595: We can mention certain numbers from our perf results.

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+* **Will enabling / using this feature result in any new calls to cloud
+  provider?**
+
+  No.
+
+* **Will enabling / using this feature result in increasing size or count
+  of the existing API objects?**
+
+  Since this feature adds a new field to pod's spec, it will increase API size
+  of Pod object depending on how many `topologySpreadConstraints` are specified.
```
alculquicondor, line 627: we can add that typically a Pod would only require 1 or 2.

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+
+* **Will enabling / using this feature result in increasing time taken by any
+  operations covered by [existing SLIs/SLOs][]?**
+
+  No.
```
alculquicondor, line 632: `framework_extension_point_duration_seconds` and other latency SLI

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -520,6 +529,128 @@ simply achieve the semantics of "PodAntiAffinity in zones" via a combination of
...
+* **Will enabling / using this feature result in non-negligible increase of
+  resource usage (CPU, RAM, disk, IO, ...) in any components?**
+
+  No.
+
+### Troubleshooting
+
+* **How does this feature react if the API server and/or etcd is unavailable?**
+
+  Running workloads won't be impacted. Submissions of new workloads using this
```
alculquicondor, line 643: I would say something like: "no new scheduling impact", but PRR to confirm.

keps/sig-scheduling/895-pod-topology-spread/README.md
alculquicondor: Also perhaps Pod Events

keps/sig-scheduling/895-pod-topology-spread/README.md
ahg-g: I would say an SLI/SLO can be setup for pod scheduling latency, this feature is replacing default spreading, so it is unlikely to increase pod scheduling latency. ```framework_extension_point_duration_seconds``` will not be a metric that we use to measure an SLO, it is more for debugging purposes.

keps/sig-scheduling/895-pod-topology-spread/README.md
alculquicondor: That might not be possible to do, since this is a Pod-level feature.

keps/sig-scheduling/895-pod-topology-spread/README.md
alculquicondor: That is true, but there is a difference: default spreading uses only the nodes that pass filters, whereas we use all nodes in Topology spreading. It might we worth saying that some increase in latency might happen, and that it can be monitored with the above metric.

keps/sig-scheduling/895-pod-topology-spread/README.md
ahg-g: I think it is possible, but needs work, and should likely be done by an external component. One can also simplify things by for example monitoring only Deployments that use topology spread in their template...

alculquicondor: In that case, we can put it as a suggestion for operators.

keps/sig-scheduling/895-pod-topology-spread/README.md
Huang-Wei: But https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml still has still field.

Huang-Wei:
Thanks for the review.

Comments addressed. @alculquicondor @ahg-g PTAL.

keps/sig-scheduling/895-pod-topology-spread/README.md
Huang-Wei: Ah, that's it...

keps/sig-scheduling/895-pod-topology-spread/README.md
```diff
@@ -583,16 +586,28 @@ extra benefit of this KEP.
...
+    - "convert to podtopologyspread.preFilterState error"
+    - "hard topology spread constraints"
+    - "internal error: get paths from key"
+  - Frequency of regular scheduling failures (with keyword "failed
+    spreadConstraint") in scheduler log.
 
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
 
-  N/A.
+  - Metric `plugin_execution_duration_seconds{plugin="PodTopologySpread"}` < 500ms.
```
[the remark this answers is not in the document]
Huang-Wei: How about <= 100 ms on 90-percentile?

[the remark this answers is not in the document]
alculquicondor: That's still high, but more reasonable. But I think we can mention the number of pods for that SLO to hold.

Huang-Wei:
@alculquicondor I'm not sure it's mandatory to have a PRR reviewer for each KEP. If so, @wojtek-t could you kindly review the PRR part? Thanks.

/assign @wojtek-t 
(feel free to unassign or assign others)

alculquicondor:
It's not enforced for 1.19, but it will be in the future. So I think I could LGTM, given that the freeze is today. Could you squash?

Huang-Wei:
@alculquicondor squashed.

PR #1800, opened 2020-05-20
Fix status of NonPreempt Kep

denkensk:
ref https://github.com/kubernetes/enhancements/issues/902
/sig scheduling


palnabarun:
LGTM!

Thanks for updating it. :slightly_smiling_face: 

/lgtm

PR #1811, opened 2020-05-26
Make NodeIPAM controller support variable sized CIDRs

liuxu623:


keps/sig-network/20200525-NodeIPAM-controller-support-variable-sized-CIDRs.md
```diff
@@ -0,0 +1,115 @@
+---
+title: NodeIPAM controller support variable sized CIDRs
```
thockin, line 2: This is annoying, but there is a newer KEP template:  https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template

Now it's a directory with a slightly different structure

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
```diff
@@ -0,0 +1,645 @@
...
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+We need use a new data structure instead of bitmap to allocate podCIDR, it should support variable sized CIDRs,
```
thockin, line 265: What I was really looking for is a plan for how you can consume and existing bitmaps and convert to the new struct.  But in digging, it seems we don't actually STORE the allocated CIDRs anywhere, so maybe you're lucky and this is easier than I thought?

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
```diff
@@ -0,0 +1,645 @@
...
+<!--
+This section should contain enough information that the specifics of your
+change are understandable.  This may include API specs (though not always
+required) or even code snippets.  If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+We need use a new data structure instead of bitmap to allocate podCIDR, it should support variable sized CIDRs, 
+interval tree is suitable. I found etcd already implement an interval tree 
+(https://github.com/etcd-io/etcd/blob/master/pkg/adt/interval_tree.go).
```
thockin, line 267: Are there good benchmarks that show the memory footprint of such a thing?  I'd like to see some evidence that using it to store thousands of CIDRs uses reasonable amounts of memory.

liuxu623: profile001
I have add benchmarks.

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
aojea: yeah, the bitmap is not persisted as in the allocators, the data is persisted in the Node object
https://github.com/kubernetes/kubernetes/blob/323f34858de18b862d43c40b2cced65ad8e24052/pkg/controller/nodeipam/ipam/range_allocator.go#L226-L238

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
aojea: I think that it would be better to compare the results of the benchmarks between the old and the new implementation, ie. with 
https://pkg.go.dev/golang.org/x/tools/cmd/benchcmp

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
```diff
@@ -0,0 +1,677 @@
...
+		end.Lsh(end, uint(bits-ones)).Add(begin, end)
+	}
+	return adt.NewStringInterval(
+		fmt.Sprintf("%0"+strconv.Itoa(length)+"s", begin),
+		fmt.Sprintf("%0"+strconv.Itoa(length)+"s", end),
+	), nil
+}
+```
+
+### Test Plan
```
aojea, line 306: it needs at least unit tests here:
for the new data structure pkg/controller/nodeipam/ipam/cidrset/cidr_set_test.go
exercising the new functionality pkg/controller/nodeipam/ipam/range_allocator_test.go

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
aojea: :thinking: , how will work 2 controllers with different cidr_set implementations at the same time?
at least the old one will not be able to do resizing o cirds

thockin: Yeah, we'll need a very well thought out deployment plan.  Specifically, covering cases like:

* Have 3-controller-manager instances
* Upgrade 1 of them to this new model
* Allocate node with the new CID size
* Leader election changes which controller-manager is in charge

During that period, some masters will still be using bitmaps.  That is going to end badly. 

We'd need at least a multi-stage rollout, where we convert controllers to use intervals but not change the config.  Once that is rolled out you can change the config.

But you also need to handle rollbacks, so you can't go back into a corrupt state.

Finally, is it really what people want to have a fixed CIDR size?  Maybe we should be shooting for something more policy oriented?

danwinship:
So... does this mean that https://github.com/kubernetes/kubernetes/issues/57130 is no longer the plan?

thockin:
No, I still think it's totally wrong, but it is supported, and I am not
against making it work slightly better, if it is not intrusive.  I'd love
to see a plan to kill it off, but I just haven't had the time myself.

keps/sig-network/20200617-NodeIPAM-controller-support-variable-sized-CIDRs/README.md
liuxu623: profile002

thockin:
Is this KEP something we want to revive?

thockin (commented):
This is still open, but I still have concerns.

It seems to be aiming at one very small use-case: changing the CIDR size.  It still leaves you with a fixed size for all nodes and a single CIDR to draw from.  That doesn't make it a very compelling risk to take.

What if, instead, we actually started making the IPAM model less clunky?

Now, I am totally making this up as I go, but just laying out the idea:

What if we had a new CIDRAllocatorType that allowed multiple CIDRs?  Could we allow each node to request a different size block?  Maybe allocations become a new type of object, like a PersistentVolume or maybe node itself carries the request.

If we REALLY need to solve this smaller-scope problem, I think the best course would be to fork and make a whole new IPAM controller rather than modify in-place.

PR #1871, opened 2020-06-24
Storage Capacity: remember NodeSelector problem

pohly:
As pointed out during alpha code
review (https://github.com/kubernetes/kubernetes/pull/91939/files#r443081515),
using NodeSelector is a too broad API because it includes filtering by
fields, which is not supported. For v1alpha1, we validate that all
field filters are empty. A better solution would be to support fields
or use a different type.


pohly:
This became obsolete: the current API proposal uses LabelSelector instead of NodeSelector.

PR #1878, opened 2020-06-30
KEP: Efficient watch resumption after kube-apiserver reboot

wojtek-t:


keps/sig-api-machinery/xxxx-progress-notify/README.md
```diff
@@ -0,0 +1,456 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+In order to support large Kubernetes clusters without huge number of etcd
```
lavalamp, line 70: This is more like background than a summary.

keps/sig-api-machinery/xxxx-progress-notify/README.md
```diff
@@ -0,0 +1,456 @@
...
+is send to the watcher.
+
+We are going to utilize this feature to solve the problems described above.
+
+1. We will work with etcd team to make the interval configurable. It should
+   be fairly simple based on: https://github.com/etcd-io/etcd/pull/11463
+
+1. We modify all watches used to propagate watchcache to set the
+   `WithProgressNotify` option. For watches being serve from etcd in case
+   of disabled watchcache, this should remain unchanged.
```
lavalamp, line 169: Alternatively, we could automatically translate to bookmarks, no?

keps/sig-api-machinery/xxxx-progress-notify/README.md
```diff
@@ -0,0 +1,456 @@
...
+   https://github.com/kubernetes/kubernetes/pull/86769#issuecomment-579171765
+	 for more details, but such small value is probably not needed. The goal is
+   to ensure that before the next kube-apiserver is restarted and watchers
+   attached to it are rebalanced to other replicas, most of them already
+   received bookmarks with resource version higher than the one from other
+   kube-apiservers were initialized. So having the period to be between
+   1 and 5 second in addition to bookmarks sent on shutdown should be enough.
+
+The POC PR can be found in: https://github.com/kubernetes/kubernetes/pull/92472
+
```
lavalamp, line 220: Did you consider the alternative of making apiserver read etcd's history when it starts instead of just the current state? I think that would also address the problem? If that makes sense could you list it as an alternative and state why it's inferior?

keps/sig-api-machinery/xxxx-progress-notify/README.md
lavalamp: I think a summary might be:

The kube-apiserver watch cache starts up with zero changes in its window. As a consequence, clients that want to resume a watch immediately after kube-apiserver reboots almost always have a RV that is out of the window. We propose to fix this by making clients aware of the most recent RV before they are disconnected.

keps/sig-api-machinery/xxxx-progress-notify/README.md
```diff
@@ -0,0 +1,456 @@
...
+- there is one object of type Y: y with rv=101
+- there is a watch W for objects of type X already synced to rv=100 and
+  attched to kube-apiserver-1
+- both kube-apiserver are up-to-date with watchcache for X synced to rv=100
+
+- T1: kube-apiserver-1 is being restarted
+- T2: watch W is being reestablished from rv=100 from kube-apiserver-2
+- T3: kube-apiserver-1 is initialized, with watcache X synced to rv=101
+- T4: y is being updated to rv=102
+- T5: kube-apiserver-2 is being restarted
```
lavalamp, line 127: I think you don't need multiple apiservers to illustrate the problem. Actually the problem is most acute when there's just a single apiserver.

If there were 100 apiservers, this wouldn't be nearly as bad of a problem, because by the time you reboot the final one, there's lots of others with sufficient history in their windows. (Of course, the change rate matters too, but in low change rate systems (which cause more out of window errors) it's also less problematic to do full re lists, as they're less busy.)

So, adding apiservers actually makes it a little better, but just going to 3 or 5 is not sufficient to fix the problem.

It might be clearer to demonstrate the problem with a single apiserver and then explain why adding a few apiservers doesn't fix it.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The kube-apiserver watch cache is initialized from etcd at the moment
+when it starts with empty transaction log. As a consequence, clients
```
lavalamp, line 72:
```suggestion
when it starts with empty change history. As a consequence, clients
```

?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,485 @@
...
+1. If no object of a given resource type is changing after kube-apiserver
+  startup, watch cache may stuck being synced to resource version not being
+	version of any object of a given type for extended period of time. See
+  https://github.com/kubernetes/kubernetes/issues/91073 for more details.
+
+To illustrate the problem better, imagine the following example:
+- there is a single kube-apiserver that we are going to restart
+- there are only two resource types: X and Y (for simplicity)
+- there is one object of type X: x with rv=100
+- there is one object of type Y: y with rv=101
```
lavalamp, line 122: Thanks, this is much easier to follow now, I think you can also remove references to type Y?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,485 @@
...
+   the moment of watch cache initialization), relist is forced
+
+Note that adding more kube-apiservers doesn't solve the problem. In
+fact, it is introducing the second problem where watch cache in different
+instances may be out of sync for long period of time. To achieve such
+situation it is enough to have object y be updated to rv=102 between
+the moment kube-apiserver-1 is initialized and kube-apiserver-2 starts
+its initialization. And we would stuck with kube-apiserver-1 synced to
+rv=101 and kube-apiserve-2 synced to rv=102 until the next change of
+any object of type X happens (even though both watch caches contain the
```
lavalamp, line 140: OK, you're talking about something I guess I didn't understand from the first draft, and actually I still don't think I understand it.

Are you saying that a low-churn resource type could get stuck in the distant past due to an unrelated high-churn type?

Whatever this second problem is, does it actually cause different effects than the first problem?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,485 @@
...
+large production clusters with tens of thousands of pods for which any
+pod was changing for long enough periods of time (e.g. tens of minutes).
+
+The goal of this proposal is to avoid both of these problems.
+
+### Goals
+
+- avoid tons of relists during kube-apiservers rolling upgrades
+- avoid different instances of kube-apiserver stuck with watchcache synced
+  to different resource versions for extended period of time
```
lavalamp, line 155: Ah, OK, now I understand what the second problem is. I think the example needs more detail about when the cache "syncs" to the global index and when in "syncs" to the highest index of an object in the collection.

Can you make the problem example more extreme like this? Imagine a collection with objects a, b, c, having RVs 1, 2, 5000 respectively. Now object c is deleted at an inconvenient time while an apiserver is starting. Does this apiserver now believe it is only synced to RV=2 for this collection?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+Pros:
+- doesn't require any etcd changes
+- no overhead when kube-apiserver is running (only initialization being more
+   expensive)
+
+Cons:
+- given kube-apiserver is performing compaction (default every 5m), lack of
+   changes of any object of type X within that period would result in inability
+   to initialize transaction log anyway; so that is not universal solution
```
lavalamp, line 483: Isn't getting the latest RV the most valuable part, though?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: done

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: Actually it is needed, because I need some non-X-type object that is bumping be global resource version.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: It's not exactly like that.

I have updated this part of clear example (without mixing it with the first problem).
Is it more clear now?

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: Actually it's not the case.

Watchcache claims to be synced to RV of the latest change of the object of a given type. Deletions are handled correctly (so e.g. if the deletion was at RV=6000), it will claim to be synced to 6000, not 2. So that works fine already.

I clairified that in line 89 [I though I wrote that already, but apparently I it wasn't the case.]

The problem is purely with initialization, because initialization is happening NOT from the objects of a given type, but from "global etcd version" (which doesn't have to represent any object of our type).

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: That doesn't solve the second problem though (i.e. you can have multiple kube-apiservers out-of-sync for extended period of time).

Based on your comments above, I think you misunderstood that second problem - so hopefully with new clarifications it's clearer what I mean here.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+- T1: kube-apiserver-1 being restarted
+- T2: kube-apiserver-1 is initialized with watchcache for X synced to
+   rv=101 (most recent etcd version) [as above]
+- T3: object y is being updated to rv=102
+- T4: kube-apiserver-2 is being restarted
+- T5: kube-apiserver-2 is initialized with watchcache for X synced to
+   rv=102 (most recent etcd version)
+
+As long as no object of type X is being touched (created, updated or
+deleted), watchcache for X will not be updates. So even though they
```
lavalamp, line 154: I still do not understand how this is different from the first problem. In both scenarios, for resource X, apiserver has an empty change window and believes X is synced to an RV number which a) is greater than any actual X object change's RV, and b) at which no X change happened.

In the second scenario there are multiple apiservers in the condition at the same time, at different RV numbers, but I don't see how that makes it worse.

I feel like there's some additional fact missing here, like if the RV given by list X from apiserver 2 can't be used to watch X on apiserver 1. But this doesn't seem like a big problem, since most list/watch calls will go to the same apiserver (and if you add a restart to break this, then it really is no different from problem 1). Additionally, for N apiservers in this condition at different RVs, the probability of a split-apiserver list/watch working is probably around 50% since the RV needn't exactly match? If it needed to exactly match the RV, the probability would be like N^-2 and that would get my attention :)

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
lavalamp: OK, that's good to know. The problem I guess is that deletion records aren't loaded on startup.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
lavalamp: Ah, so the actual need is to select the oldest RV that is greater than any change to the objects in the collection.

If it weren't for deleted objects not showing up in the list, it would be trivial, you wouldn't even need to read the change history.

But if we can read the change history, I don't agree with this con. If you can get:
* the list of changes from the history window since the last compaction, and it is empty
* C = RV of the last compaction
* O = greatest RV of an object returned by list
* G = global RV as of the time of the list
and if O < C <= G, then it is safe to pretend that you've synced to C instead of G. (given that the list is empty, this will always be the case. If the list is not empty you can still use C as the RV as long as you fill the revision history in memory with the list.)

This seems like a much less time-sensitive fix than relying on not being about to generate the condition in 250ms.

It may be reasonable to make both fixes, but if we pre-fill the cache like this, then we just need to give the client an RV newer than the last compaction, which is drastically less frequent than 250ms.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: > In the second scenario there are multiple apiservers in the condition at the same time, at different RV numbers, but I don't see how that makes it worse.

So the problem is that if you list from the newer kube-apiserver and then try to establish watch from that point in the older kube-apiserver you get "too old RV". And need to relist again. If you're unlucky, it may be indefinite...

Basically the problem is that you get different responses for the same request from two kube-apiservers for extended period of time, even though there isn't any network partition or anything like that.
I would say it's far from expected...

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: That isn't really it. Because the RV of the LIST, doesn't depend on what was happening before.
And we don't load events at all (if we would, deletions would load correctly, since these are regular watch events).

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
wojtek-t: > Ah, so the actual need is to select the oldest RV that is greater than any change to the objects in the collection.

Probably equal to the most fresh change instead of "great than any change to the objects in this collection" (we need to be able to say from which point in time we can start serving watches and be sure that nothing has happened after that). 
But yeah - this is +-1 difference, so probably doesn't matter.


> the list of changes from the history window since the last compaction, and it is empty

This one is tricky. Because even if we would have a way to expose RV of the last compaction, the all changes since then we can only get via watch. And if we establish watch, we have no way to say if it already synced to now or not...
We can probably solve that with using ProgressNotify again, but I think we would end-up with similarly if not more complicated solution (with the solution I'm proposing there aren't any corner cases)...

lavalamp: OK, here's a very stupid idea that is probably easy to implement and effective.

When filling the watch cache, currently we start watching for changes at the index given by the list.

Imagine a change that, in addition to doing that, also "binary" searches backwards in time with a series of watch requests.

E.g.:
* watch zero is at RV=X. Cache is now synced to X. We let this watch continue. (same watch we do today.)
* watch 1 is at RV=X-1000. We get an out-of-window error.
* watch 2 is at RV=X-500. We get 3 changes with RVs < X; we prepend them to our change list and now our cache is synced to X-500. Close the watch.
* watch 3 is at RV = X-750, out of window.
* watch 4 is at RV = X-625, we get 1 additional change, prepend it, cache is now synced at RV=X-625.

4 iterations is probably more than enough, but you can use this technique to find the exact limit (which might be a moving target!).

wojtek-t: Yeah - I've been thinking about that too. But imagine there are no changes after RV=X.
So when you initialized watch2 from RV=X-500 - how you can answer whether it already reached X or not?

lavalamp: Well, not getting an out-of-window error tells us something?

We could enable progress-notify on these extra watches :)

wojtek-t: > We could enable progress-notify on these extra watches :)

Sure - but with that we need to do most of changes described in the doc anyway...

I guess what I'm saying is that - sure we can do that, but it's actually not that much simpler (if at all).
[Note that we would need to be able to present history, which we don't support now, so another complication to what people, including Jordan already claim to be terribly complicated...]

lavalamp: It is a change, and it makes the watch system more complicated, but it's localized and doesn't change any client-visible interfaces. In that sense it is well contained. We can make sure the testing is good. Personally I feel like making apiserver load prior history adds a lot of robustness to the system, and making the system depend on something happening every 250ms adds brittleness.

But it may be time for @deads2k and @liggitt to weigh in.

wojtek-t: If we would really require these 250ms for something - true. But we can tolerate much more (given kube-apiserver initializes tens of seconds anyway).

wojtek-t: Or I should say it a bit differently. In order to utilize the progress notify to detect whether watch already synced to RV=X, we basically need to bump the frequency to something reasonable from the current 10m anyway. So pretty much all the changes described in this proposal would have to be implemented anyway.

wojtek-t: Or let me actually write it differently:
I'm not opposed to changing how we initialize watchcache, but based on all the arguments above (mostly need for watch progress notify to implement it, the fact that we can still have two apiservers claimed to be synced to different RVs for extended period of time), I would treat it as a complementary not a replacement of what I'm proposing.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+In this KEP we propose how to address this problem.
+
+## Motivation
+
+In order to support large Kubernetes clusters without huge number of etcd
+and kube-apiserver replicas, kube-apiserver contains an in-memory cache
+called `watch cache` from which all watch requests as well as non-quorum
+reads (in opt-in fashion) are served. Watch cache is being propagated
+directly from etcd using typical "list + watch" pattern and stores both
+the current state as well as som recent 'transaction log'. There is a
```
deads2k, line 84: typo: some

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+To illustrate the problem better, imagine the following example:
+- there is a single kube-apiserver that we are going to restart
+- there are only two resource types: X and Y (for simplicity)
+- there is one object of type X: x with rv=100
+- there is one object of type Y: y with rv=101
+- there is a watch W for objects of type X already synced to rv=100
+- watchcache for both X and Y is up-to-date
+
+- T1: kube-apiserver being restarted
+- T2: kube-apiserver is initialized with watchcache for X synced to
```
deads2k, line 132: it initialized this based on a list, correct?  if so, please indicate that inline.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+- T3: object y is being updated to rv=102
+- T4: kube-apiserver-2 is being restarted
+- T5: kube-apiserver-2 is initialized with watchcache for X synced to
+   rv=102 (most recent etcd version)
+
+As long as no object of type X is being touched (created, updated or
+deleted), watchcache for X will not be updates. So even though they
+contain the same set of objects, one claims to be synced to rv=101
+and the other claims to be synced to rv=102.
+This may result in resuming watches across api-servers to suffer from
```
deads2k, line 157: This appears to be the meat. Bold it perhaps?  Getting stuck in the state is unfortunate.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+"too old resource version" errors in a steady state.
+
+From the first glance it may look like having many kube-apiservers can
+mitigate this problem. However, if no object of type X is being changed
+for extended period of time, this doesn't help and I've seen multiple
+large production clusters with tens of thousands of pods for which any
```
deads2k, line 163: I can't parse this sentence.  missing word maybe?  No pod was changing?

deads2k:
This looks plausible to me, but given past history with the watch cache, I'd like to have a solid way to test it.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -0,0 +1,502 @@
...
+  the resource version to which watch cache is synced from then on
+- if watch cannot be reestablished, all watchers connected to it are
+  disconnected, a new quorum list is send and watchcache is reset to
+  that point in time
+
+It's important to note that while only objects of a given resource type
+are watched (thus resource version to which watch cache is synced can
+only be updated to a value that some object of that type had), the quorum
+list, even though it is only requesting objects of that resource type,
+returns "globally current version for the whole etcd".
```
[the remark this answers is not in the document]
lavalamp: Yeah, this condition is per-etcd backend.

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
lavalamp: The frequency change requires etcd work? Or is it just a setting when you start a watch?

I'm uncomfortable with the solution being strongly time dependent. I currently still think I'd like apiserver to pre-load enough items into the cache to make the chance of a race extremely low. If we can make the "race window" 10s instead of 250ms, I think I'm fine with that.

wojtek-t: > The frequency change requires etcd work? Or is it just a setting when you start a watch?

Yes - currently progress notify is hardcoded to be sent every 10m.

>  If we can make the "race window" 10s instead of 250ms, I think I'm fine with that.

I already wrote above that 250ms doesn't seem to be needed. I was thinking about something between 1s and 5s. But maybe 10s is good enough too? What we need is to ensure that in rolling-upgrade scenario, the next-to-be-updated kube-apiserver will get a progress-notify event *after* the previous is initialized.

So 10s sounds fine to me to - given we want to send bookmarks on shutdown we need to be able to catch the signal and propagate it down. So we can also wait for the next progress notify in that scenario.

Actually - the above are important details about how this can/should be tuned. I added it to the proposed solution above (in the Proposal section).

wojtek-t: @lavalamp - hopefully this addresses your concern?:
https://github.com/kubernetes/enhancements/pull/1878/files#diff-d73a72730139197a054ffcd22466d771R243-R254

wojtek-t:
> This looks plausible to me, but given past history with the watch cache, I'd like to have a solid way to test it.

@deads2k - I would like to agree on the overall idea and have the "provisional" KEP merged first. Once we're in agreement on that, I will work on filling in the remaining sections, including "Testing" (to reach the implementable state) one and we can discuss that further then - I agree it's a reasonable requirement, but I'm also fairly sure it's doable.
Does that make sense?

lavalamp:
I think this is sufficiently on the right track to go to provisional ~now.

deads2k:
> I think this is sufficiently on the right track to go to provisional ~now.

I agree. I would like the update/explanation of http connection stickiness

/approve

wojtek-t:
> I agree. I would like the update/explanation of http connection stickiness

Done - PTAL at https://github.com/kubernetes/enhancements/pull/1878/files#diff-d73a72730139197a054ffcd22466d771R256-R268

@deads2k @lavalamp - PTAL

keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
lavalamp: OK, yeah, that should keep it from being racy.

wojtek-t:
Thanks Daniel.
Will work on filling the remaining sections to move it to implementable in the next week or two.

wojtek-t:
The PR filling the remaining sections and marking the KEP as implementatble is out for review: https://github.com/kubernetes/enhancements/pull/1922

PR #1938, opened 2020-08-14
update README.md to add 1.19 enhancements sheet

kikisdeliveryservice:
Adding the 1.19 sheet 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Please do not comment on the enhancement issue to:
 All the enhancements from this repo are visualized in the Enhancements Tracking Spreadsheets.
 
 Links:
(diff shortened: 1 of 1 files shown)


palnabarun:
Thank you @kikisdeliveryservice for adding the current milestone. :smile: :tada: 

/lgtm

justaugustus:
Thanks @kikisdeliveryservice!
Could we also add a note to the Enhancements subteam playbook to do this at the start of the cycle?

/lgtm
/approve

palnabarun:
> Could we also add a note to the Enhancements subteam playbook to do this at the start of the cycle?

@justaugustus Sure. I will keep that in mind while updating the handbook. Thanks for pointing that out. :+1:

PR #1951, opened 2020-08-20
PRR questionaire for SelfLink removal KEP

wojtek-t:
/cc @lavalamp

### DIFF
--- a/keps/sig-api-machinery/1164-remove-selflink/README.md
+++ b/keps/sig-api-machinery/1164-remove-selflink/README.md
@@ -15,6 +15,13 @@
   - [Graduation Criteria](#graduation-criteria)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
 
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/1164-remove-selflink/README.md
```diff
@@ -178,7 +185,130 @@ No specific strategy is required.
...
+resource usage (CPU, RAM, disk, IO, ...) in any components?**
+  No.
+
+### Troubleshooting
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+
+_This section must be completed when targeting beta graduation to a release._
```
lavalamp, line 296: None of the questions cover this, but absence of this could cause hard to figure out problems in workloads that depend on it.

lavalamp:
Will enabling the feature flag cause the schema to change? I'm slightly nervous that removing the field completely in some release will trigger bugs that merely not setting it does not.

keps/sig-api-machinery/1164-remove-selflink/README.md
wojtek-t: > but absence of this could cause 

By "this" you mean "selflink", right?

Have you heard of any workloads using SelfLink? What this can be used for? Some identification for sharding purposes or something like that?

wojtek-t:
> Will enabling the feature flag cause the schema to change? I'm slightly nervous that removing the field completely in some release will trigger bugs that merely not setting it does not.

No - the schema will not change, it will just stop being propagated.
This is what @deads2k suggested before Alpha that we should have Beta when the field will be there but it will not be propagated, to give people change for feedback (and switch it on again before they fix their stuff if needed).

keps/sig-api-machinery/1164-remove-selflink/README.md
```diff
@@ -178,7 +185,130 @@ No specific strategy is required.
...
+* **Are there any tests for feature enablement/disablement?**
+  No.
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  If there is any component relying on the fact that SelfLink field is set,
+  it may stop working as expected.
```
deads2k, line 220: that component may stop working as expected.

keps/sig-api-machinery/1164-remove-selflink/README.md
```diff
@@ -178,7 +185,130 @@ No specific strategy is required.
...
+* **What specific metrics should inform a rollback?**
+  No generic metrics. Health of individual components should be watched.
+  Generic Kubernetes components has been updated to not rely on it.
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  Manual tests were done, SelfLink was/wasn't set as expected.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs,
+fields of API types, flags, etc.?**
+  Yes - SelfLink field in ObjectMetadata is being deprecated and removed.
```
deads2k, line 231: indicate which release this happened in please.

deads2k: > indicate which release this happened in please

nm, kep does that above

PR #1953, opened 2020-08-24
KEP: Enhancing Performance Benchmarking testcases

hanamantagoudvk:


hanamantagoudvk:
CLA is signed now

wojtek-t:
I will take a look in the next couple days.

/assign

/cc @mm4tt

mm4tt:
Thanks for opening this, looks really interesting!
I'm also a bit swamped this week, but adding to my TODO list, will try to take a look in the next couple days.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+networking scenarios :
+
+a)  IPv4 , IPv6 , Dual-Stack.
+b)  Protocols: TCP, UDP, HTTP/HTTPS.
+c)  With services such as ClusterIP , NodePort etc.
+
+We intend to submit tests to measure the following network performance metrics for various 
+scenarios such as different services (ClusterIP/NodePort), with IPv4/IPv6 configured etc.
+
+a)  Throughput (For TCP traffic)
```
aojea, line 72: we should have packet rate too, TCP throughput is "easy to manipulate" to have nice metrics (increase MTU, offloading)

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+b)  Loss rate (For UDP traffic)
+c)  Latency/Jitter (For UDP/HTTP/HTTPS)
+d)  Response time (For HTTP/HTTPs)
```
aojea, line 75: I think that response time is the 2 way latency plus the time to process the request, right?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+a)  IPv4 , IPv6 , Dual-Stack.
+b)  Protocols: TCP, UDP, HTTP/HTTPS.
+c)  With services such as ClusterIP , NodePort etc.
+
+We intend to submit tests to measure the following network performance metrics for various 
+scenarios such as different services (ClusterIP/NodePort), with IPv4/IPv6 configured etc.
+
+a)  Throughput (For TCP traffic)
+b)  Loss rate (For UDP traffic)
+c)  Latency/Jitter (For UDP/HTTP/HTTPS)
```
aojea, line 74: latency as in the time to travel from A -> B or going and coming back from A <-> B ?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+
+The testcase will cover different protocols such as TCP, UDP, HTTP/HTTPS with IPv4/IPv6 addresses configured.
+
+The testcases will cover different client-to-server pod number ratio (i.e 1:N , 1:1 , N:1)
+
+### Tools
+
+We intend to use the following tools for testing in addition to iPerf/netperf already being used
+
+a)  T-Rex: T-Rex is opensource traffic generator tool made available by CISCO.
```
aojea, line 106: we need to be aware of the limitations https://trex-tgn.cisco.com/trex/doc/trex_faq.html#_which_nics_are_supported_why_not_all_dpdk_supported_nics_supported_by_trex

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+## Proposal
+
+### Test Scenarios
+
+As captured in the link https://docs.google.com/document/d/1_760ptAHBY_c56slWhjl6PJQOrf__vQFJXkL58ewg1M/edit
+we propose to have 2 broad categories based on how client PODs can communicate with server PODs.
+
+The testcase will cover different protocols such as TCP, UDP, HTTP/HTTPS with IPv4/IPv6 addresses configured.
+
+The testcases will cover different client-to-server pod number ratio (i.e 1:N , 1:1 , N:1)
```
aojea, line 100: are we going to build a matrix, is N a variable? 
also, we should consider if the pods are in the same node or not, that will impact highly the results (pod to pod in the same node )

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Tools
+
+We intend to use the following tools for testing in addition to iPerf/netperf already being used
+
+a)  T-Rex: T-Rex is opensource traffic generator tool made available by CISCO.
+    It will be used for measuring the UDP maximum throughput (Kpps), RTT latency (min,max,avg).
```
aojea, line 107: you should expand on this, is Trex going to be deployed as a pod? Stateful or stateless mode?
If you only use one Trex pod, how do you "replay back" the generated traffic?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+b)  Siege: Siege is an open source benchmark utility. 
+    It will be used for measuring HTTP/HTTPs metrics such as total number of hits recorded, 
+    bytes transferred, response time, concurrency, and return status. 
+
+	
+### Execution Environment
+
+We propose to enhance the testcases in either of the 2 test frameworks i.e E2E or netperf.
+
+Our recommendation is to include these testcases in kubernetes CI framework,so that these
```
aojea, line 118: what requirements do you need? number of nodes, VMs resources, ... 
I know that Trex works fine for AWS, how about GCE, are there any NIC/network limitations?
Any specific CNI plugin to test with?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+testcases are run periodically (once in a week). 
+
+### Future Enhancements
+
+Performance benchmarking frameworks (E2E or netperf) can be enhanced to include testcases 
+related to various CNIs.
+
+### Benefits
+
+1) Our proposal attempts to standardize the test scenarios and cover most common performance metrics.
```
aojea, line 128: That will be awesome, for that we should define the Test cases with the output metrics first

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes we can have TCP throughput measurement for various MTU values. Already netperf framework has such testcase for different MSS size.

hanamantagoudvk: Packet Rate (packet per second) itself is not provided by tools like iperf. So we have to calculate ourselves.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes true.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes N is variable. We can define the possible values like 5/10/20 etc.
But we can limit it to few values , otherwise the testsuite runtime will increase.

Yes , forgot to capture this important point of client-server pod co-located on same node or remotely located.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+## Motivation
+
+As Kubernetes is increasingly being adopted by various telco cloud providers for deployments, 
+it brings additional networking requirements/scenarios to be supported.  
+With wide array of network functions being containerized, the need to analyse and 
+monitor the networking performance metrics such as latency, throughput etc has arisen.
+
+We have come across 2 test frameworks which have limited testcases for network performance benchmarking.
+
+a) E2E (test\e2e\network\networking_perf.go): This framework has only 2 testcases for measuring throughput.
```
mm4tt, line 54: There is also clusterloader2 which does test network performance in some way. E.g. we have probes testing NetworkLatency (data plane) - [code](https://github.com/kubernetes/perf-tests/blob/b1fe05e3f799545f79daf6e44a331b5807ec1305/clusterloader2/pkg/measurement/common/probes/probes.go). It's very limited at the moment but could be easily extended to achieve more. The advantage is that it already runs continuously at 100 and 5000 node scale

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Non-goals
+
+Benchmarking and measurements with intermediate services like ingress controllers, load-balancers 
+are out of scope of this proposal.
+
+### Challenges and Open Questions [optional]
+
+1) Which is the suitable hosting framework for proposed tests ? 
+2) Periodic execution of these proposed tests ?
```
mm4tt, line 88: They should be set up as any other K8s CI/CD job, i.e. using prow. SIG Scalability can provide resources for running these tests. See examples of scalability prow jobs in https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/sig-scalability

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,152 @@
...
+    It will be used for measuring HTTP/HTTPs metrics such as total number of hits recorded, 
+    bytes transferred, response time, concurrency, and return status. 
+
+	
+### Execution Environment
+
+We propose to enhance the testcases in either of the 2 test frameworks i.e E2E or netperf.
+
+Our recommendation is to include these testcases in kubernetes CI framework,so that these
+testcases are run periodically (once in a week).
```
mm4tt, line 119: Depending on the cost of those tests, we can run them more often. To estimate the cost (for e2e tests) we'll need to know the size of the cluster the tests will be running in and the time that the tests will take.

mm4tt (commented):
I'm supportive of this proposal. Sig scalability can provide resources needed to run this tests continuously, but I'd like to get some data on how much it's going to cost (# vCPUs and time to run a single test suite)

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: In our simulated setup, we have tried with Trex client pod (in stateless mode) and server pod with OVS installed.
But to replay the received packets at server pod , we had to use 2 vNICs. Hence both client and server pods had to use 2 vNICs. This forced us to use multus CNI. This could become a dis-advantage if i can't use secondary network CNI. 

iperf2 also gives latency (end-to-end) , but with Trex we could measure NDR/PDR (https://trex-tgn.cisco.com/trex/doc/trex_ndr_bench_doc.html).

So we need to decide here.

aojea: yeah, I have used Trex quite a bit in the past, it's really a great and powerful tools and I love to be used here.

For my Kubernetes testing with Trex I had to use it in Advanced Stateless mode  (something similar to this setup https://trex-tgn.cisco.com/trex/doc/trex_astf_vs_nginx.html#_overview ) If we do we want to do real data plane performance testing Trex is the tool, data plane performance has to be measure in packets per second to have a good understanding on how many 

iperf and others userspace tools, are not real dataplane performance, because they are constrained by the system (you can only generate the traffic that the pod is able to generate) and does not give you an idea con how many packet per second is able to move the system, because offloading and other mechanism optimize it

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: Just for taking it into account, you'll need this data also for asking for resources :)

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes once we finalize the test framework, we can implement the various testcases and see the runtime. After that, we can decide whether to include all testcases or selective ones.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Ok will take a look at it.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes we can enhance testcases in clusterloader2 as well. Can't we integrate netperf with clusterloader2 ? I thought we could reuse the code , so with lesser effort we could achieve this.

Need your help in finalizing the test framework.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: I would like to differentiate between packet rate and throughput, is common to use it interchangeable in "marketing" papers, but for performance testing we need to be specific. San we add another point with this metric?
```
b) Packet per seconds
```
One interesting thing you use to see in Linux systems is that you can have maximum throughput only using offloading and jumbo from, because you've reached the packet per second  limit and the only way to get the maximum is using bigger packets

image

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: we should define clearly what latency are we measuring, one way or two way

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Regarding requirements ,  
a) Worker nodes : The worker nodes to run the client-server pods. The number might vary based on the testcases and number of pods we want to run on them.
b) vCPUs : I believe , here the worker nodes are run on the VMs. So these VMs should have enough vCPUs assigned , so that the tools we run inside the POD can make use these vCPUs. We can pin the vCPUs to the tools , so that they run efficienlty.

Regarding the Trex, i am not sure of its limitations on various provider cloud platforms like AWS , GCE. 
I have used Trex only in my local k8s cluster.

Regarding CNIs, we plan to have generic test cases written which are agnostic to CNI.
CNI configuration should be preparatory step.
But initially i would like to test with Calico , Multus CNI.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes i agree with you regarding the iperf. But Trex brings the need for 2 vNICs. How do we deal with that ? If we can use multus CNI for such testcases, then we can use Trex. Is it possible to manage with just 1 vNIC for Trex ?

aojea: > Is it possible to manage with just 1 vNIC for Trex ?

it is possible https://trex-tgn.cisco.com/trex/doc/trex_manual.html#_dummy_ports

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: can we make it clear in the doc? is a bit picky, but once we start to report metrics it can be misleading and prone to different interpretations

hanamantagoudvk: Yes i will make changes to the document.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: @mm4tt are any "networking specifics" in the cluster used by the tests? I mean, special network drivers and these kind of things. If you have run any dpdk test before is a good thing, because this will assure Trex compatibiliy.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,188 @@
...
+
+c) Client-Server POD ratio (1:1 , N:1 , N:M)
+
+### Tools
+
+We intend to use the following tools for testing :
+
+a) iperf3: For TCP throughput measurement.
+
+b) iperf2: For UDP latency, loss rate , pps measurement.
```
aojea, line 126: is it possible to just use only one to avoid confusions?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,188 @@
...
+1) Which is the suitable hosting framework for proposed tests ? 
+2) Periodic execution of these proposed tests ?
+
+
+## Proposal
+
+### Test Scenarios
+
+Detailed test scenarios are listed in below link
+https://docs.google.com/document/d/1_760ptAHBY_c56slWhjl6PJQOrf__vQFJXkL58ewg1M/edit?usp=sharing
```
aojea, line 110: we should incorporate them to the KEP before merging for future references

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,188 @@
...
+
+Test Scenarios capture following aspects :
+
+a) Point-to-Point(P2P), Services.
+
+b) Client-Server POD location on worker node (co-located , distributed).
+
+c) Client-Server POD ratio (1:1 , N:1 , N:M)
```
aojea, line 118: I think that you can not load balance between iperf servers, 1:M and N:M client->ClusterIP should be only possible to test with siege and , per example, nginx/httpd pods as service backends.
I think that the option of using stickiness for ClusterIPs should give us the same results of 1:1, and no real value.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,188 @@
...
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-goals](#non-goals)
+  - [Challenges and Open Questions [optional]](#challenges-and-open-questions-optional)
+- [Proposal](#proposal)
```
aojea, line 28: We should add a section on reporting, how are the results going to be stored and how can we compare if we regressed or improved?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: iperf3 is a bandwidth tester tool (https://github.com/esnet/iperf/issues/547).
where as , iperf2 provides in enhanced option to measure latency (https://www.reddit.com/r/networking/comments/8j3rky/iperf_enhanced_reports_for_udp_latency/). 

Hence we can't use the same tool for both.

aojea: I wanted to mean that iperf2 measured bandwidth too, and maybe it would be simpler using one tool only for all tests.
Anyway, is up to you, was mostly curiosity

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: Yes i have modified the list of testcases accordingly.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @mm4tt  are calico and multus CNIs available for us to run these testcases ?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @mm4tt we have updated this spec with HLD(High Level Design) assuming we are going with clusterloader framework. Could you please review it ?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @aojea We have captured reporting section in HLD. Please review it.

hanamantagoudvk:
@mm4tt @wojtek-t Kindly request your review cycles on this PR.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
wojtek-t: Yes - please copy whatever is relevant to this KEP

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
wojtek-t: I think I'm actually against extending k8s e2e tests - I would prefer to use either of the two other existing ones. We're generally going into direction to not use them for performance/scalability related testing.

Re networking specific - both are about tests and are independent of how the cluster is configured. You should be able to configure your clusters as you wish independently from the test you want to run on it.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,196 @@
...
+c) clusterloader2
+
+Our recommendation is to include these testcases in kubernetes CI framework,so that these
+testcases are run periodically (once in a week). 
+
+### High Level Design 
+
+We have captured the high level design for the clusterloader2 framework in the below document :
+
+https://docs.google.com/document/d/1FhLsE7I1G14jQm7xefIUfrJBLTjQeLx809NpiXgv4lE/edit#heading=h.zetv08f6zre1
```
wojtek-t, line 153: Let's not link docs - please paste all the relevant information into the KEP.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,196 @@
...
+   
+c) clusterloader2 (https://github.com/kubernetes/perf-tests/tree/master/clusterloader2) : This framework 
+   measures only network latency. Other performance metrics are not measured.
+
+### Goals
+
+The primary goal of this proposal is to enhance the testcases to cover various 
+networking scenarios :
+
+a) IPv4 , IPv6 , Dual-Stack.
```
wojtek-t, line 72: This isn't about testcases - this is how the cluster if configured.

Can you please make it more explicit of:
- what cluster configuration/setup options you would like to test
- what specific tests you would like to run against those.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,196 @@
...
+scenarios such as different services (ClusterIP/NodePort), with IPv4/IPv6 configured etc.
+
+a)  Throughput (For TCP traffic)
+
+b)  Packets per second (For UDP traffic)
+
+c)  Loss rate (For UDP traffic)
+
+d)  Latency/Jitter (For UDP/HTTP/HTTPS)
+    (Note: Latency being measured is a one-way latency)
```
wojtek-t, line 88: How you're going to measure this one? How you will guarantee no clock skew across machines (if there is clock skew you can't really measure it, right?)

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
```diff
@@ -0,0 +1,196 @@
+---
```
wojtek-t, line 1: Please adjust to the new KEP template (kep.yaml and README.md).

wojtek-t (commented):
Please don't link docs - the KEP should be self-contained. Once this is updated I will take another pass.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @wojtek-t ,  yes i understand that if there is clock skew across machines , latency metric value might be inaccurate. 
Want to know how GCP on which these scalability tests, ensures the time synchronizations ?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @wojtek-t , do we have any CNI specific testcaes in sig-scalability or its not in the scope of sig-scalability? 
As per my knowledge of scalability tests, they are run on GCP cluster. Which CNI is the default one on GCP ?

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
hanamantagoudvk: @wojtek-t @aojea : I have updated the new KEP template. I have taken care most of the review comments. Please review.

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases.md
aojea: I would postpone ip family craziness :smile: and work with IPv4 for now

keps/sig-scalability/0001-enhance-network-performance-benchmark-testcases/README.md
```diff
@@ -0,0 +1,887 @@
...
+
+2. iperf2: For UDP latency, loss rate , pps measurement.
+
+3. Siege: Siege is an open source benchmark utility. 
+   It will be used for measuring HTTP/HTTPs metrics such as total number of hits recorded, 
+   bytes transferred, response time, concurrency, and return status. 
+
+Other tools such as Trex will be used in future for further enhancements.
+
+### Resource Requirements
```
hanamantagoudvk, line 360: @wojtek-t @mm4tt  , have one query. Is VM(worker node) over subscription , a possibility on  compute servers in google-cloud?

wojtek-t:
@hanamantagoudvk - thanks a lot for extending the original KEP and evolving it.

With the evolvements on the KEP and the fact that it will involve only changes in perf-tests repo (and the fact that it won't require any change to the Kubernetes itself), I suggest:
- closing this KEP (and corresponding issue)
- opening a PR (you can simply copy the main part of this proposal) against clusterloader2 repo https://github.com/kubernetes/perf-tests/tree/master/clusterloader2/docs

We are changing the test frameworks kind-of independently of the release cycle so I think it's unnecessary overhead to track it here.

hanamantagoudvk:
Closing this PR as this enhancement is not related to core k8s modules , but instead clusterloader2 test framework . Hence raised another PR on clusterloader2 framework :
https://github.com/kubernetes/perf-tests/pull/1498

PR #1957, opened 2020-08-28
Use a regexp instead of strings for SIGs in query

johnbelamaric:
This allows passing one or more regexp instead of strings to the `kepctl query` command. This makes discoverability way better. For example, with this and the PRR PR I have out, I could use `kepctl query --sig '.*' --prr '@johnbelamaric' --include-prs --status provisional` to get a list of all Provisional KEPs for which I am assigned as a PRR reviewer.

However, it means the CLI tool only will work if you are online, since the SIG list is loaded in the package `init` of the validations module. Maybe we should embed the list as of the build time?

### DIFF
--- a/pkg/kepctl/query.go
+++ b/pkg/kepctl/query.go
@@ -18,11 +18,12 @@ package kepctl
 
 import (
 	"fmt"
-	"strings"
+	"regexp"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


logicalhan:
> Maybe we should embed the list as of the build time? 

I like the idea, it shouldn't really change super often.

PR #1964, opened 2020-09-02
fix broken link:https://github.com/kubernetes/kubernetes/blob/master/…

CriaHu:
…pkg/kubelet/apis/cri/runtime/v1alpha2/api.proto#L734


ehashman:
This link doesn't resolve. I'm going to close this.

PR #1985, opened 2020-09-14
Removing dockershim from kubelet

dims:
Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/keps/sig-node/1985-remove-dockershim/README.md
+++ b/keps/sig-node/1985-remove-dockershim/README.md
@@ -1,0 +1,302 @@
+# KEP-1985: Removing dockershim from kubelet
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Terms](#terms)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Pros](#pros)
+  - [Cons](#cons)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Dockershim removal criteria](#dockershim-removal-criteria)
+  - [Dockershim removal plan](#dockershim-removal-plan)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Terms
+
+- **CRI:** Container Runtime Interface – a plugin interface which enables kubelet to use a wide variety of container 
+runtimes, without the need to recompile.
+
+## Summary
+
+CRI for docker (i.e. dockershim) is currently a built-in container runtime in kubelet code base. This proposal aims 
+at a deprecation and subsequent removal of dockershim from kubelet.
+
+## Motivation
+
+In Kubernetes, the CRI interface is used to talk to a container runtime, The design of CRI is to be able to run a CRI
+implementation as a separate binary. However currently the CRI of docker (a.k.a. dockershim) is part of kubelet code, runs
+as part of kubelet and is tightly coupled with kubelet's lifecycle. 
+
+This is not ideal as kubelet then has dependency on specific container runtime which leads to maintenance burden for not 
+only developers in sig-node, but also cluster administrators when critical issues (e.g. runc CVE) happen to container 
+runtimes. The pros of removing dockershim is straightforward:
+
+### Pros
+- Docker is not special and should be just a CRI implementation just like every other CRI implementation in our ecosystem. 
+- Currently, dockershim "enjoys" some inconsistent integrations for various reasons (see [legacyLogProvider](https://cs.k8s.io/?q=legacyLogProvider&i=nope&files=&repos=kubernetes/kubernetes) for example) . Removing these "features" should eliminate maintenance burden of kubelet.
+- A cri-dockerd can be maintained independently by folks who are interested in keeping this functionality
+- Over time we can remove vendored docker dependencies in kubelet.
+- Due to convenience of inheriting from this builtin shim for the container runtime, there is less incentive to move to 
+  new container runtimes. The production issues found and addressed in both Containerd and CRI-O might cause the 
+  production issues for some users, which causes a lot of maintenance burdens to Kubernetes community.
+- The community can focus and move faster on the new container runtime-related enhancements once we drop dockershim
+
+Having said that, cons of removal built-in dockershim requires lots of attention:
+
+### Cons
+- Deployment pain with a new binary in addition to kubelet.
+  - An additional component may aggravate the complexity currently. It may be relieved with docker version evolutions.
+- The number of affected users may be large.
+  - Users must change existing use experience when using Kubernetes and docker.
+  - Users have to change their existing workflows to adapt to this new changes.
+  - And other unrecorded stuff.
+- CRI is still technically in alpha, and should graduate to GA before removing dockershim from kubelet. There 
+is a [KEP 2041](https://github.com/kubernetes/enhancements/pull/2041) for graduating the CRI API version.
+- cri-dockerd will vendor kubernetes/kubernetes, that may be tough.
+- cri-dockerd as an independent software running on node should be allocated enough resource to guarantee its availability.
+
+> You can check [the discussion in sig-node mailing list](https://groups.google.com/forum/#!msg/kubernetes-sig-node/0qVzfugYhro/l6Au216XAgAJ) for more details. 
+
+### Goals
+
+- A concrete dockershim removal criteria.
+- A brief plan to remove dockershim spanning multiple releases.
+
+### Non-Goals
+
+- Refactoring or re-design of dockershim itself due to deprecation.
+
+## Proposal
+
+### Dockershim removal criteria
+
+- CRI itself is alpha. So we need another KEP to graduate CRI API.
+- kubelet has no dependency on dockershim/docker in its whole lifecycle. This is already done using the `dockerless` tag
+- All node related features are CRI generic and have no "back door" dependency on dockershim/docker.
+- Deprecate and remove, or replace all Docker-specific features.
+- Reasonable benchmark result of performance degradation after moving dockershim to out-of-tree.
+- E2E test framework has been updated with fully support of out-of-tree CRI container runtime.
+
+### Dockershim removal plan
+
+Step 1: Deprecate in-tree dockershim and decouple dockershim from kubelet.
+
+Target releases: 1.20, 1.21
+
+Actions:
+
+- Mark in-tree dockershim as "maintenance mode":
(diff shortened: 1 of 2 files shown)


keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+### Graduation Criteria
+
+- All feedback gathered from users
+- Adequate test signal quality for node e2e
+- Tests are in Testgrid and linked in KEP
+- Allowing time for additional user feedback and bug reports
+
+### Upgrade / Downgrade Strategy
+
+Not applicable.
```
neolit123, line 175: maybe we should mention this here:

```suggestion
Upgrade: Users should follow the migration guide before upgrading to a version of the kubelet that no longer includes dockershim.

Downgrade: Not applicable.
```

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [ ] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: SizeMemoryBackedVolumes
```
odinuge, line 189: This part is probably not relevant?

odinuge (commented):
Thanks for working on this @dims (and @mattjmcnaughton)! I agree on the overall idea as discussed previously, and as a sig-node contributor it would also make "my life" a lot easier. It would also help making the codebase cleaner and easier to work with. :smile: 

Still need to read through and think about the details, but overall this looks good!

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+runtimes, without the need to recompile.
+
+## Summary
+
+CRI for docker (i.e. dockershim) is currently a built-in container runtime in kubelet code base. This proposal aims 
+at a deprecation and subsequent removal of dockershim from kubelet.
+
+## Motivation
+
+In Kubernetes, CRI is the used as the "default" container runtime, while currently the CRI of docker (a.k.a. dockershim)
```
liggitt, line 60: typo? `CRI is the used as the "default" container runtime`

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+In Kubernetes, CRI is the used as the "default" container runtime, while currently the CRI of docker (a.k.a. dockershim) 
+is part of kubelet code and coupled with kubelet's lifecycle. 
+
+This is not ideal as kubelet then has dependency on specific container runtime which leads to maintenance burden for not 
+only developers in sig-node, but also cluster administrators when critical issues (e.g. runc CVE) happen to container runtimes. The pros of removing dockershim is straightforward:
+
+### Pros
+- Docker is not special and should be just a CRI implementation just like every other CRI implementation in our ecosystem. 
+- Currently, dockershim "enjoys" some backdoors for various reasons (see [legacyLogProvider](https://cs.k8s.io/?q=legacyLogProvider&i=nope&files=&repos=kubernetes/kubernetes) for example) . Removing these "features" should eliminate maintenance burden of kubelet.
```
liggitt, line 68: "backdoor" suggests security issues. possibly a term like "special privileges" or "inconsistent integrations"?

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+- A cri-dockerd can be maintained independently by folks who are interested in keeping this functionality
+- Over time we can remove vendored docker dependencies in kubelet.
+
+Having said that, cons of removal built-in dockershim requires lots of attention:
+
+### Cons
+- Deployment pain with a new binary in addition to kubelet.
+  - An additional component may aggravate the complexity currently. It may be relieved with docker version evolutions.
+- The number of affected users maybe large.
```
liggitt, line 77: typo: may be

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+  - Users must change existing use experience when using Kubernetes and docker.
+  - Users have to change their existing workflows to adapt to this new changes.
+  - And other unrecorded stuff.
+- Updating all the eco-system tools to avoid dependencies on docker.
+- Many people use the built in dockershim for in-cluster image build. While that may not be something we recommend for 
+a variety of reasons, it will be a breaking change for these users.
+- CRI is still in alpha，should probably get a 1.0 out there splitting out dockershim completely from kubelet. There
```
liggitt, line 84: suggest strengthening this language:
```suggestion
- CRI is still technically in alpha, and should graduate to GA before removing dockershim from kubelet.
```

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+- The number of affected users maybe large.
+  - Users must change existing use experience when using Kubernetes and docker.
+  - Users have to change their existing workflows to adapt to this new changes.
+  - And other unrecorded stuff.
+- Updating all the eco-system tools to avoid dependencies on docker.
+- Many people use the built in dockershim for in-cluster image build. While that may not be something we recommend for 
+a variety of reasons, it will be a breaking change for these users.
+- CRI is still in alpha，should probably get a 1.0 out there splitting out dockershim completely from kubelet. There 
+needs to be another KEP for graduating the CRI API version.
+- Existing CNI and CSI plugins may also be affected.
```
liggitt, line 86: this surprised me... are there more concrete details you can point to for these?

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+- Mark in-tree dockershim as "maintenance mode":
+  - CRI generic changes/features can continue on dockershim.
+  - WIP efforts on dockershim can continue and go to complete.
+  - dockershim/docker specific changes/features should be rejected.
+- Deprecate the legacy features of dockershim in kubelet by providing a specific timeline. Currently, kubelet still has:
+  - vendored dockershim 
+  - flags that are used to configure dockershim.
+  - support to get container logs when docker uses journald as the driver.
+  - logic of moving docker processes to a given cgroup
+  - TBD anything else?
```
liggitt, line 129: note: complete before merge

keps/sig-node/1985-remove-dockershim/README.md
danwinship: The only CNI-related CRI-vs-dockershim difference that I know of is that CRI handles pod HostPorts itself, while dockershim expects someone else to do it. (Either the plugin does it itself or else the administrator configures a chain of plugins including `portmap`.) So a plugin that previously only supported dockershim might need to be updated to also support CRI correctly (eg, https://github.com/openshift/origin/pull/16463 though that PR was slightly wrong in that it assumed the difference was CRI-O-specific, not CRI in general). But all those issues probably got worked out years ago...

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+Step 2: Release kubelet without dockershim
+
+Target releases: 1.22
+
+Actions:
+- Document and announce migration guide.
+- Release harness would build kubelet with `dockerless` tag on. So the default build will not support docker out of 
+the box. 
+- If folks need this support, they would have to build kublet by themselves as the code is still present in the
+source tree.
```
liggitt, line 145: at first glance, this doesn't match the spirit of the deprecation policy, which is intended to avoid requiring consumers to change the way they consume/run kubernetes until the deprecation period ends

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+Step 2: Release kubelet without dockershim
+
+Target releases: 1.22
+
+Actions:
+- Document and announce migration guide.
+- Release harness would build kubelet with `dockerless` tag on. So the default build will not support docker out of 
+the box. 
+- If folks need this support, they would have to build kublet by themselves as the code is still present in the
```
mauriciovasquezbernal, line 144:
```suggestion
- If folks need this support, they would have to build kubelet by themselves as the code is still present in the
```

mattjmcnaughton:
Thanks for your thought and leadership on this project @dims and @derekwaynecarr :)

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+### Cons
+- Deployment pain with a new binary in addition to kubelet.
+  - An additional component may aggravate the complexity currently. It may be relieved with docker version evolutions.
+- The number of affected users maybe large.
+  - Users must change existing use experience when using Kubernetes and docker.
+  - Users have to change their existing workflows to adapt to this new changes.
+  - And other unrecorded stuff.
+- Updating all the eco-system tools to avoid dependencies on docker.
+- Many people use the built in dockershim for in-cluster image build. While that may not be something we recommend for 
+a variety of reasons, it will be a breaking change for these users.
```
neolit123, line 83: should we recommend to those users to move to:
https://github.com/moby/buildkit
https://docs.docker.com/develop/develop-images/build_enhancements/
> Limitations
> Only supported for building Linux containers

BenTheElder: They're not using **dockershim** for build. They're using dockerd directly. They could continue to run dockerd and use it outside of kubernetes. This usage isn't really relevant..?

keps/sig-node/1985-remove-dockershim/README.md
dims: Fixed

keps/sig-node/1985-remove-dockershim/README.md
dims: Done. picked `inconsistent integrations`

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+Having said that, cons of removal built-in dockershim requires lots of attention:
+
+### Cons
+- Deployment pain with a new binary in addition to kubelet.
+  - An additional component may aggravate the complexity currently. It may be relieved with docker version evolutions.
+- The number of affected users maybe large.
+  - Users must change existing use experience when using Kubernetes and docker.
+  - Users have to change their existing workflows to adapt to this new changes.
+  - And other unrecorded stuff.
+- Updating all the eco-system tools to avoid dependencies on docker.
```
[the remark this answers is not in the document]
dims: Dropping this :)

keps/sig-node/1985-remove-dockershim/README.md
dims: Dropping this too

keps/sig-node/1985-remove-dockershim/README.md
dims: Done

keps/sig-node/1985-remove-dockershim/README.md
dims: Dropping this.

keps/sig-node/1985-remove-dockershim/README.md
dims: Dropping this.

keps/sig-node/1985-remove-dockershim/README.md
dims: Done

keps/sig-node/1985-remove-dockershim/README.md
dims: Let's tweak this as we go forward as the KEP is a living document.

keps/sig-node/1985-remove-dockershim/README.md
dims: Updated

keps/sig-node/1985-remove-dockershim/README.md
dims: cleaned up

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+This is not ideal as kubelet then has dependency on specific container runtime which leads to maintenance burden for not 
+only developers in sig-node, but also cluster administrators when critical issues (e.g. runc CVE) happen to container 
+runtimes. The pros of removing dockershim is straightforward:
+
+### Pros
+- Docker is not special and should be just a CRI implementation just like every other CRI implementation in our ecosystem. 
+- Currently, dockershim "enjoys" some inconsistent integrations for various reasons (see [legacyLogProvider](https://cs.k8s.io/?q=legacyLogProvider&i=nope&files=&repos=kubernetes/kubernetes) for example) . Removing these "features" should eliminate maintenance burden of kubelet.
+- A cri-dockerd can be maintained independently by folks who are interested in keeping this functionality
+- Over time we can remove vendored docker dependencies in kubelet.
```
dchen1107, line 72: Another two pros in my mind:

1) Due to convenience inheriting from this builtin shim for the container runtime, there is less incentive to move to new container runtime. The production issues found and addressed in both Containerd and CRI-O might cause the production issues for some users, which causes the maintenance burden to Kubernetes community.
2) The community can focus on the new container runtimes enhancements.

dims: Included these 2 suggestions. thanks @dchen1107

keps/sig-node/1985-remove-dockershim/README.md
```diff
@@ -0,0 +1,302 @@
...
+  - Ensure cluster/node e2e are 100% CRI focused.
+  - Ensure test-infra install appropriate CRI implementations in e2e machines. 
+- Ensure Windows scenarios that currently depend on docker are fully supported in alternative CRI implementations
+
+Step 2: Release kubelet without dockershim
+
+Target releases: 1.22
+
+Actions:
+- Document and announce migration guide.
```
krmayankk, line 142: is there a tracking issue for this migration guide, i think it would help to include in it what steps do existing users take for moving from docker to containerd running on there hosts ?

SergeyKanzhelev: I started this: https://github.com/kubernetes/website/pull/25787

PR #1995, opened 2020-09-20
KEP-1977: Add KEP for ContainerNotifier

xing-yang:
Enhancement issue: https://github.com/kubernetes/enhancements/issues/1977

### DIFF
--- a/keps/sig-node/1977-container-notifier/README.md
+++ b/keps/sig-node/1977-container-notifier/README.md
@@ -1,0 +1,899 @@
+# KEP-1977: Container notifier
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+    - [Phase 1](#phase-1)
+    - [Phase 2](#phase-2)
+    - [Phase 3](#phase-3)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [API Changes](#api-changes)
+    - [Phase 1 API Changes](#phase-1-api-changes)
+      - [Inline Definition - ContainerNotifier](#inline-definition---containernotifier)
+      - [PodNotification](#podnotification)
+    - [Phase 2 API Additions](#phase-2-api-additions)
+      - [Notification](#notification)
+      - [ContainerNotifierHandler](#containernotifierhandler)
+    - [Phase 3 API Additions](#phase-3-api-additions)
+  - [Kubelet Impact in Phase 2 and Beyond](#kubelet-impact-in-phase-2-and-beyond)
+    - [CRI Changes](#cri-changes)
+- [Implementation Plan](#implementation-plan)
+  - [Phase 1](#phase-1-1)
+  - [Phase 2](#phase-2-1)
+  - [Phase 3](#phase-3-1)
+- [Example Workflows](#example-workflows)
+  - [Quiesce Hooks](#quiesce-hooks)
+  - [Example Workflow with SIGHUP (Phase 2)](#example-workflow-with-sighup-phase-2)
+    - [With Probe (Phase 3)](#with-probe-phase-3)
+  - [Example Workflow to Change Log Verbosity (Phase 2)](#example-workflow-to-change-log-verbosity-phase-2)
+    - [With Probe (Phase 3)](#with-probe-phase-3-1)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Test Plan](#test-plan)
+  - [Unit tests](#unit-tests)
+  - [E2E tests](#e2e-tests)
+- [Graduation Criteria](#graduation-criteria)
+    - [Alpha Graduation](#alpha-graduation)
+  - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+  - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+This KEP proposes to introduce a mechanism to notify a selected set of `Pod`(s) to run pre-specified commands inlined in those `Pod`(s) specification.
+
+This is achieved by:
+* Adding an optional `Notifiers` field in [Container](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core) type to allow users to specify command(s)/signal(s) which can be executed/sent.
+* Introducing a core API object(`PodNotification`) to trigger a specific command/signal specified in a `Pod`.
+
+## Motivation
+
+Being able to take application consistent snapshots/backups is a hard requirement to protect stateful workloads in Kubernetes. It's required for many applications to temporally quiesce themselves before taking snapshots of their PersistentVolume(s) and unquiesce afterwards to ensure application consistency. A mechanism to send a quiesce/unquiesce command to `Pod`(s) is needed.
+
+The first attempt was to introduce an [ExecutionHook](https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/20190120-execution-hook-design.md) CRD which allows arbitrary commands to be executed against containers, and a controller which manages the lifecycle of the custom resources. This approach, though solves the application quiesce/unquiesce use cases, has been considered neither secure nor generic enough. The controller needs to be granted super user privileges on nodes.
+
+Meanwhile, there are other interesting use cases which requires a similar mechanism:
+* Users have been asking a feature to signal `Pod`s in this [issue](https://github.com/kubernetes/kubernetes/issues/24957). For example, sending a signal to a `Pod` to reload configuration, to flush/roll logs, to change log verbosity etc.
+* Mechanism to notify `Pod`s on a to-be-evicted node.
+
+With these new motivations, we are proposing a more generic design in this KEP. As the eventual goal is to have Kubelet executing the commands against `Pod`s, this approach is considered to be more secure as it will not widen the scope to grant super user privileges.
+
+### Goals
+
+#### Phase 1
+
+- Users can specify an optional list of commands in `Pod`s specification at container level.
+- Users or upper level controllers can request to execute commands in `Pod`s by creation of an API object(`PodNotification`).
+- Implement a *trusted* controller which monitors requests to execute commands, notifies the corresponding `Pod`s to run those commands, and updates status of requests corresponding to the execution results.
+
+#### Phase 2
+
+- Introduce an upper level API(`Notification`) which allows users to send `PodNotification` to a set of `Pod`s which can be matched by a label selector.
+- Extend `ContainerNotifier` API in Phase 1 to support sending signals to containers.
+- Move *trusted* controller logic into Kubelet.
+
+#### Phase 3
+
+- Add a `Probe` to verify the results from the commands/signals if needed.
+
+### Non-Goals
+
+- Writing concrete commands, for example quiesce scripts, which could be executed. It is the responsibility of users who set up pod definitions.
+
+## Proposal
+
+In phase 1, propose to introduce several API changes:
+- Adding an optional field `Notifiers` which is a list of `ContainerNotifier`s into `Container`.
+- Adding an inlined type `ContainerNotifierHandler` which defines a command.
+- Adding a core API type `PodNotification` which defines request type to trigger execution of `ContainerNotifier`(s) in a `Pod`.
+- Introduce a new gate `ContainerNotifier` to toggle this feature.
+
+A SINGLE *trusted* controller(Pod Notification Controller) will be implemented to watch `PodNotification` resources, execute the command and update their statuses accordingly.
+
+In phase 2, propose to:
+- Add a core API `Notification` type and a controller which processes `Notification` resources.
+- Add an inline `Pod` definition for signals and allows the API object to send a request to trigger delivering of those signals.
+- Move logic in the Pod Notification controller into Kubelet. Kubelet watches `PodNotification` objects, runs the command and updates statuses of `PodNotification` objects accordingly.
+
+In phase 3, a `Probe` may be added if needed as an inline `Pod` definition to verify the signal is delivered or whether the command is run and results in the desired outcome.
+
+Details of proposed API changes can be found in *API Changes* section  bellow.
+
+### API Changes
+
+#### Phase 1 API Changes
+
+##### Inline Definition - ContainerNotifier
+
+Add []ContainerNotifier to the Container struct:
+
+```
+type Container struct {
+    ......
+    // +optional
+    Lifecycle *Lifecycle `json:"lifecycle,omitempty" protobuf:"bytes,12,opt,name=lifecycle"`
+
+    // Notifiers can be triggered by PodNotification resources.
+    // Each notifier must have a unique name within the Container.
+    // Upon creation of a PodNotification, Container(s) in the specified
+    // Pod with a matching notifier name will be executed.
+    // +optional
+    Notifiers []ContainerNotifier
+    ......
+}
+
+type ContainerNotifier struct {
+    // Name of the ContainerNotifier. Name must be a valid label key. More info:
+    // https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set.
+    // Each ContainerNotifier within a Container must have a unique name.
+    // Names with the prefix `k8s.io/` are reserved for Kubernetes-defined
+    // "well-known" notifiers.
+    // Names with other prefixes such as `example.com/` or unprefixed are custom names.
+    // Immutable.
+    Name string
+
+    // The action to take when a matching PodNotification is observed.
+    // Exactly-once execution of the action is NOT guaranteed, so it's user's
+    // responsibility to make sure the handler is idempotent to get consistent
+    // behavior.
+    // For example, when a quiesce command has been ran on a database, the
+    // database will stay quiesced, and any consecutive execution of the
+    // same quiesce command should not have additional impact.
+    Handler *ContainerNotifierHandler
+
+    // Number of seconds after which the notifier will time out.
+    // Default to 1 second. Minimum value is 1.
+    // +optional
+    TimeoutSeconds int32
+}
+
+// ContainerNotifierHandler specifies an action to take when a ContainerNotifier
+// has been triggered.
+type ContainerNotifierHandler struct {
+    // Exec specifies the action to take.
+    // +optional
+    Exec *ExecAction `json:"exec,omitempty" protobuf:"bytes,1,opt,name=exec"`
+}
+```
+
+##### PodNotification
+
+The PodNotification object is a request for execution a ContainerNotifier in a `Pod`. It will be an in-tree API object in the core API group so that Kubelet can watch and trigger the actions.
+
+All containers in the specified `Pod` which have the requested "ContainerNotifier" defined in their spec will have their corresponding "ContainerNotifierHandler" executed.
+
+Note that there is no guarantee of execution ordering between containers.
+
+```
+type PodNotification struct {
+    metav1.TypeMeta
+    // +optional
+    metav1.ObjectMeta
+    // Spec defines the behavior of a notification.
+    // +optional
+    Spec PodNotificationSpec
+    // Status defines the current state of a notification.
+    // +optional
+    Status PodNotificationStatus
+}
+
+type PodNotificationSpec struct {
+    // The name of the Pod to which notification will be addressed.
+    // Note that a PodNotificaion only notifies a Pod in the same namespace.
+    // The PodNotification fails immediately if the specified Pod does not exist.
+    // Required.
+    // Immutable.
+    PodName string
+
+
+    // The name of the ContainerNotifier to trigger.
+    // The PodNotification completes immediately if no Container in the
+    // Pod has the requested ContainerNotifier defined.
+    // Required.
+    // Immutable.
+    Notifier string
+}
+```
+
+###### PodNotificationStatus
+`PodNotificationStatus` represents the current status of a `PodNotification`. It contains a list of `ContainerNotificationStatus`, a start time stamp suggesting the first time the `PodNotification` has been observed, a completion time stamp suggesting the end of the notification, and a succeeded flag representing the final state of the `PodNotification` upon completion.
+
+A `PodNotification` is considered to be successful(i.e., with succeeded flag set to `true`) only if the notification have been successfully executed in all qualified containers in the specified `Pod`.
+
+Note that the `PodNotification` controller (or Kubelet in phase 2) will NOT retry upon failures/errors returned from `ContainerNotifier`. Top level controllers or users can create another `PodNotification` to do retries.
+
+```
+type PodNotificationStatus struct {
+    // A list of ContainerNotificationStatus, with each item representing the
+    // execution result of the corresponding ContainerNotifier in Container(s)
+    // in the specified Pod.
+    // +optional
+    Containers []ContainerNotificationStatus
+
+    // If set, it suggests the timestamp(UTC) at when the first time the
+    // PodNotification has been acknowledged by the PodNotification Controller.
+    // +optional
+    StartTime *metav1.Time
+
+    // If set, it suggests the timestamp(UTC) at when the PodNotification has
+    // been marked as completed. A PodNotification is completed when the
+    // action has been executed in all eligible Containers in this Pod.
+    // Note that a completed PodNotification does NOT mean the notification
+    // has been executed successfully. Refer to `Succeeded` field for more details.
+    // +optional
+    CompleteTime *metav1.Time
+
+    // State represents the current stage of the PodNotification.
+    // Default to PodNotificationStateNew.
+    State PodNotificationStateType
+
+    // Error represents the failing reason of the PodNotification.
+    // +optional
+    Error *PodNotificationError
+}
+
+// PodNotificationStateType represents the current stage of the PodNotification.
+type PodNotificationStateType string
+const(
+  // PodNotificationStateNew suggests the PodNotification is newly created and
+  // has not been picked up yet.
+  PodNotificationStateNew PodNotificationStateType = "New"
+
+  // PodNotificationStateSucceeded suggests the PodNotidication has been
+  // successfully executed against all qualified Containers.
+  PodNotificationStateSucceeded PodNotificationStateType = "Succeeded"
+
+  // PodNotificationStateFailed suggests the PodNotidication has been executed
+  // however at least one of the qualified Containers returned an error. Details
+  // can be found in 'Containers'.
+  PodNotificationStateFailed PodNotificationStateType = "Failed"
+)
+
+// ContainerNotificationStatus represents the execution state of the handler in
+// the requested ContainerNotifier.
+type ContainerNotificationStatus struct {
+    // The name of the Container in which the handler has been executed.
+    // Required.
+    Name string
+
+    // If set, it suggests the timestamp at when the first time the execution
+    // of the action has started in UTC.
+    // +optional
+    StartTime *metav1.Time
+
+    // If set, it suggests the timestamp at when the first time the execution
+    // of the handler has completed in UTC.
+    // Note that `CompleteTime` does NOT mean the handler has been executed
+    // successfully. Refer to `Succeeded` field for more details.
+    // +optional
+    CompleteTime *metav1.Time
+
+    // Succeeded will be set to true when the handler has been executed
+    // successfully, and it will be set to false otherwise. It will not be
+    // specified if the status of execution is unknown(i.e., in execution).
+    // +optional
+    Succeeded *bool
+
+    // If specified, Error represents the last error encountered when executing
+    // the handler.
+    // +optional
+    Error *PodNotificationError
+}
+
+type PodNotificationError struct {
+    // Type of the error
+    // Required.
+    Type PodNotificationErrorType
+
+    // A human-readable message indicating details why the action failed.
+    Message string
+
+    // It indicates when the error occurred
+    Timestamp metav1.Time
+}
+
+// PodNotificationErrorType represents the type of error occurred in a
+// Notifiication.
+type PodNotificationErrorType string
+
+// More error types could be added, e.g., Forbidden, Unauthorized, etc.
+const (
+    // Pod not found error
+    PodNotificationErrorPodNotFound PodNotificationErrorType = "PodNotFound"
+
+    // The execution of a handler timed out.
+    PodNotificationErrorHandlerTimeout PodNotificationErrorType = "HandlerTimeout"
+)
+```
+
+Creators of `PodNotification` resources are responsible of the full lifecycle of them.
+
+#### Phase 2 API Additions
+
+We propose to further extend the API in two aspects:
+- A `Notification` type which allows users or upper level controllers to specify a label selector to target a set of `Pod`s to notify. This makes the feature much user friendly for use cases like application consistent snapshot, signaling a group of `Pod`s.
+- Enhancement on `ContainerNotifierHandler` to support signal.
+
+##### Notification
+```
+type Notification struct {
+    metav1.TypeMeta
+    // +optional
+    metav1.ObjectMeta
+
+    // Spec specifies the desired state of a Notification.
+    // +optional
+    Spec NotificationSpec
+    // Status indicates the current state of a Notification.
+    // +optional
+    Status NotificationStatus
+}
+
+type NotificationSpec struct {
+    // A label query over Pods under the same namespace of the Notification
+    // to notify.
+    // Note: Only Pods in the ‘Running’ phase are eligible for notification.
+    // Ref: kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+    // Required.
+    Selector metav1.LabelSelector
+
+    // Policy specifies pod selection policy. If not specified, default to
+    // "AllPods".
+    // +optional.
+    Policy *PodSelectionPolicy
+
+    // The name of the ContainerNotifier to trigger.
+    Notifier string
+
+    // Parallelism specifies the maximum number of Pods to notify at any given
+    // time concurrently, which effectively is the same as the number of
+    // uncompleted PodNotifications created from this Notification.
+    // If not specified or set to 0, all eligible Pods will be notified.
+    // If set to a number greater than the number of eligible Pods, all
+    // eligible Pods will be notified concurrently.
+    // Setting this field to a negative number fails the Notification immediately.
+    Parallelism int
+}
+
+// PodSelectionPolicy specifies a policy to scope Pods which are qualified for
+// a Notification.
+type PodSelectionPolicy string
+
+const (
+    // PodSelectionPolicyAllPods means all Pods matches the specified label
+    // selector in Notification will be notified. If this policy is specified,
+    // CompleteTime field of the NotificationStatus will not be set.
+    PodSelectionPolicyAllPods PodSelectionPolicy = "AllPods"
+
+
+    // PodSelectionPolicyPreExistingPods means only those Pods created before
+    // the Notification(i.e., create timestamps smaller than Notification's) are
+    // qualified for notification.
+    // Newly created Pods after the Notification will not be notified.
+    PodSelectionPolicyPreExistingPods PodSelectionPolicy = "PreExistingPods"
+)
+
+// NotificationStatus represents the current state of a Notification
+type NotificationStatus struct {
+    // FailedCount indicates the current number of failed PodNotification
+    // resources created from this Notification.
+    // +optional
+    FailedCount int
+
+    // SucceededCount indicates the current number of succeeded PodNotification
+    // resources created from this Notification.
+    // +optional
+    SucceededCount int
+
+    // If set, it suggests the timestamp(UTC) at when the first time the
+    // Notification has been acknowledged by the Notification Controller.
+    // +optional
+    StartTime *metav1.Time
+
+    // If set, it suggests the timestamp(UTC) at when the Notification has
+    // been marked as completed. A Notification is completed when all
+    // PodNotifications created from this Notification have completed.
+    // Note that this field will never be set if "AllPods" is specified as
+    // pod selection policy.
+    // +optional
+    CompleteTime *metav1.Time
+}
+```
+
+##### ContainerNotifierHandler
+```
+// ContainerNotifierHandler defines a specific action to be executed or a signal which will be delivered.
+// Exactly one of Exec and Signal should be set.
+type ContainerNotifierHandler struct {
+    // Exec specifies the action to take.
+    // +optional
+    Exec *ExecAction `json:"exec,omitempty" protobuf:"bytes,1,opt,name=exec"`
+
+    // Signal specifies a signal to send to the container
+    // +optional
+    // define constants for signals?
+    // validate the signals are valid?  windows?
+++    Signal string
+}
+```
+
+#### Phase 3 API Additions
+
+If needed, we may add a probe to ContainerNotifier in Phase 3 to verify that the ContainerNotifierHandler has delivered the signal or executed the command.
+
+Phase 3 is more loosely decided than phases 1 and 2 at this stage, we will have to revisit the design details.
+
+```
+type ContainerNotifier struct {
+    // Name of the ContainerNotifier specified as a DNS_LABEL. Each
+    // ContainerNotifier within a Container must have a unique name.
+    // Names with the prefix `k8s.io/` are reserved for Kubernetes-defined
+    // "well-known" notifiers.
+    // Names with other prefixes such as `example.com/` are custom names.
+    // Immutable.
+    Name string
+
+    // The action to take when a matching PodNotification is observed.
+    // Exactly-once execution of the action is NOT guaranteed, so it's user's
+    // responsibility to make sure the handler is idempotent to get consistent
+    // behavior.
+    // For example, when a quiesce command has been ran on a database, the
+    // database will stay quiesced, and any consecutive execution of the
+    // same quiesce command should not have additional impact.
+    Handler *ContainerNotifierHandler
+
+    // Number of seconds after which the notifier will time out.
+    // Default to 1 second. Minimum value is 1.
+    // +optional
+    TimeoutSeconds int32
+
++   // Add a Probe in Phase 3. Probe defined in the core API here will be used.
++   Probe *Probe
+}
+```
+
+### Kubelet Impact in Phase 2 and Beyond
+
+When moving the logic from the PodNotification controller to Kubelet, we foresee the following changes to happen in Kubelet:
+
+- Kubelet will watch PodNotification resources for Pods on the node it's running.
+
+- Kubelet will execute command/send signal specified in ContainerNotifier against containers. CRI changes are required to support signals.
+
+- Kubelet will update PodNotification status.(Possible QPS concern when trying to update 100+ of PodNotifications concurrently).
+
+- Kubelet will have some method of measuring the success/failure of a PodNotification.
+* For exec, it will depend on the return value of the ExecInContainer method.
+* For signals, it will depend on the new CRI changes.
+
+Kubelet will not retry if the ContainerNotifier call fails. It is up to upper level controllers or users who have requested the PodNotification to manually retry by creating another PodNotification resource.
+
+#### CRI Changes
+
+To support signals directly, we might need to enhance CRI. One potential approach is to add a new "Signal" interface with an input parameter "signal string". Initially, a predefined set of signals will be supported. i.e., "SIGHUP". The initial set of signals is yet to be defined.
+
+```
+// Runtime is the interface to execute the commands and provide the streams.
+type Runtime interface {
+        Exec(containerID string, cmd []string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
++        Signal(containerID string, signal string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
+
+        Attach(containerID string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
+        PortForward(podSandboxID string, port int32, stream io.ReadWriteCloser) error
+}
+```
+
+To support "SIGHUP" directly, we can translate that into a call to the docker client method `ContainerKill`.
+
+```
+https://github.com/moby/moby/blob/master/client/container_kill.go#L9
+```
+
+We can add a `SignalKill` method in `~/go/src/k8s.io/kubernetes/pkg/kubelet/dockershim/libdocker/kube_docker_client.go` and call docker client `ContainerKill`, similar to how `CreateExec` method calls docker client `ContainerExecCreate`.
+
+## Implementation Plan
+
+### Phase 1
+
+Phase 1 API definition will be added to Kubernetes. Controller logic implementation in this phase will not happen in Kubelet. Instead, it will be implemented in a single *trusted* controller which watches `PodNotification`s and execute the command via exec(with the goal being to move that into Kubelet in the next steps). In this phase, only exec will be implemented.
+
+The PodNotifier controller will be created in a separate repo under Kubernetes. The repo will be sponsored by sig-node: `https://github.com/kubernetes/containernotifier`.
+
+### Phase 2
+
+* Introduce `Notification` API and its corresponding controller.
+* Make CRI changes to support signals.
+* Move PodNotification controller logic to Kubelet.
+
+### Phase 3
+
+* Add Probe if needed.
+
+## Example Workflows
+
+### Quiesce Hooks
+
+For example, there are 3 commands that need to be executed sequentially to quiesce a mysql database before taking a snapshot of its persistent volumes: *lockTables*, *flushDisk*, and *fsfreeze*. After the snapshots have been taken, there are 2 commands to be executed sequentially to unquiesce it: *fsUnfreeze*, *unlockTables*. For simplicity, assume we only need to run *fsfreeze* for one volume and we only need to run each command in one container in one pod. These commands are defined in "Notifiers []ContainerNotifier" inside the Container.
+
+1. lockTables
+Upper level controller creates a `PodNotification` object to request the *lockTables* `ContainerNotifier`.
+PodNotification controller(or Kubelet in phase 2) noticed the creation of the `PodNotification` object. It starts to run the *lockTables* command specified in the *ContainerNotifier* and updates the `ContainerNotificationStatus` to set the StartTimestamp.
+
+When the command finishes successfully, PodNotification controller(or Kubelet in phase 2) sets the Succeed field in ContainerNotificationStatus to True.
+If it fails or times out, PodNotification controller(or Kubelet in phase 2) sets the Succeed field to False.
+
+2. flushDisk
+If the *lockTables* command succeeds, upper level controller will proceed to create another `PodNotification` object to request the *flushDisk* `ContainerNotifier`.
+
+PodNotification controller(or Kubelet in phase 2) starts to run the *flushDisk* command specified in the `ContainerNotifier` and updates the `PodNotificationStatus` correspondingly.
+
+When the *flushDisk* command finishes successfully, PodNotification controller(or Kubelet in phase 2) sets the Succeed field in the `ContainerNotificationStatus` to True.
+If it fails or times out, PodNotification controller(or Kubelet in phase 2) sets the Succeed field to False.
+
+If the *lockTables* command fails, upper level controller will create a new `PodNotification` object to request the *unlockTables* `ContainerNotifier`. It will not proceed to the next PodNotification and the snapshot creation will be marked as failure.
+
+3. fsfreeze
+If the *flushDisk* command succeeds, upper level controller will create a new `PodNotification` object to request the *fsfreeze* `ContainerNotifier`.
+
+PodNotification controller(or Kubelet in phase 2) starts to run the *fsfreeze* command specified in the `ContainerNotifier` and updates the `PodNotificationStatus` correspondingly.
+
+When the *fsfreeze* command finishes successfully, PodNotification controller(or Kubelet in phase 2) sets the Succeed field in the `ContainerNotificationStatus` to True.
+If it fails or times out, PodNotification controller(or Kubelet in phase 2) sets the Succeed field to False.
+
+4. Take snapshot
+If the *fsfreeze* command succeeds, upper level controller will proceed to take a snapshot.
+
+5. fsUnfreeze
+Upper level controller creates a `PodNotification` object to request the *fsUnfreeze* `ContainerNotifier` as long as step 3 has been executed.
+If *fsFreeze* in step 3 is called, *fsUnfreeze* should always be called.
+
+PodNotification controller(or Kubelet in phase 2) starts to run the *fsUnfreeze* command specified in the `ContainerNotifier` and updates the `PodNotificationStatus`correspondingly.
+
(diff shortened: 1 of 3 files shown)


xing-yang:
CC @yuxiangqian

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+            // +optional
+	PodSelector *metav1.LabelSelector
+ }
+```
+
+###### Notification Status
+
+NotificationStatus represents the current state of a Notification. It has a list of PodNotificationStatuses.
+
+A PodNotificationStatus represents the current state of a Notification for a specific pod. A PodNotificationStatus contains ContainerNotificationStatuses of all the containers which have the corresponding ContainerNotifier specified. PodNotificationStatus is created when a Notifier starts to run in a pod.
```
thockin, line 236: This is going to be a problem.  Imagine a notifier that selects 1000 pods.  That's not implausible.  Even 5000 is believable.  That leaves you just a few hundred bytes per pod for status, before it starts failing (too big).  Worse, it will fail in non-deterministic ways in practice.

I think we have a few choices here, and we should seek a pattern precedent (@liggitt for consideration)

1) Notification.status is a "summary".  The real status is stored in another new resource `NotificationResult` which is selected by something like "notification.kubernetes.io/uid=<notification-uid>".  Each NotificationResult holds the status for one single pod-notification.  Need to define the lifecycle management of this so we don't leave them around for too long.

2) Notification.status is a "summary".  The real status is stored in each pod, e.g. in `status.notifications[]` which would be a list of "recent" notifications and the per-container statuses

3) other ideas?

TL;DR: we can't have (practically) unbounded lists in a single resource.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+```
+
+##### Notification API Object
+
+The Notification object is a request for a ContainerNotifier. It will be an in-tree API object in the "node.k8s.io" API group so that Kubelet can monitor and trigger the actions. The external controller or a user creates the Notification object to request a ContainerNotifier.
+
+All containers in a selected pod which have a "ContainerNotifier" defined in their spec with the "ContainerNotifierName" will have their corresponding "ContainerNotifierAction" executed. Moreover, there is no guarantee of ordering on execution.
+
+```
+type Notification struct {
```
thockin, line 204: We need to define what happens in this case:

time 0: pods A, B, C are running, labelled foo=bar
time 1: create Notification selecting foo=bar
time 2: all notifiers in A, B, C are done
time 3: pod D is created, labelled foo=bar
*** Do we run the notifier on D?  If so, for how long does that apply?  If not, how do we prevent it?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+    // Signal specifies a signal to send to the container
+    // +optional
+    // define constants for signals?
+    // validate the signals are valid?  windows?
+++    Signal string
+}
+```
+
+### Phase 3 API Additions
```
thockin, line 337: I think we should say something like "Phase 3 is more loosely decided than phases 1 and 2, and if it is needed, we will have to revisit the design details".  There are just too many variables to see the needs at this point (IMO).

As such, I am not aggressively API reviewing it :)

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+type Runtime interface {
+        Exec(containerID string, cmd []string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
++        Signal(containerID string, signal string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
+
+        Attach(containerID string, in io.Reader, out, err io.WriteCloser, tty bool, resize <-chan remotecommand.TerminalSize) error
+        PortForward(podSandboxID string, port int32, stream io.ReadWriteCloser) error
+}
+```
+
+To support "SIGHUP" directly, we can either convert it to the command "kill -SIGHUP" and send it the same way as other commands, or we can translate that into a call to the docker client method `ContainerKill`.
```
thockin, line 394: you can't assume a container has a `kill` binary to execute

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+
+#### Alpha Graduation
+
+* Feature Flag is present.
+* Container notifier controller is implemented in a sig-node sponsored repo.
+* E2E tests are implemented.
+
+### Alpha -> Beta Graduation
+
+* One release has been waited to allow for feedback from users.
```
thockin, line 542: Are we willing to go Beta with the phase 1 design?  I don't think so, am I wrong?  I think you should talk about phases here.  If we decide we need phase 3, will we go beta without it?  What if we go beta and THEN decide we need phase 3, does that become alpha and we release independently?  Or do we wait for it to "catch up" and GA together.

keps/sig-node/1977-container-notifier/kep.yaml
```diff
@@ -0,0 +1,50 @@
...
+participating-sigs:
+  - sig-storage
+status: implementable
+creation-date: 2020-09-20
+reviewers:
+  - "@thockin"
+  - "@liggit"
+  - "@saad-ali"
+  - "@sjenning"
+approvers:
```
thockin, line 16: That's a lot of approvers - do we need them all?  I feel like Jordan OR me, Dawn OR Derek..

xing-yang: Sure, we'll trim down the approvers list.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,770 @@
...
+           // This contains the name of the ContainerNotifier to send to a container.
+           // Name must be unique within a container.
+           // Names are label-style, with an optional prefix. Names without the prefix
+           // such as `quiesce` are reserved for Kubernetes-defined "well-known" names.
+           // Names with a prefix such as `example.com/label-style` are custom names.
+           ContainerNotifierName string
+
+            // PodSelector specifies a label query over a set of pods.
+            // +optional
+	PodSelector *metav1.LabelSelector
```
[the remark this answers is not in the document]
xing-yang: I assume you meant list "PodSelector" as the first parameter and "ContainerNotifierName" second in NotificationSpec?

keps/sig-node/1977-container-notifier/README.md
xing-yang: Ok

keps/sig-node/1977-container-notifier/README.md
xing-yang: Removed wording using the command.

keps/sig-node/1977-container-notifier/README.md
xing-yang: We'll need phase 2 before going Beta.  Regarding phase 3, we should discuss and make a decision after phase 2 is done?

keps/sig-node/1977-container-notifier/README.md
xing-yang: Ok, I'll wait for more suggestions.

keps/sig-node/1977-container-notifier/README.md
xing-yang: We have said that Kubelet (container notifier controller) will not do retries.  So if that is the case, Kubelet (container notifier controller) will only process the pods matching the label when the Notification object is first created?  If so, Pod D will not be processed?

keps/sig-node/1977-container-notifier/README.md
xing-yang: Are there any existing examples that I could follow?

keps/sig-node/1977-container-notifier/README.md
krmayankk: Is this a namespaced object ?  Once its created, it should apply to everything  until the api object is deleted . 
Is this more of an imperative api or we want it to be declarative (i guess the latter )?
i am thinking what does the declaration mean ?
- notify all containers that match this selector at this time t
- let there be notification for all pods matching this selector as long this api exists
- i want all containers of the pods matching a selector to be quiesced (this seems more natural)
its a bit confusing to think since the lifecycle hooks are tied to the lifecycle of the container/pod and are ok to be imperative, but this api is not tied to the lifecycle but more of a command to the containers.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+           // Names with a prefix such as `example.com/label-style` are custom names.
+           ContainerNotifierName string
+ }
+```
+
+###### Notification Status
+
+NotificationStatus represents the current state of a Notification. It has a list of PodNotificationStatuses.
+
+A PodNotificationStatus represents the current state of a Notification for a specific pod. A PodNotificationStatus contains ContainerNotificationStatuses of all the containers which have the corresponding ContainerNotifier specified. PodNotificationStatus is created when a Notifier starts to run in a pod.
```
krmayankk, line 231: why a specific pod? this should contain the status of notification for all pods that match the selector right ?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+type ContainerNotifierHandler struct {
+    // Exec specifies the action to take.
+    // +optional
+    Exec *ExecAction `json:"exec,omitempty" protobuf:"bytes,1,opt,name=exec"`
+
+    // Signal specifies a signal to send to the container
+    // +optional
+    // define constants for signals?
+    // validate the signals are valid?  windows?
+++    Signal string
```
krmayankk, line 328: Do we need to start with signal in phase 1 and then add exec later if that is necessary ?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+   Name string
+   Action *ContainerNotifierAction
++   // Add a Probe in Phase 2. Probe defined in the core API here will be used.
++   Probe *Probe
+}
+```
+
+### Kubelet Impact in Phase 2 and Beyond
+
+When moving the logic from the container notifier controller to Kubelet, there will be impact on Kubelet.
```
krmayankk, line 362: its not clear from this proposal why the controller is starting with a separate controller(also not clear if its separate or part of one of the existing controllers) and then later moving to kubelet

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+https://github.com/moby/moby/blob/master/client/container_kill.go#L9
+```
+
+We can add a `SignalKill` method in `~/go/src/k8s.io/kubernetes/pkg/kubelet/dockershim/libdocker/kube_docker_client.go` and call docker client `ContainerKill`, similar to how `CreateExec` method calls docker client `ContainerExecCreate`.
+
+## Implementation Plan
+
+### Phase 1
+
+Phase 1 API definition will be added to Kubernetes. Implementation in this phase will not happen in Kubelet. Instead, it will be implemented in a single *trusted* controller which acts on Notifications via exec (with the goal being to move that into Kubelet in the next steps). In this phase, only exec will be included and signal will not be included as that involves changes to CRI.
```
krmayankk, line 403: so a separate external controller running per node (daemon set) ? In order to run the exec commands , will it exec into the pods ?

xing-yang:
CC @andrewsykim

keps/sig-node/1977-container-notifier/README.md
xing-yang: Yes.

keps/sig-node/1977-container-notifier/README.md
xing-yang: This will start as a separate controller in a new repo, and later on the logic will be moved to kubelet.
This is because there are concerns from sig-node on potential impact on kubelet.

keps/sig-node/1977-container-notifier/README.md
xing-yang: Support for signal requires changes to CRI. That means it can only happen in Kubelet.  Since we are starting with a separate controller, we can't start with signal.

Also support for exec is an important use case (the quiesce use case for application consistent snapshot). So support for exec should be added in phase 1.

keps/sig-node/1977-container-notifier/README.md
xing-yang: For a particular PodNotificationStatus, it represents status for a particular pod.  Yes, all pods matching the selector will be selected.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+   Action *ContainerNotifierAction
+
+    // Number of seconds after which the notifier times out.
+    // Defaults to 1 second. Minimum value is 1.
+    // +optional
+    TimeoutSeconds int32
+}
+
+// ContainerNotifier describes a command to run in the container.
+type ContainerNotifierAction struct {
```
thockin, line 180: I was proposing that we get rid of this layer of nesting:

```
Pod {
  Spec: {
    Containers: [{
      Notifiers: [{
        Name: foo
        TimeoutSeconds: 10
        Action: {   # note no "handler"
          Exec: {
            Command: bar
          }
        }
      }]
    }]
  }
}
```

keps/sig-node/1977-container-notifier/README.md
thockin: @krmayankk is on the issue.

This is a pseudo-imperative operation.

@xing-yang 


> Kubelet (container notifier controller) will only process the pods matching the label when the Notification object is first created

That seems reasonable at first blush, but what is the mechanism for that?  Is the rule "only applies to pods whose `metadata.creationTimestamp` is less than the Notification's `metadata.creationTimestamp` ?  If so, let's document it.  Also - less-than or less-than-or-equal?

Does this satisfy the higher-order controller's needs?  Is it plausible that a quiesce notification could be created, a snapshot started, then a new pod shows up which is not quiesced?  Even if Notifications applied to it, that would be async, so it still seems bad?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+type NotificationSpec struct {
+            // PodSelector specifies a label query over a set of pods.
+            // +optional
+            PodSelector *metav1.LabelSelector
+
+           // This contains the name of the ContainerNotifier to send to a container.
+           // Names are label-style, with an optional prefix. Names without the prefix
+           // such as `quiesce` are reserved for Kubernetes-defined "well-known" names.
+           // Names with a prefix such as `example.com/label-style` are custom names.
+           ContainerNotifierName string
```
thockin, line 223: minor: This is a needlessly long name.  Just `Notifier` or `NotifierName` perhaps?

keps/sig-node/1977-container-notifier/README.md
thockin: EndpointSlice is similar - I am not sure there's a closer example.

keps/sig-node/1977-container-notifier/README.md
thockin: We should not go GA (and maybe not beta) without phase 2, but it is phased for a reason.

keps/sig-node/1977-container-notifier/README.md
thockin: An abundance of caution

keps/sig-node/1977-container-notifier/README.md
thockin: not needed per-node, just per-cluster

keps/sig-node/1977-container-notifier/README.md
sjenning: yes, agreed

sjenning:
I would like to see example pod specs (at least the container blocks) for the examples and use cases mentioned.

I understand I should be able to think about what it would look like from the API definition, but just to make it clearer.

It exposes issues like what @thockin noticed [here](https://github.com/kubernetes/enhancements/pull/1995#discussion_r492865819)

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: @thockin using CreationTimestamp can certainly be one option to rule out newly created pods, it does not cover the case when a pod is evicted/died during hook execution. From quiescing requirements perspective, its critical to have a point-in-time "fixed" scope for snapshotting, thus if a pod died during the notification, we might want to fail the the whole notification.
we have a couple of other potential options here:
1. make notification to pod one-to-one by replacing podSelector with a pod definition. higher level controller like application snapshot controller then can have a pre-selected list and create one notification to each desired pod.
Advantage:

- simpler API
- ordering of notification is possible and more convenient to achieve for higher level controllers
disadvantages:
 a. probably does not fit well with the signal case where multiple pods need to be notified without a higher level controller

2. introducing a list of resolved pod in Notification.Spec which will be immutable once resolved. Pod selection is done prior to any real execution of notification. And that list would be the source of truth for the scope of execution.
WDYT?

keps/sig-node/1977-container-notifier/kep.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.20"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.20"
+  beta: "v1.21"
```
yuxiangqian, line 36: I think this is too aggressive given that the amount of work.

keps/sig-node/1977-container-notifier/README.md
thockin: This is an interesting problem.  If a pod dies mid-snapshot, do you fail the whole snapshot?  If a new pod a starts mid-snapshot (scale-up?) does it need to quiesce?  Maybe that just "doesn't happen" ?

Your (1) happens to ALSO solve the status cardinality problem, so is attractive in that regard.  It means the scale of things kubelet has to watch just got much bigger, though, unless we set labels to scope it (e.g. set the node-name as a label, let kubelet watch only that).

But it is ugly for humans.

Another approach would be to define Notifications as applying to all pods.  In the example above it *would* apply to D.  And E, F, G etc until the Notification is removed.  All the controller (or kubelet) could guarantee would be that every Notification that selects a given pod would be attempted once, and if there were multiple notifications, they would be attempted in some deterministic order (e.g. creationTimestamp or RV).

So to expand the example:

t0:  pods A, B, C are running, labelled foo=bar
t1:  create Notification N1 selecting foo=bar
t2:  N1 notifiers in A, B, C are done
t3:  pod D is created, labelled foo=bar
t4:  N1 notifier in D is done
t5:  create Notification N2 selecting foo=bar
t6:  N2 notifiers in A, B, C, D are done
t7:  pod E is created, labelled foo=bar
t8:  N1 notifier in E is done
t9:  N2 notifier in E is done
t10: delete Notification N1
t11: pod F is created, labelled foo=bar
t12: N2 notifier in F is done  ## note N1 was not executed
t13: delete Notification N2

This seems maybe workable, but I need you to look at it from your use-cases.  Bear in mind that rapid sequences can be observed out of order by watchers, so if your plan would be to delete N1 then create N2 you need some time in between to know that all kubelets have observed the deletion (or ensure that it is "safe" to trigger N1 on new pods).  Similarly, if you want to clean up N1 and N2 you *must* delete them in order with sufficient time between to ensure no out-of-order deletion problems.

We don't have a "serializing barrier" operation. 

Regarding "ugly for humans", maybe we need both?  Individual pod-notifications (and status) which controllers can orchestrate PLUS a selector-notification which a built-in controller would expand into pod-notifications?  This represents a broader API change, but is the more flexible model.

Thoughts?

Regarding your (2) - does that satisfy requirements?  Let me see if I get it right:

t0:  pods A, B, C are running, labelled foo=bar
t1:  create Notification N1 selecting foo=bar
t2:  a controller evaluates foo=bar one time and notes the result (A, B, C)
t3:  N1 notifiers in A, B, C are done
t4:  pod D is created, labelled foo=bar
t5:  N1 notifier in D is NOT executed

Is that right?  How does that map to the use-cases?

thockin (commented):
Corner cases are so much fun

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,773 @@
...
+     // +optional
+     Lifecycle *Lifecycle `json:"lifecycle,omitempty" protobuf:"bytes,12,opt,name=lifecycle"`
+
+    // Notifiers can be triggered by Notification resources. Each notifier must
+    // have a unique name within the container definition, but the same name may be
+    // used in different containers of the same pod. When a Notification is created
+    // which selects a pod and matches the Notifier name, the Notifier will run in the
+    // container.
+    // +optional
+    Notifiers []ContainerNotifier
```
krmayankk, line 154: do you think we could have named the lifecycle hooks and reused them ?

keps/sig-node/1977-container-notifier/README.md
krmayankk: +1

keps/sig-node/1977-container-notifier/README.md
krmayankk: Applying to all pods for as long as the notification exists makes more sense from an api point of view. Also including the pod definition would be messy.

krmayankk: > @krmayankk is on the issue.
> 
??

> This is a pseudo-imperative operation.
> 
Is that ok to have ? Do we need to update any apiconvention guidelines on when a pseudo imperative api makes more sense ?

keps/sig-node/1977-container-notifier/README.md
krmayankk: Do we need this at pod level, to mean it applies to all containers ?

keps/sig-node/1977-container-notifier/README.md
krmayankk: It would be good to think about who can send notifications ? What permissions they need on a pod ?. Since the controller is doing exec, does it mean as long as exec is opened for the controller, anyone can create and run the commands using controller as a proxy?

Another idea is should we make notifier a subresource ?

keps/sig-node/1977-container-notifier/README.md
liggitt: > NotificationStatus represents the current state of a Notification. It has a list of PodNotificationStatuses.

when we talked about this in doc form, I thought notification was scoped to a single pod in order to allow [notifying in waves](https://docs.google.com/document/d/1SWSlZoxY5zFjBKFKaATP07s3q02UenSp8R9-yRkCcwg/edit?disco=AAAAJ1RQggU) and [avoid unbounded status accumulation](https://docs.google.com/document/d/1SWSlZoxY5zFjBKFKaATP07s3q02UenSp8R9-yRkCcwg/edit?disco=AAAAKEpj35c) ... did that approach get revised since then?

xing-yang:
CC @ashish-amarnath

keps/sig-node/1977-container-notifier/README.md
xing-yang: So that means we can't have a PodSelector in the NotificationSpec.  It will just be a PodName instead.

keps/sig-node/1977-container-notifier/README.md
xing-yang: For the quiesce use case, a higher level backup controller needs to decide what pods are included. When a backup is started, the backup controller needs to figure out all the pods that need to get the quiesce command.  It can retrieve this information using a PodSelector.  After that it is a fixed list of Pods that need to be quiesced for this particular backup.  Therefore, we can’t have PodSelector in Notification.  It should be either a List of Pods that is immutable or a Pod name for a single pod, basically #2 and #1 suggested by Xiangqian.

If we decide to have a PodList in NotificationSpec, the steps would be like this:

1. An external Backup controller determines what pods need to receive the notification by using a PodSelector (this depends on backup controller logic).
2. Backup controller creates a Notification object with a PodList with the selected pods.

This way we have a concrete list of Pods. If quiesce fails for any pod/container, it should be treated as failure for the whole quiesce operation.  Typically the backup software gives user an option to complete a backup that is not quiesced successfully or fail it if any quiesce action fails. If user chooses to do a backup even if quiesce fails, the backup controller will trigger a un-quiesce command to undo anything that is already quiesced but skip doing quiesce for the remaining pod/container.  Then the backup controller will take a snapshot and mark that snapshot as not-quiesced.  User can decide what to do with a snapshot that is not quiesced.

keps/sig-node/1977-container-notifier/README.md
xing-yang: @thockin would you consider having one to one mapping between a Notification and a Pod?  That would simplify status update.

krmayankk: Just read the discussion in the doc, I think the notification per pod makes most sense and seems safer. Also the idea of building on top using a notification controller for safe rollout of notification is even better

keps/sig-node/1977-container-notifier/README.md
thockin: I don't think reusing hook definitions is worth the effort - early versions of this had a new resource just for that and I found it quite cumbersome.  It's part of the pod, just like a probe.

thockin: Re pod-level - I think that's an optimization (for laziness) which we can explore if it comes up.  I doubt it.

keps/sig-node/1977-container-notifier/README.md
thockin: ping

keps/sig-node/1977-container-notifier/README.md
thockin: > we can’t have PodSelector in Notification

I'm not sure I buy that.  This API needs to be general enough to use by humans.  We need to be clear about what happens if a new pod is created that satisfies the criteria.  Don't over-focus on just quiesce

keps/sig-node/1977-container-notifier/README.md
thockin: RBAC governs who can create Notifications in a namepace - isn't that enough?

keps/sig-node/1977-container-notifier/README.md
thockin: If we go to 1:1, do we need a higher-level API that selects pods and expands a notification for each of them?or should we add that to kubectl?  or should we just say "someone else's problem"?

If we go 1:1 does a subresource make more sense?  It gives even finer RBAC control, but we need the result to be a new resource (for status).  E.g. POST to /pods/foobar/notify -> creates a PodNotifier; kubelet watches PodNotifiers.  Any precedent for that?  Any need to get finer-grained than per-namespace?

thockin (commented):
Given the open comments should we discuss again?  Is this stalled on a decision or on hands-on-keyboards?

xing-yang:
> Given the open comments should we discuss again? Is this stalled on a decision or on hands-on-keyboards?

@thockin Yes, I think we should discuss again as there are a few open issues.  I'll try to set up a meeting.

yuxiangqian:
> Given the open comments should we discuss again? Is this stalled on a decision or on hands-on-keyboards?
>
@thockin there are couple of other things(VolumeSnapshot GA, KubeCon US etc) going on concurrently for both @xing-yang and me, and we should have more spins after this week. I am reworking some pieces of this KEP for a bit to make it more clear, and we will schedule a discussion to talk about the Notification piece. Thanks for your patience.

thockin:
NP - I have a fair bit of free time for the next few weeks, so I am circling back to all the KEPs I am tracking

keps/sig-node/1977-container-notifier/README.md
krmayankk: I like the subresource idea which i was hinting here https://github.com/kubernetes/enhancements/pull/1995/files#r493239279 . I guess i don't follow why it should result in a new PodNotifier being created? For e.g. does the POST on a log or the exec sub resource result in a new object ? The main reason i like this is its less number of new objects to manager , less strain on etcd, since the notifications are an array. It does make the RBAC more fine grained. Thinking out loud, if i am not mistaken, you can say a service account can or cannot exec to any pods in a namespace, So similar should be possible with this notify subresource

keps/sig-node/1977-container-notifier/README.md
krmayankk: see here https://github.com/kubernetes/enhancements/pull/1995/files#r555640516

thockin:
1.21 KEP freeze is in O(days) - if this wants review before then, I need updates.

yuxiangqian:
Apologies for the late update. We have revisited some of the designs in this KEP to address some of the major concerns raised in the previous version(details in this [document](https://docs.google.com/document/d/1lPXtF58qpdrpR6vsRREGTzhPRkHOc1qMbcPrnjfxrgU/edit#)).

Major changes:
1. Resolve the concerns around potential scalability issues when a Notification targets at a large number of Pods(~Ks).
2. Shrink the scope in phase 1 to reduce implementation complexity and simpler PodNotification API.
3. Resolve too-deeply nested structure concern.

keps/sig-node/1977-container-notifier/kep.yaml
xing-yang: These milestones need to be updated.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    ......
+    // +optional
+    Lifecycle *Lifecycle `json:"lifecycle,omitempty" protobuf:"bytes,12,opt,name=lifecycle"`
+
+    // Notifiers can be triggered by PodNotification resources.
+    // Each notifier must have a unique name within the Container.
+    // Upon creation of a PodNotification, Container(s) in the specified
+    // Pod with a matching notifier name will be executed.
+    // +optional
+    Notifiers []ContainerNotifier
```
shawn-hurley, line 149: Is this mutable? Should someone be able to add a notifier to a pod after it is running I guess is the question that I am asking?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    // behavior.
+    // For example, when a quiesce command has been ran on a database, the
+    // database will stay quiesced, and any consecutive execution of the
+    // same quiesce command should not have additional impact.
+    Handler *ContainerNotifierHandler
+
+    // Number of seconds after which the notifier will time out.
+    // Default to 1 second. Minimum value is 1.
+    // +optional
+    TimeoutSeconds int32
```
shawn-hurley, line 174: Should there be a Handler that is run on timeout?

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: "Containers" field in Pod can not be updated that is my understanding.
[ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#podspec-v1-core)
IMO, it also seems to be reasonable to make the "notifiers" immutable such that only a restricted/predefined set of notifiers can be triggered. Are there any specific cases you see the needs of dynamically adding/removing notifiers from a specific container?

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: How to handle timeout/errors is indeed an area that the current API proposal intentionally tries to simply and transfer some of the complexity to users. Could you give some concrete examples? One thing that bothers me for introducing  error/timeout handlers is that another "timeout" might be needed for the error/timeout handler itself. It would also complicate the API/implementation quite a bit. WDYT?

shawn-hurley: I agree that it will complicate things, what I was worried about is if I quiesce an application but  it times out, to bring the application back up to running, we would want to run an unquiesce operation? 

I guess this could be solved by the notifier having a graceful termination with a shutdown signal being sent on timeout? Wondering if it should be apart of the enhancement?

keps/sig-node/1977-container-notifier/README.md
shawn-hurley: I was wondering if it would be possible to do that, I can't think of a good use case thanks!

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
++   // Add a Probe in Phase 3. Probe defined in the core API here will be used.
++   Probe *Probe
+}
+```
+
+### Kubelet Impact in Phase 2 and Beyond
+
+When moving the logic from the PodNotification controller to Kubelet, we foresee the following changes to happen in Kubelet:
+
+- Kubelet will watch PodNotification resources cluster-wide.
```
thockin, line 471: It only needs to watch notifications for pods on this node.  That's a potentially important optimization we should plan for,

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    // It indicates when the error occurred
+    Timestamp metav1.Time
+}
+
+type PodNotificationErrorType string
+
+// More error types could be added, e.g., Forbidden, Unauthorized, etc.
+const (
+    // Pod does not exist
+    NonExistingPod PodNotificationErrorType = "NonExistingPod"
```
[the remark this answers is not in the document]
yuxiangqian: ack

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    Status NotificationStatus
+}
+
+type NotificationSpec struct {
+    // A label query over Pods under the same namespace of the Notification
+    // to notify.
+    // Note: Only Pods in the ‘Running’ phase are eligible for notification.
+    // Ref: kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase
+    // Required.
+    PodSelector metav1.LabelSelector
```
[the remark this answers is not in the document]
yuxiangqian: ack. done

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+
+    // Policy specifies pod selection policy. If not specified, default to
+    // "AllPods".
+    // +optional.
+    Policy *PodSelectionPolicy
+
+    // The name of the ContainerNotifier to trigger.
+    // Name is a DNS_LABEL.
+    Notifier string
```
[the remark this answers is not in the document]
yuxiangqian: removed this comment.

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    Parallelism int
+}
+
+type PodSelectionPolicy string
+
+const (
+    // AllPods means all Pods matches the specified label selector in Notification
+    // will be notified. If this policy is specified, CompleteTime field of the
+    // NotificationStatus will not be set.
+    AllPods PodSelectionPolicy = "AllPods"
```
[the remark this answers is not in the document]
yuxiangqian: For the "AllPod" one, we can simply remove this type, "PodSelectionPolicy" is just an optional field in spec, we can treat the default behavior to "AllPod".
For the "PreExistingPods" option, it's really to support the case for application consistency snapshot.

I see there's a lot of confusions and concerns over this model, and there is indeed another alternative which is acceptable to me:

1. Get rid of the PodSelectionPolicy completely from Notification, and the behavior will be always try to notify any newly created Pod which matches the label selector until the Notification resource is deleted.
2. For those upper level controllers which rely on completion of PodNotifications, i.e., quiesce a pod before taking a snapshot of it's volumes, they can always use the `PodNotification` type to get finer grained control. (Having the upper level controller/end user to handle the complexity, and leave the API simple).

I personally are fine for both models, and if there are not other cases, I would actually prefer the alternative to have a simpler API.

WDYT?

@xing-yang any comments?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    // +optional
+    StartTime *metav1.Time
+
+    // If set, it suggests the timestamp(UTC) at when the Notification has
+    // been marked as completed. A Notification is completed when all
+    // PodNotifications created from this Notification have completed.
+    // Note that this field will never be set if "AllPods" is specified as
+    // pod selection policy.
+    // +optional
+    CompleteTime *metav1.Time
```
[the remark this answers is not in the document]
yuxiangqian: see comments above for the alternative, if we achieve an agreement there for the proposed alternative, then we can remove this field.

I assume you meant having a GoalCount in `NotificationSpec`? That has the risk racing in the following situation:
1. prior to create notification, pod count = X
2. create a notification with label selector and goalcount = X
3. a pod died, or a new pod has been created, now the pod count = Y
4. during execution, X != Y, which ones to run the command if Y > X? or Y < X, should it fail?

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: agreed. Updated this point.

keps/sig-node/1977-container-notifier/README.md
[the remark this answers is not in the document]
yuxiangqian: also I would like to point out this is a phase 2 item, and we can always adjust the design should there be more use cases or new requirements

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,877 @@
...
+    // Type of the error
+    // Required.
+    Type PodNotificationErrorType
+
+    // A human-readable message indicating details why the action failed.
+    Message string
+
+    // Reason is a brief CamelCase string that describes the failure and it’s
+    // meant for machine parsing and tidy display in the CLI.
+    Reason string
```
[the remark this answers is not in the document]
yuxiangqian: This type used to be  a Condition type, and changed to be an error. removed reason.

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: the main issue with introducing a graceful termination in the API is that it does not fundamentally solve the problem. It's possible the graceful termination mechanism itself is not graceful enough :). Meaning, it might end up as a failure. It seems to be reasonable to me that Applications/controllers understands the semantic of returned errors of a notification, and handle it appropriately (i.e., keep on retrying an un-quiesce hook until it succeeds).

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -246,48 +246,51 @@ type PodNotificationStatus struct {
...
 
     // Succeeded will be set to true when the action has been executed
-    // successfully in all matched containers within this Pod,
+    // successfully in all qualified containers within this Pod,
     // and to false otherwise. It will not be specified if the status of
-    // of execution is unknown(i.e., in execution).
+    // of execution is unknown(i.e., still in execution).
     // +optional
-    Succeeded *metav1.Time
+    Succeeded *bool
```
thockin, line 260: I think this is better served by an enum `result` (or something like that) with values for success and failure, so we can later add other things.

Thoughts?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -312,11 +311,11 @@ type PodNotificationErrorType string
 
 // More error types could be added, e.g., Forbidden, Unauthorized, etc.
 const (
-    // Pod does not exist
-    NonExistingPod PodNotificationErrorType = "NonExistingPod"
+    // Pod not found error
+    PodNotFound PodNotificationErrorType = "PodNotificationErrorPodNotFound"
```
thockin, line 315: You changed this the opposite of what I meant :)

`PodNotifictionErrorPodNotFound PodNotificationErrorType = "PodNotFound"`

keps/sig-node/1977-container-notifier/README.md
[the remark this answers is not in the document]
thockin: > For the "PreExistingPods" option, it's really to support the case for application consistency snapshot.

I'm still not convinced that would be correct.  If a new pod started during a snapshot (buggy controller?) wouldn't you want to notify that one?  That raises a fun corner case - suppose I deliver the notification to pod P, and then there was an OOM kill during the snapshot, causing pod P to restart.  We would want to re-deliver the notification, right?  What ensures that, and how do I report it?

keps/sig-node/1977-container-notifier/README.md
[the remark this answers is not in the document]
thockin: I meant under status - the controller would update the goal count if new pods start that match the selector.  Then a client could look at the number of completions and the goal count.  When that "settles" (which is really context dependent) then we know it's "done enough".  That way it applies in all modes (all pods or pre-existing).

thockin:
I am pretty close to happy with this.

keps/sig-node/1977-container-notifier/README.md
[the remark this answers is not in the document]
yuxiangqian: IMO, these cases should ALL be treated as a failure of the Notification. The definition of PreExistingPods is to only cover those ones are running prior to the creation of a Notification.
If PreExistingPods is specified,
1. Any existing Pod crashes/restarts fails the Notification
2. Newly created Pod will *not* be notified

If re-delivery is needed, a user can always elect to use "AllPods"?

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: Ack

keps/sig-node/1977-container-notifier/README.md
yuxiangqian: make sense, will update

thockin (commented):
Thanks!

/lgtm
/approve

keps/sig-node/1995-container-notifier/README.md
```diff
@@ -0,0 +1,899 @@
...
+type PodNotificationSpec struct {
+    // The name of the Pod to which notification will be addressed.
+    // Note that a PodNotificaion only notifies a Pod in the same namespace.
+    // The PodNotification fails immediately if the specified Pod does not exist.
+    // Required.
+    // Immutable.
+    PodName string
+
+
+    // The name of the ContainerNotifier to trigger.
```
ruiwen-zhao, line 216: PodNotificationSpec only has PodName and ContainerNotifier name (which is only container-level unique). So if a pod has two containers, each of which has a ContainerNotifier with the same name, how does PodNotificationSpec distinguish between them?

xing-yang: If two containers have the same ContainerNotifier, then the same command will be run in both containers; If the intention is to run different commands in two different containers, each should use a different ContainerNotifier name.

ruiwen-zhao: Got it, as long as this is intended we should be good. Actually you already mentioned that "All containers in the specified `Pod` which have the requested "ContainerNotifier" defined in their spec will have their corresponding "ContainerNotifierHandler" executed".

keps/sig-node/1995-container-notifier/README.md
```diff
@@ -0,0 +1,899 @@
...
+// Notifiication.
+type PodNotificationErrorType string
+
+// More error types could be added, e.g., Forbidden, Unauthorized, etc.
+const (
+    // Pod not found error
+    PodNotificationErrorPodNotFound PodNotificationErrorType = "PodNotFound"
+
+    // The execution of a handler timed out.
+    PodNotificationErrorHandlerTimeout PodNotificationErrorType = "HandlerTimeout"
```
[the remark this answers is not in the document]
yuxiangqian: yes, it will be surfaced both in PodNotificationStatus aggregated field "State" as well as detailed in "Containers" which users can tell in which container the notifier failed.

ruiwen-zhao (approved):

keps/sig-node/1995-container-notifier/README.md
thockin: It notifies all containers with that named notifier

xing-yang:
KEP number in kep.yaml should be the enhancement issue number 1977.  Fixed it now.

sjenning:
/remove-lifecycle stale

Phase 1 looks acceptable to me.

The kubelet needs no change; only the pod spec.  The kubelet does not need to be notification aware in phase 1.

Completion of phase 1 will allow us to prove out this idea and root out potential issues with the design, both API and performance/scale, before lumping the controller logic into the kubelet where it will be harder to debug and isolate as a source of scaling issues.

dchen1107:
@sjenning thanks for taking another look on this. Like what I mentioned at last week's SIG Node, the new proposal narrowed down the scope a lot based on our offline review meetings and comments. Due to the overwhelm of the reviewers last cycle, we postponed this feature, and hopefully we can finish the phase 1 for 1.23 this cycle.

derekwaynecarr:
I had a number of questions/clarifications I hope we can get through in SIG Node call, otherwise I will paste here right after.

sjenning:
Derek and I talked about this.  Some thoughts.

Point in time nature of Notifications. Notification created at time X might behave differently from one created at time X+1, due to the PreExisting SelectorPolicy and Pod creation times i.e. if the PodNotification is created before the Pod, the notification fails. PodNotifications are a one shot thing.

Notification controller watches all pods for CREATE (UPDATE?) to see if the selector matches and then creates PodNotifications on container start? container ready? pod ready?  Do pods that have deletionTimestamp set get notifications?

If the pod is deleted, are all PodNotifications that reference the pod deleted too?  Is the ownerRef for the PodNotification the pod? or the Notification?

Number of PodNotification resources in the cluster would be the sum of all matching pods over all Notifications, a potentially very large number.  How are these cleaned up?  Notification controller (i.e. kubelet in the future) would have to WATCH PodNotifications cluster-wide right?  I don't see a way to apiserver-side filter the watch to only PodNotifcations referencing a pod a particular kubelet is running.  Thus for each change to a PodNotification cluster-wide, a watch event is sent to each kubelet.

QPS for the controller updating all these PodNotification resources will need to be high.  Stress on container runtime and apiserver.  A potential scaling foot-gun.  We have had issues with kubelet QPS in the past delaying pod status updates and leading to all manner of pain.

We have a number of container types including init and ephemeral containers that also use the Container type.  Do notifications work for them?  Doesn't seem to make sense for those container types.
Point in time nature of Notifications. Notification created at time X might behave differently from one created at time X+1, due to the PreExisting SelectorPolicy and Pod creation times i.e. if the PodNotification is created before the Pod, the notification fails. PodNotifications are a one shot thing.

Notification controller watches all pods for CREATE (UPDATE?) to see if the selector matches and then creates PodNotifications on container start? container ready? pod ready?  Do pods that have deletionTimestamp set get notifications?

If the pod is deleted, are all PodNotifications that reference the pod deleted too?  Is the ownerRef for the PodNotification the pod? or the Notification?

Number of PodNotification resources in the cluster would be the sum of all matching pods over all Notifications, a potentially very large number.  How are these cleaned up?  Notification controller (i.e. kubelet in the future) would have to WATCH PodNotifications cluster-wide right?  I don't see a way to apiserver-side filter the watch to only PodNotifcations referencing a pod a particular kubelet is running.  Thus for each change to a PodNotification cluster-wide, a watch event is sent to each kubelet.

QPS for the controller updating all these PodNotification resources will need to be high.  Stress on container runtime and apiserver.  A potential scaling foot-gun.  We have had issues with kubelet QPS in the past delaying pod status updates and leading to all manner of pain.

We have a number of container types including init and ephemeral containers that also use the Container type.  Do notifications work for them?  Doesn't seem to make sense for those container types.

derekwaynecarr:
> I don't see a way to apiserver-side filter the watch to only PodNotifcations referencing a pod a particular kubelet is running. Thus for each change to a PodNotification cluster-wide, a watch event is sent to each kubelet.

I think for this item if we can integrate with the `NodeAuthorizer` admission plugin, it would be good.

derekwaynecarr:
To summarize:

I think this feature needs a garbage collection strategy so we understand the anticipated life time of a PodNotification resource as both a provider/consumer.  I would think its reasonable to say that a Pod can have X active PodNotification resources per pod.  This helps us budget QPS, plan for handler costs, etc.  Completed PodNotification resources must be garbage collected in order to avoid denial of service for both the kubelet (memory growth) and kube-api (similar).  TTL on the resource is one way to do that.

Exec/probe costs are really expensive in practice and a key input in understanding node sizing, so having a fixed budget to plan for is helpful.  For awareness, pods whose liveness or readiness probes are done with too great of frequency or make heavy use of exec dominate cpu resource usage by the kubelet and runtime.  This is why having a limit on the number of concurrent notifications at admission time is helpful.

Adding ContainerNotificationStatus to include ContainerID so it maps to what is shown in PodStatus would let us know if the value is unique across time and space.

When in the pod lifecycle is the pod accepting signals, is a signal ever needed before success of particular probes (liveness/readiness/etc.)?  If not, should we scope it to a particular well-defined phase in the Pod lifecycle.

Is there a reason to not explicitly exclude init containers or ephemeral containers as notification targets?

If a pod is undergoing termination, is a signal still sent?

If a notification is created for a terminal pod, who updates the status?

If a notification is created that references a notifier that does not exist in the pod spec, what happens?  Does that fail at validation time or runtime?

Should the notification reference a pod by UUID and not just name so it is also bound in time and space?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,899 @@
...
+### Non-Goals
+
+- Writing concrete commands, for example quiesce scripts, which could be executed. It is the responsibility of users who set up pod definitions.
+
+## Proposal
+
+In phase 1, propose to introduce several API changes:
+- Adding an optional field `Notifiers` which is a list of `ContainerNotifier`s into `Container`.
+- Adding an inlined type `ContainerNotifierHandler` which defines a command.
+- Adding a core API type `PodNotification` which defines request type to trigger execution of `ContainerNotifier`(s) in a `Pod`.
```
smarterclayton, line 115: I don’t see a defense/justification for this object in the examples of this KEP (or alternatives) vs simply having a spec generation driven interaction in the pod itself for phase one.  If it’s been discussed and rejected, I’d like it added to alternatived (so i can see the critique).  If it hasn’t been discussed, I’d prefer it to the addition of a new object and controller in phase one that seem like they are “controller for controllers sake”.

Roughly, id expect pod.spec.containers.notifier to have a blank generation field that defaults to spec generation on edit of the notifier struct, or is 0 on create.  And then status to have “notifiers[].generation” that indicates if the kubelet executed the signal.  Clients clear spec generation to trigger the signal and wait for status generation to match to know when the kubelet successfully “leveled” the notifier (delivered it).

That seems a more useful primitive than the hypothetical object and controller, can be done in phase one so external controllers can invoke signals, and doesn’t require the extra alpha baggage until later (bulk notification can be done as a separate orthogonal phase around generation driven signaling).

yuxiangqian:
> To summarize:
> 
> I think this feature needs a garbage collection strategy so we understand the anticipated life time of a PodNotification resource as both a provider/consumer. I would think its reasonable to say that a Pod can have X active PodNotification resources per pod. This helps us budget QPS, plan for handler costs, etc. Completed PodNotification resources must be garbage collected in order to avoid denial of service for both the kubelet (memory growth) and kube-api (similar). TTL on the resource is one way to do that.
> 
Added a TTLSecondsAfterCompletion field to PodNotificationSpec and NotificationSpec
> Exec/probe costs are really expensive in practice and a key input in understanding node sizing, so having a fixed budget to plan for is helpful. For awareness, pods whose liveness or readiness probes are done with too great of frequency or make heavy use of exec dominate cpu resource usage by the kubelet and runtime. This is why having a limit on the number of concurrent notifications at admission time is helpful.
> 
Will address this in a separate commit.
> Adding ContainerNotificationStatus to include ContainerID so it maps to what is shown in PodStatus would let us know if the value is unique across time and space.
> 
Added
> When in the pod lifecycle is the pod accepting signals, is a signal ever needed before success of particular probes (liveness/readiness/etc.)? If not, should we scope it to a particular well-defined phase in the Pod lifecycle.
> 
Add clarification to the API. A `PodNotification` should fail immediately if a `Pod` has `PodPhase` other than `PodPending` or `PodReady`, in case of `PodPending`, the PodNotification will hold until the Pod is ready.
> Is there a reason to not explicitly exclude init containers or ephemeral containers as notification targets?
Good point, made it clear in API that InitContaines and EphemeralContainers will be ignored.
> 
> If a pod is undergoing termination, is a signal still sent?
Clarified in API, only `PodReady` and `PodPending` will be considered. For a pod undergoing termination, I am thinking about scenarios like a container can not be terminated successfully (stuck in some loop for example), and a signal is needed to sent to the container to force kill. Does that sound like a reasonable use case? Another aspect is, only containers still in "Running" state will have the notification executed.
> 
> If a notification is created for a terminal pod, who updates the status?
In phase one, that would be the **trusted** controller, and phase 2 will be Kubelet.
> 
> If a notification is created that references a notifier that does not exist in the pod spec, what happens? Does that fail at validation time or runtime?
I assume you mean PodNotification here, I am thinking an admission control to this kinda of validation to avoid the PodNotification from creation. Will address in a separate commit.
> 
> Should the notification reference a pod by UUID and not just name so it is also bound in time and space?
Good Point, added UID into PodNotificationStatus

xing-yang:
/assign @johnbelamaric 
PRR is ready for review.

keps/sig-node/1977-container-notifier/README.md
thockin: To be honest, this design was trying to minimize the number of moving parts within Pod, but if folks want to explore that as an alternative...

Pod doesn't seem to use `generation` so first we enable that.  Or use RV ?

So to trigger a notifier, we'd write 0 to the pod...notifier generation, which is roughly equivalent to creating a PodNotifier as specced today.  Right?

That does get rid of PodNotifier type in exchange for more moving parts in Pod itself.  In particular if means that any controller which converts a higher-level Notifier (eg, notify all pods under a selector) needs RBAC to mutate Pod.  That seems strictly worse to me, unless we add a subresource for notifications.

Do you perceive that as better?  I don't hate it, and it eliminates a whole category of problems.

We still need a controller.  Two in fact - one to actuate the notifications (eventually kubelet) and one to take higher level "notify by selector" and convert that to N discrete notifications (or we push that up onto all consumers of notifiers).  As you say, that's not phase 1 necessary.

thockin:
Given Clayton's comment, it seems unlikely that this will be resolved in time for 1.23 - I am hopeful, but trying to be realisitic.

johnbelamaric:
Perhaps this has been discussed, but I have to ask why this needs to be a core API.

It would seem the original motivation - run a quiesce command in each Pod - could be handled via an agent container living in the Pod that is watching a CR. Time may be better spent designing an out-of-tree solution based on a CR and a general agent that can run in a container in the Pod and act locally on those pod containers. Unless there is some reason that we need a privileged host-level process like `kubelet` to take some action, I can't see a justification for adding this level of complexity for something that can be solved without it.

thockin:
I think I disagree with John on this one, but I admit I am pretty close to it, which colors my perspective.  It might be worth trying.  For example, using a sidecar to enact your own notifications might be possible - how streamlined could it get?

johnbelamaric:
This is on my PRR list but I don't see any updates since pre-1.23. Assume this is a no-go for 1.24?

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+This KEP proposes to introduce a mechanism to notify a selected set of `Pod`(s) to run pre-specified commands inlined in those `Pod`(s) specification.
+
+This is achieved by:
+* Adding an optional `Notifiers` field in [Container](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#container-v1-core) type to allow users to specify command(s)/signal(s) which can be executed/sent.
+* Introducing a core API object(`PodNotification`) to trigger a specific command/signal specified in a `Pod`.
+
+## Motivation
+
+Being able to take application consistent snapshots/backups is a requirement to protect stateful workloads in Kubernetes. It's required by many applications to temporally quiesce themselves before taking snapshots of their PersistentVolume(s) and unquiesce afterwards to ensure that an application consistent snapshot/backup can be creatd. A mechanism to send a quiesce/unquiesce command to `Pod`(s) is needed.
```
chenxuc, line 79: `snapshot/backup can be creatd` change to:
`snapshot/backup can be created`

keps/sig-node/1977-container-notifier/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+8. Upper level controller is responsible to delete all `PodNotification` objects it has created after the commands have completed.
+
+9. Upper level controller is responsible to handle retries if a command fails by creating another `PodNotification` object.
+
+### Example Workflow with SIGHUP (Phase 2)
+
+This example involves a signal so it will be in phase 2 and beyond.
+
+For example, a user wants to send SIGHUB to a container in a `Pod`. This signal is defined in the `ContainterNotificationHandler` inside the `Container`.
```
chenxuc, line 646: `wants to send SIGHUB` change to:
`wants to send SIGHUP`

chenxuc (commented):
Found some typos. Thanks.

thockin:
I don't suppose this is going anywhere in 1.25?

xing-yang:
> I don't suppose this is going anywhere in 1.25?

@thockin Unfortunately no.  We still have those unresolved comments.  Xiangqian and I need to sync up about them.

thockin:
I don't suppose anyone is on this for 1.26?

xing-yang:
> I don't suppose anyone is on this for 1.26?

Thanks @thockin for the ping.  Unfortunately we are not ready to target 1.26.

keps/sig-node/1977-container-notifier/kep.yaml
```diff
@@ -0,0 +1,48 @@
+title: Container notifier
+kep-number: 1977
+authors:
+  - "@xing-yang"
+  - "@yuxiangqian"
+owning-sig: sig-node
+participating-sigs:
+  - sig-storage
+status: implementable
```
sftim, line 9: To help this move forward, how about we try to merge it with
```suggestion
status: provisional
```

and then refine from there?

thockin:
That's a good idea - @xing-yang any objections?

keps/sig-node/1977-container-notifier/kep.yaml
xing-yang: Sure.  We could try that.

xing-yang:
> That's a good idea - @xing-yang any objections?

@thockin Sure, let's try that.  Thanks.

thockin:
Can you make the proposed change and drop the PRR file?  Then we can merge as provisional.

xing-yang:
> Can you make the proposed change and drop the PRR file? Then we can merge as provisional.
Hi @thockin, I've dropped the PRR file and changed the status to "provisional".

thockin (commented):
Thanks!

/lgtm
/approve

thockin:
@dchen1107 can you approve this as "provisional" ?  Nobody is working on it but I don't want to lose state.

thockin:
I'm just going to close this and if anyone ever gets time, we can reopen

PR #2001, opened 2020-09-21
Add node shutdown KEP

bobbypage:
Enhancement issue: https://github.com/kubernetes/enhancements/issues/2000

### DIFF
--- a/keps/sig-node/2000-graceful-node-shutdown/README.md
+++ b/keps/sig-node/2000-graceful-node-shutdown/README.md
@@ -94,6 +94,8 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
@@ -530,15 +532,16 @@ Below are some examples to consider, in addition to the aforementioned [maturity
 
 #### Alpha -> Beta Graduation
 
-*   Addresses feedback from alpha testers
-*   Sufficient E2E and unit testing
+- Gather feedback from developers and surveys
+- Complete features A, B, C
+- Tests are in Testgrid and linked in KEP
 
 #### Beta -> GA Graduation
 
-*   Addresses feedback from beta
-*   Sufficient number of users using the feature
-*   Confident that no further API / kubeletConfig configuration options changes are needed
-*   Close on any remaining open issues / bugs
+- N examples of real-world usage
+- N installs
+- More rigorous forms of testing—e.g., downgrade tests and scalability tests
+- Allowing time for feedback
 
 **Note:** Generally we also wait at least two releases between beta and
 GA/stable, because there's no opportunity for user feedback, or even bug reports,
@@ -556,6 +559,18 @@ in back-to-back releases.
 
 [conformance tests]: https://git.k8s.io/community/contributors/devel/sig-architecture/conformance-tests.md
 -->
+
+#### Alpha -> Beta Graduation
+
+* Addresses feedback from alpha testers
+* Sufficient E2E and unit testing
+
+#### Beta -> GA Graduation
+
+* Addresses feedback from beta
+* Sufficient number of users using the feature
+* Confident that no further API / kubelet config configuration options changes are needed
+* Close on any remaining open issues & bugs
 
 ### Upgrade / Downgrade Strategy
 
--- a/keps/sig-node/2000-graceful-node-shutdown/README.md
+++ b/keps/sig-node/2000-graceful-node-shutdown/README.md
@@ -1,0 +1,866 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-2000: Graceful Node Shutdown
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Background on Linux Shutdown](#background-on-linux-shutdown)
+  - [Background on Inhibitors](#background-on-inhibitors)
+  - [Implementation](#implementation)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+Kubelet should be aware of node shutdown and trigger graceful shutdown of pods
+during a machine shutdown.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Users and cluster administrators expect that pods will adhere to expected pod
+lifecycle including pod termination. Currently, when a node shuts down, pods do
+not follow the expected pod termination lifecycle and are not terminated
+gracefully which can cause issues for some workloads. This KEP aims to address
+this problem by making the kubelet aware of the underlying node shutdown.
+Kubelet will propagate this signal to pods ensuring they can shutdown as
+gracefully as possible.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+*   Make kubelet aware of underlying node shutdown event and trigger pod
+    termination with sufficient grace period to shutdown properly
+*   Handle node shutdown in cloud-provider agnostic way
+*   Introduce minimal shutdown delay in order to shutdown node soon as possible
+    (but not sooner)
+*   Focus on handling shutdown on systemd based machines
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+*   Let users modify or change existing pod lifecycle or introduce new inner
+    pod depencides / shutdown ordering
+*   Support every linux init and ACPI event handling mechanism (focus on widely
+    used logind from systemd)
+*   Provide guarantee to handle all cases of graceful node shutdown, for
+    example abrupt shutdown or sudden power cable pull can’t result in graceful
+    shutdown
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+*   As a cluster administrator, I can configure the nodes in my cluster to
+    allocate X seconds for my pods to terminate gracefully during a node
+    shutdown
+
+#### Story 2
+
+*   As a developer I can expect that my pods will terminate gracefully during
+    node shutdowns
+
+### Background on Linux Shutdown
+
+In the context of this KEP, shutdown is referred to as shutdown of the
+underlying machine. On most linux distros shutdown can be initiated via a
+variety of methods for example:
+
+1. `shutdown -h now`
+2. `shutdown -h +30` `#schedule a delayed shutdown in 30mins`
+3. `systemctl poweroff`
+4. Physically pressing the power button on the machine
+5. If a machine is a VM, the underlying hypervisor can press the “virtual”
+   power button
+6. For a cloud instance, stopping the instance via Cloud API, e.g. via `gcloud
+   compute instances stop`. Depending on the cloud provider, this may result in
+   virtual power button press by the underlying hypervisor.
+
+
+Some of these cases will involve the machine receiving an ACPI event to change
+the power state. The machine can go from G0 (working state) to G2 (Soft Off)
+and finally to G3 (Off) [more info on ACPI
+states](https://en.wikipedia.org/wiki/Advanced_Configuration_and_Power_Interface).
+On Linux, prior to shutdown usually a system demon will listen to these events
+and perform some series of actions prior to userspace calling the
+[reboot(2)](https://man7.org/linux/man-pages/man2/reboot.2.html) systemcall
+with `LINUX_REBOOT_CMD_POWER_OFF` or `LINUX_REBOOT_CMD_HALT` to actually
+shutdown the machine.
+
+Historically, ACPI events were often handled by the
+[acpid](https://wiki.archlinux.org/index.php/Acpid) daemon which uses a variety
+of mechanisms to watch ACPI events (i.e. reading `/proc/acpi/event` or
+`/dev/input/eventX` to react to power button presses). However, in most modern
+linux distros today,
+[systemd-logind](https://www.freedesktop.org/wiki/Software/systemd/logind/)
+has taken over as the main component [reacting to ACPI
+events](https://wiki.archlinux.org/index.php/Power_management#ACPI_events) and
+initiating shutdown of the machine. On a system with
+systemd-logind, for example, a trigger of the power button will
+result in the systemd target
+[poweroff](https://www.freedesktop.org/software/systemd/man/systemd-halt.service.html)
+being run (see
+[HandlePowerKey](https://www.freedesktop.org/software/systemd/man/logind.conf.html),
+which will terminate all the systemd services running on the machine and
+eventually shut it down. However, in the context of kubernetes, systemd is not
+aware of the pods and containers running on the machine and systemd will simply
+kill them as regular linux processes.
+
+### Background on Inhibitors
+
+`systemd-logind` provides the ability for applications to delay shutdown and
+perform some series of actions before the shutdown completes through a
+mechanism called ["Inhibitor
+Locks".](https://www.freedesktop.org/wiki/Software/systemd/inhibit/)
+Applications can request to delay shutdown by taking an inhibitor lock by
+sending messages to logind over dbus. Applications can request up to
+`InhibitDelayMaxSec` (a setting configured in `logind.conf`) for delay based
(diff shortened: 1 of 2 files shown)


keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+    *   Mitigation: Kubelet does not provide graceful shutdown to pods (same as
+        today’s existing behavior)
+*   Kublet is unable to update `InhibitDelayMaxSec` in logind to match that of
+    `kubeletConfig.ShutdownGracePeriod`
+    *   If there are multiple logind configuration file overrides in
+        `/etc/systemd/logind.conf.d/`, logind will use the config file with the
+        lexicographically latest name. As a result in rare cases, the kubelet’s
+        InhibitDelayMaxSec conf file override may be overwritten by another
+        config file (possibly placed by another service on the machine).
+    *   Mitigation: Kubelet will use current value of `InhibitDelayMaxSec` from
```
karan, line 421: Shouldn't we then name it for example `00-kubelet.conf`? IOW, if the kubelet is expected to manage pod/machine shutdown, then it seems reasonable for the kubelet to (best-effort) take precendence.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+        logind as the shutdown period which may be less than
+        `kubeletConfig.ShutdownGracePeriod`.
+*   OS / Distro does not use systemd 183
```
karan, line 424: "at least"

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+    *   Mitigation: Kubelet will not provide graceful shutdown to pods (same as
+        today’s existing behavior).
+
+## Design Details
+
+<!--
```
karan, line 430: I think you want to move/add details about kubelet config API changes here.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+*   Unit tests for kubelet of handling shutdown event
+*   New E2E tests to validate node graceful shutdown (note limitation that K8S
```
karan, line 458: What cloud platforms are you going to test these on?

What systemd versions are you going to test on (likely only the ones provided by base images)?

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+
+### Background on Inhibitors
+
+`systemd-logind` provides the ability for applications to delay shutdown and
+perform some series of actions before the shutdown completes through a
+mechanism called ["Inhibitor
+Locks".](https://www.freedesktop.org/wiki/Software/systemd/inhibit/)
+Applications can request to delay shutdown by taking an inhibitor lock by
+sending messages to logind over dbus. Applications can request up to
+`InhibitDelayMaxSec` (a setting configured in logind.conf`) for delay based
```
rata, line 306: Niit-pick: I think there is a missing ` before logind.conf that makes the formatting wrong (because there is one at the end of logind.conf, closing)

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+*   Kubelet does not receive shutdown event or is able to create inhibitor lock
+    *   Mitigation: Kubelet does not provide graceful shutdown to pods (same as
+        today’s existing behavior)
+*   Kublet is unable to update `InhibitDelayMaxSec` in logind to match that of
```
rata, line 414:
```suggestion
*   Kubelet is unable to update `InhibitDelayMaxSec` in logind to match that of
```

keps/sig-node/2000-graceful-node-shutdown/kep.yaml
```diff
@@ -0,0 +1,34 @@
...
+status: provisional
+creation-date: 2020-09-21
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.19"
```
SergeyKanzhelev, line 16: should it be 1.20? (based on field description)

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+
+Users and cluster administrators expect that pods will adhere to expected pod
+lifecycle including pod termination. Currently, when a node shuts down, pods do
+not follow the expected pod termination lifecycle and are not terminated
+gracefully which can cause issues for some workloads. This KEP aims to address
+this problem by making the kubelet aware of the underlying node shutdown.
+Kubelet will propagate this signal to pods ensuring they can shutdown as
+gracefully as possible.
+
+### Goals
```
SergeyKanzhelev, line 193: may be worth mentioning for scoping - does this KEP tries to address the expedited re-scheduling of pods or notification of API server that node is about to shutdown?

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+`ShutdownGracePeriod` for the shutdown event. Kubelet will also register a
+`PrepareForShutdown` signal which will be emitted prior to the shutdown. Upon
+receiving the signal, Kubelet will have additional `ShutdownGracePeriod` time
+before the actual node will initiate the shutdown.
+
+**Handling the shutdown**
+
+Upon a shutdown occurring, Kubelet will gracefully terminate all the pods
+running on the node and taint itself with a new taint,
+`node.kubernetes.io/unschedulable` to avoid new pods being scheduled.
```
SergeyKanzhelev, line 351: will this taint removed when machine is back up after the restart?

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+*   Unit tests for kubelet of handling shutdown event
+*   New E2E tests to validate node graceful shutdown (note limitation that K8S
+    E2E tests currently only run on GCE).
```
SergeyKanzhelev, line 459: so the test would run on kind? Or actual nodes would shutdown?

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,843 @@
...
+“critical system pods”, and regular pods. Critical system pods should be
+terminated last, because for example, if the logging pod is terminated first,
+logs from the other workloads will not be captured. Critical system pods are
+identified as those that are in the `system-cluster-critical` or
+`system-node-critical` [priority
+classes](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical).
+
+Upon shutdown Kubelet will:
+
+1. Gracefully terminate all non critical system pods with a gracePeriodOverride
```
derekwaynecarr, line 365: do we need a buffer to allow the kubelet to report back to the apiserver?

derekwaynecarr:
@bobbypage @mrunalp thanks for putting this together!

I have no major issues with the proposal as-is other than mechanical questions about testing.

I am happy if we want to merge and iterate on just that section of the KEP as the other parts of the KEP all lgtm.

I will let @dchen1107 take a final pass.

/assign @dchen1107

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Reworded.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Thanks, fixed.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Fixed.

keps/sig-node/2000-graceful-node-shutdown/kep.yaml
bobbypage: Yup, you're correct, fixed.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Good point, added some more details here.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: That's a good point -- the name of the file can be set to ensure kubelet has higher chance of being selected. I'll take a closer look at the best naming here so the config has a higher chance of being selected.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: That's the goal, ideally kubelet can update the status of node and add this taint when the shutdown starts. If the node is rebooted or another node comes up with the same name, ideally the node will update the status without this taint. Still a bit TBD, figuring out how this will look like implementation wise.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: I think all of the node e2e tests run directly on GCE currently, I'm not aware if kind is used for e2e tests. It'll be a bit challenging to develop e2e tests here as for a full test, I'm guessing we'll need to perform an actual shutdown which may interfere with the test runner. Need to look into this aspect further.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: > What cloud platforms are you going to test these on?

Since the e2e tests only currently run on GCE, at least for e2e tests will be limited to GCP. We can run some manual tests on other cloud providers as a sanity check though.

> What systemd versions are you going to test on (likely only the ones provided by base images)?

As you mentioned, it'll be the same versions as provided in the base images the tests is running on.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: to clarify, are you proposing before pods are terminated we first report status to apiserver, and only then start terminating pods?

keps/sig-node/2000-graceful-node-shutdown/README.md
SergeyKanzhelev: btw, there is another thing about shutdown: https://github.com/kubernetes/kubernetes/issues/58635 . That is resulting in this KEP: https://github.com/kubernetes/enhancements/issues/551. Just for the reference on other issues in this area. Just found out about it

bobbypage:
Thanks @derekwaynecarr for taking a look! Will followup with @dchen1107 for a final pass. I put this KEP on agenda for upcoming SIG-Node, me and @mrunalp are happy to discuss this in more detail there.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+aware of the pods and containers running on the machine and systemd will simply
+kill them as regular linux processes.
+
+### Background on Inhibitors
+
+`systemd-logind` provides the ability for applications to delay shutdown and
+perform some series of actions before the shutdown completes through a
+mechanism called ["Inhibitor
+Locks".](https://www.freedesktop.org/wiki/Software/systemd/inhibit/)
+Applications can request to delay shutdown by taking an inhibitor lock by
```
dchen1107, line 304: nit: we might need to figure out the minimal systemd version for a working systemd inhibitor later.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+Thanks to @giuseppe for helping with getting systemd inhibitors working!
+
+### Implementation
+
+Introduce a new Kubelet Config setting, `kubeletConfig.ShutdownGracePeriod`,
+defaulting to 0 seconds. Upon kubelet startup,
+
+*   if the setting is greater than 0 seconds
+    *   kubelet will check with dbus current `InhibitDelayMaxSec` to check if
+        `kubeletConfig.ShutdownGracePeriod &lt;=` `InhibitDelayMaxSec`.
```
dchen1107, line 327: nit: <=

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+*   if `kubeletConfig.ShutdownGracePeriod` > `InhibitDelayMaxSec`
+    *   Kubelet will attempt to update the InhibitDelayMaxSec setting, by
+        writing a config file to `/etc/systemd/logind.conf.d/kubelet.conf`, and
+        sending a SIGHUP to logind to update the config setting to ensure that
+        the ShutdownGracePeriod from kubelet config is equal to
+        `InhibitDelayMaxSec`.
```
dchen1107, line 333: Reconfiguring the node through kubelet config is ok with me, but not sure if there are concerns from the community. It is worth mentioning at SIG Node meeting.

keps/sig-node/2000-graceful-node-shutdown/README.md
dchen1107: I also hope we can have the status back to the apiserver, but hope the status include all pods are terminated or being terminated. Not sure if we can guarantee it though.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+identified as those that are in the `system-cluster-critical` or
+`system-node-critical` [priority
+classes](https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/#marking-pod-as-critical).
+
+Upon shutdown Kubelet will:
+
+1. Gracefully terminate all non critical system pods with a gracePeriodOverride
+computed as `min(podSpec.terminationGracePeriodSeconds, ShutdownGracePeriod)`
+2. Gracefully terminate all critical system pods with gracePeriodOverride of 2
+seconds
```
dchen1107, line 368: Are we sure 2 seconds are enough here?

dchen1107: I saw you mentioned 2 second below, but since we are redesign the shutdown logic here, want to make sure that is enough for the critical system pods especially more and more system pods started to checkpoint.

dchen1107:
Overall looks good to me with some nits and questions.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+result in the systemd target
+[poweroff](https://www.freedesktop.org/software/systemd/man/systemd-halt.service.html)
+being run (see
+[HandlePowerKey](https://www.freedesktop.org/software/systemd/man/logind.conf.html),
+which will terminate all the systemd services running on the machine and
+eventually shut it down. However, in the context of kubernetes, systemd is not
+aware of the pods and containers running on the machine and systemd will simply
+kill them as regular linux processes.
+
+### Background on Inhibitors
```
marosset, line 298: Windows has similar functionality to inhibitor locks.
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-shutdownblockreasoncreate?redirectedfrom=MSDN is the API call to delay a shutdown.

Can we make the specifics of how kubelet controls the shutdown delay configurable for different platforms?

marosset: I can create a Background on Windows Shutdown section for inclusion in the KEP if that would be useful.

dchen1107:
We discussed this KEP at today's SIG node meeting. A couple more feedback raised at the meeting:

* Suggested by @marosset: Windows has similar APIs to register for shutdown events and delay shutdown. I'd love to see Windows support added here as well but not sure we'll have resources to work on this at the same time as the Linux implementations. I'll add links to the windows APIs to the doc PR tho.

We discussed it briefly, and we should include Windows Node support in KEP, but not alpha blocker here.

* Suggested by me in addition to two issues I raised above (1) 2s for critical system pods 2) reset node the config through kubelet's config): Instead of introducing the new taint, and new node condition, Kubelet could send node status by marking the node NotReady with detail message like: the node is shutting down.

dchen1107:
@bobbypage I approved your KEP for now. Please address all above comments and ping me for another review.

/approve

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,864 @@
...
+  control to kubelet
+    * Set Controller dbus property for the container scopes and set
+      `After=kubelet.service` for the containers. Systemd would then signal the
+      kubelet over dbus to delegate the container scope termination. This
+      requires more work in the kubelet and is also coupled to systemd and the
+      systemd cgroup driver.
+* Don’t handle node shutdown events at all, and have users drain nodes before
+  shutting them down.
+    * This is not always possible, for example if the shutdown is controlled by
+      some external system (e.g. Preemptible VMs).
```
andrewsykim, line 849: Does "Preemptible VMs" in all cloud providers trigger the shutdown event on termination?

andrewsykim: (assuming the OS for the VM is running a compatible systemd version)

bobbypage: That's a good question -- every cloud provider will eventually have to shutdown the VM and terminate it, so at some point the VM shutdown should be sent. 

For example, on GCE when a GCE Preemptible VM is terminated it gets 30 second time to shutdown, and the shutdown event is delivery at t-30 where t is when it will be forced shutdown. On AWS spot instances, the period is 2 minutes, but I'm unclear if the shutdown event is delivered a t-2min or at t itself. In addition to the systemd shutdown event, each cloud provider usually has a specific metadata server local to the VM that can be used to poll for preemption events. 

If the specific cloud provider doesn't actually trigger shutdown prior to the VM getting preempted, one workaround would be to have a cloud specific daemonset deployed that can poll the metadata server of the VM for preemption event and upon receiving the terminating event, simply trigger a node shutdown, i.e. `systemctl poweroff`, which will initiate kubelet graceful shutdown.

bobbypage:
Thanks @dchen1107 and @derekwaynecarr and rest of the folks from SIG-Node today for your feedback on this proposal. I will followup to address the comments here and would like to merge this and mark it implementable, so we can get the KEP in before the enhancements freeze for 1.20 which is October 6.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Updated with minimal systemd version:

> Inhibitor Locks were introduced to [systemd 183](https://lwn.net/Articles/499480/) (released in 2012).

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Fixed.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: We discussed this in SIG-Node today, I didn't hear any strong concerns from the community, so I think this should be fine for now.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: We discussed this during SIG-Node, the feedback sounded like 2 seconds may or may not work for various use cases, so I think it would makes sense to make this configurable via KubeletConfig. 

Unfortunately, this would result in two fields that need to be exposed in Kubelet Config: 

1. total shutdown grace period duration, `ShutdownGracePeriod`
2. shutdown grace period duration for critical pods, `ShutDownGracePeriodCritical` 

with the assumption that `ShutDownGracePeriodCritical` <  `ShutdownGracePeriod`

However, it sounds like this is worth doing, so I will update the KEP with this.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Thanks @marosset, it would be great to add a Windows specific section, thanks for offering your help!

I'm thinking it would be good to merge the initial version of the KEP and then you can do a followup PR to add the windows specific info, WDYT?

marosset: Sure, I can make a followup PR with the windows specific issue.

kikisdeliveryservice (commented):
> I will followup to address the comments here and would like to merge this and mark it implementable, so we can get the KEP in before the enhancements freeze for 1.20 which is October 6.

Hi there, If you want this in 1.20, you need to: Update the related Issue to get it in the milestone, add graduation criteria (alpha, beta, etc) and mark this as implementable.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,866 @@
...
+
+1. Gracefully terminate all non critical system pods with a gracePeriodOverride
+computed as `min(podSpec.terminationGracePeriodSeconds, ShutdownGracePeriod)`
+2. Gracefully terminate all critical system pods with gracePeriodOverride of 2
+seconds
+
+Kubelet will use the same existing
+[killPod](https://github.com/kubernetes/kubernetes/blob/release-1.19/pkg/kubelet/pod_workers.go#L292)
+function to perform the termination of pods, using `gracePeriodOverride` to set
+the appropriate grace period. During the termination process, normal [pod
```
rata, line 375: Hi! Rodrigo here, we discussed the coordination with sidecar KEP. How are you doing? :)

As part of this KEP or sidecar we need to make a change to honor `gracePeriodOverride` ([current code][link] waits for pod.TermiantionGracePeriod to run the pre-stop hooks). Can you do it yourself (should be like 1 line change :))? It is inlikely that the sidecar KEP will be ready for 1.20, I can help if you want, though :)


[link]: https://github.com/kubernetes/kubernetes/blob/1b4fa94c2131dd0b11b470a280b967b0b10ffe2d/pkg/kubelet/kuberuntime/kuberuntime_container.go#L604-L626

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,866 @@
...
+termination](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)
```
rata, line 376: The `killContainer()` function (called from `killPod()`) may take longer than `gracePeriodOverride` if the pre-stop hooks take too long (see [here][link]). It may take 2 seconds more (without much precision, it only takes into account the waiting time, not the go-code executed. Is probably negligible for more use cases, not sure if some exotic use cases, like running on 5G on a train or those need more precision).

I guess this is completely fine for alpha and if during testing we detect several environments not finishing all pods correctly then we can see how to improve it. WDYT?


[link]: https://github.com/kubernetes/kubernetes/blob/1b4fa94c2131dd0b11b470a280b967b0b10ffe2d/pkg/kubelet/kuberuntime/kuberuntime_container.go#L627-L630

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,866 @@
...
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+*   Kubelet does not receive shutdown event or is able to create inhibitor lock
+    *   Mitigation: Kubelet does not provide graceful shutdown to pods (same as
+        today’s existing behavior)
```
rata, line 415: I posted some comments on some edge cases where it might be possible to not properly kill all pods before shutdown. So, it can have similar effect to this risk, that's why I'm commenting here.

Maybe something worth doing to measure if we are effectively shutting down the pods before shutdown is to log after all pods were terminated (with fsync or some guarantees for it to be written to disk?), so if the log entry is missing we know something is not working as expected.

I think a simple log line might do the trick and help with bug reports (I'd not do anything more complicated unless really :)).

What do you think?

rata (commented):
@bobbypage Thanks for the KEP, LGTM. Added some simple comments and suggestions, but nothing major :)

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Thanks Sergey for pointing that out. Doesn't look like there is any activity on that KEP from October 2019, so will resolve this for now.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Thanks Mark, will ping you after to get the windows specific info added.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Updated the KEP and changed from suggesting 2 seconds to instead provide a configurable duration via a new KubeletConfig option, `ShutdownGracePeriodCriticalPods`.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: I updated this to include that the upon shutdown, kubelet should also update it's Ready status to false with the reason `Node is shutting down`. 

Regarding the pod status being reported back as well, I think this would happen as part of the regular sync loop, but it might be difficult to guarantee depending on the timing as Dawn mentioned.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Good to hear from you Rodrigo, hope you're doing well!

Thanks for in pointing that, out added a note to KEP, looks like https://github.com/kubernetes/kubernetes/issues/92432 is the corresponding issue. I think it makes to change it as part of this KEP, since this it'll impact the behavior here and for sidecar proposal.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Thanks for pointing that out, that's a good point. I agree, I don't think it should cause any major issue if the gracePeriod ends up being +- 2 seconds, but makes sense to reevaluate post alpha as you mentioned.

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: That's a good point, in my testing I saw that I also had the logs I could fetch after shutdown and could see if the pods existing successfully, so I'm not sure if explicit `fsync` is needed. But I like your idea of adding a final log statement, like "ShutdownComplete", or something along those lines so it's possible to verify if the shutdown process from kubelet perspective was completed successfully.

bobbypage: Added a note to the KEP about that in this section.

bobbypage:
I've updated the KEP based on the feedback so far (changed to use ReadyStatus and option to configure grace period critical pods) as mentioned in https://github.com/kubernetes/enhancements/pull/2001#issuecomment-701097113 . 

I've also updated the KEP and corresponding enhancement issue (https://github.com/kubernetes/enhancements/issues/2000) to implementable status targeting 1.20 as discussed during the SIG-Node meeting. Please let me know if there's any other concerns. 

Pinging @dchen1107 for final approval.

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,880 @@
...
+    *   Shutdown grace period unspecified, feature is not active
+    *   Pod’s ExecStop and SIGTERM handlers are given gracePeriodSeconds for
+        case when gracePeriodSeconds <= kubeletConfig.ShutdownGracePeriod
+    *   Pod’s ExecStop and SIGTERM handlers are given
+        kubeletConfig.ShutdownGracePeriod for case when gracePeriodSeconds >
+        kubeletConfig.ShutdownGracePeriod
+
+### Graduation Criteria
+
+<!--
```
kikisdeliveryservice, line 506: Need to add something here for alpha in 1.20

keps/sig-node/2000-graceful-node-shutdown/kep.yaml
```diff
@@ -0,0 +1,34 @@
...
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.20"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.21"
```
kikisdeliveryservice, line 20: v1.20 ^^^

kikisdeliveryservice (commented):
Noted 2 things from an enhancements team POV

keps/sig-node/2000-graceful-node-shutdown/README.md
bobbypage: Added.

keps/sig-node/2000-graceful-node-shutdown/kep.yaml
bobbypage: Updated.

dchen1107:
@bobbypage thanks for addressing our comments except Windows specific ones. Mark agreed to send the followup PRs to the KEP later.

/lgtm
/approve

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,895 @@
...
+message `Node Shutting Down`, thereby ensuring new workloads will not get
+scheduled to the node.
+
+Since some of the pods running on the node are often critical for the the
+workloads running on a node (e.g. logging pod daemonset, kubeproxy, kubedns)
+etc, we choose to split the pods running on the node into two categories,
+“critical system pods”, and regular pods. Critical system pods should be
+terminated last, because for example, if the logging pod is terminated first,
+logs from the other workloads will not be captured. Critical system pods are
+identified as those that are in the `system-cluster-critical` or
```
smarterclayton, line 363: I missed this when it came out, but this is super concerning to me.  I don't think these priority classes are "special" in that the Kubelet should hardcode their use.  By hardcoding these, we FORCE workloads that interact with the kubelet or other system infra pods to be in these two priority classes, which breaks the orthogonality of scheduling and resource behavior from the kubelet.

Either there needs to be something that selects for priority classes to treat "specially", or these need to be configurable at startup time in the kubelet.  The former is more flexible, the latter may be more acceptable, but would not allow a service provider to allow users to make that orthogonal.

smarterclayton: @bobbypage while reviewing the KEP (i was reading through it and thinking about the implications of the intersection with grace period when i noticed this), I think this has to be addressed before we go to beta.

bobbypage: Hi @smarterclayton 

Thanks so much for providing your feedback and comments. 

@mrunalp and I discussed this topic in length as part of KEP design and we decided to use priority class of  `system-cluster-critical` and `system-node-critical` to separate "core system workloads (e.g logging deamonsets, etc)" vs regular pods and use that information to determine shutdown ordering.

Unfortunately as I'm sure you're aware there is no existing declarative mechanism to describe pod shutdown ordering and as such we decided to use pod priority as a signal instead.  This is similar to for example pod admission and OOM score adjustment which also use `IsCriticalPod()` as a signal.

@mrunalp and I definitely agree this is not perfect, and happy to discuss if you have some alternative ideas on how to improve it and a potentially better signal for us to use to determine pod shutdown order logic instead for beta moving forward.

bobbypage:
We had a chat today with @smarterclayton @mrunalp @SergeyKanzhelev regarding some of questions in https://github.com/kubernetes/enhancements/pull/2001#discussion_r565507971 

The main item we discussed was around the current design of having two shutdown stages, first being shutting down user workloads followed by "critical node system workloads" and the current pattern of using `system-cluster-critical` and `system-node-critical` to separate system workloads vs user workloads.

Some notes from our discussion:

* Currently users who want to get their workload into the later "critical" shutdown phase need to configure their pod spec's priority class to be `system-cluster-critical` or `system-node-critical`. However, some pods may want to get into "critical" shutdown phase, but are using some other priority class.
   * Does it make sense to make configurable what priority classes are considered critical vs not instead of hardcoding  `system-cluster-critical` and `system-node-critical`  as today?

* Node Graceful shutdown currently splits the total shutdown time into two phases, first user workloads, followed by critical pods.
   * Does it make sense to provide users ability to partition the shutdown time into more than the existing two phases?  
       * Perhaps providing ability to configure a list of priority classes (or maybe range of priorities) to a shutdown phase and an associated amount of time for each phase?

SergeyKanzhelev:
I think one more small piece of feedback was to allow to use leftover of one phase to extend another. I.e. if user pods terminated very quickly, let critical pods to use the rest of the time.

bobbypage:
To circle back on https://github.com/kubernetes/enhancements/pull/2001#issuecomment-770072076 regarding supporting "custom" priority classes for node shutdown other than `system-cluster-critical` and `system-node-critical`, we discussed this point in SIG-node with some other folks and @mrunalp. 

Ultimately we decided that it's not clear how many users are actually making use of custom priority classes and would want to partition the shutdown time per specific priority class, making it a bit of niche requirement and more complicated. 

We decided to proceed to beta with the current design. If we'll get more feedback / data that supporting configurable shutdown time per "custom" priority class, we can always add this capability as followup post beta.

PR #2010, opened 2020-09-23
kepctl: fix query when KEPs are nested in directories

johnbelamaric:
Previously, it would give errors as it could not find the files. It assumed they were in the root of the SIG's KEP directory.


justaugustus:
Thanks!
/lgtm
/approve

PR #2013, opened 2020-09-24
Expose status.containerStatuses[*].imageID through Downward API

mitar:
See: https://github.com/kubernetes/enhancements/issues/2012

### DIFF
--- a/keps/sig-node/2012-expose-image-id-through-downward-api/README.md
+++ b/keps/sig-node/2012-expose-image-id-through-downward-api/README.md
@@ -1,0 +1,497 @@
+# KEP-2012: Expose status.containerStatuses[*].imageID through Downward API
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes](#notes)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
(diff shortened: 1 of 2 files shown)


keps/sig-node/2012-expose-image-id-through-downward-api/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+  beta: "v1.21"
+  stable: "v1.22"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: ImageIDDownwardAPI
+    components:
+      - kube-apiserver
+      - kube-controller-manager
```
mitar, line 35: I am not sure which components this touch, help here would be appreciated.

keps/sig-node/2012-expose-image-id-through-downward-api/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+disable-supported: true
```
mitar, line 36: What is this flag? `disable-supported`

justincormack:
If you are going to expose imageID why not expose other information, in particular for the purposes you describe, image lables could be useful too, as these may well contain (as per OCI standard labels) git commit info etc that are also useful here.

mitar:
> If you are going to expose imageID why not expose other information, in particular for the purposes you describe, image lables could be useful too, as these may well contain (as per OCI standard labels) git commit info etc that are also useful here.

I am open to that, but it is not my use case. Having imageID I can retrieve image labels at a later time. Also imageID is already readily available in Kubernetes while image labels I do not think they are? But yes, I am open to that as well.

mitar:
@mrunalp Any update on the review?

mitar:
OK, so the error is that it is still unclear who would approve it?

ehashman:
@mitar you need to add a PRR approver as well since this is net new. See https://github.com/kubernetes/community/blob/master/sig-architecture/production-readiness.md#submitting-a-kep-for-production-readiness-approval

keps/sig-node/2012-expose-image-id-through-downward-api/README.md
```diff
@@ -0,0 +1,497 @@
...
+  cluster required to make on upgrade, in order to maintain previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+### Version Skew Strategy
+
+Implementation of this is contained to only the executor component for containers,
+which has to pass on the digest of the image it is using to run the container.
+As such version skew is not anticipated to cause issues.
```
BenTheElder, line 280: Is it actually though?
IIRC pod validation happens in the api-server, but populating this downward API value would be in kubelet?

keps/sig-node/2012-expose-image-id-through-downward-api/kep.yaml
BenTheElder: kubelet?

keps/sig-node/2012-expose-image-id-through-downward-api/kep.yaml
BenTheElder: wether the feature can be disabled?

mitar:
Sorry, I do not have time to pursue this further.

PringBlocks:
how far away was this PR from getting merged? I'm interested in helping out if that's an option

mitar:
@PringBlocks I think it was mostly missing reviews. Feel free to reuse it and reopen it.

svrnm:
@PringBlocks , @mitar any interest on reviving this? I am more interested in the containerID, but I think they go well together

mitar:
I still think it would be useful, but I sadly do not have time to work on it anymore.

mitar:
Sure. I mean, you could fork the branch and create a new PR. I think that would be the cleanest.

svrnm:
> Sure. I mean, you could fork the branch and create a new PR. I think that would be the cleanest.

thanks, I will see how I can go about it

PR #2026, opened 2020-09-29
Add KEP 1923 - try nominated node first

chendave:
Signed-off-by: Dave Chen <dave.chen@arm.com>

### DIFF
--- a/keps/sig-scheduling/1923-try-nominated-node-first/README.md
+++ b/keps/sig-scheduling/1923-try-nominated-node-first/README.md
@@ -10,7 +10,6 @@
   - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
 - [Design Details](#design-details)
   - [Implementation Details](#implementation-details)
-  - [Alternatives](#alternatives)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha (v1.21):](#alpha-v121)
@@ -47,7 +46,7 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 If the scheduler fails to fit an incoming pod on any node, the scheduler will try to preempt lower
 priority pods running on a selected node to make room for the pod. The name of this node will be set
-in the pods' `pod.Status.NominatedNodeName`.
+in the pod's `pod.Status.NominatedNodeName`.
 
 The Node is called *Nominated* to indicate the intent for the Pod to be scheduled on it once preemption
 of other Pods finish. However, the `Pod.status.nominatedNodeName` information is not directly used in
@@ -97,70 +96,21 @@ nominated node.
 1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
    first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
 
-2. In case the nominated node doesn't suit for the incoming pod anymore, return `ErrNominateNode`
-   instead of `core.FitError`, because this will give scheduler a chance to clean up the nominated
-   node from the pod and find a new node to schedule instead of preemption again (might already have
-   another node available for scheduling during the period). A fresh new scheduling cycle will be
-   started later.
+2. In case the nominated node doesn't suit for the incoming pod anymore, return `err` got from `findNodesThatPassFilters`,
+   the `err` will be padded with more information to tell that scheduler is evaluating the feasibility of `NominatedNode`
+   and failed on that node.
 
-   A new error `ErrNominateNode` should be defined to describe what's going wrong on the nominated node.
+   If no error is returned and cannot pass all the filtering, this is possibly caused by the resource that claims to be
+   removed but has not been fully released yet, scheduler will continue to evaluate the rest of nodes to check if there
+   is any node already available for the coming pod.
 
-   If the nominated node doesn't suit for the pod anymore, the scheduling failure will be recorded
-   and the `updatePod` will be called, here we change the logic to update the pod as long as the
-   parameter `nominatedNode` is different with what pod holds in `pod.Status.NominatedNodeName`.
-   In this case, parameter `nominatedNode` is an empty string so that the nominated node will be
-   cleaned from the pod and the pod will be moved to the active queue. It lets scheduler find another
-   place for the pod in the next scheduling cycle.
-
-### Alternatives
-
-- Should keep trying on the nominated node in case the failure of scheduling?
-
-  This is the case when the pod deletion is still on the fly, the deletion of preemptor pods has
-  been triggered and sent to apiserver but has not actually been deleted by `kubelet` or container runtime.
-
-  Here are several things we need to consider, and this is why this approach is not adopted,
-
-  1. Keep trying in this scheduling cycle until the deletion is done
-
-     This will block the scheduler and we never know when the deletion will be done, something might
-     block this for a long time, for example, docker service is down and cannot get recovered.
-
-  2. Reserve the `pod.Status.NominatedNodeName` for the preemptor pod, so that the nominated node will be
-     tried in the following scheduling cycle (not clean up the `nominatedNode` on failure)
-
-     This will not resolve the issue mentioned above either, this will generate an infinite looping on the
-     nominated node.
-
-  3. There are other cases should be considered beside the pod deletion, which cause the nominated node
-     not able to fit for the preemptor anymore, for example, nominated node becomes unschedulable, another
-     node in the cluster releases enough room for the coming pod, topology update due to pod deletion on another
-     node which makes the nominated node not fits for `PodTopologySpread` filter anymore.
-
-     All those cases require us to start a fresh new scheduling cycle and find a better one instead of the
-     selected nominated node in previous cycle.
-
-
-- Should go on the preemption evaluation and try the nominated node there? update the nominated node if necessary.
-
-  In order to continue the preemption on the failure on the nominated node, scheduler should return `core.FitError`
-  so that preemption will continue.
-
-  1. [Debatable]: For the issue #3 mentioned above, assume it will continue to go on the preemption evaluation,
-     if there is another candidate node which doesn't preempt any victim pods, this node should be chosen as the
-     new nominated node, it is also true if this is done in the new scheduling cycle, the same node will be chosen
-     for both approaches, there is nearly no major difference, but the case like this looks more like should be handled
-     by the normal scheduling process instead of pod preemption phase, this is sound like anti-pattern of what is the
-     preemption designed.
-
-  2. If the nominated node doesn't fit due to the victim pods deletion is still on the fly, and nothing else is
-     changed, the nominated node will be chosen again either it goes to preemption evaluation or after the normal
-     scheduling cycle following by a preemption evaluation.
-     We got more chance to finish the deletion for the latter case, and the nominated node will be chosen
-     in the normal scheduling cycle or as the selected node in the following preemption evaluation phase.
-     pod deletion might be triggered again on that victim pod/pods if finally go to the preemption, there is no harm
-     to do that.
-     But we need to note the shorter time might be needed for the former case.
+   If scheduler still cannot find any node for the pod, scheduling will retry until matching either of the following cases,
+   - `NominatedNode` eventually released all the resource and the preemptor pod can be scheduled on that node.
+   - Another node in the cluster released enough release and pod get scheduled on that node instead.
+     [Discuss] Should scheduler clear the `NominatedNode` in this case?
+   - Resource cannot be released on the `NominatedNode` and no other candidate node could be found in the cluster, this will
+     be covered by [issue 95752](https://github.com/kubernetes/kubernetes/issues/95752).
+     
 
 ### Test Plan
 
--- a/keps/sig-scheduling/1923-try-nominated-node-first/README.md
+++ b/keps/sig-scheduling/1923-try-nominated-node-first/README.md
@@ -1,0 +1,204 @@
+# KEP-1923: Try Nominated Node First
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+- [Design Details](#design-details)
+  - [Implementation Details](#implementation-details)
+  - [Alternatives](#alternatives)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha (v1.21):](#alpha-v121)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+If the scheduler fails to fit an incoming pod on any node, the scheduler will try to preempt lower
+priority pods running on a selected node to make room for the pod. The name of this node will be set
+in the pods' `pod.Status.NominatedNodeName`.
+
+The Node is called *Nominated* to indicate the intent for the Pod to be scheduled on it once preemption
+of other Pods finish. However, the `Pod.status.nominatedNodeName` information is not directly used in
+the Pod's following scheduling attempts.
+
+Pod scheduling is split into two phases, the scheduling cycle and the binding cycle, the scheduling cycle
+primarily includes filtering and scoring.
+
+When preemption happens in a previous scheduling cycle, there is a high chance that the nominated node is
+the *only* node that satisfies the filters for the unscheduled Pod that triggered preemption.
+
+This KEP proposes to change the scheduling cycle such that nominated node of a pod is evaluated first
+and schedule the pod on that node if it fits. If the nominated node doesn't fit the pod, only then the
+scheduling cycle continues with the standard logic of evaluating the rest of the nodes in the cluster.
+
+## Motivation
+
+In real production environment, pods can have different priorites due to business needs, the preemption
+could happen to make sure higher priority pods could get scheduled.
+
+In cluster with large number of computing nodes, evaluating all nodes when scheduling a pod is time consuming.
+
+### Goals
+
+In the case where `pod.Status.NominatedNodeName` is set for an incoming pod, the scheduler will evaluate the
+nominated node first; if the nominated node doesn't fit the pod, the scheduling cycle will continue to evaluate
+the rest of the nodes in the cluster just like we do today.
+
+
+## Proposal
+
+### User Stories (Optional)
+
+Users want faster scheduling. Since it is highly likely the pod will only fit on the nominated node, the improvement
+in scheduling latency will come at negligible cost (the cost being placing the pod on a less optimal node).
+
+### Notes/Constraints/Caveats (Optional)
+
+When this feature is enabled the preemptor Pod might not be dispatched to the best candidated node in some corner case,
+e.g. another node releases the resources and becomes the best candidate while the victim pods got removed from the
+nominated node.
+
+## Design Details
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
+   first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
+
+2. In case the nominated node doesn't suit for the incoming pod anymore, return `ErrNominateNode`
+   instead of `core.FitError`, because this will give scheduler a chance to clean up the nominated
+   node from the pod and find a new node to schedule instead of preemption again (might already have
+   another node available for scheduling during the period). A fresh new scheduling cycle will be
+   started later.
+
+   A new error `ErrNominateNode` should be defined to describe what's going wrong on the nominated node.
+
+   If the nominated node doesn't suit for the pod anymore, the scheduling failure will be recorded
+   and the `updatePod` will be called, here we change the logic to update the pod as long as the
+   parameter `nominatedNode` is different with what pod holds in `pod.Status.NominatedNodeName`.
+   In this case, parameter `nominatedNode` is an empty string so that the nominated node will be
+   cleaned from the pod and the pod will be moved to the active queue. It lets scheduler find another
+   place for the pod in the next scheduling cycle.
+
+### Alternatives
+
+- Should keep trying on the nominated node in case the failure of scheduling?
+
+  This is the case when the pod deletion is still on the fly, the deletion of preemptor pods has
+  been triggered and sent to apiserver but has not actually been deleted by `kubelet` or container runtime.
+
+  Here are several things we need to consider, and this is why this approach is not adopted,
+
+  1. Keep trying in this scheduling cycle until the deletion is done
+
+     This will block the scheduler and we never know when the deletion will be done, something might
+     block this for a long time, for example, docker service is down and cannot get recovered.
+
+  2. Reserve the `pod.Status.NominatedNodeName` for the preemptor pod, so that the nominated node will be
+     tried in the following scheduling cycle (not clean up the `nominatedNode` on failure)
+
+     This will not resolve the issue mentioned above either, this will generate an infinite looping on the
+     nominated node.
+
+  3. There are other cases should be considered beside the pod deletion, which cause the nominated node
+     not able to fit for the preemptor anymore, for example, nominated node becomes unschedulable, another
+     node in the cluster releases enough room for the coming pod, topology update due to pod deletion on another
+     node which makes the nominated node not fits for `PodTopologySpread` filter anymore.
+
+     All those cases require us to start a fresh new scheduling cycle and find a better one instead of the
+     selected nominated node in previous cycle.
+
+
+- Should go on the preemption evaluation and try the nominated node there? update the nominated node if necessary.
+
+  In order to continue the preemption on the failure on the nominated node, scheduler should return `core.FitError`
+  so that preemption will continue.
+
+  1. [Debatable]: For the issue #3 mentioned above, assume it will continue to go on the preemption evaluation,
+     if there is another candidate node which doesn't preempt any victim pods, this node should be chosen as the
+     new nominated node, it is also true if this is done in the new scheduling cycle, the same node will be chosen
+     for both approaches, there is nearly no major difference, but the case like this looks more like should be handled
+     by the normal scheduling process instead of pod preemption phase, this is sound like anti-pattern of what is the
+     preemption designed.
+
+  2. If the nominated node doesn't fit due to the victim pods deletion is still on the fly, and nothing else is
+     changed, the nominated node will be chosen again either it goes to preemption evaluation or after the normal
+     scheduling cycle following by a preemption evaluation.
+     We got more chance to finish the deletion for the latter case, and the nominated node will be chosen
+     in the normal scheduling cycle or as the selected node in the following preemption evaluation phase.
+     pod deletion might be triggered again on that victim pod/pods if finally go to the preemption, there is no harm
+     to do that.
+     But we need to note the shorter time might be needed for the former case.
+
+### Test Plan
+
+Following tests will be covered or considered:
+
+- **Unit Tests**: All core changes must be covered by unit tests.
+- **Integration Tests**: Integration test will be provided if necessary, for example,
+  - enable the feature
+  - preempt the victim pods on the nominated node
+  - check pod will be scheduled on the nominated node
+- **Benchmark Tests**: A benchmark test which compares the performance before and after the change.
+  The performance improvement is visible by benchmark of `scheduling_algorithm_predicate_evaluation_seconds`.
+  Other benchmark will be created on-demand along with the code review process.
+
+
+### Graduation Criteria
+
+#### Alpha (v1.21):
+
+- [ ] New feature gate proposed to enable the feature.
+- [ ] Implementation of the new feature in scheduling framework.
+- [ ] Test cases mentioned in the [Test Plan](#test-plan).
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: TryNominatedNodeFirst
+    - Components depending on the feature gate: kube-scheduler
+
+* **Are there any tests for feature enablement/disablement?**
+  unittest will cover this.
+
+
+## Implementation History
+
+- 2020-09-29: Initial KEP sent out for review https://github.com/kubernetes/enhancements/pull/2026
--- a/keps/sig-scheduling/1923-try-nominated-node-first/README.md
+++ b/keps/sig-scheduling/1923-try-nominated-node-first/README.md
@@ -96,15 +96,17 @@ nominated node.
 1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
    first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
 
-2. In case the nominated node doesn't suit for the incoming pod anymore, return `err` got from `findNodesThatPassFilters`,
-   the `err` will be padded with more information to tell that scheduler is evaluating the feasibility of `NominatedNode`
-   and failed on that node.
+2. In case the nominated node doesn't suit for the incoming pod anymore, get `err` from `findNodesThatPassFilters` where
+   `NominatedNode` is firstly evaluated, the `err` will be padded with more information to tell that scheduler is evaluating
+   the feasibility of `NominatedNode` and failed on that node.
 
-   If no error is returned and cannot pass all the filtering, this is possibly caused by the resource that claims to be
-   removed but has not been fully released yet, scheduler will continue to evaluate the rest of nodes to check if there
-   is any node already available for the coming pod.
+   If no error is returned but `NominatedNode` cannot pass all the filtering, this is possibly caused by the resource that
+   claims to be removed but has not been fully released yet.
 
-   If scheduler still cannot find any node for the pod, scheduling will retry until matching either of the following cases,
+   For both of above cases, scheduler will continue to evaluate the rest of nodes to check if there is any node already
+   available for the coming pod.
+
+   Scheduler will retry until matching either of the following cases,
    - `NominatedNode` eventually released all the resource and the preemptor pod can be scheduled on that node.
    - Another node in the cluster released enough release and pod get scheduled on that node instead.
      [Discuss] Should scheduler clear the `NominatedNode` in this case?
--- a/keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
+++ b/keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
@@ -1,0 +1,27 @@
+title: Try Nominated Node First
+kep-number: 1923
+authors:
+  - "@chendave"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-scheduling
+status: provisional
+creation-date: 2020-09-29
+reviewers:
+  - "@alculquicondor"
+  - "@Huang-Wei"
+  - "@ahg-g"
+approvers:
+  - "@Huang-Wei"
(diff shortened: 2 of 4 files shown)


keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
+# KEP-NNNN: Try Nominated Node First
```
alculquicondor, line 1: 1923 :)

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - sig-scheduling
+status: implementable
+creation-date: 2020-09-29
+reviewers:
+  - "@sig-scheduling-leads"
+approvers:
+  - "@sig-scheduling-leads"
+see-also:
+  - "/keps/sig-scheduling/1923-try-nominated-node-first"
+stage: alpha
```
alculquicondor, line 16: I think we could target Beta directly, but that's a question for approvers.

You would need to complete the Production Readiness questionnaire in that case.

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,26 @@
...
+kep-number: 1923
+authors:
+  - "@chendave"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-scheduling
+status: implementable
+creation-date: 2020-09-29
+reviewers:
+  - "@sig-scheduling-leads"
```
alculquicondor, line 11: use specific usernames

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
```
alculquicondor, line 26: You are already marking the KEP as implementable, so this should be checked. Or you could do that in a follow up PR if you want to have a first part merged first.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+If the incoming pod that going to be scheduled has the priority defined,
+and there is no available nodes in the cluster match with the needs for
```
alculquicondor, line 48:
```suggestion
and there are no nodes in the cluster that satisfy the scheduling filters for the incoming Pod
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+the pod to be scheduled, scheduler will try to preempt lower priority pods
+running on each of the nodes in the cluster to release enough room for the
```
alculquicondor, line 50:
```suggestion
running on a selected node in the cluster to make room for the
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+pod to be scheduled, the selected node for the incoming pod after the
+preemption will be set as the value of `pod.Status.NominatedNodeName`.
+The nominated node here implies the node that is selected for the incoming
```
alculquicondor, line 53: This sounds like it is the case today, but rather this is what the KEP is going to implement.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+pod to be scheduled.
+
+Pod scheduling is split into two phases, the scheduling cycle and the binding
+cycle, the scheduling cycle primarily includes predicate and scroing.
```
alculquicondor, line 57:
```suggestion
cycle, the scheduling cycle primarily includes filtering and scoring.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+When the preemption is happened, the phase of predicate for the nominated node
```
alculquicondor, line 59: When preemption happened in a previous scheduling cycle, there is a high chance that the nominated node is the *only* node that satisfies the filters for the incoming Pod to be scheduled.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+has been properly measured, there is a high chance the selected nominated node
+is still the *only* one candidate for the incoming pod to be scheduled.
+
+This KEP proposes to validate the availability of the nominated node directly
```
alculquicondor, line 63:
```suggestion
This KEP proposes to validate the availability of the nominated node
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Motivation
+
+In real production environment, pods can have different priorites due to business
+needs, the preemption could happen to make sure high priority pods could got scheduled
+and spawned.
+
+For the cluster with huge number of computing nodes, looping over all the nodes in
+the cluster for the phase of the predicate is an unnegligible consuming for scheduler,
+besides, it's unpredicatable that how the out-of-tree plugins are defined, whether
```
alculquicondor, line 75: we don't need to mention out-of-tree plugins, the previous phrase should be enough.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+they are time consuming, this is also a factor that may increase the time for the
+phase of predicate.
+
+### Goals
+
+- In the case of `pod.Status.NominatedNodeName` is set for the incoming pod, scheduler
+  will try the nominated node first for the phase of predicate.
```
alculquicondor, line 82: filtering phase.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+- Failed to schedule on the nominated node will fall back to verify all the nodes in
```
alculquicondor, line 83: this shouldn't be a separate goal. Phrase it as part of the previous point.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+  the cluster instead of preemption again.
+- A new feature gate will be provided for the operator to opt in/out this feature.
```
alculquicondor, line 85: While in beta, but that's normal for any feature. Not sure if we should mention it.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+### User Stories (Optional)
+
+As a premium user, I want my high priority application got spin up more quickly,
+I don't care where the application is running, whether there is any other nodes
+in the cluster have more disk space, memory size etc., as long as my application
+run normally.
+
+### Notes/Constraints/Caveats (Optional)
+
+Enables this feature also implies the preemptor pod might not be dispatched
```
alculquicondor, line 98: "When this feature is enabled the preemptor Pod"

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+to the best candidated node in some corner case, e.g. another node releases
+the resources and becomes the best candidate while the victim pods got removed
+from the nominated node, but enable this feature will make scheduler to consider
+the nominated node firstly.
```
alculquicondor, line 102: the last sentence seem unnecessary.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Design Details
+
+### Implementation Details
+
+1. In the phase of predicate, which is currently implemented in the method of `findNodesThatFitPod`,
```
alculquicondor, line 108: filtering phase :)

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+   check the nominated directly if the incoming pod has the `pod.Status.NominatedNodeName` defined
+   and the feature gate is enabled.
+2. In case the nominated node is not suit the incoming pod anymore, return `fmt.Errorf` instead of
```
alculquicondor, line 111: This shouldn't be like that.

Let's say the nominated Node had 2 Pods preempted and one was removed. At this point, kube-scheduler might try to schedule the Pod and fail. But in this case we should keep trying.

Also, errors are supposed to signal internal errors, meaning that something went wrong internally. This doesn't qualify.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+   `core.FitError`, because this will give scheduler a chance to clean up the nominated node from
+   the pod and find a new node to schedule instead of preemption again (might already have during
+   the period). i.e. a fresh new scheduling cycle will be started from here.
+3. If the nominated node is not suit for the pod anymore, the scheduling failure will be recorded
+   and the `updatePod` will be called, here we change the logic to update the pod as long as the
+   parameter `nominatedNode` is different with what pod holds in `pod.Status.NominatedNodeName`.
+   In this case, parameter `nominatedNode` is an empty string so that the nominated node will be
+   cleaned from the pod and the pod will be moved to the active queue. It lets scheduler find another
+   place for the pod.
```
alculquicondor, line 120: in the next scheduling cycle

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+### Test Plan
+
+Following tests wll be covered or considered:
+
+- **Unit Tests**: All core changes must be covered by unit tests.
+- **Integration Tests**: Integration test will be provided if necessary, not visible in current stage.
```
alculquicondor, line 128: Do a verbal description of what the test cases are supposed to look like.

We need to test for one preempted pod and several preempted pods as well.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
alculquicondor: Additionally, if the Pod doesn't fit in the nominated Node, aren't we supposed to iterate over the rest of the Nodes? Please add those details.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: DONE

chendave:
@alculquicondor really appreciate for your thoroughly review and fantastic review comments!

I keep some of them open so we can continue the discussion, and btw, my response for the next several days will be slow due to the holiday here.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
alculquicondor: You have to state the problem:

"The Node is called Nominated to indicate the intent for the Pod to be scheduled on it once preemption of other Pods finish. However, the `Pod.status.nominatedNodeName` information is not directly used in the following scheduling cycles."

keps/sig-scheduling/1923-try-nominated-node-first/README.md
alculquicondor: Maybe we can go onto a preemption evaluation cycle again and try the nominated Node first. There is a chance that the same Node will still be the nominated one. But if not, it gets changed.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
alculquicondor: Yes, integration and benchmark.

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,30 @@
...
+participating-sigs:
+  - sig-scheduling
+status: implementable
+creation-date: 2020-09-29
+reviewers:
+  - "@alculquicondor"
+  - "@Huang-Wei"
+  - "@ahg-g"
+approvers:
+  - "@alculquicondor"
```
alculquicondor, line 15: I'm not a KEP approver :)

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - "@Huang-Wei"
+  - "@ahg-g"
+see-also:
+  - "/keps/sig-scheduling/1923-try-nominated-node-first"
+stage: alpha
+latest-milestone: "v1.21"
```
alculquicondor, line 21: You don't think we can make it in 1.20?

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Hi Aldo,  I seriously considered those approaches and add a section to explain why it is designed as this, pls take a look and like to see your feedback.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: I went through the integration test, I don't think the integration test is needed, since I cannot see any difference of expected result when this feature is enabled. 
And `scheduler_perf` already has the performance testcase to cover this, which is `scheduling_algorithm_predicate_evaluation_seconds`. I am not sure more testcases are needed or not.
Anyway, I have some placeholder created in this section.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: removed

chendave:
Courteously ping @alculquicondor , can we move forward this KEP?

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,29 @@
+title: Try Nominated Node First
+kep-number: 1923
+authors:
+  - "@chendave"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-scheduling
+status: implementable
```
alculquicondor, line 8: could you leave this as provisional for now?

keps/sig-scheduling/1923-try-nominated-node-first/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+creation-date: 2020-09-29
+reviewers:
+  - "@alculquicondor"
+  - "@Huang-Wei"
+  - "@ahg-g"
+approvers:
+  - "@Huang-Wei"
+  - "@ahg-g"
+see-also:
+  - "/keps/sig-scheduling/1923-try-nominated-node-first"
```
alculquicondor, line 18: why are you self referencing this KEP?

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,220 @@
...
+topology changing, victim pods deletion is still on-the-fly in the following scheduling cycle.
+Those cases will make the nominated node not fits for the coming pod anymore, and a new scheduling
+cycle is needed in those case. This is no harm since it will eventually get convergent.
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`,
+   check the nominated directly if the incoming pod has the `pod.Status.NominatedNodeName` defined
+   and the feature gate is enabled.
+2. In case the nominated node not suit the incoming pod anymore, return `ErrNominateNode` instead of
```
alculquicondor, line 119: I'm still not convinced that a single scheduling retry will be enough to discard the current nominated node. Maybe it will work if we solve kubernetes/kubernetes#94009 first

But to be clear, you are saying that if there is a nominated Node, that is the only Node we try in the scheduling cycle. Correct?

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,220 @@
...
+   `core.FitError`, because this will give scheduler a chance to clean up the nominated node from
+   the pod and find a new node to schedule instead of preemption again (might already have during
+   the period). i.e. a fresh new scheduling cycle will be started from here.
+   A new error `ErrNominateNode` should be defined to describe what's going wrong on the nominated node.
+3. If the nominated node is not suit for the pod anymore, the scheduling failure will be recorded
```
alculquicondor, line 124: The enumeration makes it feel like 3 happens after 2, but in reality 3 is giving more details on 2. I don't think we need the paragraph at all.

alculquicondor (commented):
I was actually waiting until after 1.20 release. But here go some more thoughts.

Note that I'm not a KEP approver.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

chendave:
@alculquicondor , as always, thank you for your review!

kindly ping @Huang-Wei  @ahg-g  could you please take a look? any comments for this proposal, particularly, the concern from alculquicondor .

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+If the incoming pod that going to be scheduled has the priority defined,
+and there are no nodes in the cluster that satisfy the scheduling filters
+for the incoming Pod, scheduler will try to preempt lower priority pods
+running on a selected node in the cluster to make room for the
+pod to be scheduled, the selected node for the incoming pod after the
+preemption will be set as the value of `pod.Status.NominatedNodeName`.
```
ahg-g, line 53:
```suggestion
If the scheduler fails to fit an incoming pod on any node, the scheduler will try to preempt lower priority pods
running on a selected node to make room for the pod. The name of this node will be set in the pods' `pod.Status.NominatedNodeName`.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+The Node is called *Nominated* to indicate the intent for the Pod to be
+scheduled on it once preemption of other Pods finish. However, the
+`Pod.status.nominatedNodeName` information is not directly used in the
+following scheduling cycles.
```
ahg-g, line 58:
```suggestion
Pod's following scheduling attempts.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+Pod scheduling is split into two phases, the scheduling cycle and the binding
+cycle, the scheduling cycle primarily includes filtering and scoring.
+
+When the preemption happened in a previous scheduling cycle, there is a high
```
ahg-g, line 63:
```suggestion
When preemption happens in a previous scheduling cycle, there is a high
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+chance that the nominated node is the *only* node that satisfies the filters
+for the incoming Pod to be scheduled.
```
ahg-g, line 65:
```suggestion
for the unscheduled Pod that triggered preemption.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+This KEP proposes to validate the availability of the nominated node in the
+new scheduling cycle before looping all over the nodes in the cluster for the
+phase of the filtering.
```
ahg-g, line 69:
```suggestion
This KEP proposes to change the scheduling cycle such that nominated node of a pod is evaluated first and schedule the pod on that node if it fits. If the nominated node doesn't fit the pod, only then the scheduling cycle continues with the standard logic of evaluating the rest of the nodes in the cluster.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Motivation
+
+In real production environment, pods can have different priorites due to business
+needs, the preemption could happen to make sure high priority pods could got scheduled
+and spawned.
```
ahg-g, line 75:
```suggestion
needs, preemption could happen to make sure higher priority pods get scheduled.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+For the cluster with huge number of computing nodes, looping over all the nodes in
+the cluster for the phase of the filtering is an unnegligible consuming for scheduler.
```
ahg-g, line 78:
```suggestion
In clusters with large number of nodes, evaluating all nodes when scheduling a pod is time consuming.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Goals
+
+- In the case of `pod.Status.NominatedNodeName` is set for the incoming pod, scheduler
+  will try the nominated node first for the filtering phase, failed to schedule on the
+  nominated node will fall back to verify all the nodes in the cluster instead of preemption
+  again.
```
ahg-g, line 85:
```suggestion
- In the case where `pod.Status.NominatedNodeName` is set for an incoming pod, the scheduler
  will evaluate the nominated node first; if the nominated node doesn't fit the pod, the scheduling cycle will continue to evaluate the rest of the nodes in the cluster just like we do today.
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Proposal
+
+### User Stories (Optional)
+
+As a premium user, I want my high priority application got spin up more quickly,
+I don't care where the application is running, whether there is any other nodes
+in the cluster have more disk space, memory size etc., as long as my application
+run normally.
```
ahg-g, line 94: I would phrase this as follows: Users want faster scheduling. Since it is highly likely the pod will only fit on the nominated node, the improvement in scheduling latency will come at negligible cost (the cost being placing the pod on a less optimal node).

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+This feature hopes to improve the performance with the minimal change on the existing scheduling
+framework, scheduler will short-circuit to check the nominated node directly instead of looping
+all the nodes in the cluster, and if the nominated node doesn't fit, a fresh new scheduling cycle
+will be started to get an new node for scheduling.
+
+In most situation, the new scheduling cycle is not needed, but we still got the case like the
+topology changing, victim pods deletion is still on-the-fly in the following scheduling cycle.
+Those cases will make the nominated node not fits for the coming pod anymore, and a new scheduling
+cycle is needed in those case. This is no harm since it will eventually get convergent.
```
ahg-g, line 112: I would simplify this as: 

"We will update the scheduling cycle such that nominated node of a pod is evaluated first and schedule the pod on that node if it fits. If the nominated node doesn't fit the pod, only then the scheduling cycle continues with the standard logic of evaluating the rest of the nodes in the cluster."

ahg-g: we already mentioned that two times in the summary and goals, I feel we repeated the same thing multiple times unnecessarily. So I think we can even remove this section completely and directly get into the implementation details.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`,
+   check the nominated directly if the incoming pod has the `pod.Status.NominatedNodeName` defined
```
ahg-g, line 117:
```suggestion
   check the nominated node first if the incoming pod has the `pod.Status.NominatedNodeName` defined
```

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,224 @@
...
+   and the feature gate is enabled.
+
+2. In case the nominated node doesn't suit for the incoming pod anymore, return `ErrNominateNode`
+   instead of `core.FitError`, because this will give scheduler a chance to clean up the nominated
+   node from the pod and find a new node to schedule instead of preemption again (might already have
+   another node available for scheduling during the period). A fresh new scheduling cycle will be
+   started later.
```
ahg-g, line 124: are you suggesting that the rest of the nodes will not be evaluated in the same cycle? if so, why?

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: removed.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: I changed my mind to evaluate the rest of the nodes, this paragraph will be updated shortly.

ahg-g: thanks, was this updated? I can't tell because the new patch was an amend.

alculquicondor: Keep in mind that removing the nominated node on the first failure is a source for other race conditions. For that reason, I suggest that all nodes should be evaluated and "new" nominated node calculated in the same cycle (if pod still doesn't fit).

ahg-g: right, and I am not sure we need to clear the nominatednodename if we successfully bind the pod to a node. We need to clear it only if we fail to find a node that fits and preemption fails to suggest a new one.

chendave: sorry for delay, this is not updated by far, I am still thinking about @alculquicondor 's suggestion here.  I am inclined to follow his suggestion now.

chendave: I am going to keep the `nominatedNode` untouched, there is a potential issue around the  `nominatedNode` , please see: https://github.com/kubernetes/kubernetes/issues/95752, this is the issue that we have today, and should be a separated topic instead.  I tracked the issue there and assign the issue to myself, so that we won't mess those issues together.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,204 @@
...
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+If the scheduler fails to fit an incoming pod on any node, the scheduler will try to preempt lower
+priority pods running on a selected node to make room for the pod. The name of this node will be set
+in the pods' `pod.Status.NominatedNodeName`.
```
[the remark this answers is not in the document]
chendave: Done

chendave:
@alculquicondor @ahg-g  thanks for review! this version should be simple enough.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,154 @@
...
+nominated node.
+
+## Design Details
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
+   first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
+
+2. In case the nominated node doesn't suit for the incoming pod anymore, return `err` got from `findNodesThatPassFilters`,
```
alculquicondor, line 99: I'm not sure I follow: If an error is returned, are you still suggesting to fail this scheduling cycle?

chendave: not fail the scheduling cycle, continue to try other nodes. 
the error here is the we got from `findNodesThatFitPod` which only try the nominated node firstly, and it will continue to try other nodes on failure.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
chendave: this is not a problem in the current proposal, so mark it as resolved.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+nominated node.
+
+## Design Details
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
+   first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
+
+2. In case the nominated node doesn't suit for the incoming pod anymore, get `err` from `findNodesThatPassFilters` where
```
alculquicondor, line 99: Looks reasonable now.

However, I have one suggestion. Rather than evaluating the nominated first and only the rest of the nodes if it doesn't fit, we could do those 2 steps concurrently. Otherwise we would be increasing latency by potentially 2x in the worst case scenario.

ahg-g: why 2x? it is a single node evaluated through filters only.

alculquicondor: If you have the number of Nodes equal to # of threads, it is 2x. But it's lower in general.

ahg-g: not worth the complexity in my opinion. it is not expensive to evaluate a single node through filters.

alculquicondor: We still have to run PreFilter for all nodes initially, right?
In that case, yes, it might not be worth the complexity.

ahg-g: Right, I assumed prefilter would be run once.

chendave: yes, prefilter will run only once, and the times to call those filters are not changed at all.

chendave:
Thanks!  @alculquicondor @ahg-g  if you okay this version, I will squash the commits and start to rework at this PR (https://github.com/kubernetes/kubernetes/pull/93179)

alculquicondor:
/lgtm
as provisional for 1.21

ahg-g:
/approve
/hold

hold for @Huang-Wei

chendave:
@Huang-Wei , are you good with this one?

Huang-Wei (commented):
Sry for the late review.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+If the scheduler fails to fit an incoming pod on any node, the scheduler will try to preempt lower
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+priority pods running on a selected node to make room for the pod. The name of this node will be set
+in the pod's `pod.Status.NominatedNodeName`.
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+The Node is called *Nominated* to indicate the intent for the Pod to be scheduled on it once preemption
+of other Pods finish. However, the `Pod.status.nominatedNodeName` information is not directly used in
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Pod scheduling is split into two phases, the scheduling cycle and the binding cycle, the scheduling cycle
+primarily includes filtering and scoring.
+
+When preemption happens in a previous scheduling cycle, there is a high chance that the nominated node is
+the *only* node that satisfies the filters for the unscheduled Pod that triggered preemption.
+
+This KEP proposes to change the scheduling cycle such that nominated node of a pod is evaluated first
+and schedule the pod on that node if it fits. If the nominated node doesn't fit the pod, only then the
+scheduling cycle continues with the standard logic of evaluating the rest of the nodes in the cluster.
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+## Motivation
+
+In real production environment, pods can have different priorites due to business needs, the preemption
+could happen to make sure higher priority pods could get scheduled.
+
+In cluster with large number of computing nodes, evaluating all nodes when scheduling a pod is time consuming.
+
+### Goals
+
+In the case where `pod.Status.NominatedNodeName` is set for an incoming pod, the scheduler will evaluate the
```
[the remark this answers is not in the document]
chendave: combined this suggestion and previous suggestion together.

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+When this feature is enabled the preemptor Pod might not be dispatched to the best candidated node in some corner case,
+e.g. another node releases the resources and becomes the best candidate while the victim pods got removed from the
+nominated node.
+
+## Design Details
+
+### Implementation Details
+
+1. In filtering phase, which is currently implemented in the method of `findNodesThatFitPod`, check the nominated node
+   first if the incoming pod has the `pod.Status.NominatedNodeName` defined and the feature gate is enabled.
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+   If no error is returned but `NominatedNode` cannot pass all the filtering, this is possibly caused by the resource that
+   claims to be removed but has not been fully released yet.
+
+   For both of above cases, scheduler will continue to evaluate the rest of nodes to check if there is any node already
+   available for the coming pod.
+
+   Scheduler will retry until matching either of the following cases,
+   - `NominatedNode` eventually released all the resource and the preemptor pod can be scheduled on that node.
+   - Another node in the cluster released enough release and pod get scheduled on that node instead.
```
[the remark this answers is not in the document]
chendave: Done

keps/sig-scheduling/1923-try-nominated-node-first/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: TryNominatedNodeFirst
```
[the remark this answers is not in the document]
chendave: Done

chendave:
@Huang-Wei  thanks for the review, could you pls take another look?

Huang-Wei:
Looks neat, thanks @chendave !

@alculquicondor @ahg-g One of my suggestions was to rename the feature gate from "TryNominatedNodeFirst" to "PreferNominatedNode", which I believe is more formal.

/lgtm

Feel free to cancel /hold if you agree on the new name.

alculquicondor:
yeah, that feature name sounds better to me

PR #2037, opened 2020-09-30
Adding windows privileged containers KEP

ambguo:


marosset:
Can you add a KEP.yaml as well based off of https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml

(I apologize that the other sig-windows KEPs don't have this yet, so we did not have any good examples to follow)

kikisdeliveryservice:
This is for issue #1981 Also note that the dir name is incorrect here it should be `1981-windows-privileged-container-support`

kikisdeliveryservice:
This is for issue #1981 Also note that the dir name is incorrect here it should be 1981-windows-privileged-container-support

cc: @kinarashah for review

fasaxc:
Great to see this getting off the ground; looking forward to the day when we can ship Calico as a daemonset on Windows.

marosset:
/lgtm
/appove

Merging this now in a provisional state.
We can iterate and get more reviews when we move this to implementable.

PR #2061, opened 2020-10-02
update kep to be eiligible for 1.20

zshihang:
/cc @msau42 
/cc @kikisdeliveryservice

### DIFF
--- a/keps/sig-storage/1855-csi-driver-service-account-token/README.md
+++ b/keps/sig-storage/1855-csi-driver-service-account-token/README.md
@@ -3,7 +3,6 @@
 ## Table of Contents
 
 <!-- toc -->
-
 - [Summary](#summary)
 - [Motivation](#motivation)
   - [User stories](#user-stories)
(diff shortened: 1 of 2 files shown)


keps/sig-storage/1855-csi-driver-service-account-token/README.md
```diff
@@ -173,14 +173,19 @@ Option 1 is adopted. See discussion
...
 
+#### Alpha->Beta
+
+- Deployed the feature in production and went through at least minor k8s.
+- Fixed any bugs.
+
 #### Beta->GA
 
-- Deployed the feature in production and went through at least minor k8s
+- Deployed the feature in production and went through at least minor k8s.
```
msau42, line 188: Do you want to add some stress/scale testing to make sure we can handle a lot of pods using this feature?

zshihang: done

PR #2063, opened 2020-10-02
Mark graceful node shutdown KEP as implementable

bobbypage:
* Add alpha graduation criteria
* Update implementation history

Addressing https://github.com/kubernetes/enhancements/issues/2000#issuecomment-702934662

### DIFF
--- a/keps/sig-node/2000-graceful-node-shutdown/README.md
+++ b/keps/sig-node/2000-graceful-node-shutdown/README.md
@@ -94,6 +94,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
+    - [Alpha Graduation](#alpha-graduation)
     - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
     - [Beta -&gt; GA Graduation](#beta---ga-graduation)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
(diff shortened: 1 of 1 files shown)


kikisdeliveryservice (commented):
Nice!! Thank you for the changes LGTM

mrunalp:
:+1:

keps/sig-node/2000-graceful-node-shutdown/README.md
```diff
@@ -560,6 +561,11 @@ in back-to-back releases.
 [conformance tests]: https://git.k8s.io/community/contributors/devel/sig-architecture/conformance-tests.md
 -->
 
+#### Alpha Graduation
+
+* Implemented the feature for Linux (systemd) only
+* Unit tests
```
dchen1107, line 567: If I read this correctly, you plan to mock those involved system components to add the unittest, so that you can avoid the real VM shutdown here. Am I right? I am fine with either way. But on another hand, we can build e2e tests with some fake shutdown event. We did fake some kernel issues for NPD tests in the past.

bobbypage: Thanks Dawn for pointing this out. I updated this section a bit to clarify the scope of the testing for alpha. 

My thoughts are for alpha would be to create appropriate mocks for system components (e.g. systemd, inhbitors) so we can avoid a real shutdown. For alpha graduation, I added as a goal to investigate the best way to create e2e tests for this feature, e.g. potentially by faking the shutdown event as you mentioned.

dchen1107: ACK and thanks for the updates.

PR #2072, opened 2020-10-04
Update 20200204-NetworkPolicy-verification-rearchitecture.md

jayunit100:


jayunit100:
cc @BenTheElder I added a statement to back down from the  " lets give this DSL to sig-test" statement :) lmk if that makes it digest better

PR #2074, opened 2020-10-05
Remove unresolved in manifest based webhooks KEP

vivekbagade:
Also add test plan and answer scalability questionnaire

### DIFF
--- a/keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
+++ b/keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
@@ -9,9 +9,10 @@
   - [Goals](#goals)
 - [Proposal](#proposal)
   - [Naming](#naming)
-  - [Metrics and audit annotations](#metrics-and-audit-annotations)
   - [New AdmissionConfig schema](#new-admissionconfig-schema)
   - [Reconfiguring manifest file](#reconfiguring-manifest-file)
+    - [Behavioral details of reconfiguration](#behavioral-details-of-reconfiguration)
+  - [Metrics and audit annotations](#metrics-and-audit-annotations)
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
@@ -113,25 +114,7 @@ All webhooks in the manifest file need to have unique names. If a new webhook
 configuration API object with a same name as a webhook in the manifest is added,
 both webhook would be invoked. Essentially, webhooks in the manifest file will
 be treated as belonging to a different domain from the webhooks registered
-through the API.
-
-### Metrics and audit annotations
-Metrics and audit annotations for webhooks registered through the manifest file
-will be surfaced differently. This allows the administrator to unique monitor
-activity of these webhooks as they would be the only party able to take action
-against issues with these webhooks. 
-
-All admission metrics for webhooks will have an additional label `manifest_based`
-with a value `true` for webhooks registered via manifest.
-
-Audit annotations for mutation webhooks will include another field `manifestBased`
-with a value `true` for webhooks registered via manifest.
-
-```
-<<[UNRESOLVED sig-instrumentation ]>>
-get feedback on how to indicate this in metrics
-<<[/UNRESOLVED]>>
-```
+through the API. 
 
 ### New AdmissionConfig schema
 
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -155,22 +138,66 @@ plugins:
...
-dynamically changed by editing the contents of the file.
+dynamically changed by editing the contents of the file. In addition to watching
+the file for changes, we would also read the file periodically checking for
+changes.
+
+#### Behavioral details of reconfiguration
+This section details the behavior of the mechanism when encountering non-optimal
+situations.
+
+In cases where:
```
deads2k, line 149: and if any of these are true when reading the file on a process start, does the process fail to start? I would expect it to.

deads2k:
This answers the outstanding questions that I know of.  I'm ok with this moving to implementable if we can find a solid reviewer who can sign up for the work (I'm not available).

@lavalamp any comments you'd like addressed before implementable?

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
liggitt: that matches my expectation, but should be called out explicitly

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -113,25 +114,7 @@ All webhooks in the manifest file need to have unique names. If a new webhook
...
-with a value `true` for webhooks registered via manifest.
-
-```
-<<[UNRESOLVED sig-instrumentation ]>>
-get feedback on how to indicate this in metrics
-<<[/UNRESOLVED]>>
-```
+through the API. 
 
 ### New AdmissionConfig schema
```
liggitt, line 120: I don't see this called out, but the treatment of non-absolute file paths for `webhooksFile` should be described. Absolute paths are required for the `kubeconfigFile` field in the same config, which is the simplest approach. If relative paths are allowed, I would expect them to be treated relative to the location of the admission config file, which requires replumbing some bits of admission config loading.

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -155,22 +138,66 @@ plugins:
...
-<<[/UNRESOLVED]>>
-```
+### Metrics and audit annotations
+For the administrator to uniquely monitor manifest based webhooks, two
+additional metrics would be added at [stability level `ALPHA`](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20190404-kubernetes-control-plane-metrics-stability.md#stability-classes)
+to the existing set of metrics.
+
+a. Webhook metadata: This metric would contain metadata of each webhook invoked,
+both API based and manifest based. The metadata includes the webhook type
+(mutating/validating) and if it is manifest based.
```
liggitt, line 169: detail the name and labels of the metric?

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -155,22 +138,66 @@ plugins:
...
+
+b. Manifest errors: Since the manifest file can be reconfigured post API server
+startup, we would add a metric informing users if there were errors
+loading/parsing the current manifest file.
```
liggitt, line 173: detail the name and labels of the metric?

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
```diff
@@ -155,22 +138,66 @@ plugins:
...
+
+Audit annotations for mutating webhooks would also carry an additional field
+`manifestBased` indicating the invocation of a manifest based webhook.
```
liggitt, line 176: give an example of the audit annotation including the `manifestBased` field

liggitt:
Asked for a few clarifications, then lgtm.

I can review the implementation.

keps/sig-api-machinery/1872-manifest-based-admission-webhooks/README.md
liggitt: also, add the metric to the `metrics` section of the kep.yaml file (same for all the metrics)

lavalamp:
This design has the following problems:
* Doesn't work for other types (e.g. CRDs).
* No way for status to be reported on the "virtual" resources.
* There's no longer a single source of truth.
* It's very hard for a user to figure out what the system is doing or how to change that.

So, I'm a big -1 on this design.

I'd accept a design that either
1. works for all resources, by adding a controller that reads a directory for manifests and puts them in etcd; and also configures an authz plugin to block non-system changes of those resources, OR
2. just an authz mechanism that prevents non-system changes to certain resources, plus blessing a super-super-user account that can change those resources (and then it's up to the holder of that account to fix those resources).

Honestly I think the authz addition is too heavy-handed and in the common case will prevent people from fixing clusters as frequently as it prevents them from breaking clusters; but I recognize that it is greatly desired.

liggitt:
> This design has the following problems:
>
> * Doesn't work for other types (e.g. CRDs).
> * No way for status to be reported on the "virtual" resources.
> * There's no longer a single source of truth.
> * It's very hard for a user to figure out what the system is doing or how to change that.

I'm ok with taking a step back to see if we can come up with an approach that is more coherent, and I agree there are multiple resources that have similar use cases, but some of those characteristics are features in some scenarios :-) 

* local-only policy config allows controlled rollout in a multi-server environment without worrying about another server stomping config ("my apiserver will load policy file X into etcd, then enforce whatever is present in etcd" is a way weaker guarantee than "my apiserver will enforce policy file X")
* stratifying policy specified by the cluster operator and policy specified/inspectable/changeable by the cluster user is useful in managed environments

vivekbagade:
> * No way for status to be reported on the "virtual" resources.
Webhook config does not have status field IIUC

> * There's no longer a single source of truth.
Yes, I partially agree with this. But, the source of truth moves from the API to the API server metrics which not as useful. 

> * It's very hard for a user to figure out what the system is doing or how to change that.
I don't completely agree. This mechanism is meant to provide operators with a better mechanism compared to compiled in admission control. And this method provides better visibility compared to compiled in admission control. There are audit annotations (for mutating webhooks) and metrics that help with this. As for what "users" can do to alter the behavior, this is a feature of the mechanism(extremely desirable in managed platforms) and the users would just have to reach out to their operator. 

> So, I'm a big -1 on this design.
> 
> I'd accept a design that either
> 
> 1. works for all resources, by adding a controller that reads a directory for manifests and puts them in etcd; and also configures an authz plugin to block non-system changes of those resources, OR
> 2. just an authz mechanism that prevents non-system changes to certain resources, plus blessing a super-super-user account that can change those resources (and then it's up to the holder of that account to fix those resources).
> 
> Honestly I think the authz addition is too heavy-handed and in the common case will prevent people from fixing clusters as frequently as it prevents them from breaking clusters; but I recognize that it is greatly desired.

I also want to +1 Jordan's comment on the usefulness of stratifying policy specified by the operator vs policy specified by the cluster user.

PR #2080, opened 2020-10-06
keps/NNNN-kep-template: Add note about test refactors

jayunit100:


kikisdeliveryservice (commented):
Would make sense to mention test refactoring around like 255 in Test Plans section itself/ Something like:

"Will any existing tests need to be refactored?"

PR #2082, opened 2020-10-06
Address "Built-in Defaulting" KEP feedback

apelisse:
cc @sttts @liggitt @DirectXMan12 @jpbetz

### DIFF
--- a/keps/sig-api-machinery/1929-built-in-default/README.md
+++ b/keps/sig-api-machinery/1929-built-in-default/README.md
@@ -67,13 +67,18 @@ SIG Architecture for cross-cutting KEPs).
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
-  - [CRD Normalization](#crd-normalization)
+  - [Changes to CRD Normalization](#changes-to-crd-normalization)
+  - [Changes to kubebuilder/default-gen](#changes-to-kubebuilderdefault-gen)
+  - [Default publishing](#default-publishing)
   - [Marker format](#marker-format)
-  - [Built-in Defaulting](#built-in-defaulting)
   - [Examples](#examples)
     - [Non-pointer structs](#non-pointer-structs)
     - [Struct pointers](#struct-pointers)
-    - [Non-pointer fields](#non-pointer-fields)
+    - [Non-pointer scalar fields](#non-pointer-scalar-fields)
+    - [Lists](#lists)
+    - [Lists without default](#lists-without-default)
+    - [String Maps](#string-maps)
+    - [String maps without default](#string-maps-without-default)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
@@ -105,11 +110,11 @@ checklist items _must_ be updated for the enhancement to be released.
 
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
-- [ ] (R) Graduation criteria is in place
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
 - [ ] (R) Production readiness review completed
 - [ ] Production readiness review approved
 - [ ] "Implementation History" section is up-to-date for milestone
@@ -145,7 +150,8 @@ field), non-deterministic or complicated defaults. These patterns
 might be useful in limited circumstances, but it should be easier to do
 the right thing in most cases.
 - Removes the inconsistency with CRDs that already support this
-mechanism.
+mechanism, and improves embedding of types from k8s.io/api into CRDs,
+which is currently cumbersome.
 - Solves bug where associative keys needs to either be required or
 defaulted. Default keys will now be able to be specified in the OpenAPI
 for built-in types, and SSA will be able to fill in the missing keys.
@@ -163,7 +169,7 @@ declarative defaults will be automatically carried along.
 ### Non-Goals
 
 While this proposal is probably paving the way for additional OpenAPI
-based validation, validation markers for built-in is entirely
+based validation, validation markers for built-in types is entirely
 out-of-scope for this project.
 
 This proposal also doesn't intend to remove the existing defaulting
@@ -176,11 +182,24 @@ some defaulting that doesn't require to be done as code though.
 It is important to us that the OpenAPI definition of a type should
 result in the same behavior, whether the type is built-in or a CRD.
 
-In order to do that, we need to change how CRDs are defaulted, and
-create a defaulting mechanism for built-in that would behave the same
-way.
+In order to do that, we need to change how CRDs are defaulted, and place
+specific requirements on definition of Golang types (specifically around
+omitempty on non-pointer scalars, and an implicit default of {} on
+non-pointer struct fields), and create a defaulting mechanism for
+built-in that would behave the same way.
 
-### CRD Normalization
+We will depend on tooling (defaulter-gen and kubebuilder) to enforce the
+new type requirements (via warnings or errors), to implicitly
+generate certain defaults in order to normalize the behaviour of
+built-in types and CRDs around zero values and absence/presence.
+
+Rule-of-thumb: when clients use the canonical go serialization, both
+clients for CRDs and built-ins should result in the same defaulting
+behaviour. We accept drift in the defaulting behaviour for JSON or YAML
+manifests that don't roundtrip through Golang types before being sent to
+the server (unstructured).
+
+### Changes to CRD Normalization
 
 Null values in CRDs are currently not defaulted. On the other hand,
 `null` typically means "Devoid of
@@ -192,28 +211,77 @@ values in the merge patch are given special meaning to indicate the
 removal of existing values in the
 target."(link)[https://tools.ietf.org/rfc/rfc7386.txt#:~:text=Null%20values%20in%20the%20merge]
 
-On the other hand, while `{field: null}` and `{}` probably have the same
+On the other hand, while `{"field": null}` and `{}` probably have the same
 meaning, they do not "look" the same and are hard to compare for
 clients. They are also currently treated differently by Kubernetes for
 defaulting purposes.
 
-We propose the creation of a new CRD flag: `preserveNullValues` which
-will default to `true`. This flag will be used as a ratcheting mechanism
-as we migrate to this new semantic throughout all Kubernetes types
-(which would be how all types eventually work). When false, the
-apiserver will use new semantics and remove all null values from CR
-objects. When true, the existing and deprecated semantics will be kept.
-That new flag is following the same logic as "preserveUnknownFields" and
-the new operation will happen using the same mechanisms right after
-pruning.
+Currently, null values trigger an error when nullable is false and don't
+get defaulted if a default is available.
+
+We propose:
+1. to start defaulting fields (both properties and additionalProperties)
+   with null values if a default is available and nullable unset or
+   false, or to remove that field if no default exists.
+2. to start defaulting items of arrays with null values if a default is
+   available and nullable unset or false, or to leave that null value in
+   the arrays and let validation fail if the field is not nullable.
+
+This means that we will start accepting inputs that were rejected in the
+past. There is code that would work with a new server but not with an
+old server. We assume that existing APIs that use nullable are happy
+with null values. Also, people can remove the nullable attribute if they
+want null values to start being removed. This will help aligning CRDs
+and built-in types (there are no ways to set nullable to true for
+built-in types).
 
 Note that we still consider `{}` (and 0, 0.0, "", false, []) to be
-different from `null`, so `{foo: {}}` is different from `{}` or from
-`{foo: null}` (but the latter two are considered equal).
+different from `null`, so `{"foo": {}}` is different from `{}` or from
+`{"foo": null}` (but the latter two are considered equal).
 
 As described by @sttts on many occasions, the immutability is blocked
 because we don't know how to distinguish between null and
 unspecified. We believe that this mechanism will be able to unblock it.
+
+### Changes to kubebuilder/default-gen
+
+For built-in types defaulting today takes place after marshalling Golang
+structs to JSON or ProtoBuf and then unmarshalling, back into Golang
+structs. This means that: unset non-pointer structs are always present
+after unmarshalling as empty struct ({}) zero valued non-pointer structs
+are marshalled as empty struct despite omitempty unset scalars without
+omitempty are indistinguishable from zero values after unmarshalling and
+are hence defaulted if a default is available. CRDs defaulting does not
+default zero values though.
+
+To normalize the behaviour, we propose:
+1. we auto-generate `default: {}` for non-pointer struct and forbid the
+`+default` marker
+2. we forbid or warn (liniting) about omitempty on non-pointer structs
+3. we require omitempty on scalars with default different from the
+zero-value
+4. we auto-generate `default: <zero-value>` for scalars without
+omitempty and forbid the `+default` marker
+
+We accept that manifests that don't roundtrip through Golang types
+(i.e. omitempty tags cannot lead to removal of zero valued fields) on
+the client, but are sent to the server via a dynamic client will have
+different defaulting behaviour between CRDs and built-ins (there is an
+example further down).
+
+### Default publishing
+
+CRDs do not publish default via the OpenAPI spec on the /openapi/v2
+endpoint today, but we have to decide what to do with the new default
+values of built-in types. The kube-openapi generated schemas will
+include them.
+
+We proppose to:
+1. leave CRDs not publishing defaults
+2. filter out defaults in the built-in type schemas too in a first step
+3. but possibly reconsider this behaviour in the future when the
+necessity of defaults for client-side merging (e.g. through kustomize)
+and influences on the ecosystem of default mis-use is better understood.
 
 ### Marker format
 
@@ -224,10 +292,13 @@ The marker will have the following format:
 which describes the value that the server will use if the field has a
 zero-value.
 
-`<any>` has to be in one-line YAML format (including its json form),
-representing the actual value to use by default. That is different from
-the current encoding in kubebuilder, but there seems to be consensus
-that they would be able to change that to YAML too for that new marker.
+`<any>` has to be in one-line JSON format (giving us the ability to move
+to YAML later while maintaining compatibility), representing the actual
+value to use by default.
+
+The JSON value will be unmarshalled by the apiserver when initializing,
+and that value will be then deep-copied into the object when defaulting
+happens.
 
 Namely, this will allow to preserve the concise semantics that you want
 for most values:
@@ -237,58 +308,36 @@ type Foo struct {
   // +default=32
   Integer int
 
-  // +default=bar
+  // +default="bar"
   String string
 
-  // +default=[popcorn, chips]
+  // +default=["popcorn", "chips"]
   StringList []string
 }
 ```
 while still allowing more complex expressions:
 ```
-// +default=[{port: 80, name: http}]
+// +default=[{"port": 80, "name": "http"}]
 ```
 
-Forcing the "one-line" subset allows us to sidestep a lot of the YAML
-weirdness in syntax, but we can still just use a YAML parser to parse
-the expressions if we want.
-
-### Built-in Defaulting
-
-Because built-in types are serialized and deserialized as Go structures,
-there are semantics that are imposed on us. The two main ones are:
-1. We can't differentiate between null, unspecified and zero-value,
-2. non-pointer structs are always present (and hence have to be
-recursively defaulted), they also can't be omitEmpty'd.
-
-The first point is solved by the removed nulls described in the previous
-section (and invalid zero-values can't be present in objects), and the
-second point can be solved by forcing non-pointer structs to have a `{}`
-default. To address that, we're proposing that the `default` marker is
-forbidden on non-pointer structs (pointer structs can behave as
-expected) and the default field of the openapi is automatically set to
-`{}`. This only applies to OpenAPI generated from built-in types or from
-kubebuilder. CRDs can still have any defaulting they have for
-structures, but we strongly suggest that OpenAPI generated from Go types
-should follow that semantics.
-
-Computed defaults (through defaulting functions) will not be supported
-as a declarative marker and will continue being supported through the
-existing defaulting functions.
 
 ### Examples
 
+Here are examples of the impact of the changes to CRD normalization,
+kubebuilder/openapi schema generation and the suggested semantics of the
++default marker.
+
 #### Non-pointer structs
 
-Given the following type, defined either in Kubebuilder or as a built-in type:
+Given the following type::
 ```golang
 type Root struct {
   Entry SubLevel
 }
 
 type SubLevel struct {
-  // +defaut=default-name
-  Name string
+  // +default="default-name"
+  Name string `json:"name,omitempty"`
   // +default=0
   Number int
 }
@@ -307,7 +356,7 @@ Root:
       properties:
         name:
           type: string
-          default: default-name
+          default: "default-name"
         number:
           type: integer
           default: 0
@@ -333,12 +382,12 @@ On the other hand, this is forbidden:
 ```golang
 Type Root struct {
   // Defaults on non-pointer structs are FORBIDDEN:
-  // +default={name: entry, number: 12}
+  // +default={"name": "entry", "number": 12}
   Entry SubLevel
 }
 
 Type SubLevel struct {
-  Name string
+  Name string `json:"name,omitempty"`
   Number int
 }
 ```
@@ -351,12 +400,12 @@ previous example:
 ```golang
 type Root struct {
   // This is now a pointer, we can provide a default:
-  // +default={name: pointer-name}
+  // +default={"name": "pointer-name"}
   Entry *SubLevel
 }
 
 type SubLevel struct {
-  // +defaut=default-name
+  // +defaut="default-name"
   Name string
   // +default=0
   Number int
@@ -372,13 +421,13 @@ Root:
   properties:
     entry:
       $ref: SubLevel
-      default: {name: pointer-name}
+      default: {"name": "pointer-name"}
 SubLevel:
   type: object
   properties:
     name:
       type: string
-      default: default-name
+      default: "default-name"
     number:
       type: integer
       default: 0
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -287,7 +286,7 @@ type Root struct {
 }
 
 type SubLevel struct {
-  // +defaut=default-name
+  // +defaut="default-name"
```
liggitt, line 289: in a few places:

```suggestion
  // +default="default-name"
```

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -430,21 +437,27 @@ Object:
   properties:
     invalid:
       type: string
-      default: default-name
+      default: "default-name"
+	  # This field would only be present for CRDs, for built-in,
+	  # the empty value would be accepted and treated as missing.
+	  minLength: 1
```
liggitt, line 443: nit: fix indent

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -484,27 +501,27 @@ the existing defaulting functions
...
 map key is missing. The code for this already exists in a
 [PR](https://github.com/kubernetes-sigs/structured-merge-diff/pull/83#issuecomment-498410353)
 5. Update some of the existing built-in APIs to specify their default
 when they are strictly declarative, probably focusing first on defaults
-that are part of the associative keys (e.g. `protocol: TCP`).
+that are part of the associative keys (e.g. `protocol: "TCP"`).
 
-Note that this is not changing how built-in defaults work, but is merely
-adding an extra-layer on top of that which makes type authors life easier.
+Note that we are keeping the exixsting built-in defaulting
```
liggitt, line 513:
```suggestion
Note that we are keeping the existing built-in defaulting
```

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -484,27 +501,27 @@ the existing defaulting functions
 2. Update kube-openapi to parse the `+default` marker and include it in
 the OpenAPI definition.
 3. Implement the new CRD semantics that clears `null` values from
-objects when the `x-kubernetes-remove-nulls` flag is present and true at
-the top-level (or in the CRD).
+objects when nullable is unset or false.
```
liggitt, line 504: Implement the new CRD semantics for explicit `null` values when `nullable` is unset or false:
* properties and list items with a default have the default applied, replacing the explicit null value
* list items without a default let the explicit null value remain, resulting in a validation error (matches current behavior)
* properties without a default have the explicit `null` value removed

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +192,29 @@ values in the merge patch are given special meaning to indicate the
...
-will default to `true`. This flag will be used as a ratcheting mechanism
-as we migrate to this new semantic throughout all Kubernetes types
-(which would be how all types eventually work). When false, the
-apiserver will use new semantics and remove all null values from CR
-objects. When true, the existing and deprecated semantics will be kept.
-That new flag is following the same logic as "preserveUnknownFields" and
-the new operation will happen using the same mechanisms right after
-pruning.
+We propose to start removing fields with a null value, when the openapi
+nullable flag is missing or false. We assume that existing APIs that use
```
liggitt, line 201: maybe tweak the ordering to clarify explicit null values will either be removed or defaulted. saying they'll be removed here, then defaulted later is hard to follow

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +192,29 @@ values in the merge patch are given special meaning to indicate the
...
+nullable are happy with null values. Also, people can remove the
+nullable attribute if they want null values to start being removed.
+This will help alignining CRDs and built-in types (there are no ways to
```
liggitt, line 204:
```suggestion
This will help aligning CRDs and built-in types (there are no ways to
```

liggitt (commented):
KEP needs approvers listed as well

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -430,21 +437,27 @@ Object:
...
-{"entry": {"name": "other-name", "number": 0}}
+>>> default('{name: other-name}')
+{"name": "other-name", "number": 0}
+# For built-in, additionally, we would have:
+>>> default('{name: ""}')
+{"name": "default-name", "number": 0}
 ```
 
-Note that the generated code for built-in will look like this:
+Note that a simplified version of the generated code for built-ins would look like this:
```
liggitt, line 460: give an example of the planned generated code for a complex object as well

apelisse: That's not exactly what you asked, but I think it has the same purpose

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +192,30 @@ values in the merge patch are given special meaning to indicate the
...
-pruning.
+Currently, null values trigger an error when nullable is false and don't
+get defaulted if a default is available. We propose to start defaulting
+fields with null values if a default is available, or the field will be
+removed if no default exists, but only when the openapi nullable flag is
+missing or false. We assume that existing APIs that use nullable are
+happy with null values. Also, people can remove the nullable attribute
+if they want null values to start being removed. This will help
+aligning CRDs and built-in types (there are no ways to set nullable to
+true for built-in types).
```
sttts, line 208: Be even more explicit: we start accepting inputs that were rejected in the past.

Note: this might mean that code that is working with a new server does not work with an old server version.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -400,10 +400,12 @@ which would have the following behavior, both for CRD types and built-in types:
...
 
-Non-pointer fields are also important, since they have a specific unmarshalling mechanism in Go.
+Non-pointer fields are also important, since they have a specific
+unmarshalling mechanism in Go.
 
-For CRDs, we can assume that invalid zero-values are omitted, otherwise
-the objects would technically be invalid.
+For CRDs, invalid zero-values are omitted, otherwise the objects would
+technically be invalid. We can safely default these field if they are
+unspecified.
```
sttts, line 408: what does "are omitted" mean? I cannot follow.

Are you saying we are adding a validation step before defaulting for all fields that have zero value?

Note that being invalid is not an easy property to check.

Example:

```yaml
type: string
maxLength: 3
pattern: ^....$
```

or

```
type: string
anyOf:
- minLength: 5
- maxLength: 4
```

In other words, you have to apply the full go-openapi/validator.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,8 +416,14 @@ Both these cases will be handled properly, here's an example:
 type Object struct {
      // This field has an invalid zero-value, empty string is not a
      // valid string.
-     // +default=default-name
-     Name string
+     //
+     // If this is a CRD, there would be a kubebuilder specific tag, and
+     // if this is a built-in, there would be a specific validation function,
+     // but basically:
+     // +minLength=1
```
sttts, line 423: does this mean `minLength: n` for some n>0 is hardcoded as the signal to switch to the new behaviour? If not, see comment above: it's getting complicated.

sttts: Please add an example for `type: number`. How do you want to express `!= 0.0`? Same for integer.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -430,32 +438,47 @@ Object:
...
+          if err := json.Unmarshal([]byte(`0`), &obj.Number); err != nil {
+               panic(err)
+          }
      }
 }
 ```
 
+It's still possible to specify an invalid zero-value in built-in types
+embedded into a CRD, but client-go will drop these values, thanks to
+`omitempty`.
```
sttts, line 480: What do you mean? Why does client-go matter for API semantics?

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: I'm not planning on changing any of the validation. If we allow empty strings today, we'll continue allowing empty strings with this proposal, but they won't be defaulted. Defaults will only be applied to unspecified or null values on non-nullable fields.

I agree that this doesn't solve all the problems that we would like to solve.

One such problem is for example, container protocol associative key. If someone embeds that container port structure in a CRD, and the "" string is passed for the protocol, it won't be defaulted and will do the wrong thing when it comes to using the associative key (CRD will think the key is `[$port, ""]`, but when the pod template is applied, the empty string will get defaulted to TCP and the key will become `[$port, TCP]`).

To be fair, the fact that we can't differentiate between empty strings and missing strings for built-in type is not addressed by this proposal and still needs to be solved.

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: I agree, it won't solve the api semantics, but it will do the right thing as long as people use client-go. That's already true for many things (you can specify `protocol: ""` and it should probably be an error, but client-go fixes it for you)

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: We can improve incrementally on that in a backward compatible way since this doesn't close the door to improve the zero-value semantics later, if we decide that it is worth it.

sttts: IMO this is a blocker for the whole enhancement. It clearly states in the goals:

> we will make the behavior equivalent between built-in types and CRDs

This goal is not reached.

sttts:
https://github.com/kubernetes/enhancements/pull/2082#discussion_r503173387 is serious to move forward. Not having a plan about a restriction of the approach which makes not reach the stated goal is a nogo.

keps/sig-api-machinery/1929-built-in-default/README.md
sttts:
```
x-kubernetes-prune-zero: true
```

How about something like that? And kubebuilder automatically adding that.

apelisse: Yeah, that's the obvious solution, which would be generated on "omitEmpty" non-pointers I assume. I like that this is not an option to users.

sttts: Am fine with something more advanced in kubebuilder that know the Go typesystem and when json.Marshal would apply omitempty and when it doesn't. In the CRD schema API we only have to provide the primitives such that kubebuilder can apply them accordingly.

apelisse:
I think we're good with this, can we merge this?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -201,11 +201,13 @@ Currently, null values trigger an error when nullable is false and don't
 get defaulted if a default is available. We propose to start defaulting
```
liggitt, line 201: can't comment above, but I would clarify the first paragraph of the proposal like this:

```diff
It is important to us that the OpenAPI definition of a type should
result in the same behavior, whether the type is built-in or a CRD.

In order to do that, we need to change how CRDs are defaulted, and
+ place specific requirements on definition of built-in types
+ (specifically around omitempty on non-pointer scalars,
+ and an implicit default of {} on non-pointer struct fields), and
create a defaulting mechanism for built-in that would behave the same
```

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -201,11 +201,13 @@ Currently, null values trigger an error when nullable is false and don't
...
+missing or false. This means that will start accepting inputs that were
+rejected in the past (there is code that would work with a new server
+but not with an old server). We assume that existing APIs that use
+nullable are happy with null values. Also, people can remove the
+nullable attribute if they want null values to start being removed. This
+will help aligning CRDs and built-in types (there are no ways to set
+nullable to true for built-in types).
 
 Note that we still consider `{}` (and 0, 0.0, "", false, []) to be
 different from `null`, so `{"foo": {}}` is different from `{}` or from
```
liggitt, line 213: can't comment below, but I'd suggest changing

> CRDs can still have any defaulting they have for structures, but...

to

> Existing CRD defaults defined for type=object properties will continue to be supported, but...

liggitt:
Found a few things to clarify and a few examples to fixup, mostly lgtm otherwise. Only real question I had was about the "`omitempty` on non-pointer scalar" requirement when the zero-value is set as an explicit default (https://github.com/kubernetes/enhancements/pull/2082#discussion_r507265827)

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +192,32 @@ values in the merge patch are given special meaning to indicate the
...
-That new flag is following the same logic as "preserveUnknownFields" and
-the new operation will happen using the same mechanisms right after
-pruning.
+Currently, null values trigger an error when nullable is false and don't
+get defaulted if a default is available. We propose to start defaulting
+fields with null values if a default is available, or the field will be
+removed if no default exists, but only when the openapi nullable flag is
+missing or false. This means that will start accepting inputs that were
+rejected in the past (there is code that would work with a new server
+but not with an old server). We assume that existing APIs that use
```
sttts, line 206: remove the parenthesis. It is misleading IMO. Code could also depend on the rejection of a request, i.e. work fine with an old server, but suddenly fail because a request gets accepted by a new server.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -400,10 +402,12 @@ which would have the following behavior, both for CRD types and built-in types:
 
 #### Non-pointer fields
 
-Non-pointer fields are also important, since they have a specific unmarshalling mechanism in Go.
+Non-pointer fields are also important, since they have a specific
```
sttts, line 405: what does "are also important" mean here? You mean that they need special semantics (= "always there") and therefore we have to come with something equivalent for CRDs.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -400,10 +402,12 @@ which would have the following behavior, both for CRD types and built-in types:
...
+unmarshalling mechanism in Go.
 
-For CRDs, we can assume that invalid zero-values are omitted, otherwise
-the objects would technically be invalid.
+For CRDs, invalid zero-values are omitted, otherwise the objects would
+technically be invalid. We can safely default these field if they are
```
sttts, line 409: I am still lost in these paragraphs. What do you want to say? What is motivation, what is a proposed change in behaviour?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +418,73 @@ Both these cases will be handled properly, here's an example:
...
 Object:
   type: object
   properties:
     invalid:
       type: string
-      default: default-name
+      default: "default-name"
+      # This field would only be present for CRDs, for built-in,
+      # the empty value would be accepted and treated as missing.
+      minLength: 1
```
sttts, line 450: Why do we want this?

Note that this is much more unnatural for ints, floats or bools. Why should strings be different?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +418,73 @@ Both these cases will be handled properly, here's an example:
...
     defaulted:
-      type: boolean
+      type: int
       default: false
```
sttts, line 453: type error

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +418,73 @@ Both these cases will be handled properly, here's an example:
...
-{"entry": {"name": "other-name", "number": 0}}
+>>> default('{name: other-name}')
+{"name": "other-name", "number": 0}
+# For built-in, additionally, we would have:
+>>> default('{name: ""}')
+{"name": "default-name", "number": 0}
 ```
 
-Note that the generated code for built-in will look like this:
+Note that a simplified, and non-cached version of the generated code for
```
sttts, line 467: what is a non-cached version?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -400,10 +402,12 @@ which would have the following behavior, both for CRD types and built-in types:
 
 #### Non-pointer fields
```
sttts, line 403: call them scalars. Also embedded structs are fields.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +418,73 @@ Both these cases will be handled properly, here's an example:
 type Object struct {
      // This field has an invalid zero-value, empty string is not a
      // valid string.
-     // +default=default-name
-     Name string
+     //
+     // If this is a CRD, there would be a kubebuilder specific tag, and
+     // if this is a built-in, there would be a specific validation function,
+     // but basically:
+     // +minLength=1
```
sttts, line 425: what I don't like about these constraints is that kubectl will follow them and a native PodSpec will be accepted while a CRD embedded PodSpec will be rejected.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -483,28 +512,34 @@ the additional defaulting functions, automatically call them prior to
 the existing defaulting functions
 2. Update kube-openapi to parse the `+default` marker and include it in
```
sttts, line 513: I miss the discussion and conclusion about publishing the OpenAPI defaults. For CRDs today we **do not** publish them intentionally. I suggest to do the same here.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -237,22 +243,18 @@ type Foo struct {
...
 ```
-// +default=[{port: 80, name: http}]
+// +default=[{"port": 80, "name": "http"}]
 ```
 
-Forcing the "one-line" subset allows us to sidestep a lot of the YAML
-weirdness in syntax, but we can still just use a YAML parser to parse
-the expressions if we want.
-
 ### Built-in Defaulting
```
sttts, line 258: why Built-in? You mostly talk about the CRD defaulting below ?!

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +418,73 @@ Both these cases will be handled properly, here's an example:
...
-          obj.Name = "default-name"
+          if err := json.Unmarshal([]byte(`"default-name"`), &obj.Name); err != nil {
+               panic(err)
+          }
      }
      if obj.Number == 0 {
-          obj.Number = 0
+          if err := json.Unmarshal([]byte(`0`), &obj.Number); err != nil {
+               panic(err)
+          }
```
sttts, line 479: I hope this is just a sketch and we are a bit more clever than calling Unmarshal all the time.

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: > For CRDs today we do not publish them intentionally.

I didn't know that. I don't really understand it, but I guess I'm fine if we do the same. As long as server-side apply is able to read them since that's one of the reason why we're building that mechanism.

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: I'll remove it yeah

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -431,6 +433,10 @@ type Object struct {
 }
 ```
 
+Non-pointer fields that have a default must be omitEmpty. Kubebuilder
+and built-in defaulting generators should both prevent these from
+happening.
```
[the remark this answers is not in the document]
apelisse: Yeah, there are lots of places in kubernetes where we have this done implicitly. I don't think we want to stop these from specifically mention the default, so I don't think it hurts to have the nuance.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -431,6 +433,10 @@ type Object struct {
 }
```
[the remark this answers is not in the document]
apelisse: Rephrased to: "This field would be defaulted to 0 without a default." because it's not omitempty and it's not a pointer.

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: Yes, we're clearly planning on caching the deserialized version since it doesn't change.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +194,32 @@ values in the merge patch are given special meaning to indicate the
...
+get defaulted if a default is available. We propose to start defaulting
+fields with null values if a default is available, or the field will be
+removed if no default exists, but only when the openapi nullable flag is
+missing or false. This means that will start accepting inputs that were
+rejected in the past. There is code that would work with a new server
+but not with an old server. We assume that existing APIs that use
+nullable are happy with null values. Also, people can remove the
+nullable attribute if they want null values to start being removed. This
+will help aligning CRDs and built-in types (there are no ways to set
+nullable to true for built-in types).
```
sttts, line 212: Maybe worth a short node:

We will remove null fields also from `map[string]T`s, i.e. for type that have `additionalProperties` in their schema instead of `properties`.

keps/sig-api-machinery/1929-built-in-default/README.md
sttts: This is still wrong.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +416,64 @@ Both these cases will be handled properly, here's an example:
 type Object struct {
      // This field has an invalid zero-value, empty string is not a
      // valid string.
-     // +default=default-name
-     Name string
+     // +default="default-name"
+     Name string `json:"name,omitempty"`
```
sttts, line 420: this is no example for the cases above:
1. `""` is valid
2. the default is not the zero value.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,12 +402,10 @@ which would have the following behavior, both for CRD types and built-in types:
...
 
-Non-pointer fields are also important, since they have a specific unmarshalling mechanism in Go.
-
-For CRDs, we can assume that invalid zero-values are omitted, otherwise
-the objects would technically be invalid.
+Non-pointer scalar fields also have a specific unmarshalling mechanism in Go,
+since they always have a value.
 
 For built-ins, we can differentiate two cases:
 1. The zero-value for the field is invalid (e.g. if empty string is not a valid value)
```
sttts, line 411: I don't see an example for this case. What does this even mean to be "invalid" ?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +416,64 @@ Both these cases will be handled properly, here's an example:
...
      // +default=0
      Defaulted int
 }
 ```
 
+Non-pointer scalar fields that have a default must be omitempty (unless
+their default is the zero-value). Kubebuilder and built-in defaulting
+generators should both require this.
+
 This would generate the following OpenAPI:
```
sttts, line 432: where is the golang source for the schema below?

keps/sig-api-machinery/1929-built-in-default/README.md
sttts: SSA will see the full schema. This happens as part of the OpenAPI v2 conversion.

sttts: Have you added a clarification anywhere?

sttts:
Am still not happy with the examples section. There are still a number of typos and missing cases. I would also like to see a clear distinction between the proposed 
1. pruning behaviour
2. defaulting behaviour
3. kubebuilder/defaulter-gen changes.
4. examples.
Especially the lack of separation of 1-3 vs. 4 makes it read very handwavy.

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: How do you know "" is valid?

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: Ah yes, the name doesn't match, good catch

keps/sig-api-machinery/1929-built-in-default/README.md
apelisse: No, not yet, I'm not super satisfied to hide them. it means we can't generate a good "openapi" example to test the behavior in kube-openapi without re-adding the defaults, it means client can't merge properly using listmapkey because they don't have the default either. The more expressive the schema the better, we should keep the defaults, and we should probably do so for CRDs too.

keps/sig-api-machinery/1929-built-in-default/README.md
sttts: Why does it matter whether it is valid or not? Why not the other cases? This is not obvious.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -414,48 +416,64 @@ Both these cases will be handled properly, here's an example:
 type Object struct {
      // This field has an invalid zero-value, empty string is not a
      // valid string.
-     // +default=default-name
-     Name string
+     // +default="default-name"
+     Name string `json:"name,omitempty"`
 
-     // This field is defaulted to 0 with no defaulter.
+     // This field would be defaulted to 0 without a default.
```
sttts, line 422: Why would it be default to 0 without a default? What about non-omitempty cases in general?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,64 +426,120 @@ which would have the following behavior, both for CRD types and built-in types:
...
-     // valid string.
-     // +default=default-name
-     Name string
+     // This field is omit-empty, and hence Golang unmarshalling drops a zero-value.
+     // +default="default-name"
+     Name string `json:"name,omitempty"`
 
-     // This field is defaulted to 0 with no defaulter.
-     // +default=0
+     // This field would be defaulted to 0 without a default.
```
apelisse, line 441: That defaulted below looks a little weird now.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -177,12 +178,23 @@ It is important to us that the OpenAPI definition of a type should
...
-around omitempty on non-pointer scalars, and an implicit default of {}
-on non-pointer struct fields), and create a defaulting mechanism for
+specific requirements on definition of Golang types (specifically around
+omitempty on non-pointer scalars, and an implicit default of {} on
+non-pointer struct fields), and create a defaulting mechanism for
 built-in that would behave the same way.
 
-### CRD Normalization
+We will depend on tooling (defaulter-gen and kubebuilder) to enforce the
+new type requirements (via linting, warnings or errors), to implicitly
```
sttts, line 189: nit: linting and warnings was meant to be the same

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -200,29 +212,55 @@ clients. They are also currently treated differently by Kubernetes for
...
+omitempty are indistinguishable from zero values after unmarshalling and
+are hence defaulted if a default is available. CRDs defaulting does not
+default zero values though.
+
+To normalize the behaviour, we propose:
+1. we auto-generate `default: {}` for non-pointer struct and forbid the
+   +default marker
+2. we forbid or warn (liniting) about omitempty on non-pointer structs
+3. we require omitempty on scalars with default different from the
+   zero-value.
```
sttts, line 259: see my 4th rule in the gdoc

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -408,51 +428,56 @@ which would have the following behavior, both for CRD types and built-in types:
...
 
-Both these cases will be handled properly, here's an example, for built-ins:
 ```golang
 type Object struct {
-     // This field has an invalid zero-value, empty string is not a
-     // valid name and will be defaulted.
+     // This field is omit-empty, and hence Golang unmarshalling drops a zero-value.
      // +default="default-name"
-     Invalid string `json:"name,omitempty"`
+     Name string `json:"name,omitempty"`
```
sttts, line 441: `json:"name,omitempty"` by rule 3

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -474,9 +499,46 @@ func GenerateDefault(obj *Object) {
 
 It's still possible to specify an invalid zero-value in built-in types
 embedded into a CRD, but client-go will drop these values, thanks to
-`omitempty`. Non-pointer scalar fields that have a default must be
-omitempty (unless their default is the zero-value). Kubebuilder and
-built-in defaulting generators should both require this.
+`omitempty`. On the other hand, a manifest in YAML
```
sttts, line 506: a manifest in JSON

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -474,9 +499,46 @@ func GenerateDefault(obj *Object) {
...
+
+{"entry": {"invalid": "", "number": 0}}
+
+converted to JSON directly and sent to the server would result in
```
sttts, line 510: remove "converted to JSON direction and"

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -474,9 +499,46 @@ func GenerateDefault(obj *Object) {
...
+
+```golang
+type Object struct {
+	List []Item
+}
+
+// +default="apple"
+type Item string
+```
+
```
sttts, line 535: flash this one out

keps/sig-api-machinery/1929-built-in-default/README.md
sttts: See my text in the gdoc. I still think we have to describe our plan.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -192,29 +208,62 @@ values in the merge patch are given special meaning to indicate the
...
-different from `null`, so `{foo: {}}` is different from `{}` or from
-`{foo: null}` (but the latter two are considered equal).
+different from `null`, so `{"foo": {}}` is different from `{}` or from
+`{"foo": null}` (but the latter two are considered equal).
 
 As described by @sttts on many occasions, the immutability is blocked
 because we don't know how to distinguish between null and
 unspecified. We believe that this mechanism will be able to unblock it.
 
+### Changes to kubebuilder/default-gen
```
sttts, line 243: @DirectXMan12 please take a look

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -253,17 +253,33 @@ default zero values though.
...
 To normalize the behaviour, we propose:
 1. we auto-generate `default: {}` for non-pointer struct and forbid the
-   +default marker
+`+default` marker
 2. we forbid or warn (liniting) about omitempty on non-pointer structs
 3. we require omitempty on scalars with default different from the
-   zero-value.
+zero-value
+4. we auto-generate `default: <zero-value>` for scalars without
+omitempty and forbit the `+default` marker
```
sttts: forbid

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -559,6 +579,25 @@ type Object struct {
...
+        type: string
+        default: "banana"
+```
+
+And the defaulting will have the following behavior:
+```python
+>>> default('{mapping: {foo: null, bar: apple}')
+{"mapping": {"foo": "banana", "bar": "apple"}
+```
+
```
sttts, line 600: maybe another case with null and no default => removal

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -548,6 +549,25 @@ type Object struct {
...
+        type: string
+        default: "apple"
+```
+
+And the defaulting will have the following behavior:
+```python
+>>> default('{list: [null, foo]}')
+{"list": ["apple", "foo"]}
+```
+
```
sttts, line 570: add case with null and no default => NO removal

sttts:
Only two nits left: examples for array and string map with null and no default

Other than that: LGTM

apelisse:
Added the example and rebased!

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,64 +447,216 @@ which would have the following behavior, both for CRD types and built-in types:
...
+Gives the following OpenAPI:
+```yaml
+Root:
+  type: object
+  default: {}
+  properties:
+    mapping:
+      type: object
+	  additionalProperties:
```
lavalamp, line 620: nit: tab

apelisse: 😬

lavalamp:
/approve

Maybe @DirectXMan12 can look at the section @sttts tagged them in and do the final LGTM?

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,64 +447,216 @@ which would have the following behavior, both for CRD types and built-in types:
...
+The valid Object type would generate the following OpenAPI:
 ```yaml
 Object:
   type: object
   properties:
-    invalid:
+    name:
       type: string
-      default: default-name
+      default: "default-name"
```
DirectXMan12, line 489: NIT: don't actually need the quotes here -- it's the same either way ;-)

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,64 +447,216 @@ which would have the following behavior, both for CRD types and built-in types:
...
+It's still possible to specify an invalid zero-value in built-in types
+embedded into a CRD, but client-go will drop these values, thanks to
+`omitempty`. On the other hand, a manifest in JSON
+
+{"entry": {"invalid": "", "number": 0}}
+
+sent to the server would result in different behaviour for built-ins
+
+```python
+>>> default('{name: ""}')
```
DirectXMan12, line 532: NIT: it's a bit weird that the example object changes when it looks like you're going to keep talking about the above invalid entry.

keps/sig-api-machinery/1929-built-in-default/README.md
```diff
@@ -398,64 +447,216 @@ which would have the following behavior, both for CRD types and built-in types:
...
+{"name": "default-name", "defaulted": 0}
+```
+
+and CRDs:
+
+```python
+>>> default('{name:"", number:0}')
```
DirectXMan12, line 539: `number` vs `defaulted`?

DirectXMan12 (commented):
couple of nits inline, but consider this a soft lgtm, expecially in the part I was directly tagged in

/lgtm

PR #2098, opened 2020-10-14
update SIG Instrumentation KEPs with current status and update-date

logicalhan:
Update metadata on SIG Instrumentation KEPs.


keps/sig-instrumentation/20191028-metrics-stability-to-beta.md
```diff
@@ -12,8 +12,9 @@ approvers:
   - "@brancz"
 editor: "@brancz"
 creation-date: 2019-10-28
-last-updated: 2019-10-28
-status: implementable
+last-updated: 2020-10-14
+status: implemented
```
wojtek-t, line 16: We don't set status to implemented for things that didn't yet reach GA.

logicalhan: This KEP is for promoting this framework to beta. All the requirements for beta are done, which means this is implemented.

logicalhan: I mean the title literally says 'Metric Stability to Beta'. It would be extremely strange to keep this in an implementable stage until we are at GA.

wojtek-t: We shouldn't have separate KEPs for Beta and for GA. There should be a single KEP for a feature - it should should be updated before GA.
Otherwise it's much harder to track what is happening.

PR #2113, opened 2020-10-23
WIP Add a matchName selection option to the NetworkPolicy namespace selector API

jayunit100:
#2112


keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,341 @@
...
+metadata:
+  name: mysql-allow-app
+spec:
+  podSelector:
+    matchLabels:
+      app: mysql
+  ingress:
+  - from:
+    - namespaceSelector:
+         matchName:
```
andrewsykim, line 215: Adding a `matchName` field to `namespaceSelector` is potentially a breaking change because we would have to create a new internal struct for `namespaceSelector` to something like:
```go
type NamespaceSelector struct {
  metav1.LabelSelector

  matchNames []string
}
``` 

While maybe we can get away with not breaking json/yaml serialization, I think we still break clients creating NetworkPolicy types off of the new version? If this is a breaking change, at worse we should add a new field outside namespaceSelector. But breaking changes aside, I like `matchNames` under `namespaceSelector` much better.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,341 @@
...
+
+
+### Test Plan
+
+We will add tests for this new api semantic into the exting test/e2e/ network policy test suites in upstream which cover both of these scenarios, using the validation framework outlined in https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md#motivation 
+
+### Graduation Criteria
+
+#### Alpha 
+- Add a feature gated new field to NetworkPolicy
```
andrewsykim, line 239: What's the feature gate called?

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+The ability to target namespaces using names, when building NetworkPolicy objects, has been a request from the broader k8s community for quite some time.  In the sig-network-networkpolicy API group, it was one of the most popular policies which was voted on.
+
+## Motivation
+
+NetworkPolicies are under utilized by developers when defining applications, and the most obvious tenancy boundary for a developer is that of **the Namespace**.  We thus aim to make this boundary **extremely** obvious and easy for users to target in a maximally secure manner.
+
+Although service-mesh and other technologies have been slated to obviate the need for developer driven security boundaries, these technologies aren't available in most clusters, and aren't supported by the Kubernetes API.
+
+The ability to provide granular and intuitive network boundaries between apps is part of the braoder vision to make the NetworkPolicy API a universal security construct implemented in all production applications. 
+
```
rikatz, line 117: @jayunit100 wdyt about inserting the original issue here? And maybe the ML discussion? :D

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+Meanwhile, the "exclude" implementation, will look like this:
+
+```
+  ingress:
+  - from:
+    - namespaceSelector:
+         matchName:
+         -  *
```
rikatz, line 233: As you said supporting matching of 'wildcard namespaces'  is a non goal, don't you think this syntax (with *) will lead to some confusion? Just thinking here.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Story 2
+
+As an user I want to "just add" a namespace to my allow list without having to manage the labels which might get added/removed over time from said namespace.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+- CNIs may not choose, initially to support this construct, and dilligent communication with CNI providers will be needed to make sure its widely adopted.
```
danwinship, line 197:
```suggestion
- CNIs may not choose, initially, to support this construct, and diligent communication with CNI providers will be needed to make sure it's widely adopted.
```

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+- Fractionating the Kubernetes API idiom by introducing a separate way to select objects.  We accept this cost because security is a fundamentally important paradigm that justifies breaking other paradigms, at times.
```
danwinship, line 198: There are other APIs that select objects by name in some cases. I don't think this is a "risk". It's a choice that needs to be defended, but it's not a risk.

keps/sig-network/20201014-NamespaceAsName.md
danwinship: yeah, it seems like it would be clearer to do `matchNameExcept:` or something

keps/sig-network/20201014-NamespaceAsName.md
danwinship: An alternative to changing the syntax would be to add a "virtual label" with some name that is not a valid label name (eg, `virtual/name` which is not valid because the prefix part isn't a DNS name). Then you could just do:

    - namespaceSelector:
        matchLabels:
          virtual/name: my-frontend

but also you could do

    - namespaceSelector:
        matchExpressions:
          - key: virtual/name
            operator: NotIn
            values:
              - kube-system

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+Meanwhile, the "exclude" implementation, will look like this:
+
+```
+  ingress:
+  - from:
+    - namespaceSelector:
+         matchName:
+         -  *
+         exclude:
+         - kube-system
```
danwinship, line 235: This policy says "allow connections from all namespaces except kube-system" which is the opposite of the thing you said you wanted to support and is therefore confusing.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+
+
+### Test Plan
+
+We will add tests for this new api semantic into the exting test/e2e/ network policy test suites in upstream which cover both of these scenarios, using the validation framework outlined in https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md#motivation 
+
+### Graduation Criteria
+
+#### Alpha 
+- Add a feature gated new field to NetworkPolicy
```
danwinship, line 246: In the past we have not used feature gates for new NetworkPolicy features.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+- Add validation tests in API
+
+#### Beta
+- The name selector has been supported for at least 1 minor release
+- At least one CNI provider implements the new field
+- Feature Gate is enabled by Default
+
+#### GA Graduation
+
+- At least two CNIs support the The name selector field
```
danwinship, line 257: I think "at least one" and "at least two" are too small for an NP feature. We don't currently have good support for being able to express that the CNI plugin doesn't support a NetworkPolicy feature that exists in the API. Therefore, it is bad when the feature is there but plugins don't support it. So we should aim for reasonably broad support before we start encouraging users to think it's safe to use.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+- The name selector has been enabled by default for at least 1 minor release
+
+#### Removing a Deprecated Flag
```
danwinship, line 260: (this section is irrelevant to this KEP and can be removed)

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+
+As an administrator I want to block all users from accessing the kube-system namespace as a fundamental default policy.
+
+#### Story 2
+
+As an user I want to "just add" a namespace to my allow list without having to manage the labels which might get added/removed over time from said namespace.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
```
danwinship, line 195: We need to consider how policies using the new feature will be interpreted by plugins that don't yet support the new feature. eg:

- causes the plugin to crash due to unexpectedly invalid data
- causes any `NetworkPolicyPeer` using the new feature to look like it says "allow all"
- causes any `NetworkPolicyPeer` using the new feature to look like it says "allow none"

Some of these possibilities are better than others...

kikisdeliveryservice (changes requested):
So you need to use the issue number to correctly format this as follows:
main dir:
keps/sig-network/2112/NamespaceAsName/

20201014-NamespaceAsName.md ->../2112/NamespaceAsName/README.md

add kep.yaml

See: https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template

Update README.md to have title KEP-2112:

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,341 @@
...
+- Add a `matchName` option (which is additive to the current `matchLabels` selector).
+- Add an additional semantic mechanism to "allow all" namespaces while excluding a finite, specific, list of namespaces (canonically, this might be `kube-system`, as this is an obvious security boundary which most clusters would prefer).
+
+### Non-Goals
+
+- Support matching of wildcard namespaces
+
+## Proposal
+
+In NetworkPolicy specification, inside ``NetworkPolicyPort`` object struct specify a new ``PortRange`` field:
```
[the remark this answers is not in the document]
jayunit100: :)

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+The latter comment received 9 likes as a github issue - indiciating the general popularity of this as a request, and that correlates well to feedback we've seen in the broader community as well.
+
+We thus conclude that, even though targetting an object using its name is not normal in K8s, in the case of NetworkPolicies, the overwhelming need for universal, easy to define security boundaries, makes a strong case for amending the API to support a "special" selector for namespace names that is **independent** of labels.
+
+### Goals
+
+- Add a `matchName` option (which is additive to the current `matchLabels` selector).
+- Add an additional semantic mechanism to "allow all" namespaces while excluding a finite, specific, list of namespaces (canonically, this might be `kube-system`, as this is an obvious security boundary which most clusters would prefer).
```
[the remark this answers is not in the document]
jayunit100: updated, pushing shortly...

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,348 @@
...
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+
+As an administrator I want to block all users from accessing the kube-system namespace as a fundamental default policy.
```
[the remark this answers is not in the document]
jayunit100: updated, pushing shortly

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fixed pushing shortly

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fixed

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: broadened to include 4 - am assuming that might be something like ovn, calico, cillium, antrea.

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: sure, im happy to modify this or remove it, ill bring this up at the mtng today

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fxd

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: yeah this `*` is kinda sneaky, restricted regex, will mislead people.  
`matchNameExcept` seems equally weird though to me... ill bring this up on the call today

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: done

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: yeah, i think we dont need one per dands comment ill just delete this

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: Yeah, i guess as an egress rule this makes more sense, but better yet i think removing this for now, discussing it and re-adding it later is better given ricardos comment above ...  the user story for the exclusion rules is a little messy ATM.  fixing...

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: ive never understood the virtual labels thing... are there any examples of how people hack stuff up to use these ?

keps/sig-network/20201014-NamespaceAsName.md
rikatz: @andrewsykim can you please provide the api guide discussed in the meeting? Thanks!

andrewsykim: I believe feature gates and now required for new fields to stable APIs to ensure safe rollback https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-field-in-existing-api-version

keps/sig-network/20201014-NamespaceAsName.md
thockin: This is probably my prime concern with this proposal

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,332 @@
...
+#### Story 2
+
+As an user I want to "just add" a namespace to my allow list without having to manage the labels which might get added/removed over time from said namespace.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+- CNIs may not choose, initially, to support this construct, and diligent communication with CNI providers will be needed to make sure it's widely adopted.
+- CNIs may not choose, initially to support this construct, and dilligent communication with CNI providers will be needed to make sure its widely adopted.
```
thockin, line 202: why is this line duplicated?

keps/sig-network/20201014-NamespaceAsName.md
thockin: Let's not invent ad hoc syntax. Please :)

Adding a field to metav1.LabelSelector is dangerous at best.  I find it confusing - it's not a selector in the usual kubernetes sense.  Also breaks clients which should not be taken lightly.

Virtual labels are interesting but if we want to go that route, we'd need to pause and define that whole mechanism.  Does it apply to all selectors and all resources?  Is it just for "name" or does it have other uses?  Does it show up when I GET an object?  This is a KEP of its own.

The *simplest* option here is to add `namespaceNames []string` to `NetworkPolicyPeer`.

You still have to define what consumers should do if they don't support this field.  They would see a `NetworkPolicyPeer` with no fields set which fails validation.  So to make this work we would need to specifically loosen that validation.  Define that to mean "allow none".  That, of course has to be alpha-gated.  We'd really need to work with the mail implementations to verify that they either do the right thing already or change them to treat that safely.

So (for example) the very best case: 1.20 could allow empty peers and add support for `namespaceNames[]` if and only if the  gate was enabled.  1.21 could flip the gate to default-on for both (beta or GA, TBD).  If it takes a while to get implementations doing the right thing, we may have to leave it in alpha for longer.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,332 @@
...
+         -  my-frontend
+```
+
+### Test Plan
+
+We will add tests for this new api semantic into the exting test/e2e/ network policy test suites in upstream which cover both of these scenarios, using the validation framework outlined in https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20200204-NetworkPolicy-verification-rearchitecture.md#motivation 
+
+### Graduation Criteria
+
+Note that we may not have explicity feature gates for this KEP, because NetworkPolicies typically haven't been implemented with gates.  We keep this section for the time being in any case as it organizes our thinking around the general progression of the feature over time.
```
thockin, line 235: All new field introductions need gates for at least one release.

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,332 @@
...
+
+#### Alpha 
+- Communicate CNI providers about the new field
+- Add validation tests in API which confirm several positive / negative scenarios in the test matrix for when this field is present/absent
+
+#### Beta
+- The name selector has been supported for at least 1 minor release
+- Four commonly used CNI providers implement the new field, with generally positive feedback on its usage.
```
thockin, line 243: Why is this written in terms of CNI providers instead of NetPol providers?

keps/sig-network/20201014-NamespaceAsName.md
thockin: still says 2 here?

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: updated to include these below , pushing shortly

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fixed, pushing

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fixed pushing shortly

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,362 @@
...
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-xxxx: Add Namespace as Name field to NetworkPolicy
```
rikatz, line 61: @jayunit100 suggestion to change here to KEP-2112 as this should be the issue number

keps/sig-network/20201014-NamespaceAsName.md
```diff
@@ -0,0 +1,362 @@
...
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: mysql-allow-app
+spec:
+  podSelector:
+    matchLabels:
+      app: mysql
+  ingress:
+  - from:
+      matchName:
```
rikatz, line 229: Probably here is ``namespaceNames``

rikatz: but reading loud, ``namespaceNames`` as a field doesn't sounds strange? wouldn't be better to be only ``namespaces`` here?

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: Updated

rikatz:
> So you need to use the issue number to correctly format this as follows:
> main dir:
> keps/sig-network/2112-NamespaceAsName/
> 
> (though I'm not sure that's a descriptive enough title tbh)
> 
> 20201014-NamespaceAsName.md ->../2112-NamespaceAsName/README.md
> 
> add kep.yaml
> 
> See: https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template
> 
> Update README.md to have title KEP-2112:

@kikisdeliveryservice updated Jay's PR here

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,347 @@
...
+- Both "allow all" and "allow none" could be incorrectly interpreted 
+incorrectly, so we need to make sure that plugins not implementing the feature 
+don't get misled by the new data structure.
+
+The `NetworkPolicyPeer` should be as obvious as possible to interpret in a way that is compatible with the semantics of the v1 policy API **before** this feature was added to it.  
+
+## Design Details
+
+- Add a new selector to the network policy peer data structure which can 
+switch between allowing a "namespaceNames" or "matchLabels" implementation,
```
andrewsykim, line 196: This seems to imply that a user has to set either `namespaceNames` or `namespaceSelector` but above in the goals it mentions:
```
- Add a `namespaceNames` option (which is additive to the current `matchLabels` selector).
```

Which implies we are taking the union of `namespaceNames` and `namespaceSelector`. Might be good to clarify this a bit.

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,347 @@
...
+
+1) One simple way to implement this feature is to modify the NetworkPolicyPeer 
+object, like so.  Since this doesn't involve modifying a go type (i.e. were 
+not replacing a labelSelector with a different type, we expect it 
+to be a cleaner implementation:
+
+```
+type NetworkPolicyPeer struct {
+  // new field
+  namespaceNames []string
```
andrewsykim, line 153: Can we add the API docs here?

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: updating this now, I actually thought more about this today

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: updated to `namespaceSelectorAllow` , because now that were putting this in NetworkPolicyPeer, I figured having uniform naming to align it with `namespaceSelector` (i.e. it extends the namespaceSelector rather then being a logical union with it, as we do w/ podSelectors and so on)

jayunit100:
so, after thinking more about it... now that we're modding NetworkPolicyPeer, it seems that naming the field so that it aligns with the namespaceSelector more, made sense, so I updated the design details to reflect that iteration.  

I still want  to comb through this and make sure its logically consistent all the way through again, after getting away from it for a day or so.

Thanks for all the feedback so far, I think its starting to take shape.  

I'd like to explore virtual labels again at some point, before we merge this, as that might still be an alternative which is more robust long term, but don't understand the idea well enough yet.

keps/sig-network/20201014-NamespaceAsName.md
danwinship: > You still have to define what consumers should do if they don't support this field.

> If it takes a while to get implementations doing the right thing, we may have to leave it in alpha for longer.

The feature gate should be tracking "are we satisfied with the API?". That's separate from "does everyone implement the API?" though clearly related.

#2137 is an attempt to deal with the problem of NetworkPolicy feature versioning more generically (though of course, if it merges in the same release as this feature does then it doesn't really solve the problem for _this_ feature since plugins that didn't implement namespace names probably wouldn't implementing versioning either)

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+	IPBlock *IPBlock `json:"ipBlock,omitempty" protobuf:"bytes,3,rep,name=ipBlock"`
+}
+```
+
+### User Stories (Optional)
+
+#### Story 1
+
+As an administrator i want to allow access from monitoring pods in kube-system 
+namespace, into any pod in my cluster, as a fundamental policy.
```
danwinship, line 167: NetworkPolicy is not intended to solve administrator problems, it's intended to solve user/developer problems. Thus, administrator-focused user stories do not provide justification for NetworkPolicy features.

And anyway, the administrator can configure the cluster in such a way that they are the only one who can manipulate Namespace labels, and thus they can make it so they can securely use `namespaceSelector`. In fact, they _should_ use `namespaceSelector`, because at some point they're going to realize that they should move the monitoring pods into their own namespace, and if they used labels to identify the monitoring namespace then all they have to do is properly label the new namespace and they're done, while if they had added explicit "allow from `kube-system`" policies everywhere, then they have to replace all of those policies with new "allow from `monitoring`" ones.

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+#### Story 1
+
+As an administrator i want to allow access from monitoring pods in kube-system 
+namespace, into any pod in my cluster, as a fundamental policy. 
+
+#### Story 2
+
+As an user I want to "just add" a namespace to my allow list without having to 
+manage the labels which might get added/removed over time from said namespace.
```
danwinship, line 172: This isn't a user story, it's just a restatement of the feature proposal.

You already spent a bunch of time justifying the feature in the Motivation section. You don't need user stories as well if you made your case there.

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+### Notes/Constraints/Caveats (Optional)
+
+
+### Risks and Mitigations
+
+- NetworkPolicy providers may **opt-out**, initially to support this construct, and dilligent communication with CNI providers will be needed to make sure its widely adopted, thus, this feature needs to be backwards compatible with the existing v1 api.
+- We thus need to make sure that hidden defaults don't break the meaning of existing policys, for example:
+  - if `namespaceSelectorAllow` field is retrieved by an api client which doesn't yet support it, the client doesnt crash (and the plugin doesnt crash, either).
```
danwinship, line 181: I talked about this only vaguely before because it wasn't clear at that point what the API was going to look like. At this point you can get more concrete. Specifically, when an "old" plugin sees a policy using `namespaceSelectorAllow`, it will think that it has an empty `NetworkPolicyPeer`, which is something it may not be able to deal with, since that wouldn't have passed validation before...

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+1) One simple way to implement this feature is to modify the NetworkPolicyPeer 
+object, like so.  Since this doesn't involve modifying a go type (i.e. were 
+not replacing a labelSelector with a different type, we expect it 
+to be a cleaner implementation:
+
+```
+type NetworkPolicyPeer struct {
+  // new field
+  namespaceSelectorAllow []string
```
danwinship, line 153: having `namespaceSelector` and `namespaceSelectorAllow` implies that the two fields differ in something having to do with "allowing", which is not accurate at all. The new field should have some name that makes it clear that it's about namespace names. (It does not need to specify "allow" because everything in `NetworkPolicyPeer` specifies something to allow.)

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+  - if `namespaceSelectorAllow` is empty, it has IDENTICAL semantics as if it were nil.
+    - in other words, there is no "allow all" namespaces semantic which corresponds to emptyness
+
+Overall, the `NetworkPolicyPeer` modifications should be unobtrusive with regard to how other fields interplay, regardless of its absence or presence, notwithstanding the fact that it potentially broadens the selection of the `namespaceSelector` field.
+
+## Design Details
+
+- Add a new selector to the network policy peer data structure which can switch between allowing a `namespaceSelectorAllow`, supporting a policy that is expressed like this:
+
+- A list of conventional namespaces
```
danwinship, line 193: what are "conventional" namespaces?

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+    matchLabels:
+      app: mysql
+  ingress:
+  - from:
+      namespaceSelectorAllow:
+      -  my-frontend
+      -  my-frontend-2
+```
+
+- As a more sophisticated example: The following would allow all traffic from `podName:xyz` living in `my-frontend` AND `my-frontend-2`, but NOT trafic from `my-frontend-3`.  This is because the presence of EITHER a namespaceSelectorAllow rule or a namespaceSelector, makes the podSelector subject to an **AND** filter operation.
```
danwinship, line 212: I don't think there's any good argument for allowing both `namespaceSelector` and `namespaceSelectorAllow` in the same peer.

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+    matchLabels:
+      app: mysql
+  ingress:
+  - from:
+    - namespaceSelectorAllow:
+      -  my-frontend
+      namespaceSelector:
+        matchLabels:
+          app: my-frontend-2
+      ipBlock:
```
danwinship, line 230: You are definitely not allowed to specify `ipBlock` with any other `NetworkPolicyPeer` field

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+None
+
+## Drawbacks
+
+We dont want to have > 1 way to select namespaces.
+
+## Alternatives
+
+A network policy operator could be created which translated a CRD into many networkpolicys on the fly, by watching namespaces and updating labels dynamically.  This would be a privileged container in a cluster and likely would not gain much adoption.
```
danwinship, line 346: I'm not at all sure what you're suggesting here

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+We thus conclude that, even though targeting an object using its name is not 
+normal in K8s, in the case of NetworkPolicies, the overwhelming need for 
+universal, easy to define security boundaries, makes a strong case for 
+amending the API to support a "special" selector for namespace names that is 
+**independent** of labels.
+
+### Goals
+
+- Add a `namespaceSelectorAllow` option (which is additive to the current `matchLabels` selector).
```
andrewsykim, line 128: This name is pretty confusing IMO. First, I think the term `Allow` is redundant cause you only specify namespaces you are trying to allow. Second the term `Selector` implies this selects against labels (or most users will think this at least). Prefer `namespaces` or namespaceNames`.

jayunit100: yeah i dont like this name either.  just wanted something other then `namespaceSelector` since that was taken. 

I'll go to `namespaceNames`

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: yeah, namespaceSelectorAllow was a horrible choice :) , changing to `namespaceNames`

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: per comment below, got rid of this section, its also redundant  to the motivation which is pretty data driven

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: removing section

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: updated

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
...
+- Support matching of wildcard namespaces
+- Changing the internal details of how namespaceSelector works with arbitrary syntax
+- Depending on more sophisticated features (like virtual labels) to accomplish 
+this feature without changing the API (note that, if someone were 
+to write a virtualLabels KEP, however, it might change the trajectory of this 
+KEP, which is worth discussing)
+
+## Proposal
+
+In NetworkPolicy specification, inside `NetworkPolicyPeer` specify a new `namespaceSelectorAllow` field.
```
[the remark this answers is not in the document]
jayunit100: moved

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: fixed

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: clarifying/ replacing the sentence below as well

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: added a small example and clarified it....  its that weird policy meta operator idea i keep shoping around :).

keps/sig-network/20201014-NamespaceAsName.md
jayunit100: fxd

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,371 @@
+<!-- toc -->
```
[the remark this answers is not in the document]
jayunit100: fxd

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: `podSelector && ( namespaceNames || namespaceSelector ) ` is what i meant there, updated the text. 

does that make sense , or you think thats bad to allow using both (imo namespaceNames just appends to the ALLOW rules so, theres an obvious use case where you have one policy and you have 5 namespaces selected by label and the 6th namespace you select by name)

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: fxd

keps/sig-network/2112-netpol-namespace-name-selector/README.md
jayunit100: nested the bullets to make this more clear and readable, added that old distinction

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,393 @@
...
+
+1) One simple way to implement this feature is to modify the NetworkPolicyPeer 
+object, like so.  Since this doesn't involve modifying a go type (i.e. were 
+not replacing a labelSelector with a different type, we expect it 
+to be a cleaner implementation:
+
+```
+type NetworkPolicyPeer struct {
+  // new field
+  namespaceNames []string
```
rikatz, line 174: Probably per api design, this needs to be upper case first letter, right? (also because this field needs to be exported)

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,393 @@
...
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: mysql-allow-app
+spec:
+  podSelector:
+    matchLabels:
+      app: mysql
+  ingress:
+  - from:
+    - namespaceNames:
```
thockin, line 224: This doesn't gel - NamespaceNames and NamespaceSelector should be mutually exclusive, shouldn't they?  They only exception we offer is that you can define NamespaceSelector and PodSelector, so I assume NamespaceNames and PodSelector would be legit.

keps/sig-network/2112-netpol-namespace-name-selector/README.md
```diff
@@ -0,0 +1,393 @@
...
+
+See motivation section for user stories.  This KEP collects concrete data and opinions from several individuals over the past few years, hence we do not include generic user stories.
+
+### Notes/Constraints/Caveats (Optional)
+
+### Risks and Mitigations
+
+- NetworkPolicy providers may **opt-out**, initially to support this construct, and dilligent communication with CNI providers will be needed to make sure its widely adopted, thus, this feature needs to be backwards compatible with the existing v1 api.
+- We thus need to make sure that hidden defaults don't break the meaning of existing policys, for example:
+  - if `namespaceNames` field is retrieved by an older client which **doesn't** yet support it:
```
thockin, line 151: This is a huge liability.

If you have policy that says:

```
from:
    namespaceNames:
        - foo
    podSelector:
        app: myapp
```

That explicitly means to allow traffic from namespace "foo" AND NOT from *this* namespace.  If you have an old client who doesn't know about the new field you will deserialize as:

```
from:
    podSelector:
        app: myapp
```

That explicitly means to allow traffic from *this* namespace.  That's a VERY different meaning.

I'm not sure what the best way to mitigate this is.  This is a general API evolution problem.  We're adding a new option to a one-of and we need to accommodate clients that can't see it.  In this case, we really want to fail-closed.

As proposed, this API doesn't fail-closed.

One option is to add a new sub-struct that contains BOTH the list of namespace names and the pod selector.  This is redundant BUT it dodges the failure mode.  Now the above example becomes:

 ```
from:
    namespaceList:
        namespaceNames:
          - foo
        podSelector:
          app: myapp
```

Which the old clients read as:

```
from: {}
```

I am very open to alternate ideas, but I think "old client fails closed" is probably a requirement.

Comment (thockin) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:151]: curiously, virtual labels do not suffer this failure mode...

File: keps/sig-network/2112-netpol-namespace-name-selector/README.md (hunk 34 above)
Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:224]: ah missing bullet, fixing

Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:248]: fied now

File: keps/sig-network/2112-netpol-namespace-name-selector/README.md (hunk 35 above)
Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: Ahhhh so, now that we're adding a new way to select things, if an old thing ignores that selector, a new policy will appear to be extremely promiscuous, because that new selector will be perceived as nil.

Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: by virtual label, do we roughly mean "put junk in the selector that always mismatches unless a client is aware of the secret syntax" ?  i.e. 

```       
    namespaceSelector:
        ^virtualNamespaceName: "foo"
```

if so , yeah your right thats pretty clean .

Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: Just to make sure we get to the bottom of the assumptions here (strongly leaning towards synth. labels...)... but to complete this thought of your workaround... 

2.  Considering your workaround idea...  Seems like an arbitrary distinction just making sure im correct, by `from: {}` you really mean

```
from: []
```

right?  from is a list of PeerPolicies, not a map.

... In any case, that `from: []` thingy... thats a `DENY ALL`... right ? (Theres no peers, so theres nothing to match an allow against).    
- in the current default, your allowing more traffic then you intended to
- in the deny all scenario, you are denying more traffic then you intended to 

so, not sure why one is inherently better then the other.

Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: # 3  ... Synthetic labels... same pitfall, just a different default  ?

Ok so, after thinking about (1) and (2) above, i came to the conclusion that, synthetic labels are just "different" not "better".... unless theres a hidden premise im missing.

If a client isnt capable of interprettying a synthetic match label for a namespace, then it will always fail matching the namespace, and thus it will ALSO have default "Deny All" semantics.  So Im assuming "deny all is the right behaviour when you dont understand a field" is what your going for.  

.... but of course LMK if im missin somethin, im not the best w/ this API stuff :)

Comment (thockin) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: > by virtual label, do we roughly mean "put junk in the selector that always mismatches unless a client is aware of the secret syntax"

Caveat: making it up as I go.  No secret syntax.  We own the
`*.kubernetes.io/*` prefix, so you could imagine defining that
`virtual.kubernetes.io/name` is always set to the resource's name.  Now, I
don't know what that really means implementation wise - do we actually store
that or populate it on read or somnething else.  I don't know if apimachinery
would go for it.  I don't know what other issues it would present (e.g. scale).

For sanity, we could say that any attempt to write such a label either fails or
is just discarded.

> QUESTION: just making sure im correct, by from: {} did you really mean

From API docs for `From []NetworkPolicyPeer`:  

"""
If this field is empty or missing, this rule matches all sources (traffic not
restricted by source). If this field is present and contains at least one item,
this rule allows traffic only if the traffic matches at least one item in the
from list.
"""

So the result is a list of one element, and that element having no fields set
falls into the second part: "...allows traffic only if the traffic matches..."

> that `from: []` thingy... thats a DENY ALL... right ? 

NOTE: `from: []` means "from all".  `from: [{}]` means deny all.  I specified
`from: {}` which is wrong - It should have been `[{}]`.  The list HAS an
element, but the client doesn't know what it said, so it is morally `{}`.

> ts still quite far from the original intent of allowing specific traffic from a ns/pod combination, albeit, its more restrictive rather then less.

I expect fail-closed will make people less angry than the alternative :)

> synthetic labels are just "different" not "better".... unless theres a hidden premise im missing. That is, they still result in a wrong interpretation of a rule, because, like any rule, you cannot follow it if you dont understand the grammer.

That's why no magic syntax.  Think of it as a label that every resource has
without specifying.

Comment (rikatz) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: So following what we've discussed in our meeting today, is that possible that we create a validation and/or a mutation in the api side? Like:

* If ``namespaceName > 0`` , then verify if ``namespaceSelector > 0``
* If ``namespaceSelector < 1``, then mutate the object and create a ``namespaceSelector`` that will never be selected

So one way to do this would be to create a [well known label](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/) like networking.kubernetes.io/null-namespace and in the situation above, try to select this label.

Another solution could be to select namespace using a [matchExpression](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#labelselector-v1-meta) with a never targeted expression. One could make a namespaceSelector with a labelSelector and not the same labelExpression, because per the doc:
```
A label selector is a label query over a set of resources. The result of matchLabels and matchExpressions are ANDed. An empty label selector matches all objects. A null label selector matches no objects.
```

So the following will always select nothing (unless my logics are wrong, and the ANDed in the description above is something like "the union of the results of both selectors", and not "if x && y")

```
namespaceSelector:
  matchLabels:
      noOp: true
  matchExpression:
      - {key: noOp, operator: DoesNotExist}
```
Because if someone creates a noOp label, the second statement will be false, so null selector. If there's no label, the first is false again. But again, my logic can be reaaallly wrong :)


Other way that @jayunit100  suggested would be to insert anything random to be selected.

Comment (thockin) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: I am not sure what you're getting at here - maybe you can explain a bit more?

> * If ``namespaceName > 0`` , then verify if ``namespaceSelector > 0``
> * If ``namespaceSelector < 1``, then mutate the object and create a ``namespaceSelector`` that will never be selected

What is a valid selector that can never select anything?

> So one way to do this would be to create a [well known label](https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/) like networking.kubernetes.io/null-namespace and in the situation above, mutate the network policy and try to select this label for the namespace.

Two problems.  First, we generally do not modify user input in surprising ways
(and I think this might qualify :).  Second, this means we have to define what
it means when both namespaceNames and namespaceSelector are defined at the same
time.  I am not very excited about carving out a special case here.  E.g.

If namespaceSelector is specifed AND it has one value AND that value is the
"magic word", then don't use it and instead do this other thing.

> Another solution could be to select namespace using a [matchExpression](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#labelselector-v1-meta) with a never targeted expression. One could make a namespaceSelector with a labelSelector and not the same labelExpression, because per the doc:
> ```
> A label selector is a label query over a set of resources. The result of matchLabels
> and matchExpressions are ANDed. An empty label selector matches all objects.
> A null label selector matches no objects.
> ```
>
> So the following will always select nothing (unless my logics are wrong, and the ANDed in the description above is something like "the union of the results of both selectors", and not "if x && y")
>
> ```
> namespaceSelector:
>   matchLabels:
>       noOp: true
>   matchExpression:
>       - {key: noOp, operator: DoesNotExist}
> ```
> Because if someone creates a noOp label, the second statement will be false, so null selector. If there's no label, the first is false again. But again, my logic can be reaaallly wrong :)

That's kind of a hack.  It's impressive in a way, but a hack nonetheless.  See
my above statement about magic words and special-cases.

> Other way that @jayunit100  suggested would be to insert anything random to be selected.

I wasn't there, so I don't know for sure what this means, but it feels like
we're thrashing around looking for a clever hack.  Clever is the enemy of
obvious.  I'd rather just evolve this toward something un-clever.

E.g. We have:

```
PodSelector *metav1.LabelSelector
NamespaceSelector *metav1.LabelSelector
IPBlock *IPBlock
```

We could evolve to something like:

```
// Deprecated - use Pods.Selector
PodSelector *metav1.LabelSelector
// Deprecated - use Namespaces.Selector or Pods.Namespaces.Selector
NamespaceSelector *metav1.LabelSelector

// One of the following may be set.  Consumers should check them in the same
// order they are defined here.  Over time, new options may be added.  Clients
// which do not understand the new options should fail closed.
Pods *PodsPeer
Namespaces *NamespacesPeer
IPBlock *IPBlock
```

and define:

```
type PodsPeer struct {
    // Must be non-empty
    Selector *metav1.LabelSelector
    // Optional, ANDed with selector
    Namespaces *NamespacesPeer
}

type NamespacesPeer struct {
    // One of the following may be set.
    Selector *metav1.LabelSelector
    Names []string
}
```

This is just hand-waving, but hopefully you can see where I am trying to go
with it.

Comment (jayunit100) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: Yeah I actually agree that the whole default label / noop label / timestamp label is all a hack.  It's like, even if we do it, it's like making the API worse as a whole.

**If** perfectly non surprising  backward compatibility with other v1 CNIs is a must, then This KEP feels increasingly to me like it should be either an API change (v2) 
, a new type of selector, or a potential "default/synthesized" labels, or "netpol operator" knob. 

That *if* above is maybe what I should bring up at the next sig net call :).

Comment (rikatz) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: > What is a valid selector that can never select anything?

Yes, sorry, I was writing in a rush. The statement here is "if I have inserted namespace by name selector, then I should verify if namespace by label is empty. If it's empty, then I should insert anything that turns it not empty, but never matched. This way if the CNI does not know about the namespace by name selector, it will still not select anything because it knows a namespace by label that doesn't select nothing -> default deny"

> First, we generally do not modify user input in surprising ways (and I think this might qualify :)

Indeed, this was a doubt we had during the meeting, if this was plausible or not.

> Second, this means we have to define what it means when both namespaceNames and namespaceSelector are defined at the same time.

IMO this should be an "OR", because when you specify a namespaceName you already know it's name, so it doesn't make sense to select also it's label. The namespaceSelector is for when you don't have specific namespace information (or want to select a group of namespaces).

> That's kind of a hack. It's impressive in a way, but a hack nonetheless. 

I don't know if I thank and feel 'proud' or with some sort of shame. We call this in Portuguese a "Gambiarra", but I need more time to explain the real meaning. It can be summarized as "a nasty ugly hack that no one knows how it works, why it works, but somehow it works". I agree, I don't like this suggestion, but as this was in my head as a possibility I had to write at least so we could discard it

> This is just hand-waving, but hopefully you can see where I am trying to go with it.

Yes, I got it. But going this way, are we saying that we're deprecating fields in a stable API (Netpol is v1)? If so then Jay is right in his statement and we should go to the way of a v2 api, right?

Comment (rikatz) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: Oh also I've remembered, this sort of case makes more and more concrete the needs and concerns that @danwinship has raised in this [KEP](https://github.com/kubernetes/enhancements/pull/2137).

@thockin I would suggest, if you have some bandwidth to take a look into Dan's Kep before coming back to this  (and maybe to the PortRanges) because it tries to solve this sort of problems in a way I think might be a concern not only for CNIs, but also for CSIs and other stuffs that might plug into Kubernetes using stable APIs :)

Comment (danwinship) [reply keps/sig-network/2112-netpol-namespace-name-selector/README.md:150]: So #2137 would solve this if everyone already implemented #2137...

One way to do this would be to do "NetworkPolicy v2" **in the Kubernetes sense of v2, not in the "NetworkPolicy++" sense**. We add a new `networkingv2.NetworkPolicy` that is a slightly-cleaned-up version of `networkingv1.NetworkPolicy`, and we add `minVersion` to both `networkingv1.NetworkPolicySpec` and `networkingv2.NetworkPolicySpec`. Then in the future when we want to add new features, we can add them in a nice way to `networkingv2.NetworkPolicy` and in an ugly-but-backward-compatible way to `networkingv1.NetworkPolicy` (with the expectation being that people wouldn't actually try to use the new features via the v1 API so it doesn't matter if it's ugly and hard to write).

eg, the user creates a policy:

```
# NB: "v2"
apiVersion: networking.k8s.io/v2
kind: NetworkPolicy
spec:
  podSelector:
    app: myServer
  ingress:
    - from:
        # Allow connections from all pods in namespaces labeled "special: true"
        # and from pods labeled "app: myapp" in namespace "foo"
        - namespaceSelector:
            matchLabels:
              special: true
        - namespaceNames:
            - foo
          podSelector:
            app: myapp
```

per #2137, the apiserver sees `namespaceNames` and so adds `spec.minVersion: 1.21` at creation time.

A network plugin that implements `networkingv2.NetworkPolicy` and supports the version 1.21 features can then read the policy and interpret it correctly.

A network plugin that implements `networkingv2.NetworkPolicy` but only supports 1.20 and earlier would see the `minVersion`, know that it can't implement this policy, and set appropriate `status` conditions to inform the user.

An old network plugin that doesn't implement `networkingv2.NetworkPolicy` would see the new NetworkPolicy via the `networkingv1` API, which would look something like:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
spec:
  minVersion: 1.21
  podSelector:
    app: myServer
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              special: true
    - fromV2:
        - namespaceNames:
            - foo
          podSelector:
            app: myapp
          v2Index: 1
```

which, after discarding the fields it doesn't understand is:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
spec:
  podSelector:
    app: myServer
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              special: true
```

so it would still see the `namespaceSelector` rule that it understands, but it wouldn't see the peer containing the `namespaceNames` rule at all. (So it would end up implementing a more restrictive policy rather than a less restrictive one. Of course we could define the v2↔v1 mapping any way we wanted if we wanted some different behavior here.)

danwinship: > Of course we could define the v2<->v1 mapping any way we wanted if we wanted some different behavior here.

actually it would be a lot simpler to just say that if there were any "new" fields anywhere in `ingress`, then the entire `ingress` section gets moved to `ingressV2` in the v1 version, so v1 clients would just see it as a policy that isolates some pods but doesn't allow any access to them. Likewise with `egress`.

jayunit100: Yup, we need to solve this by some kind of indirect means, tldr.

jayunit100: Hiya dan quick q as I attempt to formalize this discussion:

Can you clarify what you mean by 


"it as a policy that isolates some pods but doesn't allow any access to them"....  the "but" is making my brain spin  there, since isolating poss and not allowing access to pods are potentially the same operation, not opposed in any obvious way.

danwinship: Like, normally a NetworkPolicy says "isolate these pods (implicitly) and then allow access from X, Y, and Z". If you "hide" `spec.ingress` then the policy ends up saying "isolate these pods, and that's it"

thockin: > If perfectly non surprising backward compatibility with other v1 CNIs is a must

It's unlikely that there is a perfect solution here.  Only "acceptable", which
(IMO) means "fail closed".

Virtual labels is the closest to ideal FOR THIS SPECIFIC CASE, but it's a bit
of a detour to the objective here.

> are we saying that we're deprecating fields in a stable API

Deprecating fields is allowed, just not REMOVING them.  A v2 API would allow us
to clean up the API a little, but not really much.  We still need fidelity in
round-trip.  There's no panacea.

> per #2137, the apiserver sees namespaceNames and so adds spec.minVersion: 1.21 at creation time.

I'll go read that KEP, but I will say that micro-versioning of APIs has been
discussed, but it comes with its own set of non-trivial costs.  As you
describe, SOMEONE needs to taste the incoming objects and set the API versions,
including across update operations.  Absent a general mechanism to do this, I
have little faith it will work.

Or we make the user set it, and deal with all sorts of error handling when they
get it wrong.

> so it would still see the namespaceSelector rule that it understands, but it wouldn't see the peer containing the namespaceNames rule at all. (So it would end up implementing a more restrictive policy rather than a less restrictive one

We don't need micro-versioning to achieve this, though, just thoughful API
design?

fasaxc: Agree lack of a way to match on name is a friction point.  The Calico team have some experience here; when we convert a namespace to our datamodel, we add a label in our namespace so that the user can match on the name.  That seems to work fine for our users.

There's plenty of precedent in the k8s API for auto-added labels (such as all the ones on the Node resource).  Seems like the established norm is that those are just added to the datamodel as "real" labels that can be read as normal with kubectl.  If the label was added as a "real" label like that, support would just fall out in all the CNI plugins and users would be able to do more advanced matches where they want to do so.

danwinship: Node objects are more-or-less owned by Kubernetes itself though, so the fact that Kubernetes adds labels to Nodes isn't an example of "Kubernetes adding labels to other people's objects", it's just a slightly weird case of "owners adding labels to their own objects".

danwinship: > So #2137 would solve this if everyone already implemented #2137...

after some discussion there, #2137 solves the problem of "the apiserver knows about this feature but the network plugin doesn't" (user tries to apply a policy using this feature to a 1.22 cluster running a 1.20-era network plugin) but it doesn't solve the problem when the apiserver doesn't know about the feature (user tries to apply a policy using this feature to a 1.20 cluster) unless the user manually specifies `minVersion` on the policy, which is ugly.

jayunit100:
per discussion with @thockin and @liggitt "virtual labels" are potentially a thing, and were going to explore that avenue as a solution to this

jayunit100:
likely closing this in lieue of #2162 !

thockin:
Preemptive close to get it off my radar :)

PR #2117, opened 2020-10-27
fixes #1780

craiglpeters:


craiglpeters:
Will redo without need for rebase

PR #2120, opened 2020-10-30
Mark KEP 1659 as "implemented"

thockin:
Clear out last "UNRESOLVED"

### DIFF
--- a/keps/sig-architecture/1659-standard-topology-labels/README.md
+++ b/keps/sig-architecture/1659-standard-topology-labels/README.md
@@ -43,6 +43,7 @@ To get started with this template:
 - [Design Details](#design-details)
   - [Reserve a label prefix](#reserve-a-label-prefix)
   - [Defining the meaning of existing labels](#defining-the-meaning-of-existing-labels)
+  - [Redefining kubernetes.io/hostname](#redefining-kubernetesiohostname)
   - [Defining a third key (or not)](#defining-a-third-key-or-not)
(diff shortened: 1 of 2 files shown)


kikisdeliveryservice (commented):
should the milestone in the kep.yaml also be cleared out? since it's moving from provisional-> implementable now, 1.19 seems incorrect there?

andrewsykim (commented):
/lgtm

thockin:
@kikisdeliveryservice this is a weird one because it's really not a code change.  I'll take your suggestion and just clear it out.

thockin:
https://github.com/kubernetes/website/pull/24783 implements this, so we will have a race - implemented vs implementable :)

thockin:
2nd commit pushed - this in now implemented

andrewsykim (commented):
/approve
/lgtm

thockin:
PTAL @ahg-g

thockin:
@kikisdeliveryservice is there more I need to do to get KEP changes approved?

kikisdeliveryservice:
> @kikisdeliveryservice is there more I need to do to get KEP changes approved?

Since this seems to be docs work/arch work done outside of the release milestones, I think it's fine to just mark it as implemented (if you're all satisfied) and after that merges close the tracking issue here: https://github.com/kubernetes/enhancements/issues/1659

thockin:
Yeah, this is not release-locked :)

Thanks

PR #2134, opened 2020-11-07
Set KEP 1860 as implementable

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

Since https://github.com/kubernetes/kubernetes/pull/92312 is merged, this KEP should be updated

Ref #1860


thockin (commented):
Thanks!

/lgtm
/approve

PR #2147, opened 2020-11-12
update KEP for CSIServiceAccountToken

zshihang:


keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
```diff
@@ -15,9 +15,9 @@ approvers:
   - "@msau42"
   - "@mikedanese"
 creation-date: 2020-06-09
-last-updated: 2020-07-07
+last-updated: 2020-11-12
 status: implementable
-stage: alpha
+stage: beta
```
msau42, line 20: It's still alpha. I think when we plan to move to beta in 1.21, we can update this again along with the other beta kep requirements.

zshihang: done. thought it is implementable for beta. i changed to implemented for alpha.

keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
```diff
@@ -15,8 +15,8 @@ approvers:
   - "@msau42"
   - "@mikedanese"
 creation-date: 2020-06-09
-last-updated: 2020-07-07
-status: implementable
+last-updated: 2020-11-12
+status: implemented
```
msau42, line 19: From the kep template, the stage is defined as the current dev cycle:
https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml#L26

I think it's fine to leave it as implementable and alpha. I believe we save implemented for when the feature is GA

zshihang: done.

PR #2179, opened 2020-12-10
Promote Immutable Secrets/ConfigMaps feature to GA

wojtek-t:
/assign @johnbelamaric  - for PRR approval
/assign @msau42 @saad-ali - for SIG storage approval


wojtek-t:
/hold

[To ensure that it won't accidentaly merge without both PRR and SIG-storage approval.]

johnbelamaric:
PRR is approved - I am not using the command because it will approve the whole PR.

PR #2180, opened 2020-12-11
Tracker for production-readiness approvals

wojtek-t:


johnbelamaric:
I think this will be a conflict-magnet. Can we create a separate yaml per KEP, say `<kep-number>.yaml`?

wojtek-t:
> I think this will be a conflict-magnet. Can we create a separate yaml per KEP, say <kep-number>.yaml?

I'm afraid we may end-up with gazillions files in the directory.
OTOH, they are easily discoverable my number, so actually that might be fine.
We may consider splitting by sig into directories at the later phase if needed too...

@johnbelamaric Assuming I will split that into per-kep file, does that look reasonable to you?

johnbelamaric:
Yes. Go ahead and split them into SIG-based directories, and kep-per-file, by number.

Then, we need to update the template to tell people how to do it, and update kepval to check for it. You could even have kepctl create the file for a kep though that may be overkill. 

Thanks!

cmd/kepval/main_test.go
```diff
@@ -70,14 +73,43 @@ func TestValidation(t *testing.T) {
...
+			}
+			defer kepFile.Close()
+
+			kep := kepParser.Parse(kepFile)
+			if kep.Error != nil {
+				t.Errorf("%v has an error: %v", filename, kep.Error)
+			}
+
+			requiredPRRApproval := len(kep.Number) > 0 && kep.LatestMilestone >= "v1.21"
+			// TODO: There are a few KEPs that already target 1.21 and don't have PRR approval.
```
wojtek-t, line 94: @johnbelamaric - I'm wondering what smart thing we can do here about targeting to avoid breaking stuff.
Suggestions?

[Should we consider downgrading the latest-milestone on these maybe?]

johnbelamaric: Are we validating ever KEP on every PR? That is, we will fail for any PR now unless we do that? Or will we only fail when one of these KEPs is touched?

I think it's reasonable to change their `latest-milestone` back to 1.20 and/or remove it in order to force the PRR to be filled.

keps/prod-readiness/sig-api-machinery/1040.yaml
```diff
@@ -0,0 +1,4 @@
+kep-number: 1040
+beta:
+  approved: true
```
johnbelamaric, line 3: Do we even need this field? Isn't presence of an approver enough? Would this ever be set to `false`?

That is, since this file cannot be created without approval from a PRR approver, it would seem that the presence of this in the repo is sufficient.

pkg/kepval/prrs/validations/yaml.go
```diff
@@ -0,0 +1,98 @@
...
+	for key, value := range parsed {
+		// First off the key has to be a string. fact.
+		k, ok := key.(string)
+		if !ok {
+			return util.NewKeyMustBeString(k)
+		}
+
+		// figure out the types
+		switch strings.ToLower(k) {
+		case "alpha", "beta", "stable":
```
johnbelamaric, line 47: I think the validation should be:

- Find the `stage` from kep.yaml
- Check for presence and validity of PRR approver for that stage, if `latest-milestone` is > 1.20

johnbelamaric (changes requested):

keps/prod-readiness/sig-api-machinery/1040.yaml
wojtek-t: Fair enough - removed

pkg/kepval/prrs/validations/yaml.go
wojtek-t: That makes sense to me, but I think this logic doesn't belong here.

This part is more of a syntax validation.
The actual semantic validation depending on the kep.yaml is happening in the test itself - let me add that logic there.

wojtek-t (commented):
Comments addressed.

I also extracted https://github.com/kubernetes/enhancements/pull/2184 out of tihs PR, since these are useful changes on its own.

pkg/kepval/prrs/validations/yaml.go
wojtek-t: Added in the test itself (main_test.go)

cmd/kepval/main_test.go
wojtek-t: Yes - we exercise all keps on every PR.
I guess I will try to clean them up once this PR will be more ready to go.

johnbelamaric:
Ok, it's getting there, modulo the additional cleanup.

One thing - how do we handle it when a PRR approver leaves and is no longer in the list? `prr-approvers-emeritus`?

wojtek-t:
> One thing - how do we handle it when a PRR approver leaves and is no longer in the list? prr-approvers-emeritus?

Yeah - that would be my thought too. And validating that the approver is in any of that list.
Simple to do and quite intuitive imho.

cmd/kepval/main.go
```diff
@@ -24,6 +24,8 @@ import (
 	"k8s.io/enhancements/pkg/kepval/keps"
 )
 
+// TODO: Can we remove this file?
```
wojtek-t, line 27: @johnbelamaric - what do you think about this one?
Do you know if people are using this binary?

I think that changing this binary to what the test itself is doing might not be the best option (because e.g. we're hardcoing prod-readiness directory, which makes sense for test, but not necessary for the binary).
OTOH, without that, it's not a full solution.

So while things like "kepctl" are really binaries, here what we really need is a test.

WDYT?

johnbelamaric: yeah, I don't know what this is for

cmd/kepval/main_test.go
```diff
@@ -70,14 +73,61 @@ func TestValidation(t *testing.T) {
...
+				stageMilestone = kep.Milestone.Beta
+				stagePRRApprover = prr.Beta.Approver
+			case "stable":
+				stageMilestone = kep.Milestone.Stable
+				stagePRRApprover = prr.Stable.Approver
+			}
+			if len(stageMilestone) > 0 && stageMilestone >= "v1.21" {
+				// PRR approval is needed.
+				if len(stagePRRApprover) == 0 {
+					t.Errorf("PRR not approved for: %s", kep.Stage)
```
johnbelamaric, line 129: Can we point the user to the docs (KEP template? PRR policy doc?), so they have a pointer to how to fix the failure?

cmd/kepval/main_test.go
```diff
@@ -70,14 +73,61 @@ func TestValidation(t *testing.T) {
...
+				t.Fatalf("could not open file %s: %v\n", filename, err)
+			}
+			defer kepFile.Close()
+
+			kep := kepParser.Parse(kepFile)
+			if kep.Error != nil {
+				t.Errorf("%v has an error: %v", filename, kep.Error)
+			}
+
+			requiredPRRApproval := len(kep.Number) > 0 && kep.LatestMilestone >= "v1.21"
```
sftim, line 93: Just a query: Is there a way for a KEP to target v1.21 without a KEP number?

wojtek-t: Currently kep-number is not required.
But I want to make it required eventually (but we would need to migrate all keps to the new template).

cmd/kepval/main_test.go
wojtek-t: Sure - that makes perfect sense. Will add once your relevant PRs will merge.

keps/sig-cloud-provider/1959-service-lb-class-field/kep.yaml
```diff
@@ -24,7 +24,7 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 27: @andrewsykim FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-network/1860-kube-proxy-IP-node-binding/kep.yaml
```diff
@@ -16,12 +16,11 @@ approvers:
 
 stage: "alpha"
 
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 19: @Sh4d1 FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-network/2004-topology-aware-subsetting/kep.yaml
```diff
@@ -26,7 +26,7 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 29: @ckyiricky  @robscott  FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-network/2079-network-policy-port-range/kep.yaml
```diff
@@ -18,7 +18,7 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 21: @rikatz FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-node/2043-pod-resource-concrete-assigments/kep.yaml
```diff
@@ -27,7 +27,7 @@ stage: stable
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 30: @dashpole @vikaschoudhary16 @RenaudWasTaken @fromanirh @AlexeyPerevalov FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-scheduling/1923-prefer-nominated-node/kep.yaml
```diff
@@ -15,7 +15,7 @@ approvers:
   - "@Huang-Wei"
   - "@ahg-g"
 stage: alpha
-latest-milestone: "v1.21"
+# latest-milestone: "v1.21"
```
johnbelamaric, line 18: @chendave FYI, PRR approval will need to be part of the PR that sets this back to 1.21.

keps/sig-network/2079-network-policy-port-range/kep.yaml
rikatz: @johnbelamaric @wojtek-t ok, will review the KEP questionaire about the PRR. PR for the feature is open https://github.com/kubernetes/kubernetes/pull/97058 

Need to submit a new PR for the KEP with the questions revisited, right?

wojtek-t: Correct.
Or rather - I see that you already have filled some PRR sections (didn't look deep enough if all needed for Alpha, but potentially). So basically, you just need to have that reviewed by the PRR approvers.

keps/sig-scheduling/1923-prefer-nominated-node/kep.yaml
chendave: Done, thanks for the notification.

keps/sig-network/2079-network-policy-port-range/kep.yaml
rikatz: @wojtek-t completed some more questions there.

If you don't mind being the PRR approver there, I can change the kep.yaml to have you as the prr-approver :) 

Thanks!

wojtek-t: I'm happy to review but I'm OOO for the next 2.5 weeks - can review in the second full week of January - please assign the PR to me directly.

rikatz: No rush :) 

I'll assign you to the PR and mark you as the PRR reviewer in kep.yaml.

Thanks and happy new year

johnbelamaric:
/hold cancel
/approve
/lgtm

@wojtek-t when you return please add a pointer to help folks resolve the failures, as discussed

wojtek-t:
Opened https://github.com/kubernetes/enhancements/pull/2207 for followup.

PR #2183, opened 2020-12-15
Update README.md

wangxy518:


mrbobbytables:
The bit.ly links are what we want to keep. Please don't update links if the redirect in place works/is intentional.

/close

PR #2191, opened 2020-12-16
Migrate some storage KEPs to new template

wojtek-t:
Not status changes or other changes - just migrating existing info to new templates.

This will be useful to strengthen validation of kep.yaml eventually.


wojtek-t:
ref https://github.com/kubernetes/enhancements/issues/2220

PR #2199, opened 2020-12-18
Add some more answers into PRR review in EndPort kep

rikatz:
Signed-off-by: Ricardo Pchevuzinske Katz <ricardo.katz@gmail.com>

### DIFF
--- a/keps/sig-network/2079-network-policy-port-range/README.md
+++ b/keps/sig-network/2079-network-policy-port-range/README.md
@@ -232,20 +232,48 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  
+  Yes. One caveat here is that NetworkPolicies created with EndPort field set 
+  when the feature was enabled will continue to have that field set when the 
+  feature is disabled unless user removes it from the object. 
+  
+  If the value is dropped with the FeatureGate disabled, the field can only 
+  be re-inserted if feature gate is enabled again.
+
+  Rolling back the Kubernetes API Server that does not have this field
+  will make the field not be returned anymore on GET operations,
+  so CNIs relying on the new field wont recognize it anymore.
+
+  If this happens, CNIs will recognize the policy as a single port instead of a 
+  port range, which may break users, which is inevitable but satisfies the 
+  fail-closed requirement.
 
 * **What happens if we reenable the feature if it was previously rolled back?**
-  Nothing. Just need to check if the data is persisted in `etcd` after the 
-  feature is disabled and reenabled or if the data is missed
+  Nothing. 
 
 * **Are there any tests for feature enablement/disablement?**
  
- TBD
+  No - unit tests will be added later.
+
+ ### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+* **How can a rollout fail? Can it impact already running workloads?**
+  Not probably, but still there's the risk of some bug that fails validation, 
+  or conversion function crashes.
+
+* **What specific metrics should inform a rollback?**
+  The increase of 5xx http error count on Network Policies Endpoint
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  This will be done.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+  None
 
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
-
 * **How can an operator determine if the feature is in use by workloads?**
   
   Operators can determine if NetworkPolicies are making use of EndPort creating
@@ -282,10 +310,10 @@ _For GA, this section is required: approvers should be able to confirm the
 previous answers based on experience in the field._
 
 * **Will enabling / using this feature result in any new API calls?**
-  TBD
+  No
 
 * **Will enabling / using this feature result in introducing new API types?**
-  No, unless the new `EndPort` is considered a new API type
+  No
 
 * **Will enabling / using this feature result in any new calls to the cloud 
 provider?**
@@ -295,7 +323,7 @@ provider?**
 the existing API objects?**
 
   - API type(s): NetworkPolicyPorts
-  - Estimated increase in size: 2 bytes for each new `EndPort` specified
+  - Estimated increase in size: 2 bytes for each new `EndPort` value specified + the field name/number in its serialized format 
   - Estimated amount of new objects: N/A
 
 * **Will enabling / using this feature result in increasing time taken by any 
--- a/keps/prod-readiness/sig-network/2079.yaml
+++ b/keps/prod-readiness/sig-network/2079.yaml
@@ -1,0 +1,3 @@
+kep-number: 2079
+alpha:
+  approver: "@wojtek-t"
--- a/keps/sig-network/2079-network-policy-port-range/kep.yaml
+++ b/keps/sig-network/2079-network-policy-port-range/kep.yaml
@@ -18,7 +18,7 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-# latest-milestone: "v1.21"
+latest-milestone: "v1.21"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:


rikatz:
/uncc @dcbw @caseydavenport 

Will assign a PRR approver as soon as I get someone to approve this :)

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  Yes, but CNIs relying on the new field wont recognize it anymore. Also,
```
wojtek-t, line 235: It wouldn't recognize the field only if we rollback the Kubernetes version to version that doesn't have this field.

The question is purely about disabling the enablement, in which case the field is still around (though as you mentioned below it will get cleared on the next update).

So let's descibe those things separately:
- feature disablement (and the consequence of field being removed on next update)
- API disablement

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
...
-  feature is disabled and reenabled or if the data is missed
+  feature is disabled and reenabled. If the object "NetworkPolicy" is updated
+  during this time window between disabling and re-enabling the EndPort field 
+  from this object might be removed by PrepareForUpdate function.
 
 * **Are there any tests for feature enablement/disablement?**
  
- TBD
+ Yes, and the tests are already being implemented in the REST stragegy package
+ for NetworkPolicy.
```
wojtek-t, line 248: Can you point me to them - I think I've never seen enablement/disablement unit tests (we generally test the feature enabled and disablement, but not really what happens when enabling/disabling the feature).

[If we have them that's awesome - I'm looking for examples we can start pointing people to to duplicate the logic]

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
...
 
-### Monitoring Requirements
+ ### Rollout, Upgrade and Rollback Planning
 
 _This section must be completed when targeting beta graduation to a release._
+* **How can a rollout fail? Can it impact already running workloads?**
+  Nothing happens, this is a new Network Policy API Field
```
wojtek-t, line 254: By design - yes.

But in theory you can have a bug that e.g. fails validation if the field is absent. Or crashes on conversion functions or sth. Probably not very likely, but those are in theory possible.

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
...
+
+* **What specific metrics should inform a rollback?**
+  None
```
wojtek-t, line 257: Increased 4xx/5xx http error count on network policies endpoints maybe?

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
...
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  TBD / This tests should be done with a defined CNI, so as soon as CNIs starts
+  supporting the new field (by the graduation criteria, at least 4) this should be tested.
```
wojtek-t, line 261: We should probably do manual update tests even without CNI supporting that to see there aren't unexpected bugs.

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -282,10 +302,10 @@ _For GA, this section is required: approvers should be able to confirm the
 previous answers based on experience in the field._
 
 * **Will enabling / using this feature result in any new API calls?**
-  TBD
+  No
 
 * **Will enabling / using this feature result in introducing new API types?**
-  No, unless the new `EndPort` is considered a new API type
+  No
```
wojtek-t, line 308: nit: L318 - it's actually more than two bytes (e.g. you need also field name/number in its serialized format). [note that it matters much - just nitpicking]

wojtek-t (commented):
Thanks for filling this in - I added a couple relatively small comments.

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Oh, now I get by the context :)

By https://github.com/kubernetes/kubernetes/pull/97058/files#diff-7826f7adbc1996a05ab52e3f5f02429e94b68ce6bce0dc534d1be636154fded3R92 we test if the feature is enabled or disabled, but not disabling and then re-enabling BUT if this is something desired, I can try to work on a specific test suite only disabling and re-enabling the feature gate without any problem.

Like SetFeatureGate to true, create an object, expect the field, then disable the FeatureGate, get the object, check what fields are present, update the object, check if the field is removed.

Sounds a good approach?

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Did some changes, in the same item separated what happens when you disable the FG and what happens if you rollback to a K8s version that does not have this field yet.

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Hum, seems a reasonable metric for me. We can have 5xx errors against netpol endpoints if some validation fails yeap. I'm afraid that 4xx http error can represent something not quite a metric error (like forbidden access to netpol object per RBAC, or maybe a not found object when getting it). but 5xx seems a good metric to me

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Indeed, can do this weekend

rikatz:
@wojtek-t thank you for the kind review :)

I've added your suggestions, and left two items to deal later:
* Make some unit tests with fg enable/disable/enable again (will see how to achieve this)
* Make some manual tests with the PR to provide further information on the KEP Perf review.

Don't know if we can move forward with those changes, otherwise I compromise myself to review this weekend, or not later then next tuesday.

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,42 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  Yes, but disabling the feature gate and updating a Network Policy with existing EndPort field will remove this field.
```
wojtek-t, line 235: nit: can you please split the line to shorten it?

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,42 @@ _This section must be completed when targeting alpha to a release._
...
   Nothing. Just need to check if the data is persisted in `etcd` after the 
-  feature is disabled and reenabled or if the data is missed
+  feature is disabled and reenabled. If the object "NetworkPolicy" is updated
+  during this time window between disabling and re-enabling the EndPort field 
+  from this object might be removed by PrepareForUpdate function.
 
 * **Are there any tests for feature enablement/disablement?**
  
- TBD
+ Not yet, but we can add some tests
```
wojtek-t, line 249: nit:
No - unit tests will be added later

keps/sig-network/2079-network-policy-port-range/README.md
wojtek-t: I don't want to block the PR on it (especially for Alpha) - just mention that it will be done

wojtek-t (commented):
Just some nits - will approve once addressed.

keps/sig-network/2079-network-policy-port-range/README.md
wojtek-t: Yeah - that sounds fine to me (obviously this doesn't block approving this PR - just mention that in the PRR.

rikatz:
@wojtek-t Thank you again, corrected the small nits as requested, squashed commits so we don't polute the repo with small commits.

Thanks!

wojtek-t:
@rikatz - now you also need approval from SIG Network leads

/assign @thockin

rikatz:
Already pinged them in slack :) Thanks!

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,42 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  Yes, but disabling the feature gate and updating a Network Policy with existing
```
thockin, line 235: It should not remove the field - gates should be coded to preserve the contents of the field if it is exists, even if the feature itself is disabled.

caseydavenport: Is this basically a difference in storage vs. what is presented via the API? Seems like we should not remove the end port field from stored data, but depending on the gate we would either present the field or hide it?

Definitely want to maintain the behavior where disabling the gate -> enabling the gate -> disabling the gate leaves you back where you started.

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,42 @@ _This section must be completed when targeting alpha to a release._
...
+  Rolling back the Kubernetes API Server that does not have this field
+  will make the field not be returned anymore on GET operations,
+  so CNIs relying on the new field wont recognize it anymore.
 
 * **What happens if we reenable the feature if it was previously rolled back?**
   Nothing. Just need to check if the data is persisted in `etcd` after the 
-  feature is disabled and reenabled or if the data is missed
+  feature is disabled and reenabled. If the object "NetworkPolicy" is updated
+  during this time window between disabling and re-enabling the EndPort field 
+  from this object might be removed by PrepareForUpdate function.
```
caseydavenport, line 246: what do you mean by "might" be removed? Is there a certain scenario where this would occur or is it unknown whether or not this would / should occur?

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Hum, ok, seems reasonable. Just need in the impl (https://github.com/kubernetes/kubernetes/pull/97058/files#diff-f0064039f0ac16d6f72b07c2afe0fd12232dae1300dda4e3f64c72209666324bR102) to change this behavior to only drop IF the old netpol already does not have the field :)

Will change here the KEP and then the impl :)

About what's shown in the API I honestly don't know if it's possible to persist the data but not show it, maybe Tim knows

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: was not known, but as per previous comment it's also not desired, so will remove this from here, and implement the change to drop only if this is a new field, and not present on the old netpol object.

This is probably going also to take some more time as it's good now to implement the tests about enable->disable->enable FG

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: Here a quick question: what if the field is existing on the old object, and I only change the value? I'm ok about keeping the field or allowing the user to drop it, but what's the expected behavior when FG is disabled and the user changes only the contents?

rikatz:
@caseydavenport @thockin made some changes, but left the question about what should happen when the feature gate is disabled, the old object differs from the new one only on the EndPort value :)

Thanks

keps/sig-network/2079-network-policy-port-range/README.md
wojtek-t: The "and fieldNotUsed" pattern is what we use in this case, see e.g. this one as an example:
https://github.com/kubernetes/kubernetes/blob/f6e04cd3ad1f3a1a00fd091e1524d2d27dc56bd7/pkg/api/pod/util.go#L442

rikatz: Oh thank you! saved me a lot of time! :)

rikatz:
@wojtek-t so I've added in https://github.com/kubernetes/kubernetes/pull/97058/commits/0e11fc73b53e0c7a01f676f83f8b97ff517c1134#diff-7826f7adbc1996a05ab52e3f5f02429e94b68ce6bce0dc534d1be636154fded3R245 the Feature Enable/Disable/Enable test.

Hopefully this is what you expect :) 

I think I'll need to maybe refactor that and create a helper function instead of keep doing Update and verify in everything (it's pretty ugly now, I think) but let me know if this works as a test.

@thockin @caseydavenport I've also added the check of if the field is already present, do not drop it.

If this is OK to you folks, maybe the KEP can be updated and we can continue working on the EndPort PR :)

Thank you!

wojtek-t:
> @wojtek-t so I've added in a2e29d29 the Feature Enable/Disable/Enable test.
> Hopefully this is what you expect :)

Yes - this is great, thanks!
Eventually, we would like to have those kind of tests to test more behavior, but it's already huge step in that direction.

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,45 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  Yes, fields created before feature being disabled will not be changed/dropped 
+  if the old value is kept on the updated object. If the user removes the field
```
wojtek-t, line 236: It's probably because of my english, but I think I'm not fully parsing this sentence.

Maybe let's say:
`
Yes. One caveat here is that NetworkPolicy objects create with EndPort field set when the
features was enabled will continue to have that field set when the feature is disabled,
but will not have any effect.
`

WDYT?

[i.e. change this whole paragraph to what I wrote]

keps/sig-network/2079-network-policy-port-range/README.md
wojtek-t: Let's remove this part - this will will not be removed.

keps/sig-network/2079-network-policy-port-range/README.md
rikatz: hum, my concern here is that it actually will have effect, if the CNI knows the field.

As an example, let's say Calico is using the last client-go which knows the EndPort field. Once I disable the Feature Gate, per PrepareForCreate and PrepareForUpdate I will not be able to add a new EndPort field (if the EndPort field is not in use in that object), but existing EndPort fields can be changed or removed. So, disabling the FeatureGate only blocks users to add/change on objects that does not have this field, but does not block the CNI to consume this, or even the user to Update or Delete it.

Before reaching your comment here, I've re-written the paragraph as the following:

```
Yes, fields created before feature gate being disabled will not be dropped unless the user removes it from the object.
  If the user removes the field from the object it will be dropped and the 
  field can only be re-inserted if feature gate is enabled again.
```
How does this sounds?

rikatz:
@wojtek-t @thockin @caseydavenport added a new commit with some more corrections :)

Thanks

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,40 @@ _This section must be completed when targeting alpha to a release._
 
 * **Can the feature be disabled once it has been enabled (i.e. can we roll back
   the enablement)?**
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  Yes, fields created before feature gate being disabled will not be dropped unless the user removes it from the object.
```
wojtek-t, line 235: This isn't exactly true - if the feature is disabled there shouldn't be a way to drop the field even...

rikatz: Hum, okay, thinking here loudly (sorry for being too verbose!) 

By the impl, when we edit an object that have the NetPol EndPort, but with the FG disabled, the EndPort field appears there (like in kubectl edit)

When I remove the field, PrepareForUpdate will run, and skip the dropEndPort because the field is in use, so it will follow its normal flow. (https://github.com/kubernetes/kubernetes/blob/f0601ac64d39c601a66f8f6e6f69d55995c2b32a/pkg/registry/networking/networkpolicy/strategy.go#L62-L64)

The normal flow is to take the old object, and if it's changed add +1 into Generation and then apply the desired changes.

So probably I'll need to also add into something after the PrepareForUpdate to not drop the EndPort field, even if the user has removed it?

wojtek-t: This is a tricky topic - I think I didn't realize one thing when writing this comment.

Looking into other examples across the codebase, the only thing we're doing is we're preventing from setting the field when the FG is switched on and silently dropping it if it was set. We allow to explicitly dropping it at any time.


So, I just actually suggest a simple nit:
```
Yes. One caveat here is that NetworkPolicies created with EndPort fiel set when the feature was enabled will continue to have that field set when the feature is disabled unless user removes it from the object.
```

[The sentence below kind-of duplicates the message]

rikatz: Great :)

I've improved the sentence here consolidating your suggestion and improving mine (I don't know where was my head when  I wrote the next sentence, but it took me 3 reads to understand what I wanted to say).

Made another commit here

wojtek-t:
/lgtm
/approve

@thockin @caseydavenport for SIG approval

wojtek-t:
@rikatz - also please squash commits

rikatz:
Squashed, pinged Tim on slack so he can take a look when gets some time :)

keps/sig-network/2079-network-policy-port-range/README.md
```diff
@@ -232,20 +232,44 @@ _This section must be completed when targeting alpha to a release._
...
-  Yes, but CNIs relying on the new field wont recognize it anymore
+  
+  Yes. One caveat here is that NetworkPolicies created with EndPort field set 
+  when the feature was enabled will continue to have that field set when the 
+  feature is disabled unless user removes it from the object. 
+  
+  If the value is dropped with the FeatureGate disabled, the field can only 
+  be re-inserted if feature gate is enabled again.
+
+  Rolling back the Kubernetes API Server that does not have this field
```
thockin, line 243: Maybe clarify: The net impact of a rollback is that a port-range reverts to a single port.  This may break users, but that is inevitable in such a case.  It satisfies the fail-closed requirement.

thockin (commented):
ping me when small change is done.

rikatz:
@thockin The change is done. I've already rebased but the only change was what you've suggested on lines 243:249.

Thanks!

thockin (commented):
Thanks!

/lgtm
/approve

PR #2203, opened 2020-12-21
add kep for dynamic scoring plugin

yingnanzhang666:
DynamicScoring plugin provides an option to add some system-level preferences dynamically when we run into some temporary conditions.


yingnanzhang666:
@ahg-g @Huang-Wei @johnbelamaric PTAL

PR #2235, opened 2021-01-06
convert to new dir structure

deads2k:
This is on the path to moving to GA.  I need to

1. convert to new kep format
2. answer PRR questions
3. indicate that we'll use integration tests in lieu of e2e tests for this particular case and explain why.  Also, retarget to GA
4. retarget the issue
5. update the code to be GA in feature gate
6. indicate complete in kep


enj:
/lgtm
/approve

This is a no-op towards the new format.

PR #2244, opened 2021-01-08
Update Ephemeral Containers KEP for beta in 1.21

verb:
This updates graduation criteria and completes PRR sections.

### DIFF
--- a/keps/sig-node/277-ephemeral-containers/README.md
+++ b/keps/sig-node/277-ephemeral-containers/README.md
@@ -64,13 +64,13 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 - [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
 - [x] (R) KEP approvers have approved the KEP status as `implementable`
 - [x] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
-- [ ] (R) Graduation criteria is in place
-- [ ] (R) Production readiness review completed
-- [ ] Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
-- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [x] (R) Production readiness review completed
+- [x] Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
+- [x] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [x] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 <!--
 **Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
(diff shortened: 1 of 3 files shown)


keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -675,60 +663,109 @@ _This section must be completed when targeting alpha to a release._
...
 
 * **How can an operator determine if the feature is in use by workloads?**
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+
+  The kubelet has a gauge metric `kubelet_running_containers` that we can
+  augment to contain a count of running ephemeral containers.
+
+  TODO(verb): Open an issue to create this metric after discussing with PRR
```
johnbelamaric, line 749: Yes, please do.

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -675,60 +663,109 @@ _This section must be completed when targeting alpha to a release._
...
-    - Components exposing the metric:
+  - [x] Metrics
+    - Metric name: `apiserver_request_total{component="apiserver",resource="pods",subresource="ephemeralcontainers"}` (apiserver), `kubelet_container_errors_total{type="Ephemeral"}` (kubelet, Proposed)
+    - [Optional] Aggregation method: Aggregate by container type
+    - Components exposing the metric: kubelet
   - [ ] Other (treat as last resort)
     - Details:
 
+  Note that the kubelet SLI for this feature is a counter that increments upon
+  failure to create an ephemeral container. Right now the kubelet only surfaces
```
johnbelamaric, line 762: Would this have noise due to user error? That is, can we differentiate between a kubelet failure that should be of concern to the operator vs the user using a bad image or something? I guess we have the same issue for regular containers, so we should follow the same logic we have there.

keps/sig-node/277-ephemeral-containers/README.md
verb: Done. Opened kubernetes/kubernetes#97974 and linked from the KEP.

keps/sig-node/277-ephemeral-containers/README.md
verb: Ah, that's a good point. I agree that this "server error" vs "user error" is important here, but I'm not sure of the current behavior of the kubelet yet.  I [brought it up in the metrics issue](https://github.com/kubernetes/kubernetes/issues/97974#issuecomment-762356411) to discuss with sig-instrumentation and will make sure it's addressed.

verb:
Hi @dchen1107, I've changed some of the graduation criteria, delaying some of the test implementation and removing user feedback gathering, which I haven't found to be a strong signal in other features. Could you review this and the PRR updates and let me know if you approve of this direction? Thanks!

/assign @dchen1107

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -583,38 +583,27 @@ None of the tests for this feature are unusual or tricky.
...
+- [x] A task on https://kubernetes.io/docs/tasks/ describes how to troubleshoot
   a running pod using Ephemeral Containers.
-- [ ] A survey sent to early adopters doesn't reveal any major shortcomings.
 
 #### Beta -> GA Graduation
 
 - [ ] Ephemeral Containers have been in beta for at least 2 releases.
 - [ ] Ephemeral Containers see use in 3 projects or articles.
-- [ ] Ephemeral Container creation is covered by conformance tests
+- [ ] Ephemeral Container creation is covered by e2e-node tests.
```
dchen1107, line 599: Can we list this as beta related working item?

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -675,60 +663,113 @@ _This section must be completed when targeting alpha to a release._
...
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+
+  We will create a new gauge metric that's updated during kubelet's reconcile
+  of `v1.Pod` to track the number containers scheduled to this node in the API.
+  This will be slightly different than the existing
+  `kubelet_running_containers`, which describes the kubelet's representation of
+  containers, and will be able to label the metrics with fields that are only
+  available in the API object, such as type of container.
```
dchen1107, line 751: Are we planning to replace kubelet_running_containers with v1.Pod? It can be handled separately of course.

dchen1107:
/lgtm with some small comments.

Will approve it once @johnbelamaric is done with the review. Thanks!

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -675,60 +663,113 @@ _This section must be completed when targeting alpha to a release._
...
+  during an update. The case for this feature is slightly more complicated
+  because the field may not be set by the default update resource. This logic
+  has been in place for several releases.
+
+  The apiserver and kubelet have existing upgrade tests
+  (https://testgrid.k8s.io/google-gce-upgrade), but these likely don't exercise
+  ephemeral containers. We'll investigate whether it's possible to add ephemeral
+  containers to these existing tests.
+
+  TODO(verb): create issue to update existing upgrade tests after discussing
```
johnbelamaric, line 732: You say you followed the instructions, but do you have unit tests for it, and manual testing to show that after rolling back there is no problem, and after rolling forward again what happens?

verb: Yes, this behavior is tested in [TestDropEphemeralContainers](https://github.com/kubernetes/kubernetes/blob/v1.20.2/pkg/api/pod/util_test.go#L1450)

I tend to mix up upgrade/downgrade and feature enable/disable testing. enable/disable testing is the thing I care most about testing, and it matters in context of upgrades when the feature flag default changes. I realize now this section that I wrote in the "upgrade->downgrade->upgrade" path is more correctly about "disabled->enabled->disabled". I'm not sure if this was your intent for this section.

Anyway, I've done feature enablement testing manually, and I've just run an extra "disabled at 1.20 -> enabled at 1.21-alpha -> disabled at 1.20 -> enabled at 1.21-alpha" test, and everything looks ok.

I haven't done "enabled at 1.20 -> enabled at 1.21-alpha" testing since implementation hasn't changed in several releases, but I can if you'd like.

I've moved the bit about unstable fields to the enable/disable feature section and added descriptions of the unit and manual tests.

johnbelamaric:
PRR looks good now, thanks! I will approve after SIG node approval (otherwise it approves everything)

verb:
Thanks John & Dawn!

@dchen1107 I've squashed and rebased and this PR is ready for your approval.

PR #2245, opened 2021-01-08
Add KEP for Indexed Job in implementable status

alculquicondor:
Ref #2214  kubernetes/kubernetes#97169

/sig apps

### DIFF
--- a/keps/sig-apps/2214-indexed-job/README.md
+++ b/keps/sig-apps/2214-indexed-job/README.md
@@ -1,0 +1,594 @@
+# KEP-2214: Indexed Job
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [JobSpec API](#jobspec-api)
+  - [Pod detail](#pod-detail)
+  - [Job completion and restart policy](#job-completion-and-restart-policy)
+  - [Job parallelism](#job-parallelism)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP extends kubernetes with user-friendly support for running
+embarrassingly parallel jobs.
+
+Here, parallel means multiple pods. By embarrassingly parallel, it means that
+the pods have no dependencies between each other.
+In particular, neither ordering between pods nor gang scheduling are supported.
+
+We propose the addition of completion indexes into the Pods of a *Job
+[with fixed completion count]* to support running embarrassingly parallel
+programs, with a focus on ease of use.
+We call this new Job pattern an *Indexed Job*, because each Pod of the Job
+specializes to work on a particular index, as if the Pods where elements of an
+array.
+
+[with fixed completion count]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#parallel-jobs
+
+## Motivation
+
+Users can use some [Job patterns] to run embarrassingly parallel Jobs, but those
+approaches have downsides:
+
+- The queue patterns require setting up an external queue service and modifying
+  the Job binary to be able to connect to the queue.
+  Depending on the implementation, it is prone to race conditions when
+  coordinating which Pod works on which item.
+- The template pattern doesn't scale when the parallelism level is too high,
+  in terms of job creation and querying status.
+
+Due to these reasons, workloads where each Pod just needs a unique and
+ordered completion index, are hard to adapt to the existing Job patterns.
+
+The lack of support for this pattern in k8s forces users to implement their
+own APIs and controllers or adopt third party implementations. Each
+implementation splits the ecosystem, making it harder for higher level systems
+for Job queueing or workflows to support all of them.
+
+[Job patterns]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-patterns
+
+### Goals
+
+- Support the *array Job* pattern by adding completion indexes to each Pod
+  of a Job in *fixed completion count* mode.
+
+### Non-Goals
+
+- Support for work lists, where each Pod receives a different element of a
+  static list. This can be implemented by users from completion indexes.
+- Support for completion index in non-parallel Jobs or Jobs with a work queue.
+- All-or-nothing scheduling.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a Job author, I can create an array Job where each Pod receives an ordered
+completion index. I can use the index in my binary through an environment
+variable or a file to statically select the load the Pod should work on.
+
+```yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: parallel-work
+spec:
+  completions: 100
+  parallelism: 100
+  template:
+    spec:
+      containers:
+      - name: task
+        image: registry.example.com/processing-image
+        command: ["./process",  "--index", "$INDEX"]
+        env:
+        - name: INDEX
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['batch.alpha.kubernetes.io/job-completion-index'] 
+```
+
+### Notes/Constraints/Caveats (Optional)
+
+* An earlier proposal for [indexed Job]
+suggested the support for work lists, i.e. passing different parameters to each
+Pod. We decided to leave this out of the proposal to keep it simple and
+because work lists can be implemented in a startup script using the completion
+index as building block.
+* The semantics of an indexed Job are similar to a StatefulSet, in the sense
+that Pods have an associated index.
+However, the APIs have a major difference: a StatefulSet doesn't have completion
+semantics, as opposed to Jobs.
+
+[indexed Job]: https://github.com/kubernetes/community/blob/b21d1b27c8c748bf81283c2d89cde2becb5f2709/contributors/design-proposals/apps/indexed-job.md
+
+### Risks and Mitigations
+
+Jobs have a known issue in which more than one Pod can be started even if
+parallelism and completion are set to 1 ([reference]). In the case of indexed
+Jobs, this translates to more than one Pod having the same index.
+
+Just like for existing Job patterns, workloads have to handle duplicates at the
+application level.
+
+[reference]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures
+
+## Design Details
+
+### JobSpec API
+
+The JobSpec gets the field `completionMode` to control whether the Job should	
+be treated as an Indexed Job.	
+
+```golang
+// CompletionMode specifies how Pod completions of a Job are tracked.
+type CompletionMode string
+
+const (
+	// NonIndexedCompletion means that Pod completions of a Job are
+	// indistinguishable from each other.
+	NonIndexedCompletion CompletionMode = "NonIndexed"
+
+	// IndexedCompletion means that each Pod completion of a Job is tracked
+	// individually, being associated to a completion index.
+	IndexedCompletion CompletionMode = "Indexed"
+)
+
+type JobSpec struct {	
+  ...	
+  // CompletionMode specifies how Pod completions are tracked. It can be	
+  // `NonIndexed` (default) or `Indexed`.	
+  //
+  // `NonIndexed` means that each Pod completion is homologous to each other.	
+  // The Job is considered complete when there have been .spec.completions	
+  // successful completions.	
+  //
+  // `Indexed` means that each Pod completion needs to be tracked individually;	
+  // each Pods gets an associated completion index, which is available in the
+  // annotation	`batch.alpha.kubernetes.io/job-completion-index`.	
+  // The Job is considered complete when there is one successful Pod for each	
+  // index in the range 0 to (.spec.completions - 1).	
+  // When value is `Indexed`, .spec.completions must have a non-zero positive
+  // value and `.spec.parallelism` must be less than or equal to 10^5.
+  //
+  // More completion modes can be added in the future. If a Job controller
+  // observes a mode that it doesn't recognize, it manages the Job as in
+  // `NonIndexed`.
+  CompletionMode CompletionMode
+}	
+```
+
+As the comment describes, when `.spec.completionMode = "Indexed"`, the
+`.spec.completions` must be:
+
+- a non-zero positive value. This is to trigger Job management strategy for
+  *fixed completion count*. That is, `Indexed` mode cannot	be used for work
+  queue patterns.	
+- less than or equal to `10^6`. This is to guarantee that we can keep track of
+  completions per-index in the Job status in the future.
+
+### Pod detail
+
+The Pod and PodSpec APIs don't get any new fields. However, Pods created for
+Indexed Jobs get the annotation `batch.alpha.kubernetes.io/job-completion-index`
+with a value equal to its completion index. The annotation is immutable.
+
+The annotation can be accessed through the downward API as a file or environment
+variable.
+
+For user convenience, the Job controller adds the completion index as an
+environment variable through the downward API. That is, the Job controller
+creates Pods like so:
+
+```yaml
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+    - name: test-container
+      env:
+        - name: JOB_COMPLETION_INDEX
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['batch.alpha.kubernetes.io/job-completion-index'] 
+```
+
+The Job controller doesn't add the environment variable if there is a name
+conflict with an existing environment variable. Users can specify other
+environment variables for the same annotation.
+
+### Job completion and restart policy
+
+When dealing with Indexed Jobs, the Job controller keeps track of Pod
+completions for each index from 0 to `.spec.completions - 1`.
+Once the controller notices that a Pod finishes successfully, it will not create
+another Pod with the same index.
+
+The Job controller considers a Job completed when there is at least one
+successful Pod for each completion index.
+
+If an entire Pod fails, such as when the Pod gets kicked off the node or if a
+container of the Pod fails and `.spec.template.spec.restartPolicy = "Never"`,
+the Job controller gets the completion index for the failed Pod and creates a
+new Pod with the same index. The application needs to handle restarts in a
+different Pod.
+
+The Pod might not be immediately replaced, as the number of active Pods could
+have hit the parallelism limit. Once another Pod finishes, the Job controller
+syncs the Job, scanning for unsatisfied indexes and creates the missing Pods for
+them.
+
+The kubelet handles container restarts as usual, according to the
+`spec.template.spec.restartPolicy`.
+
+<<[UNRESOLVED TBD Beta: Track completed indexes in Job status]>>
+Once [kubernetes/kubernetes#28486](https://github.com/kubernetes/kubernetes/issues/28486)
+is resolved:
+
+The Job controller keeps track of completed indexes in
+`.status.completedIndexes`, a string that represents a list of numbers in a
+compressed format. For example, if a Job has completed indexes 2, 3, 4, 6 and 7,
+the list looks like:
+
+```golang
+CompletedIndexes: "2-4,6-7"
+```
+
+The `kubectl describe` command crops the list of indexes if it's too long:
+
+```
+Completed Indexes: [1-25,28,30-32,...]
+``` 
+<<[/UNRESOLVED]>>
+
+### Job parallelism
+
+A user can change the number of active Pods for a Job changing `.spec.parallelism`
+(note that `.spec.completions` is an immutable field).
+
+When starting a Job or increasing the parallelism, the Job controller creates
+Pods with lower completion index first, as long as there is no other completed
+or running Pod with the same index. This is to make the controller behavior more
+predictable. We do not offer guarantees on creation order based on completion
+index.
+
+Reducing parallelism is unaffected by completion index.
+
+### Test Plan
+
+Unit, integration and E2E tests cover the following Indexed Job mechanics:
+
+  - Creation with indexed Pod names and index annotations.
+  - Scale up and down.
+  - Pod failures.
+  
+Additionally, we add unit tests for API defaulting and validation, with feature
+gate enabled and disabled.
+  
+### Graduation Criteria
+
+#### Alpha
+
+- Complete features:
+  - Completion index in annotation
+  - Restart policy
+
+#### Alpha -> Beta Graduation
+
+- Complete features:
+  - Tracking completions by index in Job status
+- Gather feedback from end users and operators' developers. Open questions:
+  - Are stable Pod names necessary?
+- Tests are in Testgrid and linked in KEP
+
+#### Beta -> GA Graduation
+
+- E2E test graduates to conformance
+- 2 examples of operators using Indexed Jobs.
+
+### Upgrade / Downgrade Strategy
+
+In the event of a kube-controller-manager upgrade, there should not be any
+existing Indexed Jobs with running Pods.
+
+In the event of a downgrade, existing Indexed Jobs will run as NonIndexed Jobs.
+The controller will track the existing Pods ignoring the completion index. New
+Pods will be created without a completion index. Existing workloads that
+expected the completion index will fail. But this is expected in a downgrade.
+
+In the event of an upgrade after a downgrade, the controller will remove
+existing Pods without completion index for existing Indexed Jobs, without they
+counting towards failures or completions. The controller will create new Pods
+with indexes. This might cause a load spike in the cluster.
+
+If, instead of a downgrade, the cluster administrator disables the feature gate:
+
+  - kube-apiserver clears `.spec.completionMode` for new Jobs at creation time.
+    That is, all new Jobs are interpreted as `NonIndexed`.
+  - kube-controller-manager skips syncing existing Indexed Jobs and emits a
+    warning event. More specifically, the controller does not create new Pods,
+    track completion nor update status of the Job.
+  
+The above guarantees that the controller never creates Pods for Indexed Jobs
+without a completion index.
+
+### Version Skew Strategy
+
+This features has no node runtime implications.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: IndexedJob
+    - Components depending on the feature gate:
+      - kube-apiserver
+      - kube-controller-manager
+
+* **Does enabling the feature change any default behavior?**
+
+  No. Jobs need to opt-in with `.spec.completionMode=Indexed`.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  
+  Yes. Using the feature gate is the recommended way.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+
+  The Job controller starts managing Indexed Jobs again.
+  More details covered in [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy).
+
+* **Are there any tests for feature enablement/disablement?**
+
+  Yes, unit and integration test for the feature enabled, disabled and
+  transitions.
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
+
+* **What specific metrics should inform a rollback?**
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  Describe manual testing that was done and the outcomes.
+  Longer term, we may want to require automated upgrade/rollback tests, but we
+  are missing a bunch of machinery and tooling and can't do that now.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+fields of API types, flags, etc.?**
+  Even if applying deprecation policies, they may still surprise some users.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+  checking if there are objects with field X set) may be a last resort. Avoid
+  logs or events for this purpose.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine 
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+  - [ ] Other (treat as last resort)
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
+  - 99% percentile over day of absolute value from (job creation time minus expected
+    job creation time) for cron job <= 10%
+  - 99,9% of /health requests per day finish with 200 code
+
+* **Are there any missing metrics that would be useful to have to improve observability 
+of this feature?**
+  Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+  implementation difficulties, etc.).
+
+### Dependencies
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **Does this feature depend on any specific services running in the cluster?**
+  Think about both cluster-level services (e.g. metrics-server) as well
+  as node-level agents (e.g. specific version of CRI). Focus on external or
+  optional services that are needed. For example, if this feature depends on
+  a cloud provider API, or upon an external software-defined storage or network
+  control plane.
+
+  For each of these, fill in the following—thinking about running existing user workloads
+  and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+
+
+### Scalability
+
+_For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them._
+
+_For beta, this section is required: reviewers must answer these questions._
+
+_For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
+
+  No.
+
+* **Will enabling / using this feature result in introducing new API types?**
+
+  No.
+
+* **Will enabling / using this feature result in any new calls to the cloud 
+provider?**
+
+  No.
+
+* **Will enabling / using this feature result in increasing size or count of 
+the existing API objects?**
+
+  Yes.
+  
+  - API type(s): Job
+  - Estimated increase in size: new field of about 30 bytes.
+  
+  - API type(s): Pod, only when created with the new completion mode.
+  - Estimated increase in size: new annotation of about 50 bytes.
+
+* **Will enabling / using this feature result in increasing time taken by any 
+operations covered by [existing SLIs/SLOs]?**
+
+  No.
+
+* **Will enabling / using this feature result in non-negligible increase of 
+resource usage (CPU, RAM, disk, IO, ...) in any components?**
+
+  Additional CPU and memory increase in the controller-manager is negligible
+  and restricted to Jobs using the new completion mode.
+
+### Troubleshooting
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How does this feature react if the API server and/or etcd is unavailable?**
+
+* **What are other known failure modes?**
+  For each of them, fill in the following information by copying the below template:
+  - [Failure mode brief description]
+    - Detection: How can it be detected via metrics? Stated another way:
+      how can an operator troubleshoot without logging into a master or worker node?
+    - Mitigations: What can be done to stop the bleeding, especially for already
+      running user workloads?
+    - Diagnostics: What are the useful log messages and their required logging
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+
+* **What steps should be taken if SLOs are not being met to determine the problem?**
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
+
+## Implementation History
+
+* 2021-01-08: First version of the KEP in provisional status. Design Details
+  completed.
+
+## Drawbacks
+
+* Adds more complexity to the Job controller in terms of Pod and Pod status
+  management, as it introduces a new mode.
+* Failed Pods are removed before being replaced by new Pods, reducing end-user
+  debugging capabilities. However, failed Pod persists when the whole Job fails.
+
+## Alternatives
+
+- **Leave Indexed Job to third-party implementations**
+
+  The major painpoint is that this leaves Pod management to the third-party
+  implementation. With different implementations, the ecosystem is split, making
+  it harder for higher level Job orchestration frameworks to support all of
+  them.
+  
+  On the other hand, with the Indexed Job native support in core k8s,
+  third-party implementations can focus on application level APIs, using the Job
+  API as their underlying Pod management mechanism.
+  
+- **Completion Index in the Pod Name**
+
+  Completion indexes could also be part of the Pod name, leading to stable Pod
+  names. This allows 2 things:
+  - Uniqueness for each completion index, freeing applications from having to
+    handle duplicated indexes.
+  - Predictable hostnames, which benefits applications that need to communicate
+    to Pods of a Job (or among Pods of the same Job) without having to do
+    discovery.
+  
+  Stable pod names require the Job controller to remove failed Pods before
+  creating a new one with the same index. This has some downsides:
+  - Removing Job Pods is a breaking change. But this can be done if it's a new
+    Job execution mode accessible through a JobSpec field.
+  - Currently, the Job controller uses the tombstones of failed Pods to track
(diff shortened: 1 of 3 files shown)


alculquicondor:
cc @ahg-g @soltysh @erictune

keps/sig-apps/2214-array-job/README.md
```diff
@@ -0,0 +1,572 @@
...
+[Job patterns]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-patterns
+
+### Goals
+
+- Support the *array Job* pattern by adding completion indexes to each Pod
+  of a Job in *fixed completion count* mode.
+- Track the completion for each Job Pod per completion index.
+- Handle varying parallelism by creating Pods with lower indexes first and
+  removing Pods with higher indexes first. Jobs that do not require completion
+  indexes preserve their downscaling behavior.
```
wojtek-t, line 164: This already described the solution, not just what we want to achieve.

keps/sig-apps/2214-array-job/README.md
```diff
@@ -0,0 +1,572 @@
...
+<<[UNRESOLVED Keeping track of number of failures]>>
+In non-array Jobs, the Job controller reports the number of failures in the
+Job status from the number of existing failed Pods.
+
+Since array Jobs remove the failed Pods, the information needs to be tracked in
+other ways.
+One solution is keeping a recreation counter as an annotation in the Pod:
+`alpha.batch.k8s.io/job-pod-recreate-count`. Then, before recreating a Pod,
+the counter is extracted, incremented and added to the new Pod. The counters
+are the source-of-truth for Job failures.
```
wojtek-t, line 312: What if the conroller gets restarted between removing the previous pod and creating the new one?

alculquicondor: That would be an scenario where information is lost.

However, note that we also loose information in the current architecture: when the Pod garbage collector reaches the threshold and starts removing failed (or even succeeded) Pods.

keps/sig-apps/2214-array-job/README.md
alculquicondor: so just: "Handle varying parallelism"

ahg-g: I think the first goal alone is sufficient, the rest is implementation details.

alculquicondor: Removed

keps/sig-apps/2214-array-job/README.md
alculquicondor: We decided to remove stable names from the proposal. It might be reconsidered as a future related enhancement, after kubernetes/kubernetes#28486 is resolved and from user feedback.

I moved the details to "Alternatives"

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+- E2E test graduates to conformance
+- 2 examples of operators using Indexed Jobs.
+
+### Upgrade / Downgrade Strategy
+
+The Job controller keeps track of running or succeeded indexes using the
+annotation. In the event of a kube-controller-manager upgrade, some running or
+succeeded Pods might exist without a completion index.
+
+<<[UNRESOLVED handling Jobs with existing Pods without completion index ]>>
```
alculquicondor, line 239: @wojtek-t WDYT?

ahg-g, line 239: Introducing the annotation shouldn't impact existing workloads unless they explicitly use it. I don't expect that users would create an indexed job that is tolerant to not having the index. So I don't really see the need for addressing this scenario.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+### Goals
+
+- Support the *array Job* pattern by adding completion indexes to each Pod
+  of a Job in *fixed completion count* mode.
+
+### Non-Goals
+
+- Support for work lists, where each Pod receives a different element of a
+  static list. This can be implemented by users from completion indexes.
+- Support for completion index in non-parallel Jobs or Jobs with a work queue.
```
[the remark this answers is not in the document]
alculquicondor: This is still the case, completion index for Jobs with a work queue is a non-goal.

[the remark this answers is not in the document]
ahg-g: ah, so that is the case where only parallelism is set but not completions.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+- All-or-nothing scheduling.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a Job author, I can create an array Job where each Pod receives a unique and
```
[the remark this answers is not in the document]
alculquicondor: No. This was a leftover detail from previous version.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+              fieldPath: metadata.annotations['alpha.batch.k8s.io/job-completion-index'] 
+```
+
+### Notes/Constraints/Caveats (Optional)
+
+* An earlier proposal for [indexed Job]
+suggested the support for work lists, i.e. passing different parameters to each
+Pod. We decided to leave this out of the proposal to keep it simple and
+because work lists can be implemented in a startup script using the completion
+index as building block.
```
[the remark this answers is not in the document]
alculquicondor: If it's a short list, it can live in the startup script. But it could also live in a ConfigMap or some external source.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+* The semantics of an indexed Job are similar to a StatefulSet, in the sense
+that Pods have an associated index and Pods.
```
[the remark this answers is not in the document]
alculquicondor: A leftover :) Removed

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+However, the APIs have a major difference: a StatefulSet doesn't have completion
+semantics, as opposed to Jobs.
+
+[indexed Job]: https://github.com/kubernetes/community/blob/b21d1b27c8c748bf81283c2d89cde2becb5f2709/contributors/design-proposals/apps/indexed-job.md
+
+### Risks and Mitigations
+
+Jobs have a known issue in which more than one Pod can be started even if
+parallelism and completion are set to 1 ([reference]). In the case of indexed
+Jobs, this translates as more than one Pod having the same index.
```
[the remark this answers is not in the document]
alculquicondor: Done

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+### Pod detail
+
+The Pod and PodSpec APIs don't get any new fields. However, Pods created for
+Indexed Jobs get the annotation `alpha.batch.k8s.io/job-completion-index` with a
+value equal to its completion index. The annotation is immutable.
+
+The annotation can be accessed through the downward API as a file or environment
+variable.
+
+### Job completion and restart policy
```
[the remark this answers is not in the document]
alculquicondor: Added.
I don't think the change is significant enough to be a new mode. Now we are tracking number of completed Pods and you can have that number be greater than `.spec.completions`. That doesn't change.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+When dealing with Indexed Jobs, the Job controller keeps track of Pod
+completions for each index from 0 to `.spec.completions - 1`. When a Pod
+finishes successfully, the controller doesn't create another Pod with the same
+completion index. The Job controller doesn't remove succeeded or failed Pods.
+
+If an entire Pod fails, such as when the Pod gets kicked off the node or if a
+container of the Pod fails and `.spec.template.spec.restartPolicy = "Never"`,
+the Job controller gets the completion index for the failed Pod and creates a
+new Pod with the same index. The application needs to handle restarts in a
```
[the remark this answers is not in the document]
alculquicondor: Once a Pod finishes (failed or succeeded) we trigger another sync. So that's covered. Adding some notes.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+### Job scaling
+
+Jobs can be scaled up and down by varying `.spec.parallelism` (note that
+`.spec.completions` is an immutable field).
+
+When starting or scaling up a Job, the Job controller creates Pods with lower
+completion index first, as long as there is no other completed or running Pod
+with the same index. This is to make the controller behavior more
+predictable. We do not offer any guarantees.
```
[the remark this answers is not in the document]
alculquicondor: Done

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+
+The Job controller keeps track of running or succeeded indexes using the
+annotation. In the event of a kube-controller-manager upgrade, some running or
+succeeded Pods might exist without a completion index.
+
+<<[UNRESOLVED handling Jobs with existing Pods without completion index ]>>
+We have 3 options to handle Jobs where some Pods exist without completion index:
+1. Do not track completion indexes for Jobs that have at least one running or
+   succeed Pod without a completion index.
+   This is acceptable because Jobs are not long running. Eventually
```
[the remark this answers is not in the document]
alculquicondor: Fair point. Changed

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: This is more about the controller itself. It needs to do something with the existing Pods in terms of completion index tracking.

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
alculquicondor: Correct. Unfortunately we don't have another official name for this mode in the documentation.

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
ahg-g: but how would the pods access it? pass the whole list in a separate flag I guess?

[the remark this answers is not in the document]
alculquicondor: correct, or a static list in a bash script. There's an example in here https://github.com/kubernetes/community/blob/b21d1b27c8c748bf81283c2d89cde2becb5f2709/contributors/design-proposals/apps/indexed-job.md#how-kubectl-creates-job-specs

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+## Design Details
+
+### Indexed Jobs
+
+Indexed Jobs are Jobs [with fixed completion count], that is, Jobs that have a
+non-nil `.spec.completions` field.
+
+### Pod detail
+
+The Pod and PodSpec APIs don't get any new fields. However, Pods created for
```
erictune, line 174: Did you consider the alternative where you do change the API such as by adding `.spec.indexed` boolean?

Some pros of that approach are:

1. Could use a different path in the control code for indexed jobs.  Reduces risk that the new path in the controller hurts existing multi-completition jobs that don't need indexes.
1. Upgrade downgrade is simple:  On downgrade, all the ones with ".spec.indexed" stop working.  On upgrade, those without ".spec.indexed" don't do anything different.
2. Can inject index without modifying existing pods see subsequent comments).

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+Jobs can be scaled up and down by varying `.spec.parallelism` (note that
+`.spec.completions` is an immutable field).
+
+When starting or scaling up a Job, the Job controller creates Pods with lower
+completion index first, as long as there is no other completed or running Pod
+with the same index. This is to make the controller behavior more
+predictable. We do not offer any guarantees.
+
+Scale down is unaffected by completion index.
+
```
erictune, line 217: Can you show how `.status` changes and how `kubectl describe` changes?

In the case of an indexed job, I think something like this would be nice:
```
$ kubectl describe jobs/parallel-work
Name:           countdown
Namespace:      default
...
Parallelism:    100
Completions:    10
Pods Statuses:  10 Running / 27 Succeeded / 0 Failed
Completed Indexes: [1-25,27-29]
```

While this list could get somewhat long, it could be range-compressed like this: https://stackoverflow.com/questions/2270910/how-to-reduce-consecutive-integers-in-an-array-to-hyphenated-range-expressions

erictune (commented):
Looks good, Aldo.  Added some optional suggestions.

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
ahg-g: but in that other sync the parallelism level could be lower, and so you can't recreate the failed pod (which might be executing an index in the middle of the range)

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+Indexed Jobs are Jobs [with fixed completion count], that is, Jobs that have a
+non-nil `.spec.completions` field.
+
+### Pod detail
+
+The Pod and PodSpec APIs don't get any new fields. However, Pods created for
+Indexed Jobs get the annotation `alpha.batch.k8s.io/job-completion-index` with a
+value equal to its completion index. The annotation is immutable.
+
+The annotation can be accessed through the downward API as a file or environment
```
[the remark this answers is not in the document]
ahg-g: would that require a kubelet change? if so, perhaps it is not worth the dependency?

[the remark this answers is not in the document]
alculquicondor: I also prefer we use the existing downward API, rather than silently adding the variable. Note that the same can be achieved with a mutating webhook too.

Note that we still require having the index somewhere else (like annotations) so that the controller can use it for tracking.

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: @erictune 

1) - I think that with the field added to job explicitly you can also separate code path cleanly. We have the necessary information in the job, and for pods we can check by OwnerRef.

2) Eric - can you clarify the part of "On downgrade, all the ones with ".spec.indexed" stop working."?
  I guess I'm missing the point what exactly this boolean will be responsible for controlling?
  Setting the env var of the index? Something more?

On a related note - if modifying PodSpec, when not add "index" int field, but just bool indexed one?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,487 @@
...
+variable.
+
+### Job completion and restart policy
+
+When dealing with Indexed Jobs, the Job controller keeps track of Pod
+completions for each index from 0 to `.spec.completions - 1`. When a Pod
+finishes successfully, the controller doesn't create another Pod with the same
```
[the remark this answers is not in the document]
alculquicondor: Good point, changed.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Other than the output format, we still need to add the full list of completed indexes to the Status. Is that acceptable?

wojtek-t: Maybe I'm too paranoid, but I can easily imagine case of someone who wants to run 1M indexes.
And in some pathological scenarios where e.g. the app itself is crashing for even numbers, we can end up with
Completed Indexes: [1, 3, 5, 7, ...., 999999]
Which we can't really compress.

I'm fine with hearing "you're stupid - this will not happen in practise", but it's something lights the yellow lamp for me.

That said - I fully agree that it would be useful and doable in majority of cases.
So maybe we can do something in between, which is:
- try to compress it
- if it's too long, just cut it to e.g. 1k bytes (or whatever other number), so something like (for above pathological case):
- 
`Completed Indexes: [1, 3, 5, <the next e.g. 400 odd numbers or so>, 799, <truncated> ]`

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Answering in upgrade/downgrade section.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: From @erictune:

> Did you consider the alternative where you do change the API such as by adding .spec.indexed boolean?
>
> Some pros of that approach are:
>
> 1. Could use a different path in the control code for indexed jobs. Reduces risk that the new path in the controller hurts existing multi-completition jobs that don't need indexes.
> 1. Upgrade downgrade is simple: On downgrade, all the ones with ".spec.indexed" stop working. On upgrade, those without ".spec.indexed" don't do anything different.
> 1. Can inject index without modifying existing pods see subsequent comments).

My response:

Yes, I have considered the idea, and it's open for discussion.

why is (3) important? The index would always be added, but Pods might not use it. So, in general, Pods don't need changes.
I don't love the idea of having a field just because of upgrades (2); WDYT of my alternatives to this? (1) is a more compelling reason, but risk can be minimized with proper coverage.

Perhaps @liggitt can chime in as well

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: Hmm - that actually made me think if we don't really want to store it, because on controller restart and many of finished pod already GC-ed we would actually lose the knowledge about them...

On a second thought to the above - the number of segments can probably be bounded by `spec.parallelism`, so how large in practice that can be...

I guess I've just lost myself in all those thoughts - I think I need to think more about this.

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: Sorry - I think I didn't understand the problem yet. Why those pods may not have those annotations with index?

keps/sig-apps/2214-indexed-job/README.md
ahg-g: +1, it can also be used to create a metric to track usage.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: In the topic of GC kubernetes/kubernetes#28486

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Because they were created by an older version of the controller that didn't add them.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: My argument against having a `.spec.indexed` boolean field is because of the possible extension to KEP: "stable pod names" or "Completion Index in the Pod Name" (listed in Alternatives section).

In that case, API validation becomes more tricky, which in turns makes the API more unusable:
- `.spec.indexed=true` requires `.spec.completions!=nil` (with current KEP)
- `.spec.stableNames=true` requires `.spec.indexed=true` (extension to the KEP)

But, maybe we can do the following (kind of back to the first version of this KEP):
Add the field `.spec.completionMode` which can have values: NonIndexed (default), Indexed, IndexedAndUnique.

alculquicondor: Added back `spec.completionMode` to the proposal. PTAL

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
erictune: Not a kubelet change.  I am suggesting the following logic in Job controller (pseudocode):
```
// Make new Pod
pod = new Pod
// Start with user-provided values
pod.mergeValues(job.spec.template.spec)
// set a default value for the env var, unless there is a collision.
for container in pod.spec.containers:
    if pod.spec.containers.env[] does not contain item with name = "INDEX":
        ev =  new ContainerEnvVar
        ev.name = "INDEX"
        ev.valueFrom.fieldRef.fieldPath =  "metadata.annotations['alpha.batch.k8s.io/job-completion-index']"
        pod.spec.containers.env.append(ev)
create(pod)
```

keps/sig-apps/2214-indexed-job/README.md
erictune: That could work.  Another option is as follows:

1. `.spec.indexed` and  `.spec.stableNames` both unset or false :arrow_right: Non-indexed behavior.
2. `.spec.indexed` true and  `.spec.stableNames` set unset   :arrow_right: Indexed behavior
3. `.spec.indexed`  and  `.spec.stableNames` both set true :arrow_right:  IndexedAndUnique behavior.
4. `.spec.indexed` unset and  `.spec.stableNames` set true :arrow_right: admission defaults `.spec.indexed` to true for you and you get  IndexedAndUnique behavior.
5. Other combinations are an error.

Whether the user wants Indexed, or IndexedAndUnique, they only have to type one line of yaml.

keps/sig-apps/2214-indexed-job/README.md
erictune: I might do something like this:  

- If there are fewer than 1000 completions, then set the status field.  
    - Worst case is 500 (only odd ones) indexes complete  times 4 bytes per (e.g. `997,`) = 2kB which is no big deal, considering 1.5MB object size limit
- If completions > 1000, never set this particular status field.
  - Status is always initially empty, so clients should expect to sometimes see fields missing.
  - Clients should wait on the condition, not on random status fields.
- Document that setting this status field is not supported for large completion counts.
-  Limit frequency of status updates to be nice to api server.  = 500, 3 dig object size limit is 1.5MB. 
- Consider raising in a few releases if there is demand.

keps/sig-apps/2214-indexed-job/README.md
erictune: If it were me, I would not attempt to validate or default these fields in OpenAPI.  I would do this defaulting and validation it in admission or the controller.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Added as unresolved, also targetting Beta.

erictune: I like the idea of using `status` as storage to solve the "forgot which pods we already ran" problem.   I think @wojtek-t  is alluding to that.

However, I think you can solve that in a separate subsequent KEP.

wojtek-t: > I like the idea of using status as storage to solve the "forgot which pods we already ran" problem. I think @wojtek-t is alluding to that.

Yes - that was something i wanted us to think about that.
But I agree with you it probably shouldn't be this KEP - it should probably be a separate feature.

alculquicondor: #28486 your input would be appreciated there. I'm about to initiate a KEP on it.

keps/sig-apps/2214-indexed-job/README.md
erictune: Once you add an enum to an API, you cannot add new values to it.
See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas
But you can add new fields.  Which is why I suggest the new-field-per-feature field.

alculquicondor: Oh API nuances :(

But if I protect the new value with a feature gate, I don't see it much different from adding a new field. @liggitt ?

keps/sig-apps/2214-indexed-job/README.md
erictune: Okay, I looked at that issue.

Seems like you can use the same method for indexed Job, except that you need to update the `status.completedIndexes` atomically in step 5 too.  But that means `status.completedIndexes` always needs to get set.

We can put a tighter bound on its size if we limit parallelism.  We don't need to limit completions.
`status.completedIndexes` has format `a-b,c-d,e-f,...`.  The number of commas cannot be more than the max parallelism allowed by the system for a job. (say 1000).
The length of one list item with comma e.g. `a-b,` is `2 * log10(max_completions) + 2`.
So the max string length of `status.completedIndexes` is `max_parallelism *  (2 * log10(max_completions) + 2)`

If you set `max_parallelism = 10^3`
and you set `max_completions = 10^9`
then max length of `status.completedIndexes` is  20kB.

alculquicondor: "640K is more memory than anyone will ever need on a computer" -- (probably not) Bill Gates

Should we really limit parallelism way bellow what a k8s cluster can support?

erictune: What are you thinking as an alternative?

alculquicondor: I can think of 2:
- To not keep track of indexes in status (and keep around the successful Pods with a finalizer until the entire Job succeeds)
- Have a limit that is more future proof. We know k8s can run up to 15k nodes. Maybe 10^5 or 10^6 would be reasonable.

erictune: Blocking deletion of lots of pods, when there are lots of users running concurrent jobs, might overload the apiserver? 
Maybe @wojtek-t can be more specific on what limits should be here.

A limit of 15k pods parallelism and 100k indices would take up would take 15k * (2*5 + 2) = 180k bytes.
With 80 char lines, getting the pod status will take over 2000 lines, which is not a great experience.

alculquicondor: If using `kubectl describe` we can probably stop showing the indexes at a lower threshold.

erictune: One option could be to separate the presentation and the state storage into separate fields..
* For presentation purposes, show `status.completedIndexes` when it is less than than `x` characters long, else do not set. Do this in this KEP
* For state storage,  to address kubernetes/kubernetes#28486 without delaying pod deletion, one can store the state of completed pods in a base64 encoded bitmask, e.g. `status.completionBitmask`.  10^6 bits fits into 16k chars, which is 200 lines at 80 chars per line.  Annoying but better. 
Once this KEP and kubernetes/kubernetes#28486 are implemented, both fields would be present in indexedjobs.

alculquicondor: > For presentation purposes, show status.completedIndexes when it is less than than x characters long, else do not set. Do this in this KEP

What about calculating the string to display from the bitmask on kubectl?

keps/sig-apps/2214-indexed-job/README.md
liggitt: > @erictune:
> Once you add an enum to an API, you cannot add new values to it. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#backward-compatibility-gotchas

That also gives a path for enum fields we expect to expand in the future:

> For enumeration-like fields that expect to add new values in the future, such as reason fields, please document that expectation clearly in the API field descriptions. Clients should treat such sets of values as potentially open-ended.

We encourage people to use enums over booleans for future expansion purposes in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types:

> Think twice about bool fields. Many ideas start as boolean but eventually trend towards a small set of mutually exclusive options. Plan for future expansions by describing the policy options explicitly as a string type alias (e.g. TerminationMessagePolicy).

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: See latest commit

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
alculquicondor: I put this for alpha

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Going with enum. See updated API documentation.

erictune (approved):
/approve
/lgtm

alculquicondor:
@wojtek-t friendly reminder for PRR

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,565 @@
...
+  Jobmanagement strategy for *fixed completion count*. That is, `Indexed` mode
+  cannot	be used for work queue patterns.	
+- The `.spec.parallelism` must be up to `10^6`. This is to guarantee that we
+  can keep track of completions per-index in the Job status in the future.
+
+### Pod detail
+
+The Pod and PodSpec APIs don't get any new fields. However, Pods created for
+Indexed Jobs get the annotation `batch.alpha.kubernetes.io/job-completion-index`
+with a value equal to its completion index. The annotation is immutable.
```
janetkuo, line 206: How do you make the annotation immutable? 

I think what you meant is that the Job controller adds this annotation on Pod creation and won't try to modify it.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,565 @@
...
+apiVersion: v1
+kind: Pod
+spec:
+  containers:
+    - name: test-container
+      env:
+        - name: JOB_COMPLETION_INDEX
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['batch.alpha.kubernetes.io/job-completion-index']
```
janetkuo, line 225: Isn't this part of the Job's pod template, as User Story 1 shows?

Or it's something that automatically created for the users, without them specifying it in the pod template?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,565 @@
...
+
+```
+Completed Indexes: [1-25,28,30-32,...]
+``` 
+<<[/UNRESOLVED]>>
+
+### Job scaling
+
+Jobs can be scaled up and down by varying `.spec.parallelism` (note that
+`.spec.completions` is an immutable field).
```
janetkuo, line 278: Please avoid using "scale" here, as Jobs don't follow the normal semantics for scaling. Ref https://github.com/kubernetes/kubernetes/pull/60139

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I actually meant immutable. This is crucial to avoid user errors.

This is possible in the API strategy for Job. There is precedence for other Pod annotations.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: It's automatically added at Pod creation, but user can also specify their own. See updated note.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Fixed

janetkuo (approved):

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+      env:
+        - name: JOB_COMPLETION_INDEX
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.annotations['batch.alpha.kubernetes.io/job-completion-index'] 
+```
+
+The Job controller doesn't add the environment variable if there is a name
+conflict with an existing environment variables in the Job's Pod template or if
+the user already specified another environment variable for the same annotation.
```
wojtek-t, line 230: Why? What's the problem with having the value referenced by two separate env vars?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+
+The kubelet handles container restarts as usual, according to the
+`spec.template.spec.restartPolicy`.
+
+<<[UNRESOLVED TBD Beta: Track completed indexes in Job status]>>
+Once [kubernetes/kubernetes#28486](https://github.com/kubernetes/kubernetes/issues/28486)
+is resolved:
+
+The Job controller keeps track of completed indexes in the Job status in a
+base64 encoded bitmask.
```
wojtek-t, line 261: Can you explicitly describe how JobStatus is updated then (for completeness)?

Another question is: with the bitmask field and the proposal above, will we end-up with two separate fields for that?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+  // `NonIndexed` means that each Pod completion is homologous to each other.	
+  // The Job is considered complete when there have been .spec.completions	
+  // successful completions.	
+  // `Indexed` means that each Pod completion needs to be tracked individually;	
+  // each Pods gets an associated completion index, which is available in the
+  // annotation	`batch.alpha.kubernetes.io/job-completion-index`.	
+  // The Job is considered complete when there is one successful Pod for each	
+  // index in the range 0 to (.spec.completions - 1).	
+  // When value is `Indexed`, .spec.completions must have a non-zero positive	
+  // value and `.spec.parallelism` must be less than or equal to 10^6.
```
wojtek-t, line 185: I'm assuming that the reasoning behind this 10^6 is the underlying bitmask, right?
i.e. we're saying that tracking 10^6 items in bitmask takes O(125kB) and this is fine, right?

That sounds fine to me, but I would like to ensure that somewhere in the design section we will add a requirement that such big objects won't be updated frequently (e.g. not more often that once per couple seconds or something like that).

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: Ohh - or you're saying that "we don't solve it for Alpha at all" - just rely on existence of all pods for now?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+Additionally, we add unit tests for API defaulting and validation, with feature
+gate enabled and disabled.
+  
+### Graduation Criteria
+
+#### Alpha -> Beta Graduation
+
+- Complete features:
+  - Completion index in annotation
+  - Restart policy
```
wojtek-t, line 306: Shouldn't those features be part of Alpha?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+Pods will be created without a completion index. Existing workloads that
+expected the completion index will fail. But this is expected in a downgrade.
+
+In the event of an upgrade after a downgrade, the controller will ignore
+existing Pods without completion index for existing Indexed Jobs. This might
+cause a load spike in the cluster.
+
+If, instead of a downgrade, the feature is disabled:
+
+  - kube-apiserver sets `.spec.completionMode=NonIndexed` for new Jobs.
```
wojtek-t, line 332: I don't understand how this is done (and how you distinguish downgrade from disabling exactly). Can you be more specific?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,566 @@
...
+  Yes. Using the feature gate is the recommended way.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+
+  The Job controller starts managing Indexed Jobs again.
+  More details covered in [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy).
+
+* **Are there any tests for feature enablement/disablement?**
+
+  Yes, unit and integration test for the feature enabled and disabled.
```
wojtek-t, line 372: If we're going to have something fancy for feature disablement, than we should have (at least) unit test for disablement.

You can see an example of doing that here (under review, but the idea is there):
https://github.com/kubernetes/kubernetes/pull/97058/files#diff-7826f7adbc1996a05ab52e3f5f02429e94b68ce6bce0dc534d1be636154fded3R246

wojtek-t (commented):
I added some comments, but mostly fairly minor one.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I think it makes sense to give the user control over the names of the variables. A binary could always have some behavior defined for JOB_COMPLETION_INDEX which is different from the one provided by k8s.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Yes, this is about being able to maintain the bitmask. Although, if we are going to use it to track completion, we actually need to cap .spec.completions too. The idea is that in the future we don't depend on lingering pods. @erictune thoughts?

> e.g. not more often that once per couple seconds or something like that

Should we be doing some kind of client side throttling? In the current design, Job status updates would come after removing finalizers from finishing Pods, which already throttles Job updates in a way. But if one Pod finishes every 100ms (for example), we might still do more Job updates than what you are suggesting.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Correct. This would be a beta graduation criteria, pending on #2307

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: This is standard logic in API strategies for disabled feature gates. Reworded a bit.

I don't exactly distinguish a downgrade, as I cannot control the code of version N-1. So the paragraphs above describe how the code behaves in that scenario.

But what I can control is a feature being disabled, in which case I prefer to ignore Indexed Jobs that were already created.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: That was the intention of my text here. Reworded.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I changed the limit to be on completions, instead of parallelism

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: I guess I wasn't clear. I wasn't asking to touch customer-defined variable.

I was just asking - why we don't want to provide JOB_COMPLETION_INDEX unconditionally (either with user-propagated-value if set by user or as above if not).

I don't see a problem if you have two env vars, say FOO and JOB_COMPLETION_INDEX with the same value set to the val of that annotation.

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: OK - if you're not touching JobStatus for now at all, that's fine.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+  // `NonIndexed` means that each Pod completion is homologous to each other.	
+  // The Job is considered complete when there have been .spec.completions	
+  // successful completions.	
+  // `Indexed` means that each Pod completion needs to be tracked individually;	
+  // each Pods gets an associated completion index, which is available in the
+  // annotation	`batch.alpha.kubernetes.io/job-completion-index`.	
+  // The Job is considered complete when there is one successful Pod for each	
+  // index in the range 0 to (.spec.completions - 1).	
+  // When value is `Indexed`, .spec.completions must have a value between 1
+  // and 10^6.
```
wojtek-t, line 186: >  Yes, this is about being able to maintain the bitmask. Although, if we are going to use it to track completion, we actually need to cap .spec.completions too. The idea is that in the future we don't depend on lingering pods. @erictune thoughts?

I don't think you need to limit completions. The whole points that Eric made in one of other comments proved that it's enough to limit `parallelism` [because if you won't be storing bitmask, but rather [a-b,c-d,e-f,...] format, there will be at most "parallelism` of such pairs, so the length of that is actually bounded by parallelism.
Bounding completions is much more limiting that limiting parallelism to something as high as 10^6.

> Should we be doing some kind of client side throttling? In the current design, Job status updates would come after removing finalizers from finishing Pods, which already throttles Job updates in a way. But if one Pod finishes every 100ms (for example), we might still do more Job updates than what you are suggesting.

Yes - we should batch those, and send 1 update per X second. We can tune X later.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+existing Pods without completion index for existing Indexed Jobs, without they
+counting towards failures or completions. The controller will create new Pods
+with indexes. This might cause a load spike in the cluster.
+
+If, instead of a downgrade, the cluster administrator disables the feature gate:
+
+  - kube-apiserver sets `.spec.completionMode=NonIndexed` for new Jobs at
+    creation time.
+  - kube-controller-manager ignores existing Indexed Jobs and emits a warning
+    event.
```
wojtek-t, line 343: How exactly those are ignored? Are we killing the existing pods? Are we just not starting new ones? Are we stacking pods being finished?

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: > I don't think you need to limit completions. The whole points that Eric made in one of other comments proved that it's enough to limit parallelism [because if you won't be storing bitmask, but rather [a-b,c-d,e-f,...] format, there will be at most "parallelism` of such pairs, so the length of that is actually bounded by parallelism.

So it depends on the implementation. If we use a bitmask, then we have to limit completions. If we limit parallelism, we would have to use the compressed list format, but then the limit might have to be lower than 10^6, as we require more characters. We don't have to be concerned about display space, as we can just limit `kubectl describe` to print less characters.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I understood you correctly :)

What I'm saying is: what if a JOB_COMPLETION_INDEX variable causes an undesirable effect on the user binary?

I guess in this case they could override that environment variable with another value.

keps/sig-apps/2214-indexed-job/README.md
soltysh: I agree that tracking completed pods is orthogonal to this proposal.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+The Job controller keeps track of completed indexes in the Job status in a
+base64 encoded bitmask.
+The command `kubectl describe` displays the list of completed indexes in a
+human-readable compressed format. For example:
+
+```
+Completed Indexes: [1-25,28,30-32]
+``` 
+
+The command crops the list of indexes once it reaches 200 characters.
```
soltysh, line 271: 200 seems a lot, I don't think we have a hard limit in kubectl, but I'm positive this is too much. Also it's not relevant to this proposal, I'd drop it from here. It's implementation detail.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+In the event of an upgrade after a downgrade, the controller will remove
+existing Pods without completion index for existing Indexed Jobs, without they
+counting towards failures or completions. The controller will create new Pods
+with indexes. This might cause a load spike in the cluster.
+
+If, instead of a downgrade, the cluster administrator disables the feature gate:
+
+  - kube-apiserver sets `.spec.completionMode=NonIndexed` for new Jobs at
+    creation time.
+  - kube-controller-manager ignores existing Indexed Jobs and emits a warning
```
soltysh, line 342: Warn and run as non-indexed? Warn and ignore? Can you be more specific?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+expected the completion index will fail. But this is expected in a downgrade.
+
+In the event of an upgrade after a downgrade, the controller will remove
+existing Pods without completion index for existing Indexed Jobs, without they
+counting towards failures or completions. The controller will create new Pods
+with indexes. This might cause a load spike in the cluster.
+
+If, instead of a downgrade, the cluster administrator disables the feature gate:
+
+  - kube-apiserver sets `.spec.completionMode=NonIndexed` for new Jobs at
```
soltysh, line 340: Will there be validation in kube-apiserver not to allow setting indexed jobs when this feature is disabled?

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I don't have a strong opinion about the name. But I'm still not convinced we should always inject the value if the user is already adding it.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Removed specificity.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: the value is overriden. Reworded for clarity.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Added specificity.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Added specificity.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Correct, created a section for alpha.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: @soltysh any thoughts?

wojtek-t: Running 10^6 of pods at the same time is far from supported at this point.
Having more than 10^6 of pods within a single job sounds like a valid usecase to me.

So I would be much more comfortable saying that  "parallelism <= 10^5" (so decreasing that by order of magnitude even), than by setting a limit on the `completions`

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: We should always inject, unless user explicitly overwrites that name.

The behavior should be consistent - if you job is indexed, var X (I don't have strong opinion about name) will always be set (in your behavior you have "the var may or may not be set" - that's super confusing).

alculquicondor: Sounds good. Removed second condition.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Moved back to limit parallelism and use a compressed list format instead of bitmask.

wojtek-t:
@alculquicondor - this LGTM, please squash the commits

alculquicondor:
squashed

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -0,0 +1,577 @@
...
+  // annotation	`batch.alpha.kubernetes.io/job-completion-index`.	
+  // The Job is considered complete when there is one successful Pod for each	
+  // index in the range 0 to (.spec.completions - 1).	
+  // When value is `Indexed`, .spec.completions must have a value between 1
+  // and 10^6.
+  //
+  // More completion modes can be added in the future. If a Job controller
+  // observes a mode that it doesn't recognize, it manages the Job as in
+  // `NonIndexed`.
+  CompletionMode string
```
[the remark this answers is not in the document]
wojtek-t: I think I agree with @soltysh on making a dedicated type, but I think this will get figured out during API review.

wojtek-t:
/approve

I will give a change to @soltysh - to take another look.

keps/sig-apps/2214-indexed-job/README.md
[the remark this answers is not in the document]
alculquicondor: I already gave it a type :)

alculquicondor:
/hold
for LGTM from @soltysh 

@janetkuo could you approve?

alculquicondor:
sorry, I missed that it was already approved :)

ahg-g:
/lgtm

since I reviewed it earlier.

wojtek-t:
> I already gave it a type :)

Yeah - I realized after writing this comment :) I think that was the last non-addressed one. So let me cancel the hold - it's in very good shape now.

/hold cancel

soltysh:
Was one day late, but :+1: thanks for that :tada:

PR #2246, opened 2021-01-08
Updates the ttl-to-finished KEP to graduate the feature to Beta.

ahg-g:
Enhancement issue: #592

### DIFF
--- a/keps/sig-apps/592-ttl-after-finish/README.md
+++ b/keps/sig-apps/592-ttl-after-finish/README.md
@@ -16,10 +16,20 @@
   - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
     - [TTL Controller](#ttl-controller)
     - [Finished Jobs](#finished-jobs)
-    - [Finished Pods](#finished-pods)
     - [Owner References](#owner-references)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Graduation Criteria](#graduation-criteria)
+  - [Alpha](#alpha)
+  - [Alpha -&gt; Beta](#alpha---beta)
+  - [Beta -&gt; GA](#beta---ga)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Future Work](#future-work)
 - [Implementation History](#implementation-history)
 <!-- /toc -->
 
@@ -106,24 +116,6 @@ This allows Jobs to be cleaned up after they finish and provides time for
 asynchronous clients to observe Jobs' final states before they are deleted.
 
 
-Similarly, we will add the following API fields to `PodSpec` (`Pod`'s `.spec`).
-
-```go
-type PodSpec struct {
- 	// ttlSecondsAfterFinished limits the lifetime of a Pod that has finished
-	// execution (either Succeeded or Failed). If this field is set, once the Pod
-	// finishes, it will be deleted after ttlSecondsAfterFinished expires. When
-	// the Pod is being deleted, its lifecycle guarantees (e.g. finalizers) will
-	// be honored. If this field is unset, ttlSecondsAfterFinished will not
-	// expire. If this field is set to zero, ttlSecondsAfterFinished expires
-	// immediately after the Pod finishes.
-	// This field is alpha-level and is only honored by servers that enable the
-	// TTLAfterFinished feature.
-	// +optional
-	TTLSecondsAfterFinished *int32
-}
-```
-
 ##### Validation
 
 Because Job controller depends on Pods to exist to work correctly. In Job
@@ -157,16 +149,16 @@ The steps are as easy as:
 ### Implementation Details/Notes/Constraints
 
 #### TTL Controller
-We will add a TTL controller for finished Jobs and finished Pods. We considered
+We will add a TTL controller for finished Jobs. We considered
 adding it in Job controller, but decided not to, for the following reasons:
 
 1. Job controller should focus on managing Pods based on the Job's spec and pod
    template, but not cleaning up Jobs.
-1. We also need the TTL controller to clean up finished Pods, and we consider
+1. We also need the TTL controller to clean up finished Pods in the future, and we consider
    generalizing TTL controller later for custom resources. 
 
-The TTL controller utilizes informer framework, watches all Jobs and Pods, and
-read Jobs and Pods from a local cache.
+The TTL controller utilizes informer framework, watches all Jobs, and
+read Jobs from a local cache.
 
 #### Finished Jobs
 
@@ -191,29 +183,6 @@ When a Job is created or updated:
    * If it hasn't expired, it is not safe to delete the Job. Delay re-enqueue
      the Job after a computed amount of time when it will expire.
 1. Delete the Job if passing the sanity checks. 
-
-#### Finished Pods
-
-When a Pod is created or updated:
-1. Check its `.status.phase` to see if it has finished (`Succeeded` or `Failed`).
-   If it hasn't finished, do nothing. 
-1. Otherwise, if the Pod has finished, check if Pod's
-   `.spec.ttlSecondsAfterFinished` field is set. Do nothing if the TTL field is
-   not set. 
-1. Otherwise, if the TTL field is set, check if the TTL has expired, i.e.
-   `.spec.ttlSecondsAfterFinished` + the time when the Pod finishes (max of all
-   of its containers termination time
-   `.containerStatuses.state.terminated.finishedAt`) > now. 
-1. If the TTL hasn't expired, delay re-enqueuing the Pod after a computed amount
-   of time when it will expire. The computed time period is:
-   (`.spec.ttlSecondsAfterFinished` + the time when the Pod finishes - now).
-1. If the TTL has expired, `GET` the Pod from API server to do final sanity
-   checks before deleting it.
-1. Check if the freshly got Pod's TTL has expired. This field may be updated
-   before TTL controller observes the new value in its local cache.
-   * If it hasn't expired, it is not safe to delete the Pod. Delay re-enqueue
-     the Pod after a computed amount of time when it will expire.
-1. Delete the Pod if passing the sanity checks. 
 
 #### Owner References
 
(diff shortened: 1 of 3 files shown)


ahg-g:
@janetkuo should I move the "Finished Pods" section to a separate "Future Work" section?

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
 ## Graduation Criteria
 
-We want to implement this feature for Pods/Jobs first to gather feedback, and
-decide whether to generalize it to custom resources. This feature can be
-promoted to beta after we finalize the decision for whether to generalize it or
-not, and when it satisfies users' need for cleaning up finished resource
-objects, without regressions.
+- The feature implemented for Job, as future work, it can be extended to Pods and perhaps custom resources, but that should happen under separate feature flags.
+- Add necessary tests
+- Graduate to Beta in v1.21
```
janetkuo, line 262: Graduation criteria should be based on quality/maturity instead of releases. This can be captured in the `milestone` of kep.yaml.

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
 
 ## Graduation Criteria
 
-We want to implement this feature for Pods/Jobs first to gather feedback, and
-decide whether to generalize it to custom resources. This feature can be
-promoted to beta after we finalize the decision for whether to generalize it or
-not, and when it satisfies users' need for cleaning up finished resource
-objects, without regressions.
+- The feature implemented for Job, as future work, it can be extended to Pods and perhaps custom resources, but that should happen under separate feature flags.
+- Add necessary tests
```
wojtek-t, line 261: Can you describe the tests that should be added?

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: Agree with Janet. Also please split by alpha-beta and beta->ga graduation, see e.g. here:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1904-efficient-watch-resumption#graduation-criteria

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+  No.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  Yes.
```
wojtek-t, line 288: With caveats (as this is API field).

Please take a look at https://github.com/kubernetes/enhancements/pull/2199/files#diff-b18c53698772d9f0dcc15e9f35d90ecd0f181923f33e40115036991aa0b5f1c5R234 - it's a pretty similar case wrt to this question.

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+  It should work as expected.
+
+* **Are there any tests for feature enablement/disablement?**
+  No.
+
+### Rollout, Upgrade and Rollback Planning
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  It shouldn't impact already running workloads. This is an opt-in feature since users need to 
+  explicitly set the TTLSecondsAfterFinished parameter in the job spec, which is accepted by the
```
wojtek-t, line 300: nit: I think it's silently dropped if the feature is disabled (which is slightly different than accepted if enabled)

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+
+### Rollout, Upgrade and Rollback Planning
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  It shouldn't impact already running workloads. This is an opt-in feature since users need to 
+  explicitly set the TTLSecondsAfterFinished parameter in the job spec, which is accepted by the 
+  api-server only if the feature is enabled.
+
+* **What specific metrics should inform a rollback?**
+  Unexpected restarts of kube-controller-manager
```
wojtek-t, line 304: How about kube-apiserver?
I suggest also extended 4xx/5xx on Jobs from kube-apiserver (similarly to what I've been pointing out in #2199 that I mentioned above).

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  N/A
```
wojtek-t, line 307: Definitely not N/A - there are ways that it can potentially break.

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+fields of API types, flags, etc.?**
+  No
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  - Check the workqueue metrics
```
wojtek-t, line 318: How you can get that from workqueue metrics?

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+  - When checking the job object in the api server, the TTLSecondsAfterFinished parameter should
```
wojtek-t, line 319: nit: it's just a wording, but I'm trying to introduce a bit of consistency between answers (we will hopefully get to to more "choose from the below options" in the future here):

`Listing jobs in the cluster and checking if any has ttlSecondsAfterFinished field set.`

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+    - Metric name: `workqueue_depth`
+    - Metric name: `workqueue_queue_duration_seconds`
+    - Metric name: `workqueue_retries_total`
+    - Components exposing the metric: `kube-controller-manager`
+    - Metric name: `etcd_object_counts{resource="jobs.batch"}`
+    - Components exposing the metric: `kube-apiserver`.
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+
+99% of the jobs that needs cleanup are deleted within X minutes.
```
wojtek-t, line 336: I don't see an easy way to compute such SLO out of the existing metrics.

Can we create a metric that will basically expose this information?
Something like:
"histogram of when the job were supposed to be deleted minus when it was deleted"
[somewhat similar to what we did (or supposed to do) for cronjob: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/19-Graduate-CronJob-to-Stable#metrics ]

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+_This section must be completed when targeting beta graduation to a release._
+
+* **Does this feature depend on any specific services running in the cluster?**
+  No.
+
+### Scalability
+
+* **Will enabling / using this feature result in any new API calls?**
+  - API call type: DELETE jobs
+  - Estimated throughput: depends on job creation and completion rate.
```
wojtek-t, line 354: maybe: "at most one for each job creation? (depending on feature use)"

[that basically puts the upper bound and shows that while we don't control the throughput per-se, the total amount of requests is bounded by requests that already exist]

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
...
+* **Will enabling / using this feature result in introducing new API types?**
+  No.
+  
+* **Will enabling / using this feature result in any new calls to the cloud 
+provider?**
+ No.
+
+* **Will enabling / using this feature result in increasing size or count of 
+the existing API objects?**
+ No.
```
wojtek-t, line 366: Technically yes - we're adding a field so the object will be slightly bigger :)

keps/sig-apps/592-ttl-after-finish/kep.yaml
```diff
@@ -5,18 +5,47 @@ authors:
 owning-sig: sig-apps
 participating-sigs:
   - sig-api-machinery
+status: implemented
```
wojtek-t, line 8: implementable

[status should be changed to implemented after GA graduation]

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: updated to indicate that the field will continue to exist but should have no effect.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: Updated.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: done.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: updated the sentence

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: done.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: updated

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: updated

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: is that a metric that already exists? if so, what is the name?

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: updated to manually tested as part of alpha to beta graduation criteria

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: I was thinking we could use `workqueue_queue_duration_seconds`: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/prometheus/workqueue/metrics.go#L60

keps/sig-apps/592-ttl-after-finish/kep.yaml
ahg-g: done

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: good point.

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: apiserver_request_total has a "code" (being http code returned to the user):
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/endpoints/metrics/metrics.go#L76

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +219,194 @@ Mitigations:
...
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+fields of API types, flags, etc.?**
+  No
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  - The `workqueue_adds_total` tracks the number of finished Jobs with ttlSecondsAfterFinished set.
```
wojtek-t, line 291: I think you need to provide some label to that metric (controller name or sth like that), right?
Can you please make that explicit?

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +219,194 @@ Mitigations:
...
+  - Listing jobs in the cluster and checking if any has ttlSecondsAfterFinished field set.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine 
+the health of the service?**
+ - [x] Metrics
+    - Metric name: `ttl_after_finished_controller_rate_limiter_use`
+    - Metric name: `workqueue_adds_total`
```
wojtek-t, line 298: Same here (request for the label name)

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: Can you describe how? [I think you will be putting all updates to Job objects in the queue, so distinguishing the ones that are relevant might be hard in that metric.]

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: you mean 4xx/5xx replies when reaching to the jobs endpoint on the api-server? why would that be relevant to this feature?

wojtek-t: They don't have to be relevant. But they may potentially be, e.g. imagine tha tthe conversion function is returning an error if that field is set - that would exactly result in returning 5xx from it.

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +219,199 @@ Mitigations:
...
+* **How can a rollout fail? Can it impact already running workloads?**
+  It shouldn't impact already running workloads. This is an opt-in feature since users need to 
+  explicitly set the TTLSecondsAfterFinished parameter in the job spec, which is silently dropped
+  by the api-server if the feature is disabled.
+
+* **What specific metrics should inform a rollback?**
+  Unexpected restarts of kube-controller-manager
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  Manually tested.
```
wojtek-t, line 280: Any findings?

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: Thanks for explanation.

That said - I don't think you can use workqueue metrics to report what we want exactly:
1) We really want to wait for TTL - we don't want to delete jobs too early
2) The fact that the queue is keeping up with load doesn't mean there isn't a bug in the code and we aren't really deleting as fast as we should.

So I really think we should add a metric similar to what was done for CronJob...

ahg-g: > We really want to wait for TTL - we don't want to delete jobs too early

We are not deleting jobs too early. Not sure how you concluded that from my comment.

> The fact that the queue is keeping up with load doesn't mean there isn't a bug in the code and we aren't really deleting as fast as we should.

Where would the metric be updated? To achieve the semantics you are looking for I think we need a dedicated metric endpoint such that when queried, all jobs are listed and a histogram is computed (something like https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics).

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: oh, sorry, I thought I would be doing that before submitting the flag flip PR, I will do the test today.

janetkuo (approved):
LGTM. I'll let @wojtek-t do the PRR review.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: done.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: I had some issues with gce-enforcer (deleting firewall rules from my k8s on gce cluster), will try again tomorrow.

ahg-g: done, works as expected.

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +219,200 @@ Mitigations:
...
+
+* **Are there any tests for feature enablement/disablement?**
+  No.
+
+### Rollout, Upgrade and Rollback Planning
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  It shouldn't impact already running workloads. This is an opt-in feature since users need to 
+  explicitly set the TTLSecondsAfterFinished parameter in the job spec, which is silently dropped
+  by the api-server if the feature is disabled.
```
wojtek-t, line 274: The field isn't dropped, right?
We're preserving the fields if they were set.

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: > We are not deleting jobs too early. Not sure how you concluded that from my comment.

Sorry - I didn't mean we're deleting.
What I was saying: "what if there is a bug that leads to deleting them to early".

In general, the purpose of many of metrics isn't needed is everything works as designed. They are used if there are bugs, something doens't work as expected, etc....

> Where would the metric be updated? To achieve the semantics you are looking for I think we need a dedicated metric endpoint such that when queried, all jobs are listed and a histogram is computed (something like https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics).

What I had on mind is basically:
- when you send a "DELETE job" request, you basically export the metric  "currentTime - time when job succeeded - TTL value"

Sure - if the feature doesn't work at all we won't see anything in metrics. But assuming it eventually deletes stuff, we will see how accurately it does.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: It is dropped: https://github.com/kubernetes/kubernetes/blob/d72c056260e771e8cd01220203582de0c0015786/pkg/registry/batch/job/strategy.go#L79

wojtek-t: Hmm - which isn't really a pattern we want to have. We generally want to avoid clearing explicitly set fields, e.g. in:
https://github.com/kubernetes/kubernetes/blob/d72c056260e771e8cd01220203582de0c0015786/pkg/api/pod/util.go#L450

@liggitt - FYI

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: > In general, the purpose of many of metrics isn't needed is everything works as designed. They are used if there are bugs, something doens't work as expected, etc....

Right, but my general argument is that adding a new metric isn't going to make things a lot different compared to relying on the workqueue ones with respect to avoiding bugs.

 > when you send a "DELETE job" request, you basically export the metric "currentTime - time when job succeeded - TTL value"
> Sure - if the feature doesn't work at all we won't see anything in metrics. But assuming it eventually deletes stuff, we will see how accurately it does.

This doesn't protect us from bugs in the workqueue (your initial concern about the workqueue metrics), what if the workqueue has a bug were push sometimes work and sometimes silently doesn't (and so some jobs will get deleted and some will not or randomly delayed)? 

In any case, I updated the KEP to indicate that we will add this metric. But, our source of truth is etcd, and I think the only way to reliably achieve the semantics you want is by querying that directly.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: done.

keps/sig-apps/592-ttl-after-finish/README.md
ahg-g: This is an existing behavior, do we want to change it?

wojtek-t: I think we want - though please reach out to @liggitt (or other API approver) for final confirmation.

keps/sig-apps/592-ttl-after-finish/README.md
wojtek-t: I agree that doesn't protect us from all kind of bugs, but I think that metric itself is fairly useful on its own.

Anyway - thanks!

keps/sig-apps/592-ttl-after-finish/kep.yaml
```diff
@@ -2,21 +2,52 @@ title: TTL After Finished
...
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: TTLAfterFinished
+    components:
+      - kube-apiserver
+      - kube-controller-manager
+disable-supported: true
+
+metrics:
+  - ttl_after_finished_controller_rate_limiter_use
```
wojtek-t, line 49: Please also add ttl_after_finished_controller_time_to_deletion_seconds metric here

wojtek-t (commented):
I'm fine with it modulo one minor nit and the open comment about dropping field - please resolve this one with API approvers.

You also need sig-apps approval.

wojtek-t:
> You also need sig-apps approval.

Actually - I see Janet already approved. So I will just wait for the "drop fields" discussion resolution and will approve.

keps/sig-apps/592-ttl-after-finish/README.md
liggitt: The current behavior looks correct.. if the feature is disabled, we only preserve the field if it was already set in the **persisted** object, and clear it otherwise. That

ahg-g: done.

keps/sig-apps/592-ttl-after-finish/kep.yaml
ahg-g: done.

wojtek-t:
LGTM

@ahg-g - please squash commits and I will approve

ahg-g:
> LGTM
> 
> @ahg-g - please squash commits and I will approve

thanks for the review, done!

keps/sig-apps/592-ttl-after-finish/README.md
```diff
@@ -250,17 +257,141 @@ Mitigations:
 
 ## Graduation Criteria
 
-We want to implement this feature for Pods/Jobs first to gather feedback, and
-decide whether to generalize it to custom resources. This feature can be
-promoted to beta after we finalize the decision for whether to generalize it or
-not, and when it satisfies users' need for cleaning up finished resource
-objects, without regressions.
+- The feature implemented for Job, as future work, it can be extended to Pods and perhaps custom resources, but that should happen under separate feature flags.
```
wojtek-t, line 260: Your rebase didn't work - it seems you removed all the further commits....

wojtek-t: :(

ahg-g: ops, forgot to `git add .` :) should be fixed now

wojtek-t:
Thanks!

/lgtm
/approve

PR #2251, opened 2021-01-12
Add KEP for Pod Affinity NamespaceSelector

ahg-g:
/sig scheduling

#2249

### DIFF
--- a/keps/prod-readiness/sig-scheduling/2249.yaml
+++ b/keps/prod-readiness/sig-scheduling/2249.yaml
@@ -1,0 +1,3 @@
+kep-number: 2249
+alpha:
+  approver: "@wojtek-t"
--- a/keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
+++ b/keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
@@ -1,0 +1,732 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [x] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [x] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [x] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [x] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [x] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [x] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-2249: Namespace Selector For Pod Affinity
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [Performance](#performance)
+    - [Security/Abuse](#securityabuse)
+    - [External Dependencies](#external-dependencies)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Alternatives](#alternatives)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+By default, pod affinity/anti-affinity constraints are calculated against 
+the pods in the same namespace. The Spec allows users to expand that via
+the `PodAffinityTerm.Namespaces` list. 
+
+This KEP proposes two related features: 
+1. Adding NamespaceSelector to `PodAffinityTerm` so that users 
+   can specify the set of namespace using a label selector.
+1. Adding a new quota scope named `CrossNamespaceAffinity` that allows operators
+   to limit which namespaces are allowed to have pods that use affinity/anti-affinity
+   across namespaces. 
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+Pod affinity/anti-affinity API allows using this feature across namespaces using 
+a static list, which works if the user knowns the namespace names in advance.
+
+However, there are cases where the namespaces are not known beforehand, for which there
+is no way to use pod affinity/anti-affinity. Allowing users to specify the set of namespaces
+using a namespace selector addresses this problem.
+
+Since NamespaceSelector doubles down on allowing cross-namespace pod affinity, giving
+operators a knob to control that is important to limit the potential abuse of this feature
+as described in the risks section.
+
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+### Goals
+- Allow users to dynamically select the set of namespaces considered when using pod
+  affinity/anti-affinity
+- Allow limiting which namespaces can have pods with cross namespace pod affinity
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+<!--
+### Non-Goals
+
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+### User Stories (Optional)
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+I am running a SaaS service, the workload for each customer is placed in a separate
+namespace. The workloads requires 1:1 pod to node placement. I want to use pod 
+anti-affinity across all customers namespaces to achieve that.
+
+<!--
+
+### Notes/Constraints/Caveats (Optional)
+
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+#### Performance
+Using namespace selector will make it easier for users to specify affinity 
+constraints across a large number of namespaces. The initial implementation 
+of pod affinity/anti-affinity suffered from performance challenges, however 
+over the releases 1.17 - 1.20 we significantly improved that. 
+
+We currently have integration and clusterloader benchmarks that evaluate the extreme
+cases of all pods having affinity/anti-affinity constraints to/against each other.
+Those benchmarks show that the scheduler is able to achieve maximum throughput 
+(i.e., the api-server qps limit).
+
+#### Security/Abuse
+See previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
+
+We will address this concern by introducing a new quota scope named `CrossNamespaceAffinity`
+that operators can use to limit which namespaces are allowed to have pods with affinity terms
+that set the existing `namespaces` field or the proposed one `namespaceSelector`. 
+
+Using this new scope, operators can prevent certain namespaces (`foo-ns` in the example below) 
+from having pods that use cross-namespace pod affinity by creating a resource quota object in
+that namespace with `CrossNamespaceAffinity` scope and hard limit of 0:
+
+```yaml
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: disable-cross-namespace-affinity
+  namespace: foo-ns
+spec:
+  hard:
+    pods: "0"
+  scopeSelector:
+    matchExpressions:
+    - scopeName: CrossNamespaceAffinity
+```
+
+If operators want to disallow using `namespaces` and `namespaceSelector` by default, and 
+only allow it for specific namespaces,  they could configure `CrossNamespaceAffinity` 
+as a limited resource by setting the kube-apiserver flag --admission-control-config-file
+to the path of the following configuration file:
+
+```yaml
+apiVersion: apiserver.config.k8s.io/v1
+kind: AdmissionConfiguration
+plugins:
+- name: "ResourceQuota"
+  configuration:
+    apiVersion: apiserver.config.k8s.io/v1
+    kind: ResourceQuotaConfiguration
+    limitedResources:
+    - resource: pods
+      matchScopes:
+      - scopeName: CrossNamespaceAffinity
+```
+
+With the above configuration, pods can use `namespaces` and `namespaceSelector` only
+if the namespace where they are created have a resource quota object with 
+`CrossNamespaceAffinity` scope and a hard limit equal to the number of pods that are
+allowed to.
+
+Moreover, to prevent accidentally selecting a large number of namespaces, we will reject empty
+selectors. For example, users can do the following:
+
+```yaml
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector:
+      matchExpressions:
+        - key: workload
+          operator: In
+          values:
+          - HPC
+```
+
+but can't do the following:
+
+```yaml
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector: {}
+```
+
+For more protection, admission webhooks like gatekeeper can be used to further
+restrict the use of this field.
+
+#### External Dependencies
+We are aware of two k8s projects that could be impacted by this change, the descheduler 
+and cluster autoscaler. Cluster autoscaler should automatically consume the change
+since it imports the scheduler code, the descheduler however doesn't and needs to 
+be changed to take this feature into account. We will open an issue to inform the project
+about this update.
+
+
+<!--
+What are the risks of this proposal, and how do we mitigate? Think broadly.
+For example, consider both security and how this will impact the larger
+Kubernetes ecosystem.
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+## Design Details
+
+ Add `NamespaceSelector` field to `PodAffinityTerm`:
+
+```go
+type PodAffinityTerm struct {
+    // A label query over the set of namespaces that the term applies to.
+    // The term is applied to the union of the namespaces selected by this field
+    // and the ones listed in the namespaces field.
+    // nil selector and empty namespaces list means "this pod's namespace"
+    // An empty selector ({}) is not valid.
+    NamespaceSelector *metav1.LabelSelector
+}
+```
+
+As indicated in the comment, the scheduler will consider the union of the namespaces
+specified in the existing `Namespaces` field and the ones selected by the new 
+`NamespaceSelector` field. `NamespaceSelector`  is ignored when set to nil.
+
+We will do two precomputations at PreFilter/PreScore:
+
+- The names of the namespaces selected by the `NamespaceSelector` is computed. This set
+  will be used by Fliter/Score to match against existing pods namespaces.
+- A snapshot of the labels of the namespace of the incoming pod. This will be used when 
+  to match against the anti-affinity constraints of existing pods. 
+
+The precomputations are necessary for:
+
+- Performance.
+- Ensures a consistent behavior if namespace labels are added/removed during 
+  the scheduling cycle of a pod.
+
+Finally, the feature will be guarded by a new feature flag. If the feature is 
+disabled, the field `NamespaceSelector` is preserved if it was already set in
+the persisted Pod ojbect, otherwise it is silently dropped; moreover kube-scheduler 
+will ignore the field and continue to behave as before.
+
+
+With regards to adding the `CrossNamespaceAffinity` quota scope, the one design aspect
+worth noting is that it will be rolled out in multiple releases similar to the `NamespaceSelector`:
+when the feature is disabled, the new value will be tolerated in updates of objects
+already containing the new value, but will not be allowed to be added on create or update.
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Test Plan
+
+- Unit and integration tests covering:
+  - core changes
+  - correctness for namespace addition/removal, specifically label updates 
+    should be taken into account, but not in the middle of a scheduling cycle
+  - feature gate enabled/disabled
+- Benchmark Tests: 
+  - evaluate performance for the case where the selector matches a large number of pods 
+    in large number of namespaces
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation, and anything particularly
+challenging to test, should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+### Graduation Criteria
+
+#### Alpha -> Beta Graduation
+
+- Benchmark tests showing no performance problems 
+- No user complaints regarding performance/correctness.
+
+#### Beta -> GA Graduation
+    
+- Still no complaints regarding performance.
+- Allowing time for feedback
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. The KEP
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this enhancement:
+- [Maturity levels (`alpha`, `beta`, `stable`)][maturity-levels]
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc
+definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning)
+or by redefining what graduation means.
+
+In general we try to use the same stages (alpha, beta, GA), regardless of how the
+functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+#### Alpha -> Beta Graduation
+
+- Gather feedback from developers and surveys
+- Complete features A, B, C
+- Tests are in Testgrid and linked in KEP
+
+#### Beta -> GA Graduation
+
+- N examples of real-world usage
+- N installs
+- More rigorous forms of testing—e.g., downgrade tests and scalability tests
+- Allowing time for feedback
+
+**Note:** Generally we also wait at least two releases between beta and
+GA/stable, because there's no opportunity for user feedback, or even bug reports,
+in back-to-back releases.
+
+#### Removing a Deprecated Flag
+
+- Announce deprecation and support policy of the existing flag
+- Two versions passed since introducing the functionality that deprecates the flag (to address version skew)
+- Address feedback on usage/changed behavior, provided on GitHub issues
+- Deprecate the flag
+
+**For non-optional features moving to GA, the graduation criteria must include 
+[conformance tests].**
+
+[conformance tests]: https://git.k8s.io/community/contributors/devel/sig-architecture/conformance-tests.md
+-->
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
+
+## Production Readiness Review Questionnaire
+
+<!--
+
+Production readiness reviews are intended to ensure that features merging into
+Kubernetes are observable, scalable and supportable; can be safely operated in
+production environments, and can be disabled or rolled back in the event they
+cause increased failures in production. See more in the PRR KEP at
+https://git.k8s.io/enhancements/keps/sig-architecture/1194-prod-readiness.
+
+The production readiness review questionnaire must be completed and approved
+for the KEP to move to `implementable` status and be included in the release.
+
+In some cases, the questions below should also have answers in `kep.yaml`. This
+is to enable automation to verify the presence of the review, and to reduce review
+burden and latency.
+
+The KEP must have a approver from the
+[`prod-readiness-approvers`](http://git.k8s.io/enhancements/OWNERS_ALIASES)
+team. Please reach out on the
+[#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) channel if
+you need any help or guidance.
+
+-->
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: PodAffinityNamespaceSelector
+    - Components depending on the feature gate: kube-scheduler, kube-api-server
+  - [ ] Other
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+  No.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  Yes. One caveat is that pods that used the feature will continue to have
+  the NamespaceSelector field set even after disabling, however kube-scheduler
+  will not take the field into account.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+ It should continue to work as expected.
+
+* **Are there any tests for feature enablement/disablement?**
+  No, but we can do manual testing.
+
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  It shouldn't impact already running workloads. This is an opt-in feature since
+  users need to explicitly set the NamespaceSelector parameter in the pod spec,
+  if the feature is disabled the field is preserved if it was already set in the
+  presisted pod object, otherwise it is silently dropped.
+
+* **What specific metrics should inform a rollback?**
+  - A spike on metric `schedule_attempts_total{result="error|unschedulable"}`
+    when pods using this feature are added.
+  - A spike on `plugin_execution_duration_seconds{plugin="InterPodAffinity"}`.
+
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+No, will be manually tested.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+fields of API types, flags, etc.?**
+  No.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  The operator can query pods with the NamespaceSelector field set in pod affinity terms.
+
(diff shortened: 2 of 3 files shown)


keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+## Design Details
+
+ Add `NamespaceSelector` field to `PodAffinityTerm`:
+
+```go
+type PodAffinityTerm struct {
+    // A label query over the set of namespaces for which the term applies to.
+    // The term is applied to the union of the namespaces selected by this field
+    // and the ones listed in the `Namespaces` field.
+	NamespaceSelector *metav1.LabelSelector
```
wojtek-t, line 271: I started wondering about the following:
- anti-affinity is symetric, so we're ensuring that not only the pods we're scheduling satisfy it, but we also ensure that the pods that are already scheduled will have their anti-affinity satisfied
- with the introduction of NamespaceSelector, this relation is no longer static - in other words, by adding a label to a namespace, I can suddenly add new pods that actually should be considered for these relations

In other words, by adding labels to namespaces - I can suddenly violate existing anti-affinity, which up until now were always satisfied.

Did you think about implications of that?

ahg-g: I agree, if the labels change, the outcome will certainly change for future pods.

The other problem I see with namespace selector is performance. While for the incoming pod we can pre-calculate the set of namespaces, it might be expensive to do that for existing pods when evaluating their anti-affinity constraints.

This is likely an API anti-pattern, but what are your thought on changing it to regexp on the namespace name?

wojtek-t: I'm not huge fan of regexp - though you may want to ask API approvers about their opinion too.

What we can do though is use NamespaceSelector, but in the validation phase just accept the "kubernetes.io/metadata.name' label names (or whatever value we will decide for in #2162 - it's not a regexp, but will allow you to define a set of namespaces in a more native way...

wojtek-t: Actually - you already can do that with `Namespaces` field, so maybe that doesn't make that much sense...

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+By default, pod affinity/anti-affinity constraints are calculated against 
+the pods in the same namespace. The Spec allows users to exapnd that via
```
alculquicondor, line 139: s/exapnd/expand

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+-->
+
+## Motivation
+
+Pod affinity/anti-affinity API allows using this feature across namespaces using 
+a static list, which works if the user knowns the namespaces names in advance.
+
+However, there are cases where the namespaces are not known beforehand, for which there
+is no way to use pod affinity/anti-affinity. Allowing users to specify the set of namespaces
+using a namespace selector addresses this problem.
```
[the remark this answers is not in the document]
liggitt: affinity/anti-affinity already allows enumerating namespaces by name... I don't see how #2162 helps here...

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+As indicated in the comment, the scheduler will consider the union of the namespaces
+specified in the existing `Namespaces` field and the ones selected by the new 
+`NamespaceSelector` field. `NamespaceSelector`  is ignored when set to nil.
+
+The feature will be guarded by a new feature flag. When disabled, the api-server 
+will silently drop the `NamespaceSelector` field if set, moreover kube-scheduler 
+will ignore the field and continue to behave as before.
+
+
+<!--
```
alculquicondor, line 284: Add details on how scheduler will calculate matching namespaces.

Will namespace labels be part of the snapshot?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+    // A label query over the set of namespaces for which the term applies to.
+    // The term is applied to the union of the namespaces selected by this field
+    // and the ones listed in the `Namespaces` field.
+	NamespaceSelector *metav1.LabelSelector
+}
+```
+
+As indicated in the comment, the scheduler will consider the union of the namespaces
+specified in the existing `Namespaces` field and the ones selected by the new 
+`NamespaceSelector` field. `NamespaceSelector`  is ignored when set to nil.
```
alculquicondor, line 277: Is there a way for a NamespaceSelector to match any namespace?
Should we disallow that for security reasons?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+
+The main risk is performance, using namespace selector will make it easier for 
+users to specify affinity constraints across a large number of namespaces.
+
+The initial implementation of pod affinity/anti-affinity suffered from 
+performance challenges, however over the releases 1.17 - 1.20 we significantly
+improved that. 
+
+We currently have integration test and clusterloader benchmarks that evaluate the extreme
+cases of all pods having affinity/anti-affinity constraints to/against each other.
```
liggitt, line 245: with this proposal, the worst-case would now involve a namespace query and iterating over all the pods namespace-by-namespace

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
liggitt: I'm -1 on regexp/prefix matching against names... those are not strong boundaries (a regexp/prefix intended to match "$username-" namespaces would match also "$username-$othervalue-" namespaces)

The dynamic nature is not that different from label-based node affinity/anti-affinity... labels on nodes can change and invalidate previously scheduled pods, right?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+
+### Graduation Criteria
+
+#### Alpha -> Beta Graduation
+
+    - No complaints regarding performance.
+
+#### Beta -> GA Graduation
+    
+    - Still no complaints regarding performance.
```
liggitt, line 323: the bar for this should be higher... we should actively prove there are not performance problems with large numbers of namespaces and pods selected by affinity/anti-affinity rules, and the scheduler can keep up with pod creation/deletion and namespace label changes

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,663 @@
...
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Test Plan
+
```
liggitt, line 292: Explicitly call out correctness tests for namespace addition/removal/relabel scenarios. The scheduler's node caching has produced correctness problems in the past that we want to avoid here.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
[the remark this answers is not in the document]
liggitt: Is it important to select a subset of namespaces, or would it be sufficient to allow affinity/anti-affinity to apply across all namespaces (effectively making the pod labels globally meaningful)? That would be a simpler API surface and would remove the need to run label queries on namespaces

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: We currently iterate over all the pods on each node, and for each pod we test whether or not it is in an eligible namespace. 

However, there is indeed a performance concern when evaluating existing pods anti-affinity constraints: while for the incoming pod we can pre-calculate the set of namespaces, it might be expensive to do that for each existing pod when evaluating their anti-affinity constraints, pre-calculating and storing caching those for scheduled pods is not an option because the number of namespaces could be large, and because the labels of those namespaces could change.

Ideally the implementation shouldn't influence the API, but what are your thoughts on changing the API to a regexp field? it should be much faster to process than looking up namespaces via label selector.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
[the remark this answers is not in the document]
liggitt: links to previous discussions of the implications of this would be helpful (https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598, https://github.com/kubernetes/kubernetes/issues/43203#issuecomment-287237992)

The two main categories are performance (does this let a single pod make the scheduler do lots of expensive work) and security (does this let a badly behaved namespace disrupt other namespaces)

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: So this doesn't change how we iterate.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -294,9 +311,11 @@ proposal will be implemented, this is the place to discuss them.
 
 ### Test Plan
 
-- **Unit Tests**: core changes are covered by unit tests.
-- **Integration Tests**: core changes covered by integration tests.
-- **Benchmark Tests**: The performance should be as close as possible to the case where all pods are created in one namespace.
+- Unit and integration tests covering:
+  - core changes
+  - correctness for namespace addition/removal
```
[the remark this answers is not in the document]
ahg-g: changes should be taken into account, but not in the middle of a scheduling cycle.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -245,11 +248,14 @@ cases of all pods having affinity/anti-affinity constraints to/against each othe
...
-lookup its namespace object to match against its labels against the new selector.
-
-#### Security
+#### Security/Abuse
+see previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
 
+To address this concern, admission webhooks like gatekeeper can be used to restrict
```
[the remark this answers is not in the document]
ahg-g: I added the inverse.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,708 @@
...
+
+#### Security/Abuse
+see previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
+
+```
+<<[UNRESOLVED]>>
+We will partially address this concern by rejecting empty selectors and only
```
alculquicondor, line 259: lgtm. @liggitt ?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,709 @@
...
+### Notes/Constraints/Caveats (Optional)
+
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
```
Huang-Wei, line 239: Add a section on `External Integrator`? Specifically, descheduler may need to reimplement their logic to be aligned with the new API semantics.

alculquicondor: Doesn't descheduler use the framework plugins directly? @damemi

damemi: @alculquicondor It does not, being able to use the plugins directly is a motivator for the work to externalize the framework

ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: resolving this, lets discuss the api under the other [comment](https://github.com/kubernetes/enhancements/pull/2251#discussion_r555915819)

ahg-g:
@liggitt can you please take a look and review if your concerns are now addressed?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,662 @@
...
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+I am running a SaaS service, the workload for each customer is placed in a separate
+namespace. The workloads requires 1:1 pod to node placement. I want to use pod 
+anti-affinity across all customers namespaces to achieve that.
```
devdattakulkarni, line 234: Can you outline how this requirement would be achieved with the proposed change?

Presumably the namespaces are created on-the-fly for each customer so their names are not known ahead of time. But the labels on the namespaces would be known (say, something like stack=customer1, stack=customer2, etc.). So is the idea that for each customer's Pod somehow the PodAffinityTerm's Namespace selector would be specified specific to that customer in case of affinity? (say, for customer1's Pods it would be: 'stack=customer1') And for anti-affinity, it would be a list of all other labels (say, for customer1's Pods it would be: 'stack=customer2', 'stack=customer3', and so on).

Separately, you mention about 1:1 pod to node placement. Wouldn't this be possible with Pod's nodeSelector attribute? How is node placement related to namespace selection with regards to affinity/anti-affinity?

ahg-g: > Can you outline how this requirement would be achieved with the proposed change?
> 
> Presumably the namespaces are created on-the-fly for each customer so their names are not known ahead of time. But the labels on the namespaces would be known (say, something like stack=customer1, stack=customer2, etc.). So is the idea that for each customer's Pods somehow the PodAffinityTerm's Namespace selector would be specified specific to that customer in case of affinity? (say, for customer1's Pods it would be: 'stack=customer1') And for anti-affinity, it would be a list of all other labels (say, for customer1's Pods it would be: 'stack=customer2', 'stack=customer3', and so on).

I am not quite following your example, but the idea is correct: while the namespaces may not be known in advance because they are created dynamically, they could be created with a common label. To achieve affinity/anti-affinity across those namespaces, the pods would be created with a term that uses a NamespaceSelector targeting the common namespace label.

> 
> Separately, you mention about 1:1 pod to node placement. Wouldn't this be possible with Pod's nodeSelector attribute? How is node placement related to namespace selection with regards to affinity/anti-affinity?

The component that creates the pods wouldn't know which nodes have capacity/assigned and which are not, OTOH the scheduler does and is able to pick the node that both has capacity and satisfies the affinity constraints (among other possible things). Moreover, you may want to achieve 1:1 not across all possible pods, but across a specific set (e.g., only one workload pod per node, but the node could host other control plane pods like daemonsets)

devdattakulkarni: Agree on the point that the Pod creating component may not know about node capacities.
Regarding achieving 1:1 mapping between workload pods and nodes, I guess anti-affinity rule with Namespace label selector will make it possible. Ok.

ahg-g: > Regarding achieving 1:1 mapping between workload pods and nodes, I guess anti-affinity rule with Namespace label selector will make it possible. Ok.

Yes, exactly, and this is a fairly common case in HPC

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,662 @@
...
+We currently have integration test and clusterloader benchmarks that evaluate the extreme
+cases of all pods having affinity/anti-affinity constraints to/against each other.
+Those benchmarks show that the scheduler is able to achieve maximum throughput 
+(i.e., the api-server qps limit).
+
+#### Security/Abuse
+see previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
```
liggitt, line 263: the limited operations/validation below helps mitigate disrupting all namespaces... I can't tell if this enhancement makes it easier to disrupt specific namespaces, or is that already possible via that existing anti-affinity namespace field.

Is there opt-in required on both sides for pod scheduling to be disrupted, or can a user in namespace `foo` currently prevent regular `kube-system` pods from being scheduled by setting up an anti-affinity rule like:

```yaml
podAntiAffinity:
  requiredDuringSchedulingIgnoredDuringExecution:
  - labelSelector: {}
    namespaces: ["kube-system"]
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
```diff
@@ -0,0 +1,45 @@
...
+authors:
+  - "@ahg-g"
+owning-sig: sig-scheduling
+participating-sigs:
+status: implementable
+creation-date: 2021-01-11
+reviewers:
+  - "@alculquicondor"
+  - "@Huang-Wei"
+  - "@liggit"
```
liggitt, line 12: typo here and below: liggitt

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
liggitt: I think this makes sense... perhaps a brief example of what could and could not be expressed with these restrictions would be helpful.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
```diff
@@ -0,0 +1,45 @@
+title: Namespace Selector for Pod Affinity
+kep-number: 2249
+authors:
+  - "@ahg-g"
+owning-sig: sig-scheduling
+participating-sigs:
+status: implementable
```
liggitt, line 7: should still be provisional if there are UNRESOLVED sections

ahg-g: the plan is to get those resolved before merging.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: sg, added.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
ahg-g: sorry, done

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
liggitt: Are we happy with that? That seems like a mistake in the existing feature to me... do we want to continue to allow that with this new capability or try to address that possibility for cross-namespace interference?

ahg-g: > Are we happy with that? That seems like a mistake in the existing feature to me... 

Are we able to fix this now? Wouldn't restricting `labelSelector` now be a backward incompatible change?

> do we want to continue to allow that with this new capability or try to address that possibility for cross-namespace interference?

You mean limit labelSelector only if namespaceSelector is set? I am fine with that, but wouldn't that surprise users that it is allowed in some cases and not in others?

alculquicondor: > No opt-in on both sides, an existing pod can already prevent other pods from being scheduled on the node, the example you gave would prevent kube-system pods from being scheduled on the node where the pod with that term is scheduled.

On the flip side, it prevents the incoming Pod running on a Node where there is already a kube-system Pod scheduled.

ahg-g: > On the flip side, it prevents the incoming Pod running on a Node where there is already a kube-system Pod scheduled.

Yeah, but that is not really useful. Separating system from workload pods should be done using taint/tolerations and perhaps node affinity.

btw, since system pods are usually created with the highest priority, the existing pod will likely get preempted if it is preventing the system pod from getting scheduled via anti-affinity.

ahg-g:
/assign @wojtek-t 

I think this is ready for PRR review, only one open comment needs to be resolved.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: actual Equals and DoubleEquals don't make sense [1], the only one that does is "In", updated.

[1] https://github.com/kubernetes/kubernetes/blob/41a42ef6753a3fc2f90fa110c4a6dea0aa8304e9/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1114

alculquicondor: What about Exists?

ahg-g: Problematic because namespaces will have common labels: https://github.com/kubernetes/enhancements/pull/2162, so users can easily target all namespaces as follows 

```
matchExpressions:
        - key: kubernetes.io/metadata.name
          operator: Exists
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
liggitt: > > Are we happy with that? That seems like a mistake in the existing feature to me...
> 
> Are we able to fix this now? Wouldn't restricting `labelSelector` now be a backward incompatible change?

The current proposal doubles down on letting one namespace unilaterally prevent scheduling of pods from another namespace. Anyone who currently didn't want to allow that in their cluster would have put some sort of custom webhook admission policy check in place to prevent affinity rules using the `namespaces` field. This proposal adds a new field that exposes the same cross-namespace capability.

Put another way, a cluster someone had locked down to prevent cross-namespace interference would be vulnerable again after this was implemented unless they notice and make changes to their custom policy enforcement.

> > do we want to continue to allow that with this new capability or try to address that possibility for cross-namespace interference?
> 
> You mean limit labelSelector only if namespaceSelector is set? I am fine with that, but wouldn't that surprise users that it is allowed in some cases and not in others?

No, I wasn't suggesting changing validation between the fields. I was asking if use of this cross-namespace capability should be opt-in for the cluster admin, rather than opt-out.

ahg-g: Most clusters would want this enabled, so I don't think we want this feature to be opt-in. The feature will be disabled by default in alpha, so that will give people a full release to make the change to their custom policies. So I suggest we keep it opt-in and stress the point regarding custom policies in the release notes, what do you think?

alculquicondor: Is there an alternative way to make an API opt-in beyond alpha? That is, when the feature gate is enabled by default.

liggitt: It's not exactly opt-in, but defining a quota scope for cross-namespace affinity/anti-affinity pods would let a cluster admin easily set up a default cluster wide quota to say "by default, cross-namespace affinity/antiaffinity pods are not allowed unless I explicitly grant quota for them in a namespace". That would be coherent with the approach to priorityclass pods

ahg-g: Discussed offline. Practically adding a scope is like an on/off knob to control which namespaces can include pods that set namespaces/namespaceSelector fields (the actual hard limit is not going to be that important since a single pod can do the damage we are concerned about).

I will update the KEP to include proposing a new quota scope, so NamespaceSelector and "CrossNamespaceAffinity" will be done under one feature flag.

ahg-g: I updated the KEP to include the new quota scope. PTAL.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -251,27 +259,58 @@ constraints across a large number of namespaces. The initial implementation
...
+  configuration:
+    apiVersion: apiserver.config.k8s.io/v1
+    kind: ResourceQuotaConfiguration
+    limitedResources:
+    - resource: pods
+      matchScopes:
+      - scopeName: CrossNamespaceAffinity
+```
+
+If admins want allow that by default, but disable it for specific namespaces, then they
```
alculquicondor, line 296: what is the mechanism to disallow by default but only allow for specific namespaces?

ahg-g: the one described above, configuring it as a limited resource by default.

ahg-g: rephrased to hopefully make it clearer

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,737 @@
...
+  the scheduling cycle of a pod.
+
+Finally, the feature will be guarded by a new feature flag. If the feature is 
+disabled, the field `NamespaceSelector` is preserved if it was already set in
+the persisted Pod ojbect, otherwise it is silently dropped; moreover kube-scheduler 
+will ignore the field and continue to behave as before.
+
+
+With regards to adding the `CrossNamespaceAffinity` quota scope, there is no special design
+aspects to that.
```
liggitt, line 410: note this requires allowing a new value in the `scopeName` field, so will be rolled out in multiple releases. when the feature is disabled, the new value will be tolerated in updates of objects already containing the new value, but will not be allowed to be added on create or update

liggitt (commented):
one comment on the resourcequota scope rollout, lgtm otherwise

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: updated.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
```diff
@@ -0,0 +1,45 @@
+title: Namespace Selector for Pod Affinity
+kep-number: 2249
+authors:
+  - "@ahg-g"
+owning-sig: sig-scheduling
+participating-sigs:
```
alculquicondor, line 6: sig-auth? @liggitt ?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
```diff
@@ -0,0 +1,45 @@
...
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.21"
+  beta: "v1.22"
+  stable: "v1.24"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: PodAffinityNamespaceSelector
```
ahg-g, line 37: to force graduating the two features together(namespaceSelector and the new scope), I am using a single feature flag.

alculquicondor:
lgtm, but I have no powers :)

keps/prod-readiness/sig-scheduling/2249.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2249
+alpha:
+  approver: "@wojtek-t"
```
johnbelamaric, line 3: Re-assign to me, Wojtek is over-subscribed

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
```diff
@@ -0,0 +1,45 @@
...
+creation-date: 2021-01-11
+reviewers:
+  - "@alculquicondor"
+  - "@Huang-Wei"
+  - "@liggitt"
+approvers:
+  - "@Huang-Wei"
+  - "@liggitt"
+prr-approvers:
+  - "@wojtek-t"
```
johnbelamaric, line 17: Re-assign to me, Wojtek is over-subscribed

keps/prod-readiness/sig-scheduling/2249.yaml
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,739 @@
...
+
+### Test Plan
+
+- Unit and integration tests covering:
+  - core changes
+  - correctness for namespace addition/removal, specifically label updates 
+    should be taken into account, but not in the middle of a scheduling cycle
+  - feature gate enabled/disabled
+- Benchmark Tests: 
+  - evaluate performance for the case where the selector matches a large number of namespaces
```
wojtek-t, line 429: In particular, I would be interested in seeing a test, where most of pods would be covered (e.g. all namespaces by kube-system or sth like that).

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,739 @@
...
+  the NamespaceSelector field set even after disabling, however kube-scheduler
+  will not take the field into account.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+ It should continue to work as expected.
+
+* **Are there any tests for feature enablement/disablement?**
+  No, but we can do manual testing.
+
+The rest of the readiness questionnaire will be completed when graduating the feature to Beta.
```
wojtek-t, line 579: It seems you did that already :)

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,739 @@
...
+provider?**
+No.
+
+* **Will enabling / using this feature result in increasing size or count of 
+the existing API objects?**
+Yes, if users set the NamespaceSelector field.
+
+* **Will enabling / using this feature result in increasing time taken by any 
+operations covered by [existing SLIs/SLOs]?**
+No.
```
wojtek-t, line 664: Technically, scheduling (which is part of pod-startup time) may increase.
I know you have a plan to keep it under control, but it's probably worth mentioning anyway

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
wojtek-t: +1 for having SIG auth here.

wojtek-t (commented):
I'm fine with this, but I would like to see explicit approval from @liggitt first before approving

keps/sig-scheduling/2249-pod-affinity-namespace-selector/kep.yaml
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: added large number of pods explicitly. The plan is to have every pod with affinity to every other pod (a test we already have, but in a single namespace).

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: removed.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: clarified.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,739 @@
...
+      matchScopes:
+      - scopeName: CrossNamespaceAffinity
+```
+
+With the above configuration, pods can use `namespaces` and `namespaceSelector` only
+if the namespace where they are created have a resource quota object with 
+`CrossNamespaceAffinity` scope and a hard limit equal to the number of pods that are
+allowed to.
+
+Moreover, to prevent accidentally selecting a large number of namespaces, we will reject empty
```
ahg-g, line 319: @liggitt I am inclined to drop this requirement since we have the quota scope. This will make it harder to automate 1:1 pod placement.

alculquicondor: What exactly are you proposing to change? I think only allowing `In` is good. But the value of the label could be an empty string.

ahg-g: Proposing to have no constraints on NamespaceSelector, just like LabelSelector. While restricting to `In` could help preventing accidental uses of the field, it means namespaces should have a common label, which may be tricky to achieve in some setups (e.g. a provider trying to automate 1:1 pod to node placement).

alculquicondor: So you are saying you want to support "match any namespace"

alculquicondor: I think a provider would at least want to have "kube-system" out of the affected namespaces

ahg-g: Users could say "NotIn" kube-system (https://github.com/kubernetes/enhancements/pull/2162 will add a label for all namespaces that represent the namespace name).

ahg-g: I guess we could still disallow empty selector to avoid accidental uses of the field. NotIn should suffice.

ahg-g: I updated the KEP that we will only disallow empty selector to avoid accidental selection of all namespaces. Other operators are allowed.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+
+We will address this concern by introducing a new quota scope named `CrossNamespaceAffinity`
+that admins can use to limit which namespaces are allowed to have pods with affinity terms
+that set the existing `namespaces` field or the proposed one `namespaceSelector`. 
+
+Using this new scope, admins can prevent certain namespaces from having pods that use
+cross-namespace pod affinity by creating a resource quota object in that namespace with
+`CrossNamespaceAffinity` scope and hard limit of 0:
+
+```go
```
Huang-Wei, line 281:
```suggestion
```yaml
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+      - scopeName: CrossNamespaceAffinity
+```
+
+
+If admins want to disallow using `namespaces` and `namespaceSelector` by default, and 
+only allow it for specific namespaces,  they could configure `CrossNamespaceAffinity` 
+as a limited resource by setting the kube-apiserver flag --admission-control-config-file
+to the path of the following configuration file:
+
+```go
```
Huang-Wei, line 300:
```suggestion
```yaml
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector: {
+      matchExpressions:
+        - key: workload
+          operator: In
+          values:
+          - HPC
+  }
```
Huang-Wei, line 332: remove this line

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+With the above configuration, pods can use `namespaces` and `namespaceSelector` only
+if the namespace where they are created have a resource quota object with 
+`CrossNamespaceAffinity` scope and a hard limit equal to the number of pods that are
+allowed to.
+
+Moreover, to prevent accidentally selecting a large number of namespaces, we will reject empty
+selectors. For example, users can do the following:
+
+```go
+
```
Huang-Wei, line 323: remove this line

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+With the above configuration, pods can use `namespaces` and `namespaceSelector` only
+if the namespace where they are created have a resource quota object with 
+`CrossNamespaceAffinity` scope and a hard limit equal to the number of pods that are
+allowed to.
+
+Moreover, to prevent accidentally selecting a large number of namespaces, we will reject empty
+selectors. For example, users can do the following:
+
+```go
```
Huang-Wei, line 322:
```suggestion
```yaml
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+```
+
+but can't do the following:
+
+```go
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector: {}
+  }
+
```
Huang-Wei, line 342: remove these 2 lines

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+        - key: workload
+          operator: In
+          values:
+          - HPC
+  }
+```
+
+but can't do the following:
+
+```go
```
Huang-Wei, line 337:
```suggestion
```yaml
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,736 @@
...
+See previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
+
+We will address this concern by introducing a new quota scope named `CrossNamespaceAffinity`
+that admins can use to limit which namespaces are allowed to have pods with affinity terms
+that set the existing `namespaces` field or the proposed one `namespaceSelector`. 
+
+Using this new scope, admins can prevent certain namespaces from having pods that use
```
Huang-Wei, line 277: Is there a performance concern? As it implies upon a Pod's creation/update, if the Pod's podAffinity has namespace/namespaceSelector, you have to iterate each matched namespace, and verify if each one has a `CrossNamespaceAffinity` RQ and check if it eventually matches.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: Not really because we are only checking if the pod is using either field, we are not doing any matching.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
alculquicondor: The KEP still says: `The only valid operator is In.` is that the intention?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+spec:
+    hard:
+      pods: "0"
+    scopeSelector:
+      matchExpressions:
+      - scopeName: CrossNamespaceAffinity
+```
+
+
+If admins want to disallow using `namespaces` and `namespaceSelector` by default, and
```
alculquicondor, line 295: admins or providers?
That is, do admins need access to the control plane?

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: This paragraph is correct. I updated the comment on the field.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: admins. 
They don't need access to the control plane.

ahg-g: changed to use a more neutral term: operators.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
Huang-Wei: Hmm... I suppose you mean this admission check only applies to a Pod whose namespace has a `CrossNamespaceAffinity` RQ.

This _may_ not suffice to prevent some illegal usage: say a Pod's namespace is "foo-ns", and anti-affinity specifies `namespaces: ["kube-system", "foo-ns"]`, and `kube-system` is employed with a RQ having 0 pods on scope `CrossNamespaceAffinity`. In this case, this Pod still prevents further pods in namespace "kube-system" and "foo-ns" from scheduling (suppose the affinity term's labelSelector matches) on its node, right?

ahg-g: But the RQ should be created for the `foo-ns` namespace, not the other way around.

ahg-g: I updated the example, PTAL.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+
+We will address this concern by introducing a new quota scope named `CrossNamespaceAffinity`
+that operators can use to limit which namespaces are allowed to have pods with affinity terms
+that set the existing `namespaces` field or the proposed one `namespaceSelector`. 
+
+Using this new scope, operators can prevent certain namespaces from having pods that use
+cross-namespace pod affinity by creating a resource quota object in that namespace with
+`CrossNamespaceAffinity` scope and hard limit of 0:
+
+```go
```
Huang-Wei, line 281: Unresolved actually :)

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+    name: DisableCrossNamespaceAffinity
```
Huang-Wei, line 285: nit: use 2 spaces as indents.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+```yaml
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector: {
+      matchExpressions:
+        - key: workload
+          operator: In
+          values:
+          - HPC
+  }
```
Huang-Wei, line 331: Still unresolved.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+See previous discussion [here](https://github.com/kubernetes/kubernetes/issues/68827#issuecomment-468470598)
+NamespaceSelector will allow users to select all namespaces. This may cause a security 
+concern: a pod with anti-affinity constraint can block pods from all other namespaces 
+from getting scheduled in a failure domain.
+
+We will address this concern by introducing a new quota scope named `CrossNamespaceAffinity`
+that operators can use to limit which namespaces are allowed to have pods with affinity terms
+that set the existing `namespaces` field or the proposed one `namespaceSelector`. 
+
+Using this new scope, operators can prevent certain namespaces from having pods that use
```
Huang-Wei, line 277:
```suggestion
Using this new scope, operators can prevent certain namespaces (`foo-ns` in the example below) from having pods that use
```

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,734 @@
...
+  }
+```
+
+but can't do the following:
+
+```yaml
+podAntiAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+  - namespaceSelector: {}
+  }
```
Huang-Wei, line 340: Still unresolved :)

Huang-Wei (commented):
Some nits. LGTM otherwise.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,732 @@
...
+## Design Details
+
+ Add `NamespaceSelector` field to `PodAffinityTerm`:
+
+```go
+type PodAffinityTerm struct {
+    // A label query over the set of namespaces that the term applies to.
+	// The term is applied to the union of the namespaces selected by this field
+	// and the ones listed in the namespaces field.
+	// nil selector and nil or empty namespaces list means "this pod's namespace"
```
liggitt, line 373: nit: `a nil selector and empty namespaces list means "this pod's namespace` (both must be unset)

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,732 @@
...
+	// An empty selector is not valid.
```
liggitt, line 374: clarify what "empty" means here:
> An empty selector (`{}`) is not allowed.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
```diff
@@ -0,0 +1,732 @@
...
+
+As indicated in the comment, the scheduler will consider the union of the namespaces
+specified in the existing `Namespaces` field and the ones selected by the new 
+`NamespaceSelector` field. `NamespaceSelector`  is ignored when set to nil.
+
+We will do two precomputations at PreFilter/PreScore:
+
+- The names of the namespaces selected by the `NamespaceSelector` is computed. This set
+  will be used by Fliter/Score to match against existing pods namespaces.
+- A snapshot of the namespace labels of the incoming pod. This will be used when
```
liggitt, line 387: does this mean "a snapshot of the labels of the namespace containing the incoming pod"?

liggitt:
a few clarifying comments, lgtm overall

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: yes, rephrased.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: done.

keps/sig-scheduling/2249-pod-affinity-namespace-selector/README.md
ahg-g: updated.

ahg-g:
> looks good for squash

@liggitt did you look at the latest diff? good to squash?

liggitt:
yeah, good to squash

ahg-g:
Squashed, thanks all for the reviews.

alculquicondor:
LGTM

Issue #2220, opened 2020-12-29
Migrate all KEPs to the new template


wojtek-t:
This will facilitate strenghtening KEP validation

wojtek-t:
The missing remaining itmes:

sig-apimachinery:
- /20200309-consistent-resource-versions-semantics   [requires creating an issue]

sig-cloud-provider:
- 3 cloud-provider KEPS (alibaba, baiducloud, huaweicloud)

sig-cluster-lifecycle:
- a bunch of stuff

sig-instrumentation:
- /1209-metrics-stability - [requires a bit of cleanup]

sig-network:
-  #2446 addresses most
- node-local-dns needs to be merged

shekhar-rajak:
- I need clarification for "sig-cluster-lifecycle". I see directories and sub-directories. Should we just remove the sub-directories under the "sig-cluster-lifecycle" and follow the same pattern as others ? 

- Same for sig-instrumentation /1209-metrics-stability directory

wojtek-t:
> I need clarification for "sig-cluster-lifecycle". I see directories and sub-directories. Should we just remove the sub-directories under the "sig-cluster-lifecycle" and follow the same pattern as others ?

No - this splits it to kind of "subprojects". We should basically put the new templates there, e.g. there will be
/sig-cluster-lifecycle/kubeadm/12345-kep-title/(first kep)
/sig-cluster-lifecycle/kubeadm/23466-kep-title/(second kep)

> Same for sig-instrumentation /1209-metrics-stability directory

Let's leave it for now - I will handle it later.

wojtek-t:
@shekhar-rajak - if would be able to help with cluster-lifecycle, cloud-provider and api-machinery remaining ones it would be huge help - I will handle the remaining networking and instrumentaiton one.

shekhar-rajak:
Sure @wojtek-t , I will work on it.

neolit123:
one problem here is that all tracking issues get detached from the KEP locations. i can try updating the SIG CL tracking issues next week.

another problem is that KEP cross-ref links get broken.
i don't think we have many of these in SIG CL, but other SIGs might have them.

wojtek-t:
> another problem is that KEP cross-ref links get broken.
i don't think we have many of these in SIG CL, but other SIGs might have them.

If needed, we can add symlinks to the new locations (there are a few for other SIGs).

wojtek-t:
There are just few more:
```
[...]

PR #2252, opened 2021-01-12
Migrate cgroup-v2 kep to new template

giuseppe:
ref #2220

### DIFF
--- a/keps/sig-node/2254-cgroup-v2/README.md
+++ b/keps/sig-node/2254-cgroup-v2/README.md
@@ -1,27 +1,3 @@
----
-title: cgroups v2
-authors:
-  - "@giuseppe"
-owning-sig: sig-node
-participating-sigs:
-      - sig-architecture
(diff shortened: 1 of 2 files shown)


keps/sig-node/1370-cgroup-v2/kep.yaml
```diff
@@ -0,0 +1,21 @@
+title: cgroups v2
```
wojtek-t, line 1: It shouldn't be KEP number 1370.

You should open an issue in this repo, and the KEP number should be the number of that issue.

giuseppe: thanks, opened a new issue and linked here

wojtek-t:
Thank you!

/lgtm
/approve

derekwaynecarr:
thanks @giuseppe 

/approve
/lgtm

PR #2272, opened 2021-01-15
Promote pod deletion cost KEP to implementable

ahg-g:
#2255

/sig apps

### DIFF
--- a/keps/sig-apps/2255-pod-cost/README.md
+++ b/keps/sig-apps/2255-pod-cost/README.md
@@ -1,4 +1,4 @@
-# KEP-2255: Add pod-cost annotation for ReplicaSet
+# KEP-2255: ReplicaSet Pod Deletion Cost
 
 
 <!-- toc -->
@@ -27,7 +27,6 @@
   - [Scalability](#scalability)
   - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
-- [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
 <!-- /toc -->
 
@@ -55,122 +54,164 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 ## Summary
 
-This feature allows making a suggestion to the ReplicaSet controller, which pod of a Deployment should be deleted first when a scale-down event happens. This can prevent session disruption in stateful applications in a trivial manner.
+This feature allows applications to give a hint to the ReplicaSet controller 
+as to which pods should be deleted first on scale down.
 
 ## Motivation
 
-For some applications, it is necessary that the application can tell Kubernetes which pod can be deleted and which replica has to be protected. The reason for this is that some applications do have stateful sessions and it is not possible to put such an application into Kubernetes because of session termination resulting from "random" down-scale. If the application is able to tell Kubernetes which of the replicas contains no/few/less important active sessions, this would solve many problems. This feature is non-disruptive to the default behaviour. Only if the annotation is existing, it will make a difference in deletion order.
+Currently ReplicaSets are scaled down based on a criteria that on the
+limit prioritizes deleting pods with a more recent creation/readiness 
+timestamp. This is not ideal for some applications where the cost of 
+deleting pods is not related to how recent they were created. 
 
 ### Goals
 
-To recommend which pod gets deleted next of a ReplicaSet. This should help to avoid major reworks in existing applications architecture:
-* [45509](https://github.com/kubernetes/kubernetes/issues/45509) - Scale down a deployment by removing specific pods
+- An API that allows applications to influence the order of deleting 
+  pods when scaling down a ReplicaSet
 
 
 ### Non-Goals
 
-Guaranteed (in contrast to the recommendation stated in Goals) deletion of a selected replica.
+- Guarantees on pod deletion order
+- A controller that sets the cost of deleting the pods.
 
 ## Proposal
 
-The application can set the `controller.kubernetes.io/pod-cost` annotation to a pod through the Kubernetes API. When a downscale event happens, the pod with the lower priority value of the previously set annotation will be deleted first. If one pod of the Deployment has no priority annotation set, it will be treated as the lowest priority.
-
-If all pods have the same priority, there is no difference in the normal pod delete decision behaviour. The same applies if the pod-cost annotation is not used at all.
-
-The pod-cost annotation can be changed during operation, for example, if workload changes or a new master gets elected.
+Define a known annotation, namely `controller.kubernetes.io/pod-deletion-cost` that 
+applications can set to offer a hint on the cost of deleting a pod compared
+to other pods belonging to the same ReplicaSet. 
 
 ### User Stories (optional)
 
-
 #### Story 1
 
-In an application environment with stateful worker (user-)sessions, it is essential to keep the user sessions alive as good as possible. In case of a scale-down event, the application has to tell the scheduler, which delete decision would have the lowest impact on existing sessions.
+The different pods of an application could have different utilization levels. 
+On scale down, the application may prefer to remove the pods with lower utilization.
+To avoid frequently updating the pods, the application should update pod-deletion-cost
+once before issuing a scale down. This works if the application itself controls the down 
+scaling (e.g., the driver pod of a Spark deployment).
 
 #### Story 2
 
-An application consists of identical server processes, but one of the replicas will be the master, which should be kept as long as possible. All other replicas can be treated as cattle workload. Then the master can set the priority annotation with a high priority value as soon as it has finished its startup process. The other replicas can remain either without any priority set, or e.g. with all the same, lower priority. This ensures, that the master replica of this deployment will be protected in a downscale situation.
+On scale down, the application may want to remove pods running on the most expensive 
+nodes first. For example, remove pods from nodes running on standard VMs first
+then from ones running on preemptible/spot VMs (which can be 80% cheaper than standard VMs).
 
 
 ### Risks and Mitigations
 
-On previous Kubernetes ReplicaSet controller versions that don't implement the pod-cost annotation feature, the same application might make false assumptions about the protection of a master instance or workers with open (user-)sessions on it. As the pod-cost annotation would be only a suggestion to the ReplicaSet controller, the application developer should, however, handle the case of a failed master instance or broken user sessions. The feature is just an improvement, not a guarantee, as there might happen timing issues between setting the annotation and the next controller scale-down event.
+- Users perceive the feature as a guarantee to delete order. Documentation
+should stress the fact that this is best effort.
+
+- Users deploy controllers that update the annotation frequently causing a
+  significant load on the api server. Documentation should include best 
+  practices as to how this feature should be used (e.g., update the
+  pod-deletion-cost only before scale down). Moreover, [API priority and fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/) 
+  gives operators a new server-side knob that allows them to limit update
+  qps issued by such controllers.
+  
 
 ## Design Details
 
+The pod-deletion-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be rejected by the api-server
+with a BadRequest status code.
+
+Having the default value in the middle of the range allows controllers to cutomize
+the semantics of the cost of deleting pods that don't have the annotation set: 
+controllers can use positive pod-deletion-cost values if they always want uninitialized
+pods to be deleted first, or use negative pod-deletion-cost values if they want
+uninitialized pods to always be deleted last.
+
+When scaling down a ReplicaSet, controller-manager will prioritize deleting
+pods with lower pod-deletion-cost. Specifically, the pod-deletion-cost will be evaluated after
+step 3 and before step 4 as they are currently defined in
+[ActivePodsWithRanks](https://github.com/kubernetes/kubernetes/blob/cac933934b1301665e6e51a81c66c483f4e16c49/pkg/controller/controller_utils.go#L784-L809),
+which means the followig criteria is applied when comparing two pods regardless of their pod-deletion-cost:
+- if one is assigned a node and the other is not, then the unassigned pod is deleted first.
+- if the two pods are in different phases, then the pod in pending/unknown status is deleted first.
+- if the two pods have different readiness status, then the not ready pod is deleted first
+
+
+If none of the pods set the pod-deletion-cost annotation or all of them have the same value, then the 
+scale down behavior is not changed compared to now.
 
 ### Test Plan
 
-* Units test in kube-controller-manager package to test a variety of scenarios.
-* New E2E Tests to validate that replicas get deleted as expected e.g:
- * Replicas with lower pod-cost before replicas with higher pod-cost
- * Replicas with no pod-cost annotation set before replicas with low priority
+- Units test in kube-controller-manager package to test a variety of scenarios.
+- Integration tests to validate that:
+  - Replicas with lower pod-deletion-cost are deleted before replicas with higher pod-deletion-cost
+  - No behavior change when pod-deletion-cost is not set or all pods have the same pod-deletion-cost
 
 ### Graduation Criteria
 
 #### Alpha -> Beta Graduation
 * Implemented feedback from alpha testers
-* Thorough E2E and unit testing in place
 
 #### Beta -> GA Graduation
-* Significant number of end-users are using the feature
 * We're confident that no further API changes will be needed to achieve the goals of the KEP
 * All known functional bugs have been fixed
 
 ### Upgrade / Downgrade Strategy
 
-When upgrading no changes are needed to maintain existing behaviour as all of this behaviour is fully optional and disabled by default. To activate this feature either a user has to make an annotation to a pod in a Deployment by hand or the application annotates a pod in a Deployment through the API.
-
-When downgrading, there is no need to changing anything, as this is just a pod annotation, which is uncritical.
+There is no strategy per se. On upgrade, controller-manager will start taking into account
+pod-deletion-cost annotation for new and existing ReplicaSets that set the annotation. On 
+downgrade, controller-manager will stop taking into account pod-deletion-cost, and so
+reverting to old behavior.
 
 ### Version Skew Strategy
 
-As this feature is based on pod annotations, there is no issue with different Kubernetes versions. The lack of this feature in older versions may change the efficiency and reliability of the applications.
+N/A
 
 ## Production Readiness Review Questionnaire
 
 ### Feature enablement and rollback
 
 * **How can this feature be enabled / disabled in a live cluster?**
-  - [x] Other
-    - Make special pod annotations within a live Deployment
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: ReplicaSetPodDeletionCost
+    - Components depending on the feature gate: kube-controller-manager
+  - [ ] Other
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
 
 
 * **Does enabling the feature change any default behavior?**
-  - No
+ No.
 
 
 * **Can the feature be disabled once it has been enabled (i.e. can we rollback
   the enablement)?**
-  - One can either remove the annotations or downgrade to an older Kubernetes release
+Yes.
 
 
 * **What happens if we reenable the feature if it was previously rolled back?**
-  - Then the feature will be reenabled. Nothing special to consider here.
+It should continue to work as expected.
 
 
 * **Are there any tests for feature enablement/disablement?**
-
+We will add unit tests. 
 
 ### Rollout, Upgrade and Rollback Planning
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can a rollout fail? Can it impact already running workloads?**
-  - As the feature is a simple annoation, the worst what could happen is that either the annotation is lost or ignored. In the worst case, a pod with a higher priority gets deleted before a pod with a lower priority.
-
+It shouldn't impact already running workloads. This is an opt-in feature
+since users need to explicitly set the annotation.
 
 * **What specific metrics should inform a rollback?**
-  - None
-
+None.
 
 * **Were upgrade and rollback tested? Was upgrade->downgrade->upgrade path tested?**
-  - Was tested. Behaviour change in both directions, as expected.
-
+We will do manual testing.
 
 * **Is the rollout accompanied by any deprecations and/or removals of features,
   APIs, fields of API types, flags, etc.?**
-  - No. However, the exact same pod annotation string cannot be used for any other purposes.
-
+No.
 
 ### Monitoring requirements
 
@@ -179,39 +220,30 @@ _This section must be completed when targeting beta graduation to a release._
 * **How can an operator determine if the feature is in use by workloads?**
   - Search for pod annotations with the exact same pod-cost annotation string.
 
-
 * **What are the SLIs (Service Level Indicators) an operator can use to
   determine the health of the service?**
-  - A pod with a lower pod-cost annotation in a Deployment gets deleted first on a scale-down event.
-
+N/A
 
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  - All pods with a lower pod-cost annotation in a Deployment are deleted first on a scale-down event.
+N/A
 
 * **Are there any missing metrics that would be useful to have to improve
   observability if this feature?**
-  - N/A
+No.
 
 ### Dependencies
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **Does this feature depend on any specific services running in the cluster?**
-  - The feature requires the existing of the kube-controller-manager and the ability and permissions to set pod annotations.
-
+No.
 
 ### Scalability
 
-_For alpha, this section is encouraged: reviewers should consider these questions
-and attempt to answer them._
-
-_For beta, this section is required: reviewers must answer these questions._
-
-_For GA, this section is required: approvers should be able to confirms the
-previous answers based on experience in the field._
-
 * **Will enabling / using this feature result in any new API calls?**
-  - Whenever the application decides, that a change in pod-cost is needed for a replica, it will send out an API request and set the appropriate pod annotation(s).
+  - No, not the feature itself. However, users will want to deploy an external controller 
+    that updates the pod-deletion-cost, documentation should stress that update frequency
+    to be coarse grained.
 
 
 * **Will enabling / using this feature result in introducing new API types?**
@@ -225,21 +257,17 @@ previous answers based on experience in the field._
 
 * **Will enabling / using this feature result in increasing size or count
   of the existing API objects?**
-  Describe them providing:
-  - API type(s): Pod annotation
-  - Estimated increase in size: Size of a new annotation
-  - Estimated amount of new objects: new annotation for potentially every existing Pod
-
+  - No.
 
 * **Will enabling / using this feature result in increasing time taken by any
   operations covered by [existing SLIs/SLOs][]?**
-  - The time it takes to set/delete/change a pod annotation
+  - There are no SLOs covering scale down, but this feature should have negligible 
+    impact on scale-down latency since we are adding an additional sorting key.
 
 
 * **Will enabling / using this feature result in non-negligible increase of
   resource usage (CPU, RAM, disk, IO, ...) in any components?**
-  - The resources it takes to set/delete/change a pod annotation
-
+  - No.
 
 ### Troubleshooting
 
@@ -257,11 +285,13 @@ _This section must be completed when targeting beta graduation to a release._
 [existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
 
 ## Implementation History
-
-
-## Drawbacks
+- 2021-01-13: Initial KEP submitted as provisional
+- 2021-01-15: KEP promoted to implementable
 
 
 ## Alternatives
 
-Similar behaviour can be achieved through the Operator Framework which however will take a lot more configuration and setup work and is not a built-in Kubernetes feature.
+One alternative to using an annotation is adding an explicit API field. If the feature gets
+enough traction, we may consider promoting the annotation to a Status field.
+
+
--- a/keps/prod-readiness/sig-apps/2255.yaml
+++ b/keps/prod-readiness/sig-apps/2255.yaml
@@ -1,0 +1,3 @@
+kep-number: 2255
+alpha:
+  approver: "@wojtek-t"
--- a/keps/sig-apps/2255-pod-cost/kep.yaml
+++ b/keps/sig-apps/2255-pod-cost/kep.yaml
@@ -1,4 +1,4 @@
-title: Add pod-cost annotation for ReplicaSet
+title: ReplicaSet Pod Deletion Cost
 kep-number: 2255
 authors:
   - "@drbugfinder-work"
@@ -6,7 +6,7 @@ authors:
   - "@alculquicondor"
 owning-sig: sig-apps
 participating-sigs:
-status: provisional
+status: implementable
 creation-date: 2021-01-12
 reviewers:
   - "@ahg-g"
@@ -14,10 +14,12 @@ reviewers:
   - "@alculquicondor"
 approvers:
   - "@janetkuo"
+prr-approvers:
+  - "@wojtek-t"
 see-also:
   - https://github.com/kubernetes/kubernetes/issues/45509
-  - https://github.com/kubernetes/enhancements/issues/2255
-replaces:
+  - https://github.com/kubernetes/kubernetes/issues/4301
+
 
 # The target maturity stage in the current dev cycle for this KEP.
 stage: alpha
@@ -35,11 +37,10 @@ milestone:
 
 # The following PRR answers are required at alpha release
 # List the feature gate name and the components for which it must be enabled
-#feature-gates:
-#  - name: MyFeature
-#    components:
-#      - kube-apiserver
-#      - kube-controller-manager
+feature-gates:
+ - name: ReplicaSetPodDeletionCost
+   components:
+     - kube-controller-manager
 disable-supported: true
 
 # The following PRR answers are required at beta release


ahg-g:
assign @janetkuo @ingvagabund

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,129 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+to other pods belonging to the same ReplicaSet. 
 
-If all pods have the same priority, there is no difference in the normal pod delete decision behaviour. The same applies if the pod-cost annotation is not used at all.
+The pod-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be converted to
+the default value.
 
-The pod-cost annotation can be changed during operation, for example, if workload changes or a new master gets elected.
+When scaling down a ReplicaSet, controller-manager will prioritize deleting
+pods with higher pod-cost. Specifically, the rank passed to
```
ingvagabund, line 81: Test plan mentions:
```
Integration tests to validate that:
 * Replicas with lower pod-cost are deleted before replicas with higher pod-cost
```
whereas my understanding of `controller-manager will prioritize deleting pods with higher pod-cost` is higher pod-cost replicas are deleted before lower pod-cost replicas.

ahg-g: thanks for the catch, I updated the text in the proposal. higher-cost = lower chance of being deleted.

drbugfinder-work: To clarify this again: 
higher-cost (higher relative value) means lower priority and higher chance to delete
lower-cost (lower relative value) means higher priority, lower chance to delete

ahg-g: hmm, this is confusing, the way I was thinking about it is that pod cost is how costly it is to remove the pod. But you are suggesting to define it as the cost of the pod on the cluster.

ahg-g: the former definition is easier to implement by controllers: for example, cost could be the number of open stateful sessions.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 ### Non-Goals
 
-Guaranteed (in contrast to the recommendation stated in Goals) deletion of a selected replica.
+- Guarantees on pod deletion order
 
 ## Proposal
 
-The application can set the `controller.kubernetes.io/pod-cost` annotation to a pod through the Kubernetes API. When a downscale event happens, the pod with the lower priority value of the previously set annotation will be deleted first. If one pod of the Deployment has no priority annotation set, it will be treated as the lowest priority.
+Define a known annotation, namely `controller.kubernetes.io/pod-cost` that
```
ahg-g, line 72: @liggitt I am guessing the annotation should be `controller.alpha.kubernetes.io/pod-cost` in alpha.

keps/sig-apps/2255-pod-cost/README.md
drbugfinder-work: It is ambigous at the moment, so we have to define it.

In my PR for example I've implemented it like I described above: https://github.com/kubernetes/kubernetes/pull/87849
Of course we can also change this.

ahg-g: I think we need to choose the semantics that are easier for users to use.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
-If all pods have the same priority, there is no difference in the normal pod delete decision behaviour. The same applies if the pod-cost annotation is not used at all.
+The pod-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be converted to
+the default value.
 
-The pod-cost annotation can be changed during operation, for example, if workload changes or a new master gets elected.
+When scaling down a ReplicaSet, controller-manager will prioritize deleting
+pods with lower pod-cost. Specifically, the rank passed to 
+[ActivePodsWithRanks](https://github.com/kubernetes/kubernetes/blob/cac933934b1301665e6e51a81c66c483f4e16c49/pkg/controller/controller_utils.go#L784-L809)
+is pod-cost inverse then number of pods on node (the latter is the current
```
alculquicondor, line 83: We don't need to go into implementation details, but I find this sentence rather confusing.

There is only one rank number (step 4 in the code comment), so I don't know how to interpret the word "then"

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+applications can set to offer a hint on the cost of deleting a pod compared
+to other pods belonging to the same ReplicaSet. 
 
-If all pods have the same priority, there is no difference in the normal pod delete decision behaviour. The same applies if the pod-cost annotation is not used at all.
+The pod-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be converted to
+the default value.
 
-The pod-cost annotation can be changed during operation, for example, if workload changes or a new master gets elected.
+When scaling down a ReplicaSet, controller-manager will prioritize deleting
```
alculquicondor, line 80: Shouldn't this be the opposite?

That is, Pods without the annotation get pod-cost 0, thus lowest priority to be removed first. My mental interpretation is that a Pod that has a high cost should be deleted first.

Unless you are going with the interpretation "cost of losing the Pod". But then this confusion suggests that we need a better name for the annotation.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 #### Story 1
 
-In an application environment with stateful worker (user-)sessions, it is essential to keep the user sessions alive as good as possible. In case of a scale-down event, the application has to tell the scheduler, which delete decision would have the lowest impact on existing sessions.
+The different pods of an application could have different utilization levels. 
+On scale down, the application may prefer to remove the pods with lower utilization.
 
 #### Story 2
 
-An application consists of identical server processes, but one of the replicas will be the master, which should be kept as long as possible. All other replicas can be treated as cattle workload. Then the master can set the priority annotation with a high priority value as soon as it has finished its startup process. The other replicas can remain either without any priority set, or e.g. with all the same, lower priority. This ensures, that the master replica of this deployment will be protected in a downscale situation.
+On scale down, the application may want to remove pods running on the most expensive
```
alculquicondor, line 98: This is a completely different story from the previously written one, which is also valid. Why not add a new one?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
-To recommend which pod gets deleted next of a ReplicaSet. This should help to avoid major reworks in existing applications architecture:
-* [45509](https://github.com/kubernetes/kubernetes/issues/45509) - Scale down a deployment by removing specific pods
+- Allow applications to influence the order of deleting pods when scaling down a ReplicaSet
 
 
 ### Non-Goals
 
-Guaranteed (in contrast to the recommendation stated in Goals) deletion of a selected replica.
+- Guarantees on pod deletion order
```
alculquicondor, line 68: Another non-goal:

- A controller a sets and updates the annotation for running Pods.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 ### Risks and Mitigations
 
-On previous Kubernetes ReplicaSet controller versions that don't implement the pod-cost annotation feature, the same application might make false assumptions about the protection of a master instance or workers with open (user-)sessions on it. As the pod-cost annotation would be only a suggestion to the ReplicaSet controller, the application developer should, however, handle the case of a failed master instance or broken user sessions. The feature is just an improvement, not a guarantee, as there might happen timing issues between setting the annotation and the next controller scale-down event.
+One risk is that users perceive the feature as a guarantee to delete order. Documentation
+should stress the fact that this is best effort.
 
 ## Design Details
 
+See proposal.
```
alculquicondor, line 110: nit: only the first paragraph that is currently in "proposal" should be there. The rest is appropriate for design details.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,102 +48,130 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 ### Graduation Criteria
 
 #### Alpha -> Beta Graduation
 * Implemented feedback from alpha testers
-* Thorough E2E and unit testing in place
 
 #### Beta -> GA Graduation
-* Significant number of end-users are using the feature
 * We're confident that no further API changes will be needed to achieve the goals of the KEP
 * All known functional bugs have been fixed
```
alculquicondor, line 126: Is the feature worth a conformance test?

alculquicondor (commented):
It looks like the repo has lifted the PRR requirement, but since you already requested it, you are missing a file in `keps/prod-readiness/sig-apps`

keps/sig-apps/2255-pod-cost/README.md
alculquicondor: What about `pod-delete-cost` or `pod-downscale-cost`?

ahg-g: there is another comment discussing the same issue, lets discuss there in one place.

ahg-g: actually lets have the discussion here since this where we are defining the semantics.

ahg-g: Yes, my interpretation is the cost of deleting the pod. 

The default cost of 0 means by default the cost is in the middle of the range (like we did with priority).

I guess there is confusion about the name, so indeed it is better if we change it. How about `delete-cost`.

@ingvagabund @drbugfinder-work lets have the discussion here where the semantics of the annotation is defined.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: Is there a concrete example where this is valid? usually you don't even want to delete the master at all because that typically breaks the application, and so you want guarantees of deletion order which goes against the best effort nature of this feature.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: lets continue this discussion here: https://github.com/kubernetes/enhancements/pull/2272#discussion_r558481486

keps/sig-apps/2255-pod-cost/README.md
ahg-g: let me try to rephrase.

keps/sig-apps/2255-pod-cost/README.md
alculquicondor: You are right, the case described should have a separate deployment for the master, with higher pod priority.

keps/sig-apps/2255-pod-cost/README.md
alculquicondor: About the 0 default: if the controller that manages the annotation didn't have time to look at the new Pod, this is somewhat similar to the previous heuristic of "pod is just starting up", so safer to delete.

The opposite approach would be: the controller didn't have time to qualify this new Pod, so we better not remove it yet.

I prefer the first interpretation.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: ok, done, how about now?

alculquicondor: lgtm

keps/sig-apps/2255-pod-cost/README.md
ahg-g: The cost is relative to other pods, and so having the default value in the middle allows controllers to select what interpretation they want. They can use positive values for pod cost if they want the first interpretation (which is the typical scenario), or they can use negative values to indicate that pods processed by the controller should always be deleted before the ones that didn't (second interpretation).

`pod-delete-cost` sounds good to me, which aligns with the KEP title.

keps/sig-apps/2255-pod-cost/README.md
janetkuo: Only GA features are eligible for conformance tests. We can discuss that after this feature is GA.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,101 +48,133 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 * **Are there any tests for feature enablement/disablement?**
+No, but we can do manual testing.
 
+## Implementation History
+- 2021-01-13: Initial KEP submitted as provisional
+- 2021-01-15: KEP promoted to implementable
+
+
+<!--
```
janetkuo, line 177: There are still some mentions of "priority" in this doc. Please update those to pod cost.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,101 +48,133 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
-* Significant number of end-users are using the feature
 * We're confident that no further API changes will be needed to achieve the goals of the KEP
 * All known functional bugs have been fixed
 
 ### Upgrade / Downgrade Strategy
 
-When upgrading no changes are needed to maintain existing behaviour as all of this behaviour is fully optional and disabled by default. To activate this feature either a user has to make an annotation to a pod in a Deployment by hand or the application annotates a pod in a Deployment through the API.
-
-When downgrading, there is no need to changing anything, as this is just a pod annotation, which is uncritical.
+N/A
```
janetkuo, line 134: We should discuss what happens after cluster upgrades / downgrades, such as how the controller handles it on downgrade.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,101 +48,133 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+The different pods of an application could have different utilization levels. 
+On scale down, the application may prefer to remove the pods with lower utilization.
 
 #### Story 2
 
-An application consists of identical server processes, but one of the replicas will be the master, which should be kept as long as possible. All other replicas can be treated as cattle workload. Then the master can set the priority annotation with a high priority value as soon as it has finished its startup process. The other replicas can remain either without any priority set, or e.g. with all the same, lower priority. This ensures, that the master replica of this deployment will be protected in a downscale situation.
+On scale down, the application may want to remove pods running on the most expensive 
+nodes first. For example, remove pods from nodes running on standard VMs first
+then from ones running on preemptible/spot VMs (which can be 80% cheaper than standard VMs).
```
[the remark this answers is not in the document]
ahg-g: On scale up, sometimes stock-outs happen, and so the autoscaler may be forced to allocate standard VMs rather than preemptibles.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: There is no strategy here really, I added some text to that effect.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: ack.

keps/sig-apps/2255-pod-cost/README.md
drbugfinder-work: I think that we should not define any default value, as this should be a relative ranking which pod should be deleted first. The word "cost" is also a little bit confusing, as we all have noticed. I would rather describe it as "pod-delete-ranking" and keep the order 
smaller-value=higher-keep-priority=lower-delete-chance

greater-value=lower-keep-priority=higher-delete-chance 

I would say that the range from 0 to +MAXINT should be used for this (0 to -MININT is accepted as well).

keps/sig-apps/2255-pod-cost/README.md
drbugfinder-work: There are some legacy applications, where this behavior is needed (master gets elected and should be kept)

ahg-g: can you give an example?

keps/sig-apps/2255-pod-cost/README.md
ahg-g: > I think that we should not define any default value, as this should be a relative ranking which pod should be deleted first. 

We need to define the behavior for the case where some pods don't have the value set and some do in the same ReplicaSet. So defining a default value is necessary, otherwise we can't compare the pods that have the field set and ones that don't.

> The word "cost" is also a little bit confusing, as we all have noticed. I would rather describe it as "pod-delete-ranking" and keep the order

don't you think `pod-delete-cost` clarifies it? rank is similar to priority, a higher rank may give the impression that the pod is more important, and hence should be deleted last, so one can argue that it is just as confusing.

> smaller-value=higher-keep-priority=lower-delete-chance

I think this is tricker to use by applications. For example, in the utilization story, pods with *higher* utilization or more open sessions are preferred to be deleted last. Do you have an example where doing the opposite (smaller->lower chance) fits better?

keps/sig-apps/2255-pod-cost/README.md
drbugfinder-work: Some (legacy) databases

keps/sig-apps/2255-pod-cost/README.md
drbugfinder-work: > We need to define the behavior for the case where some pods don't have the value set and some do in the same ReplicaSet. So defining a default value is necessary, otherwise we can't compare the pods that have the field set and ones that don't.
We do not need a default. If the priority is low-value < high-value < not-annotated, always the not-annotated pod(s) get deleted. Of course one can remove an annotation from a prioritized pod and it will be in the group of pods to be deleted first (without any order/ranking).
Please see my demo https://www.youtube.com/watch?v=91APTsEsxIU

>  rank is similar to priority, a higher rank may give the impression that the pod is more important, and hence should be deleted last, so one can argue that it is just as confusing.
I was thinking of a "winners podium" when I read "rank", so for me 1st position would be the highest rank.

> I think this is tricker to use by applications. For example, in the utilization story, pods with higher utilization or more open sessions are preferred to be deleted last. Do you have an example where doing the opposite (smaller->lower chance) fits better?
I am with you at this point - if you see it as a "session counter" then it is a valid order. I'm fine with both options, we just need to define it, so there is no confusion about this anymore.

ingvagabund: I am inclined to `pod-delete-cost`. It clearly expresses the intention without specifying how the goal is achieved. A word `rank` does not say how the value is going to be perceived. Depending on context it can mean both higher/lower rank -> sooner to delete.

Although, `pod-deletion-cost` sounds a bit better to me. Just a nit. Not sure if there's a generic rule for how verbs in annotation keys are to be bend.

ingvagabund: Wrt. default value, we need one to assume if there's no annotation set. `0` will make a trouble if there are annotations with negative values. But again, it's up to a component annotating pods to take it into account. The component might just (re-)annotate any pod older than some duration with negative number so pods that are "young" enough and don't have an annotation to be given a chance to still run and thrive.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,210 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+should stress the fact that this is best effort.
 
 ## Design Details
 
+The pod-delete-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be converted to
+the default value.
+
+When scaling down a ReplicaSet, controller-manager will prioritize deleting
+pods with lower `pod-delete-cost`. Specifically, the pod-delete-cost will be evaluated after
```
ingvagabund, line 105: This is the only place where pod-delete-cost is wrapped with ``. Is the intention to use `pod-delete-cost` for a value? If not, can you drop the quotation marks?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,210 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+scale down behavior is not changed compared to now.
 
 ### Test Plan
 
 * Units test in kube-controller-manager package to test a variety of scenarios.
-* New E2E Tests to validate that replicas get deleted as expected e.g:
- * Replicas with lower pod-cost before replicas with higher pod-cost
- * Replicas with no pod-cost annotation set before replicas with low priority
+* Integration tests to validate that:
+ * Replicas with lower pod-delete-cost are deleted before replicas with higher pod-delete-cost
```
ingvagabund, line 121: More indentation of the tests? As printed in the review, `Integration tests ...`, `Replicas with lower ...` and `No behavior change ...` on the same level.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: ok, so a new replica is elected then if the master happens to be taken down, right?

keps/sig-apps/2255-pod-cost/README.md
alculquicondor: That's why I agree with Abdullah that having the default in the center (0) is the best approach. Users can assign the interpretation they want by adding positive or negative cost.

ahg-g: ok, unless I miss-understood the comments, I see that there is agreement to use `pod-delete-cost` and the semantics is the cost of deleting the pod (i.e., lower value means sooner to delete)

Regarding the default value, drbugfinder-work@ the way you are describing it is basically giving -MaxInt default value.

> 0 will make a trouble if there are annotations with negative values.

I think a default of 0 offers flexibility as I argued in https://github.com/kubernetes/enhancements/pull/2272#discussion_r558557145; this is similar to pod priority. It is too restrictive to fix it at either end of the range.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: I am not too convinced that in those situations the deployment is actively scaled up and down. You usually define a replica count for databases, and this stays constant unless one goes down involuntarily, like a zone going offline.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: thanks, dropped.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done, thanks.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,216 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+N/A
 
 ## Production Readiness Review Questionnaire
 
 ### Feature enablement and rollback
 
 * **How can this feature be enabled / disabled in a live cluster?**
-  - [x] Other
-    - Make special pod annotations within a live Deployment
+  - [x] Feature gate (also fill in values in `kep.yaml`)
```
ahg-g, line 155: @wojtek-t @liggitt I am wondering if a feature flag is necessary for this? how about if we just promote the annotation: `controller.alpha.kubernetes.io/pod-delete-cost` to `controller.beta.kubernetes.io/pod-delete-cost` to `controller.kubernetes.io/pod-delete-cost` in GA? conversely, if we use a feature flag, we can just use `controller.kubernetes.io/pod-delete-cost` from the get go since the flag will control the feature graduation, right?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -101,8 +101,14 @@ The pod-delete-cost range will be from [-MaxInt, MaxInt]. The default value is 0
 Invalid values (like setting the annotation to string) will be converted to
 the default value.
 
+Having the default value in the middle of the range allows controllers to cutomize
+the semantics of the cost of deleting pods that don't have the annotation set: 
+controllers can use positive pod-delete-cost values if they always want unintialized
```
ingvagabund, line 106: unintialized -> uninitialized

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -101,8 +101,14 @@ The pod-delete-cost range will be from [-MaxInt, MaxInt]. The default value is 0
...
+pods to be deleted first, or use negative pod-delete-cost values if they want
+unintialized pods to always be deleted last.
```
ingvagabund, line 108: unintialized -> uninitialized

ingvagabund (changes requested):

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done

keps/sig-apps/2255-pod-cost/README.md
liggitt: we've never had a good way to transition from an alpha annotation to a GA one... if an annotation is being used, I would recommend starting with the name you're going to want long-term.

keps/sig-apps/2255-pod-cost/README.md
liggitt: see comment above... a feature flag and a non-alpha annotation name seems like a better approach to me

ahg-g: sg, thanks.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: that is my thought as well, thanks.

keps/sig-apps/2255-pod-cost/README.md
alculquicondor: Aren't StatefulSets more appropriate for database use cases anyways?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,216 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+Yes.
 
 
 * **What happens if we reenable the feature if it was previously rolled back?**
-  - Then the feature will be reenabled. Nothing special to consider here.
+It should continue to work as expected.
 
 
 * **Are there any tests for feature enablement/disablement?**
+No, but we can do manual testing.
```
alculquicondor, line 180: why not unit tests?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,216 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+* **Will enabling / using this feature result in any new calls to cloud
+  provider?**
+  - No.
+
+
+* **Will enabling / using this feature result in increasing size or count
+  of the existing API objects?**
+  Describe them providing:
+  - API type(s): Pod annotation
+  - Estimated increase in size: Size of a new annotation
```
alculquicondor, line 203: You can point out that values are expected to be integers

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,191 +49,216 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 ### Dependencies
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **Does this feature depend on any specific services running in the cluster?**
-  - The feature requires the existing of the kube-controller-manager and the ability and permissions to set pod annotations.
-
-
-### Scalability
```
alculquicondor, line 203: Why not keep this? I think the status-quo is to leave the questions if they are not answered yet.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -256,12 +275,11 @@ _This section must be completed when targeting beta graduation to a release._
...
-## Implementation History
-
 
 ## Drawbacks
 
 
 ## Alternatives
 
-Similar behaviour can be achieved through the Operator Framework which however will take a lot more configuration and setup work and is not a built-in Kubernetes feature.
+-->
```
alculquicondor, line 284: An alternative we considered was adding an API field instead of the annotation.

Of course, leaving this to third-party implementations is also an option.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: this is more about upgrade/downgrade the control plane, the feature impacts more than one component.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: the key and value are strings, it is not a four byte int.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: not sure how leaving this to third-party implementations would work.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: sounds strange to have questions not answered, but I reverted the change.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: I added we will use unit tests, the upgrade->downgrade->upgrade is a different question actually.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+to other pods belonging to the same ReplicaSet. 
 
 ### User Stories (optional)
 
-
 #### Story 1
 
-In an application environment with stateful worker (user-)sessions, it is essential to keep the user sessions alive as good as possible. In case of a scale-down event, the application has to tell the scheduler, which delete decision would have the lowest impact on existing sessions.
+The different pods of an application could have different utilization levels. 
+On scale down, the application may prefer to remove the pods with lower utilization.
```
wojtek-t, line 89: I'm not convinced about utilization example. The reason is that utilization is often changing very frequently and  we don't want to have gazillions of pod updates frequently updating that annotation.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
+On scale down, the application may want to remove pods running on the most expensive 
+nodes first. For example, remove pods from nodes running on standard VMs first
+then from ones running on preemptible/spot VMs (which can be 80% cheaper than standard VMs).
 
 
 ### Risks and Mitigations
 
-On previous Kubernetes ReplicaSet controller versions that don't implement the pod-cost annotation feature, the same application might make false assumptions about the protection of a master instance or workers with open (user-)sessions on it. As the pod-cost annotation would be only a suggestion to the ReplicaSet controller, the application developer should, however, handle the case of a failed master instance or broken user sessions. The feature is just an improvement, not a guarantee, as there might happen timing issues between setting the annotation and the next controller scale-down event.
+One risk is that users perceive the feature as a guarantee to delete order. Documentation
+should stress the fact that this is best effort.
```
wojtek-t, line 101: The other risk that I want to have mentioned is performance-related, e.g. users want to update that too frequently.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 ## Design Details
 
+The pod-deletion-cost range will be from [-MaxInt, MaxInt]. The default value is 0.
+Invalid values (like setting the annotation to string) will be converted to
+the default value.
```
wojtek-t, line 107: Why not return "BadRequest" in that case? That seems more appropriate to me.

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 
 * **What happens if we reenable the feature if it was previously rolled back?**
-  - Then the feature will be reenabled. Nothing special to consider here.
+It should continue to work as expected.
 
 
 * **Are there any tests for feature enablement/disablement?**
-
+Unit tests.
```
wojtek-t, line 185: I guess there aren't any yet?

[Also - do you have tests that change the feature gate during the test? Or just test "how it works if feature" enabled and how it works if feature disabled?]

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can a rollout fail? Can it impact already running workloads?**
-  - As the feature is a simple annoation, the worst what could happen is that either the annotation is lost or ignored. In the worst case, a pod with a higher priority gets deleted before a pod with a lower priority.
-
+It shouldn't impact already running workloads. This is an opt-in feature
+since users need to explicitly set the annotation.
 
 * **What specific metrics should inform a rollback?**
-  - None
```
wojtek-t, line 163: Revert?

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
 
 
 * **Were upgrade and rollback tested? Was upgrade->downgrade->upgrade path tested?**
-  - Was tested. Behaviour change in both directions, as expected.
-
+We can do manual testing.
```
wojtek-t, line 199: s/can/will/ :)

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
- * Replicas with no pod-cost annotation set before replicas with low priority
+- Units test in kube-controller-manager package to test a variety of scenarios.
+- Integration tests to validate that:
+  - Replicas with lower pod-deletion-cost are deleted before replicas with higher pod-deletion-cost
+  - No behavior change when pod-deletion-cost is not set or all pods have the same pod-deletion-cost
 
 ### Graduation Criteria
 
 #### Alpha -> Beta Graduation
 * Implemented feedback from alpha testers
```
wojtek-t, line 138: I see that you removed "monitoring" section from the PRR. It's fine for Alpha, but please add a point that metrics are agreed on and implemented for Beta graduation criteria

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -55,163 +54,183 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
-_For beta, this section is required: reviewers must answer these questions._
-
-_For GA, this section is required: approvers should be able to confirms the
-previous answers based on experience in the field._
-
 * **Will enabling / using this feature result in any new API calls?**
-  - Whenever the application decides, that a change in pod-cost is needed for a replica, it will send out an API request and set the appropriate pod annotation(s).
+  - No, not the feature itself. However, users will want to deploy an external controller 
+    that updates the pod-deletion-cost, documentation should stress that update frequency
+    to be coarse grained.
```
wojtek-t, line 233: Add "no" for the 4th question (line 247).

keps/sig-apps/2255-pod-cost/README.md
```diff
@@ -233,13 +252,13 @@ previous answers based on experience in the field._
 
 * **Will enabling / using this feature result in increasing time taken by any
   operations covered by [existing SLIs/SLOs][]?**
-  - The time it takes to set/delete/change a pod annotation
+  - It should have negligible impact on scale-down latency since we are
```
wojtek-t, line 255: FWIW: Scale-down isn't even covered by any slos so far.

wojtek-t (commented):
Only minor comments from me.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: The idea is that an external controller would update the pod-deletion-cost right before it issues a scale down. So there are ways to implement this use case efficiently. I updated the story to clarify that.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: updated

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: I updated the monitoring section apart from the two questions related to SLI and SLO, I am not sure if there is a practical way of computing an SLI/SLO for this feature.

ahg-g (commented):
Thanks @wojtek-t 

I would like to propose having this feature in alpha but turned on by default. The reason is that this feature doesn't introduce a new field to the spec, and so there is no concern related to persisting the field on upgrades. Moreover, the feature has no impact unless the annotation is populated. Is there precedence to having an alpha feature turned on by default?

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: > To avoid frequently updating the pods, the application should update pod-deletion-cost
once before issuing a scale down.

But it's not always application that is issue'ing scale-down, right? It may be HPA for example...

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: I guess this was the answer to: https://github.com/kubernetes/enhancements/pull/2272/files#r560776024 - not this comment.

So this one is still open :)

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: I see no answers in monitoring section ...

wojtek-t (commented):
I don't know why I don't have "Resolve/Unresolve" buttons for comments...

wojtek-t:
> I would like to propose having this feature in alpha but turned on by default. The reason is that this feature doesn't introduce a new field to the spec, and so there is no concern related to persisting the field on upgrades. Moreover, the feature has no impact unless the annotation is populated. Is there precedence to having an alpha feature turned on by default?

This is interesting question. I think that at the very least it means meeting Beta-related quality criteria at least.
@deads2k @johnbelamaric - for thoughts from the PRR side.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: right, but that doesn't mean the story is not valid, there are cases where the application is in control, which is what this story is emphasizing.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: risks updated.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: oh, sorry, my bad!

do you mean add validation in the api-server on setting an annotation?

keps/sig-apps/2255-pod-cost/README.md
ahg-g: I was thinking how it works if the feature is enabled/disabled. we can do the former as well in a unit test since the feature flag will impact only controller-manager, but I honestly don't see much value if we have the former.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: I updated them to N/A since I don't think there is a practical way of computing SLO/SLI for such a feature.

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: OK - let's make it more explicit that it's the application itself that is in control of scale downs in this particular example.

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: Yes - I mean, e.g. if the annotation is set to string, reject the request with BadRequest status code.

wojtek-t:
>> I would like to propose having this feature in alpha but turned on by default. The reason is that this feature doesn't introduce a new field to the spec, and so there is no concern related to persisting the field on upgrades. Moreover, the feature has no impact unless the annotation is populated. Is there precedence to having an alpha feature turned on by default?

> This is interesting question. I think that at the very least it means meeting Beta-related quality criteria at least.
@deads2k @johnbelamaric - for thoughts from the PRR side.

Actually @ahg-g - I would suggest separating this discussion from this PR. I suggest we continue with this one as "ready for alpha" and once merged, let's open a separate PR updating "enable alpha by default" and let's have that discussion on the second PR. That will make it more clear what we're discussing too.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done. I was trying to avoid involving the api-server and have the feature implemented by kube-controller-manager only.

keps/sig-apps/2255-pod-cost/README.md
ahg-g: done, and also gave an example.

ahg-g:
> > > I would like to propose having this feature in alpha but turned on by default. The reason is that this feature doesn't introduce a new field to the spec, and so there is no concern related to persisting the field on upgrades. Moreover, the feature has no impact unless the annotation is populated. Is there precedence to having an alpha feature turned on by default?
> 
> > This is interesting question. I think that at the very least it means meeting Beta-related quality criteria at least.
> > @deads2k @johnbelamaric - for thoughts from the PRR side.
> 
> Actually @ahg-g - I would suggest separating this discussion from this PR. I suggest we continue with this one as "ready for alpha" and once merged, let's open a separate PR updating "enable alpha by default" and let's have that discussion on the second PR. That will make it more clear what we're discussing too.

Sounds good.

keps/sig-apps/2255-pod-cost/README.md
wojtek-t: Thanks!

wojtek-t (commented):
I'm fine with this - I will wait for sig-apps approval then then approve PRR.

ahg-g:
> I'm fine with this - I will wait for sig-apps approval then then approve PRR.

Thanks! @janetkuo this one ready. Please let me know if you have any further comments.

janetkuo (approved):

ahg-g:
/hold cancel

I squashed the commits.

PR #2286, opened 2021-01-17
Update volume health KEP

xing-yang:
Enhancement issue: https://github.com/kubernetes/enhancements/issues/1432

### DIFF
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -15,13 +15,15 @@
   - [External controller](#external-controller)
     - [CSI interface](#csi-interface)
     - [Node down event](#node-down-event)
-  - [External node agent](#external-node-agent)
+  - [Kubelet](#kubelet)
     - [CSI interface](#csi-interface-1)
   - [Alternatives](#alternatives)
     - [Alternative option 1](#alternative-option-1)
     - [Alternative option 2](#alternative-option-2)
     - [Alternative option 3](#alternative-option-3)
     - [Optional HTTP(RPC) service](#optional-httprpc-service)
+    - [External node agent](#external-node-agent)
+      - [CSI interface](#csi-interface-2)
 - [Graduation Criteria](#graduation-criteria)
   - [Alpha](#alpha)
   - [Alpha -&gt; Beta](#alpha---beta)
@@ -108,8 +110,8 @@ There could be conditions that cannot be reported by a CSI driver. One or more n
 
 The Kubernetes components that monitor the volumes and report events with volume health information include the following:
 
-* An external monitoring node agent on each kubernetes worker node.
-  * The node agent watches volume health of the PVCs on that node. If a PVC has an abnormal health condition, an event will to reported on the pod object that is using the PVC. If multiple pods are using the same PVC, events will be reported on multiple pods.
+* Extend Kubelet's existing volume monitoring capability to also monitor volume health on each kubernetes worker node.
+  * In addition to gathering existing volume stats, Kubelet also watches volume health of the PVCs on that node. If a PVC has an abnormal health condition, an event will to reported on the pod object that is using the PVC. If multiple pods are using the same PVC, events will be reported on multiple pods.
 * An external monitoring controller on the master node.
   * Monitoring controller reports events on the PVCs.
 
@@ -130,10 +132,12 @@ Two main parts are involved here in the architecture.
   - Trigger controller RPC to check the health condition of the CSI volumes.
   - The external controller sidecar will also watch for node failure events. This component can be enabled via a flag.
 
-- External Node Agent:
-  - The external node agent will be deployed as a sidecar together with the CSI node driver on every Kubernetes worker node.
-  - Trigger node RPC to check volume's mounting conditions.
+- Kubelet:
+  - Kubelet already collects volume stats from CSI node plugin by calling CSI function NodeGetVolumeStats.
+  - In addition to existing volume stats collected already, Kubelet will also check volume condition collected from the same CSI function and log events to Pods if volume condition is abnormal.
   - Note that currently we do not have CSI support for local storage. When the support is available, we will implement relavant CSI monitoring interfaces as well.
+
+The volume health monitoring by Kubelet will be controlled by a new feature gate called `VolumeHealth`.
 
 ## Implementation
 
@@ -447,14 +451,19 @@ The external monitoring controller reports events on the PVCs.
 * In the case that a node goes down, the controller will report an event for all PVCs on that node.
 * The external monitoring controller reports node down events on the PVCs. The node down monitoring component can be enabled via a flag.
 
-### External node agent
+### Kubelet
 
 #### CSI interface
-Call NodeGetVolumeStats() RPC to check the mounting and other health conditions. The frequency of the check should be tunable. A configure option will be available in the external node agent to adjust this value.
 
-The external node agent will go through all the pods on that node and find out all volumes used by those pods. It will watch all those volumes and call NodeGetVolumeStatus() to find out their health status.
+Kubelet already collects volume stats from CSI node plugin by calling CSI function NodeGetVolumeStats.
+https://github.com/kubernetes/kubernetes/blob/v1.21.0-alpha.2/pkg/volume/csi/csi_metrics.go#L71
 
-The external node agent reports events on the pod object. If multiple pods are using the same volume, report events on all pods.
+In addition to volume stats collected already, Kubelet will also check the mounting and other health conditions from NodeGetVolumeStats.
+
+If abnormal volume condition is detected from NodeGetVolumeStats, Kubelet will retrieve all the pods used by the particular volume and report events on the pod objects. If multiple pods are using the same volume, events will be reported on all pods. This can be done by adding logic in csi_client after the NodeGetVolumeStats call to send events to pods if volume condition is abnormal.
+https://github.com/kubernetes/kubernetes/blob/v1.21.0-alpha.2/pkg/volume/csi/csi_client.go#L608
+
+This new volume health monitoring by Kubelet will be gated by the `VolumeHealth` feature gate. If enabled, Kubelet will monitor volume health when calling NodeGetVolumeStats CSI function and report events on pods when abnormal volume condition is detected. If not enabled, Kubelet works the same as before and will not check volume health when calling NodeGetVolumeStats CSI function.
 
 ### Alternatives
 
@@ -610,11 +619,33 @@ This optional service is out of scope for this KEP but may be introduced in a fu
 
 This option is not in the main proposal because it is a push-based method while CSI uses pull-based method.
 
+#### External node agent
+
+In the initial proposal and first alpha implementation of this feature, there is an external node agent component.
+
+* An external monitoring node agent on each kubernetes worker node.
+  * The node agent watches volume health of the PVCs on that node. If a PVC has an abnormal health condition, an event will to reported on the pod object that is using the PVC. If multiple pods are using the same PVC, events will be reported on multiple pods.
+
+When moving this feature to beta, we decided to combine this with existing volume stats check in Kubelet as both are calling the same CSI function NodeGetVolumeStats.
+
+Here's the initial proposal:
+
+- The external node agent will be deployed as a sidecar together with the CSI node driver on every Kubernetes worker node.
+- The external node agent triggers node RPC to check volume's mounting conditions.
+- Note that currently we do not have CSI support for local storage. When the support is available, we will implement relavant CSI monitoring interfaces as well.
+
+##### CSI interface
+Call NodeGetVolumeStats() RPC to check the mounting and other health conditions. The frequency of the check should be tunable. A configure option will be available in the external node agent to adjust this value.
+
+The external node agent will go through all the pods on that node and find out all volumes used by those pods. It will watch all those volumes and call NodeGetVolumeStatus() to find out their health status.
+
+The external node agent reports events on the pod object. If multiple pods are using the same volume, report events on all pods.
+
 ## Graduation Criteria
 ### Alpha
 * Initial feature implementation, including:
   * External controller volume health monitoring.
-  * External node agent volume health monitoring.
+  * Kubelet volume health monitoring.
 * Implementation in the csi-mock driver.
 * Add basic unit tests.
 
@@ -627,14 +658,14 @@ This option is not in the main proposal because it is a push-based method while 
 
 ## Test Plan
 ### Unit tests
-* Unit tests for external controller and external node agent volume health monitoring. The following failure scenario will be simulated in the unit tests:
+* Unit tests for external controller and Kubelet volume health monitoring. The following failure scenario will be simulated in the unit tests:
   * VolumeNotFound
   * OutOfCapacity
   * VolumeUnmounted
   * NodeDown
 
 ### E2E tests
-* e2e tests for external controller and external node agent volume health monitoring. Hostpath CSI driver and GCE PD driver will be used for e2e tests. The following failure scenario will be tested in the e2e tests:
+* e2e tests for external controller and Kubelet volume health monitoring. Hostpath CSI driver and GCE PD driver will be used for e2e tests. The following failure scenario will be tested in the e2e tests:
   * VolumeNotFound
   * OutOfCapacity
   * VolumeUnmounted
@@ -649,35 +680,44 @@ _This section must be completed when targeting alpha to a release._
 * **How can this feature be enabled / disabled in a live cluster?**
   - [x] Other
     - Describe the mechanism:
-      This feature does not have a feature gate because it is out-of-tree.
-      It is enabled when the health monitoring controller and agent sidecars
-      are deployed with the CSI driver.
+      This feature has a feature gate called `VolumeHealth` for Kubelet.
+      It is enabled when the feature gate in turned on.
+      The health monitoring feature in external controller does not have a
+      feature gate because it is out of tree.
+      It is enabled when the health monitoring controller sidecar is deployed with the CSI driver.
     - Will enabling / disabling the feature require downtime of the control
       plane?
-      It only affects the health monitoring controller and agent sidecars.
+      Enabling the `VolumeHealth` feature gate will require downtime of Kubelet.
+      From the controller side, it only affects the health monitoring controller sidecar.
     - Will enabling / disabling the feature require downtime or reprovisioning
       of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
-      No.
+      Not for the controller side volume monitoring.
+      For Kubelet, enabling/disabling the feature requires downtime of a node.
 
 * **Does enabling the feature change any default behavior?**
-  Enabling the feature will allow events to be reported on PVCs or Pods when
+  Enabling the `VolumeHealth` feature gate will allow Kubelet to monitor volume health and
+  generate events on Pods so it will change the default behavior.
+  Enabling the feature from the controller side will allow events to be reported on PVCs when
   abnormal volume conditions are detected.
 
 * **Can the feature be disabled once it has been enabled (i.e. can we rollback
   the enablement)?**
-  Yes. Uninstalling the health monitoring controller and agent sidecars will
-  disable the feature. Existing events will not be removed but they will
-  disappear after a period of time. Disabling the feature should not break
-  an existing application as these events are for humans only, not for automation. 
+  Yes. Uninstalling the health monitoring controller sidecar will disable the feature from
+  the controller side.
+  Disabling the feature gate on Kubelet will prevent Kubelet from monitoring volume health.
+  Existing events will not be removed but they will disappear after a period of time.
+  Disabling the feature should not break an existing application as these events are for humans
+  only, not for automation.
 
 * **What happens if we reenable the feature if it was previously rolled back?**
   Events will be added to PVCs or Pods when abnormal volume conditions are
   detected again.
 
 * **Are there any tests for feature enablement/disablement?**
-  Since there is no feature gate for this feature and the only way to enable or
-  disable this feature is to install or unistall the sidecars, we cannot write
-  tests for feature enablement/disablement.
+  There will be unit tests for the feature `VolumeHealth` enablement/disablement.
+  Since there is no feature gate for this feature on the controller side and the only way to
+  enable or disable this feature is to install or unistall the sidecar, we cannot write
+  tests for feature enablement/disablement on the controller side.
 
 ### Rollout, Upgrade and Rollback Planning
 
@@ -686,16 +726,25 @@ _This section must be completed when targeting beta graduation to a release._
 * **How can a rollout fail? Can it impact already running workloads?**
   Try to be as paranoid as possible - e.g., what if some components will restart
    mid-rollout?
-   This feature does not have a feature gate. It is enabled when the health
-   monitoring controller and agent sidecars are deployed with the CSI driver.
+   This feature does not have a feature gate on the controller side. It is enabled when
+   the health monitoring controller sidecar is deployed with the CSI driver.
    So the only way for a rollout to fail is that deploying the health
-   monitoring controller or agent sidecars with the CSI driver fails. If
+   monitoring controller sidecar with the CSI driver fails. If
    the health monitoring controller cannot be deployed, no events on volume
-   condition will be reported on PVCs. If the health monitoring agent cannot
-   be deployed, no event on volume condition will be reported on the pod.
+   condition will be reported on PVCs.
+
+   If enabling the `VolumeHealth` feature fails, no event on volume condition will be
+   reported on the pod.
 
 * **What specific metrics should inform a rollback?**
-  Currently an event will be recorded on pvc/pod when the controller/agent has successfully retrieved an abnormal volume condition from the storage system. However when other errors occur in the controller/agent, the errors will be logged but not recorded as events. Before moving to beta, the controller/agent should be modified to record an event when other errors occur.
+  An event will be recorded on the PVC when the controller has successfully retrieved an
+  abnormal volume condition from the storage system. When other errors occur in the controller,
+  the errors will also be recorded as events.
+
+  In Kubelet, an event will be recorded on the Pod when Kubelet has successfully retrieved an
+  abnormal volume condition. If the call to `NodeGetVolumeStats` fails for other reasons,
+  an error will be returned and whether this will be logged as an event on the Node is up to
+  the existing Kubelet logic and will not be changed.
 
 * **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
   Describe manual testing that was done and the outcomes.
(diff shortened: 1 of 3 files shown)


keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+   mid-rollout?
+   This feature does not have a feature gate. It is enabled when the health
+   monitoring controller and agent sidecars are deployed with the CSI driver.
+   So the only way for a rollout to fail is that deploying the health
+   monitoring controller or agent sidecars with the CSI driver fails. If
+   the health monitoring controller cannot be deployed, no events on volume
+   condition will be reported on PVCs. If the health monitoring agent cannot
+   be deployed, no event on volume condition will be reported on the pod.
+
+* **What specific metrics should inform a rollback?**
```
deads2k, line 697: I think this is important.  How will I know if I an updated health monitoring controller or agent sidecar is malfunctioning?  do they crash/produce events/show metrics that fail?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+  checking if there are objects with field X set) may be a last resort. Avoid
+  logs or events for this purpose.
+  An operator can check the metrics `csi_sidecar_operations_seconds`,
```
deads2k, line 718: is this divided by the type of operation?  This KEP appears to be adding APIs and knowing which APIs are being used seems to have value.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
+  - 99% percentile over day of absolute value from (job creation time minus expected
+    job creation time) for cron job <= 10%
+  - 99,9% of /health requests per day finish with 200 code
```
deads2k, line 739: health requests on which process?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+  as node-level agents (e.g. specific version of CRI). Focus on external or
+  optional services that are needed. For example, if this feature depends on
+  a cloud provider API, or upon an external software-defined storage or network
+  control plane.
+
+  For each of these, fill in the following—thinking about running existing user workloads
+  and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]: installation of csi-external-health-monitor-controller and csi-external-health-monitor-agent sidecars
+    - Usage description:
+      - Impact of its outage on the feature: Installation of csi-external-health-monitor-controller and csi-external-health-monitor-agent sidecars are required for the feature to work. If csi-external-health-monitor-controller is not installed, abnormal volume conditions will not be reported as events on PVCs. Similarly, if csi-external-health-monitor-agent is not installed, abnormal volume conditions will not be reported as events on Pods.
```
deads2k, line 761: can you clarify what provides the new API described here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1432-volume-health-monitor/README.md#csi-changes . It seems as though that component needs to be present and up to date to function.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
+  Describe them, providing:
+  - API call type (e.g. PATCH pods): Only events will be reported to PVCs or Pods if this feature is enabled.
+  - estimated throughput
+  - originating component(s) (e.g. Kubelet, Feature-X-controller)
+  focusing mostly on:
+  - components listing and/or watching resources they didn't before
+    csi-external-health-monitor-controller and csi-external-health-monitor-agent sidecars
```
deads2k, line 782: what is the rate limit you chose for events being produced?  Do you have the "similar event collapsing" feature of the event broadcaster turned on?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+  No.
+
+* **Will enabling / using this feature result in increasing size or count of
+the existing API objects?**
+  Describe them, providing:
+  - API type(s): No
+  - Estimated increase in size: (e.g., new annotation of size 32B):
+    No
+  - Estimated amount of new objects: (e.g., new Object X for every existing Pod)
+    No
```
deads2k, line 805: aren't you producing events for each PVC and on each pod mounting that PVC?  Do you produce any events in the good case? How many do you produce in the "bad" case?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,180 @@ _This section must be completed when targeting alpha to a release._
...
+### Troubleshooting
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How does this feature react if the API server and/or etcd is unavailable?**
+  If API server and/or etcd is unavailable, it will not be able to report events on PVCs or Pods.
```
deads2k, line 831: in such a case, does the controller backoff?

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Currently we only report an event on pvc/pod when we have successfully retrieved an abnormal volume condition from the storage system, but when other errors occur in the controller we only log an error but we don't report an event.  We probably should also report an event when other errors occur.  I'll update the KEP with this.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: This is per sidecar.  For the health monitoring controller sidecar, `csi_sidecar_operations_seconds` will be measuring ListVolumes or GetVolume RPC. For the health monitoring agent sidecar, `csi_sidecar_operations_seconds` will be measuring NodeGetVolumeStats RPC.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Sorry, that line was in the original template.

The metrics `csi_sidecar_operations_seconds` includes a gRPC status code.  We can look at the code to figure out the success/failure ratio.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Those are not Kubernetes API changes.  They are in the CSI spec v1.3.   https://github.com/kubernetes-csi/external-health-monitor/blob/master/go.mod#L6

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: We have a monitor interval to control how often to check the volume health.  It if configurable with 1 minute as default.
I'm not aware there is a "similar event collapsing" feature on the event broadcaster.  Will take a look.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: The controller reports events on PVC while the agent reports events on Pod.  They work independently of each other.  It is recommended that CSI driver should not report duplicate information thru the controller and the agent.  For example, if the controller detects a failure on one volume, it should record just one event on one PVC.  If an agent detects a failure, it should record an event on every pod used by the affected PVC.

Recovery event will be reported once if volume condition changes back to normal.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: It should not backoff.  It will log error messages.

xing-yang:
@deads2k Comments are addressed.  PTAL.  Thanks.

keps/sig-storage/1432-volume-health-monitor/README.md
deads2k: > This is per sidecar. For the health monitoring controller sidecar, `csi_sidecar_operations_seconds` will be measuring ListVolumes or GetVolume RPC. For the health monitoring agent sidecar, `csi_sidecar_operations_seconds` will be measuring NodeGetVolumeStats RPC.

Can you indicate that its necessary to slice the metric by process after you aggregate?  It would have helped me.

keps/sig-storage/1432-volume-health-monitor/README.md
deads2k: > Those are not Kubernetes API changes. They are in the CSI spec v1.3. https://github.com/kubernetes-csi/external-health-monitor/blob/master/go.mod#L6

So whatever provides that API surface needs to support or the csi-external-health-monitor-agent won't be able to query it, right?  I'm imagining a skew case of old node and kubelet with new controller.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,193 @@ _This section must be completed when targeting alpha to a release._
...
+* **Will enabling / using this feature result in any new API calls?**
+  Describe them, providing:
+  - API call type (e.g. PATCH pods): Only events will be reported to PVCs or Pods if this feature is enabled.
+  - estimated throughput
+  - originating component(s) (e.g. Kubelet, Feature-X-controller)
+  focusing mostly on:
+  - components listing and/or watching resources they didn't before
+    csi-external-health-monitor-controller and csi-external-health-monitor-agent sidecars.
+    There is a monitor interval for the controller and one for the agent to control how often
+    to check the volume health. It is configurable with 1 minute as default.
```
deads2k, line 793: > We have a monitor interval to control how often to check the volume health. It is configurable with 1 minute as default.
> I'm not aware there is a "similar event collapsing" feature on the event broadcaster. Will take a look.

We found that when scaled out across many nodes, low frequency checks can still produce high volumes of events.  To control this, we created the eventrecorder and have options on it to control QPS per key.  Can you have a look and be sure that you're collapsing keys and have a slow update cadence per key?  I think kubelets have a modest burst and then do something like one event every five minutes per key to avoid overloading kube-apiserver.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: CSI driver needs to be updated to implement volume health RPCs in controller/node plugins.  The minimum kubernetes version should be 1.13: https://kubernetes-csi.github.io/docs/introduction.html#kubernetes-releases. K8s v1.13 is the minimum supported version for CSI driver for work (of course, different CSI drivers have different requirements on supported k8s versions as well).  Let's say the CSI node plugin on one node has been upgraded to support volume health while it is not upgraded on 3 other nodes, then we will only expect to see volume health events on pods running on that one upgraded node.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Sure.  Added to the KEP that we should use this "similar event collapsing" feature and also consider changing the default from 1 to 5 minutes.

deads2k:
/lgtm

/hold
holding in case you'd like someone specific from sig-storage to have a look.


For the eventrecording bits, try to track backwards from here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/record/events_cache.go#L38-L49 I no longer remember how everything gets wired, but I think we use those in kubelet.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+
+_For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them._
+
+_For beta, this section is required: reviewers must answer these questions._
+
+_For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
```
msau42, line 787: If I remember correctly, there was a part of the design where we need have informers on Pod objects. Since this is a controller that runs per-node, I think having a Pod informer is going to create significant load on the API server, and we should figure out how to address that.

msau42 (commented):
I remember sig-node and sig-instrumentation were proposed some new architecture for monitoring node components. Are we aligned with those efforts?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+  Longer term, we may want to require automated upgrade/rollback tests, but we
+  are missing a bunch of machinery and tooling and can't do that now.
+  Manual testing will be done.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs,
+fields of API types, flags, etc.?**
+  Even if applying deprecation policies, they may still surprise some users.
+  No.
+
+### Monitoring Requirements
```
ehashman, line 711: @kubernetes/sig-instrumentation-api-reviews

annajung:
Hey @xing-yang , 1.21 Enhancements Lead here

One of the requirements for PRR is to create a `prod-readiness/<sig>/<KEP number>.yaml file`, with the PRR approver's GitHub handle for the specific stage. Please see [Submitting a KEP for production readiness approval](https://github.com/kubernetes/community/blob/master/sig-architecture/production-readiness.md#submitting-a-kep-for-production-readiness-approval) for details. 

Just wanted to mention it in case you want to include it in this PR as well :)

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
+   This feature does not have a feature gate. It is enabled when the health
+   monitoring controller and agent sidecars are deployed with the CSI driver.
+   So the only way for a rollout to fail is that deploying the health
+   monitoring controller or agent sidecars with the CSI driver fails. If
```
ehashman, line 692: Because these are deployed as sidecars, if the sidecars fail, that would also result in a restart of the rest of the CSI driver, which certainly could impact running workloads, right? How are we mitigating that risk---is it up to cluster operators to add and remove these sidecars from their cluster manifests?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+  No.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+  checking if there are objects with field X set) may be a last resort. Avoid
+  logs or events for this purpose.
```
ehashman, line 718: I think the intent of this question is an on/off metric as opposed to checking for the presence of metrics provided by this feature, but PRR reviewers can hopefully clarify.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+  An operator can check the metric `csi_sidecar_operations_seconds`,
+  Container Storage Interface operation duration with gRPC error code status
+  total. It is reported from CSI external-health-monitor-controller and
+  external-health-monitor-agent sidecars. For the health monitor controller
+  sidecar, `csi_sidecar_operations_seconds` will be measuring `ListVolumes` or
+  `GetVolume` RPC. For the health monitor agent sidecar,
```
ehashman, line 724: Measuring what about `ListVolumes` and `GetVolume` -- the duration of RPC calls?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+  - [ ] Other (treat as last resort)
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
+  - 99% percentile over day of absolute value from (job creation time minus expected
+    job creation time) for cron job <= 10%
```
ehashman, line 744: I don't understand what this SLO means, can you clarify?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+    - [Optional] Aggregation method:
+    - Components exposing the metric: csi-external-health-monitor-controller and csi-external-health-monitor-agent sidecars
+  - [ ] Other (treat as last resort)
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
```
ehashman, line 742: Do we have benchmarks on volume for the API calls? 1% of 1000 calls per day 10 errors, 1% of 100,000 calls per day is 1000 5XX errors which is quite a lot. Generally the presence of any number of 5XX errors in the KAS audit logs prompt an investigation IME.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+
+  The metrics `csi_sidecar_operations_seconds` includes a gRPC status code. If the
+  status code is `OK`, the call is successful; otherwise, it is not successful. We
+  can look at the ratio of successful vs non-successful statue codes to figure out
+  the success/failure ratio.
+
+* **Are there any missing metrics that would be useful to have to improve observability
+of this feature?**
+  Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+  implementation difficulties, etc.).
```
ehashman, line 755: Are we producing operation counts in addition to durations? (I'm guessing yes if `csi_sidecar_operations_seconds` is a histogram but we might want to note the existence such metrics in here explicitly.)

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+* **Will enabling / using this feature result in increasing size or count of
+the existing API objects?**
+  Describe them, providing:
+  - API type(s): No
+  - Estimated increase in size: (e.g., new annotation of size 32B):
+    No
+  - Estimated amount of new objects: (e.g., new Object X for every existing Pod)
+    The controller reports events on PVC while the agent reports events on Pod. They work independently of each other. It is recommended that CSI driver should not report duplicate information through the controller and the agent. For example, if the controller detects a failure on one volume, it should record just one event on one PVC. If an agent detects a failure, it should record an event on every pod used by the affected PVC.
+
+    Recovery event will be reported once if the volume condition changes from abnormal back to normal.
```
ehashman, line 825: Can you please include a numeric estimate/bounds on the events?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+  This feature will periodically query storage systems to get the latest volume conditions. So this will have an impact on the performance of the operations running on the storage systems.
+
+* **Will enabling / using this feature result in non-negligible increase of
+resource usage (CPU, RAM, disk, IO, ...) in any components?**
+  Things to keep in mind include: additional in-memory state, additional
+  non-trivial computations, excessive access to disks (including increased log
+  volume), significant amount of data sent and/or received over network, etc.
+  This through this both in small and large cases, again with respect to the
+  [supported limits].
+  This will increase load on the storage systems as it periodically queries them.
```
ehashman, line 840: Given this, it would be good to include an escape valve to turn this off at runtime. Getting stuck in I/O sleep can take out an entire node.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -674,8 +679,200 @@ _This section must be completed when targeting alpha to a release._
...
+### Troubleshooting
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How does this feature react if the API server and/or etcd is unavailable?**
+  If API server and/or etcd is unavailable, error messages will be logged and the controller/agent will not be able to report events on PVCs or Pods.
```
ehashman, line 851: Is there a Prometheus dependency on service discovery from the API server for scraping the metrics?

xing-yang:
@annajung added keps/prod-readiness/sig-storage/1432.yaml.

xing-yang:
Currently the volume health monitor agent retrieves volume condition by calling CSI function NodeGetVolumeStats.
https://github.com/kubernetes-csi/external-health-monitor/blob/v0.2.0/pkg/csi-handler/pv_handler.go#L107
res, err := handler.nodeClient.NodeGetVolumeStats(ctx, &req)

Kubelet already collects volume stats from CSI node plugin also by calling CSI function NodeGetVolumeStats.
https://github.com/kubernetes/kubernetes/blob/v1.21.0-alpha.2/pkg/volume/csi/csi_metrics.go#L71
metrics, err := csiClient.NodeGetVolumeStats(ctx, mc.volumeID, mc.targetPath)

I wonder if we should consolidate these two and have Kubelet handle volume health check as well.  This way we don’t need a separate volume health monitor agent.

@msau42 @ehashman thoughts?

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: I think we can use the FieldSelector to limit the scope of the informer to watch the pods only bound to one node.

On another thought, if we have kubelet handle volume health check as part of the NodeGetVolumeStats call, we don't need a separate volume health monitor agent.
https://github.com/kubernetes/kubernetes/blob/v1.21.0-alpha.2/pkg/volume/csi/csi_metrics.go#L71

msau42:
> I wonder if we should consolidate these two and have Kubelet handle volume health check as well. This way we don’t need a separate volume health monitor agent.

That sounds reasonable to me, especially since kubelet is already calling it. The main challenge will be how to plumb it through since this is specifically called in a metrics collection path, and our use case (events) is a little different.

ehashman:
There is also work being done to try to pare down the metrics being exposed by the kubelet... cc @dashpole @bobbypage @haircommander if they want to weigh in here

xing-yang:
> The main challenge will be how to plumb it through since this is specifically called in a metrics collection path, and our use case (events) is a little different.

In Node Problem Detector, there are multiple types of exporters.  Other than the Prometheus exporter, there is also a Kubernetes exporter.  `Kubernetes exporter reports node problems to Kubernetes API server: temporary problems get reported as Events, and permanent problems get reported as Node Conditions.`  Maybe we can do something similar here?

https://github.com/kubernetes/node-problem-detector#exporter

xing-yang:
CC @NickrenREN

xing-yang:
Updated the KEP to move the node side volume health monitoring to Kubelet.

@msau42 @ehashman please take a look.

xing-yang:
> I think leveraging existing calls that kubelet is making seems reasonable to me. I think the main challenge will be how to plumb this to pod events from the current code which is in the metrics path.
> cc @ehashman @dashpole to chime in. Note that this feature is not exposing new metrics. It's using the volume metrics call we already make today to also gather health status, and report health status as Pod events.

I'm thinking we can add logic in csi_client after the NodeGetVolumeStats call to send events to pods if volume condition is abnormal.  WDYT?  @msau42 @ehashman @dashpole 
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_client.go#L608

ehashman:
/assign @dashpole 
for instrumentation review/approval

msau42:
> I'm thinking we can add logic in csi_client after the NodeGetVolumeStats call to send events to pods if volume condition is abnormal. 

Currently it's called in the metrics collection path. I'm unsure about the frequency of the path. Is it invoked periodically with some configurable interval, or is it only called if someone queries the metric endpoint?

dashpole:
It looks like volume stats are collected on an interval and cached: https://github.com/kubernetes/kubernetes/blob/28e2e12b887fe082929d3ece4b3cbd572dc15d39/pkg/kubelet/server/stats/volume_stat_calculator.go#L81

The summary API just queries the cached metrics.

Adding events is fine so long as the event message doesn't change between similar events so they can be deduplicated.

In principle, it would be better to have separate calls for getting stats and doing health checks, but i'm assuming the CSI interface this depends on is already fixed.

The only potential problem would be if you weren't able to get a pod reference to associate the event with.  But that is something that can be handled in code review.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -644,38 +680,260 @@ _This section must be completed when targeting alpha to a release._
...
+  In Kubelet, an operator can check whether the feature gate `VolumeHealth`
+  is enabled.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine
+the health of the service?**
+  - [ ] Metrics
+    - Metric name: csi_sidecar_operations_seconds
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+      csi-external-health-monitor-controller exposes the `csi_sidecar_operations_seconds` metric.
```
dashpole, line 786: it looks like the metric names need to be updated to not be sidecar related

xing-yang: I removed the reference for the health monitor agent.  This one is from the controller side so we are keeping it as a CSI sidecar.  It won't have impact on the node.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -644,38 +680,260 @@ _This section must be completed when targeting alpha to a release._
...
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
+   This feature does not have a feature gate on the controller side. It is enabled when
+   the health monitoring controller sidecar is deployed with the CSI driver.
+   So the only way for a rollout to fail is that deploying the health
+   monitoring controller sidecar with the CSI driver fails. If
+   the health monitoring controller cannot be deployed, no events on volume
+   condition will be reported on PVCs.
+
+   If enabling the `VolumeHealth` feature fails, no event on volume condition will be
```
msau42, line 736: Could kubelet crash if there's a bug in how we process the CSI call?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -644,38 +680,260 @@ _This section must be completed when targeting alpha to a release._
...
+   reported on the pod.
+
+* **What specific metrics should inform a rollback?**
+  An event will be recorded on the PVC when the controller has successfully retrieved an
+  abnormal volume condition from the storage system. When other errors occur in the controller,
+  the errors will also be recorded as events.
+
+  In Kubelet, an event will be recorded on the Pod when Kubelet has successfully retrieved an
```
msau42, line 744: We don't have kubelet metrics on the csi call itself. This is something that @Jiawei0227 was considering adding in general.

keps/sig-storage/1432-volume-health-monitor/README.md
msau42: Yes the histogram includes duration, counts and return code.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -644,38 +680,260 @@ _This section must be completed when targeting alpha to a release._
...
+
+* **What steps should be taken if SLOs are not being met to determine the problem?**
+  If SLOs are not being met, analysis should be made to understand what have caused the problem. Debug level logging should be enabled to collect verbose logs. Look at logs to find out what might have caused the events to be missed. If it indicates an underlying problem on the storage system, then storage admin can be pulled in to help find the root cause.
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
 
 ## Implementation History
 
+- 20210117: Update KEP for Beta
```
msau42, line 935: can you update this since we're going to do an alpha2?

msau42:
/lgtm
/approve
from sig-storage

/hold
for prr review.  @deads2k note that we decided to keep this in alpha and do a redesign.

xing-yang:
CC @dchen1107

PR #2302, opened 2021-01-21
Copyright 2020 updated with Copyright 2021

shekhar-rajak:
All the files updated with 2021


mrbobbytables:
@shekhar-rajak Updating the copyright isn't needed. They only have to go in when the file is first created. We'd have to update thousands upon thousands of files each year otherwise :x

/close

PR #2308, opened 2021-01-21
Add KEP for job tracking in implementable status

alculquicondor:
Ref #2307 kubernetes/kubernetes#8486

/sig apps

### DIFF
--- a/keps/prod-readiness/sig-apps/2307.yaml
+++ b/keps/prod-readiness/sig-apps/2307.yaml
@@ -1,0 +1,3 @@
+kep-number: 2307
+alpha:
+  approver: "@wojtek-t"
--- a/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
+++ b/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
@@ -1,0 +1,585 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-2307: Job tracking without lingering Pods
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [New API calls](#new-api-calls)
+    - [Bigger Job status](#bigger-job-status)
+    - [Unprotected Job status endpoint](#unprotected-job-status-endpoint)
+- [Design Details](#design-details)
+  - [API changes](#api-changes)
+  - [Algorithm](#algorithm)
+  - [Deleted Pods](#deleted-pods)
+  - [Pod adoption](#pod-adoption)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The current Job controller currently relies on completed Pods to not be removed
+in order to track the Job completion status. This proposal presents an
+alternative implementation for Job tracking that does not have this dependency.
+
+## Motivation
+
+The current approach of relying on the Pods existence is problematic for Jobs
+that require a big number of completions or for clusters with too many Jobs
+running at the same time. The finished Pods cannot be removed until the entire
+Job completes, even if the Pods failed.
+
+Furthermore, once the number of finished Pods reaches a threshold, the Pod
+garbage collection controller starts removing Pods. Today, the Job controller
+relies on the garbage collector having a big threshold.
+
+### Goals
+
+- Perform Job completion and failures tracking without relying on lingering
+  Pods.
+
+### Non-Goals
+
+- Remove Pods once they have been accounted for.
+
+## Proposal
+
+The Job controller creates Pods with a finalizer to prevent finished Pods from
+being removed by the garbage collector. The Job controller removes the finalizer
+from the finished Pods once it has accounted for them. In subsequent Job syncs,
+the controller ignores finished Pods that don't have the finalizer.
+
+### Notes/Constraints/Caveats (Optional)
+
+Due to the lack of support of atomic changes across kubernetes objects, an
+intermediate state is necessary. Before removing the Pod finalizers,
+the controller adds finished Pods to a list in the Job status. After removing
+the Pod finalizers, the controller clears the list and updates the counters.
+
+### Risks and Mitigations
+
+#### New API calls
+
+The new algorithm introduces new API calls:
+- one per Pod lifecycle to remove finalizers
+- one for each Job sync, to track the intermediate state.
+
+Note that there no new calls in the Pod creation path.
+
+On the other hand, to update the Job status once a Pod finishes, we need a total
+2 new API calls compared to the legacy algorithm. If more than one Pod finish at
+a given time, we add `n + 1` API calls, where `n` is the number finished Pods.
+
+Consider a Job with multiple Pods. With a 50 QPS limit in the job controller
+client, the controller should be able to process between 2000 to 3000 Pods.
+
+The increase in API calls is justified for the following reasons:
+- The legacy tracking cannot handle a big number of terminated Pods, across
+  any number of Jobs, at a given point.
+- In the entirety of its lifecycle, a Pod requires at least 8 API calls,
+  including status updates and events.
+  
+However, in order to prevent Jobs with big number of Pods from starving Jobs
+with fewer Pods, the Job controller might skip status updates until enough
+Pods have accumulated or enough time has passed. See [Algorithm](#algorithm)
+below for more details.
+   
+#### Bigger Job status
+
+Job status can temporarily grow if too many Pods finish at the same time or
+if the Job controller is down for some time. In this case, we do partial
+status updates. See [Algorithm](#algorithm) below.
+
+#### Unprotected Job status endpoint
+
+Changes in the status not produced by the Job controller in
+kube-controller-manager could affect the Job tracking. Cluster administrators
+should make sure to protect the Job status endpoint via RBAC.
+
+## Design Details
+
+### API changes
+
+The Job status gets a new struct to hold the uncounted Pods before they are
+added to the counters.
+
+```golang
+type JobStatus struct {
+    Succeeded int32
+    Failed    int32
+    ...
+
+    // UncountedTerminatedPods holds UIDs of Pods that have finished but
+    // haven't been accounted in the counters.
+    // If nil, Job tracking doesn't make use of this structure.
+    // +optional
+    UncountedTerminatedPods *UncountedTerminatedPods
+}
+
+// UncountedTerminatedPods holds UIDs of Pods that have finished but haven't
+// been accounted in Job status counters.
+type UncountedTerminatedPods struct {
+    // Succeeded holds UIDs of succeeded Pods.
+    Succeeded []types.UID
+    // Succeeded holds UIDs of failed Pods.
+    Failed    []types.UID
+}
+```
+
+Note: the final name of the field `uncountedTerminatedPods` will be decided
+during API review.
+
+### Algorithm
+
+The following algorithm updates the status counters without relying on finished
+Pods to be present indefinitely. The algorithm assumes that the Job controller
+could be stopped at any point and executed again from the first step without
+losing information. Generally, all the steps happen in a single Job sync
+cycle.
+
+1. The Job controller calculates the number of succeeded Pods as the sum of:
+   - `.status.succeeded`,
+   - the size of `job.status.uncountedTerminatedPods.succeeded` and
+   - the number of finished Pods that are not in `job.status.uncountedTerminatedPods.succeeded`.
+   This number informs the creation of missing Pods to reach `.spec.completions`.
+   The controller creates Pods for a Job with the finalizer
+   `batch.kubernetes.io/job-completion`.
+2. The Job controller adds Pod UIDs to the `.status.uncountedTerminatedPods.succeeded`
+   and `.status.uncountedTerminatedPods.failed` lists if the Pod:
+    - has the `batch.kubernetes.io/job-completion` finalizer, and
+    - the Pod is on Succeeded or Failed phase, respectively.
+    The controller sends a status update.
+3. The Job controller removes the `batch.kubernetes.io/job-completion` finalizer
+   from all Pods on Succeeded or Failed phase that were added to the lists in
+   `.status.uncountedTerminatedPods` in the previous step.
+4. The Job controller counts the Pods in the `.status.uncountedTerminatedPods` lists
+   that:
+   - have no finalizer, or
+   - were removed from the system.
+   The counts increment the `.status.failed` and `.status.succeeded` and clears
+   counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
+   status update.
+
+Steps 2 to 4 might deal with a potentially big number of Pods. Thus, status
+updates can potentially stress the kube-apiserver. For this reason, the Job
+controller repeats steps 2 to 4, capping the number of Pods each time, until
+the controller processes all the Pods. The number of Pods is caped by:
+- time: in each iteration, the job controller removes all the Pods' finalizers
+  it can in a unit of time in the order of tens of seconds. This allows to
+  throttle the number of Job status updates.
+- count: Preventing big writes to etcd. We limit the number of UIDs to the order
+  of hundreds, keeping the size of the slice under 20kb.
+
+If any Pod finalizer removal fails in step 3, the controller manager still
+executes step 4 with the Pods that succeeded.
+
+Steps 2 to 4 might be skipped in the scenario where a status update happened
+too recently and the number of uncounted Pods is a small percentage of
+parallelism.
+
+### Deleted Pods
+   
+In the case where a user or another controller removes a Pod, which sets a
+deletion timestamp, the Job controller treats it the same as any other Pod.
+That is, once it reaches Failed status, the controller accounts for the Pod and
+then removes the finalizer.
+
+This is different from the legacy tracking, where the Job controller does not
+account for deleted Pods. This is a limitation that this KEP also wants to
+solve.
+
+However, if the Job controller deletes the Pod (when parallelism is decreased,
+for example), the controller removes the finalizer before deleting it. Thus,
+these deletions don't count towards the failures.
+   
+### Pod adoption
+
+If a Job with `.status.uncountedTerminatedPods != nil` can adopt a Pod
+(according to the existing adoption criteria), this Pod might not have a
+finalizer.
+
+The job controller adds the finalizer in the same patch request that modifies
+the owner reference.
+
+### Test Plan
+
+- Unit tests:
+  - Job sync with feature gate enabled.
+  - Removal of finalizers when feature gate is disabled.
+  - Tracking of terminating Pods.
+- Integration tests:
+  - Job tracking with feature enabled.
+  - Tracking of terminating Pods.
+  - Transition from feature enabled to disabled and enabled again.
+  - Tracking Jobs with big number of Pods, making sure the status is eventually
+    consistent.
+- E2E test:
+  - Job tracking with feature enabled.
+
+### Graduation Criteria
+
+#### Alpha
+
+- Implementation:
+  - Job tracking without lingering Pods
+  - Removal of finalizer when feature gate is disabled.
+- Tests: unit, integration, E2E
+
+#### Alpha -> Beta Graduation
+
+- Support for [Indexed Jobs](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
+- Processing 5000 Pods per minute across any number of Jobs, with Pod creation
+  having higher priority than status updates. This might depend on
+  [Priority and Fairness](https://git.k8s.io/enhancements/keps/sig-api-machinery/1040-priority-and-fairness).
+- Metrics:
+  - latency
+  - errors
+- Tests are in Testgrid and linked in KEP
+
+#### Beta -> GA Graduation
+
+- E2E test graduates to conformance.
+- Job tracking scales to 10^5 completions per Job processed within an order of
+  minutes.
+
+### Upgrade / Downgrade Strategy
+
+When the feature `JobTrackingWithoutLingeringPods` is enabled for the first
+time, the cluster can have Jobs whose Pods don't have the
+`batch.kubernetes.io/job-completion` finalizer. It would be hard to add the
+finalizer to all Pods while preventing race conditions.
+
+We use `.status.uncountedTerminatedPods != nil` to indicate whether the Job
+was created after the feature was enabled. If this field is nil, the Job
+controller tracks Pods using the legacy tracking.
+
+The kube-apiserver sets `.status.uncountedTerminatedPods` to an empty struct
+when the feature gate `JobTrackingWithoutLingeringPods` is enabled, at Job
+creation. In alpha, apiserver leaves `.status.uncountedTerminatedPods = nil`
+for [Indexed Jobs](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
+
+When the feature is disabled after being enabled for some time, the next time
+the Job controller syncs a Job:
+1. It removes finalizers from all Pods owned by the Job.
+2. Sets `.status.uncountedTerminatedPods` to nil.
+
+### Version Skew Strategy
+
+No implications to node runtime.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: JobTrackingWithoutLingeringPods
+    - Components depending on the feature gate:
+      - kube-apiserver
+      - kube-controller-manager
+  - [ ] Other
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+
+  Yes.
+  
+  - Removing terminated Pods doesn't affect Job status.
+  - Pods removed by the user or other controllers count towards failures or
+    completions.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  
+  Yes.
+  The job controller removes finalizers in this case.
+  Since some succeeded Pods might have been removed, the job controller will
+  create new Pods to fulfill completions. But this is no different from
+  existing behavior with the legacy tracking.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+
+  Existing Jobs are tracked with legacy Pods.
+
+* **Are there any tests for feature enablement/disablement?**
+
+  Yes, we plan to add integration tests.
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
+
+* **What specific metrics should inform a rollback?**
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  Describe manual testing that was done and the outcomes.
+  Longer term, we may want to require automated upgrade/rollback tests, but we
+  are missing a bunch of machinery and tooling and can't do that now.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
+fields of API types, flags, etc.?**
+  Even if applying deprecation policies, they may still surprise some users.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+  checking if there are objects with field X set) may be a last resort. Avoid
+  logs or events for this purpose.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine 
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+  - [ ] Other (treat as last resort)
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
+  - 99% percentile over day of absolute value from (job creation time minus expected
+    job creation time) for cron job <= 10%
+  - 99,9% of /health requests per day finish with 200 code
+
+* **Are there any missing metrics that would be useful to have to improve observability 
+of this feature?**
+  Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+  implementation difficulties, etc.).
+
+### Dependencies
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **Does this feature depend on any specific services running in the cluster?**
+  Think about both cluster-level services (e.g. metrics-server) as well
+  as node-level agents (e.g. specific version of CRI). Focus on external or
+  optional services that are needed. For example, if this feature depends on
+  a cloud provider API, or upon an external software-defined storage or network
+  control plane.
+
+  For each of these, fill in the following—thinking about running existing user workloads
+  and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+
+
+### Scalability
+
+_For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them._
+
+_For beta, this section is required: reviewers must answer these questions._
+
+_For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
+
+  - PATCH Pods, to remove finalizers.
+    - estimated throughput: one per Pod created by the Job controller, when Pod
+      finishes or is removed.
+    - originating component: kube-controller-manager
+  - PUT Job status, to keep track of uncounted Pods.
+    - estimated throughput: at least one per Job sync. The job controller
+      throttles additional calls at 1 per a few seconds (precise throughput TBD
+      from experiments).
+    - originating component: kube-controller-manager.
+
+* **Will enabling / using this feature result in introducing new API types?**
+
+  No.
+
+* **Will enabling / using this feature result in any new calls to the cloud 
+provider?**
+
+  No.
+
+* **Will enabling / using this feature result in increasing size or count of 
+the existing API objects?**
+
+  - Pod
+    - Estimated increase: new finalizer of 33 bytes.
+  - Job status
+    - Estimated increase: new array temporarily containing terminated Pod UIDs.
+      The job controller caps the size of the array to less than 20kb.
+
+* **Will enabling / using this feature result in increasing time taken by any 
+operations covered by [existing SLIs/SLOs]?**
+
+  No existing SLIs/SLOs for Jobs.
+
+* **Will enabling / using this feature result in non-negligible increase of 
+resource usage (CPU, RAM, disk, IO, ...) in any components?**
+
+  Additional memory to hold terminated Pods and the status of removing their
+  finalizers.
+
+### Troubleshooting
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How does this feature react if the API server and/or etcd is unavailable?**
+
+* **What are other known failure modes?**
+  For each of them, fill in the following information by copying the below template:
+  - [Failure mode brief description]
+    - Detection: How can it be detected via metrics? Stated another way:
+      how can an operator troubleshoot without logging into a master or worker node?
+    - Mitigations: What can be done to stop the bleeding, especially for already
+      running user workloads?
+    - Diagnostics: What are the useful log messages and their required logging
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+
+* **What steps should be taken if SLOs are not being met to determine the problem?**
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
+
+## Implementation History
+
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+- Extra API calls and temporarily bigger Job status. However, without them
+  it's impossible to ever scale the Job controller to deal with greater amount
+  of Jobs or Jobs with greater amount of Pods.
+
+## Alternatives
+
+- Keep a list of created Pod UIDs, clearing them when they have been accounted.
+  This has the benefit of requiring less Job status updates. On the other hand,
+  the size of the updates is unbounded.
--- a/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
+++ b/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
@@ -76,6 +76,7 @@ SIG Architecture for cross-cutting KEPs).
   - [API changes](#api-changes)
   - [Algorithm](#algorithm)
   - [Deleted Pods](#deleted-pods)
+  - [Deleted Jobs](#deleted-jobs)
   - [Pod adoption](#pod-adoption)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
@@ -221,7 +222,7 @@ type JobStatus struct {
 type UncountedTerminatedPods struct {
     // Succeeded holds UIDs of succeeded Pods.
     Succeeded []types.UID
-    // Succeeded holds UIDs of failed Pods.
+    // Failed holds UIDs of failed Pods.
     Failed    []types.UID
 }
 ```
@@ -240,7 +241,9 @@ cycle.
 1. The Job controller calculates the number of succeeded Pods as the sum of:
    - `.status.succeeded`,
    - the size of `job.status.uncountedTerminatedPods.succeeded` and
-   - the number of finished Pods that are not in `job.status.uncountedTerminatedPods.succeeded`.
+   - the number of finished Pods that are not in `job.status.uncountedTerminatedPods.succeeded`
+     and have a finalizer.
+
    This number informs the creation of missing Pods to reach `.spec.completions`.
    The controller creates Pods for a Job with the finalizer
    `batch.kubernetes.io/job-completion`.
@@ -257,8 +260,8 @@ cycle.
    - have no finalizer, or
    - were removed from the system.
    The counts increment the `.status.failed` and `.status.succeeded` and clears
-   counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
-   status update.
+   counted Pods from `.status.uncountedTerminatedPods` lists. The controller
+   sends a status update.
 
 Steps 2 to 4 might deal with a potentially big number of Pods. Thus, status
 updates can potentially stress the kube-apiserver. For this reason, the Job
@@ -291,6 +294,12 @@ solve.
 However, if the Job controller deletes the Pod (when parallelism is decreased,
 for example), the controller removes the finalizer before deleting it. Thus,
 these deletions don't count towards the failures.
+
+### Deleted Jobs
+
+When a user or another controller deletes a Job, the job controller scans
+associated Pods and removes finalizers from them without updating any Job
+status.
    
 ### Pod adoption
 
--- a/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
+++ b/keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
@@ -1,0 +1,42 @@
+title: Job tracking without lingering Pods
+kep-number: 2307
+authors:
+  - "@alculquicondor"
+owning-sig: sig-apps
+status: implementable
+creation-date: 2020-01-21
+reviewers:
+  - "@erictune"
+  - "@lavalamp"
+  - "@soltysh"
+approvers:
+  - "@janetkuo"
+prr-approvers:
+- "@wojtek-t"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.21"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.21"
+  beta: "v1.22"
+  stable: "v1.24"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: JobTrackingWithoutLingeringPods
+    components:
+      - kube-apiserver
+      - kube-controller-manager
+disable-supported: true
+
+# The following PRR answers are required at beta release
+metrics:
+  - TBD


alculquicondor:
/assign @erictune 
for sig-apps
/assign @lavalamp 
for api changes

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+### Non-Goals
+
+- Remove Pods once they have been accounted for.
+
+## Proposal
+
+The Job controller creates Pods with a finalizer to prevent finished Pods to be
+removed by the garbage collection. The Job controller removes the finalizer from
+the finished Pods once it has accounted for them. In subsequent Job syncs, the
+controller ignores finished Pods that don't have a finalizer.
```
lavalamp, line 143:
```suggestion
controller ignores finished Pods that don't have the finalizer.
```

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+    // UncountedPodUIDs holds UIDs of Pods that have finished but haven't
+    // been accounted in the counters.
+    UncountedPodUIDs UncountedPodUIDs
+}
+
+// UncountedPodUIDs holds UIDs of Pods that have finished but haven't been
+// accounted in Job status counters.
+type UncountedPodUIDs struct {
+    // Enabled indicates whether Job tracking uses this struct.
```
lavalamp, line 183: Perhaps just make the `UncountedPodUIDs` field optional, then you don't need this?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+1. The Job controller creates missing Pods with the finalizer
+   `batch.kubernetes.io/job-completion`
+2. The Job controller adds Pod UIDs to the `.status.uncountedPodUIDs.succeeded`
+   and `.status.uncountedPodsUIDs.failed` lists if the Pod has the
+   `batch.kubernetes.io/job-completion` finalizer and the Pod is on Succeeded
+   or Failed phase, respectively. The controller sends a status update.
+3. The Job controller removes the `batch.kubernetes.io/job-completion` finalizer
+   from all Pods on Succeeded or Failed phase. We can assume that all of them
+   have been added to the lists if the previous step succeeded.
```
lavalamp, line 208: If any of these writes fail, you have to stop the whole thing and start over again. You could permit partial success by noting which finalizer removals succeeded, and only count those pods in step 4.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+4. The Job controller increments the `.status.failed` and `.status.succeeded`
+   counters with the lengths of the `.status.uncountedPodsUIDs` lists, clearing
+   them afterwards. The controller sends a status update.
+   
+In the case where a user or another controller removes a Pod, which sets a
+deletion timestamp, the Job controller removes the finalizer. This is along with
```
lavalamp, line 214: I'm not sure I follow what you're trying to say here? It's not wrong, I just don't understand why it needs to be said here.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+other Pods in step 3. Note that step 2 doesn't pay attention to deletion
+timestamps. So deleted Pods will be accounted in the lists and counters if they
+are in a Failed or Succeeded phase.
+   
+<<[UNRESOLVED Pod adoption ]>>
+If a Job with `.status.uncountedPodUIDs.enabled = true` can adopt a Pod, this
+might not have a finalizer. We can:
+1. Prevent adoption if the Pod doesn't have the finalizer. This is a breaking
+   change.
+1. Add the finalizer in the same patch request that modifies the owner
```
lavalamp, line 224: Should go with this option.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+-->
+
+### Upgrade / Downgrade Strategy
+
+When the feature `JobTrackingWithoutLingeringPods` is enabled for the first
+time, the cluster can have Jobs whose Pods don't have the
+`batch.kubernetes.io/job-completion` finalizer. It would be hard to add the
+finalizer to all Pods while preventing race conditions.
+
+We use `.status.uncountedPodUIDs.enabled` to indicate whether the Job was
```
lavalamp, line 312: What happens if we roll back to a version of the job controller that doesn't know about the finalizer?

lavalamp (commented):
Looks good, I think the downgrade scenario is the only important comment.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
wojtek-t: +1 - I don't see any disadvantages of this option.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+```golang
+type JobStatus struct {
+    Succeeded int32
+    Failed    int32
+    ...
+
+    // UncountedPodUIDs holds UIDs of Pods that have finished but haven't
+    // been accounted in the counters.
+    UncountedPodUIDs UncountedPodUIDs
```
wojtek-t, line 177: Please ensure that the size of that is bounded (controller won't put more than e.g. 100 pods in that list).
If there are more pods to remove - it should be done in few steps.

[The cost of job update api calls is amortized by finalizers removal anyway - so that sounds fine.]

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
wojtek-t: And what happens when we upgrade again :)

wojtek-t (commented):
Agree with Daniel - that looks reasonable to me.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: Yes. But after some more thought, using an explicit "Enabled" field is probably clearer, so I guess let's keep it how you have it.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: I don't think this should be much harder than 1.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: Unfortunately, I think you might need to roll out in two releases; the first release will remove the finalizer if it is present, but doesn't add it. The second release can then start adding the finalizer and implement the rest of the feature.

The alternative is to do both of a) turning off the feature flag won't disable the finalizer removal logic and b) give users a command/script which they can use to manually remove all finalizers in the event of a downgrade.

If you're going to have an alpha release anyway, then I think it's fine to omit b) since the vast majority of users won't be in this situation.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: What I don't like about the boolean is that it will eventually be useless. And the struct is only supposed to be handled by the Job controller, so I think I'm fine with the pointer if we can differentiate between `nil` and empty struct.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: I'm fine with the alternative. Is there a directory where we already have scripts like the one described?

lavalamp: No, the release notes is probably the place to put it. But if it's just alpha in the first release, probably no need.

@liggitt may have an opinion

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+  Pods.
+
+### Non-Goals
+
+- Remove Pods once they have been accounted for.
+
+## Proposal
+
+The Job controller creates Pods with a finalizer to prevent finished Pods to be
+removed by the garbage collection. The Job controller removes the finalizer from
```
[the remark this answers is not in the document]
alculquicondor: Done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+
+### Algorithm
+
+The following algorithm updates the status counters without relying on finished
+Pods to be present indefinitely. The algorithm assumes that the Job controller
+could be stopped at any point and executed again from the first step without
+losing information. Generally, all the steps happen in a single Job sync
+cycle.
+
+1. The Job controller creates missing Pods with the finalizer
```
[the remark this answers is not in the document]
alculquicondor: Done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+    - the Pod is on Succeeded or Failed phase, respectively.
+    The controller sends a status update.
+3. The Job controller removes the `batch.kubernetes.io/job-completion` finalizer
+   from all Pods on Succeeded or Failed phase. We can assume that all of them
+   have been added to the lists if the previous step succeeded. The controller
+   keeps note of the Pod updates that succedeed.
+4. The Job controller counts the Pods in the `.status.uncountedPodUIDs` lists
+   that:
+   - had no finalizer at the beginning of the sync, or
+   - had the patch in step 3 succeed, or
```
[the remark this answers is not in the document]
alculquicondor: I don't see harm in the detail.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+   `batch.kubernetes.io/job-completion`
+2. The Job controller adds Pod UIDs to the `.status.uncountedPodUIDs.succeeded`
+   and `.status.uncountedPodsUIDs.failed` lists if the Pod:
+    - has the `batch.kubernetes.io/job-completion` finalizer, and
+    - the Pod is on Succeeded or Failed phase, respectively.
+    The controller sends a status update.
+3. The Job controller removes the `batch.kubernetes.io/job-completion` finalizer
+   from all Pods on Succeeded or Failed phase. We can assume that all of them
+   have been added to the lists if the previous step succeeded. The controller
+   keeps note of the Pod updates that succedeed.
```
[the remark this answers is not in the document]
alculquicondor: adjusted

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+the controller processes all the Pods.
+
+### User deleted Pods
+   
+In the case where a user or another controller removes a Pod, which sets a
+deletion timestamp, the Job controller removes the finalizer from the Pod,
+<<[UNRESOLVED]>>
+independently of it having finished or not.
+This allows for deleted Pods to not count towards the failed counter, like it
+is today.
```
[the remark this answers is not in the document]
alculquicondor: it's not a change of behavior. Adjusting wording

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+Note that step 2 doesn't pay attention to deletion timestamps. So the Job
+controller might account the Pod in the lists and counters if the Pods
+enter Failed or Succeeded phase before or while being deleted.
```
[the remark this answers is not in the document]
alculquicondor: This is how the controller works today. Except that once the Pods are completely removed, they don't count anymore.

So maybe we should just stop counting pods with a termination timestamp altogether? That's the discussion for this UNRESOLVED section.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,601 @@
...
+<<[/UNRESOLVED]>>
+   
+### Pod adoption
+
+<<[UNRESOLVED Pod adoption ]>>
+If a Job with `.status.uncountedPodUIDs.enabled = true` can adopt a Pod, this
+might not have a finalizer. We can:
```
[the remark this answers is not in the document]
alculquicondor: The qualifications for Pod adoption don't change from current behavior. Clarified.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: removed unresolved taking option 2.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
ahg-g: Sometimes the detail add confusion rather than clarity. As a reader, I paused a little reading this wondering why are we distinguishing between the two. When I read a proposal, I expect to read about the semantics or behavior we are trying to achieve, if there is uncertainly on how to achieve that behavior, then I expect to see an implementation detail section for example. This is a nit, so take for what it is worth.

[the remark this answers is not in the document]
alculquicondor: The high-level overview is in Proposal and Notes. This is the section for nitty-gritty details.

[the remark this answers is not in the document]
ahg-g: The whole doc is a proposal (it is a key enhancement proposal) :)

there are limits to nitty-gritty details, there are ones that are useful and ones that add confusion because they don't add to the understanding of what is being proposed, my feedback is that this added confusion... again, I am replying not because I am insisting on the change, but just as a response to the justification you provided.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
ahg-g: Does the current qualification criteria include pods that failed/succeeded? If the pod was in finished/succeeded status, the absence of the finalizer could be because the pod was already processed by the controller in a previous cycle.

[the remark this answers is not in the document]
alculquicondor: In that scenario, the Pod would already have an updated owner reference.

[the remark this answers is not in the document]
ahg-g: what guarantees that? is it possible that an owner reference be cleared after it was counted for?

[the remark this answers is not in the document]
alculquicondor: Actually there is currently no criteria other than the Job selector, which is mutable.

In theory, a Pod could finish, be accounted for Job A, loose its finalizer and then be claimed by another Job B. At this point, I think it's fair to adopt this Pod and account for it in Job B.
But what if Job A claims the Pod again? Then we have no way of adopting it without counting it again.

@wojtek-t @lavalamp thoughts?

Perhaps we should actually not adopt a Pod if it doesn't have the finalizer.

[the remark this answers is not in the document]
lavalamp: You should really not have this problem, because only one job can have the "controller reference"; you should not clear that, and no one else should either.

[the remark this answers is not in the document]
ahg-g: right, users should not, but they could. While unlikely, I think this is worth a note because in practice this is a regression compared to current situation.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
liggitt: > turning off the feature flag won't disable the finalizer removal logic

I haven't read the full context, but auto-added finalizers not being removed on rollback is a really nasty failure mode. Ungated logic to clean up a finalizer starting in alpha stage could be reasonable. This is similar to what was done with the `kubernetes.io/pvc-protection` finalizer in https://github.com/kubernetes/kubernetes/pull/61324 / https://github.com/kubernetes/kubernetes/pull/61370

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
alculquicondor: Are you referring to this being documented as unexpected behavior? https://kubernetes.io/docs/concepts/workloads/controllers/job/#specifying-your-own-pod-selector

Although I'm proposing an scenario where ownership is transferred from Job A to Job B and then back to Job A by carefully updating the selectors so that there are no conflicts at a given time. But maybe we can document this as unexpected behavior too.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: > Ungated logic to clean up a finalizer starting in alpha stage could be reasonable.

That is what's being proposed.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,599 @@
...
+   - had the patch in step 3 succeed, or
+   - were removed from the system.
+   The counts increment the `.status.failed` and `.status.succeeded` and clears
+    counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
+    status update.
+
+Steps 2 to 4 might deal with a potentially big number of Pods. Thus, status
+updates can potentially stress the kube-apiserver. For this reason, the Job
+controller repeats steps 2 to 4, dealing with up to 100 Pods each time, until
+the controller processes all the Pods.
```
erictune, line 225: I think you added this in response to Daniel's comment.
@lavalamp , were you worried about too frequent of updates (not batching) or about the size of the diff from a single update (which the above seems concerned with).

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,599 @@
...
+   `.status.uncountedPodUIDs` in the previous step. The controller keeps note
+   of the Pod updates that succedeed.
+4. The Job controller counts the Pods in the `.status.uncountedPodUIDs` lists
+   that:
+   - had no finalizer at the beginning of the sync, or
+   - had the patch in step 3 succeed, or
+   - were removed from the system.
+   The counts increment the `.status.failed` and `.status.succeeded` and clears
+    counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
+    status update.
```
erictune, line 220: Talk about if there are changes to the logic where the controller decides whether to start new pods.
Does it look at the current `.status.succeeded` and compare that to `.spec.completions`?  Or does it look at  `.status.succeeded + len( status.uncountedPodUIDs.succeeded`?  The list could be non-empty due to the throttling mentioned below.  Or does it block starting new pods if there are any uncounted pods?

erictune (changes requested):

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: I don't think I'm concerned about either. I do think it's probably good to limit the size of the chunks we process just to guarantee progress (partial update is better than never completing a huge update). But I personally would probably have a time limit-- process all the pods you can in 1m. I would say 15s but we have to consider the possibility of apiserver taking the maximum amount of time to respond to a request (currently 60s). The tests probably need to test under the "apiserver is really slow" condition.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: This is a good point, "uncounted" isn't quite right, it's "terminatedAndUncounted". Yes, that logic needs to look at 3 sources of information, a) the counts, b) the existing pods which are not in c) the terminatedAndUncounted list.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: > process all the pods you can in 1m.

Uhmm. That would have to look like:
- repeatedly update K Pods in parallel while 1m hasn't passed.
- setup a channel and K workers, continue processing for 1m.

So then we have to decide on K. Which exact method we use can probably be left for code reviews.

alculquicondor: cc @wojtek-t since you were asking about batching updates in the other KEP

wojtek-t: Hmm - I don't understand why updating the status (a) with N pods and (b) with M pods is different from kube-apiserver perspective.

I think what we should do is: (a) take a time into account, (b) have a cap on number of pods addeded.

We need a limit to ensure that the object won't grow to super large case, e.g. if job-controller was down for some time and in the meantime gazillions pod finished.

alculquicondor: So I take this as: we need to cap on time and number. Updated the KEP

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: changed

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
alculquicondor: @soltysh @erictune

I would appreciate your input here. There is one case where I think it might be worth reconsidering the semantics. Let's say a Node becomes unavailable and gets removed. In this case, the garbage collection also removes the Pod. Should this be considered a failure or not? Today, it's not.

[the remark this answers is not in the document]
erictune: I think users ideally want failures caused by non-zero exit code (a problem within the user's control) to be counted separately from failures out of their control (preemption, node failure).

Does Failure today include both or just the former?   If it is just the former, we should try to keep it that way.  If it is a mixture, then needs a bit more thought.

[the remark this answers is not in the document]
alculquicondor: Today, it only includes failures from non-zero exit code. This is because other failures get the Pod removed.

Then, it sounds like we should be skipping Pods that have a termination timestamp from any calculations.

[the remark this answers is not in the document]
alculquicondor: Wait, but there's a problem with the above:
The Pod GC doesn't take into account finalizers. That means that Pods that this controller marks for deletion would also not be counted towards calculations.
Then I think the solution would be: not to count Pods that have a termination timestamp and modify the Pod GC to skip Pods with finalizers.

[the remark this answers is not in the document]
lavalamp: The impression I've gathered from what I've seen from user reports is that users get quite grumpy when a pod starts and doesn't complete for some reason, and would like to know about that. Would they like the two classes of failures conflated? I'm less sure about the answer to that.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,531 @@
...
+the controller ignores finished Pods that don't have the finalizer.
+
+### Notes/Constraints/Caveats (Optional)
+
+Due to the lack of support of atomic changes across kubernetes objects, an
+intermediate state is necessary. Before removing the Pod finalizers,
+the controller adds finished Pods to a list in the Job status. After removing
+the Pod finalizers, the controller clears the list and updates the counters.
+
+### Risks and Mitigations
```
ahg-g, line 156: is increasing object size a risk that we should mention?

alculquicondor: yes, added.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+the controller processes all the Pods. The number of Pods is caped by:
+- time: in each iteration, the job controller removes all the Pods' finalizers
+  it can in 1 minute. This allows to throttle the number of status updates.
+- count: Preventing big writes to etcd.
+
+### User deleted Pods
+   
+In the case where a user or another controller removes a Pod, which sets a
+deletion timestamp, the Job controller removes the finalizer from the Pod,
+<<[UNRESOLVED: should we count terminating pods?]>>
```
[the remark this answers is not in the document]
soltysh: My reasoning is that the controller did create that pod and the fact that it was removed for whatever reason should be treated equally with any other failure.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+independently of it having finished or not.
+This allows for deleted Pods to not count towards the failed counter, so it
+doesn't change the current behavior.
```
soltysh, line 247: The current controller can't count pods that don't exist :(

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
soltysh: Like I said above, controller created pods that failed (removed, evicted, non-zero exit) should all be treated as failed. The fact that current controller doesn't do it is one of the limitation this proposal is trying to address. Iow. with improved status calculation we'll get more accurate signal of failure. Example, if there's a bad actor in the system consistently removing pods from a job an increased failure rate on a job might make the job author alarmed, and that's a good thing, imo.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+- Implementation:
+  - Job tracking without lingering Pods
+  - Removal of finalizer when feature gate is disabled.
+- Tests: unit, integration, E2E
+
+#### Alpha -> Beta Graduation
+
+- Implementation:
+  - Index tracking for [Indexed Jobs](http://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
```
soltysh, line 287: This is not related effort.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+- Job tracking scales to 10^5 completions.
+
+### Upgrade / Downgrade Strategy
+
+When the feature `JobTrackingWithoutLingeringPods` is enabled for the first
+time, the cluster can have Jobs whose Pods don't have the
+`batch.kubernetes.io/job-completion` finalizer. It would be hard to add the
+finalizer to all Pods while preventing race conditions.
+
+We use `.status.uncountedTerminatedPodUIDs.enabled` to indicate whether the Job
```
soltysh, line 305: From the discussion above, `.enabled` is no more, values in `.status.uncountedTeminatedPodUIDs` should be sufficient.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+was created after the feature was enabled. If this boolean is false, the Job
+controller tracks Pods using the legacy algorithm.
+
+The kube-apiserver sets `.status.uncountedTerminatedPodUIDs.enabled = true` when
+the feature gate `JobTrackingWithoutLingeringPods` is enabled, at Job creation.
```
soltysh, line 310: I'm not sure if this is actually needed, the controller should be capable of reacting to the current state of the job and update its status accordingly.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+When the feature is disabled after being enabled for some time, the next time
+the Job controller syncs a Job:
+1. It removes finalizers from all Pods owned by the Job.
+2. Sets `.status.uncountedTerminatedPodUIDs.enabled = false`
```
soltysh, line 315: Sets `.status.uncountedTerminatedPodUIDs` to nil.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: JobTrackingWithoutLingeringPods
+    - Components depending on the feature gate:
+      - kube-apiserver
+      - kube-controller-manager
```
soltysh, line 332: I'd probably leave it only on kcm side, no need to touch kas.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+  Yes, there are integration tests.
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
```
soltysh, line 371: There's a risk that after you enable the flag and let it run for awhile some of the pods of a long-running job might get removed, which gets reflected in the status appropriately. But after a restart w/o the flag the legacy behaviour will override the numbers based on existing pods. Should the legacy behaviour account for that?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+kep-number: 2307
+authors:
+  - "@alculquicondor"
+owning-sig: sig-apps
+status: implementable
+creation-date: 2020-01-21
+reviewers:
+  - "@erictune"
+  - "@lavalamp"
+  - "@solthys"
```
soltysh, line 11: Typo, it's soltysh :wink:

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+// UncountedTerminatedPodUIDs holds UIDs of Pods that have finished but haven't
+// been accounted in Job status counters.
+type UncountedTerminatedPodUIDs struct {
+    // Enabled indicates whether Job tracking uses this struct.
+    Enabled   bool
+    // Succeeded holds UIDs of succeeded Pods.
+    Succeeded []string
+    // Succeeded holds UIDs of failed Pods.
+    Failed    []string
```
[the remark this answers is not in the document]
alculquicondor: It didn't work in my earlier design, but now it does.

It also allows to remove UIDs from the name of the struct

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: changed to pointer.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
alculquicondor: I added this as a request from API reviewer to be explicit about continuing to process the Pods that succeed.
Moved it as a note a paragraph below.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,534 @@
...
+   The counts increment the `.status.failed` and `.status.succeeded` and clears
+   counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
+   status update.
+
+Steps 2 to 4 might deal with a potentially big number of Pods. Thus, status
+updates can potentially stress the kube-apiserver. For this reason, the Job
+controller repeats steps 2 to 4, capping the number of Pods each time, until
+the controller processes all the Pods. The number of Pods is caped by:
+- time: in each iteration, the job controller removes all the Pods' finalizers
+  it can in 1 minute. This allows to throttle the number of status updates.
```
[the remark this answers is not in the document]
alculquicondor: changed to "order of minutes". It's actually not a small unit of time.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: We need this for any new API field to be able to handle partial upgrades in a HA control plane.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: we do need to set the struct to non-nil when the feature gate is enabled.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: true. it's a beta requirement for Indexed Job feature. Removed from here.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
alculquicondor: Ack

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Ack

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
alculquicondor: I agree that this can be seen as a limitation of the legacy tracking.

Only when the job controller itself removes the Pod it should NOT be considered a failure. This happens when parallelism is reduced or when the Job controller created extra unnecessary Pods. Added a note.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: yes

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: The legacy behavior shouldn't change if the feature is disabled (except for removing finalizers)

I added a note for this in "What happens if we reenable the feature if it was previously rolled back?"

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
alculquicondor: Sorry 😅

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,537 @@
...
+### Notes/Constraints/Caveats (Optional)
+
+Due to the lack of support of atomic changes across kubernetes objects, an
+intermediate state is necessary. Before removing the Pod finalizers,
+the controller adds finished Pods to a list in the Job status. After removing
+the Pod finalizers, the controller clears the list and updates the counters.
+
+### Risks and Mitigations
+
+1. The new implementation produces new API calls: one per Pod to remove
```
ahg-g, line 158: Can we have an estimate of the number of extra qps updates that will be used to implement this approach? 

Workloads that run many short running single-pod jobs may not benefit too much from this new approach, at the same time it will negatively impact throughput since we are using some of the controller qps limit towards adding and removing the finalizer from the pod and perhaps extra updates to the job status.

/cc @wojtek-t

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,537 @@
...
+   finalizers and one for each Job sync, to track the intermediate state.
+   However, the ability to handle higher number of Pods and Jobs at a given
+   point justifies the added cost of the new API calls.
```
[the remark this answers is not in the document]
alculquicondor: Updated

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Not sure what else I can say: there is one more API call for each Pod created by the controller. And one for each Job, each time a group of Pods finish.

Other than that, there is no other option for single-pod jobs. We still need to guarantee that the Pod is accounted, so the finalizer is also needed.

I considered just adding the Pod to the uncounted lists at Pod creation. But that doesn't work, because we don't know if the Pod will be successful or not.

ahg-g: We can estimate the impact on throughput assuming an x qps limit for the following two extremes: 1) for one pod per job and 2) many pods per job. The former is probably the more important one especially for short running jobs.

How many updates per job do we currently send for a single pod job? two (one for creating the job and one for updating the status when the job completes)?

alculquicondor: added more details and justification

alculquicondor: And @lavalamp gave me the idea of one alleviation: to skip updates sometimes. See latest commit.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,570 @@
...
+    // UncountedTerminatedPods holds UIDs of Pods that have finished but
+    // haven't been accounted in the counters.
+    // If nil, Job tracking doesn't make use of this structure.
+    // +optional
+    UncountedTerminatedPods *UncountedTerminatedPods
+}
+
+// UncountedTerminatedPods holds UIDs of Pods that have finished but haven't
+// been accounted in Job status counters.
+type UncountedTerminatedPods struct {
```
soltysh, line 218: I'm not in favor of this name, but we can work it out during API review and update the proposal afterwards.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,570 @@
...
+In the case where a user or another controller removes a Pod, which sets a
+deletion timestamp, the Job controller treats it the same as any other Pod.
+That is, once it reaches Failed status, the controller accounts for the Pod and
+then removes the finalizer.
+
+This is different from the legacy tracking, where the Job controller does not
+account for deleted Pods. This is a limitation that this KEP also wants to
+solve.
+
+However, if it's the Job controller what deletes the Pod (when parallelism
```
soltysh, line 284:
```suggestion
However, if the Job controller deletes the Pod (when parallelism
```

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,570 @@
...
+finalizer to all Pods while preventing race conditions.
+
+We use `.status.uncountedTerminatedPods != nil` to indicate whether the Job
+was created after the feature was enabled. If this field is nil, the Job
+controller tracks Pods using the legacy tracking.
+
+The kube-apiserver sets `.status.uncountedTerminatedPods` to an empty struct
+when the feature gate `JobTrackingWithoutLingeringPods` is enabled, at Job
+creation. In alpha, apiserver leaves `.status.uncountedTerminatedPods = nil`
+for [Indexed Jobs](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
```
soltysh, line 346: This is not needed, it's solemnly  controller responsibility to set .status field.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
soltysh: Nope, it's not needed, we don't do it for any of the controllers. It's controller's responsibility to set status accordingly.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: How do you suggest we detect if a Job was existing or not?

soltysh: If the feature-gate is on, the controller will look at that field, in all other cases it can safely ignore it and behave as it does today. Just like you explained that in a different spot where you said that in the worst case where a pod was already removed it will re-create those. In all other cases it will count the already completed ones.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Ack

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
soltysh: Spoke with @alculquicondor on slack, we've agreed that the flag will be for both kcm and kas, kas will be responsible for clearing this field when the feature gate is off.

alculquicondor: Leaving as discussed offline.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Leaving as discussed offline

soltysh (approved):
/lgtm

alculquicondor:
squashed

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+_For beta, this section is required: reviewers must answer these questions._
+
+_For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
+
+  - PATCH Pods, to remove finalizers.
+    - estimated throughput: one per created Pod, when Pod finishes or is removed.
```
johnbelamaric, line 487: one per Pod created by the Job controller, correct? Not for every Pod created.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+  No.
+
+* **Will enabling / using this feature result in increasing size or count of 
+the existing API objects?**
+
+  - Pod
+    - Estimated increase: new finalizer of 33 bytes.
+  - Job status
+    - Estimated increase: new array temporarily containing terminated Pod UIDs.
```
johnbelamaric, line 509: See discussion above. Would like to see estimates on how big these arrays can get in both normal operation and failure modes.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+status: implementable
+creation-date: 2020-01-21
+reviewers:
+  - "@erictune"
+  - "@lavalamp"
+  - "@soltysh"
+approvers:
+  - "@janetkuo"
+prr-approvers:
+- "@wojtek-t"
```
johnbelamaric, line 15: Please change to me, Wojtek is over scheduled.

keps/prod-readiness/sig-apps/2307.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2307
+alpha:
+  approver: "@wojtek-t"
```
johnbelamaric, line 3: Please change to me, Wojtek is over scheduled.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
ahg-g: So for one pod per job we will see ~50% drop in throughput? used to be 1 query to create the pod and 1 to update job status, now we add two more: another job status update and 1 to remove the finalizer?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+type JobStatus struct {
+    Succeeded int32
+    Failed    int32
+    ...
+
+    // UncountedTerminatedPods holds UIDs of Pods that have finished but
+    // haven't been accounted in the counters.
+    // If nil, Job tracking doesn't make use of this structure.
+    // +optional
+    UncountedTerminatedPods *UncountedTerminatedPods
```
wojtek-t, line 213: @lavalamp @erictune - IIUC, in the past we were using a convention that status contains information, than can be computed based on the cluster state.
This is more or less, why things like "AcquireTime" or "RenewTime" are in LeaseSpec:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/coordination/v1/types.go#L40
[that was Brian`s suggestion back then].

Did we give up with this rule? If not, then all `Succeeded`, `Failed`, and `UncountedTerminatedPods` technically no longer belong to status...

[I don't have strong opinion here - just want to flag this for discussion.]

@liggitt - for his API reviewer/approver opinion

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+  create new Pods to fulfill completions. But this is no different from
+  existing behavior with the legacy tracking.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+
+  Existing Jobs are tracked with legacy Pods.
+
+* **Are there any tests for feature enablement/disablement?**
+
+  Yes, there are integration tests.
```
wojtek-t, line 402: Are there? I'm assuming these are planned, but let's not mention they already exist....

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+previous answers based on experience in the field._
+
+* **Will enabling / using this feature result in any new API calls?**
+
+  - PATCH Pods, to remove finalizers.
+    - estimated throughput: one per created Pod, when Pod finishes or is removed.
+    - originating component: kube-controller-manager
+  - PUT Job status, to keep track of uncounted Pods.
+    - estimated throughput: at least one per Job sync. Extra calls are throttled
+      at 1 per minute.
```
wojtek-t, line 491: 1 per minute might be too aggressive

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
wojtek-t: I added comments below - I think we can control this (and we should) - and i agree I want to see a real number here.
But I think this is fine if we do that the right way.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/kep.yaml
wojtek-t: I triggerred that discussion, so I will continue looking into it.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
liggitt: will this throttling delay reflecting job completion in the API object? I didn't see tests explicitly for that.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: FYI, the KEP already mentions limiting the UIDs in the controller

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,573 @@
...
+   status update.
+
+Steps 2 to 4 might deal with a potentially big number of Pods. Thus, status
+updates can potentially stress the kube-apiserver. For this reason, the Job
+controller repeats steps 2 to 4, capping the number of Pods each time, until
+the controller processes all the Pods. The number of Pods is caped by:
+- time: in each iteration, the job controller removes all the Pods' finalizers
+  it can in a unit of time in the order of minutes. This allows to throttle the
+  number of Job status updates.
+- count: Preventing big writes to etcd.
```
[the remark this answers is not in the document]
alculquicondor: saying "order of hundreds"

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: reworded

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: correct, added clarification.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Added to test plan as integration.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: estimating to 10kb for about 500 UUIDs

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
lavalamp: Actually this makes the Success / Finished numbers un-reproducible from the environment. UncountedTerminatedPods is (at least partially) reproducible! I think we are just giving up on this criteria to make this work. It's worth a discussion but I don't see other options. The reproducibility criterion is (IMO) mostly aesthetic (we've never made the system do anything that relied on it and it seems quite unlikely that we will in the future) and it's not worth it to keep it for this object.

liggitt: > Actually this makes the Success / Finished numbers un-reproducible from the environment. UncountedTerminatedPods is (at least partially) reproducible! I think we are just giving up on this criteria to make this work. It's worth a discussion but I don't see other options. The reproducibility criterion is (IMO) mostly aesthetic (we've never made the system do anything that relied on it and it seems quite unlikely that we will in the future) and it's not worth it to keep it for this object.

agreed.

erictune: AIUI, the intent of this API Convention was that status could be stored in some less reliable storage for performance.  5+ years on from this intention, we haven't made use of that possibility.  And I really doubt that all clients are prepared for the event where status disappears and then is later reconstructed.  I am thinking of e.g. kubectl waiting for a to be ready, and then all the status disappears.  Is the client code going to do the right thing?  What is the right thing?  We never designed for that.  So we should remove that API convention.

wojtek-t: Thanks - for me it was also kind of aesthetic thing, but I wanted to ensure that there is a general agreement on that.
It seems there is, which is what I wanted to achieve.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,578 @@
...
+
+* **Will enabling / using this feature result in any new API calls?**
+
+  - PATCH Pods, to remove finalizers.
+    - estimated throughput: one per Pod created by the Job controller, when Pod
+      finishes or is removed.
+    - originating component: kube-controller-manager
+  - PUT Job status, to keep track of uncounted Pods.
+    - estimated throughput: at least one per Job sync. The job controller
+      throttles more calls at 1 per a few minutes (precise throughput TBD from
```
wojtek-t, line 495: nit: per few seconds, not few minutes;

[or to be more specific I think what we should do is:
- as soon as we gather N pods (to feed the 20kB limit you mentioned below)
- but at least every X(=10?) seconds, to make it somewhat responsive

wojtek-t (commented):
I'm fine with this modulo one small comment.

Please gather SIG approval and I will approve.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Changed to seconds and added more detail in the Algorithm section.

Some reviewers requested that the KEP leaves room for experimentation before settling on specific numbers.

alculquicondor:
@janetkuo this is ready for SIG approval

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
johnbelamaric: Thanks I see that now. Do we have estimates on how the caps interact with the maximum anticipated scale, and on how we can shard that work across multiple workers if necessary? That is, I still would like to understand:

1) the anticipated scale
2)the plan for handling the space and time complexity at that maximum scale

If we chunk into small units, that means it may take longer to process, etc.

alculquicondor: Let's say we remove 500 finalizers and do the corresponding status updates every 15s (this roughly corresponds with the default 50 QPS limit on the controller's client). That means that in 1 min, we can process 2000 terminated Pods. Does this sound reasonable?

Note that each finalizer removal happens in its own goroutine.

erictune: I am not sure that will be fast enough for the most demanding of batch users.   Do you have a plan to get feedback on that?

From an API Server standpoint, it would be better to batch up hundreds of finalizer removals into a single update request.  Is it a lot of work for those goroutines to not do the update themselves but to instead send a message through a channel to something else that can batch up the updates?

alculquicondor: does the apiserver currently support batch updates?

lavalamp: No, you have to do the finalizer removals individually since they are on different objects. But you should at least be able to use patch requests which won't have conflicts. And you can perform the requests in parallel and trust APF to keep you from overloading apiserver.

erictune: Which are you doing:

Option 1:
- Say 100 pods recently finished.
- make 100 calls to apiserver for each pods resource.
- make 100 calls to apiserver for the job resource

Option 2:
- Say 100 pods recently finished.
- make 100 calls to apiserver for each pods resource.
- make 1 call to apiserver for the job resource

alculquicondor: option 2, but it's actually 2 api calls to the job resource

alculquicondor: Added expected number of processed Pods per minute and a target of 5000 for beta graduation.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
johnbelamaric: I think the intent is also that this rule forces requirements on the cluster introspection mechanism to be able to surface anything important enough to track in status. If we have the ability to reconstruct status, then we know resources are not leaked out; our reconciliation is fully closed, without drift. If we put something in status that we cannot discover from the cluster itself we no longer have that guarantee - status could diverge from reality.

That said, in the case of job tracking, there are no resources leaked or lost track of. We lose track of some job stats potentially, but not things that are consuming underlying resources. So, only issue is the slippery slope.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
johnbelamaric: Thanks, and I see GA grad criteria has 10^5 for completions (for a single job I presume). So the job tracking takes about 20 minutes for a job of that size, in this plan, in the degenerate scenario where every pod just exits. Is that 5000 per job, or across all jobs? If there are 10 jobs of 100k pods running, accounting takes 200 minutes? Is that OK for our users? (really asking, I have no idea. it sounds reasonable to me).

alculquicondor: Clarified 5000 to be across any number of Jobs.

As for 10^5 criteria, clarified to be per Job and in the order of minutes. I think we can aim to be 10min, but we can tune that after alpha is implemented.

erictune (approved):

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+That is, once it reaches Failed status, the controller accounts for the Pod and
+then removes the finalizer.
+
+This is different from the legacy tracking, where the Job controller does not
+account for deleted Pods. This is a limitation that this KEP also wants to
+solve.
+
+However, if the Job controller deletes the Pod (when parallelism is decreased,
+for example), the controller removes the finalizer before deleting it. Thus,
+these deletions don't count towards the failures.
```
janetkuo, line 293: From the description here, it'd apply to suspended Jobs ([KEP 2232](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/2232-suspend-jobs/)) as well, I assume?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+    // Succeeded holds UIDs of failed Pods.
+    Failed    []types.UID
+}
+```
+
+Note: the final name of the field `uncountedTerminatedPods` will be decided
+during API review.
+
+### Algorithm
+
```
janetkuo, line 233: If the user deletes the Job, how are the finalizers of its Pods handled?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+  [Priority and Fairness](https://git.k8s.io/enhancements/keps/sig-api-machinery/1040-priority-and-fairness).
+- Metrics:
+  - latency
+  - errors
+- Tests are in Testgrid and linked in KEP
+
+#### Beta -> GA Graduation
+
+- E2E test graduates to conformance.
+- Job tracking scales to 10^5 completions per Job processed within an order of
```
wojtek-t, line 342: If we assume that "order of minutes" is 5 minutes - that would mean O(300) QPS to delete finalizers.

I don't think we should be that aggressive - it's much more than we support now.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+    // +optional
+    UncountedTerminatedPods *UncountedTerminatedPods
+}
+
+// UncountedTerminatedPods holds UIDs of Pods that have finished but haven't
+// been accounted in Job status counters.
+type UncountedTerminatedPods struct {
+    // Succeeded holds UIDs of succeeded Pods.
+    Succeeded []types.UID
+    // Succeeded holds UIDs of failed Pods.
```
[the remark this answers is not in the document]
alculquicondor: done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: yes, but applies to any other future feature that might delete pods.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+    The controller sends a status update.
+3. The Job controller removes the `batch.kubernetes.io/job-completion` finalizer
+   from all Pods on Succeeded or Failed phase that were added to the lists in
+   `.status.uncountedTerminatedPods` in the previous step.
+4. The Job controller counts the Pods in the `.status.uncountedTerminatedPods` lists
+   that:
+   - have no finalizer, or
+   - were removed from the system.
+   The counts increment the `.status.failed` and `.status.succeeded` and clears
+   counted Pods from `.status.uncountedPodsUIDs` lists. The controller sends a
```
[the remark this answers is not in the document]
alculquicondor: done

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+### Algorithm
+
+The following algorithm updates the status counters without relying on finished
+Pods to be present indefinitely. The algorithm assumes that the Job controller
+could be stopped at any point and executed again from the first step without
+losing information. Generally, all the steps happen in a single Job sync
+cycle.
+
+1. The Job controller calculates the number of succeeded Pods as the sum of:
+   - `.status.succeeded`,
```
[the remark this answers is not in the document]
alculquicondor: @lavalamp?

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
```diff
@@ -0,0 +1,585 @@
...
+   - the size of `job.status.uncountedTerminatedPods.succeeded` and
+   - the number of finished Pods that are not in `job.status.uncountedTerminatedPods.succeeded`.
```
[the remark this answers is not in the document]
alculquicondor: right, they should have a finalizer too, fixed.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: Added a section.

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
alculquicondor: I was thinking more like 10min. Is that still too much? This would be assuming that this is the only Job though.

lavalamp: IMO in the situation where you have a job that's that big, 300 QPS is perfectly reasonable-- that's probably a large fraction of what you're running on the cluster anyway. Compare it to the writes it took to actually get all those pods through their lifecycle. It's not adding much.

wojtek-t: If you're saying how it should be in ideal world - sure, I believe even more than should be possible.
If you're asking where we are now - I doubt we will be able to do that (I mean, we can handle 300 qps of course, but we can't blindly bump qps limits until we have P&F supporting things like LISTS, and also some rate-limiting that would rate-limit watches - whatever that means).

Also note - that it highly depends on how long a particular pod takes to run - if those are very short-running, then it matters, if they are long-running, then it's not that

alculquicondor: This is still in the Beta -> GA graduation, so I think it should be fine to have this as tentative goal

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
soltysh: @janetkuo I don't think we are changing this significantly much, there are slight variances how this will get calculated during execution, but the final result will be the same. Nobody should ever rely on how this number is being updated, nor partial results. I'm not personally worried about that breaking API compatibility.

janetkuo (approved):

keps/sig-apps/2307-job-tracking-wihout-lingering-pods/README.md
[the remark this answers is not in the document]
janetkuo: SGTM

wojtek-t:
Adding lgtm, as Janet`s approval was the last thing in this PR.

/lgtm

PR #2323, opened 2021-01-25
Add fsgroup mount proposal

gnufied:
xref https://github.com/kubernetes/enhancements/issues/2317

/sig storage

### DIFF
--- a/keps/sig-storage/2317-fsgroup-on-mount/README.md
+++ b/keps/sig-storage/2317-fsgroup-on-mount/README.md
@@ -1,0 +1,385 @@
+# Provide fsgroup of pod to CSI driver on mount
+
+## Table of Contents
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Currently for most volume plugins kubelet applies fsgroup ownership and permission based changes by recursively `chown`ing and `chmod`ing the files and directories inside a volume. For certain CSI drivers this may not be possible because `chown` and `chmod` are unix primitives and underlying CSI driver
+may not support them. This enhancement proposes providing the CSI driver with fsgroup as an explicit field so as CSI driver can apply this on mount time.
+
+## Motivation
+
+Since some CSI drivers(AzureFile for example) - don't support chmod/chown - we propose that fsgroup of pod to be provided to the CSI driver on `NodeStageVolume`
+and `NodePublishVolume` CSI RPC calls. This will allow the CSI driver to apply `fsgroup` as a mount option on `NodeStageVolume` or `NodePublishVolume` and kubelet can be freed
+from responsibility of applying recursive ownership and permission change.
+
+This feature hence becomes a prerequisite of CSI migration of Azure file driver and removal of Azure Cloud Provider.
+
+### Goals
+
+- Allow CSI driver to mount volumes with provided fsgroup.
+
+### Non-Goals
+
+- We are not supplying `fsGroup` as a generic ownerhip and permission handle to the CSI driver. We do not expect CSI drivers to `chown` or `chmod` files.
+
+## Proposal
+
+We are updating CSI specs by adding additional field called `volume_mount_group` to `NodeStageVolume` and `NodePublishVolume` RPC calls. The CSI proposal is available at - https://github.com/container-storage-interface/spec/pull/468 .
+
+The CSI spec change is deliberately trying to avoid asking drivers to use supplied `fsGroup` as a generic handle for ownership and permissions. The reason being - Kubernetes may expect ownership and permissions to be in a way that is very platform/OS specific. We do not think CSI driver is right place to enforce all kind of different permissions expected by Kubernetes. The full scope of that discussion is out of scope for this enhancement and interested folks can follow along on - https://github.com/container-storage-interface/spec/issues/449
+
+
+### Risks and Mitigations
+
+I am not aware of any associated risks. If a driver can not support using `fsgroup` as a mount option, it can always use `FileFSGroupPolicy` and let kubelet handle the ownership and permissions.
+
+## Design Details
+
+We are proposing that when kubelet determines a CSI driver has `VOLUME_MOUNT_GROUP` node capability, the kubelet will use proposed CSI field `volume_mount_group` to pass pod's `fsGroup` to the CSI driver. Kubelet will expect that driver will use
+this field for mounting volume with given `fsGroup` and no further permission/ownerhip change will be necessary.
+
+It should be noted that if a CSI driver advertises `VOLUME_MOUNT_GROUP` node capability then value defined in `CSIDriver.Spec.FSGroupPolicy` will be ignored and kubelet will always use `fsGroup` as a mount option.
+
+### Test Plan
+
+Unit test:
+1. Test that whenever supported pod's `fsGroup` should be passed to CSI driver via `volume_mount_group` field.
+
+For alpha feature:
+1. Update Azure File CSI driver to support supplying `fsGroup` via `NodeStageVolume` and `NodePublishVolume`.
+1. Run manual tests against azurefile CSI driver.
+
+For beta:
+1. E2E tests that verify volume readability/writability using azurefile CSI driver.
+2. E2E tests using CSI mock driver.
+
+We already have quite a few e2e tests that verify generic fsgroup functionality for existing drivers - https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/fsgroupchangepolicy.go . This should give us a reasonable
+confidence that we won't break any existing drivers.
+
+
+### Graduation Criteria
+
+#### Alpha -> Beta Graduation
+
+- Since this feature is a must-have for azurefile CSI migration, we will perform testing of the driver.
+- Currently CSI spec change is being introduced as alpha change and we will work to move the API change in CSI spec to stable.
+
+#### Beta -> GA Graduation
+
+- CSI spec change should be stable.
+- Tested via e2e and manually using azurefile CSI driver.
+
+### Upgrade / Downgrade Strategy
+
+Currently there is no way to make a volume readable/writable using azurefile and `fsGroup` unless
+pod was running as root.
+
+When feature-gate is disabled, kubelet will no longer pass `fsGroup`  to CSI drivers and such volumes will not be readable/writable by the Pod. This feature is currently broken anyways.
+
+
+<!--
+If applicable, how will the component be upgraded and downgraded? Make sure
+this is in the test plan.
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to maintain previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+### Version Skew Strategy
+
+<!--
+If applicable, how will the component handle version skew with other
+components? What are the guarantees? Make sure this is in the test plan.
+
+Consider the following in developing a version skew strategy for this
+enhancement:
+- Does this enhancement involve coordinating behavior in the control plane and
+  in the kubelet? How does an n-2 kubelet without this feature available behave
+  when this feature is used?
+- Will any other components on the node change? For example, changes to CSI,
+  CRI or CNI may require updating that component before the kubelet.
+-->
+
+## Production Readiness Review Questionnaire
+
+<!--
+
+Production readiness reviews are intended to ensure that features merging into
+Kubernetes are observable, scalable and supportable; can be safely operated in
+production environments, and can be disabled or rolled back in the event they
+cause increased failures in production. See more in the PRR KEP at
+https://git.k8s.io/enhancements/keps/sig-architecture/1194-prod-readiness.
+
+The production readiness review questionnaire must be completed and approved
+for the KEP to move to `implementable` status and be included in the release.
+
+In some cases, the questions below should also have answers in `kep.yaml`. This
+is to enable automation to verify the presence of the review, and to reduce review
+burden and latency.
+
+The KEP must have a approver from the
+[`prod-readiness-approvers`](http://git.k8s.io/enhancements/OWNERS_ALIASES)
+team. Please reach out on the
+[#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) channel if
+you need any help or guidance.
+
+-->
+
+### Feature Enablement and Rollback
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: MountWithFSGroup
+    - Components depending on the feature gate:
+      - Kubelet
+  - [ ] Other
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+  Enabling this feature-gate could result in `CSIDriver.Spec.FSGroupPolicy` to be ignored for a driver that now has
+  `VOLUME_MOUNT_GROUP` capability. Enabling this feature will cause volume to be mounted with `fsGroup`
+  of the pod rather than kubelet performing permission/ownership change. This should result in quicker
+  pod startup but still may surprise some users. This will be covered via release notes.
+
+  We expect that once a CSI driver accepts provided `fsGroup` via `volume_mount_group` and mount
+  operation is successful - the permissions should be correct on the volume. A user may write additional
+  healthcheck to determine the permissions if necessary.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  Yes - feature gate can be disabled once enabled and kubelet will fallback to its current behavior, which will
+  result in any CSI driver that requires using `fsGroup` as a mount option to not use the mount option. In which case
+  mounted volumes will be unwritable by any pods other than those running as root.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+  It will cause kubelet to use `volume_mount_group` field of CSI whenver applicable as discussed in above design. The pods running on affected
+  nods have to restarted for feature to take affect though.
+
+* **Are there any tests for feature enablement/disablement?**
+  Not yet.
+
+### Rollout, Upgrade and Rollback Planning
+
+* **How can a rollout fail? Can it impact already running workloads?**
+  One of the ways the rollout could fail is CSI driver defines `VOLUME_MOUNT_GROUP` capability but somehow does not implement correctly.
+  This change should not affect running workloads (i.e running Pods).
+
+  Rolling out the feature however may cause group permissions to be applied correctly which weren't applied before.
+
+* **What specific metrics should inform a rollback?**
+  If after enabling this feature a spike in `storage_operation_status_count{operation_name="volume_mount", status="fail-unknown"}` metric is observed
+  then cluster admin should look into identifying root cause and rolling back the feature.
+
+* **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
+  Describe manual testing that was done and the outcomes.
+  Longer term, we may want to require automated upgrade/rollback tests, but we
+  are missing a bunch of machinery and tooling and can't do that now.
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs,
+fields of API types, flags, etc.?**
+  Even if applying deprecation policies, they may still surprise some users.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
+  We are going to split the metric that captures mount and permission timings. The full details are available in - https://github.com/kubernetes/kubernetes/issues/98667
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+  - [ ] Other (treat as last resort)
+    - Details:
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  At a high level, this usually will be in the form of "high percentile of SLI
+  per day <= X". It's impossible to provide comprehensive guidance, but at the very
+  high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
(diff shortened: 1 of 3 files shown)


keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+## Proposal
+
+We are updating CSI specs by adding additional field called `volume_mount_group` to `NodeStage` and `NodePublish` RPC calls. The CSI proposal is available at - https://github.com/container-storage-interface/spec/pull/468 .
+
+The CSI spec change is deliberately trying to avoid asking drivers to use supplied `fsGroup` as a generic handle for ownership and permissions. The reason being - Kubernetes may expect ownership and permissions to be in a way that is very platform/OS specific. We do not think CSI driver is right place to enforce all kind of different permissions expected by Kubernetes. The full scope of that discussion is out of scope for this enhancement and interested folks can follow along on - https://github.com/container-storage-interface/spec/issues/449
+
+
+### Risks and Mitigations
+
+I am not aware of any associated risks. If a driver can not support using `fsgroup` as a mount option, it can always use `FileFSGroupPolicy` and let kubelet handle the ownership and permissions.
```
jsafrane, line 72: Please elaborate somewhere that there is a behavioral change: if two pods use the same volume with different `fsGroup`, then:
* Before this change, the pod that starts as second will `chown` all files, so the first pod probably can't access them.
* After this change, both pods may access the volume, if the volume can be mounted twice with different options.

This change is neligible and most probably corresponds to what in-tree Azure File plugin already does today.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+[`prod-readiness-approvers`](http://git.k8s.io/enhancements/OWNERS_ALIASES)
+team. Please reach out on the
+[#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) channel if
+you need any help or guidance.
+
+-->
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
```
jsafrane, line 239: Please file this section, it's required for alpha

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Currently for most volume plugins kubelet applies fsgroup ownership and permission based changes by recursively `chown`ing and `chmod`ing the files and directories inside a volume. For certain CSI drivers this may not be possible because `chown` and `chmod` are unix primitives and underlying CSI driver
+may not support them. This enhancement proposes providing the CSI driver with fsgroup as an explicit field so as CSI driver can apply this on mount time.
+
+## Motivation
+
+Since some CSI drivers(AzureFile for example) - don't support chmod/chown - we propose that fsgroup of pod to be provided to the CSI driver on `NodeStage`
```
jsafrane, line 51: I would emphasize that this feature is a prerequisite for CSI migration of AzureFile volume plugin and removal of Azure cloud provider

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+When Kubelet sees `CSIDriver.Spec.FSGroupPolicy` value set to `Mount` - it would use proposed CSI field `volume_mount_group` to pass pod's `fsGroup` to the CSI driver. Kubelet will expect that driver will use
+this field for mounting volume with given `fsGroup` and no further permission/ownerhip change will be necessary.
+
+### Test Plan
+
+Unit test:
+1. Test that whenever supported pod's `fsGroup` should be passed to CSI driver via `volume_mount_group` field.
+
+For alpha feature:
+1. Run manual tests against azurefile CSI driver.
```
jsafrane, line 97: The driver needs to be updated too.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+### Graduation Criteria
+
+<!--
+**Note:** *Not required until targeted at a release.*
```
jsafrane, line 124: Please file this section

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: No before this change - no chown/chmod will be performed at all and hence except the root user(or pods that run as root),  no pod will be able to read/write to the volume. 

I am not sure if this is a behaviour change. Before - only root could read/write to the pod and now other users can too.  I could add something here but I am not sure if it fits under risks and mitigations.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,451 @@
...
+    ReadWriteOnceWithFSTypeFSGroupPolicy FSGroupPolicy = "ReadWriteOnceWithFSType"
+    FileFSGroupPolicy FSGroupPolicy = "File"
+    NoneFSGroupPolicy FSGroupPolicy = "None"
+    // Indicates that the CSI driver supports applying fsgroup via mount option
+    MountFSGroupPolicy FSGroupPolicy = "Mount"
+)
+```
+
+When Kubelet sees `CSIDriver.Spec.FSGroupPolicy` value set to `Mount` - it would use proposed CSI field `volume_mount_group` to pass pod's `fsGroup` to the CSI driver. Kubelet will expect that driver will use
+this field for mounting volume with given `fsGroup` and no further permission/ownerhip change will be necessary.
```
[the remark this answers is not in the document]
gnufied: I removed this section. I thought long and hard about keeping `FSGroupPolicy` field because initially we weren't sure how CSI spec change will look like.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: updated.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: fixed

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: fixed

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: kinda fixed. PTAL

gnufied:
cc @andyzhangx

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+  Enabling this feature on its own does not changes any existing behaviour but if
+  selected CSI driver advertise `VOLUME_MOUNT_GROUP` node capability, it will result in
+  volume being mounted with pod's `fsGroup`.
```
deads2k, line 205: who sets this value?  End-user, correct?

How does an end user know if his setting took effect?

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  Yes - feature gate can be disabled once enabled and kubelet will fallback to its current behavior, which will
+  result in any CSI driver that requires using `fsGroup` as a mount option to not use the mount option. In which case
+  mounted volumes will be unwritable by any pods other than those running as root.
+
+* **What happens if we reenable the feature if it was previously rolled back?**
+  It will cause kubelet to use `volume_mount_group` field of CSI whenver applicable as discussed in above design.
```
deads2k, line 214: would pods would have to be recreated or woudl a crashlooping pod suddenly start working again?

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+* **Are there any tests for feature enablement/disablement?**
+  Not yet.
+
+### Rollout, Upgrade and Rollback Planning
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can a rollout fail? Can it impact already running workloads?**
```
deads2k, line 223: This question seems important for this feature. Does the scheduler have any way to know whether a particular node will honor this part of the podspec?

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+  Try to be as paranoid as possible - e.g., what if some components will restart
+   mid-rollout?
+
+* **What specific metrics should inform a rollback?**
```
deads2k, line 227: I think you'll want to list the metrics that you use for determining successful and failed mounts here.  And probably count cases where the pod's intent wasn't fulfilled.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+* **Is the rollout accompanied by any deprecations and/or removals of features, APIs,
+fields of API types, flags, etc.?**
+  Even if applying deprecation policies, they may still surprise some users.
+
+### Monitoring Requirements
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **How can an operator determine if the feature is in use by workloads?**
```
deads2k, line 242: see my metric comment above.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+* **Are there any missing metrics that would be useful to have to improve observability
+of this feature?**
+  Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+  implementation difficulties, etc.).
+
+### Dependencies
+
+_This section must be completed when targeting beta graduation to a release._
+
+* **Does this feature depend on any specific services running in the cluster?**
```
deads2k, line 274: If I have understood correctly, this relies on
1. the user to set the value in the pod
2. the kubelet to call the options with the featuregate set
3. the CSI driver to support the  option.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: yes `fsGroup` is being set by the enduser in their pods. The only way for a enduser to know if setting took effect is by execing into pod and checking permissions of mounted volume (or have some healthcheck script which does same).

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: Yes pods have to be re-created for the setting to take effect.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: I do not think scheduler has to know and it can be pretty expensive(code and complexity perspective) to bubble this up to scheduler.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Currently for most volume plugins kubelet applies fsgroup ownership and permission based changes by recursively `chown`ing and `chmod`ing the files and directories inside a volume. For certain CSI drivers this may not be possible because `chown` and `chmod` are unix primitives and underlying CSI driver
+may not support them. This enhancement proposes providing the CSI driver with fsgroup as an explicit field so as CSI driver can apply this on mount time.
+
+## Motivation
+
+Since some CSI drivers(AzureFile for example) - don't support chmod/chown - we propose that fsgroup of pod to be provided to the CSI driver on `NodeStage`
```
xing-yang, line 53: NodeStage -> NodeStageVolume

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+and `NodePublish` CSI RPC calls. This will allow the CSI driver to apply `fsgroup` as a mount option on `NodeStage` or `NodePublish` and kubelet can be freed
```
xing-yang, line 54: NodePublish -> NodePublishVolume

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+- Allow CSI driver to mount volumes with provided fsgroup.
+
+### Non-Goals
+
+- We are not supplying `fsGroup` as a generic ownerhip and permission handle to the CSI driver. We do not expect CSI drivers to `chown` or `chmod` files.
+
+## Proposal
+
+We are updating CSI specs by adding additional field called `volume_mount_group` to `NodeStage` and `NodePublish` RPC calls. The CSI proposal is available at - https://github.com/container-storage-interface/spec/pull/468 .
```
xing-yang, line 69: `NodeStage` and `NodePublish` -> `NodeStageVolume` and `NodePublishVolume`

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+### Risks and Mitigations
+
+I am not aware of any associated risks. If a driver can not support using `fsgroup` as a mount option, it can always use `FileFSGroupPolicy` and let kubelet handle the ownership and permissions.
+
+## Design Details
+
+We are proposing that when kubelet determines a CSI driver has `VOLUME_MOUNT_GROUP` node capability, the kubelet will use proposed CSI field `volume_mount_group` to pass pod's `fsGroup` to the CSI driver. Kubelet will expect that driver will use
+this field for mounting volume with given `fsGroup` and no further permission/ownerhip change will be necessary.
+
+It should be noted that if a CSI driver advertises `VOLUME_MOUNT_GROUP` node capability then value defined in `CSIDriver.Spec.FSGroupPolicy` will be ignored and kubelet willl always use `fsGroup` as a mount option.
```
xing-yang, line 83: typo: willl -> will

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
```
xing-yang, line 84: If a CSI driver does not make any changes proposed in this KEP, will there be any behavior change?  Please clarify that in the KEP.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+### Test Plan
+
+Unit test:
+1. Test that whenever supported pod's `fsGroup` should be passed to CSI driver via `volume_mount_group` field.
+
+For alpha feature:
+1. Update Azure File CSI driver to support supplying `fsGroup` via `NodeStage` and `NodePublish`.
```
xing-yang, line 91: There should be tests for non-Azure File CSI driver to make sure there are no regressions.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+1. Run manual tests against azurefile CSI driver.
+
+For beta:
+1. E2E tests that verify volume readability/writability using azurefile CSI driver.
+2. E2E tests using CSI mock driver.
```
xing-yang, line 96: Should there be E2E tests for CSI hostpath driver and GCE PD driver as well?

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+#### Alpha -> Beta Graduation
+
+- Since this feature is a must-have for azurefile CSI migration, we will perform testing of the driver.
+- Currently CSI spec change is being introduced as alpha change and we will work to move the API change in CSI spec to stable.
+
+#### Beta -> GA Graduation
+
+- CSI spec change should be stable.
+- Tested via e2e and manually using azurefile CSI driver.
```
xing-yang, line 126: Testing no regressions are introduced in non-azurefile CSI driver?

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,400 @@
...
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+### Graduation Criteria
+
+#### Alpha -> Beta Graduation
+
+- Since this feature is a must-have for azurefile CSI migration, we will perform testing of the driver.
```
xing-yang, line 120: Testing no regressions are introduced in non-azurefile CSI driver?

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: it is somewhat covered in section - "* **Does enabling the feature change any default behavior?**"

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: There are already tests that verify fsgroup behaviour for existing drivers.  I do not think we need specifically new tests for them - I think.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: I do not think so. As I mentioned above - there are already fsgroup based tests for those drivers.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: same as above. we already have fsgroup tests that verify behaviour of drivers that don't have this new capability.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: same as above

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: fixed

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: Do we need these sections for alpha feature? The PRR check says:

> _This section must be completed when targeting beta graduation to a release._

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: same as above.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: Updated this section. However kubelet never tracked whether pod's intent was fulfilled and I think such a metric is out of scope for this specific change because such a metric will apply to generic mechanism by which `fsGroup` is applied to the volumes (and they are in-use since k8s <1.2 or something)

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: Updated.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: added

gnufied:
@deads2k @xing-yang I have addressed  the review comments. PTAL. thanks.

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: > yes `fsGroup` is being set by the enduser in their pods. The only way for a end-user to know if setting took effect is by execing into pod and checking permissions of mounted volume (or have some healthcheck script which does same).

This seems like the sort of thing that users would have difficulty understanding the failure mode for.  Do you have recommended scripts or something to recommend to a user that wants to be sure his files are correct?

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: > Yes pods have to be re-created for the setting to take effect.

heh, don't tell me, write it in the doc :)

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: Also, can we get this updated in the doc?  I won't be the only admin and/or user wondering.

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: > I do not think scheduler has to know and it can be pretty expensive(code and complexity perspective) to bubble this up to scheduler.

So that means that mid-rollout (either turning it on or turning it off) can result in different pods of the same replicaset/deployment failing or succeeding in seemingly random ways?  I think that's worth noting in an answer to this question.

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: > Updated this section. However kubelet never tracked whether pod's intent was fulfilled and I think such a metric is out of scope for this specific change because such a metric will apply to generic mechanism by which `fsGroup` is applied to the volumes (and they are in-use since k8s <1.2 or something)

I can see an argument for not requiring it for alpha, but I don't think I follow the reasoning that because there was no metric before there shouldn't be a metric as this feature moves to beta.  Without a metric, how can you know that
1. the feature is in use
2. the feature is working.

keps/sig-storage/2317-fsgroup-on-mount/README.md
deads2k: > added

I'm not seeing it in this sectin.

gnufied: It is there:

```
This feature depends on presence of `fsGroup` field in Pod and CSI driver having `VOLUME_MOUNT_GROUP`
  capability.

  - [CSI driver]
    - CSI driver must have `VOLUME_MOUNT_GROUP` capability:
      - If this capability is not available in CSI driver then kubelet will try to use default mechanism of applying `fsGroup` to volume (which is basically `chown` and `chomid`). If underlying driver however does not support applying group permissions via `chown` and `chmod` then Pods will not run correctly.
      - Workloads may not run correctly or volume has to be mounted with permissions `777`.

```

keps/sig-storage/2317-fsgroup-on-mount/README.md
xing-yang: Can you add a link to the existing e2e tests https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/fsgroupchangepolicy.go to this KEP here?

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: hmm - if the replicaset was say working in 1.20 - then user must be running the pods as root or mounting the volume as world-writable (`777`). So this feature won't affect that. Even if `fsGroup` is applied correct to the volume, it won't affect anything that is running as root (because those pods can write to any volume anyways) or volumes with 777 permissions (because those volumes can be written by anyone anyways).

So in a nutshell - it does not have any impact of already correctly running workloads. Now if you somehow had a pod in 1.20 - which was not running as root and volumes weren't mounted with 777 permissions then those would suddenly start working. But I do not think that is the use case this section documents.

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,393 @@
...
+
+* **How can an operator determine if the feature is in use by workloads?**
+  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+  checking if there are objects with field X set) may be a last resort. Avoid
+  logs or events for this purpose.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
```
xing-yang, line 250: Can you fill out this section?

gnufied: This section is only needed for beta features as far as I can tell. We are going to do some work to split some of the metrics - https://github.com/kubernetes/kubernetes/issues/98667 . The SLIs could be determined in next release I think (when we know more about real world usage).

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: done

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: I have updated the doc about what admin/pod owner should expect here.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: sure.  Added more information about metrics and filed https://github.com/kubernetes/kubernetes/issues/98667 to capture how `fsGroup` intent was full-filled.

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: done.

deads2k:
/approve

PRR

keps/sig-storage/2317-fsgroup-on-mount/README.md
```diff
@@ -0,0 +1,384 @@
...
+      - Kubelet
+  - [ ] Other
+    - Describe the mechanism:
+    - Will enabling / disabling the feature require downtime of the control
+      plane?
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+  Enabling this feature on its own does not changes any existing behaviour but if
```
xing-yang, line 189: nit: does not changes -> does not change

xing-yang:
LGTM

keps/sig-storage/2317-fsgroup-on-mount/README.md
jsafrane: A CSI driver could have supported `chown` / `chmod` for RWO / RWX volumes (such as github.com/kubernetes-csi/csi-driver-smb, if the samba export is configured properly), it's just faster with a mount option. Users of this driver could experience a very small behavior change.

gnufied: addressed under the section - `Does enabling the feature change any default behavior?**`

keps/sig-storage/2317-fsgroup-on-mount/README.md
gnufied: this line is replaced entirely now.

Issue #2324, opened 2021-01-25
Allow transfer of VolumeSnapshot resources between namespaces


huffmanca:
# Enhancement Description
- One-line enhancement description (can be used as a release note): Allow transfer of VolumeSnapshots between namespaces.
- Kubernetes Enhancement Proposal: 
- Discussion Link: 
- Primary contact (assignee): @huffmanca 
- Responsible SIGs: sig-storage
- Enhancement target (which target equals to which milestone):
  - Alpha release target (1.22): 
  - Beta release target (x.y):
  - Stable release target (x.y):
- [ ] Alpha
  - [ ] KEP (`k/enhancements`) update PR(s):
  - [ ] Code (`k/k`) update PR(s):
  - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #2326, opened 2021-01-25
Add proposal for transfer of VolumeSnapshot

huffmanca:
xref #2324 

/sig storage


keps/prod-readiness/sig-storage/2324.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 1412
+stable:
```
johnbelamaric, line 2: should be `alpha`.

you can assign me to this one, though it looks like you are targeting 1.22?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+- As long as we reference the existing VolumeSnapshotContent, there should be no issues with encrypted VolumeSnapshotContents. The secret reference is stored on the VolumeSnapshotContent and VolumeSnapshotClass, not the VolumeSnapshot itself. If we take a different approach, such as re-creating the VolumeSnapshotContent, then we will need to ensure these credentials are propagated to the new content.
+- The VolumeSnapshot must be `readyToUse` before the transfer begins.
+- Any owners of the destination namespace must accept the transfer. By using the StorageTransfer API, this ensures an owner in the destination namespace is aware of the transfer.
+
+## Design Details
+
+Refer to [API for PVC Namespace Transfer](https://github.com/kubernetes/enhancements/pull/2110/) for 
+details concerning the StorageTransfer API.
+
+A `StorageTransferRequest` resource will be created in the source namespace:
```
gnufied, line 102: This text says a `StorageTransferRequest` will be created in source namespace but namespace of object in yaml below is `target-namespace`

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+  targetName: bar
+```
+
+`StorageTransferApproval` resource will be created in the target namespace:
+
+```yaml
+apiVersion: v1alpha1
+kind: StorageTransferApproval
+metadata:
+  name: approve-foo
```
gnufied, line 125: Same as above. The text says approval will be created in target namespace but namespace of this object is source-namespace.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+kind: StorageTransferRequest
+metadata:
+  name: transfer-foo
+  namespace: target-namespace
+spec:
+  source:
+    name: foo
+    namespace: source-namespace
+    kind: VolumeSnapshot
+  approvalName: approve-foo
```
gnufied, line 115: How does a transfer request know `approvalName`? does this gets filled later on by snapshot controller?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+  namespace: source-namespace
+spec:
+  source:
+    name: foo
+    kind: VolumeSnapshot
+  targetNamespace: target-namespace
+  requestName: transfer-foo
+```
+
+ When matching `StorageTransferRequest` and `StorageTransferApproval` resources are detected, the
```
gnufied, line 135: We should describe what the matching criteria is.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+spec:
+  source:
+    name: foo
+    namespace: source-namespace
+    kind: VolumeSnapshot
+  approvalName: approve-foo
+  targetName: bar
+```
+
+`StorageTransferApproval` resource will be created in the target namespace:
```
qinpingli, line 119: @huffmanca Could you explain more about the `StorageTransferApproval` and `StorageTransferRequest`, do they need to be created manually?
If yes, IMO, it's a little complicated to use it.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+
+```yaml
+apiVersion: v1alpha1
+kind: StorageTransferRequest
+metadata:
+  name: transfer-foo
+  namespace: target-namespace
+spec:
+  source:
+    name: foo
```
qinpingli, line 112: For the user he/she can access the target namespace, but can not access the source namespace, how he/she can get the volumesnapshot object he/she wants to transfer?

jsafrane: In the proposal it says *"provided that both the source and target namespace have a user that has agreed to the transfer."* By creating objects both in the source and target namespace we ensure this agreement actually happened. So users can't steal snapshots from other namespaces and at the same time they can't "push" their snapshots to namespaces that don't want it.

xing-yang:
@huffmanca can you change "VolumeSnapshot" in the title to "PVC or VolumeSnapshot"?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
jsafrane: IMO it's part of the agreement. User who gives the snapshots must pass this name to the user who accepts the snapshots, outside of Kubernetes API.

jsafrane: BTW, I'm checking OpenStack, which has very similar concept called [volume transfer between projects](https://docs.openstack.org/cinder/latest/cli/cli-manage-volumes.html#transfer-a-volume), and here OpenStack actually generates a token, that the current owner of the volume (old project) must give to the new owner (new project).

The KEP should note that the name should be at least something that the users can't guess. Using a random string in the example would be helpful.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,485 @@
...
+A `StorageTransferRequest` resource will be created in the source namespace:
+
+```yaml
+apiVersion: v1alpha1
+kind: StorageTransferRequest
+metadata:
+  name: transfer-foo
+  namespace: target-namespace
+spec:
+  source:
```
jsafrane, line 111: If it's really in the source namespace, you can use `TypedLocalObjectReference` with `APIGroup`, `Kind` and `Name`. The namespace is implied from the namespace of `StorageTransferRequest`

huffmanca: Couldn't we also use an `ObjectReference` in both the request and approval, and ensure that these match?

j-griffith:
Looking between #2110 and this to match things up and was thinking, maybe it would make sense to consolidate to a single KEP that proposes a "csi transfer" feature.  That way efforts can be focused on a single KEP and controller.  Initial thoughts on this it seems like it might be feasible.

keps/prod-readiness/sig-storage/2324.yaml
huffmanca: Updated, and this has been assigned.

Correct. The hope is that we can have a working design this quarter, and alpha implementation will be targeting 1.22.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: This has been updated. The `StorageTransferRequest` will be in the target namespace, and the `StorageTransferApproval` will be created in the source namespace.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: This has been updated. The `StorageTransferRequest` will be in the target namespace, and the `StorageTransferApproval` will be created in the source namespace.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: The matching criteria has been included, and was updated based on the meeting today. As of now the `StorageTransferApproval` indicates the storage source, and a match is determined only between the `StorageTransferRequest` and `StorageTransferApproval` values. It no longer validates that the source matches between the two.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: @qinpingli,

The current design requires that these objects be manually created. It's possible that users could script creation of them if they plan to transfer resources en masse; however, we intentionally need a user from each namespace to "sign off" on the transfer.

huffmanca:
@j-griffith ,

I've updated this KEP, but this is attempting to continue the work that you and mhenricks have been leading, since the most recent KEP has stalled. The goal would be to have a generic (for storage) API that can be used for both PVCs and VolumeSnapshots, at least to start with.

I agree that we're looking to create a single controller for both.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+  namespace: source-namespace
+spec:
+  source:
+    name: foo
+    kind: VolumeSnapshot
+  secrets:
+    - name: secret-name
+      namespace: secret-namespace
+      type: secret-type # such as `csi.storage.k8s.io/provisioner-secret-name`
+  targetNamespace: target-namespace
```
j-griffith, line 159: The transfer request (or grant if you go that route) should be enough to handle some of these details.  If you operate on the request/grant object you can eliminate things like targetNamespace, name and kind and just pull it from the actual transfer object maybe?  

So something like:
```
spec:
  storageTransferID: xxxxx
  transferToken: xxx
.....
```

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+kind: StorageTransferRequest
+metadata:
+  name: kAnUOwpzso
+  namespace: target-namespace
+spec:
+  approvalName: xyji6OwynW
+  targetName: bar
+```
+
+`StorageTransferApproval` resource will be created in the source namespace:
```
j-griffith, line 143: target namespace?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+
+```yaml
+apiVersion: storage.k8s.io/v1alpha1
+kind: StorageTransferApproval
+metadata:
+  name: xyji6OwynW
+  namespace: source-namespace
```
j-griffith, line 150: assuming this was meant to be "target" namespace?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+    kind: VolumeSnapshot
+  secrets:
+    - name: secret-name
+      namespace: secret-namespace
+      type: secret-type # such as `csi.storage.k8s.io/provisioner-secret-name`
+  targetNamespace: target-namespace
+  requestName: kAnUOwpzso
+```
+
+NOTE: The name of the `StorageTransferApproval` and `StorageTransferRequest` should not be easy to guess,
```
j-griffith, line 163: Should this matter?  The design should not allow things like "guessing" of the transfer name to allow unintended transfer. Obscurity isn't a great defense, but in the model if we're doing a grant/accept model there shouldn't be much room for abuse, especially if you throw in the possibility of requiring a token .

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+or backups in separate namespaces.
+
+#### Story 1
+
+The `prod` namespace contains a PVC that holds user information. VolumeSnapshots are taken of 
+this PVC on a regular basis.
+
+An existing VolumeSnapshot can be transferred to a test namespace. Once transferred, a PVC
+can be created from the snapshot for testing purposes.  
+
```
j-griffith, line 90: Since I think I might be confused on the request/approval ownership thing maybe it would be good to add that detail to this user story.  In addition to "wants to xfr from prod to test" outline what actions they take to kick the process off.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+- Create the PVC target-namespace\bar by copying the spec of source-namespace\foo and setting spec.volumeName appropriately.
+- Reset the PersistentVolume reclaim policy
+- Remove the Finalizer from the PVC source-namespace\foo
+
+For VolumeSnapshots, this performs the following:
+- Creates a VolumeSnapshot in the target namespace.
+- Binds the VolumeSnapshotContent to the newly created VolumeSnapshot. If this was a dynamically provisioned VolumeSnapshotContent, then it adjusts the VolumeSnapshotContent.Spec.Source to point to the newly created
+VolumeSnapshot.
+- Deletes the source VolumeSnapshot.
+
```
j-griffith, line 188: "transfering" and relabeling snapshots has always been something that made me kinda nervous.  For a number of storage backends that use things like cow type snapshots or linked reference to versions this introduces a pretty big problem.  The problem is that you could now have a snapshot in ns: foo that's actually lined to a snapshot in ns: bar.  The two namespaces aren't aware of this linkage on the back end and neither can delete resources.

I know some have proposed that backends handle the reference counting in this case and do a sort of 'soft delete' but that doesn't seem great because "somebody" is paying for that storage that they *think* they deleted.

To transfer snapshots I don't have a great alternative proposal, other than create a PVC from a snapshot and transfer that instead.  I'm probably overly paranoid about this though and I don't think there are any other folks that have this concern.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+- As long as we reference the existing VolumeSnapshotContent, there should be no issues with encrypted VolumeSnapshotContents. The secret reference is stored on the VolumeSnapshotContent and VolumeSnapshotClass, not the VolumeSnapshot itself. If we take a different approach, such as re-creating the VolumeSnapshotContent, then we will need to ensure these credentials are propagated to the new content.
+- The VolumeSnapshot must be `readyToUse` before the transfer begins.
+- Any owners of the destination namespace must accept the transfer. By using the StorageTransfer API, this ensures an owner in the destination namespace is aware of the transfer.
+
+## Design Details
+
+This continues the discussion from [API for PVC Namespace Transfer](https://github.com/kubernetes/enhancements/pull/2110/) 
+in regards to the StorageTransfer API.
+
+A `StorageTransferRequest` resource will be created in the target namespace:
```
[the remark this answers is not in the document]
huffmanca: I kept `StorageTransferRequest`, but adjusted `StorageTransferApproval` to `StorageTransferRequest`.

Originally the request would be created in the target namespace, and then the user in the source namespace could approve the request. From your comments below I've adjusted so that the request is created in the source namespace, and then the acceptance must come from the target namespace.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,538 @@
...
+```yaml
+apiVersion: storage.k8s.io/v1alpha1
+kind: StorageTransferRequest
+metadata:
+  name: kAnUOwpzso
+  namespace: target-namespace
+spec:
+  approvalName: xyji6OwynW
+  targetName: bar
+```
```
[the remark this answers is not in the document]
huffmanca: I've included this in both resources' spec. I think it best if the names were human readable, the token required, and automatically generated.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: Originally the request would be created in the target namespace, and then the user in the source namespace could approve the request.

I've adjusted these to be in the `StorageTransferRequest` and removed from the accept.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: I've included a token here and removed this note.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
huffmanca: This detail has been included in the user story.

j-griffith:
Thanks @huffmanca for the clarification on some of those points.  If #2110 is being combined here a lot of that content should probably be merged in here.  The one other thing I was getting at was the level of detail around the workflow here https://github.com/kubernetes/enhancements/pull/2110/files#diff-82617246992480a1d7acd7c0041813170f4e01cb996df54b3530e2db74ed9bcfR95 was really useful.  Other than that I'm obviously in favor of the feature :)  Really happy to see you've picked up the effort.

mkimuram:
@huffmanca @j-griffith @gnufied @jsafrane 

I'm thinking about the other approach to achieve this feature.
The idea is to add fields to PVC, instead of introducing new `StorageTransferRequest` API and `StorageTransferApproval` API.

I believe that this simplifies the Spec and avoid declaring the desired state of a certain resource in different multiple places. 

What do you think of it?

Details are as follows:

<details>

The fields to be added are `transfer.source` and `transfer.destination` in PVC spec. Also, some types need to be added to `status.conditions.type`.

```
spec:
  transfer:
    source:
      namespace: ""
      name: ""
    destination:
      namespace: ""
      name: ""
status: 
  conditions: 
    type: add [transferring|receiving|transferred]
```

For example, let's assume that user A has PVC `pvc-a` in namespace `ns-a` and would like to transfer it to PVC `pvc-b` in namespace `ns-b` that user B uses. 

User operations will be:
- User A: modify `pvc-a` to set `spec.transfer.destination.namespace`=`ns-b` and `spec.transfer.destination.name`=`pvc-b` (And may delete `pvc-a` later, after it is transferred).
- User B: create `pvc-b` with setting `spec.transfer.source.namespace`=`ns-a` and `spec.transfer.source.name`=`pvc-a`.

Controller behaviors will be:
- If `pvc.status.phase` == `"Bound"` && no pod is using this PVC && `spec.transfer.destination` is set: add `"transferring"` to `pvc.status.conditions.type`,
- If `pvc.status.conditions.type` has `"transferring"`: Avoid consuming this PVC by pods, 
- If `pvc.status.phase` == `"Pending"` && `spec.transfer.destination` is set && `pvc.status.conditions.type` doesn't have `"receiving"`: add `"receiving"` to `pvc.status.conditions.type`,
- If `pvc.status.phase` == `"Pending"` && `spec.transfer.destination` is set && `pvc.status.conditions.type` has `"receiving"`: check if the PVC with `spec.transfer.source.namespace` and `spec.transfer.source.name` has `pvc.status.conditions.type` `"transferring"`: 
  - then: Swap the bidirectional reference of PVC and PV for source PVC and destination PVC (Secrets in PV should be modified here, by using information from StorageClass referenced from destination PVC. Also, validations for the compatibility of PVCs need to be done before start swapping).
    - If succeeded: 
       - source PVC: Change `pvc.status.phase` to `"Lost"` and change `pvc.status.conditions.type` from `"transferring"` to `"transferred"`,
       - destination PVC: Change `pvc.status.phase` to `"Bound"` and remove `"receiving"` from `pvc.status.conditions.type`.
  - else: do nothing (Wait for destination PVC become `"transferring"`).

Above is just a rough idea that doesn't consider cancel or error cases. Also, it doesn't cover snapshot. However, we will be able to consider more in detail, once we agree that this approach looks good (I hope that this approach can be easily applied to snapshot, like adding similar fields in snapshot and just inherit them to PVCs).
</details>

mkimuram:
Implemented [PoC codes](https://github.com/mkimuram/kubernetes/tree/transfer) for further discussion.
Normal case works as expected. 

Console logs are as follows:

<details>

```
kubectl create ns ns-a
kubectl create ns ns-b

cat << EOF | kubectl create -n ns-a -f - 
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-a
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
EOF

cluster/kubectl.sh describe -n ns-a pvc pvc-a
Name:          pvc-a
Namespace:     ns-a
StorageClass:  standard
Status:        Bound
Volume:        pvc-18304fee-908d-4a59-8587-2b3d64166923
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/host-path
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      5Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Used By:       <none>
Events:
  Type    Reason                 Age   From                         Message
  ----    ------                 ----  ----                         -------
  Normal  ProvisioningSucceeded  6s    persistentvolume-controller  Successfully provisioned volume pvc-18304fee-908d-4a59-8587-2b3d64166923 using kubernetes.io/host-path

kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM        STORAGECLASS   REASON   AGE
pvc-18304fee-908d-4a59-8587-2b3d64166923   5Gi        RWO            Delete           Bound    ns-a/pvc-a   standard                20s
```
- User A: modify `pvc-a` to set `spec.transfer.destination.namespace`=`ns-b` and `spec.transfer.destination.name`=`pvc-b`

```
kubectl patch -n ns-a pvc pvc-a -p '{"spec":{"transfer": {"destination": {"namespace": "ns-b", "name": "pvc-b"}}}}'

cluster/kubectl.sh describe -n ns-a pvc pvc-a
Name:          pvc-a
Namespace:     ns-a
StorageClass:  standard
Status:        Bound
Volume:        pvc-18304fee-908d-4a59-8587-2b3d64166923
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/host-path
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      5Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Transfer:
  Destination:
    Namespace:  ns-b
    Name:       pvc-b
Used By:        <none>
Conditions:
  Type           Status  LastProbeTime                     LastTransitionTime                Reason  Message
  ----           ------  -----------------                 ------------------                ------  -------
  Transferring   True    Mon, 01 Jan 0001 00:00:00 +0000   Mon, 03 May 2021 14:51:33 +0000           Waiting for receiver PVC to receive the bound PV and block consuming this PVC.
Events:
  Type    Reason                 Age   From                         Message
  ----    ------                 ----  ----                         -------
  Normal  ProvisioningSucceeded  31s   persistentvolume-controller  Successfully provisioned volume pvc-18304fee-908d-4a59-8587-2b3d64166923 using kubernetes.io/host-path
```

- User B: create `pvc-b` with setting `spec.transfer.source.namespace`=`ns-a` and `spec.transfer.source.name`=`pvc-a`.

```
cat << EOF | kubectl create -n ns-b -f - 
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-b
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  transfer:
    source:
      namespace: ns-a
      name: pvc-a
EOF

cluster/kubectl.sh describe -n ns-b pvc pvc-b
Name:          pvc-b
Namespace:     ns-b
StorageClass:  standard
Status:        Pending
Volume:        pvc-18304fee-908d-4a59-8587-2b3d64166923
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      0
Access Modes:  
VolumeMode:    Filesystem
Transfer:
  Source:
    Namespace:  ns-a
    Name:       pvc-a
Used By:        <none>
Conditions:
  Type        Status  LastProbeTime                     LastTransitionTime                Reason  Message
  ----        ------  -----------------                 ------------------                ------  -------
  Receiving   True    Mon, 01 Jan 0001 00:00:00 +0000   Mon, 03 May 2021 14:52:15 +0000           Receiving transferred volume.
Events:       <none>

cluster/kubectl.sh describe -n ns-b pvc pvc-b
Name:          pvc-b
Namespace:     ns-b
StorageClass:  standard
Status:        Bound
Volume:        pvc-18304fee-908d-4a59-8587-2b3d64166923
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      5Gi
Access Modes:  RWO
VolumeMode:    Filesystem
Transfer:
  Source:
    Namespace:  ns-a
    Name:       pvc-a
Used By:        <none>
Events:         <none>
```

- User A:  (And may delete `pvc-a` later, after it is transferred).

```
cluster/kubectl.sh describe -n ns-a pvc pvc-a
Name:          pvc-a
Namespace:     ns-a
StorageClass:  standard
Status:        Lost
Volume:        
Labels:        <none>
Annotations:   pv.kubernetes.io/bind-completed: yes
               pv.kubernetes.io/bound-by-controller: yes
               volume.beta.kubernetes.io/storage-provisioner: kubernetes.io/host-path
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:      
Access Modes:  
VolumeMode:    Filesystem
Transfer:
  Destination:
    Namespace:  ns-b
    Name:       pvc-b
Used By:        <none>
Conditions:
  Type          Status  LastProbeTime                     LastTransitionTime                Reason  Message
  ----          ------  -----------------                 ------------------                ------  -------
  Transferred   True    Mon, 01 Jan 0001 00:00:00 +0000   Mon, 03 May 2021 14:52:54 +0000           Transferring volume has been completed.
Events:
  Type    Reason                 Age    From                         Message
  ----    ------                 ----   ----                         -------
  Normal  ProvisioningSucceeded  2m12s  persistentvolume-controller  Successfully provisioned volume pvc-18304fee-908d-4a59-8587-2b3d64166923 using kubernetes.io/host-path
  Normal  ClaimLost              25s    persistentvolume-controller  Bound PersistentVolume has been transfered. This PersistentVolumeClaim can safely be deleted.

kubectl get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM        STORAGECLASS   REASON   AGE
pvc-18304fee-908d-4a59-8587-2b3d64166923   5Gi        RWO            Delete           Bound    ns-b/pvc-b   standard                2m38s
```
</details>

mkimuram:
@saad-ali

I read through the previous discussions and noticed that I'm just trying to roll back the discussion [here](https://github.com/kubernetes/enhancements/pull/643#issuecomment-487747601) by my previous two comments.

Do we really need to introduce new generic CRDs for transfer, instead of just adding fields for transfer to specific APIs that support this feature? 

I think:
- From a user experience perspective, it would be easier to learn less APIs, unless the spec differs significantly between APIs. Also, users would expect that they can directly change a resource with the API for the resource.
- From a semantic perspective, PersistentVolumeClaim is already an abstraction that requests a PersistentVolume with a certain specification, therefore it would be the right place to specify which PersistentVolume to bound / unbound from (And this will be also true for VolumeSnapshot/VolumeSnapshotContent).
- We would be able to later introduce new generic transfer APIs that just wrap this implementation, if needed (Actually, we will need to introduce similar fields to PVC, even if we introduce additional CRDs outside, so that we can keep track of the status of PVCs. For example, if there are two `StorageTransferRequest` referencing the same source PVC and different destination PVCs or if `StorageTransferRequest` is deleted while transferring, keeping the status just in `StorageTransferRequest` won't be enough).

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,543 @@
...
+- Delete the PVC source-namespace\foo.
+- Wait for all Pods to stop using the PVC source-namespace\foo
+- Bind The PersistentVolume to target-namespace\bar.
+- Create the PVC target-namespace\bar by copying the spec of source-namespace\foo and setting spec.volumeName appropriately.
+- Reset the PersistentVolume reclaim policy
+- Remove the Finalizer from the PVC source-namespace\foo
+
+For VolumeSnapshots, this performs the following:
+- Creates a VolumeSnapshot in the target namespace.
+- Binds the VolumeSnapshotContent to the newly created VolumeSnapshot. If this was a dynamically provisioned VolumeSnapshotContent, then it adjusts the VolumeSnapshotContent.Spec.Source to point to the newly created
```
xing-yang, line 190: VolumeSnapshotContent.Spec.Source points to a VolumeHandle or a SnapshotHandle.  It does not point to a VolumeSnapshot.  It is VolumeSnapshotContent.Spec.VolumeSnapshotRef that points to a VolumeSnapshot.  Also VolumeSnapshotContent.Spec.VolumeSnapshotRef is immutable so you cannot rebind it.

The suggested approach is to create a new VolumeSnapshotContent, point Source to the existing SnapshotHandle, and point VolumeSnapshotRef to the new VolumeSnapshot.

CC @yuxiangqian

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,543 @@
...
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Currently, VolumeSnapshot resources are tied to a specific namespace. This KEP proposes a method for transferring
+these resources between namespaces, allowing users to conveniently move these resources without manual creation
+of each independent object (VolumeSnapshot, VolumeSnapshotContent).
+
+This extends the [API for PVC Namespace Transfer](https://github.com/kubernetes/enhancements/pull/2110/files) to also apply for VolumeSnapshots.
```
xing-yang, line 54: Looks like this KEP only handles VolumeSnapshot namespace transfer.  The KEP on PVC (https://github.com/kubernetes/enhancements/pull/2110) is closed.  Is there anyone working on PVC namespace transfer?

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,543 @@
...
+- Reset the PersistentVolume reclaim policy
+- Remove the Finalizer from the PVC source-namespace\foo
+
+For VolumeSnapshots, this performs the following:
+- Creates a VolumeSnapshot in the target namespace.
+- Binds the VolumeSnapshotContent to the newly created VolumeSnapshot. If this was a dynamically provisioned VolumeSnapshotContent, then it adjusts the VolumeSnapshotContent.Spec.Source to point to the newly created
+VolumeSnapshot.
+- Deletes the source VolumeSnapshot.
+
+The external-snapshotter will need to be updated to remove immutability of the following fields:
```
xing-yang, line 194: I don't think we should remove the immutability of these fields.  Instead a new VolumeSnapshotContent object should be created.

CC @yuxiangqian

xing-yang:
@mkimuram In your POC, you are proposing to add a Transfer.Source and Transfer.Destination field in the PVC spec.  At the end of the transfer, will a new PVC be created with the name and namespace based on the Transfer.Destination field?

mkimuram:
> At the end of the transfer, will a new PVC be created with the name and namespace based on the Transfer.Destination field?

No. New PVC is manually created by a user who have access to the namespace. It can be regarded as the same action for the receiver to create a `StorageTransferAccept` in the current KEP, from the UX view point. Please click the "Details" in the [previous comment](https://github.com/kubernetes/enhancements/pull/2326#issuecomment-831330042) and see the console log. In the example, User B creates a new PVC.

mkimuram:
@xing-yang 

I hope that this [diagram](https://mermaid-js.github.io/mermaid-live-editor/edit#eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgVXNlciBBXG4gICAgcGFydGljaXBhbnQgUFZDIEEgXG4gICAgcGFydGljaXBhbnQgUFZcbiAgICBwYXJ0aWNpcGFudCBQVkMgQlxuICAgIHBhcnRpY2lwYW50IFVzZXIgQlxuICAgIFVzZXIgQS0-PitQVkMgQTogQ3JlYXRlIFBWQyBBXG4gICAgTm90ZSByaWdodCBvZiBQVkMgQTogUGhhc2U6cGVuZGluZ1xuICAgIFBWQyBBLT4-K1BWIDogYmluZFxuICAgIFBWIC0-Pi1QVkMgQTogYmluZFxuICAgIE5vdGUgcmlnaHQgb2YgUFZDIEE6IFBoYXNlOmJvdW5kXG4gICAgVXNlciBBLT4-K1BWQyBBOiBNb2RpZnkgc3BlYy50cmFuc2Zlci5kZXN0aW5hdGlvbi5uYW1lPVBWQyBCXG4gICAgTm90ZSByaWdodCBvZiBQVkMgQTogUGhhc2U6Ym91bmQgQ29uZGl0aW9uOnRyYW5zZmVycmluZ1xuICAgIFVzZXIgQi0-PitQVkMgQjogQ3JlYXRlIFBWQyBCIHdpdGggc3BlYy50cmFuc2Zlci5zb3VyY2UubmFtZT1QVkMgQVxuICAgIE5vdGUgcmlnaHQgb2YgUFZDIEI6IFBoYXNlOnBlbmRpbmdcbiAgICBOb3RlIHJpZ2h0IG9mIFBWQyBCOiBQaGFzZTpwZW5kaW5nIENvbmRpdGlvbjpyZWNlaXZpbmdcbiAgICBQVkMgQi0-PlBWIDpiaW5kXG4gICAgUFZDIEEtPj5QVjp1bmJpbmRcbiAgICBQVi0-PlBWQyBBOnVuYmluZFxuICAgIE5vdGUgcmlnaHQgb2YgUFZDIEE6IFBoYXNlOmxvc3QgQ29uZGl0aW9uOnRyYW5zZmVycmVkICAgXG4gICAgUFYgLT4-UFZDIEI6IGJpbmRcbiAgICBOb3RlIHJpZ2h0IG9mIFBWQyBCOiBQaGFzZTpib3VuZFxuXG4iLCJtZXJtYWlkIjoie1xuICBcInRoZW1lXCI6IFwiZGVmYXVsdFwiXG59IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ) will help understand my intention. 

Note that while PVC's `condition` is `transferred`, the attach/detach controller is modified to prevent PVC with phase:bound from being consumed. Therefore, it is safe even if the PV is bound to two PVCs. Also, either PVC A or PVC B have reference to the PV at any time and transfer can only happens with two PVCs, no more than two.

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
mhenriks: Hi, based on the tile of this KEP, I closed #2110 but can of course reopen.  I am still very interested in this feature getting into k8s.

xing-yang:
We had a meeting to discuss this today.  Mustafa summarized all previous efforts and the two options.  Meeting notes are here: https://docs.google.com/document/d/17H1k4lqdtJwZSjNRaQue-FhMhyk14JA_MoURpoxha5Q/edit#

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,543 @@
...
+
+ When matching `StorageTransferRequest` and `StorageTransferAccept` resources are detected, the 
+ transfer process is initiated. We have the following matching criteria for the `StorageTransferRequest` 
+ named `r` and the `StorageTransferAccept` `a`:
+ - `r.metadata.name == a.spec.requestName`
+ - `r.metadata.namespace == a.spec.sourceNamespace`
+ - `r.spec.acceptName == a.metadata.name`
+ - `r.spec.token == a.spec.requestToken`
+
+Once matching is complete, a controller will begin the transfer process. For PVCs, this performs the following:
```
mkimuram, line 178: @mhenriks @huffmanca

What happens if
- two `StorageTransferRequest`s are created for the same resource and `StorageTransferAccept` for each `StorageTransferRequest` is created almost at the same time,
- two `StorageTransferAccept`s for the same `StorageTransferRequest` are created almost at the same time in different namespaces,
- two `StorageTransferRequest`s that have the same `source.kind` and `targetName` are created, and `StorageTransferAccept`s for both of them are created in the same namespace,
- a `StorageTransferRequest` is deleted before transfer completes,
- a `StorageTransferAccept` is deleted before transfer completes,
- a resource with the same criteria (namespace, name, and so on) is created after a transfer for the intended resource is completed (do users need to ensure that `StorageTransferRequest` and `StorageTransferAccept` are deleted after a successful transfer)?

mkimuram:
@bswartz 

As for updating secretRefs for PV, I've implemented a prototype on top of the previous POC [commits](https://github.com/kubernetes/enhancements/pull/2326#issuecomment-831330042). The idea is to modify a PV based on its SC's template before it is bound to the destination PV (This means that it only works well for dynamically provisioned case, not for manually provisioned case).

Please check the last two commits of [this branch](https://github.com/mkimuram/kubernetes/commits/transferwithsecret).

Current codes, or the last commit, blindly allow updating PV's secretRefs. However, we will be able to add fields, like transferring stautus/conditions or previousPVC, to PV and only allow updating secretRefs when such fields have expected values. Or we may be able to find a good timing that some of the PV's fields can be used to identify that the PV is clearly transferring. 

Tested with below commands, please also double check:

<details>

- Create storageClass which uses templates for secretRefs
```
cat << 'EOF' | kubectl apply -f -
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: csi-hostpath-sc-with-secret
provisioner: hostpath.csi.k8s.io
parameters:
  csi.storage.k8s.io/controller-publish-secret-namespace: ${pvc.namespace}
  csi.storage.k8s.io/controller-publish-secret-name: ${pvc.name}
  csi.storage.k8s.io/node-stage-secret-namespace: ${pvc.namespace}
  csi.storage.k8s.io/node-stage-secret-name: ${pvc.name}-${pv.name}
  csi.storage.k8s.io/node-publish-secret-namespace: default
  csi.storage.k8s.io/node-publish-secret-name: ${pvc.namespace}-${pvc.name}-${pv.name}
  csi.storage.k8s.io/controller-expand-secret-namespace: default
  csi.storage.k8s.io/controller-expand-secret-name: fixed-name-secret
EOF
```

- Create namespaces to test
```
kubectl create ns ns-a
kubectl create ns ns-b
```

- Create PVC pvc-a in namespace ns-a
```
cat << 'EOF' | kubectl apply -n ns-a  -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-a
spec:
  storageClassName: csi-hostpath-sc-with-secret
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
EOF
```

- Check the PV name for PVC pvc-a
```
kubectl get pvc pvc-a -n ns-a -o jsonpath='{.spec.volumeName}'
pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678
```

- Check the PV's secretRefs
```
kubectl get pv $(kubectl get pvc pvc-a -n ns-a -o jsonpath='{.spec.volumeName}') -o json | jq '.spec.csi'
{
  "controllerExpandSecretRef": {
    "name": "fixed-name-secret",
    "namespace": "default"
  },
  "controllerPublishSecretRef": {
    "name": "pvc-a",
    "namespace": "ns-a"
  },
  "driver": "hostpath.csi.k8s.io",
  "fsType": "ext4",
  "nodePublishSecretRef": {
    "name": "ns-a-pvc-a-pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678",
    "namespace": "default"
  },
  "nodeStageSecretRef": {
    "name": "pvc-a-pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678",
    "namespace": "ns-a"
  },
  "volumeAttributes": {
    "storage.kubernetes.io/csiProvisionerIdentity": "1626384546874-8081-hostpath.csi.k8s.io"
  },
  "volumeHandle": "c7df5132-e5b3-11eb-9372-0242ac110004"
}
```

- Request transfer from namespace ns-a
```
kubectl patch -n ns-a pvc pvc-a -p '{"spec":{"transfer": {"destination": {"namespace": "ns-b", "name": "pvc-b"}}}}'
```

- Accept transfer from namespace ns-b as PVC pvc-b
```
cat << EOF | kubectl create -n ns-b -f - 
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc-b
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  transfer:
    source:
      namespace: ns-a
      name: pvc-a
EOF
```

- Confirm that PV name for the pvc-b  matches the one previously bound to pvc-a
```
kubectl get pvc pvc-b -n ns-b -o jsonpath='{.spec.volumeName}'
pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678
```

- Confirm that PV's secretRefs are updated based on the transferred PVC's information and tempalte in SC
```
kubectl get pv $(kubectl get pvc pvc-b -n ns-b -o jsonpath='{.spec.volumeName}') -o json | jq '.spec.csi'
{
  "controllerExpandSecretRef": {
    "name": "fixed-name-secret",
    "namespace": "default"
  },
  "controllerPublishSecretRef": {
    "name": "pvc-b",
    "namespace": "ns-b"
  },
  "driver": "hostpath.csi.k8s.io",
  "fsType": "ext4",
  "nodePublishSecretRef": {
    "name": "ns-b-pvc-b-pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678",
    "namespace": "default"
  },
  "nodeStageSecretRef": {
    "name": "pvc-b-pvc-a5dfd2c3-e5fe-4bbe-8410-d7c6a8222678",
    "namespace": "ns-b"
  },
  "volumeAttributes": {
    "storage.kubernetes.io/csiProvisionerIdentity": "1626384546874-8081-hostpath.csi.k8s.io"
  },
  "volumeHandle": "c7df5132-e5b3-11eb-9372-0242ac110004"
}
```

</details>

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
mhenriks: > two StorageTransferRequests are created for the same resource and StorageTransferAccept for each StorageTransferRequest is created almost at the same time,
two StorageTransferAccepts for the same StorageTransferRequest are created almost at the same time in different namespaces,
two StorageTransferRequests that have the same source.kind and targetName are created, and StorageTransferAccepts for both of them are created in the same namespace,

I think these are all basically the same issue.  The transfer controller will process the first match it encounters and ignore others.  An annotation on the source can probably help us with the ignore case.  For applications with possibility of contention like this would probably be wise to have a standard naming scheme for requests/accepts and handle collisions that way

mhenriks: > a StorageTransferRequest is deleted before transfer completes,
a StorageTransferAccept is deleted before transfer completes,

I imagine a pending and running phase for a transfer.  During the pending phase (waiting for no pods using the source, etc) a transfer is easily aborted.  Once a transfer is running, things should happen pretty quickly and finalizars can make sure the process completes.  However, if an error is encountered in the running phase, the transfer can be aborted and the system remains in the current state (no rollback)

mhenriks: > a resource with the same criteria (namespace, name, and so on) in StorageTransferRequest is created after a transfer for the intended resource is completed (do users need to ensure that StorageTransferRequest and StorageTransferAccept are deleted after a successful transfer)?

Requests/Accepts should be deleted by the user or perhaps only an admin so that proper auditing occurs

mkimuram: @mhenriks

Thank you for sharing your ideas. Can this information be added to the KEP so that we can discuss more on further corner cases?

On changing the spec, below points would be worth considering.

- What information is needed in annotations to mark PVC as transferring? 
  (One annotation that has identities for both `StorageTransferRequest` and `StorageTransferAccept`, or two separate annotations for each of them?)
- Do we need to add finalizer to both `StorageTransferRequest` and `StorageTransferAccept` to have a chance to properly remove the annotations on a transferring PVC for cancelation?
- Should we consider adding `targetNamespace` to `StorageTransferRequest.Spec`?
  (Current spec seems to allow one StorageTransferRequest to have multiple target namespace candidates, as a result, above annotation logic is likely to become complex. If we change the spec this way, we can focus on one annotation that has identity for only `StorageTransferRequest`.)
- Do `StorageTransferRequest` and `StorageTransferAccept` need to have `Status` fields to manage a pending and running phase?
  (And possibly, should we consider adding completed phase to avoid running multiple times after a successful transfer?)
- Can we consider a way to rollback even after it becomes running phase?


Also, an example for the further corner cases is how to block/work with operations by other controllers, like provisioning, attaching, snapshotting, resizing, etc ... (Do we just make other controllers understand the annotations or consider adding status fields to PVC for this purpose? And if we choose the latter, should information in above annotations also be moved to status fields?)

keps/sig-storage/2324-volume-snapshot-namespace-transfer/README.md
```diff
@@ -0,0 +1,543 @@
...
+- Adds a Finalizer to the PVC source-namespace\foo
+- Set the reclaim policy of the associated PersistentVolume Retain (if not already).
+- Delete the PVC source-namespace\foo.
+- Wait for all Pods to stop using the PVC source-namespace\foo
+- Bind The PersistentVolume to target-namespace\bar.
+- Create the PVC target-namespace\bar by copying the spec of source-namespace\foo and setting spec.volumeName appropriately.
+- Reset the PersistentVolume reclaim policy
```
mkimuram, line 185: How to rest it? Who holds the original policy and where it is stored?

PR #2318, opened 2021-01-22
Updates to KEP 563: test plan, graduation criteria

bridgetkromhout:
These updates to the KEP reflect the tests added for 1.20 and the additional test-related item in the graduation criteria (dualstack tests run as pre-submit blocking for PRs).
 
Signed-off-by: Bridget Kromhout <bridget@kromhout.org>

PR #2344, opened 2021-01-26
Update to KEP 563 - documenting additional tests

bridgetkromhout:
Additional tests to go along with the ones added in https://github.com/kubernetes/enhancements/pull/2318 - these tests were fixed by @aramase in https://github.com/kubernetes/test-infra/pull/20593 and are now completing correctly.

Signed-off-by: Bridget Kromhout <bridget@kromhout.org>


thockin (commented):
Thanks!

/lgtm
/approve

PR #2363, opened 2021-01-29
Add PRR survey to OIDC Discovery KEP

mtaufen:
Add PRR survey as part of preparing to move to GA.

### DIFF
--- a/keps/sig-auth/1393-oidc-discovery/README.md
+++ b/keps/sig-auth/1393-oidc-discovery/README.md
@@ -13,6 +13,13 @@
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
   - [Graduation Criteria](#graduation-criteria)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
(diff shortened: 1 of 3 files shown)


keps/sig-auth/1393-oidc-discovery/README.md
```diff
@@ -296,11 +303,208 @@ of versioning. However, we can still treat graduation in terms of
...
+    - Details: API server logs, or ability of workloads to authenticate to
+      Relying Parties.
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+  We expect the endpoints to maintain high reliability, with reliability
+  matching that of kube-apiserver.
+
+* **Are there any missing metrics that would be useful to have to improve observability
+of this feature?**
+  It would be nice to have usage metrics for this endpoint. We haven't added
```
johnbelamaric, line 416: Is there an existing issue to cover this? I am gathering that the issue is you are serving up a static resource and so any metrics plumbing would have to be in the basic HTTP server in the api server. You don't have any code hit after generating the static data, so you have nowhere to hook in a metric?

If that's the case, perhaps you can at least create an issue and get it on the radar of api machinery, if it's not already.

johnbelamaric (commented):
/approve
/lgtm

keps/sig-auth/1393-oidc-discovery/README.md
mtaufen: Yeah more or less. Just registering a route doesn't seem to get you metrics by default. I could probably add some explicitly in the route handler, but a more transparent solution (with consistent route -> metric name mapping) would be preferable. 

Tracking issue: https://github.com/kubernetes/kubernetes/issues/98585

PR #2373, opened 2021-01-29
Move node label quota to new KEP format

ahg-g:
I didn't change the format of the README nor I added PRR. This will be left for when we want to actually implement this.

/assign @Huang-Wei


wojtek-t:
Thanks!

PR #2396, opened 2021-02-01
KEP-19: Graduate CronJob to stable: update stage and milestone

soltysh:
/assign @wojtek-t @kow3ns

### DIFF
--- a/keps/sig-apps/19-Graduate-CronJob-to-Stable/kep.yaml
+++ b/keps/sig-apps/19-Graduate-CronJob-to-Stable/kep.yaml
@@ -26,12 +26,12 @@ replaces:
   - "https://github.com/kubernetes/community/blob/master/contributors/design-proposals/apps/cronjob.md"
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: beta
+stage: stable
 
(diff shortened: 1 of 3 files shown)


keps/sig-apps/19-Graduate-CronJob-to-Stable/kep.yaml
```diff
@@ -26,12 +26,12 @@ replaces:
   - "https://github.com/kubernetes/community/blob/master/contributors/design-proposals/apps/cronjob.md"
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: beta
+stage: stable
```
wojtek-t, line 29: You need a prod-readiness file, as this one:
https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-apps/85.yaml

I will make a PRR review tomorrow (though I remember reviewing that in the past, so don't expect any significant changes needed).

soltysh: Yeah, I've noticed after initial push and updated that with the right one now.

janetkuo (approved):

PR #2406, opened 2021-02-02
Cache initial PR query results

johnbelamaric:
Caches the PR list for use across each SIG.

Still kludgy but cuts the number of GH API calls roughly in half.

/assign @deads2k 
/cc @wojtek-t @jeremyrickard


justaugustus:
@johnbelamaric -- I'm moving a bunch of things around in https://github.com/kubernetes/enhancements/pull/2528.

Are you okay with:
- waiting until we get that merged in, or:
- me pulling this work into that PR?

PR #2416, opened 2021-02-03
Update DaemonSet KEP for Alpha in 1.21

smarterclayton:
APIs and gate went into 1.20, impl will go into 1.21.

### DIFF
--- a/keps/prod-readiness/sig-apps/1591.yaml
+++ b/keps/prod-readiness/sig-apps/1591.yaml
@@ -1,0 +1,3 @@
+kep-number: 1591
+alpha:
+  approver: "@deads2k"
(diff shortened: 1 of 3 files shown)


soltysh (approved):
/lgtm
/approve

keps/prod-readiness/sig-apps/1591.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 1591
+alpha:
+  approver: "@johnbelamaric"
```
deads2k, line 3: @deads2k

deads2k:
your README.md doesn't show the featuregate.

deads2k:
in readme.md, you miss the need for a read/write cycle before re-enablement to clear out old surge values.

deads2k:
While it is not required for alpha, I think early adopters would appreciate information about how to detect and handle port conflicts during rollout.  Rejecting HostPort isn't fully sufficient since a HostNetwork pod can claim a port.  I think it will be very difficult to debug this sort of failure scenario.

lgtm otherwise.

smarterclayton:
The whole point of surge is that two run at once, so the debug is “your daemonset is never updated” due to either a crashlooping new pod (because you have to SO_REUSEPORT, or because you need to wait for exit).  The equivalent is a deployment with bad config where you can’t rollout new pods.  So in both cases if that’s your first deployment to production the only person you break is yourself, and if your update wedges with surge at least the old pod is still running (a bit like normal deployment rollout).

smarterclayton:
Good point on read write cycle, I’ll update.

deads2k:
> The whole point of surge is that two run at once, so the debug is “your daemonset is never updated” due to either a crashlooping new pod (because you have to SO_REUSEPORT, or because you need to wait for exit)

Yes, I know this and you know this.  But I suspect we'll get tired of telling people....

deads2k:
> Good point on read write cycle, I’ll update.

I think you forgot to push.  Don't forget to update the name for future blame.

smarterclayton:
Added two paragraphs covering some of the topics new pod authors may face.

dims:
the milestone updates looks good. thanks!

/approve

PR #2417, opened 2021-02-03
Update pod resource metrics KEP for beta with PRR

smarterclayton:
Pod resource metrics will go to beta in 1.21

### DIFF
--- a/keps/sig-instrumentation/1748-pod-resource-metrics/README.md
+++ b/keps/sig-instrumentation/1748-pod-resource-metrics/README.md
@@ -102,6 +102,13 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Beta -&gt; GA Graduation](#beta---ga-graduation)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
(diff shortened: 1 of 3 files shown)


keps/sig-instrumentation/1748-pod-resource-metrics/kep.yaml
```diff
@@ -6,7 +6,7 @@ owning-sig: sig-instrumentation
 participating-sigs:
   - sig-node
   - sig-scheduling
-status: implementable
+status: implemented
```
dashpole, line 9: IIRC, they reserve `implemented` to mean GA

dashpole:
This doesn't need a PRR approver since it is already marked implementable, right?

ehashman:
@dashpole when we update the latest milestone to beta, it'll trigger a PRR approver in CI.

keps/sig-instrumentation/1748-pod-resource-metrics/kep.yaml
ehashman: Yeah this should stay in `implementable`, we need to add a stage here:

https://github.com/kubernetes/enhancements/blob/6c2e020d4475d4170ffdeafb266a84bb444ac9a6/keps/NNNN-kep-template/kep.yaml#L26-L27

justaugustus: FYI -- I'm working on cleaning up the validation logic in https://github.com/kubernetes/enhancements/pull/2409.

ref: https://kubernetes.slack.com/archives/C1L57L91V/p1612288049074600, https://kubernetes.slack.com/archives/C1L57L91V/p1612385943085000

annajung:
Hi @smarterclayton, 1.21 Enhancements Lead here. 
Please make sure `kep.yaml` reflects the [latest template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) which includes `prr-approvers`. 

Also, please make sure to follow the [instruction for getting PRR approved](https://github.com/kubernetes/community/blob/master/sig-architecture/production-readiness.md#submitting-a-kep-for-production-readiness-approval). Specifically, this PR is missing a requirement for a yaml file under https://github.com/kubernetes/enhancements/tree/master/keps/prod-readiness/sig-instrumentation. 

Once you have the requirements, please reach out to get this reviewed by a PRR approval by reaching out in #prod-readiness slack channel asap.

Enhancements freeze is just around the corner: Feb. 9th. Please make sure to get all requirements completed before then. Thank you!

keps/sig-instrumentation/1748-pod-resource-metrics/kep.yaml
smarterclayton: I'll change it

ehashman (approved):
/lgtm
/approve

Needs PRR as well.

ehashman (commented):
/lgtm

only diff is @johnbelamaric assigned for PRR

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+
+-->
+
+### Feature Enablement and Rollback
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [ ] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name:
+    - Components depending on the feature gate:
+  - [x] Other
```
johnbelamaric, line 451: That's not really disabling it. So, if for example this new endpoint turns out to have a high cost when scraped, and can therefore be used as a DOS attack on the scheduler, there is no way to turn off the endpoint?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+
+* **Does enabling the feature change any default behavior?**
+
+Scraping these metrics does not change behavior of the system.
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+
+Administrators may stop scraping the endpoint, which will mean the metrics are not available.
```
johnbelamaric, line 465: That's not disabling it, it's just not using it.

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+This cannot impact running workloads unless an unlikely performance issue is triggered due to
+excessive scraping of the scheduler metrics endpoints (which is already possible today).
+
+Since the new metrics are proportionally less than the metrics an apiserver or node exposes,
+it is unlikely that scraping this endpoint would break a metrics collector.
+
+* **What specific metrics should inform a rollback?**
+
+Excessive CPU use from the Kube scheduler when metrics are scraped at a reasonable rate,
+although simply disabling optional scraping while waiting for the bug to be fixed would be
```
johnbelamaric, line 488: how is the metrics endpoint protected, couldn't anyone just scrape it?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+Administrators and visualization tools are the primary target of these metrics and
+so polling and canvassing of Kube distributions is one source of feedback.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+  - [x] Other (treat as last resort)
```
johnbelamaric, line 518: Can you list any specific ones?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+    - Details: The scheduler already has SLIs, any degradation caused by scraping this new endpoint would be visible there.
+
+* **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
+
+The metrics endpoint should consume a tiny fraction of the CPU of the scheduler (less than 5% at idle) when scraped
```
johnbelamaric, line 523: So no change to any existing scheduler SLOs, presumably?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
```diff
@@ -409,11 +416,198 @@ enhancement:
...
+
+  - Panic due to unexpected code path or incomplete API objects returned in watch
+    - Detection: The scrape of the component should fail
+    - Mitigations: Stop scraping the endpoint
+    - Diagnostics: Panic messages in the scheduler logs
+    - Testing: We do not inject fake panics because the behavior of metrics endpoints are well known and there is no background processing.
+
+* **What steps should be taken if SLOs are not being met to determine the problem?**
+
+Perform a golang CPU profile of the scheduler and assess the percentage of CPU charged to the functions
```
johnbelamaric, line 599: Is there any way (from metrics or logs) to tell how often the metrics endpoints are being scraped?

smarterclayton: The prometheus default metrics handler should report it (to the scraper), it would show up in a profile, but it does not write a log line (kind of in keeping with our expectations around kube-apiserver).

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
smarterclayton: Measuring on a production batch job cluster (prow jobs, o(1k pods)) with this change CPU use on scheduler went from 6m +/- 5m to 7m +/- 5m so basically noise.  Scrape interval for that for is 15s.  Testing with synthetic data this code path has so few allocations that the write time / copy time to the http scraper connection dominates CPU use, which is usually a good sign.

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
smarterclayton: It's protected same as the /metrics endpoint which is with delegated auth `--authorization-kubeconfig` which is an RBAC check

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
smarterclayton: The cost when scraped is the same as metrics, so the existing behavior for metrics is covered.  We do not allow metrics to be disabled (but if you disable the metrics set pre-GA it's roughly disablement).  You can disable the entire scheduler HTTP server.

I can clarify, but we generally don't make metrics optional and RBAC would be one way of preventing access to metrics, as would be having the scheduler not listen.  Anyone who can DOS the scheduler can do so today using the existing metrics, so we aren't adding a new DOS angle.

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
smarterclayton: In this case I would probably look at scheduler up / down from the scraping endpoint, then scheduler cpu.  Is that what you're looking for?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
johnbelamaric: Ok, can you just say that (disable the scheduler http server) and how to do it?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
johnbelamaric: Sure, or you can just say: Existing scheduler SLIs.

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
johnbelamaric: Ok, you can just say that: Existing scheduler SLO.

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
johnbelamaric: Ok, so if there is a metric from the default metrics handler please list it. This is for operators, so your instructions are more for someone developing it. I think the actions here would be:
- check the metric to see if it's being scraped like crazy
- if so, disable the scraper
- last resort, disable the http server for scheduler metrics, but that has obvious bad consequences for visibility

smarterclayton: Turns out we don't have those metrics on for request count.  Have we tried to impose a general statement about that?  We're kind of inconsistent but i'd expect sig-instrumentation to blanket enforce that (components that expose metrics should also expose the standard prometheus http request handler).  There are some cardinality concerns.  For right now we'd have to add something to catch that, and we don't *generally* do it.  Would you prefer me to provide a simpler path, or can we punt this to a broader "let's have better standard operational metrics for counting requests that don't blow out cardinality"?

johnbelamaric: we can punt it, maybe create an issue for sig-instrumentation?

keps/sig-instrumentation/1748-pod-resource-metrics/README.md
johnbelamaric: nit: you could just remove the template text and put the answer

johnbelamaric (commented):
/approve

PR #2439, opened 2021-02-06
Update stability KEP(s) with current status(es)

logicalhan:
Updating stability KEPs with up-to-date information.


ehashman:
@logicalhan this is the wrong doc, can you please update the kep.yaml in that directory?

PR #2463, opened 2021-02-08
move Paginated API Lists from community repo

deads2k:
Move https://raw.githubusercontent.com/kubernetes/community/master/contributors/design-proposals/api-machinery/api-chunking.md into enhancements in its current state.  This is the state we have today, without any changes.


smarterclayton:
Agree, this is the expected current state.

PR #2469, opened 2021-02-08
Update Reducing Kubernetes Build Maintenance to implementable

BenTheElder:
ref: #2420


keps/sig-testing/2420-reducing-kubernetes-build-maintenance/README.md
```diff
@@ -125,9 +125,9 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 - [x] (R) Design details are appropriately documented
 - [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
 - [x] (R) Graduation criteria is in place
-- [ ] (R) Production readiness review completed
+- [x] (R) Production readiness review completed
```
spiffxp, line 128: /cc @johnbelamaric @wojtek-t @deads2k 

We believe PRR is not relevant to this KEP, given that it's build-time tech debt removal, and this will result in no change to how kubernetes release artifacts are produced/consumed.

spiffxp: /hold
for ack

BenTheElder: er also I may have misread this particular checkbox: I read this as "filled out the template" and the checkbox below as "the contents of filling out the template were approved" (the checkbox below is not checked currently).

spiffxp (changes requested):
need some kep.yml changes:
- `stage: beta`
- drop TODO re: "if these are the right milestones"

keps/sig-testing/2420-reducing-kubernetes-build-maintenance/README.md
```diff
@@ -380,9 +380,6 @@ n/a.
 
 ## Production Readiness Review Questionnaire
 
-**TODO**: This entire section seems completely irrelevant for KEPs that do not
```
BenTheElder, line 383: separate follow up re: #2311, this should probably be limited to keps that do target production, as opposed to say, test tools or meta procedural stuff like the KEP KEP, saving everyone time in that case ...

BenTheElder:
> need some kep.yml changes:

@spiffxp: done

spiffxp (approved):
/approve
approver for SIG Testing

johnbelamaric (commented):
Agree no PRR needed, but we don't have a way to skip it today. Please add me in prr approvers in kep.yaml and create the PRR approval request file, adn I will approve

keps/sig-testing/2420-reducing-kubernetes-build-maintenance/kep.yml
```diff
@@ -16,7 +16,7 @@ approvers:
   - justaugustus
 # NOTE: there's no production change in this KEP
 prr-approvers:
-  - TBD
+  - johnbelamaric
```
BenTheElder, line 19: @johnbelamaric done.
still looking for why it apparently doesn't parse.

dims:
/approve
/lgtm

@johnbelamaric @BenTheElder feel free to remove hold

keps/sig-testing/2420-reducing-kubernetes-build-maintenance/kep.yml
johnbelamaric: kep.yml -> kep.yaml

BenTheElder: john found it: https://github.com/kubernetes/enhancements/pull/2469/commits/58ed75f953981e71564b5c7e54be64dad1ddd4c1

tengqm:
@BenTheElder IIUC, this KEP does not have any user facing changes, so there is no requirement for user facing docs (to be published on k8s website), correct?

BenTheElder:
Yes, correct. We'll need to notify developers before enacting removal but there are no user facing changes.

Issue #2200, opened 2020-12-18
KEP: Block service ExternalIPs via admission


thockin:
# Block service ExternalIPs via admission
- One-line enhancement description: Allow users to disable the externalIPs feature of Services via admission control
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/2176
- Discussion Link: NA
- Primary contact (assignee): thockin
- Responsible SIGs: sig-network
- Enhancement target (which target equals to which milestone):
  - Stable release target (x.y): 1.21
- [x] Stable
  - [x] KEP (`k/enhancements`) update PR(s): 
  - [x] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/97395
  - [x] Docs (`k/website`) update(s): https://github.com/kubernetes/website/pull/26297
-->

thockin:
Docs PR is up

thockin:
Docs merged.

PR #2478, opened 2021-02-09
KEP 2200 (block externalIPs) is implemented

thockin:


kikisdeliveryservice:
Related issue: https://github.com/kubernetes/enhancements/issues/2200

PR #2483, opened 2021-02-10
Remove the restriction on empty namespaceSelector in podaffinity

ahg-g:
#2249

We initially proposed to disallow empty namespaceSelector. Users however can select all namespaces as follows:
```
- Key: unused-label
  operator: DoesNotExist
```

Selecting all namespaces will likely be a common usage pattern, which we can easily special case and optimize in the scheduler implementation (by avoiding listing namespaces and matching against their labels since all namespaces are eligible). And so I am suggesting we drop the restriction on empty namespaceSelector to avoid forcing users to opt into the pattern above which requires more processing.

@wojtek-t @liggitt @Huang-Wei @alculquicondor


alculquicondor:
lgtm

wojtek-t:
lgtm - thanks!

PR #2524, opened 2021-02-19
Migrate remaining SIG api-machinery KEP to new template

wojtek-t:
Ref #2220 

/assign @lavalamp @deads2k


justaugustus (commented):
@wojtek-t -- Do we also want to add a symlink to prevent external 404s?

wojtek-t:
> @wojtek-t -- Do we also want to add a symlink to prevent external 404s?

I don't see that being referenced from any meaningful place other than the newly-created issue (which I will update once the PR merges). So I don't think it's really needed

justaugustus:
> I don't see that being referenced from any meaningful place other than the newly-created issue (which I will update once the PR merges). So I don't think it's really needed

SGTM!
/lgtm
/approve
/hold for SIG API Machinery

PR #2538, opened 2021-02-23
Update KEP 1904 with upgrade test results

wojtek-t:
Ref #1904 

/assign @deads2k @lavalamp


deads2k:
Thanks, those look good.

/lgtm
/approve

PR #2549, opened 2021-03-03
1602-structured-logging: add note about excluded components

neolit123:
A prior discussion determined kubectl and kubeadm
as components that should not be migrated.

/sig instrumentation
/cc @ehashman


ehashman (commented):
/lgtm
/approve

PR #2559, opened 2021-03-06
Add KEP 2558: Add a +featureGate tag to API

thockin:
Re https://groups.google.com/g/kubernetes-sig-architecture/c/UmPwm-J3ztE/m/QckZWJAABQAJ


dims:
@nikhita will be picking this up in a follow up on monday with blessing from @thockin see:
https://kubernetes.slack.com/archives/C5P3FE08M/p1615058923018400

/close

nikhita:
Created follow up KEP here - https://github.com/kubernetes/enhancements/pull/2560

PR #2561, opened 2021-03-08
new kep: Circumvent 50 tag ASG limit for EKS ManagedNodegroups

MyannaHarris:
Issue: #2556

### DIFF
--- a/keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
+++ b/keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
@@ -1,0 +1,142 @@
+# Circumvent 50 tag ASG limit for EKS ManagedNodegroups
+
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [IAM](#iam)
+- [Design](#design)
+- [Implementation Proposal](#implementation-proposal)
+- [Testing](#testing)
+- [Risks](#risks)
+- [Other Solutions Considered](#other-solutions-considered)
+<!-- /toc -->
+
+## Summary
+
+Currently an EC2 Autoscaling group can only have 50 tags. Many tags are already added to the ASG by standard components like the AWS cloudprovider for Kubernetes and by customers for billing and cost association purposes. Adding labels and taints to the ASG as tags will run into this 50 tag limit. The primary focus of this proposal is to get around the 50 tag limit for customers scaling to/from 0 nodes using Cluster Autoscaler on AWS EKS ManagedNodegroups in a way that will not limit the ManagedNodegroups service.
+
+AWS provides the EKS ManagedNodegroups service which manages the lifecycle of EC2 worker nodes that can join an EKS Kubernetes cluster. Each EKS ManagedNodegroup has an underlying ASG. ASGs and Cluster Autoscaler support scaling to/from 0.
+
+In the [AWS cloud provider][], if there are no nodes in the nodegroup, Cluster Autoscaler creates a nodeTemplate from ASG tags and some default allocatable resources. (See the [code][]) The tags for Cluster Autoscaler include resources, labels, and taints. The nodeTemplate is used to identify available node resources in the absence of an actual EC2 instance existing. We propose that, when a scaled-to-0 EKS ManagedNodegroup is being used, the Cluster Autoscaler cloud provider for AWS takes advantage of the EKS DescribeNodegroup API, which will return the latest lists of labels and taints. Unlike the ASG tags, the API will not limit the number of labels and taints that Cluster Autoscaler can discover to 50.
+
+[AWS cloud provider]: https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/aws
+[code]: https://github.com/kubernetes/autoscaler/blob/99936f3f3971c91b30def913c1c807c42c4b87a7/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L365-L371
+
+## IAM
+
+The ServiceAccount running the Cluster Autoscaler Pod (if using IAM Roles for Service Accounts (IRSA)) or associated with the instance profile (if not using IRSA) will need to have one additional policy applied to it: `eks:DescribeNodegroup`. If Cluster Autoscaler doesn’t have the correct permissions to call the API, it will fall back to using just the ASG tags.
+
+## Design
+
+We propose that the AWS specific driver of Cluster Autoscaler checks for AWS EKS ManagedNodegroups tags on the ASG that marks the ASG as a ManagedNodegroup. If the tags are present, Cluster Autoscaler pulls information from a cache struct containing DescribeNodegroup API response data. The AWS EKS ManagedNodegroups tags look like  `eks:cluster-name : <CLUSTER_NAME>` and `eks:nodegroup-name : <NODEGROUP_NAME>`. We'll use AWS EKS ManagedNodegroups tags because we automatically add them to every managed nodegroup. The DescribeNodegroup API is already in the AWS SDK that Cluster Autoscaler uses. ([Current EKS interface][])
+
+When Cluster Autoscaler finds the ManagedNodegroups tags, it will call functions on a ManagedNodegroupCache struct to get the labels, taints, and some other values. The cache struct will hold the response data from the EKS [DescribeNodegroup API][]. When the cache is accessed (the cache struct methods are called), it will check if there's data cached. If data exists, it will then check if the TTL has expired (1 minute). If there's no cached data or the TTL has expired, the DescribeNodegroup API will be called. Cluster Autoscaler will include both the ASG tag values and the values from the EKS API in its decisions. If an ASG tag contains the same resource as the EKS API (ex: a label key is used in both), Cluster Autoscaler will choose the value from the ASG tag so customers can override any value they want.
+
+This is what a DescribeNodegroup API response looks like (also see [here][]):
+```json
+HTTP/1.1 200
+Content-type: application/json
+{
+   "nodegroup": { 
+      "amiType": "*string*",
+      "capacityType": "*string*",
+      "clusterName": "*string*",
+      "createdAt": *number*,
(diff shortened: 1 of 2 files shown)


keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+When Cluster Autoscaler finds the ManagedNodegroup system tags, it will call functions on a ManagedNodegroupCache struct to get the labels, taints, and some other values. The cache struct will hold the response data from the EKS [DescribeNodegroup API][]. The cache struct methods will call the DescribeNodegroup API when the cached data’s TTL (1 minute default, but will be configurable) has expired or there’s no cached information for the ManagedNodegroup.Cluster Autoscaler will include both the ASG tag values and the values from the EKS API in its decisions. If an ASG tag contains the same resource as the EKS API (ex: a label key is used in both), Cluster Autoscaler will choose the value from the ASG tag so customers can override any value they want.
```
gjtempleton, line 35: When we say `...some other values.` here, are we intending to potentially fetch resource-based information here? If so, do we intend on handling the kubelet reserved resources?

In addition, one nit:
```suggestion
When Cluster Autoscaler finds the ManagedNodegroup system tags, it will call functions on a ManagedNodegroupCache struct to get the labels, taints, and some other values. The cache struct will hold the response data from the EKS [DescribeNodegroup API][]. The cache struct methods will call the DescribeNodegroup API when the cached data’s TTL (1 minute default, but will be configurable) has expired or there’s no cached information for the ManagedNodegroup. Cluster Autoscaler will include both the ASG tag values and the values from the EKS API in its decisions. If an ASG tag contains the same resource as the EKS API (ex: a label key is used in both), Cluster Autoscaler will choose the value from the ASG tag so customers can override any value they want.
```

MyannaHarris: Yes, `...some other values.` is in reference to the other values that DescribeNodegroup returns like capacityType, which customers could use like extended resources. Currently there is no plan to handle kubelet reserved resources in this proposal. Those may be looked at more in a separate proposal that I'm working on.

I'll fix the space.

Thank you

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,139 @@
...
+1. Cluster Autoscaler will then check the ASG tags for the EKS ManagedNodegroups tags. ([ASG tags code][])
+1. If Cluster Autoscaler finds the EKS ManagedNodegroups tags, it will call methods on the ManagedNodegroupCache to get labels, taints, and some other values.
+   1. If the cache has data for the given ManagedNodegroup, it will return that data.
+   1. If the data’s TTL has expired, the DescribeNodegroup API will be called and the data will be added to the cache and returned.
+   1. If there is no data cached for the ManagedNodegroup, the DescribeNodegroup API will be called and the data will be added to the cache and returned.
+1. Then Cluster Autoscaler will pull information from the ASG tags like it already does. If duplicate keys are found, the values from the ASG tags are kept.
+1. Cluster Autoscaler will then continue as usual.
+
+[allocatable resources code]: https://github.com/kubernetes/autoscaler/blob/99936f3f3971c91b30def913c1c807c42c4b87a7/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L366
+[ASG tags code]: https://github.com/kubernetes/autoscaler/blob/99936f3f3971c91b30def913c1c807c42c4b87a7/cluster-autoscaler/cloudprovider/aws/aws_manager.go#L370
```
gjtempleton, line 115: This link appears to lead to a blank line currently. Is this as this is new code yet to be written?

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Testing
+
+- [TC1] Test creating a DescribeNodegroup request
+- [TC2] Test pulling values from cache when data exists for the ManagedNodegroups
+- [TC3] Test pulling values from cache and calling DescribeNodegroup API if no data exists
```
gjtempleton, line 121: Should there be a test case for if the cache TTL has expired?

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+When Cluster Autoscaler finds the ManagedNodegroups tags, it will call functions on a ManagedNodegroupCache struct to get the labels, taints, and some other values. The cache struct will hold the response data from the EKS [DescribeNodegroup API][]. The cache struct methods will call the DescribeNodegroup API when the cached data’s TTL (1 minute default, but will be configurable) has expired or there’s no cached information for the ManagedNodegroup. Cluster Autoscaler will include both the ASG tag values and the values from the EKS API in its decisions. If an ASG tag contains the same resource as the EKS API (ex: a label key is used in both), Cluster Autoscaler will choose the value from the ASG tag so customers can override any value they want.
```
gjtempleton, line 35: This currently seems a bit unclear when first read as to whether the cache structure will automatically rehydrate the cache when a cache entry's TTL has expired in the background, or whether it will only happen when the cache is accessed and the TTL is found to have expired.

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Testing
+
+- [TC1] Test creating a DescribeNodegroup request
+- [TC2] Test pulling values from cache when data exists for the ManagedNodegroups
+- [TC3] Test pulling values from cache and calling DescribeNodegroup API if no data exists
+- [TC4] Test missing permissions and falling back to just ASG tags
+- [TC5] Test duplicate label keys - selecting the value from ASG tags response
+
+## Risks
```
gjtempleton, line 125: From my understanding if we want the cache TTL to be configurable, we'll likely need to add a new configuration flag to the CA, adding another cloud provider specific flag, I think this should be noted somewhere as a drawback to the design as proposed. (I couldn't see a better place to raise this currently.)

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+EKS also throttles describe API calls by default. To mitigate this issue, Cluster Autoscaler will keep a cache of DescribeNodegroup responses: ManagedNodegroupCache. Each cache bucket will have a TTL and, when the TTL expires, DescribeNodegroup will be called again. If there are errors during the call to DescribeNodegroup, Cluster Autoscaler will move on and just look at the ASG tags and existing default allocatable resource values.
+
+[EKS Best Practices]: https://aws.github.io/aws-eks-best-practices/cluster-autoscaling/cluster-autoscaling/#reducing-the-scan-interval
+
+## Other Solutions Considered
+
+Creating a configMap in the cluster that saves all of the node information for ManagedNodegroups was considered.
```
gjtempleton, line 135: I'm unclear, would this proposed solution have been done by the ManagedNodegroup control plane in AWS creating and updating this configMap in the cluster?

gjtempleton (commented):
Some more comments, generally around clarification of points.

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
MyannaHarris: I meant to link line 371. I'll fix that

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
MyannaHarris: I merged the cache test description in TC3. I'll separate them since there would be 2 paths to check.

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
MyannaHarris: I'll clarify. Currently the plan is to only refresh when the cache is accessed and the TTL is found to have expired.

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
MyannaHarris: Sounds good, I'll add that into the doc. Do you think the flag will cause a lot of push-back?

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
MyannaHarris: Yes, that's correct.

keps/sig-autoscaling/2556-circumvent-50-tag-asg-limit/README.md
gjtempleton: I'm not sure how much pushback it'll get, but I think it'll at least be questioned as there's currently only a small number of cloudprovider specific flags and it's adding some configuration complexity (at least in parsing the potential config options) for functionality only of use to a subset of users. Best to call it out explicitly at this stage.

MyannaHarris: After looking into the possible tps for calling DescribeNodegroup after this change and discussing with teammates, I think we shouldn't make the TTL configurable for now. This can be added later on if it becomes a concern. I'll update the kep with this in mind.

MyannaHarris:
After talking with Guy, this proposal is moving directly into the kubernetes/autoscaler repo since it affects users of only 1 cloud provider, doesn't require cross-SIG coordination, and is not an API change.

New PR for proposal: https://github.com/kubernetes/autoscaler/pull/3968

PR #2574, opened 2021-03-16
KEP 2551 - kubectl exit code normalization

rikatz:
xref Issue: https://github.com/kubernetes/enhancements/issues/2551


apelisse:
Note that some exit code are not actual errors. You should definitely account for that.

rikatz:
> Note that some exit code are not actual errors. You should definitely account for that.

True!! So we've discussed in todays sig-cli and I'll try to reduce the scope/amount of error codes, like having:

* An error code that represents some problem on the client side
* A small amount of error codes that might represent apiserver errors (although @soltysh pointed that some forbidden or not found errors may not be well differentiated in some parts like kubectl get, so maybe having the 'forbidden' code for write operations).
* Errors over 128 can be used to represent the sum of something forked. As an example: https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html we can use 129 (128 + 1) if some plugin, called from kubectl exists with 1 (and this may apply to diff as well!)

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,651 @@
...
+We have a number of issues across the repo, as [kubernetes #99354](https://github.com/kubernetes/kubernetes/issues/99354),
+[kubectl #847](https://github.com/kubernetes/kubectl/issues/847), 
+[kubernetes #73056](https://github.com/kubernetes/kubernetes/issues/73056), 
+[kubernetes #39767](https://github.com/kubernetes/kubernetes/issues/39767) and 
+[kubernetes #26424](https://github.com/kubernetes/kubernetes/issues/26424) that points
+to this specific problem.
+
+### Goals
+
+* Document possible exit codes for kubectl
```
sftim, line 89: Will we also cover expectations for plugins and their exit codes? Maybe that's a non-goal at this stage.

rikatz: Hey Tim,

IMO we should not normalize the plugins exit codes, but plugins enter in the 'external programs' category, which we can say that the error code is 128 + the plugin exit code (same for kubectl diff, as an example).

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,652 @@
...
+* Was the command executed correctly
+* If not, where was the problem? On the server side? Was the certificate expired? Or the resource does not exists?
+* Was that a problem on the client side? Was that a problem of an external system command?
+
+We have a number of issues across the repo, as [kubernetes #99354](https://github.com/kubernetes/kubernetes/issues/99354),
+[kubectl #847](https://github.com/kubernetes/kubectl/issues/847), 
+[kubernetes #73056](https://github.com/kubernetes/kubernetes/issues/73056), 
+[kubernetes #39767](https://github.com/kubernetes/kubernetes/issues/39767) and 
+[kubernetes #26424](https://github.com/kubernetes/kubernetes/issues/26424) that points
+to this specific problem.
```
soltysh, line 85: Maybe a few word summary + link would be nice?

keps/sig-cli/2551-return-code-normalization/README.md
soltysh: We can recommend but that's not a hard requirement.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,652 @@
...
+### Risks and Mitigations
+
+* Users already relying on specific error codes might face migration problems / false positives or negatives 
+because of some specific error code change. The proposal here is to decide if the old or new return code behavior
+will be used based on a well defined environment variable. The old behavior will be kept and followed by the 
+deprecation path will be defaulted after a number of releases.
+
+```
+<<[UNRESOLVED feature gating ]>>
+Decide if a global flag to control this behavior will be present, or if the Env Variable is enough
```
soltysh, line 150: We usually go with env var for starters.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,652 @@
...
+Define the initial error codes, and how to document them
+<<[/UNRESOLVED]>>
+```
+
+## Design Details
+
+The majority of commands already are organized as the following:
+* Run Complete to complete missing information with defaults. This runs on the client side
+* Run Validation to check command syntax and missing arguments. This runs on the client side
+* Run the command itself. This might run on the client side, be dry-run, run an external command or call the APIServer.
```
soltysh, line 169: This might actually change during the separattion of cobra from `Options` struct, so it would be nice to put it in a more generic way.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,652 @@
...
+Select the best design approach, move others to design alternatives
+<<[/UNRESOLVED]>>
+```
+
+```
+<<[UNRESOLVED dealing with kubectl exec and run exit ]>>
+kubectl exec and run uses the pod exit code as its own exit code, we should figure out how we 
+should deal with this
+https://github.com/kubernetes/kubernetes/blob/v1.22.0-alpha.0/test/e2e/kubectl/kubectl.go#L496
+https://github.com/kubernetes/kubernetes/blob/v1.22.0-alpha.0/staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go#L178-L179
```
soltysh, line 187: Yeah, that's a big one and how we can properly handle that.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,652 @@
...
+in the manner they are designed, calling another checkErr inner function.
+
+For the rollout of this approach, a new non-exported error checker function needs to be developed and make the 
+call to the old checkErr or the new function according to the value of the Environment Variable described in
+[Risks and Mitigations](#risks-and-mitigations) 
+
+### Creating new error parser functions
+
+Another design solution is to create helper functions for each steps:
+* When running Complete, Validate or other client side steps, call cmdutil.CheckClientErr(err) and exits with some well defined client error code, mapped to ErrorCodeClient
```
soltysh, line 214: As mentioned before I'd describe this in slightly more generic way, since this implementation might change.

rikatz:
/remove-lifecycle stale
I will still work on this :D 
/lifecycle active

keps/sig-cli/2551-return-code-normalization/README.md
deejross: Should this follow the 128+exit code rule proposed for plugins?

deejross: One other option is to start known kubectl error codes at 200. This would allow exec to continue returning the pod exit code, assuming it's less than 200. If the pod exit code is 200 or greater, then set it to 255. Maybe we should apply the same logic to external commands as well, such as diff to keep it consistent. Thoughts on that approach?

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: @rikatz will tackle on this

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: @deejross will take a look on this

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: @deejross to take a look into this

keps/sig-cli/2551-return-code-normalization/README.md
sftim: How about having `kubectl exec` take a parameter `--use-exit-code` that defaults to true, and that you can force to false if you want success to mean “we were able to `exec` _something_”?

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+Usually they use an external plugin to verify if the manifests are linted, and issue a warning if not. Joice wants to
+safelly ignore those lint errors from the plugin on the pipeline, as the exit code for linting might be well known, but 
+wants to warn users when the apply command fails because of differences between the manifests.
+
+#### Story 2
+Bruce Wayne, the security administrator of the Gotham Inc Company is following the development of a new product. Bruce asked
+for the developers to warn when a new deployment fails because of the lack of some permission, so those permissions can be 
+updated for the pipeline to work correctly.
+
+The developers are making a lot of changes, and they keeps asking for Bruce to look for every pipeline execution, even those that
```
soltysh, line 123:
```suggestion
The developers are making a lot of changes, and they keep asking for Bruce to look for every pipeline execution, even those that 
```

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+### User Stories
+
+#### Story 1
+Joice, the SRE of a big company is automating the deployment of the infrastructure with a lot of kubectl commands.
+Usually they use an external plugin to verify if the manifests are linted, and issue a warning if not. Joice wants to
+safelly ignore those lint errors from the plugin on the pipeline, as the exit code for linting might be well known, but 
+wants to warn users when the apply command fails because of differences between the manifests.
+
+#### Story 2
+Bruce Wayne, the security administrator of the Gotham Inc Company is following the development of a new product. Bruce asked
```
soltysh, line 119:
```suggestion
Bruce Wayne, the security administrator of the Gotham Inc company is following the development of a new product. Bruce asked
```

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+
+One thing that should be done is map all the error codes/ints as constants in some file, so they can be automatically
+documented.
+
+### Error Codes
+The following table represents the proposed error codes and the condition describing its usage.
+
+| Code  | Description                                                                                           |
+| ----- | ----------------------------------------------------------------------------------------------------- |
+| 1-200 | Reserved for exit codes from exec and external commands                                               |
```
soltysh, line 169: I was wondering if we need to reserve all 200, if just the initial 100 isn't sufficient? But that's something we'll need to empirically figure out as we go.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+| 201   | Catch-all for errors where the condition is unknown or no better codes exist to describe it           |
+| 202   | Missing or improper use of keyword, command, or argument                                              |
+| 203   | Client configuration error, invalid or missing configuration                                          |
+| 204   | Network failure, API could not be reached                                                             |
+| 205   | Authentication failure, identity could not be determined                                              |
+| 206   | Authorization failure, identity was determined, but does not have access to requested resource(s)     |
```
soltysh, line 175: This and not found are hard to distinguish due to server returning 404 Not Found in both cases, due to security reasons.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+The con of using this design approach is that the Error Validation functions are pretty hard to change/evolve
+in the manner they are designed, calling another checkErr inner function.
+
+For the rollout of this approach, a new non-exported error checker function needs to be developed and make the 
+call to the old checkErr or the new function according to the value of the Environment Variable described in
+[Risks and Mitigations](#risks-and-mitigations) 
+
+### Creating new error parser functions
+
+Another design solution is to create helper functions for each steps:
```
soltysh, line 216: I was thinking about it earlier this week and I think that implementing a set of new error codes inside kubectl where each command could use them and then having `CheckErr` return new error codes with that env of yours and just -1 in the backwards compatible way as it does today. Which is close with what you're describing here.

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,637 @@
...
+Consider the following in developing a version skew strategy for this
+enhancement:
+- Does this enhancement involve coordinating behavior in the control plane and
+  in the kubelet? How does an n-2 kubelet without this feature available behave
+  when this feature is used?
+- Will any other components on the node change? For example, changes to CSI,
+  CRI or CNI may require updating that component before the kubelet.
+-->
+
+## Production Readiness Review Questionnaire
```
soltysh, line 341: You're missing file in https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/ and make sure to update the template with current one.

soltysh (approved):
A couple of nits, but mostly this looks good
/hold
for you to get PRR file added and nits addressed 
/lgtm
/approve

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: agreed. This was a discussion I was on with Ross, that maybe the code reservation can be something more sparse or in a "range" other than reserving all the 200

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: We can leave this as an UNRESOLVED tag so we are stating this is not solved yet

keps/sig-cli/2551-return-code-normalization/README.md
rikatz: sorry @soltysh is there any action on this? I'm a bit slow this latest days :)

rikatz:
PRR added
/assign @johnbelamaric

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,421 @@
...
+Not applicable, as this is just a kubectl change
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+
+- [X] Feature gate (also fill in values in `kep.yaml`)
```
johnbelamaric, line 295: it's not really a feature gate, please just document it as an env var, I think the template calls this "Other"

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,421 @@
...
+
+Yes, just remove the Env Var.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+kubectl will start issuing different and standardized exit codes again when errors occurs.
+
+###### Are there any tests for feature enablement/disablement?
+
+No.
```
johnbelamaric, line 312: Presumably you can create a test that checks a return code (or codes) to make sure enablement is working?

keps/sig-cli/2551-return-code-normalization/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.24"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.24"
+
```
johnbelamaric, line 29: you're missing some fields from the latest template, please update

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,421 @@
...
+This behavior is feature gated through an Environment Variable and can be enabled or 
+disabled without impact to upgrade or downgrade between versions 
+
+### Version Skew Strategy
+
+Not applicable, as this is just a kubectl change
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
```
johnbelamaric, line 290: I think you have an older version of the questions, please review the latest template and include any new questions as well. Thanks.

johnbelamaric (commented):
Minor tweaks needed to PRR but looks good in general.

keps/sig-cli/2551-return-code-normalization/kep.yaml
```diff
@@ -0,0 +1,29 @@
+title: kubectl return code normalization 
+kep-number: 2551
+authors:
+  - "@rikatz"
+  - "@deejross"
+owning-sig: sig-cli
+status: implementable
+creation-date: 2021-03-16
+reviewers:
```
apelisse, line 9: I took a look thanks!

rikatz:
@johnbelamaric fixed based on your comments, thanks :)

keps/sig-cli/2551-return-code-normalization/README.md
```diff
@@ -0,0 +1,408 @@
...
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
```
soltysh, line 44: Nit: some of these checkboxes need to be checked.

keps/sig-cli/2551-return-code-normalization/README.md
soltysh: Let's "close" this list before going to beta, for alpha let's leave it open.

keps/sig-cli/2551-return-code-normalization/README.md
soltysh: Exactly what I said above :)

soltysh (approved):
Some minor nits, but this is good, thx!
/lgtm
/approve

PR #2581, opened 2021-03-21
Added KEP for custom metrics router

arjunrn:
For #2580


keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,651 @@
...
+existing information already exposed in most metrics adapters. The router can
+leverage existing projects like the
+[custom-metrics-apiserver](https://github.com/kubernetes-sigs/custom-metrics-apiserver)
+and the [metrics client](https://github.com/kubernetes/metrics).
+
+### Goals
+
+* Allows for multiple metric sources in a cluster Existing metrics sources
+* should be compatible with solution Leverage existing functionality in metric
+* sources for discovery and routing.
```
zroubalik, line 84:
```suggestion
* Allows for multiple metric sources in a cluster 
* Existing metrics sources should be compatible with the solution
* Leverage existing functionality in metric sources for discovery and routing.
```

keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,651 @@
...
+table. One of the requirements of this routing table is that routing entries
+should be able to be added and deleted without a complete re-indexing of all
+metrics. And finally there is the request router. This component receives the
+metric requests which are proxied from the API server and then consults the
+routing table and queries the correct metric source and returns the result.
+
+![Router Architecture](custom_metrics_router.png)
+
+#### Metrics Discovery Controller
+`CustomMetricsSource` is the proposed CRD which should be used to register
```
zroubalik, line 132: The current name could give users an impression, that it servers custom metrics only. What about `MetricsSourceProvider` or a similar name?

keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,651 @@
...
+
+## Motivation
+
+Currently, there can only be a single [custom or external metrics
+source](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-metrics-apis)
+in a kubernetes cluster. This is a problem because the different custom
+metric providers which have been developed since this feature was added
+provide different metrics for different domains. For instance the
+[prometheus-metrics-adapter](https://github.com/kubernetes-sigs/prometheus-adapter)
+exposes metrics stored in Prometheus. Similarly there are metrics adapters
```
zroubalik, line 67:
```suggestion
exposes metrics stored in Prometheus and [KEDA](https://github.com/kedacore/keda) uses external metrics for event driven based autoscaling. Similarly there are metrics adapters
```

brancz:
The intention of this is sound. Last time I proposed something like this apimachinery folks weren't sure how they felt about this approach. Maybe someone can weigh in @kubernetes/sig-api-machinery-misc . In some sense, as you mention in the proposal this is basically a per-namespace APIService.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
arjunrn: I'm open to other names. I wanted to keep the name similar to the `APIService` object already present and with the `custom-metrics-apiserver`.

arjunrn:
@brancz Thanks for inputs. I'm not sure what you mean by "this is basically a per-namespace APIService" because the new CRD I'm proposing is not namespaced.
And about the metrics API offering watch. The way I see when the `metrics-client` is extended to support it the router can implement it as well proxying the watch.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
zroubalik: Yeah, I just would like to avoid the `Custom` prefix, since this should deal with external metrics as well, otherwise I don't have a strong opinion on the name. 
`APIService` objects are named `v1beta[12].custom.metrics.k8s.io` or `v1beta1.external.metrics.k8s.io`, that depends on the type of metrics.

lavalamp:
This can be a drop-in replacement of the existing custom metric apiserver, right? no apiserver or controller manager changes needed?

arjunrn:
@lavalamp Yes, this requires no changes in the apiserver or the controller manager. But it's not a replacement but a proxy for several custom metric apiservers. There's a diagram in the KEP but it doesn't get rendered correctly in the PR. Here's what it will look like.
custom_metrics_router

brancz:
I'm generally happy with this. Apimachinery should be as well though and perform an API review.

keps/sig-instrumentation/2580-custom-metrics-router/kep.yaml
```diff
@@ -0,0 +1,30 @@
+title: KEP Template
+kep-number: NNNN
```
ehashman, line 2: Please ensure you fill this file out.

ehashman:
/assign @logicalhan 

SIG Instrumentation approver

lavalamp:
> requires no changes in the apiserver or the controller manager

Sorry for slow response, I have no objections from API Machinery.

ehashman:
We discussed this at our last instrumentation meeting. I believe that this enhancement is out of tree, so following the KEP cycle is optional and the release team does not need to track this PR. PRR is also optional.

arjunrn:
@serathius So how do we proceed here?

keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,658 @@
...
+
+This CRD uses almost the same structure as the
+[APIService](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#apiservice-v1-apiregistration-k8s-io)
+used to register APIs in the kube-apiserver. In the kube-apiserver higher
+priority `APIservices` for the same API group are always used for queries of
+that type. However when 2 services have the same priority, the requests could
+be sent to either service with equal probability. In the case of the metrics
+service it might not make sense to do that same because the returned metric
+values may differ causing the replica count of the target to fluctuate.
+Instead the age of the metrics API service registration object can be used as
```
barkbay, line 179: Could you elaborate on what object age is used here  ? And how ?

barkbay (commented):
Thanks for this PR !
I just left a few questions.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
arjunrn: All kubernetes resources(custom or otherwise) have a creation timestamp was is set when the resource is created. The routing table would contain entries of providers which expose a given metric. This would be sorted in increasing order based on the creation timestamp.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,658 @@
...
+The [API and client](https://github.com/kubernetes/metrics) for resource,
+custom and external metrics are used by the HPA controller to query for
+metrics from any registered custom/external metric providers. Technically the
+client is instantiated with the kube-apiserver as the target for the requests
+which in turn forwards it to the metrics server based on the API group i.e.
+`metrics.k8s.io`, `custom.metrics.k8s.io` or `external.metrics.k8s.io`.
+
+#### API Discovery
+API servers expose endpoints which can be used by a client to find the
+correct resource version and discover information about which API resources
```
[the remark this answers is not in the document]
arjunrn: The router would rely on the official [metrics-client](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/metrics) and I assume that the client uses the discovery API to find the preferred version and then query that. I might be wrong. Is there any use case you can think of which would require an override of the default behavior.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
```diff
@@ -0,0 +1,658 @@
...
+type MetricType string
+
+const (
+  CustomMetricsType   = "CustomMetrics"
+  ExternalMetricsType = "ExternalMetrics"
+)
+
+type CustomMetricsSourceSpec struct {
+  Service               Service      `json:"service"`
+  InsecureSkipTLSVerify bool         `json:"insecureSkipTLSVerify"`
```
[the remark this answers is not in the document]
arjunrn: Added the `caBundle` to the spec.

djamaile:
I am interested in a feature on this. Any progress on getting the merged? looks like it is almost read.

arjunrn:
@ehashman Maybe you could help here. What should we do with this KEP? Merge/close? I can transfer the ownership of the router I have implemented and become a maintainer for it also with someone from sig-instrumentation.

dgrisonnet:
If I remember correctly from the sig-instrumentation meeting where this KEP was discussed, the path forward was to move the KEP to either https://github.com/kubernetes-sigs/instrumentation or https://github.com/kubernetes-sigs/custom-metrics-apiserver since it isn't related to k/k.
The original reason for the KEP to be opened here was that we wanted to discuss if it was possible to have aggregated APIs namespaced, but now that this solution has been dismissed, I am fine with moving the KEP to the custom-metrics-apiserver repo.

I personally think that it makes sense to have the router living in its own repo in kubernetes-sigs as part of the project owned by sig-instrumentation since it is an essential add-on to the metrics API story for users who want to use multiple metrics servers. wdyt @serathius @brancz @s-urbaniak?

keps/sig-instrumentation/2580-custom-metrics-router/README.md
barkbay: My concern with the creation timestamp is that it is not obvious for the user to get it, even displayed in a column it is hard to figure out which one is the most recent among a bunch of resources.

I was wondering if sorting resources by `name` would be a better choice:
* `metadata.name` is an immutable field, if a resource is deleted and then recreated with the same name the creation timestamp will change, but not its name. It makes the behaviour more predictable.
* I don't have a good example for K8S but on OpenShift this is [how SCC are prioritized](https://docs.openshift.com/container-platform/4.7/authentication/managing-security-context-constraints.html#scc-prioritization_configuring-internal-oauth). My experience as a user is that it is a nice way to solve that problem.

What do you think ?

keps/sig-instrumentation/2580-custom-metrics-router/README.md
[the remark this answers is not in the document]
barkbay: > Is there any use case you can think of which would require an override of the default behavior.

No, I think you're right. Thanks for the clarification 👍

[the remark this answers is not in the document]
barkbay: > The router would rely on the official [metrics-client](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/metrics) and I assume that the client uses the discovery API to find the preferred version and then query that

Sorry, I remember now 😄 : I'm not sure it is possible to use the metrics-client to discover the list of metrics.
For example in [your proof concept](https://github.com/arjunrn/custom-metrics-router/blob/ade1ceb84ff9a78dc5948efd8ec9a8d325266aae/pkg/metricsclient/metricsclient.go#L103) you use the discovery client:

```golang
c.discoveryClient.ServerResourcesForGroupVersion(customMetricsAPI.SchemeGroupVersion.String())
```

In the snippet above `customMetricsAPI.SchemeGroupVersion.String()` would be statically resolved according to the imported version of `customMetricsAPI`. Which might not be the same than the one selected by the metrics client.

But I don't think it's a problem since we can reuse the `AvailableAPIsGetter` from the metrics client, for example if we keep `cmClient.NewAvailableAPIsGetter(discoveryClient)` in the [Client structure](https://github.com/arjunrn/custom-metrics-router/blob/master/pkg/metricsclient/metricsclient.go#L36):

```golang
version, err := c.customMetricsAvailableAPIsGetter.PreferredVersion()
if err != nil {
	return nil, err
}
resources, err := c.discoveryClient.ServerResourcesForGroupVersion(version.String())
if err != nil {
	return nil, fmt.Errorf("failed to get resource for %s: %v", customMetricsAPI.SchemeGroupVersion, err)
}
```

But this is an implementation detail, and you can still ignore my initial comment in the context of the KEP.

keps/sig-instrumentation/2580-custom-metrics-router/README.md
arjunrn: The reason for using timestamp is to reduce the chances of accidentally changing the provider which exposes a given metric. This might make it harder to debug but the metric value would be stable. Assume the user registers a new provider and is unaware that it exposes a metric which is currently in use. The new metric value is different to the existing metric value. This means the scaling target would suddenly scale up or down probably causing an outage.
Also to improve debugging I think the routing table could be written out to a configmap which can be examined for debugging. There is also a need to provide overrides for specific metric which this KEP doesn't address but we can cross that bridge when we get there.

barkbay (commented):
I was wondering if the KEP should provide some details about what happens if it is not possible to get the metrics list from one of the `MetricsSource`. For example, say that a `MetricsSource` with a high priority has a temporarily failure and it is not possible to initialize or refresh the list of the metrics. Does it prevent the router to serve requests ? Should we add a parameter somewhere to configure the expected behaviour in that situation ?

brancz:
I think what @dgrisonnet said makes sense to me.

djamaile:
@arjunrn are you going to move the PR?

mastersingh24:
Did this get moved?

djamaile:
Any update?

arjunrn:
Sorry everyone, I have not had the time to work on this and won't have be able to any time soon. If someone wants to pick this let me know. I can also transfer the ownership of the prototype repo to you or to the sigs org.

mastersingh24:
@arjunrn  - Happy to try to take this forward.    Is this supposed to move to https://github.com/kubernetes-sigs/custom-metrics-apiserver ?

arjunrn:
@mastersingh24 It can also be moved to https://github.com/kubernetes-sigs/instrumentation according to this [comment](https://github.com/kubernetes-sigs/custom-metrics-apiserver). I'm closing this PR because it should be added this repo.

mastersingh24:
@scarby  - I have not.  Would love to try to move this forward, but could use some help moving it somewhere.

dcfranca:
At the moment, we are running a `kube-metrics-adapter`, but we want to migrate to Keda
Is there any workaround to manage to run both during the migration?

PR #2594, opened 2021-04-01
Support Multiple Cluster CIDRs

rahulkjoshi:
Issue: https://github.com/kubernetes/enhancements/issues/2593

### DIFF
--- a/keps/sig-network/2594-multiple-cluster-cidrs/README.md
+++ b/keps/sig-network/2594-multiple-cluster-cidrs/README.md
@@ -1,0 +1,1072 @@
+# KEP-2594: Enhanced NodeIPAM to support Discontiguous Cluster CIDR
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Add more pod IPs to the cluster](#add-more-pod-ips-to-the-cluster)
+    - [Add nodes with higher or lower capabilities](#add-nodes-with-higher-or-lower-capabilities)
+    - [Provision discontiguous ranges](#provision-discontiguous-ranges)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Pre-Requisites](#pre-requisites)
+  - [New Resource](#new-resource)
+    - [Expected Behavior](#expected-behavior)
+    - [Example: Allocations](#example-allocations)
+  - [Controller](#controller)
+    - [Data Structures](#data-structures)
+    - [Dual-Stack Support](#dual-stack-support)
+    - [Startup Options](#startup-options)
+    - [Startup](#startup)
+    - [Reconciliation Loop](#reconciliation-loop)
+    - [Event Watching Loops](#event-watching-loops)
+      - [Node Added](#node-added)
+      - [Node Updated](#node-updated)
+      - [Node Deleted](#node-deleted)
+      - [ClusterCIDRConfig Added](#clustercidrconfig-added)
+      - [ClusterCIDRConfig Updated](#clustercidrconfig-updated)
+      - [ClusterCIDRConfig Deleted](#clustercidrconfig-deleted)
+  - [kube-controller-manager](#kube-controller-manager)
+  - [Test Plan](#test-plan)
+    - [Unit Tests and Benchmarks](#unit-tests-and-benchmarks)
+    - [Integration Tests](#integration-tests)
+    - [End-to-End Tests](#end-to-end-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha to Beta Graduation](#alpha-to-beta-graduation)
+    - [Beta to  GA Graduation](#beta-to--ga-graduation)
+    - [Make the Controller the new default](#make-the-controller-the-new-default)
+    - [Mark the RangeAllocator as deprecated](#mark-the-rangeallocator-as-deprecated)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Upgrades](#upgrades)
+    - [Downgrades](#downgrades)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Share Resources with Service API](#share-resources-with-service-api)
+    - [Pros](#pros)
+    - [Cons](#cons)
+  - [Nodes Register CIDR Request](#nodes-register-cidr-request)
+    - [Pros](#pros-1)
+    - [Cons](#cons-1)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone /
+release*.
+
+-   [ ] (R) Enhancement issue in release milestone, which links to KEP dir in
+    [kubernetes/enhancements](not the initial KEP PR)
+-   [ ] (R) KEP approvers have approved the KEP status as `implementable`
+-   [ ] (R) Design details are appropriately documented
+-   [ ] (R) Test plan is in place, giving consideration to SIG Architecture and
+    SIG Testing input (including test refactors)
+-   [ ] (R) Graduation criteria is in place
+-   [ ] (R) Production readiness review completed
+-   [ ] (R) Production readiness review approved
+-   [ ] "Implementation History" section is up-to-date for milestone
+-   [ ] User-facing documentation has been created in [kubernetes/website], for
+    publication to [kubernetes.io]
+-   [ ] Supporting documentation—e.g., additional design documents, links to
+    mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Today, when Kubernetes' NodeIPAM controller allocates IP ranges for podCIDRs for
+nodes, it uses a single range allocated to the cluster (cluster CIDR). Each node
+gets a range of a fixed size from the overall cluster CIDR. The size is
+specified during cluster startup time and cannot be modified later on.
+
+Kubernetes' IPAM capabilities are an optional behavior that comes with
+Kubernetes out of the box. It is not required for Kubernetes to function, and
+users may use alternate mechanisms.
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new
+CIDR allocator that can be controlled by a new resource `ClusterCIDRConfig`.
+This would enable users to dynamically allocate more IP ranges for pods. The new
+functionality would remain optional, and be an enhancement for those using the
+built-in IPAM functionality.
+
+## Motivation
+
+Today, IP ranges for podCIDRs for nodes are allocated from a single range
+allocated to the cluster (cluster CIDR). Each node gets a range of a fixed size
+from the overall cluster CIDR. The size is specified during cluster startup time
+and cannot be modified later on. This has multiple disadvantages:
+
+*   There is just one cluster CIDR from which all pod CIDRs are allocated. This
+    means that users need to provision the entire IP range up front accounting
+    for the largest cluster that may be created. This can waste IP addresses.
+*   If a cluster grows beyond expectations, there isn't a simple way to add more
+    IP addresses.
+*   The cluster CIDR is one large range. It may be difficult to find a
+    contiguous block of IP addresses that satisfy the needs of the cluster.
+*   Each node gets a fixed size IP range within a cluster. This means that if
+    nodes are of different sizes and capacity, users cannot allocate a bigger
+    pod range to a given node with larger capacity and a smaller range to nodes
+    with lesser capacity. This wastes a lot of IP addresses.
+
+### Goals
+
+*   Support multiple discontiguous IP CIDR blocks for Cluster CIDR
+*   Support node affinity of CIDR blocks
+*   Extensible to allow different block sizes allocated to nodes
+*   Does not require master or controller restart to add/remove ranges for pods.
+
+### Non-Goals
+
+*   Not providing a generalized IPAM API to Kubernetes. We plan to enhance the
+    RangeAllocator’s current behavior (give each Node a /XX from the Cluster
+    CIDR as its `PodCIDR`)
+*   No change to the default behavior of a Kubernetes cluster.
+    *   This will be an optional API and can be disabled (as today’s NodeIPAM
+        controllers may also be disabled)
+
+## Proposal
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new
+CIDR allocator that can be controlled by a new resource 'ClusterCIDRConfig'.
+This enables users to dynamically allocate more IP ranges for pods. In addition,
+it gives users the capability to control what ranges are allocated to specific
+nodes as well as the size of the pod CIDR allocated to these nodes.
+
+### User Stories
+
+#### Add more pod IPs to the cluster
+
+A user created a cluster with an initial clusterCIDR value of 10.1.0.0/20. Each
+node is assigned a /24 pod CIDR so the user could create a maximum of 16 nodes.
+However, the cluster needs to be expanded but the user does not have enough IPs
+for pods.
+
+With this enhancement, the user can now allocate an additional CIDR for pods;
+eg. 10.2.0.0/20 with the same configuration to allocate a /24 pod CIDR. This
+way, the cluster can now grow by an additional 16 nodes.
+
+#### Add nodes with higher or lower capabilities
+
+A user created a cluster with an ample sized cluster CIDR. All the initial nodes
+are of uniform capacity capable of running a maximum of 256 pods and they are
+each assigned a /24 pod CIDR. The user is planning to add more nodes to the
+system which are capable of running 500 pods. However, they cannot take
+advantage of the additional capacity because all nodes are assigned a /24 pod
+CIDR. With this enhancement the user configures a new allocation which uses the
+original cluster CIDR but allocates a /23 instead of a /24 to each node. They
+use the node selector to allocate these IPs only to the nodes with the higher
+capacity.
+
+#### Provision discontiguous ranges
+
+A user wants to create a cluster with 32 nodes each with a capacity to run 256
+pods. This means that each node needs a /24 pod CIDR range and they need a total
+range of /19. However, there aren't enough contiguous IPs in the user's network.
+They can find 4 free ranges of size /21 but no single contiguous /19 range.
+
+Using this enhancement, the user creates 4 different CIDR configurations each
+with a /21 range. The CIDR allocator allocates a /24 range from any of these /21
+ranges to the nodes and the user can now create the cluster.
+
+### Notes/Constraints/Caveats
+
+This feature does not expand the ability of the NodeIPAM controller to change
+the `Node.Spec.PodCIDRs` field. Once that field is set, either by the controller
+or a third party, it will be treated as immutable. This is particularly relevant
+in situtaitons where users start modifying or deleting the `ClusterCidrConfig`.
+Under no circumstances will the controller attempt to revoke the allocated
+CIDRs (more details on this are discussed below).
+
+### Risks and Mitigations
+
+-   Racing kube-controller-managers. If multiples of the controller are running
+    (as in a HA control plane), how do they coordinate?
+    -   The controllers will coordinate using the existing
+        kube-controller-manager leader election.
+
+## Design Details
+
+### Pre-Requisites
+
+This KEP assumes that the only consumer of the `--cluster-cidr` value is the
+NodeIPAM controller. [KEP
+2450](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2450-Remove-knowledge-of-pod-cluster-CIDR-from-iptables-rules)
+proposed modifications to the kube-proxy to remove it's dependence on a
+monolithic ClusterCIDR.  The kube-proxy flag `--detect-local-mode` must be set
+to `NodeCIDR` to properly handle nodes having discontiguous Pod CIDRs.
+
+Users not using kube-proxy must ensure that any components they have installed
+do not assume Kubernetes has a single continguous Pod CIDR.
+
+### New Resource
+
+```go
+type ClusterCIDRConfig struct {
+  metav1.TypeMeta
+  metav1.ObjectMeta
+
+  Spec ClusterCIDRConfigSpec
+  Status ClusterCIDRConfigStatus
+}
+
+type ClusterCIDRConfigSpec struct {
+    // This defines which nodes the config is applicable to. A nil selector can
+    // be applied to any node.
+    // +optional
+    NodeSelector *v1.LabelSelector
+
+    // This defines the IPv4 CIDR assignable to nodes selected by this config.
+    // +optional
+    IPv4 *ClusterCIDRSpec
+
+    // This defines the IPv6 CIDR assignable to nodes selected by this config.
+    // +optional
+    IPv6 *ClusterCIDRSpec
+}
+
+type ClusterCIDRSpec struct {
+    // An IP block in CIDR notation ("10.0.0.0/8", "fd12:3456:789a:1::/64")
+    // +required
+    CIDR string
+
+    // Netmask size (e.g. 25 -> "/25") to allocate to a node.
+    // Users would have to ensure that the kubelet doesn't try to schedule more
+    // pods than are supported by the node's netmask (i.e. the kubelet's
+    // --max-pods flag)
+    // +required
+    PerNodeMaskSize int
+}
+
+type ClusterCIDRConfigStatus struct {
+}
+```
+
+#### Expected Behavior
+
+-   `NodeSelector`, `IPv4`, and `IPv6` are immutable after creation.
+
+-   `IPv4.PerNodeMaskSize` and `IPv6.PerNodeMaskSize` must specify the same
+    number of IP addresses:
+
+    ```32 - IPv4.PerNodeMaskSize == 128 - IPv6.PerNodeMaskSize```
+
+-   Each node will be assigned all Pod CIDRs from a matching config.
+    Consider the following example:
+    
+    ```go
+    {
+        IPv4: {
+            CIDR:            "10.0.0.0/20",
+            PerNodeMaskSize: "22",
+        },
+        IPv6: {
+            CIDR:            "fd12:3456:789a:1::/64"
+            PerNodeMaskSize: "118",
+        },
+    }
+    ```
+    Only 4  nodes may be allocated from this `ClusterCIDRConfig` as only 4 IPv4
+    Pod CIDRs can be partitioned from the IPv4 CIDR. The remaining IPv6 Pod
+    CIDRs may be used if referenced in another `ClusterCIDRConfig`.
+
+-   In case of multiple matching ranges, attempt to break ties with the
+    following rules:
+    1.  Pick the `ClusterCIDRConfig` whose `NodeSelector` matches the most
+        labels on the `Node`. For example, `{'node.kubernetes.io/instance-type':
+        'medium', 'rack': 'rack1'}` before `{'node.kubernetes.io/instance-type':
+        'medium'}`.
+    1.  Pick the `ClusterCIDRConfig` with the fewest Pod CIDRs allocatable. For
+        example, `{CIDR: "10.0.0.0/16", PerNodeMaskSize: "16"}` (1 possible Pod
+        CIDR) is picked before `{CIDR: "192.168.0.0/20", PerNodeMaskSize: "22"}`
+        (4 possible Pod CIDRs)
+    1.  Pick the `ClusterCIDRConfig` whose `PerNodeMaskSize` is the fewest IPs.
+        For example, `27` (32 IPs) picked before `25` (128 IPs).
+    1.  Break ties arbitrarily.
+
+-   An empty `NodeSelector` functions as a default that applies to all nodes.
+    This should be the fall-back and not take precedence if any other range
+    matches. If there are multiple default ranges, ties are broken using the
+    scheme outlined above.
+
+-   When breaking ties between matching `ClusterCIDRConfig`, if the most
+    applicable (as defined by the tie-break rules) has no more free allocations,
+    attempt to allocate from the next highest matching `ClusterCIDRConfig`. For
+    example consider a node with the labels:
+    ```go
+    {
+        "node": "n1",
+        "rack": "rack1",
+    }
+    ```
+    If the following `ClusterCIDRConfig` are programmed on the cluster, evaluate
+    them from first to last using the first config with allocatable CIDRs. In
+    the example below, the `CluserCIDRConfig` have already been sorted according
+    to the tie-break rules.
+    ```go
+    {
+        NodeSelector: { MatchLabels: { "node": "n1", "rack": "rack1" } },
+        IPv4: {
+            CIDR:            "10.5.0.0/16",
+            PerNodeMaskSize: 26,
+        }
+    },
+    {
+        NodeSelector: { MatchLabels: { "node": "n1" } },
+        IPv4: {
+            CIDR:            "192.168.128.0/17",
+            PerNodeMaskSize: 28,
+        }
+    },
+    {
+        NodeSelector: { MatchLabels: { "node": "n1" } },
+        IPv4: {
+            CIDR:            "192.168.64.0/20",
+            PerNodeMaskSize: 28,
+        }
+    },
+    {
+        NodeSelector: nil,
+        IPv4: {
+            CIDR:            "10.0.0.0/8",
+            PerNodeMaskSize: 26,
+        }
+    }
+    ```
+
+-   The controller will add a finalizer to the `ClusterCIDRConfig` object
+    when it is created.
+
+-   On deletion of the `ClusterCIDRConfig`, the controller checks to see if any
+    Nodes are using `PodCIDRs` from this range -- if so it keeps the finalizer
+    in place and periodically polls Nodes. When all Nodes using this
+    `ClusterCIDRConfig` are deleted, the finalizer is removed.
+
+#### Example: Allocations
+
+```go
+[
+    {
+        // Default for nodes not matching any other rule
+        NodeSelector: nil,
+        IPv4: {
+            // For existing clusters this is the same as ClusterCIDR
+            CIDR:            "10.0.0.0/8",
+            // For existing API this is the same as NodeCIDRMaskSize
+            PerNodeMaskSize: 24,
+        }
+    },
+    {
+        // Another range, also allocate-able to any node
+        NodeSelector: nil,
+        IPv4: {
+            CIDR:            "172.16.0.0/14",
+            PerNodeMaskSize: 24,
+        }
+    },
+    {
+        NodeSelector: { "node": "n1" },
+        IPv4: {
+            CIDR:            "10.0.0.0/8",
+            PerNodeMaskSize: 26,
+        }
+    },
+    {
+        NodeSelector: { "node": "n2" },
+        IPv4: {
+            CIDR:            "192.168.0.0/16",
+            PerNodeMaskSize: 26,
+        }
+    },
+    {
+        NodeSelector: { "node": "n3" },
+        IPv4: {
+            CIDR:            "5.2.0.0/16",
+            PerNodeMaskSize: 26,
+        }
+        IPv6: {
+            CIDR:            "fd12:3456:789a:1::/64",
+            PerNodeMaskSize: 122,
+        }
+    },
+  ...
+]
+```
+
+Given the above config, a valid potential configuration might be:
+
+```
+{"node": "n1"} --> "10.0.0.0/26"
+{"node": "n2"} --> "192.16.0.0/26"
+{"node": "n3"} --> "5.2.0.0/20", "fd12:3456:789a:1::/122"
+{"node": "n4"} --> "172.16.0.0/24"
+```
+
+### Controller
+
+Implement a new
+[NodeIPAM controller](https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/nodeipam)
+The controller will set up watchers on the `ClusterCIDRConfig` objects and the
+`Node` objects.
+
+This controller relies on being a single writer (just as the current NodeIPAM
+controller does as well). In the case of HA control planes with multiple
+replicas, there will have to be some form of leader election to enforce only 1
+active leader. This KEP proposes re-using the kube-controller-manager leader
+election to pick a active controller.
+
+#### Data Structures
+
+We will use maps to store the allocated ranges and which node is using the
+range. Because the number of nodes is expected to be on the order of thousands,
+more sophisticated data structures are likely not required.
+
+Prior investgations [here](https://github.com/kubernetes/kubernetes/pull/90184)
+suggest that maps storing allocations will perform well under the number of
+nodes we expect.
+
+#### Dual-Stack Support
+
+To assign both IPv4 and IPv6 Pod CIDRs to a Node, the `IPv4` and `IPv6` fields
+must be both set on the object. The controller does not have an in-built notion
+of single-stack or dual-stack clusters. It uses the tie-break rules specified
+[above](#expected-behavior) to pick a `ClusterCIDRConfig` from which to allocate
+Pod CIDRs for each Node.
+
+The controller makes no guarantees that all Nodes are single-stack or that all
+Nodes are dual-stack. This is to specifically allow users to upgrade existing
+clusters.
+
+#### Startup Options
+
+The following startup options will be supported (via the
+kube-controller-manager). They are optional, and intended to support migrating
+from the existing NodeIPAM controller:
+-   `serviceCIDRs` : In some situations, users have Service CIDRs which
+    overlap with their Pod CIDR space. The controller will not allocate any IPs
+    which fall within the provided Service CIDRs.
+    
+    Currently, this is specified to the kube-controller-manager by the
+    `--service-cluster-ip-range` flag.
+-   `clusterCIDR` : Users can specify to Kubernetes which CIDR to use for Pod
+    IPs. This is a widely read configuration specified by the
+    `--cluster-cidr` flag.
+-   `nodeCIDRMaskSize` (in single-stack IPv4) : Defines the size of the per-node
+    mask in the single-stack IPv4 case.
+
+    Currently this is specified to the kube-controller-manager by the
+    `--node-cidr-mask-size` flag.
+-   `nodeCIDRMaskSizeIPv4` and `nodeCIDRMaskSizeIPv6` (in dual-stack mode):
+    Defines the size of the per-node masks for IPv4 and IPv6 respectively.
+
+    Currently this is specified to the kube-controller-manager by the
+    `--node-cidr-mask-size-ipv4` and `--node-cidr-mask-size-ipv6` flags.
+
+#### Startup
+
+-   Fetch list of `ClusterCIDRConfig` and build internal data structure
+-   If they are set, read the `--cluster-cidr` and `--node-cidr-mask-size` flags
+    and attempt to create `ClusterCIDRConfig` with the name
+    "created-from-flags-\<hash\>".
+    -   In the dual-stack case, the flags `--node-cidr-mask-size-ipv4` and
+        `--node-cidr-mask-size-ipv6` are used instead, they will also be used as
+        necessary.
+    -   The "created-from-flags-\<hash\>" object will always be created as long
+        as the flags are set. The hash is arbitrarily assigned.
+    -   If an object with the name "created-from-flags-\<hash>" already exists,
+        but it does not match the flag values, the controller will delete it and
+        create a new object. The controller will ensure (on startup) that there
+        is only one non-deleted `ClusterCIDRConfig` with the name
+        "create-from-flags\<hash>". This will allow users to change the flag
+        values and stop using the old values. 
+-   Fetch list of `Node`s. Check each node for `PodCIDRs`
+    -   If `PodCIDR` is set, mark the allocation in the internal data structure
+        and store this association with the node.
+    -   If `PodCIDR` is set, but is not part of one of the tracked
+        `ClusterCIDRConfig`, emit a K8s event but do nothing.
+    -   If `PodCIDR` is not set, save Node for allocation in the next step.
+        After processing all nodes, allocate ranges to any nodes without Pod
+        CIDR(s) [Same logic as Node Added event]
+
+#### Reconciliation Loop
+
+This go-routine will watch for cleanup operations and failed allocations and
+continue to try them in the background.
+
+For example if a Node can't be allocated a PodCIDR, it will be periodically
+retried until it can be allocated a range or it is deleted.
+
+#### Event Watching Loops
+
+##### Node Added
+
+If the Node already has a `PodCIDR` allocated, mark the CIDRs as used.
+
+Otherwise, go through the list of `ClusterCIDRConfig`s and find ranges matching
+the node selector from each family. Attempt to allocate Pod CIDR(s) with the
+given per-node size. If that `ClusterCIDRConfig` cannot fit a node, search for
+another `ClusterCIDRConfig`.
+
+If no `ClusterCIDRConfig` matches the node, or if all matching
+`ClusterCIDRConfig`s are full, raise a K8s event and put the Node on the
+reconciliation queue (infinite retries). Upon successfully allocating CIDR(s),
+update the node object with the podCIDRs.
+
+##### Node Updated
+
+Check that its Pod CIDR(s) match internal allocation.
+
+-   If node.spec.PodCIDRs is already filled up, honor that allocation and mark
+    those ranges as allocated.
+-   If the node.spec.PodCIDRs is filled with a CIDR not from any
+    `ClusterCIDRConfig`, raise a K8sEvent.
+-   If the ranges are already marked as allocated for some other node, raise
+    another error event (there isn’t an obvious reconciliation step the
+    controller can take unilaterally).
+
+##### Node Deleted
+
+Release said Node’s allocation from the internal data-structure.
+
+If this Node is the last one using a particular `ClusterCIDRConfig` that has
+been slated for deletion, trigger the deletion flow again (so that the finalizer
+is removed and internal data structures are cleaned up).
+
+##### ClusterCIDRConfig Added
+
+Install a finalizer on the `ClusterCIDRConfig` called "networking.kubernetes.io/cluster-cidr-config-finalizer".
+
+Update internal representation of CIDRs to include the new range. Every failed
+Node Allocation is stored in a queue, that will be tried again with the new
+range by the reconciliation loop.
+
+##### ClusterCIDRConfig Updated
+
+_`NodeSelector`, `IPv4`, and `IPv6` are immutable so any updates should be
+rejected_
+
+##### ClusterCIDRConfig Deleted
+
+1.  Update internal data structures to mark the range as terminating (so new
+    nodes won't be added to it)
+1.  Search the internal representation of the CIDR range to see if any Nodes are
+    using the range.
+    1.  If there are no nodes using the range, remove the finalizer and cleanup
+        all internal state.
+    1.  If there are nodes using the range, wait for them to be deleted before
+        removing the finalizer and cleaning up. 
+
+### kube-controller-manager
+
+The flag `--cidr-allocator-type` will be amended to include a new type
+"ClusterCIDRConfigAllocator".
+
+The list of current valid types is
+[here](https://github.com/kubernetes/kubernetes/blob/1ff18a9c43f59ffed3b2d266b31e0d696d04eaff/pkg/controller/nodeipam/ipam/cidr_allocator.go#L38).
+
+### Test Plan
+
+#### Unit Tests and Benchmarks
+
+-   Ensure that the controller scales to ~5,000 nodes -- memory usage and
+    reasonable allocation times
+
+#### Integration Tests
+
+-   Verify finalizers and statuses are persisted appropriately
+-   Test watchers
+-   Ensure that the controller handles the feature being disabled and re-enabled:
+    -   Test with some Nodes already having `PodCIDR` allocations
+
+#### End-to-End Tests
+
+-   Run through some sample workflows. Just a few for example:
+    -   Adding a node
+    -   Adding a ClusterCIDRConfig
+    -   Deleting a ClusterCIDRConfig that is in use
+-   Run through the [user stories](#user-stories):
+    -   Expand the ClusterCIDR (existing nodes without alloations are
+        allocated and new nodes also get ranges.
+    -   Use `NodeSelector` to allocate different sized CIDRs to different nodes.
+    -   Create and use discontiguous ranges.
+
+### Graduation Criteria
+
+#### Alpha to Beta Graduation
+
+-   Gather feedback from users about any issues
+-   Tests are in testgrid
+
+#### Beta to  GA Graduation
+
+-   Wait for 1 release to receive any additional feedback
+
+#### Make the Controller the new default
+
+After the GA graduation, change the default NodeIPAM allocator from
+RangeAllocator to ClusterCIDRConfigAllocator. This will involve changing the
+default value of the flag on the kube-controller-manager
+(`--cidr-allocator-type`).
+
+#### Mark the RangeAllocator as deprecated
+
+In the same release that the ClusterCIDRConfigAllocator is made the default,
+mark the RangeAllocator as deprecated.
+
+After 2 releases, the code can be removed.
+
+### Upgrade / Downgrade Strategy
+
+#### Upgrades
+
+There is no change to the defaults as part of the alpha, so existing clusters
+will upgrade seemlessly.
+
+If we want to use the new controller, users will have to change the
+`--cidr-allocator-type` flag on the kube-controller-manager. The new controller
+will respect the existing flags for `--cluster-cidr` and
+`--node-cidr-mask-size`.
+
+Users will also have to change the kube-proxy flags as outlined in [KEP
+2450](https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/2450-Remove-knowledge-of-pod-cluster-CIDR-from-iptables-rules).
+The flag `--detect-local-mode` must be set to `NodeCIDR` to properly handle
+nodes having discontiguous Pod CIDRs.
+
+#### Downgrades
+
+Users may "downgrade" by switching back the `--cidr-allocator-type` to
+"RangeAllocator". If users only use the existing flags (`--cluster-cidr` and
+`--node-cidr-mask-size`), then downgrade will be seamless. The Node `PodCIDR`
+allocations will persist even after the downgrade, and the old controller can
+start allocating PodCIDRs
+
+If users use the `ClusterCIDRConfig` resource to specify CIDRs, switching to the
+old controller will maintain any Node `PodCIDR` allocations that have already
+been created. Users will have to manually remove the finalizer from the
+`ClusterCIDRConfig` objects before they can be deleted.
+
+### Version Skew Strategy
+
+<!--
+If applicable, how will the component handle version skew with other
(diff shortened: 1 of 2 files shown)


keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+-->
+
+## Design Details
+
+### New Resource
+```
+ClusterCIDRConfig {
+    # Nodes may only have 1 range from each family.
+    # TODO: maybe this is redundant and can just be inferred by the controller.
+    IPFamily        FamilyType
```
dcbw, line 161: Can't this be determined by `IPCIDRBlock`? eg, isn't it more work to validate that IPFamily is the same as IPCIDRBlock, when all you're going to really use is the block?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to maintain previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+### Version Skew Strategy
```
aojea, line 393: this is going to be funny :smile: 
https://github.com/kubernetes/kubernetes/issues/95948

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
```
aojea, line 708: One alternative to avoid the validation problem with overlapping ranges, is to have only one resources for the whole cluster that contains an array with the CIDRRanges

```
ClusterCIDR {
  Type      CIDRType
  Ranges []CIDRRange
}

  IPFamily  string # Perhaps omit – inferable from CIDRBlock
  CIDRBlock string # Example "10.0.0.0/8" or "fd12:3456:789a:1::/64"
  Selector  v1.LabelSelector # Specifies which Services or Nodes can be
                               # assigned IPs from this block.
  BlockSize string # How large of an IP block to allocate. For services
                   # this would always be "/32". Example "/24"
```

with cluster CIDRs we don't expect a big number so I don't think the object can grow much

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+    IPV6 FamilyType = "IPV6"
+)
+```
+
+#### Expected Behavior
+
+- Each node will be assigned up to one range from each `FamilyType`.
+    - **TODO:** what is the tie-break if multiple ranges match
+- An empty `NodeSelector` functions as a default that applies to all nodes. This should be the fall-back and not take precedence if any other range matches.
+- `IPFamily`, `IPCIDRBlock`, and `PerNodeMaskSize` are immutable after creation.
```
aojea, line 191: why do you need IPFamily? it can be obtained from the IPCIDRBlock,no?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+]
+```
+
+Given the above config, a valid potential configuration might be:
+
+```
+{“np”: “np1”} --> "10.0.0.0/26"
+{“np”: “np2”} --> "192.16.0.0/26"
+{“np”: “np3”} --> "5.2.0.0/20", "fd12:3456:789a:1::/112"
+{“np”: “np4”} --> "172.16.0.0/24"
```
aojea, line 249: hmm, will this break the premise that all pods should be able to reach all pods in the cluster?
should we allow this mix or "enforce" all nodes to have the same families?
I prefer the later

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+```
+
+#### Expected Behavior
+
+- Each node will be assigned up to one range from each `FamilyType`.
+    - **TODO:** what is the tie-break if multiple ranges match
+- An empty `NodeSelector` functions as a default that applies to all nodes. This should be the fall-back and not take precedence if any other range matches.
+- `IPFamily`, `IPCIDRBlock`, and `PerNodeMaskSize` are immutable after creation.
+- Reject any changes that would change existing node allocations:
+    - Can not delete a `CIDRRangeConfig` that is in use by a node
```
aojea, line 193: how do you know if it is in use?
do we use to the PV/PVC model?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: oh, scratch it, is mentioned before
>     # TODO: maybe this is redundant and can just be inferred by the controller.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+    - If `PodCIDR` is set, but is not part of one of the tracked `ClusterCIDRConfig`s, emit a K8s event but do nothing.
+    - If `PodCIDR` is not set, save Node for allocation in the next step.
+After processing all nodes, allocate ranges to any nodes without Pod CIDR(s) [Same logic as Node Added event]
+
+#### Event Watching Loops
+
+##### Node Added
+Go through the list of `ClusterCIDRConfig`s and find ranges matching the node selector from each family. Attempt to allocate Pod CIDR(s) with the given per-node size. If that `ClusterCIDRConfig` cannot fit a node, search for another `ClusterCIDRConfig`.
+
+If no `ClusterCIDRConfig` matches the node, or if all matching `ClusterCIDRConfig`s are full, raise a K8s event and put the Node back on the queue for allocation (infinite retries).
```
aojea, line 273: what is doing current controller? is doing the same?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: Don't we still need to deal with overlapping ranges as the user could specify overlapping CIDRBlocks within the array?

I think overlapping ranges are fine if we coalace all of the allocated ranges in one in memory data structure (maybe a trie) in the backend. In other words, the user can specify overlapping ranges, but the allocator is smart enough to not allocate overlapping ranges.

aojea: this is related to the commented in the sig-network meeting, right now you don't have access to this information during the API operations.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Yeah agreed. Dropping IPFamily -- I was on the fence about it already.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+## Design Details
+
+### New Resource
+```
+ClusterCIDRConfig {
```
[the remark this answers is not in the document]
rahulkjoshi: Done.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+    # Nodes may only have 1 range from each family.
+    # TODO: maybe this is redundant and can just be inferred by the controller.
+    IPFamily        FamilyType
+    
+    # An IP block in CIDR notation ("10.0.0.0/8", "fd12:3456:789a:1::/64")
+    # +required
+    IPCIDRBlock     string
```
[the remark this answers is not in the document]
rahulkjoshi: Makes sense.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
thockin: Agree, I think.  It's a little daunting to expose so much of the inner workings, I admit.  I think we would strongly urge people to create pairs of configs in dual-stack mode.  IIRC there's not a "primary family" notion for pod CIDRs like there is with Services.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+### Controller
+
+Implement a new
+[NodeIPAM controller](https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/nodeipam)
+The controller will set up watchers on the `ClusterCIDRConfig` objects and the
+`Node` objects.
+
+#### Data Structures
+
+We want to use a set of radix trees to store allocation. This is a commonly used
```
thockin, line 303: I wonder if we even need to go this far.  We're talking about O(thousands) of Nodes - 15000 is the largest I know of, but even assuming 64K, we can probably afford to optimize for maintenance.   Even a simple sorted array might cut it.  I don't know what libs exist for this in Go, but let's not shoot the moon.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+pattern (linux uses it to store routing tables). This will track the ranges that
+are already allocated. We will also store the range allocated to each node. This
+is to handle conflicts in case some entity modifies the podCIDR value of the
+Node object.
+
+#### Startup
+
+-   Fetch the values of the cluster-cidr and node-cidr-mask-size arguments
```
thockin, line 311: This is in case of no config resources defined?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+-   Fetch list of `Node`s. Check each node for `PodCIDRs`
+    -   If `PodCIDR` is set, mark the allocation in the internal data structure
+        and store this association with the node.
+    -   If `PodCIDR` is set, but is not part of one of the tracked
+        `ClusterCIDRConfig`s, emit a K8s event but do nothing.
+    -   If `PodCIDR` is not set, save Node for allocation in the next step.
+        After processing all nodes, allocate ranges to any nodes without Pod
+        CIDR(s) [Same logic as Node Added event]
+
+#### Event Watching Loops
```
thockin, line 324: You'll probably want a parallel "repair" loop that just tries to rebuild the state, verify it, fix what it can, sleep, try again.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
thockin: I like your sense of humor. :)

thockin (commented):
"K8s will reject attempts to delete a CIDRRange that’s currently in use" - how?   See other comment on referential integrity.

Other than RI, I am most worried about conflict detection, but maybe the leader election is good enough.  What I was thinking was that it would be nice and simple to have an intermediate type, call it `CIDRAlloc` whose name was the allocated block.  If I am allocating "10.9.3.0/24" I create a CIDRAAlloc named "10.9.3.0/24" and link it to both the ClusterCIDRConfig and the Node.  That way, nobody could ever allocate the same range twice and you could run allocators in parallel (leaning on etcd to detect conflicts).

If I have two allocators and one chooses "10.9.3.0/24" at the same time as another that chooses ""10.9.3.128/25", my plan is not going to work.  This is how @aojea and I were hand-waving at service IP allocation.  Is it a requirement to do multiple sizes on the same range?

sdmodi:
> Other than RI, I am most worried about conflict detection, but maybe the leader election is good enough. What I was thinking was that it would be nice and simple to have an intermediate type, call it `CIDRAlloc` whose name was the allocated block. If I am allocating "10.9.3.0/24" I create a CIDRAAlloc named "10.9.3.0/24" and link it to both the ClusterCIDRConfig and the Node. That way, nobody could ever allocate the same range twice and you could run allocators in parallel (leaning on etcd to detect conflicts).
> 
> If I have two allocators and one chooses "10.9.3.0/24" at the same time as another that chooses ""10.9.3.128/25", my plan is not going to work. This is how Antonio Ojea and I were hand-waving at service IP allocation. Is it a requirement to do multiple sizes on the same range?

I think it is fine to have a conflict in range specification. Lets say we have two resources range1 and range 2 which are 10.9.3.0/24 and 10.9.2.128/25 respectively. We can continue allocating IP addresses from 10.9.3.0/24. It is possible that the label selectors for the two ranges are different. Even in this case we can reasonably argue that as long as we ensure that nodes that match range2 label selectors get IPs from 10.9.2.128/25 we are honoring the CIDR range inputs. 

Doing multiple sizes on the same range is quite powerful. It allows users that use a single CIDR range to avoid IP wastage.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: I finally understand what you mean by not having access. We cannot access the backend data structures. Can we access any data structures other than the new object that is being added? Or is the entire validation only on the static content being added?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: We didn't have a good understanding of how validations would work. We don't have an idea of how we would do this.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: Yes. The range allocator keeps trying and does not bring up the node.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Today, IP ranges for podCIDRs for nodes are allocated from a single range allocated to the cluster (cluster CIDR). Each node gets a range of a fixed size from the overall cluster CIDR. The size is specified during cluster startup time and cannot be modified later on.
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new CIDR allocator that can be controlled by a new resource `ClusterCIDRRange`. This would enable users to dynamically allocate more IP ranges for pods.
```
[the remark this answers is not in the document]
sdmodi: Right now the way this is scoped, we do not require existing Node IPAM allocators to populate what ranges have been allocated. If we did this, then it would force existing implementations to create a new object when a new range is allocated. 

I view this as a good second step if there is consensus that this is a good idea. We can start with just requiring the object for input purposes and then enhance it to be consumable by other parts of the system

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+
+* Not providing a generalized IPAM API to Kubernetes. We plan to enhance the RangeAllocator’s current behavior (give each Node a /XX from the Cluster CIDR as its `PodCIDR`)
+* No change to the default behavior of a Kubernetes cluster.
+    * This will be an optional API and can be disabled (as today’s NodeIPAM controllers may also be disabled)
+
+## Proposal
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new CIDR allocator that can be controlled by a new resource 'CIDRRange'. This enables users to dynamically allocate more IP ranges for pods. In addition, it gives users the capability to control what ranges are allocated to specific nodes as well as the size of the pod CIDR allocated to these nodes.
+
+### User Stories (Optional)
```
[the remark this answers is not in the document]
sdmodi: Yes. We are still saying only one per node. 

We do need to decide whether it is a better model to let CNI implementations do this. The fundamental question is whether we want Kubernetes to do the IPAM or not. Today Kubernetes does do IPAM out of the box and it is lacking a lot of functionality. We are proposing we enhance these capabilities.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+    IPV4 FamilyType = "IPV4"
+    IPV6 FamilyType = "IPV6"
+)
+```
+
+#### Expected Behavior
+
+- Each node will be assigned up to one range from each `FamilyType`.
+    - **TODO:** what is the tie-break if multiple ranges match
+- An empty `NodeSelector` functions as a default that applies to all nodes. This should be the fall-back and not take precedence if any other range matches.
```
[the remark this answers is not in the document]
sdmodi: This is something we hadn't thought of. In this particular case, the end goal is to select only one range. We cannot match all selectors if they are across range objects. If we have two CIDR range objects range1 with node selector "color:blue" and range2 with node selector "size:small", it is unclear what range is being selected for a small blue node.

Maybe we say that the behaviour is undefined. Or maybe we add a new label type 'RangeSelector' and force all nodes to carry a label of that type?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+{“np”: “np3”} --> "5.2.0.0/20", "fd12:3456:789a:1::/112"
+{“np”: “np4”} --> "172.16.0.0/24"
+```
+
+### Controller
+
+Implement a new [NodeIPAM controller](https://github.com/kubernetes/kubernetes/tree/master/pkg/controller/nodeipam) The controller will set up watchers on the `ClusterCIDRConfig` objects and the `Node` objects.
+
+#### Data Structures
+We want to use a set of radix trees to store allocation. This is a commonly used pattern (linux uses it to store routing tables). This will track the ranges that are already allocated. We will also store the range allocated to each node. This is to handle conflicts in case some entity modifies the podCIDR value of the Node object.
```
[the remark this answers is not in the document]
sdmodi: I think the n in this case would be the number of nodes. We can still use a simple list and keep this linear. O(n^2) would probably get expensive.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
danwinship: oh, I just meant having the network plugin create an object/objects saying "The following CIDRs are in use by the pod network: ...". Not necessarily saying which specific subnets of it have been allocated where.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
thockin: There are two parts here, which makes it a little weird.

Incoming API ops hit admission control and then the REST stack.  We must synchronously validate the data (or not?).  We can define what it means to validate (e.g. just "within a valid range" or "in range and not colliding").  Either way this is async to other operations on CIDRRange resources (e.g. we could validate it and then the range is removed immediately after that - we'd still have an invalid object).

It's not unheard of for admission control to be aware of other objects, but it's not common.  The REST stack is a little weirder.  That's where validation usually happens, and the validation code itself is supposed to be "static".  There are a few places (e.g. Service) that do weird stuff like etcd ops and validating the input is in-range to a flag, so again not unheard of, but uncommon.

The second part - the controller - is all async and much, much easier to reason about.

So IN THEORY we could look at the CIDR ranges during either admission (more self-contained) or validation.  We still have to decide if it is really worthwhile to validate that far, since we can still lose the race.  We don't have interlocks between objects and unless we actually implement referential integrity we can't prevent an object from being deleted or an address from being used twice (apiservers run in parallel)

So the racy path would be:
t0: controller writes node A `podCIDR = 10.10.10.0/24` to apiserver 1
t1: apiserver 1 validates that CIDRange 10.10.0.0/16 allows this allocation AND that no other node is using it
t2: buggy controller writes node B `podCIDR = 10.10.10.0/24` to apiserver 2
t3: apiserver 2 validates that CIDRange 10.10.0.0/16 allows this allocation AND that no other node is using it
t4: apiserver 1 approves the write and it goes to etcd
t5: apiserver 2 approves the write and it goes to etcd

== Nodes A and B have the same CIDR

Another bad path:
t0: controller writes node A `podCIDR = 10.10.10.0/24` to apiserver 1
t1: apiserver 1 validates that CIDRange 10.10.0.0/16 allows this allocation AND that no other node is using it
t2: user deletes the CIDRRange 
t3: apiserver 1 approves the write and it goes to etcd

== Node A has a CIDR that is not covered by any range

To get around the first case we either trust the controller(s) or we force some verifiable contention, causing whomever loses the race to fail the transaction and force the controller to retry.  E.g. the Service bitmap serves this role, but can end up never resolving under extreme contention.  Or you just trust the allocator to have done the right thing (what we do today).

To get around the second case we could use a finalizer - the IPAM controller adds a finalizer like "in-use" to the range it allocates from.  When the user deletes the range, it will hang as long as nodes exist that use the range.  The controller knows that range is being deleted, so it won't allocate more from it.  As nodes are removed, the controller keeps track and when that hits 0, removes the finalizer.  Need to think about what corner cases exist here.

Alternately, implement admission control to reject delete ops while in-use, but I don't think we do that anywhere, so unprecedented.

This still leaves the controller(s) to do safe allocations, which I don't think etcd can help with.  That means we can only have 1 active controller which is normal.  But there is talk of eventually sharding controllers, and this would fail that unless it also shards ranges.

I think this topic warrants more design and pros/cons of approaches.  We may just be making it harder than it needs to be?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: maps will work, you have an example and some benchmarks here https://github.com/kubernetes/kubernetes/pull/90184

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: I think there's 2 approaches we could take here.

1. Use a admission controller that validates any changes before they're committed. So in this case, the request to delete the CIDRRangeConfig would come to the controller. The controller knows if a range is in use, so it could reject the request.
2. Use the PV/PVC model of putting the object in a "terminating" state and delete it when the last node using it goes away. We can probably attach a condition to the object listing the nodes still using the config.

I'm leaning towards option 2. My thinking is that this is a model that we've used before so should be familiar. Are there any major roadblocks you can think of if we were to try option 2.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+
+#### Example Config
+```
+[
+{
+    IPFamily:        IPV4,
+    # For existing clusters this is the same as ClusterCIDR
+    IPCIDRBlock:     "10.0.0.0/8",
+    # Default for nodes not matching any other rule
+    NodeSelector:    nil,
```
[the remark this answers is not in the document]
rahulkjoshi: Ack I mde NodeSelector a pointer.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+
+#### Expected Behavior
+
+-   Each node will be assigned up to one range from each `FamilyType`.
+    -   **TODO:** what is the tie-break if multiple ranges match
+-   An empty `NodeSelector` functions as a default that applies to all nodes.
+    This should be the fall-back and not take precedence if any other range
+    matches.
+-   `CIDRBlock`, and `PerNodeMaskSize` are immutable after creation.
+-   Reject any changes that would change existing node allocations:
```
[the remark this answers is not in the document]
rahulkjoshi: @aojea made an interesting point below -- we could adopt the model of PV/PVC where the controller puts finalizers on all the CIDRRangeConfigs that are in use. Attempting to delete a range in use would put it in a "terminated" phase so that no new Nodes can draw from that range, but the User must delete all nodes using that range before it is truly deleted.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+    # can be applied to any node.
+    # +optional
+    NodeSelector    *v1.LabelSelector
+
+    # Netmask size (e.g. 25 -> "/25") to allocate to a node.
+    # Users would have to ensure that the kubelet doesn't try to schedule
+    # more pods than are supported by the node's netmask (i.e. the kubelet's
+    # --max-pods flag)
+    # +required
+    PerNodeMaskSize int
```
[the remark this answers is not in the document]
rahulkjoshi: Yeah, we're allowing overlapping CIDR ranges as part of this proposal. The thought is that we can rely on leader election to prevent overlapping/conflicting allocations to nodes.

Fundamentally, I don't think this is a new problem -- even the existing controller relies on kube-controller-manager's leader election. I don't believe having multiple CIDR blocks makes this problem worse (at least as far as I can see).

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,776 @@
...
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Today, IP ranges for podCIDRs for nodes are allocated from a single range allocated to the cluster (cluster CIDR). Each node gets a range of a fixed size from the overall cluster CIDR. The size is specified during cluster startup time and cannot be modified later on.
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new CIDR allocator that can be controlled by a new resource `ClusterCIDRRange`. This would enable users to dynamically allocate more IP ranges for pods.
+
+## Motivation
```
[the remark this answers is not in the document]
rahulkjoshi: Ack. Added a section to the summary to call this out explicitly.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+allocated to the cluster (cluster CIDR). Each node gets a range of a fixed size
+from the overall cluster CIDR. The size is specified during cluster startup time
+and cannot be modified later on. This has multiple disadvantages: * There is
+just one cluster CIDR from which all pod CIDRs are allocated. This means that
+users need to provision the entire IP range up front accounting for the largest
+cluster that may be created. This can waste IP addresses. * If a cluster grows
+beyond expectations, there isn't a simple way to add more IP addresses. * The
+cluster CIDR is one large range. It may be difficult to find a contiguous block
+of IP addresses that satisfy the needs of the cluster. * Each node gets a fixed
+IP range for clusters. This means that if nodes are of different sizes and
```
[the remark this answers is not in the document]
rahulkjoshi: done.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,863 @@
...
+
+## Design Details
+
+### New Resource
+
+```
+type ClusterCIDRConfig struct {
+    # An IP block in CIDR notation ("10.0.0.0/8", "fd12:3456:789a:1::/64")
+    # +required
+    CIDRBlock     string
```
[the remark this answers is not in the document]
rahulkjoshi: Ack.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: That sounds good. I reworked the section and added a link to your benchmarks.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Yeah, I reworked to call that out explicitly.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
rahulkjoshi: I was looking at a few other examples and I think the nodeSelector picking a random appropriate range is not unprecedented.
1. Pod's nodeSelector -- In this case the pod will be scheduled on any node that matches the selector. The semantics of pods make it obvious that we would only schedule it once. Similarly, my argument here is that the semantics of a node's PodCIDRs field make it "obvious" that we allocate only a single range from each IP Family
2. Persistent Volumes and Claims (might have misunderstood this) -- The documentation [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#selector) suggests that we filter the PVs applicable to a claim down to only PVs that match the selector and then pick any one which satisfies the claims other requirements

Based  on those examples, I'm thinking that our behavior be as follows:

> When attempting to allocate PodCIDRs for a Node, filter the ClusterCIDRConfigs down to those whose  NodeSelector matches that node. If one of the ClusterCIDRConfigs still has free IPs, allocate from one arbitrarily. Otherwise fall-back to ClusterCIDRConfigs with no selector.

What do you think @danwinship?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
+
+### Share Resources with Service API
```
thockin, line 803: To complete this though (because unifying is so compelling absent the gory details), I'll sketch some more thoughts.  I'm not 100%, but I think I am leaning against this.

Suppose we define `CIDR` as an API type:

```
type CIDR struct {
    Purpose CIDRPurpose // Services, Pods, ...
}
```

The name of the CIDR is the CIDR itself.  E.g.  `10.0.0.0/16`.

To allocate a sub-range, you create a `CIDRClaim` object, such as `10.0.1.0/24`.

The ServiceAllocator finds all CIDRs where `purpose` == `Service`.  It always allocates a /32 or /128 address.

The NodePodIPAMAllocator would need an indirection to specify which nodes can use which ranges and masks.  E.g.

```
type PodCIDRAllocationPolicy struct {
    NodeSelector  // these nodes
    CIDRSelector  // can use these CIDRs
    Mask int     // with this mask size
}
```

That's a fair bit of complexity, and it trades 2 resources (ServiceCIDR, PodCIDR) for 2 resources (PodCIDR, PodCIDRAllocationPolicy).  It doesn't really make anything easier - the controllers are actually MORE complicated now.  And it doesn't actually solve the NodePort allocation problem, so we're stuck with another resource, but different in form and function.

It's not clear to me whether there's any other use case for IP or CIDR allocation or for port allocation that helps to amortize this cost.  If there are other use-cases, the balance could shift.

Ergo, my lean towards "generalizing is not worth the effort".

thockin (commented):
Overall I am pretty OK with this.  I've asked @caseydavenport and @dcbw to look at it from their POV (but anyone is welcome!).

A few notes:

Did we discuss doing a PoC outside of k/k ?

All of the PRR needs to be filled in eventually.  Given the looming deadline, this seems likely to miss.

I think the fact that this RELIES on single-writer needs to be emphasized.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+    # --max-pods flag)
+    # +required
+    PerNodeMaskSize int
+}
+```
+
+#### Expected Behavior
+
+-   Each node will be assigned up to one range from each `FamilyType`.
+    -   **TODO:** what is the tie-break if multiple ranges match
```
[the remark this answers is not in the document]
caseydavenport: Alternatively we leave this to be defined as non-deterministic. In general I don't like doing that but it might be more friendly to say "If multiple CIDRs select this node you better be happy with any one of them winning" than saying "If multiple CIDRs select the node here is a complex way of determining which will win".

Sometimes too much knowledge can be a burden on the user. Plus I think you can't really get away from the arbitrary tie-break in the end - I can imagine users having multiple CIDRs each with the same specificity of selector, same allocation size, and all with space available (e.g., the disjoint CIDR case). I guess at that point you could just break the tie based on numerically smaller / larger address.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+more sophisticated data structure are likely not required.
+
+Prior investgations [here](https://github.com/kubernetes/kubernetes/pull/90184)
+suggest that maps storing allocations will perform well under the number of
+nodes we expect.
+
+#### Startup
+
+-   Fetch list of `ClusterCIDRConfig` and build internal data structure
+-   If node `ClusterCIDRConfig` are specified, fetch the values of the
```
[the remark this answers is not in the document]
aojea: I assume the new controller will be under a featuregate , and enabling it will disable the current ipam controller.
What if the new controller creates itself the objects with the ClusterCIDR config from the flags?
This will keep compatibility and will allow to start to deprecate the flags ...

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+    -   If `PodCIDR` is not set, save Node for allocation in the next step.
+        After processing all nodes, allocate ranges to any nodes without Pod
+        CIDR(s) [Same logic as Node Added event]
+
+#### Reconciliation Loop
+
+This go-routine will watch for cleanup operations and failed allocations and
+continue to try them in the background.
+
+For example if a Node can't be allocated a PodCIDR, it will be periodically
```
aojea, line 340: so, this is for handling a case where we don't have free cidrs, and then a node is deleted so we suddenly have one free cidr that we can use for this node that was waiting ... it is like an infinite requeue thing for the controller , right?

aojea: deleting, i just read the next section :)

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: :upside_down_face:

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: even more simple, what if we delegate the responsibility to the user, using a Status field set by the controller to say: CIDRRange OK or INVALID

The IPAM controller will just not use the CIDRRanges with invalid status, and it has to be the user to delete it and create a correct one ... is kind of the PV/PVC model, but in  one step only ...

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
aojea: > I guess at that point you could just break the tie based on numerically smaller / larger address.

whatever but deterministic please

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+
+We will use maps to store the allocated ranges and which node is using the
+range. Because the number of nodes is expected to be on the order of thousands,
+more sophisticated data structure are likely not required.
+
+Prior investgations [here](https://github.com/kubernetes/kubernetes/pull/90184)
+suggest that maps storing allocations will perform well under the number of
+nodes we expect.
+
+#### Startup
```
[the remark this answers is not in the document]
neolit123: kubeadm follows a "minimal viable cluster" model, that often suits both a basic and an advanced use case.
but the need to switch a cluster to a bigger CIDR can be a problem for all users.

ideally the proposal in this KEP, should ensure backwards compatibility with the existing configuration surface (which today is passing flags to core components) to configure a single CIDR. if some flags will be deprecated / removed, we should follow the official policy and alert users. migration guides should be written, et cetera.

kubeadm will likely continue to deploy one single arbitrary size CIDR by default.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: I've highlighted in the doc that this controller requires a single writer (and therefore probably some sort of leader-election).

I'll add a section in alternatives considered for the allocation objects being stored, so we have this in our back pocket should the need arise.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
rahulkjoshi: Makes sense -- I've updated the design doc to mention that, along with some scale tests by @aojea

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+
+Using this enhancement, the user creates 4 different CIDR configurations each
+with a /21 range. The CIDR allocator allocates a /24 range from any of these /21
+ranges to the nodes and the user can now create the cluster.
+
+### Notes/Constraints/Caveats (Optional)
+
+A major precondition for this feature is to prevent churn on the IPs assigned to
+Nodes. Specifically, we want to prevent the controller from attempting to change
+a Node’s assigned PodCIDRs. For example, K8s will reject attempts to delete a
```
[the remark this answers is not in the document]
rahulkjoshi: I've gone through and added a lot more clarity around how we want to use finalizers. Hopefully that makes sense and is workable.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
rahulkjoshi: I ended up going with @aojea 's suggestion. The controller will create a ClusterCIDRConfig from the flags, which should result in a relatively seamless upgrade. It can use a well-defined name to ensure it only creates it once -- maybe we do a more sophisticated check to see if the flags already overlap with existing objects, not sure if that's necessary.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,882 @@
...
+
+Update internal representation of CIDRs. Every failed Node Allocation is stored
+in a queue, that will be tried again with the new range by the reconciliation
+loop.
+
+##### ClusterCIDRConfig Updated
+
+_`CIDR`, and `PerNodeMaskSize` are immutable_
+
+_Admission Controller: If the change to `NodeSelector` would deselect a node
```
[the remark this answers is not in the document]
rahulkjoshi: Yeah, i've scrapped this idea and instead we'll handle updates to NodeSelector as instructions on future behavior.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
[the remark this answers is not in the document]
rahulkjoshi: For now we plan to keep the flags. I've started outlining this in the Upgrade/downgrade stories, but users would just change their controller type and keep using all the same other configurations.

Maybe down the line (after this new controller and API has gone GA), we can revisit those flags, but I doubt deprecating them is high priority. As @neolit123 points out, we'll probably treat those flags as good starting points for users who "just want a cluster" and eventually the defaults will be setup so that they can seamlessly expand their CIDRs as their needs change.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+    NodeSelector *v1.LabelSelector
+
+    // Netmask size (e.g. 25 -> "/25") to allocate to a node.
+    // Users would have to ensure that the kubelet doesn't try to schedule
+    // more pods than are supported by the node's netmask (i.e. the kubelet's
+    // --max-pods flag)
+    // +required
+    PerNodeMaskSize int
+
+    // Is this a good idea?
```
thockin, line 245: What is the intent of this?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+type ClusterCIDRConfigStatus {
+  Conditions []metav1.Conditions
+}
+
+type ClusterCIDRConfigStatusType string
+type ClusterCIDRConfigStatusReason string
+var (
+  // If the "active" condition is true, this ClusterCIDRConfig can be used by
+  // the controller to allocate PodCIDRs for matching nodes.
+  ClusterCIDRConfigActive ClusterCIDRConfigStatusType = "active"
```
thockin, line 261: Why do we need active/terminating - the finalizer + deletionTimestamp indicates the same thing, no?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+
+#### Expected Behavior
+
+-   Each node will be assigned up to one range from each `FamilyType`. In case
+    of multiple matching ranges, attempt to break ties with the following rules:
+    1.  Pick the `ClusterCIDRRange` whose `NodeSelector` matches the most
+        labels on the `Node`. For example,
+        `{'node.kubernetes.io/instance-type': 'medium', 'rack': 'rack1'}`
+        before `{'node.kubernetes.io/instance-type': 'medium'}`.
+    1.  Break ties arbitrarily.
```
thockin, line 287: maybe first choose the ClusterCIDRRange with the longest mask (smallest alloc) ?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+
+-   An empty `NodeSelector` functions as a default that applies to all nodes.
+    This should be the fall-back and not take precedence if any other range
+    matches. If there are multiple default ranges, ties are broken using the
+    scheme outlined above.
+
+-   `CIDR` and `PerNodeMaskSize` are immutable after creation.
+
+-   The controller will add a finalizer to the ClusterCIDRConfig object when it
```
thockin, line 296: I'm not convinced the conditions are needed

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+nodes, it uses a single range allocated to the cluster (cluster CIDR). Each node
+gets a range of a fixed size from the overall cluster CIDR. The size is
+specified during cluster startup time and cannot be modified later on.
+
+Kubernetes' IPAM capabilities are an optional behavior that comes with
+Kubernetes out of the box. It is not required for Kubernetes to function, and
+users may use alternate mechanisms.
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new
+CIDR allocator that can be controlled by a new resource `ClusterCIDRRange`. This
```
thockin, line 115: At some point I want to discuss the name - it doesn't roll of the tongue :)

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+more sophisticated data structures are likely not required.
+
+Prior investgations [here](https://github.com/kubernetes/kubernetes/pull/90184)
+suggest that maps storing allocations will perform well under the number of
+nodes we expect.
+
+#### Startup
+
+-   Fetch list of `ClusterCIDRConfig` and build internal data structure
+-   If they are set, read the `--cluster-cidr` and `--node-cidr-mask-size` flags
```
thockin, line 409: Under what conditions do we create this?  Always?  Only if there are no other ranges defined?  Only if there is no range that exactly matches the flags?

thockin (commented):
This looks pretty good.  What we need to clarify is whether this can be a CRD + controller or whether it needs to be a builtin.

If we said it had to be a CRD, what could not be made to work?  Could we device some upgrade path from the legacy node allocator to this new controller?

thockin:
Also time to un-"draft" this?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+
+##### Node Deleted
+
+Release said Node’s allocation from the internal data-structure.
+
+If this Node is the last one using a particular `ClusterCIDRConfig` that has
+been slated for deletion, trigger the deletion flow again (so that the finalizer
+is removed and internal data structures are cleaned up).
+
+##### ClusterCIDRConfig Added
```
aojea, line 469: I commented previously that we can do cross validation in the controller instead of the API rest, i.e. a new ClusterCIDR is added that is invalid because it overlaps or any other reason.
We can set the Status to "invalid" or something similar, to inform the user that is wrong, and the controller, will just omit it.
Since they are inmutable, the user will have to delete and create a new one with the correct configuration

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,980 @@
...
+
+-   Verify finalizers and statuses are persisted appropriately
+-   Test watchers
+
+#### End-to-End Tests
+
+-   Run through some sample workflows. Just a few for example:
+    -   Adding a node
+    -   Adding a ClusterCIDRConfig
+    -   Deleting a ClusterCIDRConfig that is in use
```
aojea, line 526: I think the user stories describe good e2e tests, we should have e2e for each user story.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
thockin: We *can* but I am not sure we need to.  If we *ever* want to do this, though, we need some way to mark it as "accepted" so we don't race.  E.g.  This would be bad:

* Add new invalid CIDR
* New node tries to use it
* Controller marks it as invalid

Instead:

* Add new invalid CIDR, state=pending
* New node tries to use it, but can't because state = pending
* Controller marks it as invalid

Again, I am not really sure we NEED to do this - we should think thru what errors it would likely catch.

aojea: I was assuming nodes only will use cidrs with status active

> ClusterCIDRConfigStatusType = "active"

Is not similar on how pvc/pv works?

How do we "protect" the cluster then from overlapping or duplicate cidrs?

thockin: What would make a CIDR "invalid" ?  They are explicitly allowed to overlap (so different nodes can get different sizes from the same range).

aojea: ah,  I wasn't considering that overlapping was allowed, you are right, that was my argument for the cross validation ... 
from the top of my head, a conflict with the service ip range (subnet equal to or included into) or wrong ip families (IPv4 only cluster and IPv6 CIDR) are the only ones that can cause problems

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Acknowledged. Adding.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Whoops, that's me playing around. Didn't mean to upload that.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: It's more verbose, for sure. active/terminating by themselves aren't too helpful, I agree. I think they add value when the reasons come into play -- it would help make it clear what's happening if you can see {"active", "cidr-exhausted"} and you know why we're not allocating from that range. Similarly {"terminating", "cidr-in-use"} might explain why the finalizer isn't being cleaned up -- it help users who don't know this Config that well from worrying the controller is broken.

But I don't feel too strongly about this. So if the preferred guidance is to avoid status shenanigans, I'm fine removing it.

rahulkjoshi:
@thockin 

> What we need to clarify is whether this can be a CRD + controller or whether it needs to be a builtin.
My hope was that this comes built-in as a successor to the current NodeIPAM controller. Making this a CRD and separate controller slightly erodes our story around the ease of use for users -- rather than just changing a flag on your kube-controller-manager, now you're installing a crd and running a new controller (in addition to changing the flag to disable the old NodeIPAM controller).

keps/sig-network/2594-multiple-cluster-cidrs/README.md
thockin: We can always add API if we need it, but we can't easily take it away.

thockin:
On the topic of CRD.

On one hand this is replacing a built-in and we probably want lots of people to use this.

On the other hand, CRDs are supposed to be the path forward for new APIs and Kubernetes shouldn't have been doing IPAM in the first-place.

I know there's work afoot to make CRDs less painful - I suggest we start with this as a CRD to prove it works and then make a case for moving it into core.

Fair?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: I think those are both interesting problems.

1. Overlapping with the service range. The current IPAM controller takes a serviceCIDR as an input and doesn't allocate anything there. Should we continue that same scheme? https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodeipam/node_ipam_controller.go#L79
2. Wong ip family is also tricky. How does Kubernetes communicate single stack vs dual stack to various components. Is there a well-known flag we can plug in to?  or maybe another Kubernetes resource?
    - Maybe we need to give this controller a flag to tell it to give every node a IPv6 address?
    - If we have both ipv4 and ipv6 objects, then we're in dual stack mode. Otherwise, it's a single stack mode.

thockin: Once you have multiple Service CIDRs and multiple Pod CIDRs, does it make sense to try to prevent overlaping?  Especially if both of those operating modes are optional?  I guess that could argue for a single IPAM admission check that understands the different modes and prevents collisions.  Doesn't sound TOO hard?

> If we have both ipv4 and ipv6 objects, then we're in dual stack mode. Otherwise, it's a single stack mode.

I like this.

aojea:
> I know there's work afoot to make CRDs less painful - I suggest we start with this as a CRD to prove it works and then make a case for moving it into core.
> 
> Fair?

is the CRD == alpha, or alpha starts when we move from CRD to builtin?
Do we need to update the graduation criteria ? 

This LGTM, there are some doubts and nits that I'm sure they will be solved once the development starts and we have more users feedback.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,981 @@
...
+logs or events for this purpose.
+-->
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+<!--
+Pick one more of these and delete the rest.
+-->
+
+-   [ ] Metrics
```
aojea, line 725: We already have metrics in the current allocator, it should be good to have similar ones
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/nodeipam/ipam/cidrset/metrics.go#L26-L68

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,981 @@
...
+  metav1.ObjectMeta
+
+  Spec ClusterCIDRConfigSpec
+  Status ClusterCIDRConfigStatus
+}
+
+type ClusterCIDRConfigSpec {
+    // An IP block in CIDR notation ("10.0.0.0/8", "fd12:3456:789a:1::/64")
+    // +required
+    CIDR string
```
mskrocki, line 231: Any chance we could have this object provide both IP families at once? example:
```
IPv4: (optional)
  CIDR
  PerNodeMaskSize
IPv6: (optional)
  CIDR
  PerNodeMaskSize
```
Where user can define one or both families. Reasoning is:
1. Ease of configuration
2. I think with current approach we might end up with race condition for DualStack clusters i.e. the allocator will see only the first IPv4 CIDR object and assign IP from it, but there is another one with the IPv6 and it would not be used

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: How about just CIDRRange?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: I think we can make it the ClusterCIDRRange with the longest mask followed by the picking the lower IP address if there is still a tie.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
sdmodi: If we are using the new node IPAM controller, we should always create this backend data structure. Is there a disadvantage to doing so?

sdmodi (approved):

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,981 @@
...
+  ClusterCIDRConfigInUse ClusterCIDRConfigStatusReason "cidr-in-use"
+)
+
+```
+
+#### Expected Behavior
+
+-   Each node will be assigned up to one range from each `FamilyType`. In case
+    of multiple matching ranges, attempt to break ties with the following rules:
+    1.  Pick the `ClusterCIDRRange` whose `PerNodeMaskSize` is the fewest IPs.
```
mskrocki, line 277: could this be more elaborate and pick first the ClusterCIDRRange with lowest maxAvailableCIDRs in the set?
e.g. `CIDR=x.x.x.x/16, PerNodeMaskSize=16` (max=1) over `CIDR=x.x.x.x/24, PerNodeMaskSize=26`(max=4)

With your current description the second ClusterCIDRRange would be choosen.

aojea:
I don't know if we have considered the overlapping with the service-cidr 
https://github.com/kubernetes/kubernetes/issues/103898#issuecomment-886421151
this is something that hits users from time to time and current code takes into account

rahulkjoshi:
@aojea and @thockin please take another look. Specifically, the "Dual-stack" section and the "startup options" and the "startup" section -- those are what's changed.

I've also fleshed out the Production Readiness section, so after we merge this, we can start that review as well.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: there is a section for dual-stack support https://github.com/kubernetes/enhancements/pull/2594/files#diff-2ca3d193cb87b069c0dbfa36429b5e380af3b02647cd4d22127cb5789b361e4dR359

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,992 @@
...
+range. Because the number of nodes is expected to be on the order of thousands,
+more sophisticated data structures are likely not required.
+
+Prior investgations [here](https://github.com/kubernetes/kubernetes/pull/90184)
+suggest that maps storing allocations will perform well under the number of
+nodes we expect.
+
+#### Dual-Stack Support
+
+Before assigning any Pod IPs to nodes, the controller will check all Nodes for
```
aojea, line 361: I think that you want to mean `PodCIDRs` instead of `Pod IPs`

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,992 @@
...
+dual-stack), then the controller enters that mode.
+
+If there are no Nodes or all their `PodCIDR` allocations are empty, the
+controller chekcks and the list of `ClusterCIDRConfigs` for their IPFamilies. If
+there are both IPv4 and IPv6 Pod CIDRs specified, it enters dual-stack mode,
+otherwise it uses single-stack mode.
+
+The controller will wait until the first Node is added before picking a mode. So
+users who want to specify dual-stack must first add 2 `CluterCIDRConfigs` (one
+each for IPv4 and IPv6). Only after creating those resources should Nodes be added.
```
aojea, line 372: just to confirm, migrating a node from single-stack to dual-stack is not possible, you have single-stack nodes, you have to create new nodes, right?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Dropped the conditions.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,992 @@
...
+_`CIDR`, and `PerNodeMaskSize` are immutable_
+
+Update any internal state as required.
+
+If the `NodeSelector` changed to include a node that is currently awaiting a
+PodCIDR allocation, it will be allocated by the reconiliation loop. 
+
+If the `NodeSelector` changes to unselect a Node currently using that range,
+there is no change to the Node's allocation. Users will have to manually
+restart/recreated that Node to reset it's `PodCIDR` allocation.
```
aojea, line 481: is this worth the effort?
is not better to play with the labels on the nodes instead?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,992 @@
...
+    -   Deleting a ClusterCIDRConfig that is in use
+-   Run through the [user stories](#user-stories):
+    -   Expand the ClusterCIDR (existing nodes without alloations are
+        allocated and new nodes also get ranges.
+    -   Use `NodeSelector` to allocate different sized CIDRs to different nodes.
+    -   Create and use discontiguous ranges.
+
+### Graduation Criteria
+
+The initial release of this feature will be as a CRD and custom controller.
```
aojea, line 530: how the custom controller will inherit the flags of the controller manager?

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,992 @@
...
+However, as part of the alpha -> beta graduation, this feature will be merged
+into the core Kubernetes API.
```
aojea, line 532: gateway friends did this dance recently https://github.com/kubernetes-sigs/gateway-api/pull/707 , we should ask them about the experience
cc: @robscott

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: That's a fair point. I was thinking we could use this for swaps, but since the finalizer will keep the old range around while it's in use, you can delete the old range, create a new one with the same selector and the appropriate CIDR, and then start restarting the nodes.

I guess the only benefit to making it editable is it lets users have more flexibility -- but maybe we should start with un-editable and then open it up if really needed.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: Hmm that's a good point -- I was banking on this being a built-in controller running in kube-controller-manager. We're also rely on kube-controller-manager's leader election so this might be a rather tricky situation.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: That's correct. What I'm unsure of is whether we want to support migrating the cluster from single-stack to dual-stack while nodes are running.

For example, the model proposed here says that if any Node is in single-stack mode, the controller will only allocate single-stack CIDRs to new nodes even if there are IPv4 and IPv6 ClusterCIDRConfigs provided. The only way to migrate your cluster is to add both IPv4 and IPv6 ClusterCIDRConfigs and then delete all nodes. Then once you start adding nodes again, the controller will allocate them in dual-stack mode. Is this an acceptable workflow for us -- I have seen concern about mixing single-stack and dual-stack nodes hence this setup.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
rahulkjoshi: yes, good catch. Fixed.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: I think that this is well encapsulated and the API is relatively simple, so we can avoid the CRD/custom controller , but better ask for permission first to @thockin ;)

keps/sig-network/2594-multiple-cluster-cidrs/README.md
aojea: we can omit this comment if we don't start with CRDs/custom controller https://github.com/kubernetes/enhancements/pull/2594#discussion_r688071399

rahulkjoshi:
@thockin @aojea @mskrocki Let's get this finalized and merged in. The KEP deadline is fast approaching and we still need PRR sign-off. I think they stop accepting new requests soon.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,1027 @@
...
+        },
+        IPv6: {
+            CIDR:            "fd12:3456:789a:1::/64"
+            PerNodeMaskSize: "118",
+        },
+    }
+    ```
+    Only 4  nodes may be allocated from this `ClusterCIDRConfig` as only 4 IPv4
+    Pod CIDRs can be partitioned from the IPv4 CIDR. The remaining IPv6 Pod
+    CIDRs may be used if referenced in another `ClusterCIDRConfig`.
```
mskrocki, line 295: Should we allow redefinition of CIDRs (overlapping)?

rahulkjoshi: I think we should -- there's value in letting users specify a large IPv6 that they reuse across a bunch of IPv4 blocks. Otherwise, users have to start being very deliberate about manually partitioning which seems rather toilsome.

keps/sig-network/2594-multiple-cluster-cidrs/README.md
```diff
@@ -0,0 +1,1027 @@
...
+    RangeAllocator’s current behavior (give each Node a /XX from the Cluster
+    CIDR as its `PodCIDR`)
+*   No change to the default behavior of a Kubernetes cluster.
+    *   This will be an optional API and can be disabled (as today’s NodeIPAM
+        controllers may also be disabled)
+
+## Proposal
+
+This proposal enhances how pod CIDRs are allocated for nodes by adding a new
+CIDR allocator that can be controlled by a new resource 'CIDRRange'. This
```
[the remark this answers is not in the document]
rahulkjoshi: Dang I thought I got all of those typos. Thanks.

mskrocki:
LGTM

thockin (commented):
Thanks!

/lgtm
/approve

PR #2597, opened 2021-04-02
update reducing kubernetes build maintenance status

BenTheElder:
see: https://github.com/kubernetes/enhancements/issues/2420

This KEP is now done except for the passive phase out of older release branches still using bazel, and any remaining cleanup we may consider after that is done.


ameukam:
/lgtm
/hold
Remove hold when ready.

PR #2611, opened 2021-04-08
Add a KEP for supporting AllPorts services.

prameshj:


keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+
+## Design Details
+
+Changes are required to APIServer validation, kube-proxy and controllers that use the ServicePort field.
+
+1) New Validation checks:
+
+   * allPorts can be set to true only for ClusterIP and LoadBalancer services.
+   * Only one ServicePort object per protocol can be specified in the ServiceSpec, if allPorts is set to true.
+
```
aojea, line 111: Can the service mutate from ClusterIP to LoadBalancer?
Is `allPorts` immutable once set?

keps/sig-network/2610-allports-services/kep.yaml
```diff
@@ -0,0 +1,46 @@
...
+kep-number: 2610
+authors:
+  - "@prameshj"
+owning-sig: sig-network
+participating-sigs:
+  - sig-cloud-provider
+status: provisional
+creation-date: 2021-04-08
+reviewers:
+  - "@thockin"
```
aojea, line 11: add me, one of the dans, andrew, or some people from other company to have some diversity

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
```
aojea, line 524: is an alternative to create a pure IP service, ie., it can also be a new Service Type IPOnly, RoutedIP (I'm bad at names :) and we just forward all IP packets to that IP, no matter port of protocol?
 can we rid of the ServicePorts field this way?

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+Today, a Kubernetes Service accepts a list of ports to be exposed by it. It is possible to specify any number of ports, including the entire port range, by enlisting them in the service spec. This can be tedious if the service needs a large number of ports. This KEP proposes to add a new field to the ServicePort spec to allow exposing the entire port range( 1 to 65535). 
+
+## Motivation
+
+There are several applications like SIP apps or RTP which needs a lot of ports to run multiple calls or media streams. Currently the only option is to specify every port in the
+Service Spec. A request for port ranges in Services has been open in https://github.com/kubernetes/kubernetes/issues/23864. Implementing port ranges are challenging since iptables/ipvs do not support port ranges. Hence, the proposal to set a single field in order to expose the entire port range and implement the service clients and endpoints accordingly.
+
+### Goals
+
+* Allow users to optionally expose the entire port range via a Service (of Type LoadBalancer or ClusterIP).
```
jayunit100, line 68: @uablrek ~ can you comment on the kproxy supportability for this ? re: ipvs/iptables/nftables/...

uablrek: That iptables/ipvs doesn't support port-ranges is not entirely true, but they can *not* re-map ports in a range, like with an offset. The ports are unchanged.
From ipvsadm(8) man page;
```
       -A, --add-service
              Add a virtual service. A service address is uniquely defined  by
              a triplet: IP address, port number, and protocol. Alternatively,
              a virtual service may be defined by a firewall-mark.
```
As you can see you must use `firewall-mark` to load-balance a range or an entire ip. The rule to set the fwmark is done by ip(6)tables and can specify port-ranges and protocol (or not).

uablrek:
I would like to see a valid use-case for `ClusterIP`.

I have not developed SIP apps for many years and never RTP but I from contacts with SIP app developers it seem they *always* use headless services because they need control over the clients.

If all-ip services can be limited to `type:LoadBalancer` and headless service as described in https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-813371970 it will simplify in very many ways. I estimate I can implement it myself for ipvs in ~100mh (e2e tests excluded) with a feature-gate in a way that would leave the existing implementation intact. Just avoiding the API change would save considerable time.

The stake-holders should not be asked "do you want all-ip services for ClusterIP?" since the answer would be "yes" (why not), but something like; "are you prepared to delay all-ip service services for ~1y at least, most likely pushed to kube-proxy v2 (which is under way), to get ClusterIP support?"

uablrek:
For `ipvs` firewall-marks (fwmark) must be used (as I have commented). I have made a test (without K8s). I share it below for the discussion.

Setup;
```bash
ipvsadm -A -f 1 -s rr
ipvsadm -A -f 1 -6 -s rr
ipvsadm -a -f 1 -r 192.168.1.1:0 -m
ipvsadm -a -f 1 -6 -r [1000::1:192.168.1.1]:0 -m
# (repeat for all real servers)
iptables -t mangle -A PREROUTING -i eth2 -d 10.0.0.0/24 -j MARK --set-mark 1
ip6tables -t mangle -A PREROUTING -i eth2 -d 1000::/120 -j MARK --set-mark 1
```
Port ranges, protocol etc. are defined with `iptables` with all possibilities iptables provides.

Status after 100 connects;
```
$ ipvsadm -Ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
FWM  1 rr
  -> 192.168.1.1:0                Masq    1      0          25        
  -> 192.168.1.2:0                Masq    1      0          25        
  -> 192.168.1.3:0                Masq    1      0          25        
  -> 192.168.1.4:0                Masq    1      0          25        
FWM  1 IPv6 rr
  -> [1000::1:c0a8:101]:0         Masq    1      0          25        
  -> [1000::1:c0a8:102]:0         Masq    1      0          25        
  -> [1000::1:c0a8:103]:0         Masq    1      0          25        
  -> [1000::1:c0a8:104]:0         Masq    1      0          25        
```

Fwmark becomes a resource that must be handles by `kube-proxy`. However, it is a node-local resource and I see no reason to centralize allocation (as with ipam), it would only complicate things.

keps/sig-network/2610-allports-services/README.md
uablrek: I think this is the right way to implement an all-ip service. Otherwise at least SCTP must be included.

uablrek: I think this is the way to implement all-ip services but still with `type:LoadBalancer` but with no ports.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+
+* Supporting Port Ranges in a Service.
+* Changing the default behavior of Service ports.
+
+## Proposal
+
+The proposal here is to introduce an “allPorts” boolean field to the [servicePort API.](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#serviceport-v1-core)
+This field is introduced within ServicePort rather than ServiceSpec, because a ServicePort entry is needed to infer the protocol.
+
+This field will be supported for Service Type=ClusterIP and ServiceType=LoadBalancer. This is not applicable to ExternalName services. NodePort services are not supported.
```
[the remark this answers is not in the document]
aojea: > At any rate, saying that allPorts LoadBalancer services are not also NodePort services may be an API break...

I think we burnt that bridge here https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/1864-disable-lb-node-ports

keps/sig-network/2610-allports-services/README.md
prameshj: Yes, the service can mutate, allPorts field will be mutable. Since it is not specified in the validation, the default behavior is that mutation is allowed.

keps/sig-network/2610-allports-services/kep.yaml
prameshj: Done, thanks!

keps/sig-network/2610-allports-services/README.md
prameshj: That's a good idea. I wonder if this should be the main approach, rather than an alternative. The one big change is for LoadBalancer controllers to also process services of this new type. @thockin @rikatz any thoughts?

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Today, a Kubernetes Service accepts a list of ports to be exposed by it. It is possible to specify any number of ports, including the entire port range, by enlisting them in the service spec. This can be tedious if the service needs a large number of ports. This KEP proposes to add a new field to the ServicePort spec to allow exposing the entire port range( 1 to 65535).
```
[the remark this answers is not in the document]
prameshj: Done.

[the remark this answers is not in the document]
prameshj: ack.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+
+### Non-Goals
+
+* Supporting Port Ranges in a Service.
+* Changing the default behavior of Service ports.
+
+## Proposal
+
+The proposal here is to introduce an “allPorts” boolean field to the [servicePort API.](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#serviceport-v1-core)
+This field is introduced within ServicePort rather than ServiceSpec, because a ServicePort entry is needed to infer the protocol.
```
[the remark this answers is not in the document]
prameshj: The mixed protocol feature requires specifying more than one port - https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types, so the allPorts validation will need to allow one port per Protocol. Updated this section.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+### Goals
+
+* Allow users to optionally expose the entire port range via a Service (of Type LoadBalancer or ClusterIP).
+
+### Non-Goals
+
+* Supporting Port Ranges in a Service.
+* Changing the default behavior of Service ports.
+
+## Proposal
```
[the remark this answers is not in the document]
prameshj: Will add this.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+Service Spec. A request for port ranges in Services has been open in https://github.com/kubernetes/kubernetes/issues/23864. Implementing port ranges are challenging since iptables/ipvs do not support port ranges. Hence, the proposal to set a single field in order to expose the entire port range and implement the service clients and endpoints accordingly.
+
+### Goals
+
+* Allow users to optionally expose the entire port range via a Service (of Type LoadBalancer or ClusterIP).
+
+### Non-Goals
+
+* Supporting Port Ranges in a Service.
+* Changing the default behavior of Service ports.
```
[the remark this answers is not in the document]
prameshj: Is there a use-case for it? It appears straightforward to add support except for the corner case - if the ExternalIP is actually a local IP on the node.  Then, are opening up all ports on the node to external traffic - I am not sure this is desirable.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,538 @@
...
+2) Populate the protocol field.
+3) Populate a valid port in the port field.
+   This is required to make sure we do not break other Service API clients (like DNS) that might always expect a non-zero port value. The [APIServer validation](https://github.com/kubernetes/kubernetes/blob/b9ce4ac212d150212485fa29d62a2fbd783a57b0/pkg/apis/core/validation/validation.go#L4373) could be relaxed, but we might break other clients.
+
+By default, a NodePort will assigned for the single valid port in the service, if it is of Service Type=LoadBalancer. Traffic Forwarding will work as usual for that NodePort.
+NodePort allocation can be disabled by setting "allocateLoadBalancerNodePorts: false" on the service spec.
+
+### Risks and Mitigations
+
+A user could accidentally expose the entire port range on their cluster nodes, by enabling it for a Service. This is mostly applicable to LoadBalancer services, that are typically
```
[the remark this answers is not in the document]
prameshj: This is mostly to point out that the firewall rules need to be configured appropriately. If someone opened up firewall for nodes to accept traffic to ANY IP on the service ports, they could run into this.

prameshj:
> I would like to see a valid use-case for `ClusterIP`.
> 
> I have not developed SIP apps for many years and never RTP but I from contacts with SIP app developers it seem they _always_ use headless services because they need control over the clients.
> 
> If all-ip services can be limited to `type:LoadBalancer` and headless service as described in [kubernetes/kubernetes#23864 (comment)](https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-813371970) it will simplify in very many ways. I estimate I can implement it myself for ipvs in ~100mh (e2e tests excluded) with a feature-gate in a way that would leave the existing implementation intact. Just avoiding the API change would save considerable time.
> 
> The stake-holders should not be asked "do you want all-ip services for ClusterIP?" since the answer would be "yes" (why not), but something like; "are you prepared to delay all-ip service services for ~1y at least, most likely pushed to kube-proxy v2 (which is under way), to get ClusterIP support?"

Thanks @uablrek . This would still require an API validation change to allow LoadBalancer + ClusterIP: None, but no need for any change to ports validation. After this, allPorts LoadBalancer services will not necessarily be valid ClusterIP services. @danwinship Would that be a breaking change?

uablrek:
Even more narrow; type:LoadBalancer + clusterIP:None + no ports.

keps/sig-network/2610-allports-services/README.md
serverwentdown: A slight variation of the above could be:
```
apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  selector:
    app: MyApp
  type: LoadBalancer
  protocols:
  - TCP
  - UDP
```

uablrek: - SCTP

serverwentdown: Explicitly declaring the protocols would help in cases where the application speaks protocols that are not TCP and UDP. 
```
protocols:
- SCTP
- L2TP
- 4in4
- 6in4
```

serverwentdown: This also makes it a future possibility to allow muti protocol services:
```
ports:
- protocol: TCP
  port: 6789
protocols:
- TCP
- ICMP
```
The default value of `protocols` can be populated from `ports`, (in most cases `[TCP]`)

prameshj: I added the "IPOnly" service type approach and "Loadbalancer + headless service + no ports" approach as alternatives. I think they both are achieving the same thing - Loadbalance at ip-level without specifying ports or protocol. But adding a new type of service is more overhead than modifying existing validation rules if they make sense.

I am not too familiar with the use-cases, but is opening up the loadbalancer service to all protocols problematic? I know iptables allows specifying -p "all" for all protocols and cloud-provider firewall rules should have a way to achieve the same as well. There is the security aspect of opening up all ports and protocols for a given LB IP with a very simple yaml change. 

We could do "LoadBalancer + headless service + no ports + new protocol list field" in that case. The downside is that LoadBalancer services are no longer guaranteed to be valid ClusterIP services.  LoadBalancer healthcheck does not depend on it..there could be other clients that expect a valid ClusterIP?

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
danwinship: I don't know if there is a use case for it; you're the one proposing the feature

danwinship:
> @danwinship Would that be a breaking change?

In some cases in the API, it's OK to loosen validation and allow cases that were previously forbidden. But even that can run into problems if there are clients that need to be able to understand all objects of a given type. So, eg, in the case of Services, any change to LoadBalancer behavior that might cause existing external cloud provider implementations to do the wrong thing (or crash!) should probably be considered breaking.

It might work better to add a new `WholeIPLoadBalancer` type rather than trying to figure out a way to use `LoadBalancer` plus weird signalling via other fields. (Hopefully existing cloud providers / third-party proxy implementations should just ignore services of types they don't recognize...)

It's _possible_ that we won't be able to add all-ports/whole-IP to `v1.Service` in a clean way, and we might have to think about doing it in terms of some new API like `Gateway` instead.

uablrek:
> any change to LoadBalancer behavior that might cause existing external cloud provider implementations to do the wrong thing (or crash!) should probably be considered breaking.

Then to limit allIP services to headless services would be safe since cloud providers would sort them out for now anyway?

keps/sig-network/2610-allports-services/README.md
rikatz: > is an alternative to create a pure IP service, ie., it can also be a new Service Type IPOnly, RoutedIP
I like it :D

> The one big change is for LoadBalancer controllers to also process services of this new type.
I guess once this is supported in alpha, beta, etc cloud controllers should adopt this and maybe we can turn this as a graduation policy, like at least one cloud provider supporting the new type (although gatekeeping with cloud providers support might be much more harder to graduate the feature) 

About the new object type, If this service have the field ExternalIP it's still a LoadBalancer, but from a different type (as Antonio stated, a IPService). Anyway, I'm not sure all cloud providers support the LoadBalancer type without relying on NodePorts.

@thockin Maybe this could also define a subresource called ExternalIP, and with a subresource we can have different RBACs right? So would that be possible to say, for example 'only user X can set ExternalIP subresource in any service object'?

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
prameshj: Yea, I would prefer to mark this as unsupported.

prameshj:
> > any change to LoadBalancer behavior that might cause existing external cloud provider implementations to do the wrong thing (or crash!) should probably be considered breaking.
> 
> Then to limit allIP services to headless services would be safe since cloud providers would sort them out for now anyway?

I will add this topic to the next sig-network meeting for discussion.

keps/sig-network/2610-allports-services/README.md
thockin: In this case a single fwmark can be used to cover all services.  If we allowed arbitrary ranges, we'd quickly run out of marks.  Unless I misunderstand how they are used?

prameshj: How about defining allPorts service as fwmark-service? That doesn't appear to make it a persistent service. iptables rules will set the packet mark. 

If we used the --tcp-service/--udp-service option with the persistent behavior, this might be desirable based on use-case? If the clients are going to use a random port to reach the service, it is probably ok to send packets to the same service port to the same backend?

thockin: The fwmark is a single 32 bit int as far as IPVS is concerned - no bitmask.  So while there are a lot of them, it collides with anyone else who uses fwmark.  We'd need to allocate a DURABLE mark per-service and add iptables rules for them, then consume them in ipvs.  They have to survive kube-proxy restarts because ipvs is not atomic (like iptables-restore).  

The persistent feature, if I understand is 3-tuple not 5-tuple.  Maybe that is OK?  It seems surprising...

prameshj: true.. managing the mask can be messy, since we would also need to deal with conflicting mark values used by other components (Calico, other networking addons).

I missed that the persistent feature is 3-tuple, not 5.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,553 @@
...
+In order to expose the entire port range on a supported service, a user needs to:
+1) Create a single ServicePort entry in the ServiceSpec and set "allPorts" to True.
+2) Populate the protocol field.
+3) Populate a valid port in the port field.
+   This is required to make sure we do not break other Service API clients (like DNS) that might always expect a non-zero port value. The [APIServer validation](https://github.com/kubernetes/kubernetes/blob/b9ce4ac212d150212485fa29d62a2fbd783a57b0/pkg/apis/core/validation/validation.go#L4373) could be relaxed, but we might break other clients.
+
+It is possible to specify more than 1 port in the serviceSpec when using the [Mixed Protocol Feature](https://kubernetes.io/docs/concepts/services-networking/service/#load-balancers-with-mixed-protocol-types). There will one ServicePort object per protocol.
+By default, a NodePort will assigned for each valid port(one per protocol) in the service, if it is of Service Type=LoadBalancer.
+Traffic Forwarding will work as usual for those NodePorts.
+NodePort allocation can be disabled by setting "allocateLoadBalancerNodePorts: false" on the service spec.
```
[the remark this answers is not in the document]
uablrek: I think `allocateLoadBalancerNodePorts: false` will only prevent _automatic_ allocation of nodePorts. If you define an explicit nodePort it will be allocated.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,553 @@
...
+  Traffic redirection to endpoints from service VIP will be IP-based, without a port or protocol filter.
+
+  This requires modifying existing LoadBalancer controller to also provision resources for this type of Service.
+  Adding a new service type is quite a bit of API overhead.
+
+*  Restrict AllPorts support to LoadBalancer services only. Allow LoadBalancer services to be headless [by relaxing the validation the check.](https://github.com/kubernetes/kubernetes/blob/036cab71a6faefa84b10a199a61bcdc38e3572c3/pkg/apis/core/validation/validation.go#L4177)
+   Default behavior is to allow traffic for all Protocols. A list of allowed protocols can be specified in a Protocols list(new field) in the ServiceSpec.
+
+   This approach breaks the assumption that all LoadBalancer services are valid ClusterIP services.
+
```
aojea, line 546: funny alternative, create a kubectl plugin that translate the range to a service object with all ports in that range :upside_down_face:

keps/sig-network/2610-allports-services/README.md
andrewsykim: I think the port forwarding behavior might be do-able with ipvs using ipsets, but I would need to play around with it a bit to be sure.

I think a larger problem with IPVS could be regarding it's "bind-all" behavior (https://github.com/kubernetes/kubernetes/issues/72236), where a all ports Service will conflict with a real host port. Long standing issue that we need to address, but have not come up with a good path forward yet.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,553 @@
...
+The proposal here is to introduce an “allPorts” boolean field to the [servicePort API.](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#serviceport-v1-core)
+This field is introduced within ServicePort rather than ServiceSpec, because a ServicePort entry is needed to infer the protocol.
+
+This field will be supported for Service Type=ClusterIP and ServiceType=LoadBalancer. This is not applicable to ExternalName services. NodePort services are not supported.
+A NodePort service accepts traffic on a given port on the Node, redirecting it to the specified targetPort of the service endpoints. Support AllPorts for a NodePort service means -
+traffic to any port on the node, will be forwarded to its endpoints on the same port. This could potentially break networking on the node, if traffic for, say, port 22 got forwarded to \<endpoint IP\>:22.
+
+In order to expose the entire port range on a supported service, a user needs to:
+1) Create a single ServicePort entry in the ServiceSpec and set "allPorts" to True.
```
andrewsykim, line 88: Curious if we've explored adding an `endPort` field, similar to what we added for port ranges for NetworkPolicy. Are there use-cases for _all ports_ or just a really large range of ports?

andrewsykim (commented):
Need to double-check this, but there could be an issue with IPVS regarding its current "bind-all" behavior (kubernetes/kubernetes#72236), where an all ports Service will conflict with a real host port bindings. Long-standing issue that we need to address, but have not come up with a good path forward yet.

keps/sig-network/2610-allports-services/README.md
uablrek: I have assumed it is more "entire IP" than "all ports". I.e no protocol, just L3.

thockin:
FTR: This is exceedingly unlikely to make 1.22 - too many open issues

keps/sig-network/2610-allports-services/README.md
prameshj: The use-case is mostly got a large range of ports rather than allPorts. Exposing the entire port range + protocols (i.e making it an IP-level LoadBalancer service) simplifies some aspects around iptables rules, env variables for service ports, SRV records etc.

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
prameshj: I agree.. the multiple restrictions are confusing. I think the next step here is to 1) Edit this KEP with the LoadBalancer+ Headless service proposal. 2) Any bigger changes should probably be part of Gateway L4.
I will refactor this KEP.

keps/sig-network/2610-allports-services/README.md
prameshj: yup! That will work until the port list makes the manifest too big. 

I just checked that upto 10k ports are being accepted fine.

wojtek-t: I'm not a fan of that - we should avoid increasing size of object if possible (also - kubectl is not the only way to iterate with apiserver, so kubectl plugin is not enough imho).

aojea: I thought about this more as a POC to gauge the interest, I went through the original issue and I don't have clear that this is the real demand, I'd love to get a better description of the problem, not of the solution.
Some of the use cases explained are based on singleton "services", some people say that they use hostNetwork pods to solve the problem. 
A quick look to the most common deployments of webrtc/media servers only show active/passive, so it will not benefit from a service port forwarding at all, unless you use sticky, but again, that is not the best solution for this problem

wojtek-t: So what you were saying is that instead of any changes that are proposed here, just expand on kubectl side?
[I guess I didn't fully understand that.]
So gather better understanding of what exactly people need?

If so - that makes more sense, but I'm not sure if we would actually get the better understanding of it thanks to it. Maybe we need to make a research instead?

aojea: > So what you were saying is that instead of any changes that are proposed here, just expand on kubectl side?
> [I guess I didn't fully understand that.]
> So gather better understanding of what exactly people need?

DanWinship's comments in the issue are spot on https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-435957275
I went through the thread several times and I still didn't find the use case :)

Any solution exposed here is very complex and add alot of dependencies ... to an already complicated and overloaded object like Services.
I mean, if this was something simple as: "add new field PortRange" and ready to go ... but is it worth it?

prameshj: I see your point @aojea. If providing automation to translate port range into list of service ports is sufficient, maybe we should start there. Adding thousands of ports does come with overhead of SRV records and service port env variables which are probably going to be unused. But if folks are already using Services with such large number of ports, automating it isn't introducing a new issue. Let me update the issue - https://github.com/kubernetes/kubernetes/issues/23864

prameshj: I added a few survey questions to get more data on use-cases. https://forms.gle/ig2KUkUF3F8a23hs9 @aojea any other questions we should include here? I will post this link in the issue once we have finalized it.

aojea: > DanWinship's comments in the issue are spot on [kubernetes/kubernetes#23864 (comment)](https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-435957275)
> I went through the thread several times and I still didn't find the use case :) I just wanted to check if after adding the command and we close the issue it gets reopened

@danwinship can you PTAL to the form and the questions?

> I added a few survey questions to get more data on use-cases. https://forms.gle/ig2KUkUF3F8a23hs9 @aojea any other  questions we should include here? I will post this link in the issue once we have finalized it.

uablrek: I still think my proposal is valid https://github.com/kubernetes/kubernetes/issues/23864#issuecomment-813371970.

This will have none of the problems discussed, e.g with SRV records or env variables, etc. It is a clean extension, easy (well, fairly easy) to implement.

It solves external access to allIP only, but that is what's missing. The use-case is not to send RTP/SIP within the cluster between PODs, and besides a headless service can be used for that, the same as for external traffic with the proposed extension.

aojea: > The use-case is not to send RTP/SIP within the cluster between PODs,

but what if it does and we implement an incomplete solution?
My point is,  are we forcing the API to solve something very specific that maybe we can solved with an external controller that generates the endpoints directly, or with some tooling in kubectl, ... or is this a general problem that really requires an API change? what goes to api has to be supported like forever and enforced by tests 

I think that the form is a great idea as next step, can you check if the questions are ok or we miss something?

uablrek: The function is not incomplete IMHO, it rather adds external access for headless services. If the DNS lookup could be used from and external machine like from a POD, the problem would be solved. But that is impossible, so an allIP function for type loadBalancer allows external access, but in a slightly different way.
And any discussion about nodePorts is a dead-end. That does not work. <-period. So AWS must ignore type:LoadBalncer for headless services. However _that_ may be a problem if they don't.

prameshj: It will be helpful to get this data from the wider group of folks asking for port range support in https://github.com/kubernetes/kubernetes/issues/23864

If we get consensus that LB + Headless + AllPorts is sufficient for their usecase, it will help establish this approach as widely applicable. What do you think @uablrek? Can you take a look at  https://forms.gle/ig2KUkUF3F8a23hs9? I am planning to send this survey on the issue. Any other questions we need to include?

uablrek: @prameshj lgtm. Perhaps ask what cloud-provider they are using or private-cloud (own installation). Cloud-providers that _require_ NodePort load-balancing can not use an full-IP service without some changes.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,553 @@
...
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
```
howardjohn, line 38: Can we look into how implementable this is by common networking implementations? CNIs (calico, cilium, etc), service meshes (linkerd, istio), and load balancer implementations (GKE/AWS, metallb, etc) are all things that could plausible have a very hard time supporting this.

As an Istio maintainer, I can tell you this would be a challenge to implement for us.

keps/sig-network/2610-allports-services/README.md
prameshj: Thanks @uablrek , added. @danwinship Do you recommend any other additions to https://forms.gle/ig2KUkUF3F8a23hs9?

keps/sig-network/2610-allports-services/README.md
prameshj: Thanks for raising this @howardjohn. I would like to get more clarity on the use-cases for the AllPorts feature to figure out the actual proposal here and revisit this.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,487 @@
...
+* Allow users to optionally expose the entire port range via a Service (of Type LoadBalancer or ClusterIP).
+
+### Non-Goals
+
+* Supporting Port Ranges in a Service.
+* Changing the default behavior of Service ports.
+
+## Proposal
+
+The proposal here is to introduce an `allPorts` boolean field to the [service API.](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#servicespec-v1-core)
```
khenidak, line 81: The new field has to be `*bool` not `bool`.  Also it would make the KEP alot relatable if you can add
1. sample service.
2. upgrade downgrade rules. e.g. what will happen to existing services if the user upgraded to newer version with this feature enabled? .. similarly what will happen to services when the feature is turned of?
3. The behavior re conversions. Services are have some weirder behavior (than most other api objects) . Example headful->headless, ClusterIP->LoadBalancer. etc. The list will help minimizing surprises during development of code.

keps/sig-network/2610-allports-services/README.md
khenidak: Services of type `allPorts` don't have node ports. So it is primarily
1. `ClusterIP:port` which will be dnated to `pod:port` 
2. `ExternalIP:port` which will be dnated to `pod:port`

aojea: loadbalancers with no nodeports? those are a thing with https://github.com/kubernetes/enhancements/issues/1864

uablrek: There should just be a few allPorts services in a cluster, say max 16, and if fwmark is used only for those it should be managable. It is also a per-node resource so kube-proxy can manage them locally, e.g re-read after re-start. See also https://github.com/kubernetes/enhancements/pull/2611#issuecomment-817108069.

uablrek:
Please see https://github.com/kubernetes/enhancements/pull/2611#discussion_r684024484 and the rest of the thread.

For `ipvs` fwmark becomes a resource that must be managed and is limited as @thockin pointed out. Because of this I think there must be some cluster-wide config parameter that limits the number of allport services.

aojea:
> For `ipvs` fwmark becomes a resource that must be managed and is limited as @thockin pointed out. Because of this I think there must be some cluster-wide config parameter that limits the number of allport services. Perhaps defaults to zero (not many cluster need this I think).

https://github.com/kubernetes/enhancements/pull/2611#discussion_r629454337

Andrew commented on the thread about using ipsets/iptables,  can we use iptables for this feature too?
It is a very simple iptable rules vs a very complex fwmark/ipvs combo

uablrek:
I can't see any way to configure ipvs for multiple (all) ports without using fwmark and from https://github.com/kubernetes/enhancements/pull/2611#discussion_r628565966 it looks like @thockin came to the same conclusion. The forwarded ports can be limited by ipset/iptables rules, but fwmark is necessary for multi-port loadbalancing with ipvs afaik.

Or do you mean that the ipvs proxier shall use iptables only (not ipvs) for multi(all) port services.

aojea:
> Or do you mean that the ipvs proxier shall use iptables only (not ipvs) for multi(all) port services?

this ^^^ ,  
```
-A PREROUTING -d SVC_IP -j DNAT --to-destination POD_IP
```

uablrek:
And the "probablility-ladder" for load-balancing I assume, a'la;
```
KUBE-SEP-ZEA5IFF4B6XMVVC6  all  --  0.0.0.0/0            0.0.0.0/0            /* default/mserver-preferdual:mconnect */ statistic mode random probability 0.25000000000
KUBE-SEP-24KO6U27VFOTKCKY  all  --  0.0.0.0/0            0.0.0.0/0            /* default/mserver-preferdual:mconnect */ statistic mode random probability 0.33333333349
KUBE-SEP-PB5AVSXCPATU3XAH  all  --  0.0.0.0/0            0.0.0.0/0            /* default/mserver-preferdual:mconnect */ statistic mode random probability 0.50000000000
KUBE-SEP-SUDVAOE6QPJL34Q7  all  --  0.0.0.0/0            0.0.0.0/0            /* default/mserver-preferdual:mconnect */
```

uablrek:
If stake-holders would agree to limit this to `externalTrafficPolicy: Local` this could be _really_ simple :smile:

prameshj:
> If stake-holders would agree to limit this to `externalTrafficPolicy: Local` this could be _really_ simple 😄

How does Local Traffic Policy make things easier? Looks like there are 2 options here:

1) Use ipset/iptables to do the DNAT implementation
2) Accept the behavior in IPVS where a 5-tuple match will forward packets to the same backend.

Option 1) seems easy enough, WDYT @uablrek @andrewsykim ?

uablrek:
> How does Local Traffic Policy make things easier?

My thought was that it should be just a simple DNAT rule as in https://github.com/kubernetes/enhancements/pull/2611#issuecomment-895061013, but I forgot that there may be many local pods. Please ignore the comment.

uablrek:
I would not dismiss `fwmark` for ipvs just yet. Please see https://github.com/kubernetes/enhancements/pull/2611#discussion_r684024484;

> There should just be a few allPorts services in a cluster, say max 16 (a small confugurable number?), and if fwmark is used only for those it should be managable. It is also a per-node resource so kube-proxy can manage them locally, e.g re-read after re-start. See also #2611 (comment).

My point is that all-ip services are expected to be very few, im many cases just one I believe. I agree that handling fwmark for all services is horrible (not possible), but for a very limited number of services I think it is ok.

uablrek:
A all-port service limit can be per-proxier, i.e a parameter in the "ipvs" section of kube-proxy config. In the kep it can be noted something as;

> Some kube-proxy modes or kube-proxy replacements (cilium, calico, kpng, ect) may put a restriction on the number of all-ip services.

prameshj:
> I would not dismiss `fwmark` for ipvs just yet. Please see [#2611 (comment)](https://github.com/kubernetes/enhancements/pull/2611#discussion_r684024484);
> 
> > There should just be a few allPorts services in a cluster, say max 16 (a small confugurable number?), and if fwmark is used only for those it should be managable. It is also a per-node resource so kube-proxy can manage them locally, e.g re-read after re-start. See also #2611 (comment).
> 
> My point is that all-ip services are expected to be very few, in many cases just one I believe. I agree that handling fwmark for all services is horrible (not possible), but for a very limited number of services I think it is ok.

I am wary of adding an upper limit, especially if it can be avoided. Is the clientIP affinity behavior an issue for AllPorts service clients?

keps/sig-network/2610-allports-services/README.md
prameshj: resolving this, since this proposal is not creating a new service type anymore.

keps/sig-network/2610-allports-services/README.md
prameshj: Captured this in the spec, resolving.

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
prameshj: Is this referring to https://github.com/kubernetes/kubernetes/issues/97076, CVE-2020-8554? Allowing allPorts with externalIP does make the scope of the CVE wider, in theory. One could MITM requests on any port to a particular externalIP by setting `allPorts` to true. I added this as unsupported.

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
prameshj: Resolving, since this is not applicable with the current proposal.

keps/sig-network/2610-allports-services/README.md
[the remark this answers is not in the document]
prameshj: I have refactored the KEP and marking this as resolved.

keps/sig-network/2610-allports-services/README.md
prameshj: Done.

keps/sig-network/2610-allports-services/README.md
prameshj: We now have a somewhat fleshed out proposal - I have marked this as an "Open question" so we can iterate after merging this initial PR.

keps/sig-network/2610-allports-services/README.md
prameshj: Also, from a survey that was sent out a couple of months back, 70% of responders indicated that  IP-level loadbalancing solves their usecase - https://docs.google.com/presentation/d/1FO9H55-gnDh2RIqOZMDoP4OPVbaIaPhKl9C5to4uNNE/edit#slide=id.gdf6ff10943_0_56

prameshj:
> Need to double-check this, but there could be an issue with IPVS regarding its current "bind-all" behavior ([kubernetes/kubernetes#72236](https://github.com/kubernetes/kubernetes/issues/72236)), where an all ports Service will conflict with a real host port bindings. Long-standing issue that we need to address, but have not come up with a good path forward yet.

Good point, I think this conflict can get exacerbated, since connectivity to the clusterIP service will be allowed on any port. I added it as a risk.

prameshj:
@aojea @uablrek @andrewsykim @thockin @khenidak @danwinship - I have refactored the KEP and moved the 2 open items into a separate section. Also added user stories/more details on service transitions. PTAL when you get a chance. I would like to merge this and iterate in a followup.

uablrek:
> > Need to double-check this, but there could be an issue with IPVS regarding its current "bind-all" behavior ([kubernetes/kubernetes#72236](https://github.com/kubernetes/kubernetes/issues/72236)), where an all ports Service will conflict with a real host port bindings. Long-standing issue that we need to address, but have not come up with a good path forward yet.
> 
> Good point, I think this conflict can get exacerbated, since connectivity to the clusterIP service will be allowed on any port. I added it as a risk.

The problem in ([kubernetes/kubernetes#72236](https://github.com/kubernetes/kubernetes/issues/72236)) is that the clusterIP or loadBalancerIP is actually assigned to an interface `kube-ipvs0` in main netns. An all-port (full-ip) service does not make it worse, rather better since all traffic to the address will be directed to the POD so for instance port 22 im main netns will **not** be exposed.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,569 @@
...
+
+2) A client(1.2.3.4) sends a request to a.b.c.d:8888 - it is received on a cluster node where:
+
+  a) firewall rules allow it
+
+  b) kube-proxy iptables rules DNAT it to one of the backend pod IPs - p.q.r.s (pod1)
+
+  c) request is received on pod1 with source IP - 1.2.3.4 and destination p.q.r.s:8888
+
+  d) pod1 responds directly to 1.2.3.4.
```
aojea, line 148: It seems that we can get rid of externalTrafficPolicy=local too and simplify this even more

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,569 @@
...
+#### Alpha
+
+- Add a new field `allPorts` to Service, but it can only be set when the feature gate is on.
+
+
+#### Alpha -> Beta Graduation
+
+- Ensure that the main clients of Service/Endpoints API - kube-proxy, kubelet, loadbalancer controllers have added support to handle empty ports. All supported node versions for the given master version(that graduates `allPorts` to Beta) should handle this case correctly.
+- Tests are in Testgrid and linked in KEP
+
```
aojea, line 242: do we want to gate on implementation of the feature on external components and cloud providers? 
_at least X major cloud provider implements the feature_

aojea:
I'm +1 on this approach, we heavily discussed this and explored a lot of options and think that at this point we need some real feedback, feature gates are great here since having it in alpha protects and give us time to fix mistakes

keps/sig-network/2610-allports-services/kep.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - "@aojea"
+  - "@danwinship"
+  - "@uablrek"
+  - "@bowei"
+  - "@freehan"
+approvers:
+  - "@thockin"
+prr-approvers:
+  - "@johnbelamaric"
```
wojtek-t, line 20: please reassign to me - John seem to be overloaded this cycle.

Also - you will need to fill in the necessary PRR questions for alpha (not sure if you want to do that in this PRR or in a once that will move it move implementable.

prameshj: Thanks Wojciech. I have reassigned to you.. This will not be targeting alpha in 1.23, most likely the next release.

wojtek-t: OK - then I will get back to it after feature freeze.

uablrek:
I have used the simplified setup proposed by @aojea in https://github.com/kubernetes/enhancements/pull/2611#issuecomment-895061013 as example in a proposed blog post about [kpng](https://github.com/kubernetes-sigs/kpng); https://github.com/kubernetes/website/pull/29783

uablrek:
I suggest to close this KEP and implement all-ip services off-tree as a custom load-balancer with kpng as indicated in https://kubernetes.io/blog/2021/10/18/use-kpng-to-write-specialized-kube-proxiers/

The ClusterIP thing complicated things as I pointed out in https://github.com/kubernetes/enhancements/pull/2611#issuecomment-817106608;

> The stake-holders should not be asked "do you want all-ip services for ClusterIP?" since the answer would be "yes" (why not), but something like; "are you prepared to delay all-ip service services for ~1y at least, most likely pushed to kube-proxy v2 (which is under way), to get ClusterIP support?"

And here we are almost a year later and the progress on this feature is basically halted.

Considering the interrest and :+1: 's in https://github.com/kubernetes/kubernetes/issues/23864 this _is_ a highly wanted feature. However to please everybody, like Istio and nodePort-only clusters (AWS) will stall implementation forever.

IMHO it's better to write a custom proxier that would be useful for the wast majority of users.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+The [APIServer validation](https://github.com/kubernetes/kubernetes/blob/b9ce4ac212d150212485fa29d62a2fbd783a57b0/pkg/apis/core/validation/validation.go#L4162) that disallows empty `ports` will be relaxed, when `allPorts` is set.
+
+The value of `allPorts` field can be toggled on supported service types.
+
+### Supported service types
+
+Setting this field will be supported for:
+
+* ServiceType=ClusterIP
+* ServiceType=LoadBalancer
```
sftim, line 95: How will this work with cloud load balancers that only forward to an assigned port and / or only listen on an assigned port?

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+
+1) A service with ClusterIP a.b.c.d is configured with `allPorts` set to true. This service has 2 endpoints - pod1, pod2.
+
+2) A client(1.2.3.4) sends a request to a.b.c.d:8888 - it is received on a cluster node where:
+
+  a) firewall rules allow it
+
+  b) kube-proxy iptables rules DNAT it to one of the backend pod IPs - p.q.r.s (pod1)
+
+  c) request is received on pod1 with source IP - 1.2.3.4 and destination p.q.r.s:8888
```
sftim, line 146: This is assuming TCP, but we should be explicit.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+   There is a warning in kube-dns for [empty port services](https://github.com/kubernetes/dns/blob/077a43e83e648ba5f04bae18ffcb824edc9db967/pkg/dns/dns.go#L320) that could be removed after GA of AllPorts.
+
+6) There will be no environment variable of the form "_SERVICE_PORT" for services with `allPorts` set. This [codepath](https://github.com/kubernetes/kubernetes/blob/e1f971d5c2a1002c4e90471d064f87f297740aba/pkg/kubelet/envvars/envvars.go#L48) currently assumes non-zero ports and that will be updated.
+
+#### Open Questions to be resolved:
+
+1) How should IPVS implementation be handled?
+Options are to a) use iptables for DNAT b) use ipvs rules that results in same 5-tuple requests being sent to the same backend pod. c) create allPorts services as "fwmark-service" and assign a unique mark for each of them.
+
+2) Identfy how Service Mesh(Istio)/Calico/MetalLB can support AllPorts.
```
sftim, line 214: Can I have a UDP all-ports Service and a TCP all-ports Service on the same ClusterIP?

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+* To verify that all users(kube-proxy, kubelet, various controllers) of Service/Endpoints can handle empty ports.
+* To check the default value of `allPorts` on each type of Service.
+
+E2E tests:
+
+* To verify that default behavior of `allPorts` does not break any existing e2e tests.
+* Test setting `allPorts` on a new Service and connecting to the service VIP on a few different ports.
+* Test setting `allPorts` on an existing Service and connecting to the service VIP on a few different ports.
+* Test unsetting `allPorts` on a service and specifying a single port allows traffic only for that port.
+* Test setting `allPorts` explicitly to false.
```
sftim, line 230: We should test with multiple protocol families, not just TCP.

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+
+Today, a Kubernetes Service accepts a list of ports to be exposed by it.
+It is possible to specify any number of ports (as long as the service is within the max object size limit), by listing them in the service spec.
+This can be tedious if the service needs a large number of ports.
+This KEP proposes to add a new field to the Service spec to allow exposing the entire port range(1 to 65535).
+
+## Motivation
+
+There are several applications like SIP/RTP/Gaming servers that need a lot(1000+) of ports to run multiple calls or media streams.
+Currently the only option is to specify every port in the Service Spec. A request for port ranges in Services has been open in https://github.com/kubernetes/kubernetes/issues/23864. Implementing port ranges are challenging since iptables/ipvs do not support remapping port ranges. Also, in order to specify several non-contiguous port ranges, the user will have to expose the entire valid port range. Hence, this proposal to set a single field in order to expose the entire port range and implement the service clients and endpoints accordingly.
```
sftim, line 69: I think nftables _can_ do this remapping. If that's right, another way to view the restriction is that kube-proxy doesn't come with a way to access that remapping (due to accessing netfilter via the older iptables API).

sftim (commented):
Some thoughts

keps/sig-network/2610-allports-services/README.md
```diff
@@ -0,0 +1,570 @@
...
+[A survey](https://docs.google.com/forms/d/1FOOG2ZoQsnJLYAjnhEtSPYmUULWFNe88iXR7gtFcP7g/edit) was sent out to collect the use-cases for AllPorts support - [results.](http://tiny.cc/allportsslides)
```
cresta-dongzhao, line 70: This link is not reachable any more.

neshkeev:
Is there any progress on the topic? Will Services ever allow to open a range of ports at once?

thockin:
There has not been progress.  IIRC the main sticking points for "all ports" are:

1) The best way to express "all ports" would be to define no `Ports[]` entries or `Ports[].Port = 0`, but that's a breaking change, from a certain POV (I think we could get past this)
2) IPVS does not have a clear way to implement this feature
3) Service Mesh support seems suspect

It may be better to define this in the L4 Gateway model.

PR #2616, opened 2021-04-09
Graduate Indexed Job to beta

alculquicondor:
Ref #2214

/sig apps

### DIFF
--- a/keps/sig-apps/2214-indexed-job/README.md
+++ b/keps/sig-apps/2214-indexed-job/README.md
@@ -155,12 +155,22 @@ semantics, as opposed to Jobs.
 
 ### Risks and Mitigations
 
-Jobs have a known issue in which more than one Pod can be started even if
-parallelism and completion are set to 1 ([reference]). In the case of indexed
-Jobs, this translates to more than one Pod having the same index.
+- More than one pod per index
+  Jobs have a known issue in which more than one Pod can be started even if
+  parallelism and completion are set to 1 ([reference]). In the case of indexed
+  Jobs, this translates to more than one Pod having the same index.
 
-Just like for existing Job patterns, workloads have to handle duplicates at the
-application level.
+  Just like for existing Job patterns, workloads have to handle duplicates at the
+  application level.
+
+- Jobs with a high number of parallelism produce starvation on small jobs
+  
+  This problem is not unique to Indexed Jobs, but the new API might motivate
+  use cases with higher degree of parallelism.
+  
+  In a Job sync, the controller will be limited to create or delete up to 500
+  Pods. The controller processes the remaining operations in subsequent syncs,
+  which it schedules with no delay.
 
 [reference]: https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures
 
@@ -212,7 +222,7 @@ type JobStatus struct {
 
   // CompletedIndexes holds the completed indexes when .spec.completionMode =
   // "Indexed" in a text format. The indexes are represented as decimal integers
-  // separated by commas. The numbers are listed in increasing order. Two or
+  // separated by commas. The numbers are listed in increasing order. Three or
   // more consecutive numbers are compressed and represented by the first and
   // last element of the series, separated by a hyphen.
   // For example, if the completed indexes are 1, 3, 4, 5 and 7, they are
@@ -335,15 +345,15 @@ gate enabled and disabled.
 #### Alpha -> Beta Graduation
 
 - Complete features:
-  - Indexed Jobs when tracking completion without lingering Pods
+  - Indexed Jobs when tracking completion with finalizers.
     [kubernetes/enhancements#2307](https://github.com/kubernetes/enhancements/issues/2307).
     
     Keeping the size of .status.completedIndexes is desirable to reduce load
     on watchers. We will evaluate holding of from counting completed Pods that
     have an outlying index. That is, contiguous indexes would be counted first.
     This allows to keep the size of the compressed list small.
-- Gather feedback from end users and operators' developers. Open questions:
-  - Are stable Pod names necessary?
+  - Add metrics.
+- Enable feature gate IndexedJob by default.
 - Tests are in Testgrid and linked in KEP
 
 #### Beta -> GA Graduation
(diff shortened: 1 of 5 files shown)


keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -424,64 +424,60 @@ _This section must be completed when targeting beta graduation to a release._
 
 * **What specific metrics should inform a rollback?**
 
+  - job_sync_duration_seconds shows significantly more latency for Indexed Jobs.
+  - job_sync_total shows errors for Indexed Jobs only.
+  - job_finished_total shows that Indexed Jobs don't finish.
```
soltysh, line 429: Any particular reasons why these metrics will be for Indexed Jobs only? IMO it seems reasonable to expose them for all jobs.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -424,64 +424,60 @@ _This section must be completed when targeting beta graduation to a release._
...
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+
+  - per-day percentage of job_sync_total with label result=error <= 1%
```
soltysh, line 463: I'd like to have clear information what will be counted as an error towards this SLO? For example, I wouldn't include update conflicts as errors. Maybe it's reasonable to either explain what is an error for this metric?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -424,64 +424,60 @@ _This section must be completed when targeting beta graduation to a release._
...
+  - 99% percentile over day for job_sync_duration_seconds is:
+    - <= 5s for single pod Jobs 
+    - <= 1m for parallel jobs with a few tens of pods.
```
soltysh, line 466: Do you have any metrics already? Few tens of pods sync shouldn't be longer than 10s.

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -424,64 +424,60 @@ _This section must be completed when targeting beta graduation to a release._
...
+    - <= 1h for parallel jobs with a few thousands of pods.
```
soltysh, line 467: That 1h is definitely too long for a single sync.

soltysh (commented):
I left a few questions.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I don't plan to add them just for Indexed Jobs, but there will be a label for the mode.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: This is based off the pod creation SLOs, which are at 1s per pod https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos

In general, these number are very conservative, as we don't know in which kind of machine the control plane will be running.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I was just thinking of whether the `syncJob` method returns error. But we can be more restrictive. We can discuss in SIG meeting.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: not if you consider https://github.com/kubernetes/community/blob/master/sig-scalability/slos/slos.md#steady-state-slisslos

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: We didn't get a chance to discuss this in the meeting.

I'm thinking that this metric is a tool for cluster operators to quickly see if things as working as expected in dashboards and easily set alerts.

Let's say there is a rogue controller that is updating job pods, causing conflicts all the time. If we silence Conflict errors, an operator might think that the cluster is operating normally, as no alerts are triggered.

We should not think of this metric as "there is a bug in the controller" (although it can also be used for that), but about making sure that the controller is progressing and there are no agents degrading its operations.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Updated to reflect that users can use a label to compare Indexed to NonIndexed

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: From @wojtek-t 

> What exactly "job_sync_duration_seconds" is reporting?

> Is it reporting the duration of a single "sync in a controller"? Or processing a job?
If the former, it should be fairly fast right? [I'm assuming we're creating pods themselves asynchronously?]

It is reporting a single sync in a controller. We create/delete pods in parallel, but we wait for all of them to succeed. What would be the recommendation? Can we simply assume that all of them would finish within 1s?

wojtek-t: > It is reporting a single sync in a controller. We create/delete pods in parallel, but we wait for all of them to succeed.

Thanks for explanation.
We can't really assume they will finish fast then, as we still have QPS limits on the client-side.

The problem I see here, is that if a single sync can take that long, then one very large job (or N very large jobs where N=number of workers of the controller) can actually starve everything else for a really long time. That's far from perfect.

What we should do is to ensure that in a single sync we create no more than N pods (say 500 - I think 500 is used by ReplicaSet controller). And that gives you much better predictability (you can say that a single sync always has to finish within M seconds).
That has a drawback that it adds some latency if you indeed would be able to create e.g. 10k pods at once, but I think the impact is relatively small and you solve the starvation problem.

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: I agree that 1h is waay to much (and I don't really see the relevance with the link you pasted Aldo).

Also see - my comment above (it would help addressing this one too).

keps/sig-apps/2214-indexed-job/kep.yaml
```diff
@@ -20,7 +20,7 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.21"
+latest-milestone: "v1.22"
```
wojtek-t, line 23: Update stage above

wojtek-t (commented):
I'm mostly fine with is modulo the SLO-related comment I added.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: I'm fine with adding a limit to the number of created pods. Should we do it for deletions too (happens when users reduces parallelism or the job reached failure limits)?

Then for 500 pods, assuming 50 calls per second, we are talking about 10s for pods + 1s for Job update. Sounds reasonable? This will increase a bit with job tracking with finalizers, but we can discuss that in its beta graduation.

keps/sig-apps/2214-indexed-job/kep.yaml
alculquicondor: Will do in next update :)

keps/sig-apps/2214-indexed-job/README.md
wojtek-t: Yes - we should do for deletions too.

O(10s) SLO sounds waaay better than 1m or 1h.

> This will increase a bit with job tracking with finalizers, but we can discuss that in its beta graduation.

It adds one additional call per 500 calls, right? So I would say it's negligible.

alculquicondor: Right, big jobs are barely affected. And indexed jobs can actually skip some calls :)

In any case, I'll add SLOs for parallelism=1, 50 and 500

wojtek-t: > Right, big jobs are barely affected. And indexed jobs can actually skip some calls :)

Smaller jobs are not affected at all, because they will never not create a 500-item batch.
So only large jobs may be affected.

wojtek-t: > In any case, I'll add SLOs for parallelism=1, 50 and 500

I would really like to have one SLO (without any steps).

alculquicondor: > Smaller jobs are not affected at all

we still have to remove finalizers from the pods and job (but we are going off topic for the purpose of this PR).

> I would really like to have one SLO (without any steps).

will do, assuming 500 operations limit.

alculquicondor: updated

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -419,69 +419,76 @@ _This section must be completed when targeting alpha to a release._
...
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+
+  - per-day percentage of job_sync_total with label result=error <= 1%
+  - 99% percentile over day for job_sync_duration_seconds is <= 11s, assuming
```
wojtek-t, line 473: Let's not be that aggressive maybe. 15s?

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -419,69 +419,76 @@ _This section must be completed when targeting alpha to a release._
...
+    a client-side QPS limit of 50 calls per second.
```
wojtek-t, line 474: BTW - can you please also update the KEP to say that we need to divide those large sync into batches of 500 (or some other N if you have opinion on exact number).

wojtek-t (commented):
Just some smaller comments.

Also waiting for @soltysh to approve first from the SIG perspective before approving PRR.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: it's right below :)

alculquicondor (commented):
Updated SLO to 15s and squashed

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -419,69 +419,76 @@ _This section must be completed when targeting alpha to a release._
...
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+
+  - per-day percentage of job_sync_total with label result=error <= 1%
+  - 99% percentile over day for job_sync_duration_seconds is <= 15s, assuming
+    a client-side QPS limit of 50 calls per second.
+  
+  In a Job sync, the controller creates or deletes up to 500 Pods. The
+  controller processes the remaining operations in subsequent syncs, which
+  it schedules with no delay.
```
wojtek-t, line 478: I think we should move it to the design section, rather than hide it here.

wojtek-t (commented):
I'm generally fine with this modulo this minor comment - waiting for @soltysh approval first though.

keps/sig-apps/2214-indexed-job/README.md
alculquicondor: Moved to risks and mitigations

keps/sig-apps/2214-indexed-job/README.md
```diff
@@ -419,69 +419,74 @@ _This section must be completed when targeting alpha to a release._
...
+  If the new kube-controller-manager crashes, it's possible that an older
+  version of it would pick it up. In 1.21, when the IndexedJob feature is
+  disabled (default), the controller would not sync Indexed Jobs, that is: the
+  controller doesn't create or delete Pods and doesn't update Job status.
+  Running Pods are not affected.
 
 * **What specific metrics should inform a rollback?**
 
+  - job_sync_duration_seconds shows significantly more latency for label
+    mode=Indexed Jobs than mode=NonIndexed.
```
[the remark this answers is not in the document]
alculquicondor: I think mode is enough. @logicalhan is there any preference between shorter or more descriptive label names?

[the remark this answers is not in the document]
alculquicondor: ah, missed that. I homogenized to `mode` for now. We can always change it during code review if people feel strongly about the long name.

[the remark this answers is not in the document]
soltysh: My vote would be towards `completion_mode` since this is consistent with the API, but I won't block this KEP on it, it can be updated afterwards.

soltysh (approved):
/lgtm
/approve

chrisnegus:
Hi @alculquicondor  ! I'm @chrisnegus , a Docs Shadow for release 1.22.
This enhancement is marked as Needs Docs for the 1.22 release.

Please follow the steps detailed in the documentation to [open a PR against dev-1.22 branch](https://kubernetes.io/docs/contribute/new-content/new-features/#open-a-placeholder-pr) in the kubernetes/website repo. This PR can be just a placeholder at this time and must be created before *Fri July 9, 11:59 PM PDT*.

Also, take a look at [Documenting for a release](https://kubernetes.io/docs/contribute/new-content/new-features/#for-developers-or-other-sig-members) to familiarize yourself with the docs requirement for the release.

Thank you!

PR #2618, opened 2021-04-12
KEP-277: Use Pod for /ephemeralcontainers

verb:
This change was requested by SIG auth for compatibility with admission controllers.

### DIFF
--- a/keps/sig-node/277-ephemeral-containers/README.md
+++ b/keps/sig-node/277-ephemeral-containers/README.md
@@ -186,16 +186,22 @@ restrictions:
 
 ### Creating Ephemeral Containers
 
-1.  A client fetches the current list of Ephemeral Containers in a pod using
-    `GetEphemeralContainers` in the generated client. This returns a
-    `v1.EphemeralContainers` to which the client appends a new
-    `EphemeralContainer` and then calls `UpdateEphemeralContainers`.
-1.  The apiserver validates and performs the pod update, copying the new 
-    ephemeral container into `Pod.Spec.EphemeralContainers`.
+Ephemeral containers are described in the `EphemeralContainers` field of
+`Pod.Spec`. This must be updated using the `/ephemeralcontainers` subresource,
+similarly to updating `Pod.Status` via `/status`.
+
+The end-to-end process for creating an ephemeral container is:
+
+1.  Fetch a `Pod` object from the `/pods` resource.
+1.  Modify the object and write it back to the pod's `/ephemeralcontainers`
+    subresource, for example using `UpdateEphemeralContainers` in the generated
+    client. (Patching is also supported on `/ephemeralcontainers`.)
+1.  The apiserver validates the update.
     1.  Pod validation fails if container spec contains fields disallowed for
         Ephemeral Containers or the same name as a container in the spec or
         `EphemeralContainers`.
-    1.  API resource versioning resolves update races.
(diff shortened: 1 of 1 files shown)


keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -186,16 +186,22 @@ restrictions:
...
+`Pod.Spec`. This must be updated using the `/ephemeralcontainers` subresource,
+similarly to updating `Pod.Status` via `/status`.
+
+The end-to-end process for creating an ephemeral container is:
+
+1.  Fetch a `Pod` object from the `/pods` resource.
+1.  Modify the object and write it back to the pod's `/ephemeralcontainers`
+    subresource, for example using `UpdateEphemeralContainers` in the generated
+    client. (Patching is also supported on `/ephemeralcontainers`.)
+1.  The apiserver validates the update.
```
liggitt, line 199: insert a step before this one making it clear that the endpoint only honors data inside spec.ephemeralcontainers

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -629,10 +635,7 @@ via this subresource. `EphemeralContainerStatuses` is updated in the same manner
 as everything else in `Pod.Status` via `/status`.
 
 `Pod.Spec.EphemeralContainers` may be updated via `/ephemeralcontainers` as per
-normal (using PUT, PATCH, etc) except that existing Ephemeral Containers may not
-be modified or deleted. Deleting Ephemeral Containers is not supported in the
```
liggitt, line 633: was the restriction to modify/delete existing ephemeral containers lifted?

keps/sig-node/277-ephemeral-containers/README.md
verb: Added, thanks.

keps/sig-node/277-ephemeral-containers/README.md
verb: Deletion was proposed and accepted in #1690, but it hasn't been implemented yet. Happy to revisit.

#1690 doesn't explicitly take a stance on in-place modifications of the mutable `Container` fields, but I don't intend to support it. I've re-added the "may not be modified" section here.

verb:
FYI, implementation in kubernetes/kubernetes#101034. WIP because I intend to give it another pass tomorrow before sending for review.

keps/sig-node/277-ephemeral-containers/README.md
```diff
@@ -629,10 +637,8 @@ via this subresource. `EphemeralContainerStatuses` is updated in the same manner
...
-initial implementation to reduce complexity. It could be added in the future,
-but see *Killing Ephemeral Containers* below for additional constraints.
+normal (using PUT, PATCH, etc) except that existing Ephemeral Containers may
+not be modified.
```
dchen1107, line 641: Since we removed delete the existing ephemeral container above, are you referring that the policy is the deletion / removal of an ephemeral container is permitted?

dchen1107:
One small question above.

/lgtm
/approve

keps/sig-node/277-ephemeral-containers/README.md
verb: That's correct, we'll permit deleting ephemeral containers.

PR #2622, opened 2021-04-13
KEP 2438: Dual-stack apiserver

danwinship:
Make the apiserver service dual-stack in dual-stack clusters

### DIFF
--- a/keps/sig-network/2438-dual-stack-apiserver/README.md
+++ b/keps/sig-network/2438-dual-stack-apiserver/README.md
@@ -1,0 +1,596 @@
+# KEP-2438: Dual-Stack API Server
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1 - External Clients](#story-1---external-clients)
+    - [Story 2 - APIServer Tooling](#story-2---apiserver-tooling)
+    - [Story 3 - Wrong-Single-Stack Internal Clients](#story-3---wrong-single-stack-internal-clients)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [kube-apiserver Command-Line Arguments](#kube-apiserver-command-line-arguments)
+  - [Pod environment / client-go](#pod-environment--client-go)
+  - [Service and EndpointSlice Reconciling](#service-and-endpointslice-reconciling)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Handling of the &quot;<code>kubernetes</code>&quot; Service](#handling-of-the--service)
+    - [Handling of &quot;<code>kubernetes</code>&quot; Endpoints and EndpointSlices](#handling-of--endpoints-and-endpointslices)
+    - [Service vs EndpointSlice Sync Issues](#service-vs-endpointslice-sync-issues)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The initial implementation of [dual-stack networking] in Kubernetes
+left the apiserver as a single-stack Service. This moves it to
+dual-stack (in dual-stack clusters).
+
+[dual-stack networking]: ../563-dual-stack/README.md
+
+## Motivation
+
+### Goals
+
+- Ensure kube-apiserver can listen dual-stack (subject to appropriate
+  feature gates):
+
+    - Allow specifying dual-stack `--bind-address` on the command
+      line.
+
+    - Autodetect dual-stack bind addresses if possible when no
+      explicit `--bind-address` is specified.
+
+- Allow kube-apiserver to advertise dual-stack functionality to other
+  components (subject to appropriate feature gates):
+
+    - Allow specifying dual-stack `--advertise-address` on the command
+      line (and likewise, autodetecting dual-stack advertise addresses
+      from the bind address when `--advertise-address` is not
+      explicitly specified).
+
+    - Publish dual-stack `EndpointSlice`s for the "`kubernetes`"
+      Service when dual-stack apiserver IPs are available.
+
+    - When configured with both dual-stack bind/advertise addresses
+      and dual-stack service CIDRs, create the "`kubernetes`" service as
+      `ipFamilyPolicy: RequireDualStack`, and assign it the first IP
+      address in both the IPv4 CIDR range and the IPv6 CIDR range.
+
+- Update `rest.InClusterConfig()` to make use of dual-stack IPs, when
+  available.
+
+### Non-Goals
+
+- Changing the definition of `KUBERNETES_SERVICE_HOST`
+
+## Proposal
+
+### User Stories
+
+#### Story 1 - External Clients
+
+As a cluster administrator, I want my apiserver to be reachable by
+both IPv4-only and IPv6-only clients _outside the cluster_, without
+them needing to use a proxy or NAT mechanism.
+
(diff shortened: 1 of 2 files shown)


keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,521 @@
...
+related infrastructure (eg, TLS certificates, endpoints for external
+load balancers or DNS names, etc).
+
+#### Story 3 - Internal Clients (?)
+
+As a developer, I want to run a pod in a dual-stack cluster but only
+access the apiserver via the cluster's secondary IP family (eg, via
+IPv6 in an IPv4-primary cluster).
+
+(This has been brought up, but it is not clear that there is really a
```
aojea, line 118: is not this one https://github.com/kubernetes/enhancements/issues/563#issuecomment-814560270?

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,521 @@
...
+the first apiserver was upgraded/downgraded, and then not changing
+after that. In a more complicated upgrade/downgrade scenario, there
+might be some flapping; eg, the first master upgrades and changes the
+Service to dual-stack, but then the second master reboots before
+upgrading, causing the old apiserver to restart and revert the Service
+definition back to its single-stack form, before then being upgraded
+and changing it back to dual-stack again.
+
+However, at all times, the Service should fully usable; it should
+never claim to be dual-stack unless there actually are dual-stack
```
aojea, line 258: a dual-stack service will always be fully usable if the primary ip family matches the single-stack family, right?

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,521 @@
...
+advertised IPv6 endpoints that didn't work).
+
+Failing that, the release notes would have to document that in the
+event of a downgrade it might be necessary to manually clean the
+EndpointSlices up.
+
+[kubernetes #101070]: https://github.com/kubernetes/kubernetes/issues/101070
+
+### Version Skew Strategy
+
```
aojea, line 308: I think that this sections refer to run several apiservers in parallel with different versions, that you touch in the upgrade/downgrade section

keps/sig-network/2438-dual-stack-apiserver/README.md
danwinship: I'm not talking about generic promises we make about dual-stack Services, I'm just talking about the specific temporary states of the apiserver service during this particular upgrade.

And actually, there's a case I wasn't thinking about: because the Service is only changed at apiserver startup, but the Endpoints get resynced while running, you can end up in a case where the Service is dual-stack, but there is no EndpointSlice for the secondary IP family. In that case, trying to connect to the Service's secondary IP would fail with "connection refused". Although, I guess as long as clients behave sanely, they'd cope with that anyway; this would only be a problem if some client tried to connect _only_ to the secondary-family ClusterIP for some reason, rather than trying to connect _only_ to the primary-family ClusterIP, or trying both in sequence/parallel.

keps/sig-network/2438-dual-stack-apiserver/README.md
danwinship: No, the template asks "how will the component handle version skew with other components?" As you said, self-version-skew is covered in the upgrade/downgrade section.

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,575 @@
...
+IP is unreachable from that pod.
+
+There was [discussion in an earlier KEP] about whether it is actually
+safe to add new `KUBERNETES_*` environment variables, given that we
+don't restrict users from using variables with such names for their
+own purposes...)
+
+It's possible that there is no solution that is nice enough that it's
+actually worth trying to make it work for the admittedly-edge-case-y
+scenario of wrong-single-stack pods...
```
aojea, line 199: this ... also, there is a workardound that is using `kubernetes.default` as `KUBERNETES_HOST`

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,575 @@
...
+should proceed smoothly. People who enable the feature gate manually
+during the alpha period may encounter problems with
+Service/EndpointSlice flapping during upgrades, and may need to do
+some manual cleanup on downgrade.
+
+The new code should also behave like the "half-new dual-stack" case
+when an apiserver is configured to be single-stack but it sees that
+other apiservers are advertising dual-stack. This will prevent
+flapping during a roll-out of a single-stack to dual-stack
+configuration change.
```
aojea, line 278: we had some fun with the opposite , dual -> single, do we want to support it?

aojea: scratch it , you cover it below

keps/sig-network/2438-dual-stack-apiserver/README.md
aojea: that you already cover below too :) ... useless comment

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,575 @@
...
+times about what endpoints are available _in the cluster's primary IP
+family_, and thus will always agree on the contents of the
+single-stack `Endpoints` resource.
+
+For `EndpointSlice`, current versions of kube-apiserver create a
+single `EndpointSlice` object named "`kubernetes`", and assume that
+other apiservers will do the same; they don't look at any other
+`EndpointSlice` objects when reconciling. To create dual-stack
+EndpointSlices, a new apiserver would need to create two objects with
+different names. Even if new apiservers use the same name as the old
```
aojea, line 387: should we provide more details on the new names for the secondary endpoint slice?
right now, there is only one, for compatibility I think it has to be the primary ip family, but how will be named the secondary?
```
kubectl get endpointslices
NAME         ADDRESSTYPE   PORTS   ENDPOINTS    AGE
kubernetes   IPv4          6443    172.18.0.2   21h
```

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,593 @@
...
+left the apiserver as a single-stack Service. This moves it to
+dual-stack (in dual-stack clusters).
+
+[dual-stack networking]: ../563-dual-stack/README.md
+
+## Motivation
+
+### Goals
+
+- Ensure kube-apiserver can listen dual-stack (subject to appropriate
```
khenidak, line 68: side comment. As we are moving `IPv6DualStack` feature to stable, you will probably need to define a new feature.

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,593 @@
...
+  - If `--bind-address` or `--advertise-address` is specified as a
+    single IP, then the apiserver is single-stack, even if dual
+    `--service-cluster-ip-range`s are specified.
+
+  - If `--bind-address` or `--advertise-address` is specified as a
+    pair of IPs, and `--service-cluster-ip-range` is specified as a
+    pair of CIDRs, then the apiserver is dual-stack, and should fail
+    if it cannot bind sockets for both address families.
+
+  - (Incompatible values for `--bind-address`, `--advertise-address`,
```
khenidak, line 177: I think when `--advertise-address` is dual-stack but `--bind-address` isn't then we **should** fail. Because simply you can not advertise address you are not bound to its family. But the other way around we *should not* fail. 

For example: A cluster may have monitoring tools running on control plane nodes that are running with alternative IP family. IOW I may want to bind to both family but i don't want to advertise both families.

khenidak:
I think this is implementable. The exact details of some of the things that need to change may have not been completely identified but we know the scope and the sharp edges of the changes.

@thockin  @danwinship

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,593 @@
...
+all non-early-adopters get smooth upgrades and downgrades.
+
+For early adopters, or if we want to only have the feature be alpha
+for a single release, and not have completely smooth behavior for
+users with N-2 version skew, there are a few potential problems:
+
+#### Handling of the "`kubernetes`" Service
+
+kube-apiserver attempts to force the "`kubernetes`" Service definition
+to match its expectations on startup, but does not attempt to
```
thockin, line 365: I think this "does not reconcile" is a bug and when we eventually fix that, it will flap.  I was thinking about alternatives...

EPSlice already says that a given IP *might* be in more than one slice (e.g. while rebalancing).  What if EACH apiserver wrote their OWN epslices.  Single-stack servers would write 1, dual-stack would write 2.   Cleanup on downgrade is tricky - which slices are "mine" ?  Maybe we need some other way to ident apiserver replicas so they can operate independently in this manner?

thockin (commented):
The only thing I don't get is the anti-flapping when apiservers are split-opinion.  Is it just "it will sync up eventually" ?

I'm OK to merge this and iterate.

thockin:
Let's follow up on the topics of in-cluster and split-opinion.

/lgtm
/approve

thockin:
Added this to the 1.23 sheet

keps/sig-network/2438-dual-stack-apiserver/README.md
aojea: ... unless you bind to 0.0.0.0 or :: , when you bind to all addresses and families, but you only one to advertise only specific ones

keps/sig-network/2438-dual-stack-apiserver/README.md
aojea: should we create a name convention for the slices name? 
that will simplify the problem `kubernetes` and `kubernetes-ipv6` ?

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,593 @@
...
+the case of a dual-stack IPv6-primary cluster that contains some nodes
+that are on an IPv4-only network, which can therefore only host
+IPv4-only pods.
+
+[it has been seen in the wild]: https://github.com/kubernetes/enhancements/issues/563#issuecomment-814560270
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
```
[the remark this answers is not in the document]
danwinship: more "not thought about"

keps/sig-network/2438-dual-stack-apiserver/README.md
```diff
@@ -0,0 +1,593 @@
...
+        apiserver.
+
+  - API server knows about `DualStackAPIServer` but either the feature
+    is not enabled on this apiserver, or else this apiserver has a
+    single-stack configuration:
+
+      - Advertises its own *single-stack* endpoint IP to etcd
+
+      - Forces the `kubernetes` Service to single-stack if there are
+        no dual-stack apiservers in etcd. Leaves the Service unchanged
```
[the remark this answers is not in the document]
danwinship: Based on what's in etcd. If you're using [`--endpoint-reconciler-type=master-count`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controlplane/reconcilers/instancecount.go), it would see that some other apiserver had created dual-stack EndpointSlices and react to that. (Although this would be flaky... maybe we want to disallow `--endpoint-reconciler-type=master-count` with dual-stack apiservers.) If you're using [`--endpoint-reconciler-type=lease`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controlplane/reconcilers/lease.go) then the endpoint IPs are stored in etcd separately from the Endpoints/EndpointSlice objects... although we will need to figure out how to change that code in a way that doesn't break existing single-stack apiservers...

keps/sig-network/2438-dual-stack-apiserver/README.md
danwinship: > Maybe we need some other way to ident apiserver replicas so they can operate independently in this manner?

yeah, we already have this; the "lease"-based endpoint reconciling. Each apiserver writes its IPs to a special object in etcd, and then each apiserver can regenerate the Endpoints/EndpointSlices based on that

danwinship:
(needs a re-lgtm because I had to regen the table of contents)

PR #2623, opened 2021-04-13
added scalability concerns to topology manager KEP

SergeyKanzhelev:
Raised in https://github.com/kubernetes/kubernetes/issues/100906. Adding a TBD record to make sure it is not lost on GA graduation


klueska:
@SergeyKanzhelev this is related to (but mostly independent of) the TopologyManager.

If any KEP should be updated, it is the CPUManager KEP, not the TopologyManager one.

That said, it is a simple, non-invasive optimization, that doesn’t affect the behavior of the system at all (other than to avoid extra update calls through the CRI), so I’m not convinced any KEP needs to be updated at all.

PR #2638, opened 2021-04-19
Update probe grace period KEP to beta

ehashman:
/sig node

Issue #2238

### DIFF
--- a/keps/prod-readiness/sig-node/2238.yaml
+++ b/keps/prod-readiness/sig-node/2238.yaml
@@ -1,3 +1,5 @@
 kep-number: 2238
 alpha:
   approver: "@johnbelamaric"
+beta:
+  approver: "@johnbelamaric"
--- a/keps/sig-node/2238-liveness-probe-grace-period/README.md
+++ b/keps/sig-node/2238-liveness-probe-grace-period/README.md
@@ -8,8 +8,6 @@
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
   - [Configuration example](#configuration-example)
-  - [User Stories (Optional)](#user-stories-optional)
-  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [Test Plan](#test-plan)
@@ -29,7 +27,6 @@
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
-- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
 <!-- /toc -->
 
 ## Release Signoff Checklist
@@ -325,16 +322,33 @@ _This section must be completed when targeting beta graduation to a release._
   Try to be as paranoid as possible - e.g., what if some components will restart
    mid-rollout?
 
+  This change is purely additive. Already-running workloads on a node will not
+  have this field set.
+
 * **What specific metrics should inform a rollback?**
+
+  Increases in error rates on shutdowns initiated by probes; unexpected
+  shutdown times for pods with containers that have probes configured.
 
 * **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
   Describe manual testing that was done and the outcomes.
   Longer term, we may want to require automated upgrade/rollback tests, but we
   are missing a bunch of machinery and tooling and can't do that now.
 
+  For the API server, rollback can be tested by:
+
+  - Turning on the feature flag
+  - Creating a workload with the field
+  - Disabling the feature flag
+
+  We will remove the feature flag from the kubelet for beta to avoid worrying
+  about the n-2 skew case.
+
 * **Is the rollout accompanied by any deprecations and/or removals of features, APIs,
 fields of API types, flags, etc.?**
   Even if applying deprecation policies, they may still surprise some users.
+
+  No.
 
 ### Monitoring Requirements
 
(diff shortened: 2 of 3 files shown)


keps/prod-readiness/sig-node/2238.yaml
```diff
@@ -1,3 +1,5 @@
 kep-number: 2238
 alpha:
   approver: "@johnbelamaric"
+beta:
+  approver: "@johnbelamaric"
```
johnbelamaric, line 5: Looks like Elana approved this, so change beta approver to her. Thanks!

ehashman: @johnbelamaric it's my KEP, I can't review my own KEP :joy:

johnbelamaric: Doh! I should have known better, I remember now from last cycle this one is yours :)

keps/sig-node/2238-liveness-probe-grace-period/README.md
```diff
@@ -329,17 +322,34 @@ _This section must be completed when targeting beta graduation to a release._
...
   Longer term, we may want to require automated upgrade/rollback tests, but we
   are missing a bunch of machinery and tooling and can't do that now.
 
+  For the API server, rollback can be tested by:
+
+  - Turning on the feature flag
+  - Creating a workload with the field
+  - Disabling the feature flag
+
+  We will remove the feature flag from the kubelet for beta to avoid worrying
```
johnbelamaric, line 344: If the field is set, and then the flag disabled in apiserver, I assume the field will not be present when the kubelet reads the spec?

ehashman: Right now, it's only unset when the feature flag is disabled *and* the field is not present, looks like: https://github.com/kubernetes/kubernetes/blob/6296c6e4d336f9d3e4cc80e0af5d4bc74f330d0d/pkg/api/pod/util.go#L553-L565

Would I need to make this change?

johnbelamaric: Well, if we don't, then wouldn't we still need the feature flag in kubelet? Turning off the feature at the api server alone won't disable it if the kubelet still gets the value. What's the standard practice for this in the API?

ehashman: I'm not sure. @derekwaynecarr wdyt?

keps/sig-node/2238-liveness-probe-grace-period/README.md
```diff
@@ -198,11 +189,13 @@ quickly.
   available behind a feature flag.
 - Appropriate tests are written.
 
-_Below graduation criteria are tentative._
-
 #### Beta
 
 - Feature flag is defaulted on.
+- Remove feature gate from [kubelet](https://github.com/kubernetes/kubernetes/pull/99375#issuecomment-794680869).
```
derekwaynecarr, line 195: To support n-2 skew, I think we need do the following:

- the feature gate is not present on the kubelet
- the feature gate is not defaulted to on until kube-apiserver is +2 versions of the kubelet having the support.

keps/sig-node/2238-liveness-probe-grace-period/README.md
derekwaynecarr: I think it is safe to remove the feature gate from the kubelet under beta criteria.

This starts the clock for all nodes to understand this new field.
In two releases, we default the kube-apiserver to enable the feature gate so all environments will have kubelets connected that understand the feature natively.

For users that want to exercise the new field earlier, they need to enable the feature gate on the kube-apiserver and ensure all connected kubelets are at equal level N.

derekwaynecarr (changes requested):
as written, i think the feature goes to beta, but disabled by default on kube-apiserver until 2+ releases when all known kubelets will support the new field.  for users that want to consume the feature earlier in kube-apiserver, they must enable the field and ensure all connected clients are at level n.  basically, we cant default the feature on yet, but we can still declare a beta milestone.

keps/sig-node/2238-liveness-probe-grace-period/README.md
johnbelamaric: My concern is about the scenario where:
 - feature gate is not present in kubelet
 - user turns on feature in apiserver, sets the new field
 - user turns off feature in apiserver

As written, in this scenario, the feature would still be active for the workloads with the value set, because the kubelet cannot turn off that feature, and it still sees the new field.

johnbelamaric: I still don't see this scenario handled properly if you remove the feature gate from kubelet early.

keps/sig-node/2238-liveness-probe-grace-period/README.md
johnbelamaric: See my question below wrt enable/disable scenario.

liggitt: > Turning off the feature at the api server alone won't disable it if the kubelet still gets the value. What's the standard practice for this in the API?

Turning off a feature flag does not clear previously persisted data from the API field, it prevents data in the field in new objects (or in updates of existing objects that don't already use the field).

johnbelamaric: Right, so my point is that we can't remove the kubelet feature flag, because if we do, then the user has no control over those old objects still containing the value, and no way to then back out the feature on those objects already with the value (in the enable then disable scenario). Except to delete it entirely and recreated it. 

So, the question is: if we want to remove the kubelet feature flag early, would it be OK to blank the field on output from the api server when the feature flag is disabled? That would prevent the kubelet from seeing the new value and effectively disable the functionality without having to keep the kubelet feature flag.

derekwaynecarr: @johnbelamaric that makes sense.  the other option is if we could say the feature flag defaulted to on for kubelet, but off for kube-apiserver, but I am not aware of when we have done that in the past (or if we did, my memory escapes me).

ehashman: I think the easiest way forward is probably to blank out the field when the feature flag is off. I will update.

ehashman:
@johnbelamaric made some small updates to blank out value from API server when feature flag is disabled, PTAL.

keps/sig-node/2238-liveness-probe-grace-period/README.md
johnbelamaric: Ok, let's move forward with this as plan-of-record, if Jordan or other API reviewers hate it we can revisit...

johnbelamaric:
This is approved for PRR. Derek, if you're good with it go ahead an approve.

keps/sig-node/2238-liveness-probe-grace-period/README.md
derekwaynecarr: this may be possible, i just cant remember what i did on this and if it handled the write side but not the read side.

PR #2643, opened 2021-04-21
Tweak PRR guidance to consider HA

johnbelamaric:
Update the "how can a rollout fail" to include rollbacks. Add text to poke people to think about HA and large clusters.

/assign @deads2k @ehashman @wojtek-t


wojtek-t:
/lgtm

Thanks!

PR #2660, opened 2021-04-29
Mark KEP 1972 as withdrawn

wangyysde:
KEP 1972 is stable since v1.20. Should we mark it as withdrawn  in v1.22?


keps/sig-node/1972-kubelet-exec-probe-timeouts/kep.yaml
```diff
@@ -15,16 +15,17 @@ approvers:
   - "@derekwaynecarr"
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: stable
+stage: implemented
```
pacoxu, line 19: line 8
> status: implemented
status is already implemented here

wangyysde: Thanks! I have modified the file again.

PR #2664, opened 2021-04-29
CSIServiceAccountToken going GA

zshihang:
/cc @msau42

### DIFF
--- a/keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
+++ b/keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
@@ -15,13 +15,14 @@ approvers:
   - "@msau42"
   - "@mikedanese"
 creation-date: 2020-06-09
-last-updated: 2021-02-04
+last-updated: 2021-04-29
 status: implementable
-stage: beta
-latest-milestone: "v1.21"
+stage: stable
+latest-milestone: "v1.22"
 milestone:
   alpha: "v1.20"
   beta: "v1.21"
+  stable: "v1.22"
 feature-gates:
   - name: CSIServiceAccountToken
     components:
--- a/keps/prod-readiness/sig-storage/1855.yaml
+++ b/keps/prod-readiness/sig-storage/1855.yaml
@@ -1,3 +1,5 @@
 kep-number: 1855
 beta:
   approver: "@deads2k"
+stable:
+  approver: "@deads2k"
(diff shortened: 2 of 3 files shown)


keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
```diff
@@ -15,13 +15,14 @@ approvers:
   - "@msau42"
   - "@mikedanese"
 creation-date: 2020-06-09
-last-updated: 2021-02-04
+last-updated: 2021-04-29
 status: implementable
-stage: beta
-latest-milestone: "v1.21"
+stage: stable
```
msau42, line 20: In the README.md, can you also update the Implementation history?

For scalability, have we tested in a large cluster if the 1 token request per second is ok?

deads2k:
PRR still looks good (https://github.com/kubernetes/enhancements/blob/master/keps/sig-storage/1855-csi-driver-service-account-token/README.md#scalability)

/approve

keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
zshihang: @liggitt do we have load test for bound tokens?

for this feature, the load is # of pods * # of csi volumes. since we have 110 pods limit and the cache layer in kubelet, that number would not be too large. i think it is nice to have one but not sure it is a blocker.

liggitt: I think large numbers of pods with unique tokens is covered by existing scale tests, but @wojtek-t @mm4tt can confirm

JamesLaverack:
xref https://github.com/kubernetes/enhancements/issues/2047

keps/sig-storage/1855-csi-driver-service-account-token/kep.yaml
mm4tt: @marseel

zshihang: done for implementation history.

re scalability, current scale test for access tokens is using secret-based tokens, which is different from the bound tokens here. however, since we enable the BoundServiceAccountTokenVolume by default since 1.21, every pod will have one projected token mounted. then the scale test for pods should cover this.

msau42: Thanks for confirming!

PR #2667, opened 2021-04-30
Updating KEP 536 to note that it has been replaced by KEP 2433

robscott:
This also commits us to removing the implementation in the 1.22 cycle.

/sig network
/assign @thockin


thockin (commented):
Thanks!

/lgtm
/approve

PR #2670, opened 2021-05-03
Update CSI Windows KEP for GA

jingxu97:
update CSI windows kep

### DIFF
--- a/keps/prod-readiness/sig-windows/1122.yaml
+++ b/keps/prod-readiness/sig-windows/1122.yaml
@@ -1,0 +1,3 @@
+kep-number: 1122
+stable:
+  approver: "@deads2k"
--- a/keps/sig-windows/1122-windows-csi-support/README.md
+++ b/keps/sig-windows/1122-windows-csi-support/README.md
@@ -33,6 +33,13 @@
     - [Beta -&gt; GA Graduation](#beta---ga-graduation)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
@@ -51,6 +58,8 @@
 - [ ] Design details are appropriately documented
 - [ ] Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
 - [ ] Graduation criteria is in place
+- [ ] Production readiness review completed
+- [ ] Production readiness review approved
 - [ ] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
@@ -293,7 +302,7 @@ There will be two command line parameters that may be passed to csi-proxy.exe:
 2. kubelet-pod-path: String parameter pointing to the path used by the kubelet to store pod specific information. This should map to the value returned by [getPodsDir](https://github.com/kubernetes/kubernetes/blob/e476a60ccbe25581f5a6a9401081dcee311a066e/pkg/kubelet/kubelet_getters.go#L48). By default it will be set to: `C:\var\lib\kubelet\pods` Parameters to `LinkPath` (detailed below) will need to be under this path.
 
 ##### CSI Proxy GRPC API
-The following are the main RPC calls that will comprise a v1alpha1 version of the CSIProxyService API. A preliminary structure for Request and Response associated with each RPC is described below. Note that the specific structures as well as restrictions on them are expected to evolve during Alpha phase and are expected to be in a final form at the end of Beta. As the API evolves, the section below will be kept up-to-date
+The following are the main RPC calls that will comprise a v1alpha1 version of the CSI Proxy API Group. A preliminary structure for Request and Response associated with each RPC is described below. Note that the specific structures as well as restrictions on them are expected to evolve during Alpha phase and are expected to be in a final form at the end of Beta. As the API evolves, the section below will be kept up-to-date
 
 ```
 service CSIProxyService {
@@ -694,7 +703,7 @@ message DismountVolumeResponse {
 
 In accordance with standard Kubernetes conventions, the above API will be introduced as v1alpha1 and graduate to v1beta1 and v1 as the feature graduates. Beyond a vN release in the future, new RPCs and enhancements to parameters will be introduced through vN+1alpha1 and graduate to vN+1beta1 and vN+1 stable versions as the new APIs mature.
 
-Members of CSIProxyService API may be deprecated and then removed from csi-proxy.exe in a manner similar to Kubernetes deprecation (policy)[https://kubernetes.io/docs/reference/using-api/deprecation-policy/] although maintainers will make an effort to ensure such deprecation is as rare as possible. After their announced deprecation, a member of CSIProxyService API must be supported:
+Members of CSI Proxy API Group may be deprecated and then removed from csi-proxy.exe in a manner similar to Kubernetes deprecation (policy)[https://kubernetes.io/docs/reference/using-api/deprecation-policy/] although maintainers will make an effort to ensure such deprecation is as rare as possible. After their announced deprecation, a member of CSI Proxy API Group must be supported:
 1. 12 months or 3 releases (whichever is longer) if the API member is part of a Stable/vN version.
 2. 9 months or 3 releases (whichever is longer) if the API member is part of a Beta/vNbeta1 version.
 3. 0 releases if the API member is part of an Alpha/vNalpha1 version.
@@ -709,7 +718,7 @@ For all RPC invocations, csi-proxy.exe will log events to the Windows applicatio
 
 #### Enhancements in Kubernetes/Utils/mounter
 
-Once the [PR](https://github.com/kubernetes/utils/pull/100/files) lands, a mounter/mount_windows_using_csi_proxy.go in Kubernetes/Utils/mounter package can be introduced. It will implement the mounter and hostutil interfaces against the CSIProxyService API.
+Once the [PR](https://github.com/kubernetes/utils/pull/100/files) lands, a mounter/mount_windows_using_csi_proxy.go in Kubernetes/Utils/mounter package can be introduced. It will implement the mounter and hostutil interfaces against the CSI Proxy API Group.
 
 #### Enhancements in CSI Node Plugins
 
@@ -747,20 +756,26 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
 
 #### Alpha -> Beta Graduation
 
-- csi-proxy.exe supports v1beta1 version of the CSIProxyService API.
+- csi-proxy.exe supports v1beta1 version of the CSI Proxy API Group.
 - end-2-end tests in place with a CSI plugin that can support Windows containers and pass all existing CSI plugin test scenarios.
 
 #### Beta -> GA Graduation
 
 - In-tree storage plugins that implements support for Windows (AWS EBS, GCE PD, Azure File and Azure Disk as of today) can use csi-proxy.exe along with other enhancements listed above to successfully deploy CSI plugins on Windows nodes.
-- csi-proxy.exe supports v1 stable version of the CSIProxyService API.
-- Successful usage of csi-proxy.exe with support for v1 version of CSIProxyService API in Windows nodes by at least two storage vendors.
+- csi-proxy.exe supports v1 stable version of the CSI Proxy API Group.
+- Successful usage of csi-proxy.exe with support for v1 version of CSI Proxy API Group in Windows nodes by at least two storage vendors.
 
 ### Upgrade / Downgrade Strategy
 
-In order to install a CSI Node Plugin or upgrade to a version of a CSI Node Plugin that uses an updated version of the CSIProxyService API not supported by the currently deployed version of csi-proxy.exe in the cluster, csi-proxy.exe will need to be upgraded first on all nodes of the cluster before deploying or upgrading the CSI Node Plugin. In case there is a very old CSI Node Plugin in the cluster that relies on a version of CSIProxyService API that is no longer supported by the new version of csi-proxy.exe, the previously installed version of csi-proxy.exe should not be uninstalled from the nodes. Such scenarios are expected to be an exception.
+As explained in (#csi-proxy-grpc-api-graduation-and-deprecation-policy), CSI proxy will expose multiple API versions according to the API versioning policy, to facilitate drivers upgrading to a newer API version and multiple drivers using different API versions.
+
+In order to install a CSI Node Plugin or upgrade to a version of a CSI Node Plugin that uses an updated version of the CSI Proxy API Group not supported by the currently deployed version of csi-proxy.exe in the cluster, csi-proxy.exe will need to be upgraded first on all nodes of the cluster before deploying or upgrading the CSI Node Plugin.
+
+In case there is a very old CSI Node Plugin in the cluster that relies on a version of CSI Proxy API Group that is no longer supported by the new version of csi-proxy.exe, the previously installed version of csi-proxy.exe should not be uninstalled from the nodes. Such scenarios are expected to be an exception.
 
 Different nodes in the cluster may be configured with different versions of csi-proxy.exe as part of a rolling upgrade of csi-proxy.exe. In such a scenario, it is recommended that csi-proxy.exe upgrade is completed first across all nodes. Once that is complete, the CSI Node Plugins that can take advantage of the new version of csi-proxy.exe may be deployed.
+
+However, in case there is unavoidable case that CSI Node Plugin might get upgraded first before csi-proxy is upgraded, the driver must fall back to older versions if a newer API version of CSI proxy is not available. An example is [PD CSI driver implementation](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/pull/738)
 
 Downgrading the version of csi-proxy.exe to one that is not supported by all installed versions the CSI Node Plugins in the cluster will lead to loss of access to data. Further, if a cluster is downgraded from a version of Kubernetes where the plugin watcher supports Windows nodes to one that does not, existing Windows workloads that were using CSI plugins to access storage will no longer have access to the data. This loss of functionality cannot be handled in an elegant fashion.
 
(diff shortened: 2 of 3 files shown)


keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+-->
+It is up to the cloud provider to tie CSI-proxy version to cluster version. When clusters rollout and rollback, the CSI-proxy should be udpated accordingly.
+
+###### What specific metrics should inform a rollback?
+
+<!--
+What signals should users be paying attention to when the feature is young
+that might indicate a serious problem?
+-->
+N/A
```
msau42, line 889: Can we use CSI proxy binary resource usage as a signal?

Or if CSI driver error metrics?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+<!--
+Describe manual testing that was done and the outcomes.
+Longer term, we may want to require automated upgrade/rollback tests, but we
+are missing a bunch of machinery and tooling and can't do that now.
+-->
+N/A
```
msau42, line 898: What about csi-proxy binary version upgrades/downgrades?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+<!--
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+Windows CSI Node plugin needs to use CSI-proxy to perform volume/disk/file operations. By mornitoring the CSI node plugin, CSI-proxy is monitoed.
```
msau42, line 911: typo: monitored

So if CSI driver returns an error, it's either a csi driver problem or a csi-proxy problem, and someone would need to look at logs to debug further.

Would it be beneficial to have separate metrics so that it's easier to distinguish where the problem is coming from? It could be helpful for debugging performance issues as well.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+
+###### How can an operator determine if the feature is in use by workloads?
+
+<!--
+Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+checking if there are objects with field X set) may be a last resort. Avoid
+logs or events for this purpose.
+-->
+This can be determined by monitoring the usage of CSI node plugin.
```
msau42, line 920: It would have to be a combination of two metrics: the csi metric and whether or not the node is windows.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+Volume/file operation latency might be different from different storage vendors. For example, it takes a few minutes to format a disk for Windows
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+<!--
+Pick one more of these and delete the rest.
+-->
+
+- [x] Metrics
+  - Metric name: Volume/file operation latency for CSI
```
msau42, line 966: Can you specify the exact metric name?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+implementation difficulties, etc.).
+-->
+No
+
+### Dependencies
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+No
```
msau42, line 985: Does CSI proxy depend on powershell, windows os versions? Any packages or libraries that need to be installed on the host?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+<!--
+Things to keep in mind include: additional in-memory state, additional
+non-trivial computations, excessive access to disks (including increased log
+volume), significant amount of data sent and/or received over network, etc.
+This through this both in small and large cases, again with respect to the
+[supported limits].
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+-->
+No
```
msau42, line 1086: CSI proxy will be making OS system calls. Could any of those system calls be expensive?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+<!--
+For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them.
+
+For beta, this section is required: reviewers must answer these questions.
+
+For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field.
+-->
+Stress tests can performed through storage test cases. Different cloud provider should have their own scalability tests.
```
msau42, line 1016: Can we reuse any existing kubernetes storage tests so that each cloud provider doesn't have to write custom tests?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+### Troubleshooting
+
+<!--
+This section must be completed when targeting beta to a release.
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+-->
+1. If CSI-proxy is not install
```
msau42, line 1097: These two can go under the "other failure modes" section

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,6 +777,348 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+-->
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+
 ## Implementation History
 
 07/16/2019: Initial KEP drafted
```
msau42, line 1124: Also add GA update entry here

msau42 (commented):
@ddebroy do any of the existing design sections need to be updated? For example, upgrade/downgrade strategy

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: resource usage seems not a strong signal. If CSI proxy is not installed, or is not running or install the wrong version, when CSI driver makes API call, it will have error messages.

We also have csi-proxy logs which can be used indicate whether it is running correctly or not.

jingxu97: Or pdcsi_operation_error_rate for Windows sort of metric can indicate there might be some issue with csi-proxy

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: Different provider might use different way to upgrade/downgrade. For GKE, this is be controlled by setting environment variable to control csi-proxy version which can tied to GKE cluster version.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: csi-proxy has its own log too. User can check csi-proxy to detect error message.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: updated

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: I put the notes in wrong place here

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: should be csi_operations_seconds_bucket
https://github.com/kubernetes/kubernetes/pull/98979/files

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: All the windows versions used in kubernetes has all the necessary packages or libraries, so no need to install anything here.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: The existing OSS stress tests can be used https://github.com/kubernetes/kubernetes/blob/v1.21.0/test/e2e/storage/testsuites/volume_stress.go#L106

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: The one takes longer latency is format disk, but I don't see it consumes much cpu or memory. All others seems light

keps/sig-windows/1122-windows-csi-support/README.md
msau42: I think we want some sort of metric to easily detect and alert. Logs are usually only looked at once a problem is identified. It seems like we can use the volume_operation metric to cover this.

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Yes, but should we at least do some basic upgrade/downgrade testing (could be manual) so that we know the basic mechanics will work and can provide guidance to distros on the proper/supported ordering?

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Or the new csi_operations one

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: we could provide manual test guidance using pd csi driver. Will it work?

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Are Windows versions explicitly specified in Kubernetes or can you install Kubernetes on any OS version? For example in Linux, we don't really have specific version requirements.

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Could rescan, or list disks be expensive? Have we scale tested a lot of volumes on a single node?

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: Yes, windows has version support
https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#windows-os-version-support

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: we can try to run some manual tests to verify. 

All persistent volume related metrics are TBD here
https://github.com/kubernetes/community/blob/master/sig-scalability/configs-and-limits/thresholds.md

keps/sig-windows/1122-windows-csi-support/README.md
ddebroy: Windows event logs (which monitor the health of all Windows Services [that csi-proxy supports and can be configured to run as]) can also provide errors and warnings around any major issues with csi-proxy. Windows event logs will be integrated into Kubernetes tooling as part of https://github.com/kubernetes/enhancements/pull/2271

keps/sig-windows/1122-windows-csi-support/README.md
ddebroy: In the context of CSI proxy (running as a Process or Windows Service), should this section focus on:
[a] specific latencies and extra resources consumed as a result of driving operations through [1] CSI plugins using CSI proxy vs [2] in-tree plugins with support for Windows?
OR
[b] the overall resources associated with running operations (such as list disk/format volumes/etc) associated with CSI plugins vs not running CSI plugins on Windows?

If [a], I think Jing already did this analysis earlier and maybe we can post it here?

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Events and logs are good for troubleshooting specific issues, but not for detecting fleet-wide trends. I think this is where metrics are valuable. We already have volume_operation and csi_operation metrics that will cover this path.

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Yes some development documentation on how you could test upgrades/downgrades would be helpful. We can also investigate if the existing kubernetes statefulset upgrade test can also be leveraged for Windows.

keps/sig-windows/1122-windows-csi-support/README.md
msau42: I think a) is the more important measurement, especially considering csi migration.

ddebroy (commented):
I guess we should also consider updating the API section at https://github.com/kubernetes/enhancements/pull/2670/files#diff-a1f0449cd50e68225198d063fd235586c87379109ee3d1c1ed6fe3663d33c816L299 in a separate PR to align with the latest feedback around v1.

msau42:
/assign @ehashman 
for prr review

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: added. Thanks!

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+- The rest of volume operations in the range of seconds.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+<!--
+Pick one more of these and delete the rest.
+-->
+
+- [x] Metrics
+  - Metric name: Volume/file operation latency for CSI (csi_operations_seconds_bucket See details https://github.com/kubernetes/kubernetes/pull/98979/files)
```
msau42, line 968: remove "bucket" from the metric name

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+This through this both in small and large cases, again with respect to the
+[supported limits].
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+-->
+There are two aspects of resource usage in the context of CSI proxy
+
+[a] specific latencies and extra resources consumed as a result of driving operations through CSI plugin using CSI proxy vs. In-tree plugin with Support for Windows:
+
+The prelimiary investigation on this is available. There are relatively small extra delay caused by CSI proxy for certain volume operations. There will be future performance improvement work on this area.
```
msau42, line 1092: link?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+Describe the consequences on existing workloads (e.g., if this is a runtime
+feature, can it break the existing applications?).
+
+NOTE: Also set `disable-supported` to `true` or `false` in `kep.yaml`.
+-->
+It depends on cloud provider implementation. For example in GCE, environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+Nothing
```
msau42, line 851: Windows CSI volume operations will work again.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+<!--
+Pick one of these and delete the rest.
+-->
+
+- [ ] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name:
+  - Components depending on the feature gate:
+- [x] Other
+  - Describe the mechanism:
+  CSI-proxy is a binary that can be deployed (or running as a service) directly on each Windows node during node start up time. For example, in gce, CSI-proxy
```
msau42, line 821: Note here that the deployment depends on the cluster provider.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+  binary is downloaded and running as Windows service in node [startup script](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/k8s-node-setup.psm1#L424)
+  An environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
+  - Will enabling / disabling the feature require downtime of the control
+    plane?
+  No
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+  If csi-proxy binary is deployed as a binary or windows service during node start up time, enabling this feature will require reprovisioning of a node.
```
msau42, line 829: Could someone install and launch csi-proxy without rebooting or creating a new node?

The reprovisioning of a node seems like a kube-up specific thing?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+<!--
+Try to be as paranoid as possible - e.g., what if some components will restart
+mid-rollout?
+
+Be sure to consider highly-available clusters, where, for example,
+feature flags will be enabled on some API servers and not others during the
+rollout. Similarly, consider large clusters and how enablement/disablement
+will rollout across nodes.
+-->
+It is up to the cloud provider to tie CSI-proxy version to cluster version. When clusters rollout and rollback, the CSI-proxy should be udpated accordingly.
```
msau42, line 881: Is it a requirement to tie the csi-proxy version to a Kubernetes version?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+
+###### What specific metrics should inform a rollback?
+
+<!--
+What signals should users be paying attention to when the feature is young
+that might indicate a serious problem?
+-->
+For detecting fleet-wide trends, volume_operation and csi_operation metrics performed on Windows node could be used for indicating issues on CSI-proxy.
```
msau42, line 889: Note that these metrics are exposed by kubelet.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -768,12 +777,366 @@ Downgrading the version of csi-proxy.exe to one that is not supported by all ins
...
+###### Are there any tests for feature enablement/disablement?
+
+<!--
+The e2e framework does not currently support enabling or disabling feature
+gates. However, unit tests in each component dealing with managing data, created
+with and without the feature, are necessary. At the very least, think about
+conversion tests if API types are being modified.
+-->
+No
+### Rollout, Upgrade and Rollback Planning
```
msau42, line 862: @ddebroy @jingxu97 how much of the upgrade strategy section has changed? My understanding is now CSI proxy exposes multiple versioned named pipes to facilitate multiple drivers using different API versions, and drivers upgrading to newer API versions.

And on the driver side, @mauriciopoppe did some new experiments so that drivers could fall back to older versions if a newer API version was not available.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: done

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: updated

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,387 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+  - Describe the mechanism:
+  CSI-proxy is a binary that can be deployed (or running as a service) directly on each Windows node. For example, in GCE, CSI-proxy
+  binary is downloaded and running as Windows service in node [startup script](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/k8s-node-setup.psm1#L424)
+  An environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
+  - Will enabling / disabling the feature require downtime of the control
+    plane?
+  No
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+  If csi-proxy binary is deployed as a binary or windows service during node start up time, enabling this feature will require reprovisioning of a node in this case.
```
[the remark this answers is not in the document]
jingxu97: upadated.

ehashman:
Hey @msau42, this KEP is currently missing from the enhancements spreadsheet, can @kubernetes/sig-windows-leads please add it for inclusion?

ddebroy:
I think @jsturtevant already did that? ref: https://kubernetes.slack.com/archives/CN5JCCW31/p1620756591039500

keps/sig-windows/1122-windows-csi-support/README.md
msau42: I agree technically there is no dependency between csi proxy and kubernetes versions (besides 1.18). However, for csi drivers that are not cloud-specific, their support matrix would be simpler if they could indicate a Kubernetes version and Kubernetes distros adhere to our recommendations. Otherwise one k8s distro on 1.20 may support csi-proxy v1beta1, but another distro may only support csi-proxy v1beta2, etc. This becomes especially complicated when a driver also has to take into consideration the csi sidecar version dependencies.

JamesLaverack:
xref https://github.com/kubernetes/enhancements/issues/1122

jingxu97:
addressed comments. PTAL @ddebroy @msau42 @ehashman

johnbelamaric:
Please set the `latest-milestone` and `stage` in the `kep.yaml`, and create the PRR approval file. You can assign me as PRR approver.

/unassign @ehashman 
/assign @johnbelamaric

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+Pick one of these and delete the rest.
+-->
+
+- [ ] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name:
+  - Components depending on the feature gate:
+- [x] Other
+  - Describe the mechanism:
+  CSI-proxy is a binary that can be deployed (or running as a service) directly on each Windows node. For example, in GCE, CSI-proxy
+  binary is downloaded and running as Windows service in node [startup script](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/windows/k8s-node-setup.psm1#L424)
```
ddebroy, line 836: grammar nit: `running as` => `run as a`

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+  An environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
+  - Will enabling / disabling the feature require downtime of the control
+    plane?
+  No
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+  If csi-proxy binary is deployed as a binary or windows service during node start up time, one way to enable this feature is to reprovision of a node during cluster upgrade.
+  System admin could potenticallyinstall and launch CSI-proxy after node is starting.
```
ddebroy, line 844: nit - space between "potentially" and "install"

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+<!--
+Describe the consequences on existing workloads (e.g., if this is a runtime
+feature, can it break the existing applications?).
+
+NOTE: Also set `disable-supported` to `true` or `false` in `kep.yaml`.
+-->
+It depends on cloud provider implementation. For example in GCE, environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
```
ddebroy, line 862: `cloud provider implementation` => `Kubernetes distribution` since csi-proxy need not be running on cloud hosts only.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+with and without the feature, are necessary. At the very least, think about
+conversion tests if API types are being modified.
+-->
+No
+### Rollout, Upgrade and Rollback Planning
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+It is up to cloud provider to have their own plan. For example, GKE will use version control to rollout, upgrade and rollback this feature.
```
ddebroy, line 882: `cloud provider` => `Kubernetes distribution`

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+- [ ] Other (treat as last resort)
+  - Details:
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+<!--
+Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+implementation difficulties, etc.).
+-->
+No
```
ddebroy, line 996: We can potentially call out enabling prometheus for operations in the future :-)

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
+<!--
+Look at the [existing SLIs/SLOs].
+
+Think about adding additional work or introducing new steps in between
+(e.g. need to do X to start a container), etc. Please describe the details.
+
+[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
+-->
+No
```
ddebroy, line 1091: The SLIs call out statefulsets. How about stating that an in-tree plugin like PD/Azure-Disk/EBS that supports Windows today may use csi-proxy in the future due to CSI Migration. This could introduce higher latencies (such as node mount/initial format/etc) if the pods of the statefulset are targeting Windows and the pods mount PVs backed by CSI plugins using csi proxy.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
+[a] specific latencies and extra resources consumed as a result of driving operations through CSI plugin using CSI proxy vs. In-tree plugin with Support for Windows:
+
+The preliminary investigation using GCE PD CSI driver on this is available [link](https://docs.google.com/document/d/10aXLjJs8HvloY0zQJaMsOMn0S0AyrVyfh8iC9T7A1yg/edit). There are relatively small extra delay caused by CSI proxy for certain volume operations. There will be future performance investigation and improvement work on this area.
+
+[b] the overall resources associated with running operations (such as list disk/format volumes/etc) associated with CSI plugins vs not running CSI plugins on Windows?
+
+The tests performed on this area has not shown noticeable overhead. Plan to test on larger scale in the future.
+
+### Troubleshooting
```
ddebroy, line 1114: How about stating: "CSI Proxy is a thin stateless API layer between CSI node plugins (running on Windows) and Windows OS APIs (surfaced through Win32 or Powershell cmdlets). Troubleshooting of APIs associated with CSI Proxy should be performed using either logs from the CSI plugin or Windows OS event logs."

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+    - Detection: How can it be detected via metrics? Stated another way:
+      how can an operator troubleshoot without logging into a master or worker node?
+    - Mitigations: What can be done to stop the bleeding, especially for already
+      running user workloads?
+    - Diagnostics: What are the useful log messages and their required logging
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+-->
+1. If CSI-proxy is not install
```
ddebroy, line 1142: nit: how about "CSI Proxy binary is not installed and running"

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+2. CSI-proxy version and Nodoe Plugin version do not match
```
ddebroy, line 1143: Spelling nit: Node

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
 Downgrading the version of csi-proxy.exe to one that is not supported by all installed versions the CSI Node Plugins in the cluster will lead to loss of access to data. Further, if a cluster is downgraded from a version of Kubernetes where the plugin watcher supports Windows nodes to one that does not, existing Windows workloads that were using CSI plugins to access storage will no longer have access to the data. This loss of functionality cannot be handled in an elegant fashion.
 
 ### Version Skew Strategy
 
 Beyond the points in the above section (Upgrade/Downgrade strategy), there are no Kubernetes version skew considerations in the context of this KEP.
 
+The mininum Kubernetes version to support CSI Windows is 1.8. Though techinicially the API versions in CSI proxy is independent of Kubernetes versions, we recommand the following versions between CSI proxy and Kubernetes. 
+
+Status | Min K8s Version | Recommended K8s Version
```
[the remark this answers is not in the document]
jingxu97: Yes, but because the PR to update the doc is not yet merged. I will work on that and update the link here once it is merged.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: fixed

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: fixed

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: added

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: good point, added

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: added

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: updated

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: fixed

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
 Downgrading the version of csi-proxy.exe to one that is not supported by all installed versions the CSI Node Plugins in the cluster will lead to loss of access to data. Further, if a cluster is downgraded from a version of Kubernetes where the plugin watcher supports Windows nodes to one that does not, existing Windows workloads that were using CSI plugins to access storage will no longer have access to the data. This loss of functionality cannot be handled in an elegant fashion.
 
 ### Version Skew Strategy
 
 Beyond the points in the above section (Upgrade/Downgrade strategy), there are no Kubernetes version skew considerations in the context of this KEP.
 
+The mininum Kubernetes version to support CSI Windows is 1.8. Though techinicially the API versions in CSI proxy is independent of Kubernetes versions, we recommand the following versions between CSI proxy and Kubernetes.
```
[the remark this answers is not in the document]
jingxu97: udpated

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -758,22 +767,397 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+As explained in (#csi-proxy-grpc-api-graduation-and-deprecation-policy), CSI proxy will expose multiple API versions according to the API versioning policy, to facilitate drivers upgrading to a newer API version and multiple drivers using different API versions.
+
+In order to install a CSI Node Plugin or upgrade to a version of a CSI Node Plugin that uses an updated version of the CSIProxyService API not supported by the currently deployed version of csi-proxy.exe in the cluster, csi-proxy.exe will need to be upgraded first on all nodes of the cluster before deploying or upgrading the CSI Node Plugin.
```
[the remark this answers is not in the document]
jingxu97: updated all the places

jingxu97:
@johnbelamaric I updated PR. PTAL. Thank you!

ddebroy (approved):
/lgtm
/approve

keps/sig-windows/1122.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 1122
```
johnbelamaric, line 1: this file is in the wrong directory. it should be `keps/prod-readiness/sig-windows/1122.yaml`

jingxu97: fixed it. Thank you!

keps/prod-readiness/sig-windows/1122.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 1122
+stable:
+  approver: "@johnbelamaric"
```
johnbelamaric, line 3: thanks! handing off to @deads2k though

deads2k, line 3: I'll pick this one up.

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+<!--
+Describe the consequences on existing workloads (e.g., if this is a runtime
+feature, can it break the existing applications?).
+
+NOTE: Also set `disable-supported` to `true` or `false` in `kep.yaml`.
+-->
+It depends on Kubernetes distribution. For example in GCE, environment variable ENABLE_CSI_PROXY is implemented to disable/enable CSI-proxy installation.
```
deads2k, line 862: If I disable this feature, do I need to restart my windows nodes to disconnect previously mounted volumes?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+It is up to the cloud provider to tie CSI-proxy version to cluster version. When clusters rollout and rollback, the CSI-proxy should be updated accordingly.
+The minimum version of Kubernetes required for CSI Windows support is 1.18
+
+###### What specific metrics should inform a rollback?
+
+<!--
+What signals should users be paying attention to when the feature is young
+that might indicate a serious problem?
+-->
+For detecting fleet-wide trends, volume_operation and csi_operation metrics exposed by kubelet on Windows node could be used for indicating issues on CSI-proxy.
```
deads2k, line 905: Is there any way to know how many nodes have CSI volumes enabled for windows?

Does the scheduler have any way to know which windows nodes are capable of providing CSI volumes?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+For instance, if this is a pod-related feature, it should be possible to determine if the feature is functioning properly
+for each individual pod.
+Pick one more of these and delete the rest.
+Please describe all items visible to end users below with sufficient detail so that they can verify correct enablement
+and operation of this feature.
+Recall that end users cannot usually observe component logs or access metrics.
+-->
+
+- [x] Events
+  - Event Reason: Windows Pods which are using the PVCs (using CSI Node plugin) can start.
```
deads2k, line 950: Do I also get an event for those PVCs which are failing?

Can I get metrics showing how many failed volume bindings there are per kubelet?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+    - Detection: How can it be detected via metrics? Stated another way:
+      how can an operator troubleshoot without logging into a master or worker node?
+    - Mitigations: What can be done to stop the bleeding, especially for already
+      running user workloads?
+    - Diagnostics: What are the useful log messages and their required logging
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+-->
+1. CSI Proxy binary is not installed and running
```
deads2k, line 1144: how can I detect that this is happening?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+2. CSI-proxy version and Node Plugin version do not match
+
+The error message from CSI Node plugin will show, for example
```
deads2k, line 1147: is this for case 1 or case 2?

keps/sig-windows/1122-windows-csi-support/README.md
```diff
@@ -747,33 +756,410 @@ All E2E storage tests covering CSI plugins will be ported to Windows workloads a
...
+```
+"transport: Error while dialing open \\\\.\\\\pipe\\\\csi-proxy-disk-v1beta2: The system cannot find the file specified."
```
deads2k, line 1149: Is there a way to expose this (or a similar, identifiable message) in a kubelet log or an event?

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: if disable this feature, you cannot use CSI driver to disconnect mounted volumes. You could RDP to windows node, and run command line to remove the mount point,

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: scheduler does not know about which node is capable of using CSI (has CSI plugin installed). This is true for both linux and Windows.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: In Pod event, there will be error message related to volume mount failed. 

there is operation error rate metric

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: the error message is for both case 1 and 2.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: both has the same error message. if CSI proxy version is lower than Node plugin requested API version, it is basically the same thing as CSI-proxy is not running.

keps/sig-windows/1122-windows-csi-support/README.md
jingxu97: this error message is shown too in events for Pod which is trying to mount PVC.

jingxu97: I will update this with the information

keps/sig-windows/1122-windows-csi-support/README.md
msau42: Should we mention that nodes should be drained first before uninstalling csi-proxy?

jingxu97: sure, that's a good point.

keps/sig-windows/1122-windows-csi-support/README.md
mauriciopoppe: I think we could have a section about the steps we discussed to upgrade the client first (pd-csi) and then the server (csi-proxy binary) independently.

keps/prod-readiness/sig-windows/1122.yaml
deads2k, line 3: can you update to @deads2k so we can keep track?

deads2k:
the PRR looks good.  Please update the approver to @deads2k if you would.  I'll approve to be sure you don't miss freeze

/approve

ddebroy (commented):
/lgtm

PR #2677, opened 2021-05-05
Initialize map before inserting elements

swetharepakula:
Validation errors are masked because of the following error:
```
panic: assignment to entry in nil map [recovered]
panic: assignment to entry in nil map
```

Example run: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/enhancements/2673/pull-enhancements-verify/1389750336436047872


ehashman:
/close

duplicate of #2676

PR #2689, opened 2021-05-07
cgroupv2: update latest-milestone

giuseppe:

### DIFF
--- a/keps/sig-node/2254-cgroup-v2/README.md
+++ b/keps/sig-node/2254-cgroup-v2/README.md
@@ -210,7 +210,8 @@ these controllers is to use eBPF.
   A good candidate for running cgroup v2 test is Fedora 31 that has
   already switched to default to cgroup v2.
 
-- Beta: e2e tests coverage and performance testing.
+- Beta: e2e tests coverage and performance testing.  Verify that both
+  the CPU and Memory Manager work.
 
 - GA: Assuming no negative user feedback based on production
   experience, promote after 2 releases in beta.
--- a/keps/sig-node/2254-cgroup-v2/README.md
+++ b/keps/sig-node/2254-cgroup-v2/README.md
@@ -9,6 +9,18 @@
 - [Non-Goals](#non-goals)
 - [User Stories](#user-stories)
 - [Implementation Details](#implementation-details)
+- [Design](#design)
+  - [Test Plan](#test-plan)
+    - [Needed Tests](#needed-tests)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
 - [Proposal](#proposal)
   - [Dependencies on OCI and container runtimes](#dependencies-on-oci-and-container-runtimes)
   - [Current status of dependencies](#current-status-of-dependencies)
@@ -17,7 +29,6 @@
   - [Phase 1: Convert from cgroups v1 settings to v2](#phase-1-convert-from-cgroups-v1-settings-to-v2)
   - [Phase 2: Use cgroups v2 throughout the stack](#phase-2-use-cgroups-v2-throughout-the-stack)
 - [Risk and Mitigations](#risk-and-mitigations)
-- [Graduation Criteria](#graduation-criteria)
 <!-- /toc -->
 
 ## Summary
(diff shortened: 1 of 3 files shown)


giuseppe:
@ehashman is this enough or is there more metadata I need to add/modify?

mrunalp:
Can we capture that we have to check that CPU manager/Memory manager works with cgroups v2?

mrunalp:
We are also missing PRR review for this since this KEP predates that process. We will have to get it updated for PRR.

keps/sig-node/2254-cgroup-v2/kep.yaml
```diff
@@ -21,5 +21,5 @@ see-also:
 replaces:
 superseded-by:
 
-latest-milestone: "0.0"
+latest-milestone: "v1.19"
 stage: "alpha"
```
ehashman, line 25: Latest milestone should be 1.22 / stage beta, right?

giuseppe: correct.

I'll fix it, thanks for the review!

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -210,7 +210,8 @@ these controllers is to use eBPF.
   A good candidate for running cgroup v2 test is Fedora 31 that has
   already switched to default to cgroup v2.
 
-- Beta: e2e tests coverage and performance testing.
+- Beta: e2e tests coverage and performance testing.  Verify that both
+  the CPU and Memory Manager work.
```
ehashman, line 215: @giuseppe please ensure you add test plans and fill out the PRR questionnaire. See https://github.com/kubernetes/enhancements/blame/master/keps/NNNN-kep-template/README.md#L335

ehashman (changes requested):

keps/sig-node/2254-cgroup-v2/README.md
giuseppe: added the tests plan to the KEP

johnbelamaric: We need the PRR questions filled out too please.

giuseppe: thanks for the review.  Added now

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,185 @@ This proposal aims to:
...
+N/A.  Each node can be configured separately.
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+N/A.  It requires a reboot to be enabled.  If the workload accesses directly the
+cgroup file system, then also the workload must be enabled for cgroup v2.
+
+###### What specific metrics should inform a rollback?
+
+Pods not being healthy.
```
johnbelamaric, line 152: Maybe pods not getting the right resource constraints? Any way to detect that?

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,185 @@ This proposal aims to:
...
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+The cgroup file system inside of the containers will use cgroup v2 instead of cgroup v1.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+Looking at the node configuration.  If the node is using cgroup v2, then also the pods
```
johnbelamaric, line 166: Any specific commands we can run? Any metrics or Node status fields?

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,185 @@ This proposal aims to:
...
+- [ ] Metrics
+  - Metric name:
+  - [Optional] Aggregation method:
+  - Components exposing the metric:
+- [X] Other (treat as last resort)
+  - Details: not a service
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+N.
```
johnbelamaric, line 195: typo

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,185 @@ This proposal aims to:
...
+
+N/A
+
+###### What are other known failure modes?
+
+N/A
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+Reboot the node on cgroup v1
```
johnbelamaric, line 243: That would be to fix it. Which is fine but maybe say "If SLOs are not being met, reboot the node in cgroup v1 to disable this feature."

keps/sig-node/2254-cgroup-v2/README.md
mrunalp: One could inspect the pod cgroups. However, they will have to look at the conversion table in this KEP as some field values map directly while others don't.

keps/sig-node/2254-cgroup-v2/README.md
mrunalp: One can run `cat /proc/self/cgroup` on the node to check if it is running in cgroups v2 mode. Node status field sounds like a good idea to me. @derekwaynecarr wdyt?

derekwaynecarr: since enabling cgroup v2 on a linux host is a prerequisite action to even running a kubelet, and you would not rollout a linux host on v2 without special consideration, i dont see any unique value in reporting kubelet is on a v2 host versus what i would see in normal monitoring systems on those hosts.

odinuge:
Sorry to be a bit late at the game now, as I didn't see this before now, but;

> @mrunalp: Can we capture that we have to check that CPU manager/Memory manager works with cgroups v2?

Well, it does not work right now. Since we constantly update all containers, runc and their device handling will fail, resulting in k8s just crashing. With crun it sort of work, but some design choices in crun make it fail with a lot of other stuff (metrics etc.). :/

PR #2691, opened 2021-05-07
Update random pod scaledown KEP for beta

damemi:
Ref: https://github.com/kubernetes/enhancements/issues/2185

### DIFF
--- a/keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
+++ b/keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
@@ -34,15 +34,15 @@
 
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
 - [x] (R) KEP approvers have approved the KEP status as `implementable`
 - [x] (R) Design details are appropriately documented
 - [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
 - [x] (R) Graduation criteria is in place
 - [x] (R) Production readiness review completed
 - [x] Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
-- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [x] "Implementation History" section is up-to-date for milestone
+- [x] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 
@@ -206,6 +206,7 @@ Alpha (v1.21):
 
 Beta (v1.22): 
 - Enable LogarithmicScaleDown feature gate by default
+- Enable `deleted_pod_age_ratio` metric
 
 Stable (v1.23):
 - Remove LogarithmicScaleDown feature gate
@@ -263,46 +264,48 @@ _This section must be completed when targeting alpha to a release._
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can a rollout fail? Can it impact already running workloads?**
-  Try to be as paranoid as possible - e.g., what if some components will restart
-   mid-rollout?
+  This should not affect running workloads, though there is the possibility that the logic 
+  panics which would cause kube-controller-manager to crash
 
 * **What specific metrics should inform a rollback?**
+  Increased pod deletions could indicate runaway/hot-loop failures in the scaledown logic.
+  Availability of applications may also be affected. Though the intent of this is to provide 
+  better available through more distributed victim selection, in cases of desired binpacking 
+  pods may remain running on undesired nodes.
 
 * **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
-  Describe manual testing that was done and the outcomes.
-  Longer term, we may want to require automated upgrade/rollback tests, but we
-  are missing a bunch of machinery and tooling and can't do that now.
+  This will be manually tested before the graduation to beta
 
 * **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
 fields of API types, flags, etc.?**
-  Even if applying deprecation policies, they may still surprise some users.
+  No
 
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can an operator determine if the feature is in use by workloads?**
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+  The scaledown behavior of all replicasets will be affected by this featuregate being 
+  enabled, so somehow monitoring them will be necessary to determine it
 
 * **What are the SLIs (Service Level Indicators) an operator can use to determine 
 the health of the service?**
-  - [ ] Metrics
-    - Metric name:
+  - [x] Metrics
+    - Metric name: deleted_pod_age_ratio
     - [Optional] Aggregation method:
-    - Components exposing the metric:
+    - Components exposing the metric: kube-controller-manager
   - [ ] Other (treat as last resort)
-    - Details:
+  
+The metric `deleted_pod_age_ratio` will provide a histogram of the ratio between the 
+chosen `deleted pod`'s age over the current `youngest pod`'s age, for pods where the sort 
+algorithm falls back to age. (Pod age is the final criteria in the sorting algorithm, so we don't 
+want to measure this ratio for deletions which don't use this feature, as those may validly fall 
+outside the desired range).
 
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+  There should be no values `>2` in the above metric when the Pod Cost annotation is unset 
+  (see https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/2255-pod-cost) and 
+  the pod's deletion was based on a timestamp comparison (rather than, for example, pod state).
 
 * **Are there any missing metrics that would be useful to have to improve observability 
 of this feature?**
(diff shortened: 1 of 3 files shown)


damemi (commented):
/cc @alculquicondor @soltysh 
I've tried updating this with the PRR answers required for beta. Though, imo many of them seem irrelevant since this is a minor change to a previously-undocumented behavior. For example, I don't think we need a metric to expose this behavior (but it could be easy to add a simple boolean, or in some way measure the logarithmic performance of cumulative scaledowns on the cluster). Please let me know if you feel there are better responses we could provide.

alculquicondor:
You also need to update the file in `/keps/prod-readiness`

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
...
 
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can an operator determine if the feature is in use by workloads?**
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+  Only by the presence of the feature gate
```
alculquicondor, line 283: Explicitly say that all replicasets are affected.

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
...
     - Metric name:
     - [Optional] Aggregation method:
     - Components exposing the metric:
-  - [ ] Other (treat as last resort)
-    - Details:
+  - [x] Other (treat as last resort)
+    - Details: The health of this feature is dependent on the overall health of the 
+  control plane services, specifically kube-controller-manager but also to an extent 
+      services like API server, kubelet, etc.
+      We could add a metric to KCM to indicate that this feature gate is enabled
```
alculquicondor, line 295: Remove this speculation. I don't think we should do that.

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can a rollout fail? Can it impact already running workloads?**
-  Try to be as paranoid as possible - e.g., what if some components will restart
-   mid-rollout?
+  Rollouts should have no effect, as this only relates to scaledown of replicasets
 
 * **What specific metrics should inform a rollback?**
+N/A
```
alculquicondor, line 269: If there were metrics for number of pods deleted, maybe those could be used. Like, if there's a problem on the implementation, potentially you could see an uptick of pod deletions.

Another thing could be that the availability of applications is affected. Although the intent of the feature is that availability increases (because of better spreading).

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
...
 * **What are the SLIs (Service Level Indicators) an operator can use to determine 
 the health of the service?**
   - [ ] Metrics
     - Metric name:
     - [Optional] Aggregation method:
     - Components exposing the metric:
-  - [ ] Other (treat as last resort)
-    - Details:
+  - [x] Other (treat as last resort)
+    - Details: The health of this feature is dependent on the overall health of the
```
wojtek-t, line 292: This isn't helpful - it's true for almost any feature.

What we need is a signal to determine if the kcm is doing what it should be doing.

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
...
+  Rollouts should have no effect, as this only relates to scaledown of replicasets
 
 * **What specific metrics should inform a rollback?**
+N/A
 
 * **Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?**
-  Describe manual testing that was done and the outcomes.
-  Longer term, we may want to require automated upgrade/rollback tests, but we
-  are missing a bunch of machinery and tooling and can't do that now.
+  N/A
```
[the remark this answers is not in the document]
alculquicondor: Sure, I can do it once the code changes are in.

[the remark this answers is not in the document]
damemi: > Sure, I can do it once the code changes are in.

Would that just be the beta promotion change https://github.com/kubernetes/kubernetes/pull/101767? I'm also confused about how this should be tested with the upgrade->downgrade->upgrade path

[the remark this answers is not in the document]
wojtek-t: Please add a text that you will test that before the feature will graduate to beta.
Also - please update the KEP within the release cycle [see https://github.com/kubernetes/enhancements/pull/2538 as an example where we did that.]

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,36 @@ _This section must be completed when targeting alpha to a release._
...
     - Components exposing the metric:
-  - [ ] Other (treat as last resort)
-    - Details:
+  - [x] Other (treat as last resort)
+    - Details: The health of this feature is dependent on the overall health of the 
+  control plane services, specifically kube-controller-manager but also to an extent 
+      services like API server, kubelet, etc.
+      We could add a metric to KCM to indicate that this feature gate is enabled
 
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
```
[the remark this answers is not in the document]
wojtek-t: I understand the intention, but I don't think it's a good SLO - SLI/SLO needs precise definition and this value will heavily depend on the load (and can change over time).

That may be a good metric informing rollback though...

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,39 @@ _This section must be completed when targeting alpha to a release._
...
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can an operator determine if the feature is in use by workloads?**
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+  The scaledown behavior of all replicasets will be affected by this featuregate being 
+  enabled, so somehow monitoring them will be necessary to determine it
```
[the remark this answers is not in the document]
wojtek-t: I think the existing answer is fine - it will be used by all workloads.

Now we just need a way to verify that it really works - but it's more the SLI question below.

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
wojtek-t: Can we come up with some metrics for this? Even if whitebox-y, that's still much better than nothing.

Maybe we can expose histogram of something like:
- |age of deleted pod| / |age of yougest pod|
[We expect all values to be within [1,2], so having values outside of it is a good signal that something is wrong.

[Well - technically it will collide with https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/2255-pod-cost so maybe we additionally need a metric label whether such annotation was set or not]
[And then the SLO would be that there are no values >2 when the pod_cost_set=false]

damemi: > |age of deleted pod| / |age of yougest pod|

are these currently available, or is there work needed to add them to KCM? @soltysh where are the KCM metrics set up so that we can start work on that?

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,46 +263,40 @@ _This section must be completed when targeting alpha to a release._
...
 
 * **What are the SLIs (Service Level Indicators) an operator can use to determine 
 the health of the service?**
-  - [ ] Metrics
-    - Metric name:
+  - [x] Metrics
+    - Metric name: deleted_pod_age_ratio
```
wojtek-t, line 292: If this is going to be the metric that I proposed in: https://github.com/kubernetes/enhancements/pull/2691#discussion_r629877889

can you please define it somewhere in the KEP (and probably add a beta criteria that this metric is implemented)?

[If it's going to be a different metric, please define that too :) ]

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,46 +263,40 @@ _This section must be completed when targeting alpha to a release._
...
 
 * **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+  There should be no values `>2` in the above metric when the Pod Cost annotation is set
```
wojtek-t, line 298: s/set/unset/, right?

[Or to be more specific - if this is unset for all pods of the replicaset...]

Basically (it's a comment toward my ask above to define the metric in the kep), I think that the metric should have a field of "pod-cost-annotation-exists" (better name welcome :-) ) that replicaset controller will be computing and setting when exporing  that metric (and then the check is trivial).

wojtek-t (commented):
Some suggestions/answers - hope these are useful.

keps/sig-apps/2185-random-pod-select-on-replicaset-downscale/README.md
```diff
@@ -263,37 +263,39 @@ _This section must be completed when targeting alpha to a release._
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can a rollout fail? Can it impact already running workloads?**
-  Try to be as paranoid as possible - e.g., what if some components will restart
-   mid-rollout?
+
```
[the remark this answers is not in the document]
soltysh: Maybe something along:

It shouldn't impact already running workloads. One possible problem during rollout is a panic in the algorithm which might crash controller-manager.

soltysh (approved):
One suggestion for that one missing bit, the rest
/lgtm
/approve

damemi:
Updated with the latest feedback, and squashed.

One note about the metric (mentioned this to @alculquicondor offline) is that the defined range of `[1,2]` is only valid for pods where their deletion was based on sorting by timestamp. 

Similar to how this conflicts with the Pod Cost Annotation, there are [other criteria](https://github.com/damemi/kubernetes/blob/a8d105ab724ccdb45b4ff380d84fd356e833991e/pkg/controller/controller_utils.go#L792-L801) which could put a pod older than `2 * youngest_pod_age` as higher priority for deletion, such as a `Pending` status phase. I've tried to make this clear in the definitions of the new metric.

wojtek-t:
/lgtm
/approve

Thanks!

PR #2693, opened 2021-05-07
Client Exec Proxy

nckturner:
A KEP to propose a client-side proxy for a number of client authentication use cases.

### DIFF
--- a/keps/prod-readiness/sig-auth/2718.yaml
+++ b/keps/prod-readiness/sig-auth/2718.yaml
@@ -1,0 +1,3 @@
+kep-number: 2718
+alpha:
+  approver: "@johnbelamaric"
--- a/keps/sig-auth/2718-20210511-client-exec-proxy/README.md
+++ b/keps/sig-auth/2718-20210511-client-exec-proxy/README.md
@@ -1,0 +1,853 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-2718: Client Executable Proxy
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Proxy Proposal](#proxy-proposal)
+  - [Authentication to Proxy](#authentication-to-proxy)
+  - [Caching](#caching)
+  - [Proxy Shutdown](#proxy-shutdown)
+  - [API](#api)
+  - [Test Plan](#test-plan)
+    - [Integration](#integration)
+    - [E2E](#e2e)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative Proposal: Request Replacement](#alternative-proposal-request-replacement)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+Client authentication in Kubernetes can be achieved via one of a handful of
+mechanisms: by passing a bearer token directly to kubectl using the `--token`
+flag, by inserting a bearer token into the kubeconfig, by inserting a client
+certificate and key into the kubeconfig, or by defining an exec-based token
+plugin in the kubeconfig which supports calling an arbitrary binary which is
+expected to return a bearer token and/or client certificate/key.  The client
+exec proxy adds an extension point for additional authentication mechanisms
+that are difficult to set up with the current architecture.  Client request
+signing, and an external TLS certificate authenticator are two use cases that
+would benefit from this proposal.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+The motivation for this KEP stems from the inability of the existing
+authentication mechanisms to satisfy the author's requirement to implement a
+Kubernetes client that implements AWS's request signing process, known as
+Signature Version 4 (sig-v4)
+[[1]](https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html)
+[[2]](https://github.com/kubernetes/kubernetes/issues/92535). In order to
+implement sig-v4, a canonical version of the request must be created, combined
+with any required additional metadata, signed by the secret key, and the
+signature appended to the request in a header or query string parameter.  In
+order to achieve this, the request signer must have access to the entire
+request, so the existing exec-based authentication is unsatisfactory.  A
+generic solution is desirable to satisfy other future request signing or
+request modifying feature requests.
+
+The solution that exists today is to use the existing proxy configuration in
+client-go.  The includes configuring a proxy via the environment variables,
+HTTP_PROXY or HTTPS_PROXY, or using the explicit proxy URL supported by
+kubeconfig [[3]](https://github.com/kubernetes/kubernetes/pull/81443).  The
+problem with this solution is that the operator experience of setting up and
+managing an external proxy is poor.  For example, to implement a request
+signing proxy for kubelet, one has to manage the proxy process with an init
+system of some kind, and secure the proxy endpoint, both of which require
+significant, non-trivial configuration outside of Kubernetes.
+
+Another use case is using an external TLS certificate signer
+[[4]](https://github.com/kubernetes/enhancements/pull/1749).  This would allow
+usage of TLS client certificates within client-go, by delegating digital key
+operations to external processes, for example, an HSM.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+The goals of this KEP include:
+1. Defining a specification which enables Kubernetes clients to do arbitrary
+request modification before sending the request to the API server, via an
+extensible mechanism to support the following use cases:
+    - Client request signing.
+    - Using hardware protected keys for request signing or mTLS.
+    - Adding or manipulating arbitrary headers to the request.
+2. Defining a specification which would provide the ability for the client-side
+proxy server to be executed and communicated with according to configuration
+solely contained in a kubeconfig file.
+3. Defining a specification which is interoperable with an existing client-side
+proxy setup, e.g. a proxy configured with the HTTP_PROXY environment variable.
+4. Defining a specification for a client-side proxy server that is easy to
+secure without taking additional setup steps, by:
+    - Using a unix domain socket for the communication between the client and
+      the proxy, secured with file permissions and TLS.
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+The following are non-goals of this KEP:
+1. Solving use cases outside of the above-mentioned client authentication
+   workflows.
+2. Implementing a specific flavor of request signing (like SigV4), or adding
+   support for any hardware security module protocol directly in the Kubernetes
+   client libraries.
+3. Defining any new mechanisms for validating request signatures server-side.
+   This is because they can be validated by a front-proxy, which is already a
+   feature of Kubernetes.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+This proposal is to implement an extension in Kubernetes client libraries to
+forward requests to a proxy process for request modification.  The proxy will
+receive an HTTP request and will have the opportunity to sign the request,
+attach headers, use hardware protected keys for mTLS, and/or modify the
+request.  The final request will be constructed and forwarded to the apiserver.
+
+It is already possible, by setting the appropriate HTTP_PROXY, HTTPS_PROXY
+environment variables, to configure a Kubernetes client to send requests to a
+proxy.  While this may work in some situations, it is not ideal for a generic
+request signing solution across all client configurations.  Firstly, http
+proxies are already commonly integrated in client networking configurations, so
+for these clients setting up an additional proxy configuration might not be
+possible, or at least not straightforward.  Secondly, this approach requires a
+long running external proxy process which introduces operational complexity.
+Thirdly, communicating with a proxy for request signing requires a secure
+communication channel between the client and the proxy, which means in a case
+where multiple users share a client computer (for example, a shared bastion),
+configuring and securing a local proxy to prevent signing requests as a
+different user becomes more complicated.
+
+### Risks and Mitigations
+
+<!--
+What are the risks of this proposal, and how do we mitigate? Think broadly.
+For example, consider both security and how this will impact the larger
+Kubernetes ecosystem.
+
+How will security be reviewed, and by whom?
+
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+The risks of this proposal are similar to the existing exec credential
+provider.  There are risks associated with execing a binary, for example in a
+scenario where users are downloading binaries from untrustworthy sources to use
+in their cluster.  However, this is not a departure from existing security
+posture of the exec credential feature present in k8s.io/client-go today.
+
+## Design Details
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Proxy Proposal
+
+The proxy exec plugin is a long-running proxy process which is executed by the
+client on first use.  When executed, the exec proxy plugin will bind and listen
+on a socket, and will a output URL on stdout, over which the client would proxy
+all API server bound requests.  The proxy binary location is defined in the
+`ExecConfig` section of a kubeconfig `user`.
+
+The proxy plugin process will be executed from client-go the first time it is
+needed, on transport setup.  If the client fails to connect to the socket
+because no one is listening, client-go will reattempt to exec the plugin.
+Client-go will support both plugins binaries which are themselves the proxy
+process, and additionally plugin binaries which complete, but are responsible
+for starting a proxy process and returning a URL to the proxy.
+
+The client will send its request to the proxy plugin process, and the proxy
+will forward the request to the API server, rather than returning the modified
+request to the client to be sent (see alternatives considered).  The downside
+of this approach is that the proxy must duplicate the some of the transport
+functionality of client packages, like respecting existing proxy configuration,
+but allows for more flexibility for proxy plugin developers.
+
+### Authentication to Proxy
+
+Options:
+
+1. The client will authenticate to its proxy using mTLS.  On exec, the proxy
+   will pass back client key and certificate authority data for the client to
+   use for authentication when establishing a connection.
+2. An alternative method for authentication to the proxy would be file
+   permission based.  The GID owner of the client process could passed to the
+   proxy over stdin, which the proxy should give read/write permissions on the
+   socket to.  However, this could be less portable if we need to use a non UDS
+   communication mechanism for windows, or if we support TCP/IP in the future.
+3. A third mechanism for authentication for windows based systems would be
+   using named pipes, which support the Win32 access control model.
+
+An option will be chosen during implemenation of alpha.
+
+### Caching
+
+The connection from the client to the proxy can be cached in the client based
+on the cluster details, the existing HTTP_PROXY configuration, the exec-proxy
+URL, and the certificates used for mTLS.
+
+### Proxy Shutdown
+
+In the case where the client dies unexpectedly, due to a crash or SIGKILL, the
+proxy should also die.  To handle this, the proxy will attempt to take a file
+lock that the parent holds and quit if it ever successfully takes the lock.
+Before exec, the client should take the lock.  The lock file path will be
+provided to the proxy in the ExecCredentialSpec.
+
+Alterntatively, the proxy could quit after a certain amount of idle time.  For
+example, proxies that are idle for more than 5 minutes will be required to
+quit.  Idle quit is simpler, but the lock approach handles the kubectl use case
+better, so it is preferred.
+
+### API
+
+The communication over stdin and stdout with the exec'd binary will be done
+with using the existing ExecCredentialSpec and ExecCredentialStatus objects.
+
+```go
+// ExecCredentialSpec holds request and runtime specific information provided by
+// the transport.
+type ExecCredentialSpec struct {
+	// Cluster contains information to allow an exec plugin to communicate with the
+	// kubernetes cluster being authenticated to. Note that Cluster is non-nil only
+	// when provideClusterInfo is set to true in the exec provider config (i.e.,
+	// ExecConfig.ProvideClusterInfo).
+	// +optional
+	Cluster *Cluster `json:"cluster,omitempty"`
+
+	// Interactive declares whether stdin has been passed to this exec plugin.
+	Interactive bool `json:"interactive"`
+
+	// New:
+
+	// LockFilePath specifies a lock file to be taken by the client
+	// +optional
+	LockFilePath string `json:"lockFilePath,omitempty"`
+}
+
+// ExecCredentialStatus holds credentials for the transport to use.
+//
+// Token and ClientKeyData are sensitive fields. This data should only be
+// transmitted in-memory between client and exec plugin process. The Exec plugin
+// itself should at least be protected via file permissions.
+type ExecCredentialStatus struct {
+	// ExpirationTimestamp indicates a time when the provided credentials expire.
+	// +optional
+	ExpirationTimestamp *metav1.Time `json:"expirationTimestamp,omitempty"`
+	// Token is a bearer token used by the client for request authentication.
+	Token string `json:"token,omitempty" datapolicy:"token"`
+	// PEM-encoded client TLS certificates (including intermediates, if any).
+	ClientCertificateData string `json:"clientCertificateData,omitempty"`
+	// PEM-encoded private key for the above certificate.
+	ClientKeyData string `json:"clientKeyData,omitempty" datapolicy:"security-key"`
+
+	// New:
+
+	// ProxyConfig is the config returned by the exec-proxy
+	// +optional
+	ProxyConfig *ProxyConfig `json:"proxyConfig,omitempty"`
+}
+
+
+// ProxyConfig configures the client to proxy requests via the returned address.
+type ProxyConfig struct {
+	// URL is the location of the exec proxy listener.
+	URL string `json: "url"`
+
+	CertificateAuthorityData []byte `json:"certificate-authority-data"`
+}
+```
+
+### Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation, and anything particularly
+challenging to test, should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing
+guidelines][testing-guidelines] when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+Both integration and e2e tests will be required in addition to unit tests.
+
+#### Integration
+
+Integration tests will cover:
+
+* Authentication between client and proxy.
+* Connection from client to apiserver through proxy.
+* Proxy use with kubectl (possible test location: [command-line integration
+  test
+  suite](https://github.com/kubernetes/kubernetes/tree/ec560b9737537be8c688776461bc700e8ddedb9d/test/cmd)).
+* UDS connections from the client to the proxy will be tested.
+
+#### E2E
+
+The e2e tests will test communication from client to apiserver through proxy in
+a real cluster. The e2e tests will require a proxy implementation and
+apiserver, etcd, and a client.
+
+### Graduation Criteria
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. The KEP
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this enhancement:
+- [Maturity levels (`alpha`, `beta`, `stable`)][maturity-levels]
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc
+definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning)
+or by redefining what graduation means.
+
+In general we try to use the same stages (alpha, beta, GA), regardless of how the
+functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+#### Alpha -> Beta Graduation
+
+- Gather feedback from developers and surveys
+- Complete features A, B, C
+- Tests are in Testgrid and linked in KEP
+
+#### Beta -> GA Graduation
+
+- N examples of real-world usage
+- N installs
+- More rigorous forms of testing—e.g., downgrade tests and scalability tests
+- Allowing time for feedback
+
+**Note:** Generally we also wait at least two releases between beta and
+GA/stable, because there's no opportunity for user feedback, or even bug reports,
+in back-to-back releases.
+
+#### Removing a Deprecated Flag
+
+- Announce deprecation and support policy of the existing flag
+- Two versions passed since introducing the functionality that deprecates the
+  flag (to address version skew)
+- Address feedback on usage/changed behavior, provided on GitHub issues
+- Deprecate the flag
+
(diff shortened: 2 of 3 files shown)


keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+In general we try to use the same stages (alpha, beta, GA), regardless of how the
+functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+#### Alpha -> Beta Graduation
```
enj, line 407: For this to graduate to beta I would like to see an example plugin command that does TLS offload based on some local key pair (i.e. the TLS bits to the API server are being handled by the proxy and not kubectl).  This would then be used in e2e tests to prove out the functionality.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+The goals of this KEP include:
+1. Defining a specification which enables Kubernetes clients to do arbitrary request modification before sending the request to the API server, via an extensible mechanism to support the following use cases:
+    - Client request signing.
+    - Using hardware protected keys for request signing or mTLS.
+    - Adding or manipulating arbitrary headers to the request.
+2. Defining a specification which would provide the ability for the client-side proxy server to be executed and communicated with according to configuration solely contained in a kubeconfig file.
+3. Defining a specification which is interoperable with an existing client-side proxy setup, e.g. a proxy configured with the HTTP_PROXY environment variable.
+4. Defining a specification for a client-side proxy server that is easy to secure without taking additional setup steps, by:
+    - Using a unix domain socket for the communication between the client and the proxy, secured with file permissions.
```
enj, line 184: Please remove the UDS bits.  That is a separate feature KEP that is orthogonal to this one.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+The goals of this KEP include:
+1. Defining a specification which enables Kubernetes clients to do arbitrary request modification before sending the request to the API server, via an extensible mechanism to support the following use cases:
+    - Client request signing.
+    - Using hardware protected keys for request signing or mTLS.
+    - Adding or manipulating arbitrary headers to the request.
+2. Defining a specification which would provide the ability for the client-side proxy server to be executed and communicated with according to configuration solely contained in a kubeconfig file.
```
enj, line 181: Not 100% sure what this trying to say.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+It is already possible, by setting the appropriate HTTP_PROXY, HTTPS_PROXY environment variables, to configure a Kubernetes client to send requests to a proxy.  While this may work in some situations, it is not ideal for a generic request signing solution across all client configurations.  Firstly, http proxies are already commonly integrated in client networking configurations, so for these clients setting up an additional proxy configuration might not be possible, or at least not straightforward.  Secondly, this approach requires a long running external proxy process which introduces operational complexity.  Thirdly, communicating with a proxy for request signing requires a secure communication channel between the client and the proxy, which means in a case where multiple users share a client computer (for example, a shared bastion), configuring and securing a local proxy to prevent signing requests as a different user becomes more complicated.
+
+### User Stories (Optional)
```
enj, line 213: Please fill this out.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+}
+
+```
+
+The ExecProxyConfig includes the address where the proxy is expected to listen.
+```go
+type ExecProxyConfig struct {
+	// The socket path where the proxy should listen.
+	UnixSocketPath string `json:"unixSocketPath"`
+}
```
enj, line 320: This API is exactly backwards, the proxy port is returned by the exec plugin.  It is not input config.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+### Proxy Proposal
+
+The request signing proxy is a long-running proxy process which is executed by the client on first use.  This proposal would add configuration options to allow an executed binary to return a localhost port or UDS socket address, over which the client would proxy all API server bound requests.  An additional configuration section for the kubeconfig defines a proxy binary location.
+
+Questions to answer:
+
+1. What is the lifetime of the proxy plugin?
+
+    The proxy plugin process will be executed from a goroutine (executor) the first time it is needed, i.e. on transport setup.  The executor goroutine will complete when the exec command completes.  The client-go library will have an additional work queue.  If the client fails to connect to the socket because no one is listening, an item will be added to the work queue.  An additional goroutine worker will be waiting for work items, and will spawn a new executor goroutine.
```
enj, line 266: None of this makes sense to be.  The exec plugin is the proxy.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+   The client will send its request to the proxy plugin process, and the proxy will forward the request to the API server, rather than returning the modified request to the client to be sent (see alternatives considered).  The downside of this approach is that the proxy must duplicate the some of the transport functionality of client packages, but allows for more flexibility for proxy plugin developers.
+
+1. Who is responsible for creating the unix domain socket?
+
+   The listening unix domain socket is created by the proxy plugin process.
+
+1. What does the the entire process look like?
+
+   The client attempts a connection to the unix domain socket address (file path) and if it fails, it executes the proxy plugin process, passing the listening address.  The listening unix domain socket is then created by the proxy plugin process and bound to the file path that is passed by the client.  The client then connects to address and upon successful connection, the client sends all requests destined for the kube-apiserver to the proxy plugin.
```
enj, line 282: Drop all the UDS bits.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+	// +optional
+	// The destination Kubernetes cluster.
+	Cluster *Cluster `json:"cluster"`
+}
+
+type ExecProxyResponse struct {
+	metav1.TypeMeta
+}
+```
```
enj, line 344: I would expect new fields in the existing exec plugin code.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+- name: nic@cluster1
+  user:
+    exec:
+      apiVersion: client.authentication.k8s.io/v1alpha1
+      args:
+      - --region=us-west-2
+      command: /usr/local/bin/sigv4proxy
+      execProxy:
+        unixSocketPath: /var/run/kubelet/signer.sock
```
enj, line 357: This is a non-starter.  I want to be able to download a kubeconfig and have it work on any host that has the exec plugin.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: I'm just saying, the configuration should be entirely in the kubeconfig and we shouldn't define any new configuration files for any configuration we might need to pass to the exec proxy plugin.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
```diff
@@ -0,0 +1,711 @@
...
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-NNNN: Client Executable Proxy
```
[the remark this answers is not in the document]
nckturner: Agree, but didn't have a lot of good ideas.  Open to suggestions.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: I think there could be a default proxy port/path, but I imagine that there are many use cases where a user would need to customize this.  I included it in the input as an override of the default, to avoid having to configure it elsewhere, different for each proxy plugin.  But I think either approach is technically possible, right?

keps/sig-auth/NNNN-20201209-client-exec-proxy/kep.yaml
```diff
@@ -0,0 +1,53 @@
...
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.22"
+  beta: "v1.23"
+  stable: "v1.24"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: ClientExecProxy
```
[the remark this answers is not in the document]
nckturner: Yeah I'll remove this -- I was thinking it would protect the feature when it was used in kubelet, for example, but I am not sure that makes sense.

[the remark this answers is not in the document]
nckturner: Or if it does make sense in that context, I can just remove the reference to kubectl.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: Yeah, the exec plugin is (or can be) the proxy.  The only thing that matters is that the proxy is listening on the expected listener after the exec command is run.  The above description is leaving it optional -- the exec plugin can either somehow start the proxy and return, or it can be the proxy and not return, except in error.  In the case where the proxy stop listening, an item will be added to a queue which will trigger a restart attempt of the exec plugin.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: I added a struct to the existing ExecConfig:

	type ExecConfig struct {
		ExecProxy ExecProxyConfig `json:"execProxy"``
	}

For the request/response, I am not a huge fan of the Spec/Status that is used by the exec credential plugin, (those make sense in the controller context, but not here) and I don't think an Exec Credential plugin needs to overlap with an Exec Proxy plugin, but if you feel differently we can discuss. There _are_ duplicate fields, so it could make sense.

keps/sig-auth/2718-20210511-client-exec-proxy/kep.yaml
```diff
@@ -0,0 +1,54 @@
...
+reviewers:
+  - "@mikedanese"
+  - "@enj"
+  - "@liggitt"
+approvers:
+  - "@mikedanese"
+  - "@enj"
+  - "@liggitt"
+prr-approvers:
+  - "@deads2k"
```
deads2k, line 21: Sure.  Also, please create a file like https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-auth/1393.yaml when the overall design is worked out.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: Thinking more, I agree here.  The proxy address being returned by the plugin makes a lot more sense to avoid having to do something to tell clients apart some other way.  I.e. as I understand, UDS clients wouldn't be differentiating themselves if connecting to the same listening socket path, so we'd need to use some protocol on top which would add complexity.  I'll play around with this in a proof-of-concept but currently planning to update the KEP accordingly.  Were there other things I'm not thinking about?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,811 @@
...
+	// +optional
+	ClientCertificate []byte `json:"client-certificate"`
+}
+
+type Address struct {
+	// +optional
+	UnixSocket UnixSocket
+
+	// +optional
+	Port int `json:"port"`
```
mikedanese, line 423: Why not just a URL?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,811 @@
...
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Proxy Proposal
+
+The proxy exec plugin is a long-running proxy process which is executed by the
+client on first use.  The proxy binary location is defined in the `ExecConfig`
```
mikedanese, line 326: Do we have one proxy per process, one per client, one per config, etc...? How does the caching work?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,811 @@
...
+for somehow starting a proxy process and returning its listening address.
+
+The client will send its request to the proxy plugin process, and the proxy
+will forward the request to the API server, rather than returning the modified
+request to the client to be sent (see alternatives considered).  The downside
+of this approach is that the proxy must duplicate the some of the transport
+functionality of client packages, but allows for more flexibility for proxy
+plugin developers.
+
+#### Extending ExecConfig
```
mikedanese, line 345: Please add a section explaining authentication to the proxy endpoint.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,811 @@
...
+
+### Proxy Proposal
+
+The proxy exec plugin is a long-running proxy process which is executed by the
+client on first use.  The proxy binary location is defined in the `ExecConfig`
+section of a kubeconfig `user`.  When executed, the exec proxy plugin will bind
+and listen on a socket, and will return either a port or a UDS socket address
+on stdout, over which the client would proxy all API server bound requests.
+
+The proxy plugin process will be executed from client-go the first time it is
```
mikedanese, line 331: Please explain how proxy shutdown is expected to work.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,811 @@
...
+Client-go will support both plugins binaries which are themselves the proxy
+process, and additionally plugin binaries which complete, but are responsible
+for somehow starting a proxy process and returning its listening address.
+
+The client will send its request to the proxy plugin process, and the proxy
+will forward the request to the API server, rather than returning the modified
+request to the client to be sent (see alternatives considered).  The downside
+of this approach is that the proxy must duplicate the some of the transport
+functionality of client packages, but allows for more flexibility for proxy
+plugin developers.
```
mikedanese, line 343: What functionality will need to be reimplemented?

mikedanese:
Thanks. The problem/motivation make sense to me and the approach is reasonable. Comments are mostly seeking clarification and additional specification in some areas.

nckturner:
Thanks @mikedanese, getting back to this.  Will address comments today/this week.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Done.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Currently thinking one proxy per process, where the connection to the proxy would be cached based on the input parameters passed over stdin (which includes the cluster information). Added a note to that effect.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: I think that makes sense to me and I updated the KEP to reflect that.  @enj initially suggested the port so he might want to add his thoughts here.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Added a section on proxy shutdown.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: I'm going to resolve this because the reference is outdated, but we can continue the discussion about windows and UDS.

nckturner:
@mikedanese @enj ping for additional review

nckturner:
also @micahhausler @anguslees @jaypipes

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,834 @@
...
+	Env []ExecEnvVar `json:"env"`
+
+	// Preferred input version of the ExecInfo. The returned ExecCredentials MUST use
+	// the same encoding version as the input.
+	APIVersion string `json:"apiVersion,omitempty"`
+
+	// ...
+
+	// +optional
+	ProxyConfig ExecProxyConfig `json:"proxyConfig"`
```
zshihang, line 403: 1. why use an empty struct as bool?
2. instead of adding a boolean value in ExecConfig, we could add a new `ExecConfig` field in `type AuthInfo struct {` (per-user) or `type Cluster struct {` (per-cluster) and call it proxy.

the kubeconfig would be something like this
```
- name: nic@cluster1
  user:
    proxy:
      apiVersion: client.authentication.k8s.io/v1alpha1
      args:
      - --region=us-west-2
      command: /usr/local/bin/sigv4proxy
```

nckturner: 1. It is a placeholder for now, I removed the contents of the struct but left it until I came back to this piece.
2. Good suggestion.  I think per-user makes sense and aligns with a per-client-process proxy.

keps/sig-auth/NNNN-20201209-client-exec-proxy/README.md
nckturner: Resolving because out of date.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,854 @@
...
+### Authentication to Proxy
+
+When communicating to the proxy over TCP/IP, the authentication method to the
+proxy is certificate based. The proxy passes back a client certificate and
+certificate authority that the client can use on connection.
+
+When communicating to the proxy over a unix domain socket, the authentication
+method to the proxy is file permission based.  The GID owner of the client
+process is passed to the proxy over stdin, which the proxy should give
+read/write permissions on the socket to.
```
enj, line 342: I would like to reduce this down to just mTLS over UDS.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,854 @@
...
+
+In the case where the client dies unexpectedly, due to a crash or SIGKILL, the
+proxy should also die.  This could be accomplished with 2 approaches: the proxy
+could quit after a certain amount of idle time, or the proxy could attempt to
+take a file lock that the parent holds and quit if it ever successfully takes
+the lock.  Both are reasonable approaches, so idle quit will be used due to its
+simplicity.  Proxies that are idle for more than 5 minutes will be required to
+quit.
+
+### API
```
enj, line 359: I disagree with effectively the entire structure of this API.  I expect this to be an extension to the existing client-go credential plugin API, specifically as extension to `ExecCredentialStatus`.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,854 @@
...
+
+#### Alpha
+
+- Implementation of either TCP/IP or UDS connections to proxy.
+
+#### Alpha -> Beta Graduation
+
+- Both TCP/IP and UDS connections to proxy supported.
+- Implementation of UDS (or equivalent) on windows.
+- Example proxy implementation.
```
enj, line 537: A full reference implementation demonstrating TLS offload.

enj (changes requested):
While I remain positive on the idea, I disagree with the API proposed here.

Take a look at https://github.com/enj/kubernetes/compare/master...enj:enj/f/exec_proxy to see an approximation of the design I am looking for.

At a high level, I want the existing status response of client-go credential plugins to be enhanced to allow a plugin to return a proxy CA bundle and UDS socket path (I am less stressed by Windows after getting some metrics data around this though I will still want this design communicated over the MLs to get feedback).  This combined with the existing information sent via the spec/status is enough for client-go to build an mTLS connection to proxy and for the proxy to connect to the API server with whatever authentication mechanism it wants.

Since we do not have API consensus, this is too late for the 1.24 enhancement deadline.  However, I would like to get this KEP into an implementable state as soon as possible, even if implementation code cannot be merged until 1.25+.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Moved to existing `ExecCredential` API.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Done.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,854 @@
...
+The connection from the client to the proxy can be cached in the client based
+on the parameters in the ExecProxyRequest: the proxyURL and cluster details.
+
+### Proxy Shutdown
+
+In the case where the client dies unexpectedly, due to a crash or SIGKILL, the
+proxy should also die.  This could be accomplished with 2 approaches: the proxy
+could quit after a certain amount of idle time, or the proxy could attempt to
+take a file lock that the parent holds and quit if it ever successfully takes
+the lock.  Both are reasonable approaches, so idle quit will be used due to its
```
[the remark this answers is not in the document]
nckturner: Updated.

keps/sig-auth/2718-20210511-client-exec-proxy/kep.yaml
```diff
@@ -0,0 +1,54 @@
+title: KEP Template
```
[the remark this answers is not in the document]
nckturner: Fixed.

nckturner:
@enj Updated with your feedback.

keps/prod-readiness/sig-auth/2718.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2718
+stable:
```
rhockenbury, line 2: This is targeting `alpha` not `stable`, correct?

enj: Yes.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+
+The connection from the client to the proxy can be cached in the client based
+on the cluster details, the existing HTTP_PROXY configuration, the exec-proxy
+URL, and the certificates used for mTLS.
+
+### Proxy Shutdown
+
+In the case where the client dies unexpectedly, due to a crash or SIGKILL, the
+proxy should also die.  To handle this, the proxy will attempt to take a file
+lock that the parent holds and quit if it ever successfully takes the lock.
```
enj, line 353: Where is this in the API?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+	Interactive bool `json:"interactive"`
+
+	// New:
+
+	// HttpProxy is the proxy configuration taken from a `Cluster` config or
+	// from http_proxy and https_proxy environment variables. This allows
+	// existing proxy configuration to continue working as is, the exec-proxy
+	// will configure itself to proxy requests to this address.
+	// +optional
+	HttpProxy string `json:"httpProxy,omitempty"`
```
enj, line 386: This information is already included in the `Cluster` object.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+#### Alpha
+
+- Implementation of UDS connections to proxy.
+
+#### Alpha -> Beta Graduation
+
+- Implementation of UDS (or equivalent) on windows.
+- Full reference proxy implementation with TLS offload.
+- Gather feedback from developers, SIG Auth, SIG CLI.
+- Tests are in Testgrid and linked in KEP.
```
enj, line 529: These need to work on macOS, windows and linux.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to maintain previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+For clients moving from no proxy to proxy, a modification to the kubeconfig
+must be made and a proxy binary must be located as specified in the kubeconfig.
+For clients moving from one version of the proxy to another, the proxy binary
+must be upgraded.
```
enj, line 548: This is not correct right?  You may need to update the exec plugin to return a proxy config in its status response, but the kubeconfig could remain unchanged?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+-->
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+* **How can this feature be enabled / disabled in a live cluster?**
+  - [x] Feature gate (also fill in values in `kep.yaml`)
+    - Feature gate name: ClientExecProxy
+    - Components depending on the feature gate: kubelet, kubectl,
```
enj, line 600: How will this work for `kubectl` which does not support feature gates AFAIK?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+request to the client to be sent (see alternatives considered).  The downside
+of this approach is that the proxy must duplicate the some of the transport
+functionality of client packages, like respecting existing proxy configuration,
+but allows for more flexibility for proxy plugin developers.
+
+### Authentication to Proxy
+
+The client will authenticate to its proxy using mTLS.  On exec, the proxy will
+pass back client key and certificate authority data for the client to use for
+authentication when establishing a connection.
```
enj, line 335: Some details on the implementation may be interesting here.  For example, I think we could us Go's reverse proxy code to implement this.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+-->
+
+#### Alpha
+
+- Implementation of UDS connections to proxy.
+
+#### Alpha -> Beta Graduation
+
+- Implementation of UDS (or equivalent) on windows.
+- Full reference proxy implementation with TLS offload.
```
enj, line 527:
```suggestion
- Full reference proxy implementation with TLS offload.  This will be built using code from client-go that provides a harness to offload to a Go `crypto.Signer`.
```

enj (commented):
Still need to review some of the later parts of the KEP doc.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: Also need tests that show that SPDY based APIs like `pods/exec` still work.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+  logs or events for this purpose.
+
+* **What are the SLIs (Service Level Indicators) an operator can use to determine
+the health of the service?**
+  - [ ] Metrics
+    - Metric name:
+    - [Optional] Aggregation method:
+    - Components exposing the metric:
+  - [ ] Other (treat as last resort)
+    - Details:
```
enj, line 674: Number of requests?
Proxy latency?
Failed requests?
Lots of proxy restarts?

This code would be available to webhooks that interface with the API server so this remains relevant.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj, line 529: Also need tests with webhooks since they share this same auth code.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+localhost address from the server process on exec.  The server process listens
+at the unix domain socket address, and the client connects to it.  The client
+attempts to exec the server binary if it is unable to connect using the
+provided address.
+
+This proposal was not preferred to the chosen proposal because, after
+discussions at SIG-Auth, the overall feeling was that defining the request
+response protocol between the client and the signing process would be more
+cumbersome, and be less suited to a generic solution that accommodated any
+proxy related use case.
```
enj, line 844:
```suggestion
proxy related use case.  It is also not possible to implement TLS offload with
this approach.
```

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+    - Adding or manipulating arbitrary headers to the request.
+2. Defining a specification which would provide the ability for the client-side
+proxy server to be executed and communicated with according to configuration
+solely contained in a kubeconfig file.
+3. Defining a specification which is interoperable with an existing client-side
+proxy setup, e.g. a proxy configured with the HTTP_PROXY environment variable.
+4. Defining a specification for a client-side proxy server that is easy to
+secure without taking additional setup steps, by:
+    - Using a unix domain socket for the communication between the client and
+      the proxy, secured with file permissions.
```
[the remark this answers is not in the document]
nckturner: Correct.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: We can also have platform specific implementation of this and use locks in the generic implementation. On linux, you can use PR_SET_PDEATHSIG:

https://cs.opensource.google/go/go/+/refs/tags/go1.19.1:src/syscall/exec_linux.go;l=48-52;drc=126c22a09824a7b52c019ed9a1d198b4e7781676

Which simplifies the implementation of exec plugins that only need to work on linux. On windows, you can use job objects.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,837 @@
...
+	ProxyConfig *ProxyConfig `json:"proxyConfig,omitempty"`
+}
+
+
+// ProxyConfig configures the client to proxy requests via the returned address.
+type ProxyConfig struct {
+	// URL is the location of the exec proxy listener.  Supports unix domain
+	// socket locations in the format unix:///path/to/socket.
+	URL string `json: "url"`
+	// CertificateAuthorityData is required if an exec-proxy is used.
```
mikedanese, line 409: Is this always necessary? Why?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,837 @@
...
+	// ProxyConfig is the config returned by the exec-proxy
+	// +optional
+	ProxyConfig *ProxyConfig `json:"proxyConfig,omitempty"`
+}
+
+
+// ProxyConfig configures the client to proxy requests via the returned address.
+type ProxyConfig struct {
+	// URL is the location of the exec proxy listener.  Supports unix domain
+	// socket locations in the format unix:///path/to/socket.
```
mikedanese, line 407: Do we allow unnamed sockets? Do we allow anonymous sockets?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,837 @@
...
+Integration tests will cover:
+
+* Authentication between client and proxy.
+* Connection from client to apiserver through proxy.
+* Proxy use with kubectl (possible test location: [command-line integration
+  test
+  suite](https://github.com/kubernetes/kubernetes/tree/ec560b9737537be8c688776461bc700e8ddedb9d/test/cmd)).
+* UDS connections from the client to the proxy will be tested.
+
+#### E2E
```
mikedanese, line 447: Given that this is client <-> API, an integration test should be sufficient. I would not expect that we would need to cover this with an e2e unless/until we want to add this to conformance.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: What did we decide in SIG Auth? Prefix fields with alpha while this functionality is in the alpha stage?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: Lots of proxy restarts seems reasonable. I'm not convinced there's much value in the others.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: The mechanism has to reflect user choice / opt-in.  The fields being alpha is meaningless.  My preference would be to check an interface which is implemented by feature gates on everything but kubectl.  kubectl would use an environmental variable based implementation.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: If we need lock files, they have to be present in the API.  Also, we will provide a reference implementation so the bulk of implementers should not have to care.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: Because the connection between the proxy and client-go is always over mutual TLS.  We do not rely on the UDS to provide any security guarantees especially since they behave differently on windows.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,837 @@
...
+for these clients setting up an additional proxy configuration might not be
+possible, or at least not straightforward.  Secondly, this approach requires a
+long running external proxy process which introduces operational complexity.
+Thirdly, communicating with a proxy for request signing requires a secure
+communication channel between the client and the proxy, which means in a case
+where multiple users share a client computer (for example, a shared bastion),
+configuring and securing a local proxy to prevent signing requests as a
+different user becomes more complicated.
+
+### Risks and Mitigations
```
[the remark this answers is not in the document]
enj: IMO we do not need to do anything because we do not rely on the UDS to provide any security properties.  https://github.com/kubernetes/enhancements/pull/2693#discussion_r986893116

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: We probably should stick with simple file based sockets.  I am unsure if the other ones even work on windows.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: IIRC @nckturner wanted to use this feature in EKS for kubelet authn to the API server.  That may be an interesting test scenario to exercise in CI since it is a distinctly different type of client than say `kubectl`.

keps/prod-readiness/sig-auth/2718.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2718
+alpha:
+  approver: "@deads2k"
```
johnbelamaric, line 3: Please re-assign to me

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
johnbelamaric: Yes, env var is usually what we do for kubectl to enable alpha flags/features, as I recall. The primary point of gating CLI features is so that the user of those features is sure to be aware that the feature is alpha, and will not accidentally use it thinking it is GA.

Presumably for controllers consuming these libraries, they will have a way to know this is alpha?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
johnbelamaric: As a client-side proxy component, how would you monitor this? Metrics for that specific proxy process would seem to be out of scope of this KEP.

johnbelamaric (commented):
A few comments but looks pretty good from a PRR perspective.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
zshihang: there are scenarios that mtls is not necessary. should we leave the decision to the users?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Is there a difference between unnamed and anonymous sockets?  Between unnamed, abstract and named/pathnamed, it looks like all are actually supported by the windows implementation of `AF_UNIX` [1].  Abstract might actually make sense for us because it avoids the use of the filesystem (but they still have a name), and if we are not relying on the filesystem for security then it simplifies the implementation by avoiding path collisions and avoiding the need to clean up the socket.  Unnamed doesn't make sense to me because it would require a different API and the socket would have to be passed over `fork()`.  I think that would be a more difficult API for plugin implementers.

[1]: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

mikedanese: Sorry, meant to say abstract, not anonymous.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: If we leave the decision to the users, then we need to allow them to use the socket permissions and the directory permissions in which the socket resides to provide security guarantees.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: > IIRC @nckturner wanted to use this feature in EKS for kubelet authn to the API server. That may be an interesting test scenario to exercise in CI since it is a distinctly different type of client than say kubectl.

That use case is correct.  We'd definitely want a scenario like this one in an E2E test.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: > Presumably for controllers consuming these libraries, they will have a way to know this is alpha?
I think the feature gate documentation should be sufficient. [1]

[1]: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: > socket permissions

We use socket permissions all over the place (docker socket, KMS plugin, device plugin, CSI driver)... without issue. It's a well understood permission model. While not particularly critical for kubectl, if efficiency ever becomes a concern (e.g. usage expands to cluster components like kubelet) then we will want to revisit the mTLS requirement.

enj: Using socket permissions on server components that can be tightly controlled is not the same as using it on end user desktop that may have more than one user and questionable security controls (and an unbounded set of custom software).  We can certainly measure the overhead of mTLS but I suspect it is negligible on any modern hardware.  Unlike the API server, this proxy can do connection level mTLS checks instead of per request mTLS checks, which will further reduce any issues.  And again, the reference implementation will do this all for you - you only have to specify the business logic that is custom to your proxy, not this boilerplate.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
enj: https://github.com/microsoft/WSL/issues/4240

enj: https://github.com/grpc/grpc/issues/22285#issuecomment-1169211526
https://github.com/microsoft/WSL/issues/5272#issuecomment-1236620129

nckturner: Good find @enj

enj: > [microsoft/WSL#4240 (comment)](https://github.com/microsoft/WSL/issues/4240#issuecomment-620805115)

I have confirmed with `@sunilmut` that this is still true.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: If they already using an exec plugin, then yes.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Maybe we can start with 1) number of failed requests and 2) number of proxy restarts.  @johnbelamaric there are other metrics in client-go so I think we can follow the same pattern. [1]  They would be exposed via a metrics endpoint in kube-controller-manager, for example, but not in kubectl.  

[1]: https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/leaderelection/metrics.go

johnbelamaric: +1 great.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -683,6 +683,9 @@ of this feature?**
   Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
   implementation difficulties, etc.).
 
+  - Number of proxy restarts
```
johnbelamaric, line 686: this section is for metrics that were *not* added, is that what you intended? Also my previous question applies to this metric - isn't this proxy an external component, how would we know it has restarted (maybe I missed something?)

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -683,6 +683,9 @@ of this feature?**
...
+  - Number of failed requests
```
johnbelamaric, line 687: If you actually mean to add these metrics, please use the names so people know what to look for in Prometheus

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
johnbelamaric: sorry, actually what I meant was "number of proxy restarts" - we are not deploying the actual proxy in this feature, right? We are just a client to it, via the domain socket? So how would we know if it restarted?

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+functionality of client packages, like respecting existing proxy configuration,
+but allows for more flexibility for proxy plugin developers.
+
+### Authentication to Proxy
+
+The client will authenticate to its proxy using mTLS.  On exec, the proxy will
+pass back client key and certificate authority data for the client to use for
+authentication when establishing a connection.
+
+An alternative method for authentication to the proxy would be file permission
```
mikedanese, line 337: Add unnamed pipes as a third alternative and don't pick one. We will decide during the implementation.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -0,0 +1,845 @@
...
+
+	// ProxyConfig is the config returned by the exec-proxy
+	// +optional
+	ProxyConfig *ProxyConfig `json:"proxyConfig,omitempty"`
+}
+
+
+// ProxyConfig configures the client to proxy requests via the returned address.
+type ProxyConfig struct {
+	// URL is the location of the exec proxy listener.  Supports unix domain
```
mikedanese, line 406: Remove the comment about what is supported.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: Let's remove this comment for now and work it out over the implementation.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
mikedanese: @johnbelamaric in the existing credential plugin architecture, the binary actually invokes the plugin process: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins

Same thing here.

johnbelamaric:
Ok, well, I am just waiting for SIG approval - PRR is fine. Deadline is close...

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: I guess it should be framed as "number of times we have to exec the proxy because of a connection failure"

nckturner: Looks like we already have a metric representing exec plugin calls: https://github.com/kubernetes/client-go/blob/d3b97581d2e22eef0f02a6667e19dfc193df7a3e/plugin/pkg/client/auth/exec/metrics.go#L93-L111

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
```diff
@@ -663,9 +667,8 @@ _This section must be completed when targeting beta graduation to a release._
 * **What are the SLIs (Service Level Indicators) an operator can use to determine
 the health of the service?**
   - [ ] Metrics
-    - Metric name:
-    - [Optional] Aggregation method:
-    - Components exposing the metric:
+    - Metric name: `rest_client_exec_plugin_with_proxy_call_total`
```
johnbelamaric, line 670: Is that just a count? can we differentiate errors? (non-blocking question)

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Added `LockFilePath` to the API.

keps/sig-auth/2718-20210511-client-exec-proxy/README.md
nckturner: Yes, we will be able to differentiate errors, the implementation follow the example of: https://github.com/kubernetes/client-go/blob/d3b97581d2e22eef0f02a6667e19dfc193df7a3e/plugin/pkg/client/auth/exec/metrics.go

enj:
/lgtm
/approve

This is close enough for us to figure out the final details during the implementation.

PR #2703, opened 2021-05-10
KEP-2702: Graduate HPA v2beta2 API to GA

supriya-premkumar:
Changes:
	1. Adds respective README.md and kep.yaml


keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,422 @@
...
+
+For each of these, fill in the following—thinking about running existing user workloads
+and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+-->
+
+### Scalability
```
josephburnett, line 291: N/A

josephburnett (changes requested):

johnbelamaric:
You can assign me as PRR approver. Thanks.

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,422 @@
...
+
+
+### Implementation Details
+
+
+### Risks and Mitigations
+
+* v1-v2 [conversion loss](https://github.com/kubernetes/kubernetes/issues/80481) of multiple CPU targets
+* v2beta1 has the significant amount of boilerplate and overhead in maintaining conversion routines for multiple public APIs.
+* Avoid existing confusion with `SelectPolicy` [enumeration](https://github.com/kubernetes/api/blob/2c3c141c931c0ab1ce1396c3152c72852b3d37ee/autoscaling/v2beta2/types.go#L149-L156)
```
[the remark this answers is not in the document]
supriya-premkumar: Ack. Sounds good

josephburnett:
@jrsapi 
> Current KEP/README with test and monitoring plans

This KEP already has a testing plan.  And the monitoring plan template says `This section must be completed when targeting beta to a release.`.  This API is already in beta.  And we are not graduating any features, just the API.  Do we still need a monitoring plan?

josephburnett:
@johnbelamaric 
Thanks for being the PRR approver.  What do we need to do next?  Will you review this as-is, or do I need to fill out something else?

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,443 @@
...
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This document outlines required steps to graduate autoscaling v2beta2 API to GA.
+
+## Motivation
+
+
```
sftim, line 34: ?

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/kep.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - "@supriya-premkumar"
+  - "@josephburnett"
+owning-sig: sig-autoscaling
+reviewers:
+  - "@josephburnett"
+  - "@gjtempleton"
+approvers:
+  - "@josephburnett"
+  - "@gjtempleton"
+prr-approvers:
```
johnbelamaric, line 13: you don't need this here anymore (sorry bad template). just having it in the other file is enough

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/kep.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - "@johnbelamaric"
+creation-date: 2021-05-10
+last-updated: 2021-05-10
+status: implementable
+
+latest-milestone: "1.22"
```
johnbelamaric, line 19: v1.22, I don't think the tooling figured that out.

keps/prod-readiness/sig-autoscaling/2702.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2702
+beta:
```
johnbelamaric, line 2: stable

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,422 @@
...
+you need any help or guidance.
+-->
+
+### Feature Enablement and Rollback
+
+<!--
+This section must be completed when targeting alpha to a release.
+-->
+
+###### How can this feature be enabled / disabled in a live cluster?
```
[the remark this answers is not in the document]
johnbelamaric: yeah, this is going to stable so it's no longer applicable. if there is an existing feature gate though, it would be good to document it so that folks seeing this documentation when on older versions will know what to do.

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,422 @@
...
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### How can an operator determine if the feature is in use by workloads?
```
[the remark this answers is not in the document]
johnbelamaric: The question isn't whether it's supported, it's whether it's in use. So in this case the answer would be that some v2 resources exist. Not sure if we automatically have any metrics around presence or count of specific resource types.

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
johnbelamaric: Why is the Scalability section N/A?

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
```diff
@@ -0,0 +1,443 @@
...
+
+<!--
+Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+checking if there are objects with field X set) may be a last resort. Avoid
+logs or events for this purpose.
+-->
+
+###### How can someone using this feature know that it is working for their instance?
```
johnbelamaric, line 228: Please provide answers here and in the other monitoring questions. How do we monitor this functionality?

johnbelamaric:
> @jrsapi
> 
> > Current KEP/README with test and monitoring plans
> 
> This KEP already has a testing plan. And the monitoring plan template says `This section must be completed when targeting beta to a release.`. This API is already in beta. And we are not graduating any features, just the API. Do we still need a monitoring plan?

Well all the more so if you are targeting GA. The template doesn't take into account stuff that was beta before the template was written. But when going to GA you should definitely be able to tell us how to monitor your components and how to know if they are working. Also ideally you may have some info on troubleshooting and/or other failure modes you may have seen in the field.

johnbelamaric:
I understand you are just graduating the API, but isn't that the same as saying the features are GA quality now? I am not sure I see a difference?

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This document outlines required steps to graduate autoscaling v2beta2 API to GA.
+
+## Motivation
```
josephburnett, line 32: "The HPA v2 series APIs were first introduced in November, 2016 (5 years ago).  The primary feature of the v2 series is adding support for multiple and custom metrics.  The structure was improved slightly in the v2beta2 API which became available in May 2018 and has remained largely unchanged since then.  The v2beta2 API has been used extensively and informally treated as stable.  The motivation for this KEP is to push it over the line to make it formally so.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+run `kubectl get hpa.v2.autoscaling -oyaml` to see your objects in the v2 format.
+
+###### Does enabling the feature change any default behavior?
+
+<!--
+Any change of default behavior may be surprising to users or break existing
+automations, so be extremely careful here.
+-->
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
```
josephburnett, line 138:
```
The feature can be enabled by adding `autoscaling/v2` to the `--runtime-config` flag: 
https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/staging/src/k8s.io/apiserver/pkg/server/options/api_enablement.go#L45

Adding `api/all` will also include `autoscaling/v2`.

The feature can be disabled by removing the `--runtime-config` entry.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### How can an operator determine if the feature is in use by workloads?
```
josephburnett, line 203: > The question isn't whether it's supported, it's whether it's in use. So in this case the answer would be that some v2 resources exist. Not sure if we automatically have any metrics around presence or count of specific resource types.

@johnbelamaric All HPA objects are stored in v1 format on disk.  They are up converted the requested version.  And down converted upon update.

So there are no specifically versioned objects the user needs to worry about.  Am I understanding your question correctly?

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+For each of these, fill in the following—thinking about running existing user workloads
+and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+-->
+
+### Scalability
```
josephburnett, line 290: > Why is the Scalability section N/A?

@johnbelamaric I was thinking along the lines of this graduation will not change the scalability characteristics of the v2beta2 API.  But let me try and address the scalability changes from v1 to v2:

```
The HPA v2 APIs allow users to configure multiple metrics, each with a separate target.  A recommendation is
calculated for each metric and the largest recommendation is used.  The more metrics are added to a given
HPA the longer it will take to reconcile.  The HPA is single-threaded processing recommendations one-at-a-time.
When default reconciliation period is 15 seconds.  If there is too much work to do reconciliation will slow down 
and happen less frequently than every 15 seconds.  This will cause autoscaling to be less responsive at high scale.

Previously v1 scaled along two dimensions, number of HPA and number of pods selected by each HPA (linearly).  
Now it will scale with the number of metrics defined in HPAs and the number of pods selected each metric (again 
linearly).

Additionally v2 adds a behavior structure which allows the user configure that rate and delay of scaling and down.
  Enforcing these constraints require storing previous recommendations and scaling events in memory.  The 
longer the configured interval the more memory is used.  The maximum window allows is 60 minutes 
([code](https://pkg.go.dev/k8s.io/api/autoscaling/v2beta2#HPAScalingRules)) so 240 recommendations / events per 
configured metric.  Each recommendation is an `int32` and `time.Time`.  Each scaling event is an `int32`, 
a `time.Time` and a `bool` ([code](https://pkg.go.dev/k8s.io/api/autoscaling/v2beta2#HPAScalingRules)) so 
the memory footprint is relatively small.  It will scale linearly with the number of metrics defined and the size 
of the HPA's configured window.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+are missing a bunch of machinery and tooling and can't do that now.
+-->
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+<!--
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
```
josephburnett, line 197:
```
The HPA requires the `metrics.k8s.io` APIs to be available in the cluster to operate.  This API 
is served by the Metrics Server.  An operator can verify the Metrics Server is available to provide 
resource metrics to the HPA by running the command `kubectl get apiservices` and looking for 
the status of `v1beta1.metrics.k8s.io` (version subject to change).  Operators should take care 
to make sure Metrics Server is up and running to maintain resource autoscaling.

The v2 HPA requires the `custom.metrics.k8s.io` and `external.metrics.k8s.io` APIs as well to 
retrieve custom and external metrics.  There is no default implementation of these APIs and 
cluster operators must install an "adapter" for their metrics backend (e.g. [Prometheus](https://github.com/kubernetes-sigs/prometheus-adapter)).  An operator can verify the adapter
 is working properly by running the same kubectl for apiservices and looking for the `v1beta1.custom.metrics.k8s.io` and `v1beta1.external.metrics.k8s.io` APIs (usually served 
by the same adapter).  Care should be taken to ensure the adapter and specific metrics 
backend is available to maintain custom metric autoscaling.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+For instance, if this is a pod-related feature, it should be possible to determine if the feature is functioning properly
+for each individual pod.
+Pick one more of these and delete the rest.
+Please describe all items visible to end users below with sufficient detail so that they can verify correct enablement
+and operation of this feature.
+Recall that end users cannot usually observe component logs or access metrics.
+-->
+
+- [ ] Events
+  - Event Reason:
```
josephburnett, line 223:
```
The event type `Normal`, reason `SuccessfulRescale`, note `New size: N; reason: FOO` 
indicates autoscaling is operating normally.

Abnormal events type `Warning` include reasons such as `FailedRescale` and 
`FailedComputeMetricsReplicas` and will include details about the error in the note.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+Pick one more of these and delete the rest.
+Please describe all items visible to end users below with sufficient detail so that they can verify correct enablement
+and operation of this feature.
+Recall that end users cannot usually observe component logs or access metrics.
+-->
+
+- [ ] Events
+  - Event Reason: 
+- [ ] API .status
+  - Condition name:
```
josephburnett, line 225:
```
There are three condition types which indicate the operating status of the HPA.  They are 
`ScalingEnabled`, `AbleToScale` and `ScalingLimited` (see type 
[comments](https://pkg.go.dev/k8s.io/api/autoscaling/v2beta2#HorizontalPodAutoscalerConditionType
))

Under normal operating circumstances `ScalingEnabled` and `AbleToScale` should be status `true`, 
indicating the HPA is successfully reconciling the scale.  `ScalingLimited` indicates user configuration
 is limiting the "ideal" scale with a minimum, maximum, rate or delay.  Which limit is the cause will be 
indicated in the message.  It's normal for this to be `true` or `false` periodically.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+Recall that end users cannot usually observe component logs or access metrics.
+-->
+
+- [ ] Events
+  - Event Reason: 
+- [ ] API .status
+  - Condition name: 
+  - Other field: 
+- [ ] Other (treat as last resort)
+  - Details:
```
josephburnett, line 228:
```
The HPA status includes the current observed metric values, one for each given target.  Using these
values an operator can verify the HPA is maintaining the desired target for the dominant metric.  

The operator can also see the number of pods the HPA observed under `status.currentReplicas` and 
the most recent recommendation under `status.desiredReplicas`.

The latest observed generation is echoed back in status so an operator can verify the HPA is keeping
up-to-date with configuration changes.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,426 @@
...
+* The KEP [test plan](https://github.com/kubernetes/enhancements/blob/15d330a932e0aae220ff719c391f7d815492088f/keps/sig-autoscaling/20190307-configurable-scale-velocity-for-hpa.md#test-plan) includes unit tests
+* Add e2e tests for container resource targets
+* Add conformance tests
+
+### Graduation Criteria
+
+The following code changes must be made for graduating to GA
+
+* Move API objects to `v2` and support conversion internally
+
```
josephburnett, line 86:
```
* Add behavior and container target E2E tests.
```

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
johnbelamaric: Ok, I suppose that makes sense. In that case, just the existence of HPA objects. We're sort of backfilling docs here for a long standing feature. In general as we see from the previous questions, I would prefer we answer the PRR from the point of view of "what will help cluster operators" with respect to this feature rather than from the point of view of "does specifically switching from v2beta2 to v2 have a specific impact with this question". That is, the questions are geared around the idea that going to GA means it's stable and folks will want to treat it as such, so "what does that mean for the operator?". How do they support it? How do they know it's working and if it's not working how do they fix the problem if it's not?

Given that this feature has been around a long time, you probably have all that knowledge, just not in one place. So taking it to GA gives us the opportunity to put it all in one place. Alternatively if you do have it in one place, a link there could suffice.

keps/sig-autoscaling/2702-graduate-v2beta2-autoscaling-api-to-GA/README.md
supriya-premkumar: Ack. Updated it. Thank you.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,511 @@
...
+- [ x ] Other (treat as last resort)
+  - Details:
+  The HPA status includes the current observed metric values, one for each given target. Using these
+  values an operator can verify the HPA is maintaining the desired target for the dominant metric.
+  The operator can also see the number of pods the HPA observed under `status.currentReplicas` and the most
+  recent recommendation under `status.desiredReplicas`.
+  The latest observed generation is echoed back in status so an operator can verify the HPA is keeping up-to-date
+  with configuration changes.
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
```
johnbelamaric, line 301: Do you have SLOs/SLIs for HPA?

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,511 @@
...
+implementation difficulties, etc.).
+-->
+
+### Dependencies
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### Does this feature depend on any specific services running in the cluster?
```
johnbelamaric, line 344: What happens if no metrics server is installed? What happens if it fails?

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,511 @@
...
+  - estimated throughput
+  - originating component(s) (e.g. Kubelet, Feature-X-controller)
+Focusing mostly on:
+  - components listing and/or watching resources they didn't before
+  - API calls that may be triggered by changes of some Kubernetes resources
+    (e.g. update of object X triggers new updates of object Y)
+  - periodic API calls to reconcile state (e.g. periodic fetching state,
+    heartbeats, leader election, etc.)
+-->
+No
```
johnbelamaric, line 403: Maybe say, "No, not in comparison to using the existing v2beta2 APIs, but of course using HPA results in new API calls as described above.".

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
josephburnett: We actually have quite good documentation IMO which has been updated as the newer v2 fields have been added: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/  This includes quite a bit of background, algorithm details, and some good [operator notes](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-metrics-apis).

johnbelamaric (commented):
Thanks this is looking a lot better. Still a bit more to do but this will be really helpful for operators.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
johnbelamaric: Awesome. Please include links to that.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
josephburnett: No, we don't.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
josephburnett: I think I addressed that under "Monitoring Requirements" as what to keep working in the cluster.  Maybe that should be moved here instead.

TL;DR: without Metrics Server autoscaling on resource metrics will not work.  Without the a custom metrics adapter and the backing metric store running, custom and external metrics will not work.  If there are multiple metrics defined and one is not available, scale up will continue but scale down will not (for safety).

supriya-premkumar:
Thank you so much @johnbelamaric for the thorough review. Are we missing anything else?

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
johnbelamaric: So there are no metrics at all from the autoscaler? Probably something you want to add though I won't block this KEP for it.

johnbelamaric:
OK, I am happy with PRR. Please get SIG approval and then I will approve.

supriya-premkumar:
@gjtempleton Could you please LGTM the PR if everything looks good?
We might be able to make it before the enhancement freeze, provided that the bot merges the PR in time. 🙂

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
```diff
@@ -0,0 +1,579 @@
...
+<!--
+Pick one more of these and delete the rest.
+-->
+
+- [ ] Metrics
+  - Metric name:
+  - [Optional] Aggregation method:
+  - Components exposing the metric:
+- [ ] Other (treat as last resort)
+  - Details:
```
gjtempleton, line 380: I think we should remove this as we currently have no SLIs on the operation of the HPA, and don't intend to add them as part of this graduation currently.

gjtempleton (commented):
One minor point, other than that it looks good to me.

keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md
supriya-premkumar: Ack, updated it. Thank you.

jrsapi:
Greetings @josephburnett @gjtempleton @supriya-premkumar! 
After reviewing the merged [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/README.md#test-plan) and approved [PRR](https://github.com/kubernetes/enhancements/blob/master/keps/sig-autoscaling/2702-graduate-hpa-api-to-GA/kep.yaml). This enhancement is now being tracked for the 1.22 milestone. 

Thanks!

jrsapi:
Greetings @josephburnett @gjtempleton,
Enhancement shadow checking with a reminder that we are 2 weeks away from code freeze (July 8, 2021). Can you link which k/k PR/s are needed to implement this enhancement for the 1.22 milestone? This KEP is marked at risk without this information. 

Thanks!

jrsapi:
Greetings @josephburnett @gjtempleton,
 I've received an update from @supriya-premkumar that this KEP will not be part of the 1.22 milestone. I'm looking to confirm if that is the case. 
Code freeze is this Thursday, July 8th and this KEP is at risk. Also this Friday, July 9th is the Docs placeholder PR deadline. Here are [instructions on opening](https://kubernetes.io/docs/contribute/new-content/new-features/#open-a-placeholder-pr) a PR in dev-1.22 branch.

Thanks!

gjtempleton:
Hi @jrsapi, that's correct, we're not going to be able to address everything we need to in time for meeting the 1.22 deadlines and will be moving it out to 1.23.

PR #2713, opened 2021-05-11
Pod priority based graceful node shutdown

mrunalp:
@derekwaynecarr @bobbypage 

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

### DIFF
--- a/keps/prod-readiness/sig-node/2712.yaml
+++ b/keps/prod-readiness/sig-node/2712.yaml
@@ -1,0 +1,3 @@
+kep-number: 2712
+alpha:
+  approver: "@deads2k"
--- a/keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
+++ b/keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
@@ -1,0 +1,553 @@
+# KEP-2712: Pod Priority Based Graceful Node Shutdown
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Implementation](#implementation)
+  - [Migration from the Node graceful shutdown feature](#migration-from-the-node-graceful-shutdown-feature)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha Graduation](#alpha-graduation)
+    - [Alpha -&gt; Beta Graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA Graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+
+## Release Signoff Checklist
+
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [x] (R) Graduation criteria is in place
+- [x] (R) Production readiness review completed
+- [ ] Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Kubelet graceful shutdown should take the pod priority values into account to
+determine the order in which the pods are stopped.
+
+## Motivation
+
+The [node graceful shutdown KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2000-graceful-node-shutdown) added support to the kubelet to detect
+that a node is shutting down and making sure that the pods are gracefully
+stopped before allowing the shutdown to proceed. 
+
+The feature added flags to specify the total time for shutdown and the
+time to reserve for shutting down critical pods.
+
+However, there is a need to allow more fine grained control over the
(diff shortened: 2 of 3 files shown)


keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+| 0                      |60 seconds     |
+
+
+In the above case, the pods with custom-class-b will go into the same bucket
+as custom-class-c for shutdown.
+
+If there are no pods in a particular range, then the kubelet will skip
+over the range entirely.
+
+If this feature is enabled and no configuration is provided, then no ordering
```
derekwaynecarr, line 159: can you specify what min systemd version is required for this?

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+
+## Design Details
+
+The configuration will be controlled by a new Kubelet Config setting,
+`kubeletConfig.PodPriorityShutdownGracePeriods`:
+
+```
+type PodPriorityShutdownGracePeriod struct {
+	Priority int32
+	ShutdownGracePeriod metav1.Duration
```
derekwaynecarr, line 190: i think this should include the unit in the field name (seconds)

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+    - Will enabling / disabling the feature require downtime or reprovisioning
+      of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
+      - yes (will require restart of kubelet)
+
+* **Does enabling the feature change any default behavior?**
+  Any change of default behavior may be surprising to users or break existing
+  automations, so be extremely careful here.
+
+    * The main behavior change is that during a node shutdown, pods running on
+      the node will be terminated gracefully.
```
deads2k, line 306: Is there an upper bound on how long the kubelet will wait?  Do unprivileged pod authors get to choose this value and potentially prevent nodes from shutting down for a very long time?

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+
+* **Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?**
+  Also set `disable-supported` to `true` or `false` in `kep.yaml`.
+  Describe the consequences on existing workloads (e.g., if this is a runtime
+  feature, can it break the existing applications?).
+
+    * Yes, the feature can be disabled by either disabling the feature gate.
```
deads2k, line 314: And this requires restarting the kubelet?  

Do the pods have to be evicted?  Based on quick read, it doesn't seem like we'll have to.

keps/prod-readiness/sig-node/2712.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2712
+beta:
```
deads2k, line 2: When did this go through alpha?

mrunalp: copy paste error, fixing.

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
mrunalp: You mean on a node where it is already enabled? One can restart the kubelet with the feature gate turned off without evicting the pods.

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
mrunalp: Yes, it will be bounded. Adding a sentence.

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+to 3 minutes, then it could be migrated to:
+
+|Pod priority class value|Shutdown period|
+|------------------------|---------------|
+| 1000000001             |180 seconds    |
+| 0                      |120 seconds    |
+
+Kubelet will be modified to only work with the config proposed in this KEP or the
+Node shutdown KEP. If both are specified, then it will be treated as a configuration
+error.
```
bobbypage, line 176: Maybe add,

If neither are specified, then Graceful Node Shutdown feature is disabled.

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,544 @@
...
+Instead of looking at the pod priority class names, we can instead
+look at the pod priority class values to allow more control over pod
+shutdown order.
+
+### Goals
+
+*   Make the kubelet use shutdown configuration based on pod priority values for
+    graceful shutdown.
+
+### Non-Goals
```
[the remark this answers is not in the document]
mrunalp: done

deads2k:
PRR lgtm.

/approve

The PRR.  The rest of KEP is up to sig-node.

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
mrunalp: done

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,553 @@
...
+| 100000                 |300 seconds    |
+| 1000                   |120 seconds    |
+| 0                      |60 seconds     |
+
+
+In the above case, the pods with custom-class-b will go into the same bucket
+as custom-class-c for shutdown.
+
+If there are no pods in a particular range, then the kubelet will skip
+over the range entirely.
```
sftim, line 158: What does “skip over the range entirely” mean? 0-second grace period perhaps?

mrunalp: If we allocate 2 minutes for pods in a 1000 - 10000 range and there are no actual pods then we don't artificially wait for 2 minutes but move on to the next range.

sftim, line 158:
```suggestion
If there are no pods in a particular range, then the kubelet does not wait
for pods in that priority range. Instead, the kubelet immediately skips to the
next priority class value range.
```

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
```diff
@@ -0,0 +1,553 @@
...
+|------------------------|---------------|
+| 100000                 |300 seconds    |
+| 10000                  |180 seconds    |
+| 1000                   |120 seconds    |
+| 0                      |60 seconds     |
+
+The above table implies that any pod with priority value >= 100000 will get
+300 seconds to stop, any pod with value >= 10000 and < 100000 will get 180
+seconds to stop, any pod with value >= 1000 and < 10000 will get 120 seconds to stop. 
+Finally, all other pods will get 60 seconds to stop.
```
sftim, line 139: I would illustrate that the shutdown periods don't need to be ordered (assuming that's true)
```suggestion
| 100000                 |10 seconds     |
| 10000                  |180 seconds    |
| 1000                   |120 seconds    |
| 0                      |60 seconds     |

The above table implies that any pod with priority value >= 100000 will get
just 10 seconds to stop, any pod with value >= 10000 and < 100000 will get 180
seconds to stop, any pod with value >= 1000 and < 10000 will get 120 seconds to stop. 
Finally, all other pods will get 60 seconds to stop.
```

sftim: (this also matches how I'd set up a node - once there are no workload Pods, actually I don't mind if the node shutdown is less than graceful).

mrunalp: Yes, they don't have to be ordered.

mrunalp: Thanks!

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
mrunalp: Accepted, thanks!

keps/sig-node/2712-pod-priority-based-graceful-node-shutdown/README.md
mrunalp: done

derekwaynecarr:
thanks @mrunalp for the updates.

thanks for all those that helped provide review.

/approve
/lgtm

PR #2723, opened 2021-05-12
Add new sections in kep

gnufied:
Fill up implementation history, drawback and alternatives etc. 

xref #2317


gnufied:
cc @reylejano @xing-yang

PR #2724, opened 2021-05-12
Update KEP 1602 milestone metadata

serathius:
/cc @reylejano @ehashman


ehashman:
Should this be moved to be owned by WG Logging?

ehashman (commented):
/approve
/lgtm
/hold

Don't want this to miss the deadline, just am wondering if we should be transferring ownership of this so Instrumentation only needs to ack, not approve.

serathius:
Will look into splitting KEP ownership, let's merge this for 1.22 enhancement freeze.
/unhold

PR #2733, opened 2021-05-13
KEP-2086: update beta milestone for v1.22

andrewsykim:
Signed-off-by: Andrew Sy Kim <kim.andrewsy@gmail.com>

### DIFF
--- a/keps/sig-network/2086-service-internal-traffic-policy/README.md
+++ b/keps/sig-network/2086-service-internal-traffic-policy/README.md
@@ -56,7 +56,7 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 ## Summary
 
-Add a new field `spec.trafficPolicy` to Service that allows node-local and topology-aware routing for Service traffic.
+Add a new field `spec.internalTrafficPolicy` to Service that allows node-local and topology-aware routing for Service traffic.
 
 ## Motivation
 
@@ -76,17 +76,15 @@ for internal Service traffic.
 
 ## Proposal
 
-Introduce a new field in Service `spec.trafficPolicy`. The field will have 4 codified values:
+Introduce a new field in Service `spec.internalTrafficPolicy`. The field will have 2 codified values:
 1. Cluster (default): route to all cluster-wide endpoints (or use topology aware subsetting if enabled).
-2. Topology: route to endpoints using topology-aware routing. See Topology Aware Hints KEP for more details.
-2. PreferLocal: route to node-local endpoints if it exists, otherwise fallback to behavior from Cluster.
-3. Local: only route to node-local endpoints, drop otherwise.
+2. Local: only route to node-local endpoints, drop otherwise.
 
-A feature gate `ServiceTrafficPolicy` will also be introduced for the alpha stage of this feature.
-The `trafficPolicy` field cannot be set on Service during the alpha stage unless the feature gate is enabled.
+A feature gate `ServiceInternalTrafficPolicy` will also be introduced this feature.
+The `internalTrafficPolicy` field cannot be set on Service during the alpha stage unless the feature gate is enabled.
 During the Beta stage, the feature gate will be on by default.
 
-The `trafficPolicy` field will not apply for headless Services or Services of type `ExternalName`.
+The `internalTrafficPolicy` field will not apply for headless Services or Services of type `ExternalName`.
 
 ### User Stories (Optional)
 
@@ -103,7 +101,6 @@ Traffic should never bounce to a daemon on another node.
 ### Risks and Mitigations
 
 * When the `Local` policy is set, it is the user's responsibility to ensure node-local endpoints are ready, otherwise traffic will be dropped.
-* Using the `Local` or `PreferLocal` policy may result in imbalanced traffic for pods in a Service. It is the user's responsibility to handle this.
 
 ## Design Details
 
@@ -113,8 +110,6 @@ type ServiceInternalTrafficPolicyType string
 
 const (
 	ServiceTrafficPolicyTypeCluster     ServiceTrafficPolicyType = "Cluster"
-	ServiceTrafficPolicyTypeTopology    ServiceTrafficPolicyType = "Topology"
-	ServiceTrafficPolicyTypePreferLocal ServiceTrafficPolicyType = "PreferLocal"
 	ServiceTrafficPolicyTypeLocal       ServiceTrafficPolicyType = "Local"
 )
 
@@ -123,51 +118,54 @@ type ServiceSpec struct {
 	...
 	...
 
-	// trafficPolicy denotes if the traffic for a Service should route
-	// to cluster-wide endpoints or node-local endpoints. "Cluster" routes traffic
-	// to a Service to all cluster-wide endpoints. "Topology" routes traffic based on
-	// topology hints. "PreferLocal" will route internal traffic to node-local endpoints
-	// if one exists, otherwise it will fallback to the same behavior as "Cluster".
-	// "Local" routes traffic to node-local endpoints only, traffic is dropped
-	// if no node-local endpoints are ready. When externalTrafficPolicy is "Cluster",
-	// traffic from external sources will be routed based on the trafficPolicy. When
-	// externalTrafficPolicy is "Local", trafficPolicy is ignored for traffic from
-	// external sources.
+	// InternalTrafficPolicy specifies if the cluster internal traffic
+	// should be routed to all endpoints or node-local endpoints only.
+	// "Cluster" routes internal traffic to a Service to all endpoints.
+	// "Local" routes traffic to node-local endpoints only, traffic is
+	// dropped if no node-local endpoints are ready.
+	// The default value is "Cluster".
+	// +featureGate=ServiceInternalTrafficPolicy
 	// +optional
-	// +feature-gate=ServiceTrafficPolicy
-	TrafficPolicy ServiceTrafficPolicyType `json:"trafficPolicy,omitempty"`
+	InternalTrafficPolicy *ServiceInternalTrafficPolicyType `json:"internalTrafficPolicy,omitempty" protobuf:"bytes,22,opt,name=internalTrafficPolicy"`
 }
 ```
 
-This new field will intersect with externalTrafficPolicy in the following ways:
-* if `externalTrafficPolicy=Cluster`, traffic will be routed based on `trafficPolicy` for external sources
-* if `externalTrafficPolicy=Local`, `externalTrafficPolicy` will take precedent over `trafficPolicy`, but only for external sources.
+This field will be independent from externalTrafficPolicy. In other words, internalTrafficPolicy only applies to traffic originating from internal sources.
 
 Proposed changes to kube-proxy:
-* when `trafficPolicy=Cluster`, default to existing behavior today.
-* when `trafficPolicy=Topology`, use topology hints from EndpointSlice API.
-* when `trafficPolicy=PreferLocal`, route to endpoints in EndpointSlice that matches the local node's topology (topology defined by `kubernetes.io/hostname`),
-fall back to "Cluster" behavior if there are no local endpoints.
-* when `trafficPolicy=Local`, route to endpoints in EndpointSlice that maches the local node's topology, drop traffic if none exist.
+* when `internalTrafficPolicy=Cluster`, default to existing behavior today.
+* when `internalTrafficPolicy=Local`, route to endpoints in EndpointSlice that maches the local node's topology, drop traffic if none exist.
+
+Overlap with topology-aware routing:
+
+| ExternalTrafficPolicy | InternalTrafficPolicy | Topology | External Result | Internal Result |
+| - | - | - | - | - |
+| - | - | Auto | Topology | Topology |
+| Local | - | Auto | Local | Topology |
+| Local | Local | Auto | Local | Local |
 
 ### Test Plan
 
 Unit tests:
-* unit tests validating API strategy/validation for when `trafficPolicy` is set on Service.
-* unit tests exercising kube-proxy behavior when `trafficPolicy` is set to all possible values.
+* unit tests validating API strategy/validation for when `internalTrafficPolicy` is set on Service.
+* unit tests exercising kube-proxy behavior when `internalTrafficPolicy` is set to all possible values.
 
 E2E test:
-* e2e tests validating default behavior with kube-proxy did not change when `trafficPolicy` defaults to `Cluster`. Existing tests should cover this.
-* e2e tests validating that traffic is preferred to local endpoints when `trafficPolicy` is set to `PreferLocal`.
-* e2e tests validating that traffic is only sent to node-local endpoints when `trafficPolicy` is set to `Local`.
+* e2e tests validating default behavior with kube-proxy did not change when `internalTrafficPolicy` defaults to `Cluster`. Existing tests should cover this.
+* e2e tests validating that traffic is only sent to node-local endpoints when `internalTrafficPolicy` is set to `Local`.
 
 ### Graduation Criteria
 
 Alpha:
-* feature gate `ServiceTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.trafficPolicy`. Otherwise field is dropped.
-* kube-proxy handles traffic routing for 4 initial internal traffic policies `Cluster`, `Topology`, `PreferLocal` and `Local`.
+* feature gate `ServiceInternalTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.internalTrafficPolicy`. Otherwise field is dropped.
+* kube-proxy handles traffic routing for 2 initial internal traffic policies `Cluster`, and `Local`.
 * Unit tests as defined in "Test Plan" section above. E2E tests are nice to have but not required for Alpha.
 
+Beta:
+* integration tests exercising API behavior for `spec.internalTrafficPolicy` field of Service.
+* e2e tests exercising kube-proxy routing when `internalTrafficPolicy` is `Local`.
+* feature gate `ServiceInternalTrafficPolicy` is enabled by default.
+* consensus on how internalTrafficPolicy overlaps with topology-aware routing.
 
 ### Upgrade / Downgrade Strategy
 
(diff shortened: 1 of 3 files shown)


keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -123,51 +119,46 @@ type ServiceSpec struct {
...
-* kube-proxy handles traffic routing for 4 initial internal traffic policies `Cluster`, `Topology`, `PreferLocal` and `Local`.
+* feature gate `ServiceInternalTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.internalTrafficPolicy`. Otherwise field is dropped.
+* kube-proxy handles traffic routing for 2 initial internal traffic policies `Cluster`, and `Local`.
 * Unit tests as defined in "Test Plan" section above. E2E tests are nice to have but not required for Alpha.
 
+Beta:
+* integration tests exercising API behavior for `spec.internalTrafficPolicy` field of Service.
+* e2e tests exercising kube-proxy routing when `internalTrafficPolicy` is `Local`.
+* feature gate `ServiceInternalTrafficPolicy` is enabled by default.
+* consensus on how internalTrafficPolicy overlaps with topology-aware routing.
```
andrewsykim, line 161: TBH I lost track of where we landed on this specific topic, but we should probably reach consensus on the overlap prior to going to Beta.

andrewsykim (commented):
/assign @thockin @maplain @robscott

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: My current understanding is that we removed PreferLocal because it can now be handled with topology-aware routing (without endpointslice subsetting). What is our current thinking around how `internalTrafficPolicy: Local` and the new topology-aware API overlap? Are there plans for topology-aware routing to support something like `RequireNode`??

robscott: Here's my understanding: *TrafficPolicy=Local always have precedence but can be used in conjunction with topology. Here's how I'm interpreting it anyway:

| ExternalTrafficPolicy | InternalTrafficPolicy | Topology | External Result | Internal Result |
| - | - | - | - | - |
| - | - | Auto | Topology | Topology |
| Local | - | Auto | Local | Topology |
| Local | Local | Auto | Local | Local |

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -76,17 +76,15 @@ for internal Service traffic.
...
+Introduce a new field in Service `spec.internalTrafficPolicy`. The field will have 2 codified values:
 1. Cluster (default): route to all cluster-wide endpoints (or use topology aware subsetting if enabled).
-2. Topology: route to endpoints using topology-aware routing. See Topology Aware Hints KEP for more details.
-2. PreferLocal: route to node-local endpoints if it exists, otherwise fallback to behavior from Cluster.
-3. Local: only route to node-local endpoints, drop otherwise.
+2. Local: only route to node-local endpoints, drop otherwise.
 
-A feature gate `ServiceTrafficPolicy` will also be introduced for the alpha stage of this feature.
-The `trafficPolicy` field cannot be set on Service during the alpha stage unless the feature gate is enabled.
+A feature gate `ServiceInternalTrafficPolicy` will also be introduced for the alpha stage of this feature.
```
robscott, line 83: Nit: I don't think this is only for alpha:
```suggestion
A feature gate `ServiceInternalTrafficPolicy` will also be introduced for this feature.
```

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -123,51 +119,46 @@ type ServiceSpec struct {
...
-	// topology hints. "PreferLocal" will route internal traffic to node-local endpoints
-	// if one exists, otherwise it will fallback to the same behavior as "Cluster".
-	// "Local" routes traffic to node-local endpoints only, traffic is dropped
-	// if no node-local endpoints are ready. When externalTrafficPolicy is "Cluster",
-	// traffic from external sources will be routed based on the trafficPolicy. When
-	// externalTrafficPolicy is "Local", trafficPolicy is ignored for traffic from
-	// external sources.
+	// InternalTrafficPolicy specifies if the cluster internal traffic
+	// should be routed to all endpoints or node-local endpoints only.
+	// "Cluster" routes internal traffic to a Service to all endpoints.
```
robscott, line 124: This probably needs a bit more nuance since it's not necessarily all endpoints. I don't think the specifics need to block this PR though.

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -123,51 +119,46 @@ type ServiceSpec struct {
...
 
 Alpha:
-* feature gate `ServiceTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.trafficPolicy`. Otherwise field is dropped.
-* kube-proxy handles traffic routing for 4 initial internal traffic policies `Cluster`, `Topology`, `PreferLocal` and `Local`.
+* feature gate `ServiceInternalTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.internalTrafficPolicy`. Otherwise field is dropped.
+* kube-proxy handles traffic routing for 2 initial internal traffic policies `Cluster`, and `Local`.
 * Unit tests as defined in "Test Plan" section above. E2E tests are nice to have but not required for Alpha.
 
+Beta:
+* integration tests exercising API behavior for `spec.internalTrafficPolicy` field of Service.
```
robscott, line 158: I don't see these in the test plan above, do you have any specific ones you're thinking of?

robscott (commented):
Mostly LGTM, just a couple questions.

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: Thanks Rob! That makes sense to me.

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -123,51 +119,46 @@ type ServiceSpec struct {
...
-	// to cluster-wide endpoints or node-local endpoints. "Cluster" routes traffic
-	// to a Service to all cluster-wide endpoints. "Topology" routes traffic based on
-	// topology hints. "PreferLocal" will route internal traffic to node-local endpoints
-	// if one exists, otherwise it will fallback to the same behavior as "Cluster".
-	// "Local" routes traffic to node-local endpoints only, traffic is dropped
-	// if no node-local endpoints are ready. When externalTrafficPolicy is "Cluster",
-	// traffic from external sources will be routed based on the trafficPolicy. When
-	// externalTrafficPolicy is "Local", trafficPolicy is ignored for traffic from
-	// external sources.
+	// InternalTrafficPolicy specifies if the cluster internal traffic
```
wojtek-t, line 122: L104 - remove reference to "PreferLocal"

keps/sig-network/2086-service-internal-traffic-policy/README.md
wojtek-t: I like this table - it explains a lot. Can we put this into KEP?

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -207,54 +192,55 @@ Yes, the feature gate can be disabled, but Service resource that have set the ne
 
 * **What happens if we reenable the feature if it was previously rolled back?**
 
-New Services should be able to set the `trafficPolicy` field. Existing Services that have the field set already should not be impacted.
+New Services should be able to set the `internalTrafficPolicy` field. Existing Services that have the field set already should not be impacted.
```
wojtek-t, line 195: Actually, existing services with the field set will be impacted because their behavior will be reenabled (i.e. we will start enforcing local policy again).

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -207,54 +192,55 @@ Yes, the feature gate can be disabled, but Service resource that have set the ne
...
 
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can an operator determine if the feature is in use by workloads?**
 
-TBD for beta.
+* Check Service to see if `internalTrafficPolicy` is set to `Local`.
+* Check iptable rules on kube-proxy to see if cluster IP routing is only being sent to node-local endpoints.
```
wojtek-t, line 229: Operators can't reliably do that if they manage hundreds or thousands of clusters.

Can we add a metric to kube-proxy that will be exposing it (e.g. some counter that incremenets when we add node-local endpoints or sth like that?)

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -207,54 +192,55 @@ Yes, the feature gate can be disabled, but Service resource that have set the ne
...
 
 * **What are the SLIs (Service Level Indicators) an operator can use to determine
 the health of the service?**
 
-TBD for beta.
+N/A
```
wojtek-t, line 234: Disagree - see my comment above.

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: This was just copied from what's in the API spec now, I agree we should re-evaluate the wording here.

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: No, I haven't thought that deep into the details yet.

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: I could, but there's not much precedent for this. If we did it for internal traffic policy, we'd want to do it for a bunch of other error cases with kube-proxy. Do we really want that?

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: done!

wojtek-t: Thanks!

keps/sig-network/2086-service-internal-traffic-policy/README.md
wojtek-t: @thockin - thoughs?

I would say that the fact that we're missing observability now doesn't mean that we should continue keeping it non-observable. So I would say that let's start adding for new stuff at least.

thockin: It seems reasonable that we could add metrics for kube-proxy, to capture its perceived breakdown of things.  We have not much to build on here, though.

I could imagine a metric that simply tracks the number of services labelled by external and internal traffic policy, but that doesn't say much about this feature.  We could add a metric that counts local vs non-local endpoints but that's not useful unless we carve it up by internal and external traffic policy.  I'm worried that the logic to carve up the metrics becomes about as complicated as the logic it is tracking.

What are we really trying to answer?  Something like "are cases where policy is local staying local?", right?  We can't expose per-service metrics, and absolute numbers are not super useful.  I'm skeptical that the code would even have a path for recognizing "local policy, non-local endpoint" to be counted.

What about something simple like "blackholes" labelled by ITP ?  If a service has no endpoints, that is interesting.  If a service has ITP=local and no endpoints, that's alertable (probably).

intersecting that with the other kep, maybe label by ITP and by conditions?

robscott: > If a service has ITP=local and no endpoints, that's alertable (probably)

Not too familiar with all the use cases here, but one I can imagine would involve different sets of nodes for different workloads. In that case traffic from a needs to reach b where ITP=local, but both a and b only run on the same subset of nodes. Technically that would result in a lot of blackholes for other nodes where the Service existed but local endpoints did not. Not sure how big of a concern that is or if it meaningfully limits the usefulness of that metric, it's just one of the ways I could imagine ITP being used.

thockin: I agree it's possible, and the good news is we don't have to define alerts here.  :)  My point stands - when it comes to traffic policy "working" we want to answer 2 questions:

1) Are we finding the local endpoints? (blackholes = 0)
2) Are we ONLY finding the local endpoints?

The 2nd is much harder to instrument without splitting the impl up, I think.  The code path that would accumulate non-local endpoints for a local service doesn't exist because it is non-sense, and we don't carry an intermediate representation that we can cross-check ex post facto (which we can and maybe should fix),

E.g.  convert the sync loop to 2 stages.  First, loop over all services and gather the endpoints for each "path" (internal/external client).  Second, convert that into iptables/ipvs.  Then you can run sanity checks in between like "if service is ITP local and has non-local endpoints, there's a bug".  

Unsurprisingly, this sounds quite a lot like kpng.  So maybe adding blackholes is a good start and we make sure kpng can handle such other metrics?

@mcluseau

wojtek-t: Great discussion. And I buy the argument that doing (2) would be hard and the logic would be similarly complex to the code itself.

So if we can just export metric about blackholes - that sounds fine to me (and that seems more important imho).
So yeah - let's just expose blackholes.

andrewsykim: Added some words to add the "blackhole" metrics, we can discuss naming of the actual metric during PR review

keps/sig-network/2086-service-internal-traffic-policy/README.md
aojea: I can't see any valuable integration test for this field ... should be better not committing to it?
we can always add them later but this will stop the graduation

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -103,7 +101,6 @@ Traffic should never bounce to a daemon on another node.
 ### Risks and Mitigations
 
 * When the `Local` policy is set, it is the user's responsibility to ensure node-local endpoints are ready, otherwise traffic will be dropped.
-* Using the `Local` or `PreferLocal` policy may result in imbalanced traffic for pods in a Service. It is the user's responsibility to handle this.
```
aojea, line 104: We have to add something like:

The special service `kubernetes` created by the apiserver will never have `internalTrafficPolicy` local or something like this

JamesLaverack:
xref https://github.com/kubernetes/enhancements/issues/2086

keps/sig-network/2086-service-internal-traffic-policy/README.md
thockin: I kind of agree with Antonio

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -123,51 +118,54 @@ type ServiceSpec struct {
...
-* kube-proxy handles traffic routing for 4 initial internal traffic policies `Cluster`, `Topology`, `PreferLocal` and `Local`.
+* feature gate `ServiceInternalTrafficPolicy` _must_ be enabled for apiserver to accept values for `spec.internalTrafficPolicy`. Otherwise field is dropped.
+* kube-proxy handles traffic routing for 2 initial internal traffic policies `Cluster`, and `Local`.
 * Unit tests as defined in "Test Plan" section above. E2E tests are nice to have but not required for Alpha.
 
+Beta:
+* integration tests exercising API behavior for `spec.internalTrafficPolicy` field of Service.
+* e2e tests exercising kube-proxy routing when `internalTrafficPolicy` is `Local`.
+* feature gate `ServiceInternalTrafficPolicy` is enabled by default.
+* consensus on how internalTrafficPolicy overlaps with topology-aware routing.
```
thockin, line 168: now that the table is in - remove this line?

thockin (commented):
Thanks!

/lgtm
/approve

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -207,54 +199,55 @@ Yes, the feature gate can be disabled, but Service resource that have set the ne
...
 
 * **How can a rollout fail? Can it impact already running workloads?**
 
-TBD for beta.
+Rollout should have minimal impact because the default value of `internalTrafficPolicy` is `Cluster`, which is the default behavior today.
 
 * **What specific metrics should inform a rollback?**
 
-TBD for beta.
+N/A
```
wojtek-t, line 218: Probably the blackhole metrics should be used here.

keps/sig-network/2086-service-internal-traffic-policy/README.md
```diff
@@ -207,54 +199,55 @@ Yes, the feature gate can be disabled, but Service resource that have set the ne
...
 
 ### Monitoring Requirements
 
 _This section must be completed when targeting beta graduation to a release._
 
 * **How can an operator determine if the feature is in use by workloads?**
 
-TBD for beta.
+* Check Service to see if `internalTrafficPolicy` is set to `Local`.
+* A new metric will be added to represent Services that are being "blackholed" (internalTrafficPolicy=Local and no endpoints).
```
wojtek-t, line 236: I'm assuming it's "per-node" metric (i.e. exported by kube-proxy), right? If so - can we make it explicit.

[Also - please add a TODO to fill in the metric name here once it is implemented].

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: done

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: done

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: fixed, thanks

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: Me too, will re-evalute whether integration tests are worth adding during PR

keps/sig-network/2086-service-internal-traffic-policy/README.md
andrewsykim: That's managed by kube-apiserver, so I don't see how it's a risk? Can you expand?

thockin: We don't disclaim every other field, I don't think we need to disclaim this one.

thockin (commented):
/lgtm
/approve

wojtek-t:
I'm fine with this from PRR POV.

That said, an exception has to be filed, right? @andrewsykim 

/approve PRR
/hold

andrewsykim:
@annajung do you think this needs an exception? SIG approval was on-time, but PRR approval was missed?

annajung:
Hey @andrewsykim since the enhancement freeze deadline requires all KEPs to be merged with all requirements (including PRR) and approvals, this does require an exception. I recommend [requesting an exception](https://github.com/kubernetes/sig-release/blob/master/releases/EXCEPTIONS.md#requesting-an-exception) as soon as possible to get it reviewed by the release team. Thanks! cc @JamesLaverack

andrewsykim:
Exception filed, thanks

annajung:
Exception [approved](https://groups.google.com/g/kubernetes-sig-release/c/PTRqQTorb7c/m/izp7WjZBBAAJ)
/hold cancel

PR #2734, opened 2021-05-13
KEP-1645: make clusterset optional

rosenhouse:
Draft changes, based on our conversation at SIG-Multicluster on May 11

### DIFF
--- a/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
+++ b/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
@@ -93,9 +93,9 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Exporting Services](#exporting-services)
     - [Restricting Exports](#restricting-exports)
   - [Importing Services](#importing-services)
-  - [ClusterSet Service Behavior Expectations](#clusterset-service-behavior-expectations)
+  - [Multicluster Service Behavior Expectations](#multicluster-service-behavior-expectations)
     - [Service Types](#service-types)
-    - [ClusterSetIP](#clustersetip)
+    - [MultiClusterIP](#multiclusterip)
     - [DNS](#dns)
     - [EndpointSlice](#endpointslice)
     - [Endpoint TTL](#endpoint-ttl)
@@ -234,16 +234,34 @@ implementation.  The "Design Details" section below is for the real
 nitty-gritty.
 -->
 #### Terminology
+- **multicluster namespace** - A placeholder name for a group of
+  Kubernetes namespaces with the same name on different clusters.
+  This API specifies the semantics for sharing services
+  amongst the namespaces that belong to a multicluster namespace.
+
+  Implementations of this API are responsible for defining and tracking
+  multicluster namespaces and their membership. The specific mechanism
+  is out of scope of this proposal.
 
 - **clusterset** - A placeholder name for a group of clusters with a high
   degree of mutual trust and shared ownership that share services amongst
   themselves. Membership in a clusterset is symmetric and transitive. The set
   of member clusters are mutually aware, and agree about their collective
-  association. Within a clusterset, [namespace sameness] applies and all
-  namespaces with a given name are considered to be the same namespace.
-  Implementations of this API are responsible for defining and tracking
-  membership in a clusterset. The specific mechanism is out of scope of this
-  proposal.
+  association. Within a clusterset, [namespace sameness] applies such that all
+  namespaces with a given name are considered members of a same-named
+  *multicluster namespace*.
+
+  > Note: Clusterset is an optional, high-trust model for defining multicluster
+  namespaces. Implementations are not required to implement clustersets.
+  Instead, they may provide other means for defining multicluster namespaces, for
+  example by allowing directional, pair-wise federation between same-named
+  namespaces of mutually-untrusting clusters.  However, the clusterset model is
+  easy to reason about and a natural model for implementors to target.
+
+  Implementations of this API that choose to implement clustersets are responsible
+  for defining and tracking their membership. The specific mechanism is out of
+  scope of this proposal.
+
 - **mcs-controller** - A controller that syncs services across clusters and
   makes them available for multi-cluster service discovery and connectivity.
   There may be multiple implementations, this doc describes expected common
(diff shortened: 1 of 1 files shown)


keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
...
+  association. Within a clusterset, [namespace sameness] applies such that all
+  namespaces with a given name are considered members of a same-named
+  *multicluster namespace*.
+
+  > Note: Clusterset is an optional, high-trust model for defining multicluster
+  namespaces. Implementations are not required to implement clustersets.
+  Instead, they may provide other means for defining multicluster namespaces, for
+  example by allowing directional, pair-wise federation between same-named
+  namespaces of mutually-untrusting clusters.  However, the clusterset model is
+  easy to reason about and a natural model for implementors to target.
```
JeremyOT, line 259: This to me is the core of the edit. We don't want to remove ClusterSet from the spec. MCS is defined based on ClusterSets and I don't think we want to change that. What we do want to say is that it is valid to implement the MCS API outside of ClusterSets, but that not all features may work as expected and you'll need to understand what that means for your chosen implementation - e.g. does namespace sameness only apply selectively? Do all clusters receive the import?

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
 nitty-gritty.
 -->
 #### Terminology
+- **multicluster namespace** - A placeholder name for a group of
```
pmorie, line 237: I'm very name-coining averse in the area of multicluster so I have some hesitation about this name. Basically no matter what name we choose for anything in this area, there is a chance that someone will reason incorrectly about it given implications of the name. How about the term 'client namespace' instead here? I would hate for someone to read into this an implication that there will be some technology called 'multicluster namespaces' that comes out of this SIG. I know there is a note below, but snippets of this doc might be pasted that don't include the note.

I hope that doesn't come across as nitpicky, I have had many many years of dispelling mistaken impressions about the multicluster area of kube community that were based on such assumptions folks had (understandably) made.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
...
-has been exported within the clusterset, referenced by namespaced name.
+has been exported within a multicluster namespace.
 `ServiceImport` resources will be managed by the MCS implementation's
 mcs-controller.
 
-If multiple clusters export a `Service` with the same namespaced name, they will
-be recognized as a single combined service. For example, if 5 clusters export
-`my-svc.my-ns`, each importing cluster will have one `ServiceImport` named
-`my-svc` in the `my-ns` namespace and it will be associated with endpoints from
+If multiple clusters export a `Service` into the same multicluster namespace, they will
```
pmorie, line 303: how does it sound if you said 'into a client namespace' here, and ...

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
...
+be recognized as a single combined service from the importing clusters. For example,
+if 5 clusters export `my-svc` into a multicluster namespace `my-ns`, then each importing
```
pmorie, line 305: ... and here you said:

> into client namespace `my-ns`

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
...
+cluster associated with the multicluster namespace will have one `ServiceImport`
```
pmorie, line 306: ... and here you said:

> with the client namespace

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -362,8 +385,8 @@ proposal will be implemented, this is the place to discuss them.
 -->
 ### Exporting Services
 
-Services will not be visible to other clusters in the clusterset by default.
-They must be explicitly marked for export by the user. This allows users to
+Services will not be visible to other clusters by default.
+They must be explicitly marked for export. This allows users to
```
pmorie, line 389: I think being marked for export by the user is an important part of the original that should be retained.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -449,30 +472,30 @@ status:
   - type: Conflict
     status: "True"
     lastTransitionTime: "2020-03-30T01:33:55Z"
-    message: "Conflicting type. Using \"ClusterSetIP\" from oldest service export in \"cluster-1\". 2/5 clusters disagree."
+    message: "Conflicting type. Using \"MultiClusterIP\" from oldest service export in \"cluster-1\". 2/5 clusters disagree."
```
pmorie, line 475: I think I'd rather see "MulticlusterServiceIP" here, which is aligned with the name of this concept, wdyt?

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -449,30 +472,30 @@ status:
...
 same namespaced name._
 
-This requires that within a clusterset, a given namespace is governed by a
-single authority across all clusters. It is that authority’s responsibility to
-ensure that a name is shared by multiple services within the namespace if and
-only if they are instances of the same service.
+This requires that the exporting of services within a multicluster namespace
+be governed by a single authority across all clusters.  This may be achieved in different ways:
+- For example, in the case of a clusterset model with high-trust amongst clusters, an authority would ensure that the namespaced-name of a service is used across multiple clusters if and only if they are instances of the same service.
+- In the case of a federated, non-clusterset model, with low-trust between clusters, the implementation of this API must have outside knowledge of cluster-namespace ownership and use that to validate `ServiceExport`s being reconciled.
```
pmorie, line 491: My preference would be to avoid use of the term 'federation' here, it is nearly universally confusing in practice, especially given the presence of projects with that name in this community.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -550,8 +578,8 @@ type ServiceImport struct {
 type ServiceImportType string
 
 const (
-  // ClusterSetIP are only accessible via the ClusterSet IP.
-  ClusterSetIP ServiceImportType = "ClusterSetIP"
+  // MultiClusterIP are only accessible via the Multicluster IP.
```
pmorie, line 581: Let's just call this specifically what it is - the IP of the multicluster service

pmorie (commented):
Looks like a quality edit, but I have some concerns about names of various things, as I explain below :)

Thank you for raising this case, it's great to get feedback from the community on these concepts!

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -234,54 +234,77 @@ implementation.  The "Design Details" section below is for the real
...
 -->
 #### Terminology
+- **multicluster namespace** - A placeholder name for a group of
+  Kubernetes namespaces with the same name on different clusters.
+  This API specifies the semantics for sharing services
+  amongst the namespaces that belong to a multicluster namespace.
+
+  Implementations of this API are responsible for defining and tracking
+  multicluster namespaces and their membership. The specific mechanism
+  is out of scope of this proposal.
```
JeremyOT, line 244: I'm concerned about how this weakens assumptions around this API and puts more on the consumer to figure out how each implementation works, which doesn't help with the consistency we're going for. As we discussed in the weekly meeting, it's certainly possible to build a reasonable system that manages membership/scope for each namespace separately and there's value in that, but it's more challenging to reason about. It should be possible to use the MCS API with such system, but I don't think that should be the baseline.

Ideally, consumers can assume implementations follow namespace sameness across all clusters involved in MCS unless they've adopted an implementation that provides a more fine grained membership concept, in which case they can still use the MCS API but either they or the implementation will need to provide appropriate guidance/guardrails/feedback around implementation-specific behavior.

JeremyOT (commented):
Thanks for this!

I'm concerned about the idea of removing ClusterSet from the spec and what it does to intuitiveness. My first thought was to make it very clear that implementations can use the MCS API outside of ClusterSets and just need to make their behavior clear to users, but what if instead, we focus on the ClusterSet as the customization point.

In that world: MCS works with ClusterSets, ClusterSets imply namespace sameness across member clusters, but some implementations may have more fluid definitions of a ClusterSet. What do you think?

rosenhouse:
Thanks for the reviews everyone, your feedback is pushing me to re-think the approach here.

Also: I'm going to be working on other things for the next few months.  So time to close this!

PR #2739, opened 2021-05-13
Update the KEP to call out why we don't plan to create groups for sha…

vinayakankugoyal:
…red access to certificate files.


neolit123 (commented):
we had a discussion on slack with @vinayakankugoyal . he found that we can omit some of the original planned user/groups in the initial KEP.

/lgtm
/approve
/ok-to-test

PR #2740, opened 2021-05-13
Cgroups v2: Update latest milestone

mrunalp:
Replaces #2689 to address comments

### DIFF
--- a/keps/sig-node/2254-cgroup-v2/README.md
+++ b/keps/sig-node/2254-cgroup-v2/README.md
@@ -132,7 +132,7 @@ yes, it is enough to restart the node on cgroup v1
 
 ###### What happens if we reenable the feature if it was previously rolled back?
 
-it should work seamlessly without any difference
+It should work seamlessly without any difference
 
 ###### Are there any tests for feature enablement/disablement?
 
@@ -149,7 +149,8 @@ cgroup file system, then also the workload must be enabled for cgroup v2.
 
 ###### What specific metrics should inform a rollback?
 
-Pods not being healthy.
+Pods not being healthy. One could inspect if the pods are getting the cgroups
+set correctly referencing the conversion table in this KEP.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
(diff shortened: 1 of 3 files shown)


mrunalp:
@johnbelamaric could you take a look? I have addressed your comments here. (It is late for @giuseppe ) so opened this to try to make the deadline.

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,186 @@ This proposal aims to:
...
+N/A.  Each node can be configured separately.
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+N/A.  It requires a reboot to be enabled.  If the workload accesses directly the
+cgroup file system, then also the workload must be enabled for cgroup v2.
+
+###### What specific metrics should inform a rollback?
+
+Pods not being healthy. One could inspect if the pods are getting the cgroups
```
derekwaynecarr, line 152: i anticipate that a kubelet would start all pods or none.  i doubt you ever drop down to the cgroup tree to see if values are set appropraitely, but it would clearly be observable from any host monitoring system that is cgroup aware.

derekwaynecarr (approved):
/approve
/lgtm

its important to note that the kubelet is running on a host that the linux admin has preconfigured.  as part of bringing that host, it is assumed that the linux admin knows how they configured their operating system, and how they intend to monitor processes on that system independent of kubelet launched workloads.  we know this feature does not work when kubelet doesnt start, or it it is unable to launch _any_ pod.  when we start leveraging cgroup v2 unique features for memory qos (separate merged kep), we will start to see more unique items to monitor and measure improvement around particularly for oom avoidance.

keps/sig-node/2254-cgroup-v2/README.md
```diff
@@ -52,6 +63,186 @@ This proposal aims to:
...
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+The cgroup file system inside of the containers will use cgroup v2 instead of cgroup v1.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+An operator could run `cat /proc/self/cgroup` on a node to check if it is running in cgroups v2 mode.
```
johnbelamaric, line 167: Ok. Please consider in the future (GA) making this sort of information available via the API or a metric. Like in Node status or something. Because if I am monitoring 1000s of clusters with 1000s of nodes each, this is not convenient. I won't block this now though for that.

johnbelamaric (commented):
/approve

keps/sig-node/2254-cgroup-v2/README.md
mrunalp: Thanks!

Issue #2771, opened 2021-06-02
Supports https for httpGet of Lifecycle


wzshiming:
# Enhancement Description
- One-line enhancement description (can be used as a release note): Supports https for httpGet of Lifecycle
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/2772
- Discussion Link: 
- Primary contact (assignee):
- Responsible SIGs: 
  - sig-node
- Enhancement target (which target equals to which milestone):
  - Alpha release target (x.y): v1.23
  - Beta release target (x.y): v1.24
  - Stable release target (x.y): v1.26
- [ ] Alpha
  - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/2772
  - [x] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/102474
  - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #2772, opened 2021-06-02
KEP-2771: Supports https for httpGet of Lifecycle

wzshiming:
Supports https for httpGet of Lifecycle

KEP #2771
PR kubernetes/kubernetes#102474

### DIFF
--- a/keps/sig-node/2771-pod-lifecycle-https/README.md
+++ b/keps/sig-node/2771-pod-lifecycle-https/README.md
@@ -1,0 +1,581 @@
+# KEP-2771: Supports https for httpGet of Lifecycle
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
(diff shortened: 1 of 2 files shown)


keps/sig-node/2771-pod-lifecycle-https/README.md
```diff
@@ -0,0 +1,579 @@
...
+### Goals
+
+- Supports https for httpGet of Lifecycle
+
+### Non-Goals
+
+## Proposal
+
+### User Stories
+
```
aojea, line 89: what about something like
```
as a Kubernetes user I want to be able to use https for my pod lifecycle hooks
```

wzshiming: Thanks, updated

keps/sig-node/2771-pod-lifecycle-https/README.md
```diff
@@ -0,0 +1,581 @@
...
+For example, consider both security and how this will impact the larger
+Kubernetes ecosystem.
+How will security be reviewed, and by whom?
+How will UX be reviewed, and by whom?
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+## Design Details
+
+Supports HTTPS scheme for httpGet of Lifecycle, and skip TLS verification.
```
khenidak, line 109: I am not sure we should support skip `tls verification`. At least it must not be the default behavior. Must be an opt in. 

If we assume that the reason of supporting TLS is security of the lifecycle hook, then it does not make sense to skip TLS verification.

aojea: See discussion here https://github.com/kubernetes/kubernetes/pull/102474#issuecomment-854704317

The other dimension is if we want to make http lifecycle hooks equal to probes, that does skip TLS verification ... this is a tough one

PR #2777, opened 2021-06-02
KEP 2775: kubectl delete interactivity and delay

eddiezane:
ref: #2775

### DIFF
--- a/keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
+++ b/keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
@@ -1,0 +1,579 @@
+# KEP-2775: kubectl delete interactivity and delay
+
+<!--
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests for meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+When a namespace is deleted it also deletes all of the resources under it. The deletion runs without further confirmation, and can be devastating if accidentally run against the wrong namespace (e.g. thanks to hasty tab completion use).  
+
+```  
+kubectl delete namespace prod-backup  
+```
+
+When all namespaces are deleted essentially all resources are deleted. This deletion is trivial to do with the `--all` flag, and it also runs without further confirmation. It can effectively wipe out a whole cluster.  
+
+```
+kubectl delete namespace --all
+```  
+
+There are certainly things cluster operators should be doing to help prevent this user error (like locking down permissions) but it doesn't matter how many mistakes have to pile up to create a perfect storm of a bad thing because we're allowing a bad thing to happen without a confirmation gate.
+
+This KEP proposes two protections when deleting resources with kubectl.
+
+## Motivation
+
+Over the years we've heard stories from users who've accidentally deleted resources in their clusters - often all of their resources. This trend seems to be rising lately as newer folks venture into the Kubernetes/DevOps/Infra world. Experienced users are also not accident proof and have shared tragic tales.
+
+### Goals
+
+* Mitigate the potential for accidental imperative deletes
+
+### Non-Goals
+
+* Server side solutions
+
+## Proposal
+
+1. `kubectl delete [--all | --all-namespaces]` will warn about the destructive action that will be performed and artificially delay for x seconds allowing users a chance to abort.
+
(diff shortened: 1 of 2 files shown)


keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
```diff
@@ -0,0 +1,579 @@
...
+
+* Mitigate the potential for accidental imperative deletes
+
+### Non-Goals
+
+* Server side solutions
+
+## Proposal
+
+1. `kubectl delete [--all | --all-namespaces]` will warn about the destructive action that will be performed and artificially delay for x seconds allowing users a chance to abort.
```
briantopping, line 93: I'd like to see the same behavior with `kubeadm reset` when the node being acted upon would remove etcd quorum.

briantopping (commented):
Looks good! One request, the KEP is currently oriented toward `kubectl` when I believe it ought to be a system-wide initiative to stop existentially destructive behaviors. `kubeadm reset` is in this realm, but not all resets are existential. When such a situation is detected, the same behavior ought to be used. Are there other systems of this nature that should also be covered?

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
neolit123: `kubeadm reset` shows a confirmation prompt, unless the user passes `-f`.

```
$ kubeadm reset 
[reset] Reading configuration from the cluster...
[reset] FYI: You can look at this config file with 'kubectl -n kube-system get cm kubeadm-config -oyaml'
[reset] WARNING: Changes made to this host by 'kubeadm init' or 'kubeadm join' will be reverted.
[reset] Are you sure you want to proceed? [y/N]: 
```

it's phases are documented, can be executed on demand or skipped:
https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-reset/

briantopping: Maybe the resolution to this is after the user presses 'y' and the cluster is determined to go unstable after such a request, a second warning is displayed alerting to such.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
```diff
@@ -0,0 +1,579 @@
...
+> Warning! You are deleting the following resources:
+> Name:    Group:        Kind:
+> my-pod    core/v1      Pod
+> Proceeding in x seconds. ctrl+c to interrupt
+
+If the `--interactive` flag is passed for imperative deletes we will prompt the user for confirmation. As above we need to determine the prompt -- a generic message or explicitly lay out the resources.
+
+```
+kubectl delete namespace prod -i
+Are you sure? (y/n)
```
thockin, line 173: suggest default to N.  e.g. ` Are you sure? (y/N)`

thockin (commented):
I'd like to entertain the idea of API-side delete-inhibition.  Do you think that should be here or a different KEP?

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
eddiezane: Good call. Changed and explicitly called out.

eddiezane:
> I'd like to entertain the idea of API-side delete-inhibition. Do you think that should be here or a different KEP?

I think that should be in a different KEP. I liked what @brendandburns shared about the locking Azure does.

> We can add an annotation ("k8s.io/confirm-delete: true") to a Pod and if that annotation is present, prompt for confirmation of the delete. We might also consider "k8s.io/lock" which actively blocks the delete.

> We could also support those annotations at a namespace level if we wanted to.

> This is similar to Management Locks that we introduced in Azure (https://docs.microsoft.com/en-us/rest/api/resources/managementlocks) for similar reasons to prevent accidental deletes and force an explicit action (remove the lock) for a delete to proceed.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
```diff
@@ -0,0 +1,579 @@
...
+The delay introduced in this KEP will cause CI/CD jobs to take x seconds longer unless authors add `-i=false`. As stated above we can potentially mitigate this by skipping if `CI=true` is set.
+
+## Alternatives
+
+We considered showing the confirmation prompt if a TTY was detected. After further research it was determined that we would run into issues with TTY detection due to platforms and pipelines spoofing a TTY in an attempt to match the output of developer's terminals (e.g. color).
+
+https://groups.google.com/g/kubernetes-dev/c/y4Q20V3dyOk/m/LOe7Id1DBgAJ
+https://groups.google.com/g/kubernetes-dev/c/y4Q20V3dyOk/m/W0y2fD-NAAAJ
+
+During RFC discussion several ideas were shared. One of them was a locking mechanism for resources that would prevent deletion. This is considered orthogonal to this KEP because users don't intend to delete the resources they are accidentally deleting - not that there are things that should never be deleted.
```
raghvenders, line 571: I agree with the statement - intent not to delete resource but accidentally being deleted. However, there are things that cannot be deleted are at different levels in Kubernetes like Kube-* namespaces, default namespaces if you try to delete it using the same kubectl command, it is 403 forbidden. But it is still possible if you clear the cluster, technically hierarchical.

Do you think is it worth re-phrasing  _not that there are things that should never be deleted_ ?

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
raghvenders: Is y/N is the standard across Kubernetes ? In some places, we do use Yes/No.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
```diff
@@ -0,0 +1,579 @@
...
+
+### Non-Goals
+
+* Server side solutions
+
+## Proposal
+
+1. `kubectl delete [--all | --all-namespaces]` will warn about the destructive action that will be performed and artificially delay for x seconds allowing users a chance to abort.
+
+2. Add a new `--interactive | -i` flag to `kubectl delete` that will require confirmation before deleting resources imperatively (i.e. not `-f`). This flag will be false by default. If this flag is explicitly set to false it will not prompt or artificially delay.
```
liggitt, line 95: how does this interact with other consumers of stdin in kubectl (e.g. [client-go credential plugins that make use of stdin](https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/541-external-credential-providers), or manifest reading from stdin like `-f -`)

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
liggitt, line 93: is this being proposed for delete invocations for:
* all `--all` invocations
* cluster-scoped resources
* namespacesd

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
```diff
@@ -0,0 +1,579 @@
...
+
+These two changes introduce significant protections for users and do not break backwards compatibility. Scripts will see an x second delay unless they explicitly set `-i=false`.
```
liggitt, line 97: > Scripts will see an x second delay unless they explicitly set `-i=false`.

I expected us to attempt to detect interactive invocations and only delay in those cases. That would limit the delay for scripts to environments which mimic interactive terminals, which seems better.

eddiezane: > I think I'd rather they set -i false explicitly
that is obvious AT THE CALL SITE as opposed to looking for ${CI} being set somewhere else
it's a big declaration of "I meant to do that"

eddiezane: These were @thockin's thoughts.

> I think I'd rather they set `-i false` explicitly
that is obvious AT THE CALL SITE as opposed to looking for `${CI}` being set somewhere else
it's a big declaration of "I meant to do that"
When it comes to scary stuff I'll take "annoyingly explicit" every time
e.g. there's not a `${UNSAFE_RM}` variable that implies `-f`

I did digging on detecting a TTY and it's a pretty mixed bag. A handful of CI/CD providers spoof a TTY to get nicer color output. I don't think we can reliably go that route.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
eddiezane: I think it will matter where in the call stack we read stdin. The client-go credential plugins probably happen early the flow? I see this living near the end right inside the delete command.

Will need to play around with it with. We can have some integration tests for these.

What would you like to see about it in the KEP?

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
eddiezane: > is this being proposed for delete invocations for:
> 
> * all `--all` invocations

Yes.

> * cluster-scoped resources
> * namespacesd

Not specifically. Only when deleted with `--all` or `--all-namespaces`.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
eddiezane: I'll check for precedent.

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
liggitt: > A handful of CI/CD providers spoof a TTY to get nicer color output. I don't think we can reliably go that route.

If the issue is false positive detection of interactive invocations, introducing delays to those false positives seems ok to me

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
liggitt: > The client-go credential plugins probably happen early the flow?

I think it depends on where the client config is constructed.

> What would you like to see about it in the KEP?

Describing specific places we use stdin in conjunction with `kubectl delete` calls today (`-f -`, exec plugins, etc), and that this `--interactive` option cannot be used in combination with those (and exercising the behavior when both are attempted with a testcase)

keps/sig-cli/2775-kubectl-delete-interactivity-delay/README.md
raghvenders: Just tried locally, how initial development looks like (pardon for the verbose/warning :) )
Screen Shot 2021-06-20 at 8 36 18 PM

@thockin @liggitt @eddiezane - If the interactive flag is not given, it suppose to go for delay. we can delay it using ticker but every second is expect to log as a line like ( Ignore the Order ascending or descending )

->resources will be deleted in 9 sec
->resources will be deleted in 8 sec... ??

Not sure if it is possible to do like state changes/clock in node/UI  - resources will be delete in (9..8..7.....) ?

mkimuram:
Could you check [this comment](https://github.com/kubernetes/enhancements/pull/2840#issuecomment-925258781) and give feedback?
I think that this KEP shares some use cases and motivations with KEP-2839 (In-use protection KEP), so I would appreciate it if the people involved to this KEP also feedback to KEP-2839.

eddiezane:
Going back to the drawing board on this.

ardaguclu:
For the people interested on this KEP, there was an another attempt in this proposed KEP https://github.com/kubernetes/enhancements/pull/3896. That would be great, if you can leave your insights on it too. Thanks.

Issue #2779, opened 2021-06-10
support custom queue for the scheduler


suigh:
# Enhancement Description
- One-line enhancement description (can be used as a release note): support custom queue for the scheduler
- Kubernetes Enhancement Proposal: 
- Discussion Link: https://github.com/kubernetes/kubernetes/issues/102271
- Primary contact (assignee): @suigh
- Responsible SIGs: sig-scheduling
- Enhancement target (which target equals to which milestone):
  - Alpha release target (1.22):
  - Beta release target (x.y):
  - Stable release target (x.y):
- [x] Alpha
  - [x] KEP (`k/enhancements`) update PR(s):
  https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler
  - [ ] Code (`k/k`) update PR(s):
  - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

suigh:
close as discussed in #2781.

PR #2781, opened 2021-06-10
support custom queue for the scheduler

suigh:
docs for KEP issue #2779

### DIFF
--- a/keps/prod-readiness/sig-scheduling/2779.yaml
+++ b/keps/prod-readiness/sig-scheduling/2779.yaml
@@ -1,3 +1,3 @@
 kep-number: 2779
-beta:
+alpha:
   approver: "@wojtek-t"
--- a/keps/prod-readiness/sig-scheduling/2779.yaml
+++ b/keps/prod-readiness/sig-scheduling/2779.yaml
@@ -1,0 +1,3 @@
+kep-number: 2779
+beta:
+  approver: "@wojtek-t"
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -375,5 +375,5 @@ sharing policy will not be broken. The drawback is that all the pods in the job 
 be impacted, it cannot handle the case that only some pods need to be touched.
 
 Another alternative is to make the pod that will break the resource sharing policy with
-`Unschedulable`. The drawback is that making such a decision needs the status of all other
-pods, the logic is complex, and the performance will be impacted.
+`Unschedulable`. The drawback is that making such a decision needs to know whether other
+pods can run or not, the logic is complex, and the performance will be impacted.
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -100,7 +100,8 @@ intact.
 
 ### Non-Goals
 
-Support custom rules for pod selection in the current internal queue of scheduler.
+1. Support custom rules for pod selection in the current internal queue of scheduler.
+2. Enable scheduler developers to change the internal queue, or queue sorting plugin.
 
 ## Proposal
 
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -119,14 +119,20 @@ But the default scheduler works as before and thus won't be impacted.
 
 ## Design Details
 
-The current internal queue is coupled with `cache`, `Option` and other implementations,
-this solution will not decouple them as this has no impact on the goal of custom queue.
-The coupled implementations will be exposed to users as custom queue needs to access them.
+The current internal queue is coupled with `cache`, `Option` and other implementations
+(e.g., the queue is initialized with specified parameters like `LessFunc`, `SharedInformerFactory`,
+and `Option` by [factory.go#create](https://github.com/kubernetes/kubernetes/blob/f1f0183d2bbcde33024b2a05d6f39df32f11e037/pkg/scheduler/factory.go#L172)),
+this solution will not decouple them as this has no impact on the goal of custom queue,
+and such logics can be enhanced in the future on demand.
+
+Some of the items like `SchedulingQueue`, `Option` that are only used by the internal
+queue now will be exposed at [interface.go](https://github.com/kubernetes/kubernetes/blob/fb9cafc99be94a73d9b92545164dbf336bbd230a/pkg/scheduler/framework/interface.go),
+users of custom queue can access them.
 
 To implement a custom queue, what an user needs to do is to copy the codes of the current
-internal queue (e.g., scheduling_queue.go, heap.go), update some logics to get the custom
-queue, and then pass the custom queue to the scheduler. With this way, user's logics will
-override the logics of the current internal queue.
+internal queue (e.g., scheduling_queue.go), update some logics to get the custom queue,
+and then pass the custom queue to the scheduler like the plugins. With this way, user's
+logics will override the logics of the current internal queue.
 
 The codes will be updated as following:
 
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -84,6 +84,14 @@ Given that the business needs may vary greatly, it'd be desirable to provide a r
 implement their custom queue implementation, while the upstream maintainers focus
 on keeping the core small and extensible, and thus only maintain the internal queue piece.
 
+This KEP is not used to change the internal queue, or queue sorting plugin, quite the contrary,
+that logic will be kept as is. The main idea is to give scheduler developers an option to manage the whole
+queuing logic, including how to prioritize/pop/backoff/flush pods. It's all up to the scheduler developers.
+To ensure the developers can have full control of the queuing logic, the whole interface of the internal
+queue will be exposed instead of part of it that can only meet above examples. By exposing the whole interface,
+we needn't update the interface that need to be exposed again and again for coming demands, this can save
+developers' effort as they needn't raise a KEP for such requests, this can also save community's effort.
+
 ### Goals
 
 1. Support custom queue implementation of the scheduler.
@@ -173,7 +181,6 @@ the interface `SchedulingQueue`, users must implement it.
 
 #### Beta -> GA Graduation
 
-- Only necessary interface and structs are exposed to users.
 - Allowing time for feedback to ensure that the new interface sufficiently expresses users requirements.
 
 ### Upgrade / Downgrade Strategy
@@ -366,14 +373,30 @@ Why should this KEP _not_ be implemented?
 
 ## Alternatives
 
-An alternative is to enhance the current queue to meet some requests. The drawback
-is that we cannot handle all the potential requests as the effort is big, we should
-give the ball to the users, and they can do what they want.
+For the multi-tenancy support case the mentioned in `motivation` section, with this KEP, scheduler developers
+can add some sub-queues, one sub-queue is for one user and put the pods for this user into it, when the
+scheduler wants to pop a pod, the new logic will go through the sub-queues, get the most under-used sub-queue
+(by comparing the `target resources by ratio`/`allocated resources` of every sub-queue) and pop one pod from it.
+Instead of comparing all the pending pods, we only need to compare the sub-queues here, which is much faster.
+The most important benefit for the developers is that they have full control of this logic, they can update the
+logic at anytime at will, needn't talk with k8s community at all.
 
-Another alternative is to suspend a job and don't let it go unless the resource
-sharing policy will not be broken. The drawback is that all the pods in the job will
-be impacted, it cannot handle the case that only some pods need to be touched.
+An alternative is to enhance the current queue to meet this request, resorting all the pending pods when trying
+to get a pod to pop. The drawback is that this special change is only for the multi-tenancy case, and we need a way
+to control the impact (E.g., add a flag), the users who needn't multi-tenancy support cannot benefit from it.
+There will be many similar requests in the future as different customers have different requests, to meet all the
+requests, the changes will make the logic of internal queue complex and hard to maintain.
 
-Another alternative is to make the pod that will break the resource sharing policy with
-`Unschedulable`. The drawback is that making such a decision needs to know whether other
-pods can run or not, the logic is complex, and the performance will be impacted.
+Another alternative is to have a high level controller to suspend a job and don't let it go unless the resource
+sharing policy will not be broken, this is a good solution to support multi-tenancy case at job level, and needn't
+a KEP to update the current design. The drawback is that all the pods in a job will be impacted by this way, it
+cannot handle the case that only some pods in a job need to be touched. The controller needs to predict whether
+a job can run or not before resuming a job, or the resource sharing may be broken, and preemption need to be
+triggered to re-balance the resource sharing, some workload will be interrupted, this is another drawback.
+
+Another alternative is to make the pod that will break the resource sharing policy with `Unschedulable` before
+binding it with a node. This is the ideal solution as it is very clean and no negative impact. The drawback is
+that making such a decision needs to know whether other pods can run or not, we need a whole picture of all the
+pods and nodes, the relations for pods-to-pods, pods-to-nodes and nodes-to-nodes, and how to adjust the
+decision once there are changes (E.g., a node cannot be accessed anymore), the logic is complex, and the
+performance should be low when there are thousands of pods and nodes.
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -119,49 +119,65 @@ But the default scheduler works as before and thus won't be impacted.
 
 ## Design Details
 
+The current internal queue is coupled with `cache`, `Option` and other implementations,
+this solution will not decouple them as this has no impact on the goal of custom queue.
+The coupled implementations will be exposed to users as custom queue needs to access them.
+
+To implement a custom queue, what an user needs to do is to copy the codes of the current
+internal queue (e.g., scheduling_queue.go, heap.go), update some logics to get the custom
+queue, and then pass the custom queue to the scheduler. With this way, user's logics will
+override the logics of the current internal queue.
+
 The codes will be updated as following:
 
-1. User implements a custom queue with the interface SchedulingQueue,
-the name is myQueue, and registers it
+1. Make the interface `SchedulingQueue` as a public one, so that users can implement it
+for the custom queue.
+
+2. Provide a new function `WithCustomQueue`, users can register the custom queue with this
+function like other plugins.
 
     ```go
-        command := app.NewSchedulerCommand(
-            app.WithPlugin(coscheduling.Name, coscheduling.New),
-            ......
-          + app.WithCustomQueue(myQueue.New),
-        )
+         command := app.NewSchedulerCommand(
+             app.WithPlugin(coscheduling.Name, coscheduling.New),
+             ......
+       +     app.WithCustomQueue(customQueue.New),
+         )
     ```
 
-2. The registered custom queue will be recorded by updated Registry struct.
+3. The registered custom queue will be recorded by updated `Registry` struct.
 
     ```go
       type Registry struct {
-        Pf          map[string]PluginFactory
-        CustomQueue schedulingQueue.SchedulingQueue
+          Pf          map[string]PluginFactory
+          CustomQueue schedulingQueue.SchedulingQueue
       }
     ```
 
-3. In function (c *Configurator) create(), pass the custom queue to the scheduler
-if it exists.
+4. In function `(c *Configurator) create()`, custom queue will be passed to the scheduler
+if it exists, or the current internal queue will be used.
 
     ```go
-      if c.registry.CustomQueue != nil {
-          podQueue = c.registry.CustomQueue
-      } else {
-          podQueue = schedulingQueue.NewSchedulingQueue()
-      }
+      - podQueue := internalqueue.NewSchedulingQueue(
+      -         lessFn,
+      +
+      + var podQueue framework.SchedulingQueue
+      + if c.registry.CustomQueue != nil {
+      +         podQueue = c.registry.CustomQueue
+      + } else {
+      +         podQueue = &internalqueue.PriorityQueue{}
+      + }
+      + podQueue.Init(...)
     ```
 
-4. Basic inputs for queue's initialization will not be changed, a new interface `Init` is added in
-SchedulingQueue, users must implement it.
+5. Basic inputs for queue's initialization will not be changed, a new method `Init` is added in
+the interface `SchedulingQueue`, users must implement it.
 
     ```go
-    -func NewPriorityQueue(
-    +func (p *PriorityQueue) Init(
-        lessFn framework.LessFunc,
-        informerFactory informers.SharedInformerFactory,
-        opts ...Option,
-    )
+      + func (p *PriorityQueue) Init(
+      +        lessFn framework.LessFunc,
+      +        informerFactory informers.SharedInformerFactory,
+      +        opts ...framework.Option,
+      + ) {
     ```
 
 ### Test Plan
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -131,7 +131,7 @@ manner.
 
 A practical design for phase 1 is described as below:
 
-1. Make `SchedulingQueue` a public interface. Users can choose to implement their own queue.
+1. Make `SchedulingQueue` a public interface which is private now in ([scheduling_queue.go](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126)). Users can choose to implement their own queue.
 
 2. Provide a new function `WithCustomQueue`, so users can register the custom queue with this
 function like other plugins.
@@ -144,32 +144,10 @@ function like other plugins.
          )
     ```
 
-3. The registered custom queue will be recorded by updated `Registry` struct.
-
-    ```go
-      type Registry struct {
-          Pf          map[string]PluginFactory
-          CustomQueue schedulingQueue.SchedulingQueue
-      }
-    ```
-
-4. In function `(c *Configurator) create()`, custom queue will be passed to the scheduler
+3. The registered custom queue will be passed to the scheduler
 if it exists, or the current internal queue will be used.
 
-    ```go
-      - podQueue := internalqueue.NewSchedulingQueue(
-      -         lessFn,
-      +
-      + var podQueue framework.SchedulingQueue
-      + if c.registry.CustomQueue != nil {
-      +         podQueue = c.registry.CustomQueue
-      + } else {
-      +         podQueue = &internalqueue.PriorityQueue{}
-      + }
-      + podQueue.Init(...)
-    ```
-
-5. Basic inputs for queue's initialization will not be changed, a new method `Init` is added in
+4. Basic inputs for queue's initialization will not be changed, a new method `Init` is added in
 the interface `SchedulingQueue`, users must implement it.
 
     ```go
@@ -179,12 +157,6 @@ the interface `SchedulingQueue`, users must implement it.
       +        opts ...framework.Option,
       + ) {
     ```
-
-To get started with the implementation, an easy way is to copy the codes of the current
-internal queue (i.e., scheduling_queue.go), and then update the logic in-place to adapt to
-the business needs (it's very like how some out-of-tree PostFilter plugin refers to the
-defaultpreemption plugin). With the custom queue handle, users can pass it to the scheduler like the
-plugins. Following this way, the user's custom scheduler queue will override the upstream internal queue.
 
 ### Test Plan
 
@@ -201,6 +173,7 @@ plugins. Following this way, the user's custom scheduler queue will override the
 
 #### Beta -> GA Graduation
 
+- Only necessary interface and structs are exposed to users.
 - Allowing time for feedback to ensure that the new interface sufficiently expresses users requirements.
 
 ### Upgrade / Downgrade Strategy
@@ -396,3 +369,11 @@ Why should this KEP _not_ be implemented?
 An alternative is to enhance the current queue to meet some requests. The drawback
 is that we cannot handle all the potential requests as the effort is big, we should
 give the ball to the users, and they can do what they want.
+
+Another alternative is to suspend a job and don't let it go unless the resource
+sharing policy will not be broken. The drawback is that all the pods in the job will
+be impacted, it cannot handle the case that only some pods need to be touched.
+
+Another alternative is to make the pod that will break the resource sharing policy with
+`Unschedulable`. The drawback is that making such a decision needs the status of all other
+pods, the logic is complex, and the performance will be impacted.
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -1,0 +1,423 @@
+# KEP-2779: support custom queue for the scheduler
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests for meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Provide an interface so that user can implement a custom queue of scheduler,
+with such a queue, user can sort the pods in the queue with specified rules 
+and get some functions that are not supported now.
+
+## Motivation
+The current queue (PriorityQueue in pkg\scheduler\internal\queue\scheduling_queue.go) 
+for pod selection in the scheduler is good enough for most cases, while it cannot handle 
+all the requests even with a custom scheduler based on the kubernetes scheduler plugins.
+
+One example:
+There are many pods from 2 users (userA and userB), and the target is to ensure resource
+usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
+the resource usage in Less function cannot work, because the resource usage for 
+userA/userB will be updated dynamically once a pod is selected and bound, while the 
+algorithm (heapsort) that used by the current queue to get the next pods depends on static
+data.
+
+Another example:
+The interval of function flushUnschedulableQLeftover is hard coded to 30 seconds, user
+cannot change it.
+
+These issues can be resolved if user can implement the queue by themselves, then user
+can get what they want easily, and code maintainer needn't handle various requests.
+
+### Goals
+
+Support custom queue of the scheduler.
+
+### Non-Goals
+
+Support custom rules for pod selection in the current queue of scheduler
+
+## Proposal
+
+Provide an interface like current scheduler plugins, users can provide a
+custom queue of scheduler and register it, then the kubernetes
+scheduler will use this custom queue for pod selection.
+
+This is an extension of the current scheduler plugins, users can control
+more details of the scheuler with this enhancement.
+
+Pros:  
+Users can get full control of the pod selection logic in the queue, they
+can select pods with custom logic.
+
+The scheduler plugins design is enhanced, the custom queue and the extension
+points can work together to meet more requests.
+
+Cons:  
+Users need to understand the details of the current queue which has 13
+functions now, it is not easy to implement a custom queue.
+
+### Risks and Mitigations
+
+If custom queue is not enabled or provided, the current queue will work as
+before, no other users will be impacted.
+
+## Design Details
+
+The codes will be updated as following:
+
+1. User implements a custom queue with the interface SchedulingQueue,
+the name is myQueue, and registers it
+
+  ```
+	command := app.NewSchedulerCommand(
+		app.WithPlugin(coscheduling.Name, coscheduling.New),
+                        …...
+	+	app.WithCustomQueue(myQueue.New),
+	)
+  ```
+
+2. The registered custom queue will be recorded by updated Registry struct.
+
+  ```
+  type Registry struct {
+    Pf          map[string]PluginFactory
+    CustomQueue schedulingQueue.SchedulingQueue
+  }
+  ```
+
+3. In function (c *Configurator) create(), pass the custom queue to the scheduler
+if it exists.
+  ```
+  if c.registry.CustomQueue != nil {
+      podQueue = c.registry.CustomQueue
+  } else {
+      podQueue = schedulingQueue.NewSchedulingQueue()
+  }
+   ```
+
+### Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Consider the following in developing a test plan for this enhancement:
+- Will there be e2e and integration tests, in addition to unit tests?
+- How will it be tested in isolation vs with other components?
+
+No need to outline all of the test cases, just the general strategy. Anything
+that would count as tricky in the implementation, and anything particularly
+challenging to test, should be called out.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+### Graduation Criteria
+
+<!--
+**Note:** *Not required until targeted at a release.*
+
+Define graduation milestones.
+
+These may be defined in terms of API maturity, or as something else. The KEP
+should keep this high-level with a focus on what signals will be looked at to
+determine graduation.
+
+Consider the following in developing the graduation criteria for this enhancement:
+- [Maturity levels (`alpha`, `beta`, `stable`)][maturity-levels]
+- [Deprecation policy][deprecation-policy]
+
+Clearly define what graduation means by either linking to the [API doc
+definition](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning)
+or by redefining what graduation means.
+
+In general we try to use the same stages (alpha, beta, GA), regardless of how the
+functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+#### Alpha
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- Gather feedback from developers and surveys
+- Complete features A, B, C
+- Additional tests are in Testgrid and linked in KEP
+
+#### GA
+
+- N examples of real-world usage
+- N installs
+- More rigorous forms of testing—e.g., downgrade tests and scalability tests
+- Allowing time for feedback
+
+**Note:** Generally we also wait at least two releases between beta and
+GA/stable, because there's no opportunity for user feedback, or even bug reports,
+in back-to-back releases.
+
+**For non-optional features moving to GA, the graduation criteria must include
+[conformance tests].**
+
+[conformance tests]: https://git.k8s.io/community/contributors/devel/sig-architecture/conformance-tests.md
+
+#### Deprecation
+
+- Announce deprecation and support policy of the existing flag
+- Two versions passed since introducing the functionality that deprecates the flag (to address version skew)
+- Address feedback on usage/changed behavior, provided on GitHub issues
+- Deprecate the flag
+-->
+
+### Upgrade / Downgrade Strategy
+
+N/A
+
+### Version Skew Strategy
+
+N/A
+
+## Production Readiness Review Questionnaire
+
+<!--
+
+Production readiness reviews are intended to ensure that features merging into
+Kubernetes are observable, scalable and supportable; can be safely operated in
+production environments, and can be disabled or rolled back in the event they
+cause increased failures in production. See more in the PRR KEP at
+https://git.k8s.io/enhancements/keps/sig-architecture/1194-prod-readiness.
+
+The production readiness review questionnaire must be completed and approved
+for the KEP to move to `implementable` status and be included in the release.
+
+In some cases, the questions below should also have answers in `kep.yaml`. This
+is to enable automation to verify the presence of the review, and to reduce review
+burden and latency.
+
+The KEP must have a approver from the
+[`prod-readiness-approvers`](http://git.k8s.io/enhancements/OWNERS_ALIASES)
+team. Please reach out on the
+[#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) channel if
+you need any help or guidance.
+-->
+
+### Feature Enablement and Rollback
+
+N/A
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Other
+  - Describe the mechanism: Restart the scheduler with/without custom queue.
+  - Will enabling / disabling the feature require downtime of the control
+    plane? Yes
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? Yes
+
+###### Does enabling the feature change any default behavior?
+
+Yes, the logic in queue of scheduler will be updated by user's logics.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+This feature will work as we described before. The history operations have no
+impact on the behavior.
+
+###### Are there any tests for feature enablement/disablement?
+
+N/A
+
+### Rollout, Upgrade and Rollback Planning
+
+N/A
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+N/A
+
+###### What specific metrics should inform a rollback?
+
+N/A
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+N/A
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+This is a feature of scheduler, and operator can decide which scheduler will
+be used.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] Events
+  - Event Reason: the custom queue can raise events for some actions
+- [ ] API .status
+  - Condition name: 
+  - Other field: 
+- [ ] Other (treat as last resort)
+  - Details:
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+The performance is only impacted by the logics in custom queue, not impacted by the
+design structure.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- [ ] Metrics
+  - Metric name:
+  - [Optional] Aggregation method:
+  - Components exposing the metric:
+- [x] Other (treat as last resort)
+  - Details: the performance is only impacted by the custom queue's logic.
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+No
+
+### Dependencies
+
+N/A
+
+###### Does this feature depend on any specific services running in the cluster?
+
+No
+
+### Scalability
+
+N/A
+
+###### Will enabling / using this feature result in any new API calls?
+
+No
+
+###### Will enabling / using this feature result in introducing new API types?
+
+No
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+No
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+No
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any components?
+
+No
+
+### Troubleshooting
+
+The user who provides custom queue is responsible for providing troubleshooting messages/events.
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+It depends on the related logics in custom queue.
+
+###### What are other known failure modes?
+
+No
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+Using the current queue as a workaround and check the logics in custom queue.
+
+## Implementation History
+
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
+
+An alternative is to enhance the current queue to meet some requests. The drawback
+is that we cannot handle all the potentail requests as the effort is big, we should 
+give the ball to the users, and they can do what they want.
--- a/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
+++ b/keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
@@ -61,32 +61,32 @@ the scheduler seamlessly. This can satisfy the business needs such as sophistica
 pods sorting and internal queuing mechanics.
 
 ## Motivation
-The current internal queue implementation ([scheduling_queue.go#PriorityQueue]
-(https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126))
+
+The current internal queue implementation ([scheduling_queue.go#PriorityQueue](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126))
 works for most cases, but the limited interfaces exposure makes it hard to
 implement a scheduler plugin for sophisticated requirements.
 
 One requirement in terms of multi-tenancy support:  
 There are many pods from 2 users (userA and userB), and the target is to ensure resource
 usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
-the resource usage in `Less` function doesn't work properly, because the resource usage for 
-userA/userB will be updated dynamically once a pod is selected and bound, while the 
+the resource usage in `Less` function doesn't work properly, because the resource usage for
+userA/userB will be updated dynamically once a pod is selected and bound, while the
 algorithm (heapsort) that used by the current queue to get the next pods depends on static
 data.
 
-Another requirement in terms of custom function support:  
-For example, the interval of function flushUnschedulableQLeftover cannot be updated dynamically, while
-user want it to be shorter or longer at different times as shorter interval can get pods to be scheduled
-faster and longer one can get less logs flushed.
+Other requirements come from current semi-exposed functions/parameters:
+For example, the interval of function flushUnschedulableQLeftover, although configurable,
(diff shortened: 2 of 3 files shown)


keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+The codes will be updated as following:
+
+1. User implements a custom queue with the interface SchedulingQueue,
+the name is myQueue, and registers it
+
+  ```
+	command := app.NewSchedulerCommand(
+		app.WithPlugin(coscheduling.Name, coscheduling.New),
+                        …...
+	+	app.WithCustomQueue(myQueue.New),
```
denkensk, line 126: I think It maybe  better to make the CustomQueue as a plugin and register it as other plugins like:
```go
	command := app.NewSchedulerCommand(
		app.WithPlugin(coscheduling.Name, coscheduling.New),
	+	app.WithPlugin(myQueue.Name, myQueue.New),
)

**Review by denkensk — COMMENTED** [review]
It's a particularly good idea.

Some suggestion:
1. It may be better to make the queue as the new extension point of scheduling framework. Then the users can implement their custom queue as the plugins and register or construct like other plugins.
2. Can you give a brief description of the interface to be implemented in custom queue?

**Comment by suigh** [comment]
@denkensk 
Thanks for your suggestions, please see my answers here

`It may be better to make the queue as the new extension point of scheduling framework. Then the users can implement their custom queue as the plugins and register or construct like other plugins.`
My first idea is like this one, and I switched to the current one as 
1. we store the registered plugin in the following struct
type Registry struct {
	map[string]PluginFactory
}
type PluginFactory = func(configuration runtime.Object, f framework.Handle) (framework.Plugin, error)

`PluginFactory ` is not the expected function to create a custom queue as the queue needn't the parameters in it

2. With this solution, we will have to identify the input's type  -- for plugin or for queue somewhere, and then assign the custom queue to the scheduler. This check is not straightforward, add a new function like `WithCustomQueue` seems better.


`Can you give a brief description of the interface to be implemented in custom queue?`
User will implement the interface `SchedulingQueue` for the current queue. User can just copy the `pkg\scheduler\queue\scheduling_queue.go` and update something. I will add an example into the  SIG_scheduling after updating the k8s codes.

`Some concerns:
It maybe difficult for users to implement a complete custom queue. Is it considered that only part of the interface is exposed like pop or add, or the data structure like heap?`
Considered this before, and choose the current one because:
1. by this way, the solution will be complex than the current one, and when user ask for more functions, we will have to update the codes again.
2. for an user who only want to update one or several functions, the user can just copy the `scheduling_queue.go` and update it, and left other functions there., the effort is same as this way.

File: keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md (hunk 1 above)
Comment (suigh) [reply keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md:126]: Thanks for the suggestion, please see my answer below, it explains why `WithPlugin` is not re-used.

File: keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md [hunk 2]
```diff
@@ -0,0 +1,423 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Provide an interface so that user can implement a custom queue of scheduler,
```
Huang-Wei, line 58:
```suggestion
Provide an interface so that users can implement a custom queue of scheduler.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+with such a queue, user can sort the pods in the queue with specified rules 
+and get some functions that are not supported now.
```
Huang-Wei, line 60:
```suggestion
With such, users can build their custom queue implementation at will, and plumb into
the scheduler seamlessly. This can satisfy the business needs such as sophisticated
pods sorting and internal queuing mechanics.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Motivation
+The current queue (PriorityQueue in pkg\scheduler\internal\queue\scheduling_queue.go)
```
Huang-Wei, line 63:
```suggestion
The current internal queue implementation ([scheduling_queue.go#PriorityQueue](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126)).
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+## Summary
+
+Provide an interface so that user can implement a custom queue of scheduler,
+with such a queue, user can sort the pods in the queue with specified rules 
+and get some functions that are not supported now.
+
+## Motivation
+The current queue (PriorityQueue in pkg\scheduler\internal\queue\scheduling_queue.go) 
+for pod selection in the scheduler is good enough for most cases, while it cannot handle 
+all the requests even with a custom scheduler based on the kubernetes scheduler plugins.
```
Huang-Wei, line 65:
```suggestion
works for most cases, but the limited interfaces exposure makes it hard to
implement a scheduler plugin for sophisticated requirements.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+One example:
```
Huang-Wei, line 67:
```suggestion
One requirement in terms of multi-tenancy support:
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+There are many pods from 2 users (userA and userB), and the target is to ensure resource
+usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
+the resource usage in Less function cannot work, because the resource usage for
```
Huang-Wei, line 70:
```suggestion
the resource usage in `Less` function doesn't work properly, because the resource usage for 
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+There are many pods from 2 users (userA and userB), and the target is to ensure resource
+usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
+the resource usage in Less function cannot work, because the resource usage for 
+userA/userB will be updated dynamically once a pod is selected and bound, while the 
+algorithm (heapsort) that used by the current queue to get the next pods depends on static
+data.
+
+Another example:
+The interval of function flushUnschedulableQLeftover is hard coded to 30 seconds, user
+cannot change it.
```
Huang-Wei, line 77: It's now customizable, so no longer a valid motivation.

Huang-Wei: So, can you compile at least one more motivation here?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+These issues can be resolved if user can implement the queue by themselves, then user
+can get what they want easily, and code maintainer needn't handle various requests.
```
Huang-Wei, line 80:
```suggestion
Given that the business needs may vary greatly, it'd be good to provide a replaceable
`SchedulerQueue` interface and plumbing mechanics. So that the developers can
implement their custom queue implementation, while the upstream maintainers focus
on keeping the core small and extensible, and thus only maintain the internal queue piece.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Goals
+
+Support custom queue of the scheduler.
```
Huang-Wei, line 84: Support custom queue implementation of the scheduler

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Non-Goals
+
+Support custom rules for pod selection in the current queue of scheduler
```
Huang-Wei, line 88:
```suggestion
Support custom rules for pod selection in the current internal queue of scheduler.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Proposal
+
+Provide an interface like current scheduler plugins, users can provide a
+custom queue of scheduler and register it, then the kubernetes
+scheduler will use this custom queue for pod selection.
```
Huang-Wei, line 94:
```suggestion
scheduler will use this custom queue for pod management.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+This is an extension of the current scheduler plugins, users can control
+more details of the scheuler with this enhancement.
+
+Pros:  
+Users can get full control of the pod selection logic in the queue, they
```
Huang-Wei, line 100:
```suggestion
Users can get full control of the pod queuing logic, they
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+can select pods with custom logic.
```
Huang-Wei, line 101:
```suggestion
can pop/re-queue/backoff pods with custom logic.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+points can work together to meet more requests.
+
+Cons:  
+Users need to understand the details of the current queue which has 13
+functions now, it is not easy to implement a custom queue.
+
+### Risks and Mitigations
+
+If custom queue is not enabled or provided, the current queue will work as
+before, no other users will be impacted.
```
Huang-Wei, line 113: A poor-implemented queue may not function well, in both functionality and performance.
But the default scheduler works as before and thus won't be impacted.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+
+The codes will be updated as following:
+
+1. User implements a custom queue with the interface SchedulingQueue,
+the name is myQueue, and registers it
+
+  ```
+	command := app.NewSchedulerCommand(
+		app.WithPlugin(coscheduling.Name, coscheduling.New),
+                        …...
```
Huang-Wei, line 125:
```suggestion
                        ......
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+before, no other users will be impacted.
+
+## Design Details
+
+The codes will be updated as following:
+
+1. User implements a custom queue with the interface SchedulingQueue,
+the name is myQueue, and registers it
+
+  ```
```
Huang-Wei, line 122:
```suggestion
```go
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+  if c.registry.CustomQueue != nil {
+      podQueue = c.registry.CustomQueue
+  } else {
+      podQueue = schedulingQueue.NewSchedulingQueue()
+  }
+   ```
+
+### Test Plan
+
```
Huang-Wei, line 150: Please refer to other KEPs to fill this section as well as others.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,423 @@
...
+3. In function (c *Configurator) create(), pass the custom queue to the scheduler
+if it exists.
+  ```
+  if c.registry.CustomQueue != nil {
+      podQueue = c.registry.CustomQueue
+  } else {
+      podQueue = schedulingQueue.NewSchedulingQueue()
+  }
+   ```
+
```
Huang-Wei, line 148: Can you also add here to explain how it explains with other internal scheduler structures like Cache, Assumed state, NominatedPods, etc. Any other interfaces are needed? and how. What's the risk here?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.22"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.22"
+  beta: "v1.23"
+  stable: "v1.24"
```
Huang-Wei, line 30: We won't be able to finish in alpha in this release:

```suggestion
  alpha: "v1.23"
  beta: "v1.24"
  stable: "v1.26"
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: support custom queue for the scheduler
+    components:
+      - kube-scheduler
+disable-supported: false
+
+# The following PRR answers are required at beta release
+metrics:
+  - my_feature_metric
```
Huang-Wei, line 42: We can follow up with PRR later.

Huang-Wei (commented):
Thanks @suigh. Here is the first round of review.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Update the message to:

Another requirement in terms of custom function support:  
For example, the interval of function flushUnschedulableQLeftover cannot be updated dynamically, while
user want it to be shorter or longer at different times as shorter interval can get pods to be scheduled
faster and longer one can get less logs flushed.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: I removed the space. Why adding `go`? A typo?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Updated with 

```
- **Unit Tests**: All core changes must be covered by unit tests.
- **Integration Tests**: One integration test for the custom queue.
- **Benchmark Tests**: The performance benchmark test result is same as before if custom queue is not used.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: A new interface Init will be aded to handle the initialization parameters from the k8s framework, it is used to replace the old function NewPriorityQueue, the logic inside them are same.
Packages queue and heap will be exposed to end users, then users can access/use them and the mentioned internal scheduler structures. 


```
4. Basic inputs for queue's initialization will not be changed, a new interface `Init` is added in
SchedulingQueue, users must implement it.
-func NewPriorityQueue(
+func (p *PriorityQueue) Init(
 	lessFn framework.LessFunc,
 	informerFactory informers.SharedInformerFactory,
 	opts ...Option,
)
5. For the scheduler internal packages under ([internal] (https://github.com/kubernetes/kubernetes/tree/b6c75bee15e150628fcc240ab32dba6190d254e4/pkg/scheduler/internal)), queue and heap will be moved out of this folder so that users can access and use the structures inside them, cache and parallelize will not be touched.
```

suigh:
@Huang-Wei 
Thanks very much for the detailed review, I had updated the documents, please help review the new versions.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
Huang-Wei: Yeah, that's a fair point.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
Huang-Wei: that's for pretty formatting.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
Huang-Wei: > Packages queue and heap will be exposed to end users, then users can access/use them and the mentioned internal scheduler structures.

I don't think I like this idea. The internal structure should be restricted for internal usage only.

Can you list the blockers (due to missing plumbing mechanics) that a pure out-of-tree implementation won't work properly?

Also, I'm also thinking about how `Option` can be more general and thus more usable for an out-of-tree queue implementation. Right now the internal `priorityQueueOptions` are fixed data structures, and hence not extensible.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -119,52 +124,52 @@ The codes will be updated as following:
...
+ 	lessFn framework.LessFunc,
+ 	informerFactory informers.SharedInformerFactory,
+ 	opts ...Option,
+)
+```
 
-No need to outline all of the test cases, just the general strategy. Anything
-that would count as tricky in the implementation, and anything particularly
-challenging to test, should be called out.
+5. For the scheduler internal packages under ([internal] (https://github.com/kubernetes/kubernetes/tree/b6c75bee15e150628fcc240ab32dba6190d254e4/pkg/scheduler/internal)), queue and heap will be moved out of this folder so that users can access and use the structures inside them, cache and parallelize will not be touched.
```
Huang-Wei, line 166: remove the blank between text and URL link.

Again, I don't think it's feasible to move queue and heap out.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Thanks for this suggestion. 
Downloaded an extention for markdownlint to ensure the markdown format can be followed.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: OK, let's keep queue and heap in the original location.  Removed `5` from the document.

The current blockers in my mind are `SchedulingQueue` and `PriorityQueueOptions` (maybe there will be more, I will write some codes to verify it), and I'm planning to move them to `pkg\scheduler\framework\interface.go` so that the users can access them, this is a right loaction for you?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Item `5` was removed, queue and heap will be internal packages as before.

suigh: The current blockers in my mind are SchedulingQueue and PriorityQueueOptions (maybe there will be more, I will write some codes to verify it), and I'm planning to move them to pkg\scheduler\framework\interface.go so that the users can access them, this is a right loaction for you?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: close this thread as there is same thread below.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Checked with codes, following items in the current queue need to be exposed, and I'm planning to move them into `pkg\scheduler\framework\interface.go`

```
type PreEnqueueCheck func(pod *v1.Pod) bool

type PriorityQueueOptions struct {}

type Option func(*PriorityQueueOptions)

type SchedulingQueue interface {}
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
Huang-Wei: Please don't fold leave unresolved comments.

suigh: Oh, sorry. I misunderstand this comment, I think there will be another round review at `beta release` ....

suigh:
@Huang-Wei 
Updated doc as per the discussion in section `Design Details`, please help review. :)

What is updated:
1: for the coupled items (e.g., cache, option) with internal queue, this solution will not decouple them, we will just expose the coupled ones that need to be accessed by the custom queue. This can make the solution simple, and we can refine the codes in the future if necessary.
2: to get a custom queue, an user can copy the `scheduling_queue.go` and update it, then register to the scheduler, this is easy and straightforward.
3: update some previous messages to make the logic clear.

suigh:
some update:
1. document was updated at `Design Details` section, added more details about what will do and what will not do. The current design will not provide perfect interface as it is complex and we already can get what we want by the current design.

2. tried this solution by implementing a custom queue ( get a file named scheduling_queue.go, implemented similiar logic in scheduling_queue.go) and registered it like other plugins, the result is OK, pods were scheduled by the logic in customer quque, this solution works.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+functions now, it is not easy to implement a custom queue.
+
+### Risks and Mitigations
+
+A poor-implemented queue may not function well, in both functionality and performance.
+But the default scheduler works as before and thus won't be impacted.
+
+## Design Details
+
+The current internal queue is coupled with `cache`, `Option` and other implementations
```
Huang-Wei, line 122: The scheduler _internal_ queue, as its name implies, was originally implemented for internal usage. Due to that, some structs (like `cache`, `Option`) are not designed for external extension, and the internal queue is initialized with specified parameters like `LessFunc`, `SharedInformerFactory`,
and `Option` by [factory.go#create](https://github.com/kubernetes/kubernetes/blob/f1f0183d2bbcde33024b2a05d6f39df32f11e037/pkg/scheduler/factory.go#L172)). Although refactoring the internal queuing mechanics to be reusable/extensible is the eventual goal, we'd like to approach that gradually, and thus is not the goal of phase 1 of this design. In phase 1, we'd like to exercise the idea of "a replaceable scheduler queue" in a "minimal viable product" manner.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+Given that the business needs may vary greatly, it'd be desirable to provide a replaceable
+`SchedulerQueue` interface and plumbing mechanics. So that the developers can
+implement their custom queue implementation, while the upstream maintainers focus
+on keeping the core small and extensible, and thus only maintain the internal queue piece.
+
+### Goals
+
+Support custom queue implementation of the scheduler.
+
```
Huang-Wei, line 90: Add a bullet: Roll out the design gradually. In phase 1, keep current internal queue implementation intact.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+Some of the items like `SchedulingQueue`, `Option` that are only used by the internal
+queue now will be exposed at [interface.go](https://github.com/kubernetes/kubernetes/blob/fb9cafc99be94a73d9b92545164dbf336bbd230a/pkg/scheduler/framework/interface.go),
+users of custom queue can access them.
+
+To implement a custom queue, what an user needs to do is to copy the codes of the current
+internal queue (e.g., scheduling_queue.go), update some logics to get the custom queue,
+and then pass the custom queue to the scheduler like the plugins. With this way, user's
+logics will override the logics of the current internal queue.
+
+The codes will be updated as following:
```
Huang-Wei, line 137: A practical design for phase 1 is described as below:

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+1. Make the interface `SchedulingQueue` as a public one, so that users can implement it
```
Huang-Wei, line 139:
```suggestion
1. Make `SchedulingQueue` a public interface. Users can choose to implement their own queue.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+for the custom queue.
+
+2. Provide a new function `WithCustomQueue`, users can register the custom queue with this
```
Huang-Wei, line 142:
```suggestion
2. Provide a new function `WithCustomQueue`, so users can register the custom queue with this
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+and such logics can be enhanced in the future on demand.
+
+Some of the items like `SchedulingQueue`, `Option` that are only used by the internal
+queue now will be exposed at [interface.go](https://github.com/kubernetes/kubernetes/blob/fb9cafc99be94a73d9b92545164dbf336bbd230a/pkg/scheduler/framework/interface.go),
+users of custom queue can access them.
+
+To implement a custom queue, what an user needs to do is to copy the codes of the current
+internal queue (e.g., scheduling_queue.go), update some logics to get the custom queue,
+and then pass the custom queue to the scheduler like the plugins. With this way, user's
+logics will override the logics of the current internal queue.
```
Huang-Wei, line 135: Remove these 2 paragraphs as they are custom implementation details (or suggestions).

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+the interface `SchedulingQueue`, users must implement it.
+
+    ```go
+      + func (p *PriorityQueue) Init(
+      +        lessFn framework.LessFunc,
+      +        informerFactory informers.SharedInformerFactory,
+      +        opts ...framework.Option,
+      + ) {
+    ```
+
```
Huang-Wei, line 188: Alternatively, you can ass a stanza about the implementation practice/suggestion here:

```
To get started with the implementation, an easy way is to copy the codes of the current
internal queue (i.e., scheduling_queue.go), and then update the logic in-place to adapt to
your business needs  (it's very like how some out-of-tree PostFilter plugin refers to the defaultpreemption plugin).
With the custom queue handle, you can pass it to the scheduler like the plugins.
Following this way, the user's custom scheduler queue will override the upstream internal queue.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+### Test Plan
+
+- **Unit Tests**: All core changes must be covered by unit tests.
+- **Integration Tests**: One integration test for the custom queue.
```
Huang-Wei, line 192:
```suggestion
- **Integration Tests**: At least one integration test to craft a custom queue to exercise an end-to-end flow.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+- **Benchmark Tests**: The performance benchmark test result is same as before if custom queue is not used.
+
+### Graduation Criteria
+
```
Huang-Wei, line 196: Describe the plan of how to roll it out here. Like phase 1, phase 2, etc.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+functionality is accessed.
+
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+#### Alpha
+
+- Feature implemented behind a feature flag
```
Huang-Wei, line 224: I guess we don't need a feature gate here. But we can mention the maturity level by using the terms alpha, beta, etc.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,449 @@
...
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- Gather feedback from developers and surveys
+- Complete features A, B, C
```
Huang-Wei, line 230: Polish this as well as other uncompleted ones.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: updated.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: added.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: removed.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: added.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Updated as following, is this OK?
```
### Graduation Criteria

#### Alpha -> Beta Graduation

- Users can implement their own custom queues.
- No user complaints regarding correctness.

#### Beta -> GA Graduation

- Allowing time for feedback to ensure that the new interface sufficiently expresses users requirements.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Removed this part, and add some details about the `Graduation Criteria`, mentioned by above answers.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: This document is for the solution of phase 1 (provide interface to implement custom queue), and there is no solution details for phase 2 (refine the codes, decouple the queue with some structures), should we mention the `Graduation Criteria` of phase 2 in this document?

Huang-Wei:
@suigh LGTM generally. Please leave some time for other reviewers to take a look.

alculquicondor:
I have this in my queue. I will take a look later this week or next week.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
```diff
@@ -0,0 +1,42 @@
+title: support custom queue for the scheduler
+kep-number: 2779
+authors:
+  - "@suigh"
+owning-sig: sig-scheduling
+status: provisional
+creation-date: 2021-06-09
+reviewers:
+  - "@Huang-Wei"
+  - TBD
```
alculquicondor, line 10: Please add me

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+see-also:
+  - "/keps/sig-scheduling/624-scheduling-framework"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.22"
```
alculquicondor, line 24: 1.23

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.23"
+  beta: "v1.24"
+  stable: "v1.26"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: support custom queue for the scheduler
```
alculquicondor, line 35: I suppose we don't need a feature gate for this feature, as it wouldn't change behavior.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
alculquicondor: Say TDB instead.

keps/prod-readiness/sig-scheduling/2779.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2779
+beta:
```
alculquicondor, line 2: alpha

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+Given that the business needs may vary greatly, it'd be desirable to provide a replaceable
+`SchedulerQueue` interface and plumbing mechanics. So that the developers can
+implement their custom queue implementation, while the upstream maintainers focus
+on keeping the core small and extensible, and thus only maintain the internal queue piece.
+
+### Goals
+
+1. Support custom queue implementation of the scheduler.
+2. Roll out the design gradually. In phase 1, keep current internal queue implementation
```
alculquicondor, line 90: Please clarify: is it a goal to change the default queueing plugin? If not, please add a Non Goal. Regardless, point 2 needs is not very clear on the intent.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Provide an interface so that users can implement a custom queue of scheduler.
```
alculquicondor, line 58: which kind of user? end user or custom scheduler developers?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+With such, users can build their custom queue implementation at will, and plumb into
```
alculquicondor, line 59: which scheduler? the default scheduler? IIUC, the intent is to easy the development of custom schedulers.

The phrasing of this paragraph makes it seem that you are proposing an API to configure the queue at runtime, which is not the case.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+There are many pods from 2 users (userA and userB), and the target is to ensure resource
+usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
+the resource usage in `Less` function doesn't work properly, because the resource usage for
+userA/userB will be updated dynamically once a pod is selected and bound, while the
+algorithm (heapsort) that used by the current queue to get the next pods depends on static
+data.
+
+Other requirements come from current semi-exposed functions/parameters:
+For example, the interval of function flushUnschedulableQLeftover, although configurable,
+is immutable after it gets initialized. A requirement is to adjust the interval dynamically
```
alculquicondor, line 79: What is the use case for this?

The intend of the flush is to make sure we are not leaving pods without scheduling retries. Ideally, it shouldn't exist and retries should entirely be based on cluster events. But realistically, it's hard to catch all events.
But depending on your use case, you might just want a mechanism to trigger the retry of *certain* pods.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+intact.
+
+### Non-Goals
+
+Support custom rules for pod selection in the current internal queue of scheduler.
+
+## Proposal
+
+Provide an interface like current scheduler plugins, users can provide a
+custom queue of scheduler and register it, then the kubernetes
```
alculquicondor, line 100: at which time? build time, configuration time, runtime?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+This is an extension of the current scheduler plugins, users can control
+more details of the scheduler with this enhancement.
+
+Pros:  
+Users can get full control of the pod queuing logic, they can pop/re-queue/backoff pods with
+custom logic.
+
+The scheduler plugins design is enhanced, the custom queue and the extension
+points can work together to meet more requests.
```
alculquicondor, line 111: s/requests/requirements

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+## Motivation
+
+The current internal queue implementation ([scheduling_queue.go#PriorityQueue](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126))
+works for most cases, but the limited interfaces exposure makes it hard to
+implement a scheduler plugin for sophisticated requirements.
+
+One requirement in terms of multi-tenancy support:  
+There are many pods from 2 users (userA and userB), and the target is to ensure resource
+usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1. Sorting the pods by
+the resource usage in `Less` function doesn't work properly, because the resource usage for
```
alculquicondor, line 72: Can you share a sample design of how you would implement a queue that satisfies this requirement?

Perhaps we can put the interface in a different place so that you don't have to implement the entire queue. I would like to see Alternative proposals.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+    ```go
+         command := app.NewSchedulerCommand(
+             app.WithPlugin(coscheduling.Name, coscheduling.New),
+             ......
+       +     app.WithCustomQueue(customQueue.New),
+         )
+    ```
+
+3. The registered custom queue will be recorded by updated `Registry` struct.
```
alculquicondor, line 147: This sounds like implementation detail that can be left for code reviews. I would probably ask to not touch the registry. But again, we don't need to discuss this in the KEP.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+    ```go
+      type Registry struct {
+          Pf          map[string]PluginFactory
+          CustomQueue schedulingQueue.SchedulingQueue
+      }
+    ```
+
+4. In function `(c *Configurator) create()`, custom queue will be passed to the scheduler
```
alculquicondor, line 156: Also irrelevant for the KEP

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+usage. Due to that, some structs (like cache, Option) are not designed for external extension,
+and the internal queue is initialized with specified parameters like LessFunc,
+SharedInformerFactory, and Option by [factory.go#create](https://github.com/kubernetes/kubernetes/blob/f1f0183d2bbcde33024b2a05d6f39df32f11e037/pkg/scheduler/factory.go#L172). Although refactoring the internal queuing mechanics to be reusable/extensible is the eventual
+goal, we'd like to approach that gradually, and thus is not the goal of phase 1 of this design.
+In phase 1, we'd like to exercise the idea of "a replaceable scheduler queue" in a "minimal viable product"
+manner.
+
+A practical design for phase 1 is described as below:
+
+1. Make `SchedulingQueue` a public interface. Users can choose to implement their own queue.
```
alculquicondor, line 134: The interface definition should be part of the KEP

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+    ```go
+      + func (p *PriorityQueue) Init(
+      +        lessFn framework.LessFunc,
+      +        informerFactory informers.SharedInformerFactory,
+      +        opts ...framework.Option,
+      + ) {
+    ```
+
+To get started with the implementation, an easy way is to copy the codes of the current
```
alculquicondor, line 183: Irrelevant for the KEP

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
alculquicondor: I think you should have *some* idea of what phase 2 would look like before we proceed.

alculquicondor (commented):
As part of the alternatives, consider this:

(I'm assuming your use case is around batch, not serving applications).
Have you considered using a higher level controller and the Job.suspend feature to govern resource sharing? The idea is that you would suspend Jobs at creation. The Job would only be un-suspended (resumed) when the Pods for the Job wouldn't violate fairness or resource quotas or whatever policy you want to implement https://kubernetes.io/docs/concepts/workloads/controllers/job/#suspending-a-job

This solution is good for 2 reasons:
- There are no unschedulable Pods competing for scheduling time.
- The resume policy doesn't have to be based on the current size of the cluster, which makes sense when you have a cluster autoscaler.

denkensk:
> Have you considered using a higher level controller and the Job.suspend feature to govern resource sharing? The idea is that you would suspend Jobs at creation. The Job would only be un-suspended (resumed) when the Pods for the Job wouldn't violate fairness or resource quotas or whatever policy you want to implement

---

I like the idea. In fact, we have implemented higher level controller and used it in the production environment in Alibaba. We are preparing to open source the design and code. I'll invite you to help review. But this does not conflict with this kep. We pay more attention to the job level like tf-job mpi-job and so on. Maybe we can cooperate to solve some batch scheduling problems @Huang-Wei @alculquicondor @suigh

alculquicondor:
I would like tf-operator (and every batch API) to use k8s Jobs underneath so it's possible to suspend them.
See kubeflow/tf-operator#1303 for the ongoing discussion.

alculquicondor:
mpi-operator might be able to migrate to k8s Jobs sooner kubeflow/mpi-operator#373

denkensk:
Thanks for your information @alculquicondor 
We chose a similar way to express suspend by adding annotation here
https://github.com/kube-queue/api/blob/bffe1acb5aa961be9f171ff0fa2afa47f02f5381/pkg/apis/scheduling/v1alpha1/type.go#L99
And we also try to add the related processing logic to TF-operator and MPI-operator  @zw0610
But if you can change every batch API in the ecosystem) to use k8s Jobs underneath，It will bring us a big help

alculquicondor:
So back to the proposal in this KEP. I'm not convinced that the scheduling queue sorting is gonna guarantee any type of resource sharing, like this requirement listed in the motivation:

> There are many pods from 2 users (userA and userB), and the target is to ensure resource usage (E.g. userA gets x CPUs) ratio between userA and userB is 1:1.

The scheduling queue sorting is kind of like a heuristic about which pod should be tried next. But let's say you are able to implement the sorting based on users' usage ratio. We try to schedule Pod A and suddenly the API server was overwhelmed and couldn't accept the binding operation. Then Pod A is added back to the queue. But since binding happens in a separate routine, maybe we are already scheduling Pod B from another user. Then Pod B succeeds and you were not able to guarantee the ratio.

Then, what I think we should be doing is not care too much about the queue sorting. But then we can have a PreFilter (and perhaps a Permit) plugin that would return `Unschedulable` if such pod would violate the resource sharing.

If what I'm proposing doesn't make sense, I would like to see why in the "Alternatives" section of the KEP.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
suigh: added

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
suigh: updated to 1.23

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
suigh: comment out this part as suggetsed.

keps/prod-readiness/sig-scheduling/2779.yaml
suigh: updated.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Merged items 3/4 into one item, remove the code details, the new message is:

3. The registered custom queue will be passed to the scheduler if it exists, or the current internal queue will be used.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: As mentioned above, merged it with item 3 and removed the code details.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: There is no change for the interface `SchedulingQueue`, just make it from private to publish, this is why its details are not listed. To avoid confusing, provide following message for more details:

1. Make `SchedulingQueue` a public interface which is private now in ([scheduling_queue.go](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go#L126)). Users can choose to implement their own queue.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: this part was removed.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: For phase 2, what we will do is
1. refine the codes, make the interface simple and easy to use, decouple this feature with other parts of the scheduler
2. refine some designs with user's feedback

Update the message as following:
```
 #### Beta -> GA Graduation

+- Only necessary interface and structs are exposed to users.
 - Allowing time for feedback to ensure that the new interface sufficiently expresses users requirements.
```

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/kep.yaml
suigh: updated.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: For phase 2, what we will do is

refine the codes, make the interface simple and easy to use, decouple this feature with other parts of the scheduler
refine some designs with user's feedback
Update the message as following:

suigh:
@alculquicondor 
Thanks very much for your review and suggestions. 

For the solution of suspending a job, it works for sharing resources at job level, but it doesn't work for pod level sharing. E.g., if we can govern the resource sharing by stopping half of running pods in a job (the stopped pods will be re-created later), suspending the whole job is overkilled. 

Agree that `I'm not convinced that the scheduling queue sorting is gonna guarantee any type of resource sharing...`, only queue sorting cannot resolve this issue, we also need preemtption to preempt the pods from over-used users, then the pods from under-used users can run, and the KEP is to ensure the pods from under-used user can run first. Without this KEP, ping-pong preemption may happen as the pods from over-used users can be preempted and then get resource before other pods, and this step can be repeated.

For suggestion `But then we can have a PreFilter (and perhaps a Permit) plugin that would return Unschedulable if such pod would violate the resource sharing.`, I tried this one before, it is a little complex for developers, expecially when there are thousands of pods with different requests from differnt users, it is hard to say if a pod should run or not. For example, podA is from an over-used user, running it will violate the resource sharing, should we just mark it as `Unschedulable`? It is not true as if no other pods can run at all, podA should run to make full use of the cluster resources. It is hard to know the status of all other pods, even we can know, the performance will be impacted by going through all the pods, so it is hard to mark it as `Unschedulable`. With this KEP, we can control the sorting logic and always try the pods from under-used users first, and try other pods later, the above issue will not happen, the resources in the cluster can be fully used.

With this KEP and the scheduler plugin framework, developers can handle the pods with dependencies (e.g., the case that mentioned by me, some AI workload, etc) with straightforward logic, it is easy to maintain, and performance can be improved.

Updated "Alternatives" for the discussions in the readme, please help review them.

suigh:
@alculquicondor 
Thanks very much for your review and suggestions. 

For the solution of suspending a job, it works for sharing resources at job level, but it doesn't work for pod level sharing. E.g., if we can govern the resource sharing by stopping half of running pods in a job (the stopped pods will be re-created later), suspending the whole job is overkilled. 

Agree that `I'm not convinced that the scheduling queue sorting is gonna guarantee any type of resource sharing...`, only queue sorting cannot resolve this issue, we also need preemtption to preempt the pods from over-used users, then the pods from under-used users can run, and the KEP is to ensure the pods from under-used user can run first. Without this KEP, ping-pong preemption may happen as the pods from over-used users can be preempted and then get resource before other pods, and this step can be repeated.

For suggestion `But then we can have a PreFilter (and perhaps a Permit) plugin that would return Unschedulable if such pod would violate the resource sharing.`, I tried this one before, it is a little complex for developers, expecially when there are thousands of pods with different requests from differnt users, it is hard to say if a pod should run or not. For example, podA is from an over-used user, running it will violate the resource sharing, should we just mark it as `Unschedulable`? It is not true as if no other pods can run at all, podA should run to make full use of the cluster resources. It is hard to know the status of all other pods, even we can know, the performance will be impacted by going through all the pods, so it is hard to mark it as `Unschedulable`. With this KEP, we can control the sorting logic and always try the pods from under-used users first, and try other pods later, the above issue will not happen, the resources in the cluster can be fully used.

With this KEP and the scheduler plugin framework, developers can handle the pods with dependencies (e.g., the case that mentioned by me, some AI workload, etc) with straightforward logic, it is easy to maintain, and performance can be improved.

Updated "Alternatives" for the discussions in the readme, please help review them.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: custom scheduler developers, updated this part to show it explicitly.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: The developers will get a scheduler binary with the custom queue, then they can replace the default scheduler binary with it, or run it as the secondary scheduler of the k8s.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: To resolve the issue that mentioned by me, I will build some sub-queues of the queue, one sub-queue for one user. Once schedule wants to pop a pod from the queue, the queue will go through the sub-queues, select the most under-used, and pop a pod from it. 

The purpose of this solution is to expose the entire queue, so that scheduler developer can control all the details in a queue, and we needn't expose any new function/structs/etc in the future. This is used to save our effort. :)

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: This case is used to show that there are many possible customer cases that we cannot handle with the current internal queue, and we need a solution for customers to implement the queue at their will. This is also the reason that why we want the customer the implement the whole queue.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: This feature is only for the [scheduling_queue.go](https://github.com/kubernetes/kubernetes/blob/ee459b8969ed2abfed79a07d4ac9d41f13f18ce6/pkg/scheduler/internal/queue/scheduling_queue.go), no plugin will be touched. Do we need to plugin here?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: At build time, the custom queue will be part of the scheduler binary after building.  Readme updated.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: a typo?

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -396,3 +369,11 @@ Why should this KEP _not_ be implemented?
 An alternative is to enhance the current queue to meet some requests. The drawback
 is that we cannot handle all the potential requests as the effort is big, we should
 give the ball to the users, and they can do what they want.
```
alculquicondor, line 371: I disagree with this statement.

I'm more in favor of keeping the interface small with clear use cases.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -396,3 +369,11 @@ Why should this KEP _not_ be implemented?
...
+
+Another alternative is to suspend a job and don't let it go unless the resource
+sharing policy will not be broken. The drawback is that all the pods in the job will
+be impacted, it cannot handle the case that only some pods need to be touched.
+
+Another alternative is to make the pod that will break the resource sharing policy with
+`Unschedulable`. The drawback is that making such a decision needs the status of all other
+pods, the logic is complex, and the performance will be impacted.
```
alculquicondor, line 379: In the proposal, don't you have to know the status of all pods during sorting? It looks like the latency is moving from one place to another.

alculquicondor (commented):
In general, I think the motivation is incomplete.

I would like to see how the interface would be used to satisfy the use cases you describe. The use cases themselves are still too high level. More detail will help. This would also inform better the "alternatives"

Then, your beta->GA criteria is basically to rewrite the whole interface and expose only "the necessary interface and structs". That's not the intent of graduation criterias. 
A graduating feature should have minimal variances from what is already released as alpha or beta. What you are proposing is that we rethink everything in the next release. I would like to know how this interface could look like now. Which brings us back to the point above: we need to understand better the use case.

Huang-Wei:
The key focus of this KEP is not to change the internal queue, or queue sorting plugin. Quite the contrary, that logic will be kept as is. The main idea is to give plugin developers an option to manage the whole queuing logic, including how to prioritize/pop/backoff/flush pods. It's all up to the plugin developers. And in terms of changes to the scheduler framework, it needs to make some functions public and tweak some parameters.

Another motivation of this KEP, IMHO, is to support multi-subqueue, which can be mapped to user-defined multi-tenancy model. If the plugin developers can implement their own scheduling queue, they're capable to manage pods for different tenants at ease. As we know, in the default scheduler, sorting in a global queue is somewhat awkward. I see some solutions are tagging pods to work it around, but building their own multi-subqueues can be more efficient and solve the problem fundamentally.

What @alculquicondor mentioned, managing the pods at creation time on the controller side is also a nice alternative. @suigh I think it'd be nice to compare the pros/cons, and polish the motivations/alternatives sections a bit. I'd expect to see at least compelling use cases that a custom scheduling queue can achieve (or achieve efficiently), while others cannot - which will make this KEP convicable.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: The target of his KEP is to enable scheduler developers to implement their own logic to manage the order the pods. We can expose part of the interface for some known cases now, we will have to expose more interfaces in the future to handle new cases as developers' requests are un-preditable, this needs effort and maybe a new KEP. To expose the whole interface, we can handle the potentail requests by this KEP, needn't extra effort in the future.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: Oh, the `status` here is not clear. I update it to `The drawback is that making such a decision needs to know whether other
pods can run or not, ......`. To mark a pod as `Unschedulable`, the alternative needs to predict whether the competitor pods can run or not, this is complex.

In my proposal, the pods for different users will be put into different queues, and once the scheduler want to pop a pod, it will select the most under-used queue and pop one, the scheduler needn't care about other pods in other queues, logic is simple, there is no latency.

suigh:
@alculquicondor @Huang-Wei 
Thanks very much for your review and suggestions, the readme is updated as following with your comments, would you please help review again?

1. Add following message at `Motivation` section to explain why we need to exponse the whole interface instead part of it.
```
This KEP is not used to change the internal queue, or queue sorting plugin, quite the contrary,
that logic will be kept as is. The main idea is to give scheduler developers an option to manage the whole
queuing logic, including how to prioritize/pop/backoff/flush pods. It's all up to the scheduler developers.
To ensure the developers can have full control of the queuing logic, the whole interface of the internal
queue will be exposed instead of part of it that can only meet above examples. By exposing the whole interface,
we needn't update the interface that need to be exposed again and again for coming demands, this can save
developers' effort as they needn't raise a KEP for such requests, this can also save community's effort.
```

2. Removed `Only necessary interface and structs are exposed to users` from ` Beta -> GA Graduation` section as this message may confuse the reader. What we will do is to expost the whole interface, not part of it.

3. Add more details in the `Alternatives` section to show what we do to resolve the multi-tenancy case, and what's the pros/cons when comparing with the possible alternatives.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
```diff
@@ -84,6 +84,14 @@ Given that the business needs may vary greatly, it'd be desirable to provide a r
 implement their custom queue implementation, while the upstream maintainers focus
 on keeping the core small and extensible, and thus only maintain the internal queue piece.
 
+This KEP is not used to change the internal queue, or queue sorting plugin, quite the contrary,
```
alculquicondor, line 87: This should go as Non-Goal

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
alculquicondor: Can't you keep the same information about usage ratios in the PreFilter plugin? The latency should be similar.

alculquicondor:
In general, this change would be quite simple for us to maintain (just an exported interface)

However, I want to make sure you are not shooting yourselves in the foot by taking the challenge of building your own queue that might not solve the requirements you have.

Also, if you succeed in your multitenancy efforts, perhaps it's something that could be upstreamed. But as it is currently stated (you have a completely alternative implementation), it's gonna be harder to justify.

keps/sig-scheduling/2779-support-custom-queue-for-the-scheduler/README.md
suigh: The above message is a little long for `Non-Goals`, how about updating the readme as following?

Add message `Enable scheduler developers to change the internal queue, or queue sorting plugin` in the `Non-Goals` section to show that it is not our goal.

Keep the orignal message in the `Motivation` section to let others know the details of our design.

suigh:
For question `Can't you keep the same information about usage ratios in the PreFilter plugin? The latency should be similar.`

I guess your suggestion is to drop the pods from over-used user in PreFilter plugin, and only try to bind the pods from under-used user. This will cause latency when there are thousands of pods from over-used user. In the worst case, after we go through all the pods from under-used user and cannot get pod to run, we will have to re-check the dropped pods from over-used user.

suigh:
> In general, this change would be quite simple for us to maintain (just an exported interface)
> 
> However, I want to make sure you are not shooting yourselves in the foot by taking the challenge of building your own queue that might not solve the requirements you have.
> 
> Also, if you succeed in your multitenancy efforts, perhaps it's something that could be upstreamed. But as it is currently stated (you have a completely alternative implementation), it's gonna be harder to justify.

Thanks very much for your reminder. :)

I had implemented similar solution in my own repo, it can meet my multi-tenancy request. While it is hard for me to open source it as it is implemented on a non-kubernetes-official structure, and it is hard to sync up with the upstream. This is why this KEP was logged. Once this KEP was finished, I will add some codes at `kubernetes-sigs/scheduler-plugins` to show the details about the multi-tenancy solution.

alculquicondor:
Can @suigh  and @denkensk come to the next sig meeting?

suigh:
> Can @suigh and @denkensk come to the next sig meeting?

I can attend the meeting, and I'm not sure if Alex (@denkensk) is available or not. How long can we take to talk this ticket in the meeting? I didn't attend this meeting before.

> Another alternative: instead of exposing the entire SchedulingQueue, what about adding an interface just for `activeQ`?
Yes, it is a another solution that can help the multi-tenancy case. This solution is not used because:
1. scheduler developer cannot change the logic for backoff and unschedulable queues. E.g., in my project, backoff queue is not necessary at all, remove it can improve the performance.
2. developer don't have full control of the SchedulingQueue,  if a developer want to change something, she/he will have to talk with community, the effort is big.

denkensk:
> Can @suigh and @denkensk come to the next sig meeting?

Sure

suigh:
@alculquicondor, @denkensk,
I checked the meeting time, it is 10:00 AM PDT and 1:00 AM for my local time CST,  it is a little early.  Can we have the meeting in another time?   There is once a meeting at 5:00 PT, and 8:PM for CST, it is a good time, but it is cancelled. Can we reuse the time?

alculquicondor:
@Huang-Wei WDYT?

alculquicondor:
> in my project, backoff queue is not necessary at all, remove it can improve the performance.

That's the kind of things I would like to avoid. Removing the backoff can lead to starvation.

ahg-g:
I recently came back from vacation, I would like to take a look at this proposal, likely today or tomorrow.

suigh:
> > in my project, backoff queue is not necessary at all, remove it can improve the performance.
> 
> That's the kind of things I would like to avoid. Removing the backoff can lead to starvation.

Since developers have full control of the codes, the developers should handle all the related issues. If developers don't want to change the backoff queue, they can just leave it untouched, the result is same as only exposing the activeQ. We should try to remove the limitations on developers.

ahg-g:
The KEP is simply proposing a refactor to allow passing in different SchedulingQueue implementation on scheduler instantiation. 

I don't see this worthy of a KEP as it is proposed; also, as proposed, I don't think we will want to document this as a supported customization feature (like we do for plugins).

I think what would be worthy of support is a new SchedulingQueue interface that is abstract enough of the current internal implementation. For example, an interface that doesn't expose the details of Active/backoff/unschedulable queues etc.

Huang-Wei:
> @Huang-Wei WDYT?

If 5:00 PM PST works for both of you, I can setup a meeting next week.

Huang-Wei:
> I don't see this worthy of a KEP as it is proposed;

Probably this is a good way to move it forward, like what we did for some framework changes ([1](https://github.com/kubernetes/kubernetes/pull/103383), [2](https://github.com/kubernetes/kubernetes/pull/101394)) where the design doc was linked (but not a KEP).

Given that we don't have a clear plan for phase 2, a viable way is to claim minimum changes in a document that makes SchedulingQueue customizable, and evolve SchedulingQueue's abstraction in the long run. But if we go with a KEP, we may have to have concrete plans in the very beginning. @suigh WDYT?

ahg-g:
If other community members can't attend, then it is probably better to continue the discussion in this open forum.

suigh:
Making SchedulingQueue customizable with a document instead of KEP is good to me, can we go by this way? If yes, I will write a design document and update more details about motivation.

What we should in my mind:
1: Write a document about how to make the SchedulingQueue customizable instead of a KEP, and log a PR for the implementation. With this way, this feature is not a public one, but the developers can use this feature to implement their custom queues.
2: After the codes in step 1 are develiered, we will work on the SchedulingQueue abstraction, make it clean and simple for public usage.
3: Make this feature as a public one with a KEP when SchedulingQueue is clean/simple enough.

suigh:
@alculquicondor, @denkensk, @Huang-Wei, @ahg-g 

Thanks very much for your suggestions, help and patience on this KEP. 

Seems this is not the right time to expose the whole interface of the internal queue now, I will try simplifly the internal queue first as @ahg-g suggested, and maybe expose the refined interfacein the future. 

Close this KEP now, and thanks for your time again. :)

alculquicondor:
/close

Follow the discussion in kubernetes/kubernetes#102271 or open a new one

PR #2783, opened 2021-06-11
apf: estimate width of LIST

tkashem:


keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+- `Most Recent`: served from etcd via quorum read
+- `Not older than`: at least as new as the provided `resourceVersion`
+- `Exact`: at the exact resource version provided
+- `Continue Token, Exact`: return data at the resource version of the initial paginated list call  
+- `Any`: strong consistency is not required, data at any resource version may be served
+
+We want to feed the semantics in to the width estimation process, we can use the following back of the
+envelope mapping:
+```
+S = {'Most Recent'=4 | 'Not older than'=2 | Exact=2 | 'Continue Token, Exact'=2 | 'Any'=1}
```
wojtek-t, line 1267: This is making it a magic from the very beginning and I think I have hard time with intuitive explanation of those constants. In particular, I see no reason why "Most Recent" and "Exact" should really have the same cost - they do exactly the same thing.

I would suggest dropping the S for now and if really needed leaving it for further releases.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+
+We want to feed the semantics in to the width estimation process, we can use the following back of the
+envelope mapping:
+```
+S = {'Most Recent'=4 | 'Not older than'=2 | Exact=2 | 'Continue Token, Exact'=2 | 'Any'=1} 
+```
+
+Estimate `N`:
+
+- If the specified value of `limit` is invalid then the apiserver will reject the request? in this case we can
```
wojtek-t, line 1272: We don't even have to discuss, because it will be rejected anyway. So you can set whatever you want, because it doesn't matter.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+  set `N=1`.
+- For a valid value of `limit` we know the upper bound of the total number of objects that will be processed, so `N=L`.
```
wojtek-t, line 1274: As I mentioned during the last P&F meeting - this isn't really true.

The `limit` is the upper bound for number of `returned objects`, not for the ones to process. In the most pathological case, one can imagine setting limit to 1 and setting a selector to something that doesn't match any object and we will have to process every single object.

So we should use N=L if there is empty selector and #total number of objects otherwise.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+- If `limit` is not specified, we should default `N` (a pessimistic default is `500`?), or we can tune it based
```
wojtek-t, line 1275: Pestimistic defaulting is not the best idea, given that amount of objects may be very different. We should be using metrics for this (we know that value).

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+  - If it is not set, then it is the first of the `chunked` LIST request, depending on `resourceVersion` its semantics
+    could be `Most Recent` (most expensive).
+  - If it is set, it's a consecutive request and its semantics should be `Exact` which makes it more likely to be
+    served from watch cache. But due to HA configuration, the consecutive request(s) may land in a separate 
+    apiserver where the watch cache is stale.
+    
+
+Estimate `W`:
+```
+W = S * ceil(transform(N))
```
wojtek-t, line 1287: As mentioned above - I would start with W = transform(N) directly. We can tune in the future if needed.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
...
+```
+
+Serving the request gets exponentially more expensive as the number of objects increase.
```
wojtek-t, line 1290: This isn't exponential. In fact - it's already describe above:
https://github.com/kubernetes/enhancements/pull/2783/files#diff-0d81f61bda1a026d678a311aa592601d41014434a5a3c91e17a3a3dfe1d4527aR1169

I don't yet know the best value of A and B, but let's not try to figure them out in KEP - we will tune them based on experiments.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
```diff
@@ -1225,6 +1226,97 @@ Note that given the estimation for duration of processing the requests is
 automatically corrected (both up and down), there is no need to change that
 in the initial version.
 
+
+#### Estimate width of LIST request
```
wojtek-t, line 1230: BTW - let's not introduce another section for it and incorporate it in the above one (determining the width)

keps/sig-api-machinery/1040-priority-and-fairness/README.md
MikeSpreitzer: If it is rejected during "execution" (from an APF point of view), then it matters.

keps/sig-api-machinery/1040-priority-and-fairness/README.md
MikeSpreitzer: As far as I understand, traversing the objects and evaluating the selector is not the majority of the CPU cost in a LIST request.  How about we make it `N=f*L`, where `f` is a fraction we determine experimentally?

wojtek-t: For an individual item it's usually not the majority I agree.

But in a pathological case I described above, if there are e.g. 100,000 pods but my selector is not matching any of them, even if I set limit L=1, then the cost is actually proportional to this 100,000.

So in other words - I would much rather prefer to overestimate (for overload-protection) than underestimate (as this can lead to overloading).

keps/sig-api-machinery/1040-priority-and-fairness/README.md
wojtek-t: @MikeSpreitzer - can you clarify what do you mean by "during execution"? Do you mean:
(a) it was waiting for some time in the queue and was rejected?
(b) something else
If (a), can you clarify why it matters?

PR #2788, opened 2021-06-17
CSR Duration KEP

enj:

/sig auth
/assign @liggitt @smarterclayton @mikedanese

### DIFF
--- a/keps/sig-auth/2784-csr-duration/README.md
+++ b/keps/sig-auth/2784-csr-duration/README.md
@@ -89,6 +89,7 @@ that can be used to request a particular duration.
 ### Non-Goals
 
 - Requiring signers to honor the requested duration
+- Configuring existing in-tree consumers of the CSR API (i.e. kubelets) to make use of this feature
 
 ## Proposal
 
@@ -133,8 +134,10 @@ structs found in the `k8s.io/api/certificates/v1` and `k8s.io/api/certificates/v
 packages.
 
 A new optional `ExpirationSeconds` field will be added to this struct.  The go
-doc comment describes the behavior of this field.  A `*uint32` is used because
-the field is optional, must be positive and must not overflow JSON parsers.
+doc comment describes the behavior of this field.  A `*int32` is used because
+the field is optional and must not overflow JSON parsers (an unsigned type is
+avoided as we want to provide a detailed validation error on a negative input
+instead of a difficult to understand decoding error).
 
 ```go
 // CertificateSigningRequestSpec contains the certificate request.
@@ -164,7 +167,7 @@ type CertificateSigningRequestSpec struct {
   // The minimum valid value for expirationSeconds is 600, i.e. 10 minutes.
   //
   // +optional
-  ExpirationSeconds *uint32 `json:"expirationSeconds,omitempty" protobuf:"varint,8,opt,name=expirationSeconds"`
+  ExpirationSeconds *int32 `json:"expirationSeconds,omitempty" protobuf:"varint,8,opt,name=expirationSeconds"`
 
   // go doc omitted for brevity
   Usages []KeyUsage `json:"usages,omitempty" protobuf:"bytes,5,opt,name=usages"`
@@ -183,12 +186,17 @@ and signing is asynchronous which necessitates a buffer), `10` minutes seems lik
 appropriate minimum to prevent accidental DOS against the CSR API.  Furthermore,
 `10` minutes is a short enough lifetime that revocation is not of concern.
 
+Metrics will be included to show if requested expirations are being extended or
+truncated (i.e. is the requested duration being honored by the signer).
+
 ### Test Plan
 
 Unit tests covering:
 
 1. Validation logic for minimum duration
 2. `pkg/controller/certificates/authority.PermissiveSigningPolicy` updates to handle `expirationSeconds`
+3. Metrics
+4. CSR REST storage ignores `spec.expirationSeconds` when the feature gate is disabled
 
 Integration test covering:
 
@@ -200,8 +208,7 @@ Integration test covering:
 
 #### Alpha
 
-- This design will start at the beta phase and the functionality will always be enabled
-- There will be no feature gate associated with this design (discussed below)
+- This design will start at the beta phase and the functionality will be enabled by default
 
 This design represents a small, optional change to an existing GA API.  Thus it
 prioritizes rollout speed to allow clients to start using this functionality
@@ -217,7 +224,9 @@ during version skews (discussed below).
 
 #### GA
 
-- Confirm with cert-manager that the new functionality addresses their use case
+- Confirm with [cert-manager](https://github.com/jetstack/cert-manager/pull/3646) that the new functionality addresses their use case
+- Confirm with [pinniped](https://pinniped.dev) that the new functionality addresses their use case
+- Confirm that no other metrics are necessary
 - Wait one release after beta to allow bugs to be reported
 
 The existing conformance tests for the certificates API (`test/e2e/auth/certificates.go`)
@@ -237,7 +246,7 @@ could request a duration of a month but the signer could truncate the duration t
 its internal maximum of two weeks).  Thus this design emphasizes feature rollout
 speed to aid in ecosystem adoption instead of data durability.  Combined with the
 simplicity of implementation and low risk nature of the proposal, the alpha stage
-and related feature gate have been omitted from this design.
+has been omitted from this design.
 
 Clients that do not set the `spec.expirationSeconds` field will observe no change
 in behavior, even during upgrades and downgrades.
@@ -277,7 +286,7 @@ represents the status quo.
 
 In this scenario, the requested `spec.expirationSeconds` may be ignored because
 old API servers will silently drop the field on update.  This is harmless
-represents and the status quo.
+and represents the status quo.
 
 The CSR API is resilient to split brain scenarios as unknown fields are silently
 dropped and the `spec` fields are immutable after creation [1] [2] [3].
@@ -292,15 +301,11 @@ dropped and the `spec` fields are immutable after creation [1] [2] [3].
 
 ###### How can this feature be enabled / disabled in a live cluster?
 
-- Other
-  - Describe the mechanism:
-      As written above, this functionality will be enabled by default with no configuration options.
-  - Will enabling / disabling the feature require downtime of the control
-    plane?
-      N/A
-  - Will enabling / disabling the feature require downtime or reprovisioning
-    of a node? (Do not assume `Dynamic Kubelet Config` feature is enabled).
-      N/A
+- Feature gate
+  - Feature gate name: `CSRDuration` (enabled by default)
+  - Components depending on the feature gate:
+    - kube-apiserver
+    - kube-controller-manager
 
 ###### Does enabling the feature change any default behavior?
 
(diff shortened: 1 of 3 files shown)


keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+  //
+  //   1. Old signer that is unaware of the field (such as the in-tree
+  //      implementations prior to v1.22)
+  //   2. Signer whose configured maximum is shorter than the requested duration
+  //   3. Signer whose configured minimum is longer than the requested duration
+  //
+  // The minimum valid value for expirationSeconds is 600, i.e. 10 minutes.
+  //
+  // +optional
+  ExpirationSeconds *uint32 `json:"expirationSeconds,omitempty" protobuf:"varint,8,opt,name=expirationSeconds"`
```
smarterclayton, line 167: What error do you get when someone sends you -1?  Is it a prototbuf / json decoding error with a 429?  If not, I'd probably prefer having int32 so that we can properly validate.

liggitt: agree... I can't find any other examples of uint in our rest API

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+1. Creating and approving CSRs and asserting that certificate signer controllers such as
+   `pkg/controller/certificates/signer.NewKubeAPIServerClientCSRSigningController` honor
+   `spec.expirationSeconds` by checking the duration of the issued certificate.
+
+### Graduation Criteria
+
+#### Alpha
+
+- This design will start at the beta phase and the functionality will always be enabled
+- There will be no feature gate associated with this design (discussed below)
```
liggitt, line 204: even if this starts in beta, I'd still expect a gate

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+#### Beta
+
+- Feature fully implemented as described in this design
+- Unit tests completed and enabled
+- Integration test completed and enabled
+- [CSR docs](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests) updated with details about usage of the `spec.expirationSeconds` field
+
+#### GA
+
+- Confirm with cert-manager that the new functionality addresses their use case
```
liggitt, line 220: would like more than a single consumer referenced here

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+interpret the data correctly as defined by the feature being implemented.
+
+In the case of this design, data durability is not of concern as clients, no matter
+what, have to assume that signers may ignore the requested duration completely
+even after they have been updated to understand the field (for example the client
+could request a duration of a month but the signer could truncate the duration to
+its internal maximum of two weeks).  Thus this design emphasizes feature rollout
+speed to aid in ecosystem adoption instead of data durability.  Combined with the
+simplicity of implementation and low risk nature of the proposal, the alpha stage
+and related feature gate have been omitted from this design.
```
liggitt, line 240: I could buy accelerating to beta, but not dropping the feature gate

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+      N/A
+
+###### Does enabling the feature change any default behavior?
+
+Existing clients would continue to leave `spec.expirationSeconds` unset and thus
+would observe no difference in behavior.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+N/A
```
liggitt, line 312: reference the feature gate... if use of the shorter expirations reveals a problem, controller-manager should be able to turn off honoring the field using a gate

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+Since it is optional for signers to honor `spec.expirationSeconds`, this design
+is fully tolerant of API server and controller manager rollouts/rollbacks that
+fail or get wedged in a partial state.  The worse case scenario is that the
+`spec.expirationSeconds` field is ignored, which mimics the status quo.  Clients
+must always check the duration of the issued certificate to determine if the
+requested `spec.expirationSeconds` was honored.
+
+###### What specific metrics should inform a rollback?
+
+N/A
```
liggitt, line 336: do we want visibility to clients requesting expirations that get truncated?

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+that can be used to request a particular duration.
+
+### Goals
+
+- Allow clients to request a duration for requested certificates
+- Allow the CSR API to be easily [adopted in the Kubernetes ecosystem](https://github.com/jetstack/cert-manager/pull/3646)
+
+### Non-Goals
+
+- Requiring signers to honor the requested duration
```
liggitt, line 91: is configuring existing in-tree consumers of the CSR API (kubelets) to make use of this feature a non-goal?

keps/sig-auth/2784-csr-duration/README.md
enj: Not sure if such a metric would be useful because there is no way for a client to know if the requested expiration is aligned with the certificate controller config.  Since we do not have metrics related to the CSR controllers today so I left this out.

keps/sig-auth/2784-csr-duration/README.md
enj: So a feature gate that:

1. Is enabled by default
2. Starts out at beta

?

keps/sig-auth/2784-csr-duration/README.md
enj: IMO, it is a non-goal but I could see part of the GA requirements being some kind of validation that the kubelet CSR flow could use this... which may require the config?  I would prefer to defer this wiring to the existing efforts for CSR rotation but honestly I cannot remember where we are with that.

keps/sig-auth/2784-csr-duration/README.md
deads2k: > So a feature gate that:
> 
> 1. Is enabled by default
> 2. Starts out at beta

Yes.  That gives us a way to turn it off in a running cluster without rolling back.

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+Unit and integration tests will cover cases where `spec.expirationSeconds` is
+specified and cases where it is left unspecified.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+Since it is optional for signers to honor `spec.expirationSeconds`, this design
+is fully tolerant of API server and controller manager rollouts/rollbacks that
+fail or get wedged in a partial state.  The worse case scenario is that the
```
deads2k, line 329: Seems like the worst case scenario is that a kube-controller-manager or a critical approver attempts to use this field, gets a decoding error, a nil pointer, or some other fatal failure and panics in the presence of this field, causing CSRs to linger unapproved and unsigned, causing all kubelet credentials to expire and fail.

I want a feature gate to turn this off if necessary.

keps/sig-auth/2784-csr-duration/README.md
deads2k: Without a metric, how will a cluster-admin know that this feature is functioning?  How will he know that he has approvers that are not honoring the value in the long term?

It seems like a thing worth tracking.

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,473 @@
...
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+N/A
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+Use `kubectl` to determine if CSRs with `spec.expirationSeconds` set are being
```
deads2k, line 355: This answer is only half of the API. The other half is seeing it honored in the result.

keps/sig-auth/2784-csr-duration/README.md
enj: Added explicit non-goal.

keps/sig-auth/2784-csr-duration/README.md
enj: Added explicit feature gate.

keps/sig-auth/2784-csr-duration/README.md
enj: Changed to `*int32`.

keps/sig-auth/2784-csr-duration/README.md
enj: Added pinniped as a consumer.

keps/sig-auth/2784-csr-duration/README.md
enj: Added explicit metrics around this.

keps/sig-auth/2784-csr-duration/README.md
enj: Added `.status.certificate` details.

deads2k:
The design and the PRR both lgtm.  The exception request was approved. 

/lgtm
/approve
/hold

placing a hold to allow additional reviewers this week.

keps/sig-auth/2784-csr-duration/README.md
```diff
@@ -0,0 +1,518 @@
...
+
+- Feature fully implemented as described in this design
+- Unit tests completed and enabled
+- Integration test completed and enabled
+- [CSR docs](https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests) updated with details about usage of the `spec.expirationSeconds` field
+
+#### GA
+
+- Confirm with [cert-manager](https://github.com/jetstack/cert-manager/pull/3646) that the new functionality addresses their use case
+- Confirm with [pinniped](https://pinniped.dev) that the new functionality addresses their use case
```
liggitt, line 228: non-blocking for merge of this doc, but can you confirm with them this addresses their use case ~now, not in a release?

liggitt (commented):
lgtm as well

keps/sig-auth/2784-csr-duration/README.md
mattmoyer: This looks good for the Pinniped use case as well. Our planned usage is for a credential exchange API just like the second user story:

1. External user authenticates to a trusted service (using some custom application-specific credential).
1. Trusted service creates a CSR with a short `spec.expirationSeconds` (probably the 10 minute minimum).
1. Trusted service approves its own CSR and relies on a default signer to sign it.
1. Trusted service retrieves the signed certificate and returns it to the authenticated external user.

We would quickly add support for new flow but keep some fallback code that uses a custom signer until we no longer support current Kubernetes versions. We can probe for this feature dynamically and select the modern or fallback implementation appropriately.

smarterclayton:
Sorry for the delay, I thought I dropped a "this looks good and Jordan and David hit all of my comments already", but it looks like I never pressed comment.

This looks good to me, no further comments above what was already landed.

supriya-premkumar:
Hi @enj 👋🏽. Supriya here, 1.22 Enhancements Shadow.
For the enhancement to be included in the 1.22 milestone, it must meet the following criteria:

- Complete all the Beta release PRR questionnaire(upgrade/rollback test)
- The PR must be approved and merged by **EOD June 25, 2021**

Thank you!

PR #2794, opened 2021-06-23
Update NoSwap SwapBehavior to LimitedSwap

ehashman:
Per https://github.com/kubernetes/kubernetes/pull/102823/files#r653144961

### DIFF
--- a/keps/sig-node/2400-node-swap/README.md
+++ b/keps/sig-node/2400-node-swap/README.md
@@ -343,7 +343,7 @@ type KubeletConfiguration struct {
 
 type MemorySwapConfiguration struct {
 	// Configure swap memory available to container workloads. May be one of
-	// "", "NoSwap": workloads cannot use swap
+	// "", "LimitedSwap": workload combined memory and swap usage cannot exceed pod memory limit
 	// "UnlimitedSwap": workloads can use unlimited swap, up to the allocatable limit.
 	SwapBehavior string
 }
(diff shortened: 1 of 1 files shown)


keps/sig-node/2400-node-swap/README.md
```diff
@@ -343,7 +343,7 @@ type KubeletConfiguration struct {
 
 type MemorySwapConfiguration struct {
 	// Configure swap memory available to container workloads. May be one of
-	// "", "NoSwap": workloads cannot use swap
+	// "", "LimitedSwap": workload combined memory and swap usage cannot exceed pod memory limit
```
mrunalp, line 346: Can we clarify that v1 and v2 behavior will diverge here?

ehashman: It won't diverge in the current implementation, as we're not disallowing swap usage on cgroupsv2. If we want to allow for a "noswap" option in the future, we can add that for v2 only.

mrunalp: Yes, but we can clarify that this LimitedSwap option only applies to cgroups v1 and not v2.

ehashman: It applies to both right now.

ehashman: @mrunalp and I synced offline; I was unaware that cgroupsv2 doesn't support a combined memory+swap limit, and must configure them separately.

This behaviour difference is now addressed below!

mrunalp (approved):
/lgtm

derekwaynecarr:
Thank you for making the updates.  Thanks @sjenning for sharing your insights with @ehashman and myself during the review.

/approve
/lgtm

PR #2807, opened 2021-07-01
Add new changes to recover from expansion KEP

gnufied:
xref https://github.com/kubernetes/enhancements/issues/1790

/sig storage


keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,73 +106,139 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
+                newSize = allocatedSize
+            }
+        }
+
+    }
+    return *newSize
+}
+```
+
+One more key design choice is - once expansion on node failed and user requested lower size (by modifying `pvc.Spec.Resources`) - kubelet must not retry expanding the same volume to previous `allocatedResources`, because external-resizer will consider `ResizeFailedOnNode` state as a terminal error and allow lowering of `allocatedResources` - and at the same time kubelet will try to expand the volume to higher `allocatedResources`. This will be solved in kubelet by following code:
```
gnufied, line 156: Discussed this offline with @jsafrane and he correclty pointed out that - this check here is not good enough. There is still a chance that while kubelet is increasing the volume size, user may reduce the size and resize-controller may pick lower value (because PVC still has `ResizeFailedOnNode` condition).

To fix - we are going to need additional `ResizeStatus` value. I will update the KEP.

gnufied: We managed to solve this problem with new design.

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
 
-To solve aforementioned problem - we propose that, a new field will be added to PVC, called `pvc.Status.AllocatedResources`. When a PVC is created - this field defaults to `pvc.Spec.Resources` but when user expands the PVC,
-and when expansion-controller starts volume expansion - it will set `pvc.Status.AllocatedResources` to user requested value in `pvc.Spec.Resources` before performing expansion. The quota calculation will be updated to use `max(pvc.Spec.Resources, pvc.Status.AllocatedResources)` which will ensure that abusing quota will not be possible.
+To solve aforementioned problem - we propose that, a new field will be added to PVC, called `pvc.Status.AllocatedResources`. When user expands the PVC, and when expansion-controller starts volume expansion - it will set `pvc.Status.AllocatedResources` to user requested value in `pvc.Spec.Resources` before performing expansion. The quota calculation will be updated to use `max(pvc.Spec.Resources, pvc.Status.AllocatedResources)` which will ensure that abusing quota will not be possible.
```
msau42, line 115: The resizing status also has to be updated to InProgress at this step right?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
 
-When user reduces `pvc.Spec.Resources`, expansion-controller will set `pvc.Status.AllocatedResources` to lower value (thereby giving quota back to the user) - only if current actual size of volume is less than or equal to `pvc.Spec.Resources` after entire control-plane and node side expansion is successfully finished. It will fetch actual size of the volume by using `ControllerGetVolume` CSI RPC call. It is possible to track completion of resizing operation in external-resizer via function - https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/controller/controller.go#L394.
+Resizing operation will always work towards full-filling size recorded in `pvc.Status.AllocatedResources` and only when previous operation has finished(i.e `pvc.Status.ResizeStatus` is nil) or when previous operation has failed with a terminal error - it will use new user requested value from `pvc.Spec.Resources`.
```
msau42, line 117: Can you be more explicit about what are the conditions to signal "in progress", "finished" and "terminal error"?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
+When user reduces `pvc.Spec.Resources`, expansion-controller will set `pvc.Status.AllocatedResources` to lower value only if:
 
-*Note:* This proposal expects that users can not modify `pvc.Status` directly and cheat quota system. In general this should be fine because users should not have access to edit status of almost anything. Link to discussion on slack - https://kubernetes.slack.com/archives/CJUQN3E4T/p1620059624022100
+1. If `pvc.Status.ResizeStatus` is `ControllerExpansionFailed` (indicating that previous expansion to last known `allocatedResources` failed with a final error) and previous control-plane has not succeeded.
```
msau42, line 121: what does "previous control plane has not succeeded" mean?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
+Resizing operation will always work towards full-filling size recorded in `pvc.Status.AllocatedResources` and only when previous operation has finished(i.e `pvc.Status.ResizeStatus` is nil) or when previous operation has failed with a terminal error - it will use new user requested value from `pvc.Spec.Resources`.
 
-If CSI driver does not have `GET_VOLUME` controller capability(or `ControllerGetVolume` does not report volume size) and `pvc.Spec.Resources` < `pvc.Status.AllocatedResources` (i.e user is attempting to reduce size of a volume that expansion controller previously tried to expand) - then although expansion-controller will try volume expansion with value in `pvc.Spec.Resources` - it will not reduce reported value in `pvc.Status.AllocatedResources`, which will result in no quota being restored to the user. In other words - for CSI drivers that don't have `GET_VOLUME` controller capability - `pvc.Status.AllocatedResources` will report highest requested value and reducing `pvc.Spec.Resources` will not result in reduction of used quota.
+When user reduces `pvc.Spec.Resources`, expansion-controller will set `pvc.Status.AllocatedResources` to lower value only if:
```
msau42, line 119: "only if one of the following is true"

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
 
-*Note:* This proposal expects that users can not modify `pvc.Status` directly and cheat quota system. In general this should be fine because users should not have access to edit status of almost anything. Link to discussion on slack - https://kubernetes.slack.com/archives/CJUQN3E4T/p1620059624022100
+1. If `pvc.Status.ResizeStatus` is `ControllerExpansionFailed` (indicating that previous expansion to last known `allocatedResources` failed with a final error) and previous control-plane has not succeeded.
+2. If `pvc.Status.ResizeStatus` is `NodeExpansionFailed` and SP supports node-only expansion (indicating that previous expansion to last known `allocatedResources` failed on node with a final error).
```
msau42, line 122: How do you detect a driver supports node-only expansion?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
+3. If `pvc.Status.ResizeStatus` is `nil` or `empty` and previous `ControllerExpandVolume` has not succeeded.
```
msau42, line 123: In what scenario would this be possible? If ControllerExpandVolume has not succeeded, wouldn't it be in failed state?

keps/sig-storage/1790-recover-resize-failure/README.md
msau42: Can you also clarify, is this done in in-tree controller or csi sidecar?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
 ##### Case 0 (default PVC creation):
 - User creates a 10Gi PVC by setting - `pvc.spec.resources.requests["storage"] = "10Gi"`.
-- API server sets `pvc.Status.AllocatedResources` to "10Gi" on creation.
 
 ##### Case 1 (controller+node expandable):
 - User increases 10Gi PVC to 100Gi by changing - `pvc.spec.resources.requests["storage"] = "100Gi"`.
-- `pvc.Status.AllocatedResources` still reports `10Gi`.
 - Quota controller uses `max(pvc.Status.AllocatedResources, pvc.Spec.Resources)` and adds `90Gi` to used quota.
 - Expansion controller starts expanding the volume and sets `pvc.Status.AllocatedResources` to `100Gi`.
+- Expansion controller also sets `pvc.Status.ResizeStatus` to `ControllerExpansionInProgress`.
```
msau42, line 137: Do we need to set this at the same time as increasing AllocatedResources to avoid a race?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -103,75 +112,99 @@ reduces the PVC request size, for both CSI and in-tree plugins they are designed
...
-- Expansion controller notices that `pvc.Spec.Resources` < `pvc.Status.AllocatedResources` (meaning user tried to reduce size).
-- Expansion controller fetches actual size of the volume using `ControllerGetVolume` CSI RPC call.
-- Expansion controller sees that `actual_volume_size`(100Gi) >= `pvc.Status.AllocatedResources` (100Gi), so it does not make any modifictions to `pvc.Status.AllocatedResources`.
+- Expansion controller notices that previous expansion to last known `allocatedresources` is still in-progress.
+- Expansion controller keeps expanding the volume to `allocatedResources`.
+- Expansion to `100G` succeeds and `pv.Spec.Capacity` and `pvc.Status.Capacity` report new size as `100G`.
+- Although `pvc.Spec.Resources` reports size as `20G`, expansion to `20G` is never attempted.
+- Quota controller sees no change in storage usage by the PVC because `pvc.Status.AllocatedResources` is 100Gi.
+
+##### Case 4(Malicious User and rounding to GB/GiB bounaries)
```
msau42, line 182: Do we need the new driver call to return actual size anymore? What if user increases size quickly twice in a row, and then lowers back down (eg to 50GB, 100GB, back down to 51GB)? Can we detect if the first expansion success is for the 50GB or 100GB?

I think since we don't set allocatedResources until right before we start expanding should be able to handle this?

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -360,7 +393,7 @@ _This section must be completed when targeting beta graduation to a release._
 
 ## Drawbacks
 
-- One drawback is while this KEP allows an user to reduce requested PVC size, it does not automatically reduce quota.
+- One drawback is while this KEP allows an user to reduce requested PVC size, it does not allow reduction in size all the way to same value as original size recorded in `pvc.Status.Cap`. The reason is currently resize controller and kubelet work on a volume one after the other and we need special logic in both kubelet and control-plane to process reduction in size all the way upto original value. This is somewhat racy and we need a signaling mechanism between control-plane and kubelet when `pvc.Status.AllocatedResources` needs to be adjusted. We plan to revisit and address this in next release.
```
msau42, line 396: At a first glance I think we should be able to use the new ResizeStatus to do something like `orig size == new size && resizeStatus is pending/inprogress/failed`. Agree we revisit before moving to beta.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: does not have `CONTROLLER_EXPAND_VOLUME` capability in control-plane? I could document that.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: Generally speaking yeah it is not possible. But from implementation perspective this condition "completes" the exhaustive branching check we perform when using new size from `pvc.Spec.Resources`. So this statement is merely for checking a box.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: Yes - whenever control-plane modifies `pvc.Status` it should submit both the changes as part of same patch request. I will make a note of it.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: In this case it means `pv.Spec.Cap < pvc.Spec.Resources`.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: Yes in general as such we don't need drivers to report *actual* size in `ControllerExpandVolume` RPC call's response, but  `pv.Spec.Capacity` is set to value reported by the driver - so if they do start reporting it, it will be used.

And also yeah - I think rapidly expanding and then shrinking works fine because we don't set `allocatedResources` until we start expanding the volume.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: fixed

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: noted.

keps/sig-storage/1790-recover-resize-failure/README.md
msau42: would something like "previous control-plane resize attempt" be more clear?

keps/sig-storage/1790-recover-resize-failure/README.md
msau42: Can you add a flow for this rapid expansion then shink case? ie something like:
1. user increases size to 50GB
2. resize to 50GB starts but still in progress
3. user increases size to 100GB
4. first resize operation timesout for some reason
5. ...

deads2k:
is this KEP moving to beta?  If so, you need to update https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-storage/1790.yaml

deads2k:
> is this KEP moving to beta? If so, you need to update https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-storage/1790.yaml

Also, it looks like several sections of PRR that are required for moving to beta are missing.  In particular, metrics informing rollback are missing and other parts of those later sections.

gnufied:
@deads2k we could not merge the implementation changes in 1.22 and hence this is the first time the feature will be going alpha.let me update `kep.yaml`

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: fixed (I think)

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: I added a smaller state diagram inline which clarifies how external-resizer determines new size.

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -94,6 +96,14 @@ As part of this proposal, we are mainly proposing three changes:
 
 1. Relax API validation on PVC update so as reducing `pvc.Spec.Resoures` is allowed, as long as requested size is `>pvc.Status.Capacity`.
 2. Add a new field in PVC API called - `pvc.Status.AllocatedResources` for the purpose of tracking used storage size. By default only api-server or resize-controller can set this field.
+3. Add a new field in PVC API called - `pvc.Status.ResizeStatus` for purpose of tracking status of volume expansion. Following are possible values of newly introduced `ResizeStatus` field:
+   - nil // General default because type is a pointer.
+   - empty // When expansion is complete, the empty string is set by resize controller or kubelet.
```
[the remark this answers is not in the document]
gnufied: It can be nil or empty and don't think we need an explicit default of empty. It might help us later down the road when we want to distinguish between values that were never set vs values that just has been cleared off via resize operation.

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: added

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -203,6 +243,10 @@ If CSI driver does not have `GET_VOLUME` controller capability(or `ControllerGet
   so it should not break any of existing automation. This means that if `pvc.Status.AllocatedResources` is available it will be
   used for calculating quota.
 
+  One more thing to keep in mind is - enabling this feature in kubelet while keeping it disabled in external-resizer will cause
```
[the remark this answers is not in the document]
gnufied: added a note about how it will work with older kubelet.

PR #2809, opened 2021-07-05
Update use of Job annotation instead of finalizer

alculquicondor:
And add note about keeping the legacy behavior

#2307

/sig apps


soltysh (approved):
/lgtm
/approve

PR #2874, opened 2021-08-18
Graduate JobTrackingWithFinalizers to beta in 1.23

alculquicondor:
Ref #2307

Updates graduation criteria and completes PRR.

### DIFF
--- a/keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
+++ b/keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
@@ -30,7 +30,7 @@
   - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
     - [How can this feature be enabled / disabled in a live cluster?](#how-can-this-feature-be-enabled--disabled-in-a-live-cluster)
     - [Does enabling the feature change any default behavior?](#does-enabling-the-feature-change-any-default-behavior)
-    - [Can the feature be disabled once it has been enabled (i.e. can we roll back](#can-the-feature-be-disabled-once-it-has-been-enabled-ie-can-we-roll-back)
+    - [Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?](#can-the-feature-be-disabled-once-it-has-been-enabled-ie-can-we-roll-back-the-enablement)
     - [What happens if we reenable the feature if it was previously rolled back?**](#what-happens-if-we-reenable-the-feature-if-it-was-previously-rolled-back)
     - [Are there any tests for feature enablement/disablement?**](#are-there-any-tests-for-feature-enablementdisablement)
   - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
@@ -331,16 +331,16 @@ the owner reference.
   - Removal of finalizer when feature gate is disabled.
   - Support for [Indexed Jobs](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
 - Tests: unit, integration.
+
 #### Alpha -> Beta Graduation
 
-- Processing 5000 Pods per minute across any number of Jobs, with Pod creation
-  having higher priority than status updates, using [Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control).
-  We rename the existing workload-high priority to workload-medium and, when
-  the feature gate JobTrackingWithFinalizers is enabled, we add a workload-high
-  priority that matches the job-controller ServiceAccount for Pod creation.
+- Pod processing throughput per minute (mix of creating and counting finished Pods),
+  assuming an average Job .spec.parallelism=10.
+  - Up to 2500 Pods (~3000 queries) for a 50 QPS client limit for the job controller.
+  - Up to 5000 (~6000 queries) Pods for a 100 QPS client limit for the job controller.
 - Ensure that tracking Jobs with big number of Pods doesn't cause starvation of
   smaller jobs.
-- Metrics for latency and errors
+- Metrics for latency, counting updates and errors.
 - Job E2E tests are in Testgrid with the feature enabled and linked in KEP
 
 #### Beta -> GA Graduation
@@ -416,8 +416,7 @@ No implications to node runtime.
   - Pods removed by the user or other controllers count towards failures or
     completions.
 
-#### Can the feature be disabled once it has been enabled (i.e. can we roll back
-  the enablement)?
+#### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
   
   Yes.
   The job controller removes finalizers in this case.
@@ -436,8 +435,6 @@ No implications to node runtime.
 ### Rollout, Upgrade and Rollback Planning
 
 #### How can a rollout fail? Can it impact already running workloads?
-  Try to be as paranoid as possible - e.g., what if some components will restart
-   mid-rollout?
   
   The change doesn't affect running Pods. If the component restarts
   mid-rollout into an older version, the Job controller switches to tracking
@@ -453,8 +450,10 @@ No implications to node runtime.
 #### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
   
   Integration tests cover feature gate disablement.
-  A manual upgrade->downgrade->upgrade flow will be executed to ensure that a
-  running Job falls back to tracking without finalizers.
+
+  A manual upgrade->downgrade->upgrade flow will be executed prior to graduation
+  to ensure that a running Job falls back to tracking without finalizers. The
+  KEP will be updated with the findings of the test.
 
 #### Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
 fields of API types, flags, etc.?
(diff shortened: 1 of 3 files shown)


keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -305,28 +307,33 @@ the owner reference.
...
   - Support for [Indexed Jobs](https://git.k8s.io/enhancements/keps/sig-apps/2214-indexed-job)
-- Tests: unit, integration, E2E
-
+- Tests: unit, integration.
 #### Alpha -> Beta Graduation
 
 - Processing 5000 Pods per minute across any number of Jobs, with Pod creation
-  having higher priority than status updates. This might depend on
-  [Priority and Fairness](https://git.k8s.io/enhancements/keps/sig-api-machinery/1040-priority-and-fairness).
+  having higher priority than status updates, using [Priority and Fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control).
```
alculquicondor, line 314: @lavalamp this is what we discussed offline. This is the behavior I would like, but maybe this is the wrong approach due to backwards compatibility. Let me know if you have better suggestions.

lavalamp: 5000/60 = 83 QPS; if each request is 100ms, that's a concurrency of ~9, which is not very much. You probably don't even need to adjust the default APF levels (you do need to adjust or turn off the client side rate limit).

Note that pod creations have a bunch of follow-on requests and there's not much point in giving this special priority to create them if the rest of the system can't then keep up.

What you should do instead is look at the shares given to workload-high and the other PLs and figure out what overall in-flight limit needs to be such that workload high ends up with at least that much concurrency, and document that, so operators know how big of a cluster they need to get that throughput.

Note that to get your maximum throughput, you need to multiplex your requests across all apiservers, which is challenging in some deployments. But it may be more important than e.g. making a new PL.

alculquicondor: If I determine that we need an update of the existing PLs, should that be (and can it be) dependent on the feature gate for this KEP?

lavalamp: That would be very awkward, so I'd recommend no. Not every k8s cluster needs this to go this fast, so I think in that case we'd most likely ask you to just document what needs to be done. But let's see what that is before making any decisions.

wojtek-t: > I think we are good with the current default APF configuration. And we can recommend users to set the client QPS to 100 if they expect to have 2500 pods finishing per minute or disable it completely if they expect to process more than 5k pods per minute.

SGTM

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -305,28 +330,33 @@ the owner reference.
...
+  We rename the existing workload-high priority to workload-medium and, when
```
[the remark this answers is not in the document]
wojtek-t: +1 for removing it - please proceed

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,105 @@ _This section must be completed when targeting alpha to a release._
...
   Yes.
   
   - Removing terminated Pods doesn't affect Job status.
   - Pods removed by the user or other controllers count towards failures or
     completions.
 
-* **Can the feature be disabled once it has been enabled (i.e. can we roll back
-  the enablement)?**
+#### Can the feature be disabled once it has been enabled (i.e. can we roll back
+  the enablement)?
```
wojtek-t, line 420: nit: move to preview line

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,105 @@ _This section must be completed when targeting alpha to a release._
...
+  - An increase in `job_sync_duration_seconds`. Users should expect a higher
+    duration than previous versions of the job controller due to the new API
+    calls.
+  - Stale `job_sync_total` or `job_finished_total`.
 
-### Monitoring Requirements
+#### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+  
+  Integration tests cover feature gate disablement.
+  A manual upgrade->downgrade->upgrade flow will be executed to ensure that a
```
wojtek-t, line 456: Can you please ack that those will get run before feature will graduate? And that the KEP will get updated with their results [sth like: https://github.com/kubernetes/enhancements/pull/2538 ]

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,105 @@ _This section must be completed when targeting alpha to a release._
...
-  Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
-  checking if there are objects with field X set) may be a last resort. Avoid
-  logs or events for this purpose.
+### Monitoring Requirements
 
-* **What are the SLIs (Service Level Indicators) an operator can use to determine 
-the health of the service?**
-  - [ ] Metrics
-    - Metric name:
+#### How can an operator determine if the feature is in use by workloads?
```
wojtek-t, line 466: I think some metric would be useful here. How about a counter summarizing how many pod additions to uncountedTerminatedPods were done?

wojtek-t (commented):
I'm fine with it modulo the comments I added.

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Ack

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Added, but rather when they are removed from `uncountedTerminatedPods`

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Done

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Updated.

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
wojtek-t: Sure - that's better.

wojtek-t:
@alculquicondor - I'm generally fine with this; please work with SIG approvers to have that approved (@soltysh :) ) and I will approve after this

soltysh (approved):
/lgtm
/approve

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,106 @@ _This section must be completed when targeting alpha to a release._
...
   create new Pods to fulfill completions. But this is no different from
   existing behavior with the legacy tracking.
 
-* **What happens if we reenable the feature if it was previously rolled back?**
+#### What happens if we reenable the feature if it was previously rolled back?**
 
   Existing Jobs are tracked with legacy Pods.
 
-* **Are there any tests for feature enablement/disablement?**
+#### Are there any tests for feature enablement/disablement?**
```
wojtek-t, line 431: Have those been added? If not, can we add a similar comment as for upgrade tests?

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,106 @@ _This section must be completed when targeting alpha to a release._
...
+    duration than previous versions of the job controller due to the new API
+    calls.
+  - Stale `job_sync_total` or `job_finished_total`.
 
-* **Is the rollout accompanied by any deprecations and/or removals of features, APIs, 
-fields of API types, flags, etc.?**
-  Even if applying deprecation policies, they may still surprise some users.
+#### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+  
+  Integration tests cover feature gate disablement.
```
wojtek-t, line 452: How about enablement?

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
```diff
@@ -380,112 +408,106 @@ _This section must be completed when targeting alpha to a release._
...
 
-### Monitoring Requirements
+  A manual upgrade->downgrade->upgrade flow will be executed prior to graduation
```
wojtek-t, line 454: prior to Beta graduation

wojtek-t (commented):
Just some minor comments - LGTM after applied.

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Yes, we have them. Updated.

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: That too. Updated.

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Fixed

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
wojtek-t: Can you please add links?

wojtek-t (commented):
Sorry - one last nit

keps/sig-apps/2307-job-tracking-without-lingering-pods/README.md
alculquicondor: Done :)

PR #2875, opened 2021-08-18
Mark 1693 warnings KEP as implemented

liggitt:
fixes #1693


lavalamp:
/approve

Bot doesn't like that extra d :)

PR #2927, opened 2021-09-01
pkg/repo: only parse kep.yaml files as KEPs

spiffxp:
`kepctl query` was attempting to parse non-kep.yaml files as KEPs and
erroring out before loading all KEPs.

I opted to filter out non-kep.yaml files from the loading code rather
than allow kepctl to continue loading files, so as to incentivize
fixing kep.yaml files to be valid ASAP


justaugustus (approved):
/lgtm
/approve

PR #2982, opened 2021-09-21
KEP-555: omit managed fields from audit log

tkashem:


keps/sig-api-machinery/555-server-side-apply/README.md
```diff
@@ -286,6 +287,67 @@ The conversion between the two and creating the diff was complex and would have
...
+	// +optional
+	OmitManagedFields bool `json:"omitManagedFields,omitempty"`
+}
+
+type PolicyRule struct {
+	// +optional
+	OmitManagedFields *bool `json:"omitManagedFields,omitempty"`
+}
+```
+The above API changes will be introduced in `v1`, `v1beta1` and `v1alpha1` of `audit.k8s.io`
```
sttts, line 309: we never roundtrip these, do we? So technically we could do this in v1 only.

sttts:
Sgtm.

/assign @deads2k @liggitt

keps/sig-api-machinery/555-server-side-apply/README.md
lavalamp: But there's no harm in doing it in all, no?

lavalamp:
/approve
/lgtm

/hold

(hold for someone from sig auth to ack)

liggitt:
looks reasonable to me

keps/sig-api-machinery/555-server-side-apply/README.md
sttts: only that we encourage use of v1 by not implementing new stuff in old versions.

keps/sig-api-machinery/555-server-side-apply/README.md
```diff
@@ -286,6 +287,67 @@ The conversion between the two and creating the diff was complex and would have
...
+
+A new field `OmitManagedFields` is added to both `Policy` and `PolicyRule` making the following possible:
+- `Policy.OmitManagedFields` sets the default policy for omitting managed fields globally.
+  - the default value is `false`, managed fields are not omitted, this retains the current behavior.
```
ritazh, line 313: Any plans for changing the default value in a future release since the fields may not be very useful? making it opt-out instead?

tkashem: not that i am aware of, primarily we didn't want to quietly drop managed fields from audit log. The level `RequestResponse` stipulates that the response object in its "entirety" is being audited, so if we are to drop any fields from the audit log then I think it should be an opt-in by the operator.

sttts:
sig-auth has reviewed with @ritazh and @liggitt.

/hold cancel

PR #3062, opened 2021-11-29
Update Efficient Watch Resumption KEP to Stable

wojtek-t:
Ref #1904 

/sig api-machinery

/assign @deads2k @lavalamp

### DIFF
--- a/keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
+++ b/keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
@@ -300,8 +300,8 @@ Alpha should provide basic functionality covered with tests described above.
 #### Beta -> GA Graduation
 
 - Enabled in Beta for at least two releases without complaints
-- Rolling-upgrade of kube-apiservers in 5k-node cluster test is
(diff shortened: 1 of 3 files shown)


keps/sig-api-machinery/1904-efficient-watch-resumption/README.md
```diff
@@ -300,8 +300,8 @@ Alpha should provide basic functionality covered with tests described above.
 #### Beta -> GA Graduation
 
 - Enabled in Beta for at least two releases without complaints
-- Rolling-upgrade of kube-apiservers in 5k-node cluster test is
-   automated and running periodically.
+- Benefits of the feature during rolling upgrade of kube-apiserver at scale
+   (thousands of nodes) proven in production
```
wojtek-t, line 304: We still aren't testing HA clusters in upstream. Given that it will take more time (we're still trying to converge on what exact framework we should be using) and the fact that:
- we're somewhat-continuously testing it as prt of GKE
- we've already seen a proven effect/benefit from it on real production clusters

I don't want to block GA-ing this feature on that.

PR #3164, opened 2022-01-20
Fix link for gateway api implementations

MadhavJivrajani:
Fixes a small link in the gateway API KEP.


thockin (commented):
Thanks!

/lgtm
/approve

PR #3171, opened 2022-01-21
exclude kubeadm from PRR

neolit123:
- Update the PRR readme with a note that kubeadm is out of scope.
- Remove existing PRR approvals for kubeadm in
keps/prod-readiness/sig-cluster-lifecycle. These were rubberstamps and the KEPs do not contain an actual PRR.
- Update kepval/approval.go to skip kubeadm KEP paths.

some context on why this change is being proposed:
https://github.com/kubernetes/enhancements/issues/2067#issuecomment-1018408831

> this has been a recurring topic every release

role book PR:
https://github.com/kubernetes/sig-release/pull/1831

### DIFF
--- a/keps/prod-readiness/README.md
+++ b/keps/prod-readiness/README.md
@@ -19,4 +19,8 @@ items.
 Full documentation on the production readiness review process can be found on
 the SIG Architecture page [here][prod-readiness].
 
+**Note:** The kubeadm tool is considered out-of-tree and does not fall under PRR.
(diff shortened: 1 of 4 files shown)


ehashman:
PRR team consensus in https://kubernetes.slack.com/archives/CPNHUMN74/p1643393111358089 is that you should still assign a PRR reviewer so that we can confirm kubeadm KEPs are out of scope of PRR.

/close

PR #3201, opened 2022-02-01
ExecProbeTimeout PRR and re-promoting to GA

SergeyKanzhelev:
I suggest we treat ExecProbeTimeout as it enters GA now and filled out the PRR. I suggest we implement two new improvements for the feature:

- Probe duration metrics: https://github.com/kubernetes/kubernetes/issues/101035
- Container events on the exceeded timeout

Also making documentation on workload migration to be a GA blocker.

/sig node
KEP: #1972
/priority important-soon


johnbelamaric:
OK, PRR looks good. Thanks Sergey

derekwaynecarr:
this looks good to me.

/approve
/lgtm

PR #3240, opened 2022-03-15
Graduate SeccompDefault feature to beta

saschagrunert:
Updating the SeccompDefault feature to beta, targeting for the v1.25 cycle.

Refers to https://github.com/kubernetes/enhancements/issues/2413


keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -143,14 +143,33 @@ be extended to cover the new behavior if enabled.
...
-- [ ] No major bugs reported in the previous cycle
+- [x] Enable the feature per default
+- [x] Making the feature API aware
+- [x] No major bugs reported in the previous cycle
+
+##### API Awareness
+
+We add conversion methods to the API Server with the graduation of the feature
+to beta. The beta graduation does not mean that it's enabled by default, which
+got introduced in Kubernetes v1.24.
```
cbandy, line 160: Is this accurate? [This KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/3136-beta-apis-off-by-default#non-goals) seems to say that feature gates were not part of that policy:

> Feature gates control new features (not just new APIs) and they are on by default for beta features. This KEP is not changing the lifecycle flow for feature gates.

saschagrunert: I think so, looking at the conversation in: https://github.com/kubernetes/kubernetes/pull/109819#discussion_r871637251

saschagrunert:
@kubernetes/sig-architecture PTAL as well :)

dims:
@saschagrunert ACK. will do later this week unless John or Derek are able to before me

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -143,14 +143,33 @@ be extended to cover the new behavior if enabled.
...
+profile as well as handling upgrade scenarios. If the feature is enabled, then
+created workloads not specifying any `securityContext.seccompProfile.type` will
+default to `RuntimeDefault`.
+
+This also means for an upgrade of a cluster, where the feature has not been
+enabled, that all empty `seccompProfile.type` fields will be converted to
+`Unconfined` to explicitly exclude already running workloads.
+
+Changing seccomp profiles of running workloads requires a restart of them, which
+incorporates with the previously described conversion.
```
dims, line 172: Curious, what happens under this case?
- existing cluster (from say 1.24) `seccompProfile.type` is empty.
- during upgrade (to 1.25) the cluster operator switches on this flag
- after upgrade they figure out there is a problem and switches off this flag

what would be the end result of the string in `seccompProfile.type` after steps above?

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -143,14 +143,33 @@ be extended to cover the new behavior if enabled.
...
+##### API Awareness
+
+We add conversion methods to the API Server with the graduation of the feature
+to beta. The beta graduation does not mean that it's enabled by default, which
+got introduced in Kubernetes v1.24.
+
+The API Awareness allows us giving users direct feedback about the chosen seccomp
+profile as well as handling upgrade scenarios. If the feature is enabled, then
+created workloads not specifying any `securityContext.seccompProfile.type` will
+default to `RuntimeDefault`.
```
liggitt, line 165: The kubelet feature added in a previous release was opt-in. Will this be opt-in as well? If so, more than a feature flag is needed, since the opt-in goes beyond feature graduation. 

It also means this defaulting should only be done on pod creation, not in API defaulting for existing pods. This is because:
1. it depends on configuration, and conversion should be stateless
2. applying new defaults to existing fields in pods on API server upgrade triggers instant restarts of pods across nodes, which causes outages (https://github.com/kubernetes/kubernetes/issues/63814)

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -143,14 +143,33 @@ be extended to cover the new behavior if enabled.
...
+
+This also means for an upgrade of a cluster, where the feature has not been
+enabled, that all empty `seccompProfile.type` fields will be converted to
+`Unconfined` to explicitly exclude already running workloads.
```
liggitt, line 169: Existing objects should not default an existing field on upgrade. See above comment on triggering kubelet detection of changes that trigger restarts.

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: It would be still `RuntimeDefault`, which would not solve the issue for the user. 

What we could do is to enable it only for new workloads and converting everything else to `Unconfined` if not set.

keps/sig-node/2413-seccomp-by-default/README.md
liggitt, line 165: What are the implications for defaulting securityContext and seccomp profile for workloads which do not specify an operating system, but are scheduled to windows notes? At best that seems misleading, at worst does it get rejected by the windows kubelet?

keps/sig-node/2413-seccomp-by-default/README.md
liggitt: I think we're limited to setting the field on new objects (just on pods?) at creation time

saschagrunert: Sounds good, I'll reflect that in the KEP. What would be a reason to only set it on pods rather than all containers (including init containers)?

liggitt: sorry, I meant just on Pod objects, rather than on other pod-spec-containing objects like Deployments, ReplicaSets, etc

saschagrunert: Yes, technically we only set it on the pod level, from which containers inherit.

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: Looking at the [Windows docs](https://kubernetes.io/docs/concepts/windows/intro/) it seems that seccomp is supported by the currently available container runtimes. So I would not distinguish the field by the host OS or architecture to keep it simple. It could be still possible that seccomp is not supported from the Kernel or build time perspective. Container runtimes should then ignore the field and unset it in the OCI spec (which is the current behavior).

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -143,14 +143,31 @@ be extended to cover the new behavior if enabled.
...
+- [x] Enable the feature per default
+- [x] Making the feature API aware
+- [x] No major bugs reported in the previous cycle
+
+##### API Awareness
+
+We add conversion methods to the API Server with the graduation of the feature
+to beta. This means that the Kubelet feature gate `SeccompDefault` will be
+enabled by default, while users still have to enable the kubelet configuration
+`SeccompDefault` or use the `--seccomp-default` CLI flag.
```
saschagrunert, line 161: @liggitt @dims from a technical perspective, assuming that the conversion of new workloads will be done on creation in the API Server: How could the information flow that the feature is enabled in the kubelet to the API Server?

saschagrunert: cc @kubernetes/sig-architecture @kubernetes/sig-node-proposals any thoughts on that?

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -280,53 +297,32 @@ _This section must be completed when targeting beta graduation to a release._
...
 - **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
 
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+  None
```
deads2k, line 312: Once you know how many seccomp or related failures you have, then there will likely be enough to suggest a guarantee here.

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -378,26 +375,22 @@ _This section must be completed when targeting beta graduation to a release._
...
-    - Detection: How can it be detected via metrics? Stated another way:
-      how can an operator troubleshoot without logging into a master or worker node?
-    - Mitigations: What can be done to stop the bleeding, especially for already
-      running user workloads?
-    - Diagnostics: What are the useful log messages and their required logging
-      levels that could help debug the issue?
-      Not required until feature graduated to beta.
-    - Testing: Are there any tests for failure mode? If not, describe why.
+
+  None
```
deads2k, line 382: If the feature is disabled on kubelets, but the seccomp.type is set, what happens to the pod?  Are the values not honored?  That seems like a significant problem.

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -280,53 +297,32 @@ _This section must be completed when targeting beta graduation to a release._
...
 - **What are the SLIs (Service Level Indicators) an operator can use to determine
   the health of the service?**
 
-  - [ ] Metrics
-    - Metric name:
-    - [Optional] Aggregation method:
-    - Components exposing the metric:
-  - [ ] Other (treat as last resort)
-    - Details:
+  There are none required because the feature acts as toggle switch.
```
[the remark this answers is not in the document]
saschagrunert: Yep, added two SLIs

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: Agreed and added the SLO for the SLIs.

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: If `seccomp.type` is `RuntimeDefault` and the feature is disabled then it will apply `RuntimeDefault`. I don't see any other failure modes than workload termination or misbehavior (described in the SLIs now). Do you have anything in mind?

keps/sig-node/2413-seccomp-by-default/README.md
```diff
@@ -280,53 +298,48 @@ _This section must be completed when targeting beta graduation to a release._
...
 - **What are the reasonable SLOs (Service Level Objectives) for the above SLIs?**
-  At a high level, this usually will be in the form of "high percentile of SLI
-  per day <= X". It's impossible to provide comprehensive guidance, but at the very
-  high level (needs more precise definitions) those may be things like:
 
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99,9% of /health requests per day finish with 200 code
+  - The workload availability, functionality and health is exactly the same with
```
deads2k, line 328: Is there a metric that tracks pod crashloops?  That would be enough to allow a cluster-admin to notice a problem in aggregate, right?

keps/sig-node/2413-seccomp-by-default/README.md
deads2k: I wasn't sure if there were edges around specifying a seccomp type that doesn't exist.

saschagrunert: Cool, looks like we're then all set. :blush:

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: I'm dropping the API awareness to keep simplicity.

saschagrunert:
@liggitt I dropped the API awareness, so I guess there is no API review required any more

@mrunalp I added the new test plan section, which basically covers a new serial e2e test if the feature gate is enabled by default

@deads2k incorporated the PRR review, PTAL

Thank you all! :pray:

keps/sig-node/2413-seccomp-by-default/README.md
saschagrunert: Yep, changed the section to mention that.

PR #3381, opened 2022-06-10
Fix failing PRR approvers checks

wojtek-t:
Verification got broken by merging: https://github.com/kubernetes/enhancements/pull/3281


OWNERS_ALIASES
```diff
@@ -188,8 +188,8 @@ aliases:
     - johnbelamaric
     - deads2k
     - wojtek-t
-# prod-readiness-approvers-emeritus:
-#    - ehashman
+  prod-readiness-approvers-emeritus:
+    - ehashman
```
wojtek-t, line 192: @dims @johnbelamaric - This PR is failing validation, because our testing is fetching OWNERS_ALIASES from head.
So merging this PR (if we can somehow do that manually) would fix the problem.

If we can't do that - we need to temporarily re-add Elana to prod-readiness-approvers back, and only then merge this PR.

JamesLaverack (approved):
/lgtm

api/groups.go
```diff
@@ -156,6 +159,8 @@ func (f *RemoteGroupFetcher) FetchPRRApprovers() ([]string, error) {
 
 	var result []string
 	result = append(result, config.Data[f.PRRApproversAlias]...)
+	// TODO: Figre out if we want to treat emeritus approvers differently.
+	result = append(result, config.Data[f.PRRApproversEmeritusAlias]...)
```
wojtek-t, line 163: Actually - if it's impossible to merge manually, then what we can do is to:
- append explicitly "ehashman" here
- get rid of it in a ubsequent PR 

Let me do that now maybe.

OWNERS_ALIASES
dims: +1

api/groups.go
dims: thanks!

PR #3450, opened 2022-08-02
KEP 3448: support removing ephemeral containers on completion

infiniteregrets:
- One-line PR description: This KEP proposes to add a mechanism to remove ephemeral containers once they complete.

- Issue link: https://github.com/kubernetes/enhancements/issues/3448

- Other comments: This is a WIP, while I am still figuring out how the KEP process works and what would be the best way to put the idea here, please feel free to correct me and help me finish this proposal as there are so many things I am not completely sure of. Would appreciate any help/corrections. Thanks!


tallclair:
/assign @verb 

I'd appreciate if this KEP could flesh out the motivation section a bit more. As is, it's not clear to me why we want this change.

PR #3462, opened 2022-08-11
KEP 1489: Mark CSI migration for Cinder as done

jsafrane:
CSI migration for OpenStack Cinder was GA in 1.24, mark the KEP as `implemented`.

And close the enhancement issue:
Fixes: #1489

cc @xing-yang @msau42


jsafrane:
cc @Jiawei0227 for tracking

PR #3494, opened 2022-09-04
Fix the broken link to scheduler extenders design

SataQiu:
Fix the broken(404) link for scheduling framework


alculquicondor:
/retitle Fix the broken link to scheduler extenders design

PR #3585, opened 2022-10-05
961: maxUnavailable for StatefulSets

krmayankk:
Updated milestone and dates
Added graduation criteria for Beta


soltysh (changes requested):
You're missing the following:
- `keps/prod-readiness/sig-apps/961.yaml` updated to cover beta stage
- you need to update current README.md with the recent template see https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md
- make sure this passes all the pre-checks

PR #3594, opened 2022-10-06
reduce scope for alpha extending metrics stability

logicalhan:
This amends KEP-3498 to reduce scope slightly for alpha.

/sig instrumentation
/assign @dashpole @dgrisonnet


dashpole:
Is this just because of the large number of existing alpha metrics?

PR #3596, opened 2022-10-06
amend test section for component slis

logicalhan:
Since the feature will be default disabled, the e2e tests will be a beta requirement.


dashpole (approved):
/lgtm

PR #3605, opened 2022-10-10
remove dynamic kubelet config completely

SergeyKanzhelev:
/sig node
KEP: #281
/kind cleanup


SergeyKanzhelev:
Since the work was planned and ongoing and there were no feedback recieved that will change plans, I'd suggest to take this KEP in 1.26

PR: https://github.com/kubernetes/kubernetes/pull/112643

PR #3634, opened 2022-10-21
Fix link rot

jboxman:
Fixes a few KEP links that are no longer valid

But maybe these should point to blob instead of tree?


sftim (commented):
/ok-to-test

PR #3654, opened 2022-11-10
WIP - KEP-740 service account external signing

ivelichkovich:
- One-line PR description:
Revisiting KEP-740 with an updated design to avoid a potential race condition.

- Issue link:

- Other comments:
- Several years ago we had consensus on this, would like to get consensus and move along to implementable but this is still a draft


enj:
Please start fresh with a new KEP based on the latest template.  The KEP requirements have dramatically changed over the last few years.  I consider this KEP to be withdrawn due to age.

PR #3667, opened 2022-11-24
Update API proposal for KEP 3157

wojtek-t:
Ref #3157 

/assign @deads2k @lavalamp 

/cc @p0lyn0mial @jpbetz @liggitt

### DIFF
--- a/keps/sig-api-machinery/3157-watch-list/README.md
+++ b/keps/sig-api-machinery/3157-watch-list/README.md
@@ -85,7 +85,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Proposal](#proposal)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
-  - [Required changes for a WATCH request with the RV=&quot;&quot; and the ResourceVersionMatch=MostRecent](#required-changes-for-a-watch-request-with-the-rv-and-the-resourceversionmatchmostrecent)
+  - [Required changes for a WATCH request with the SendInitialEvents=true](#required-changes-for-a-watch-request-with-the-sendinitialeventstrue)
     - [API changes](#api-changes)
     - [Important optimisations](#important-optimisations)
     - [Manual testing without the changes in place](#manual-testing-without-the-changes-in-place)
@@ -179,7 +179,7 @@ The kube-apiserver is vulnerable to memory explosion.
 The issue is apparent in larger clusters, where only a few LIST requests might cause serious disruption.
 Uncontrolled and unbounded memory consumption of the servers does not only affect clusters that operate in an 
 HA mode but also other programs that share the same machine.
-In this KEP we propose a potential solution to this issue.
+In this KEP we propose a solution to this issue.
 
 ## Motivation
 
@@ -257,7 +257,7 @@ The "Design Details" section below is for the real
 nitty-gritty.
 -->
 
-In order to lower memory consumption while getting a list of data and make it more predictable, we propose to use consistent streaming from the watch-cache instead of paging from etcd.
+In order to lower memory consumption while getting a list of data and make it more predictable, we propose to use streaming from the watch-cache instead of paging from etcd.
 Initially, the proposed changes will be applied to informers as they are usually the heaviest users of LIST requests (see [Appendix](#appendix) section for more details on how informers operate today).
 The primary idea is to use standard WATCH request mechanics for getting a stream of individual objects, but to use it for LISTs.
 This would allow us to keep memory allocations constant.
@@ -266,17 +266,17 @@ plus a few additional allocations, that will be explained later in this document
 The rough idea/plan is as follows:
 
 - step 1: change the informers to establish a WATCH request with a new query parameter instead of a LIST request.
-- step 2: upon receiving the request from an informer, contact etcd to get the latest RV. It will be used to make sure the watch cache has seen objects up to the received RV. This step is necessary and ensures we will serve consistent data, even from the cache.
-- step 2a: send all objects currently stored in memory for the given resource.
+- step 2: upon receiving the request from an informer, compute the RV at which the result should be returned (possibly contacting etcd if consistent read was requested). It will be used to make sure the watch cache has seen objects up to the received RV. This step is necessary and ensures we will meet the consistency requirements of the request.
+- step 2a: send all objects currently stored in memory for the given resource type.
 - step 2b: propagate any updates that might have happened meanwhile until the watch cache catches up to the latest RV received in step 2.
 - step 2c: send a bookmark event to the informer with the given RV.
 - step 3: listen for further events using the request from step 1.
 
 Note: the proposed watch-list semantics (without bookmark event and without the consistency guarantee) kube-apiserver follows already in RV="0" watches.
 The mode is not used in informers today but is supported by every kube-apiserver for legacy, compatibility reasons.
-A watch started with RV="0" may return stale. It is possible for the watch to start at a much older resource version that the client has previously observed, particularly in high availability configurations, due to partitions or stale caches
+A watch started with RV="0" may return stale data. It is possible for the watch to start at a much older resource version that the client has previously observed, particularly in high availability configurations, due to partitions or stale caches.
 
-Note 2: informers need consistent lists to avoid time-travel when switching to another HA instance of kube-apiserver with outdated/lagging watch cache.
+Note 2: informers need consistent lists to avoid time-travel when initializing after restart to avoid time travel in case of switching to another HA instance of kube-apiserver with outdated/lagging watch cache.
 See the following [issue](https://github.com/kubernetes/kubernetes/issues/59848) for more details.
 
 
@@ -310,7 +310,7 @@ required) or even code snippets. If there's any ambiguity about HOW your
 proposal will be implemented, this is the place to discuss them.
 -->
 
-### Required changes for a WATCH request with the RV="" and the ResourceVersionMatch=MostRecent
+### Required changes for a WATCH request with the SendInitialEvents=true
 
 The following sequence diagram depicts steps that are needed to complete the proposed feature.
 A high-level overview of each was provided in a table that follows immediately the diagram.
@@ -328,11 +328,11 @@ Whereas further down in this section we provided a detailed description of each 
  </tr>
 <tr>
   <th>2.</th>
-  <th>The watch cache contacts etcd for the most up-to-date ResourceVersion.</th>
+  <th>If needed, the watch cache contacts etcd for the most up-to-date ResourceVersion.</th>
 </tr>
 <tr>
   <th>2a.</th>
-  <th>The watch cache starts streaming initial data. The data it already has in memory.</th>
+  <th>The watch cache starts streaming initial data it already has in memory.</th>
 </tr>
 <tr>
   <th>2b.</th>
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -447,17 +447,35 @@ It replaces its internal store with the collected items (syncWith) and reuses th
...
+    //   is send and watch stream is continued to be send.
+    //   If RV is unset, this is interpreted as "consistent read" and the
+    //   bookmark event is send when the state is synced to the moment when
+    //   request started being processed.
+    // - ResourceVersionMath = Exact
+    //   Unsupported error is returned.
+    // - ResourceVersionMatch unset (or set to any other value)
+    //   BadRequest error is returned.
+    SendInitialEvents bool
+}
```
wojtek-t, line 478: If we would agree that this is the right API, then the proposal needs to be updated in couple other places, but I won't do that, until we agree for the API change.

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -447,17 +447,35 @@ It replaces its internal store with the collected items (syncWith) and reuses th
...
+type ListOptions struct {
+    ...
+
+    // SendInitialEvents, when set together with Watch option,
+    // being the watch stream with synthetic init events to build the
+    // whole state of all resources followed by a synthetic "Bookmark"
+    // event containing a ResourceVersion after which the server
+    // continues streaming events.
+    //
+    // When SendInitialEvents option is set, we require ResourceVersionMatch
```
p0lyn0mial, line 462: It will give users even more options to establish a watch request. Not sure if it is good, they might be even more confused. Especially those that are used to the current semantics.

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -447,17 +447,35 @@ It replaces its internal store with the collected items (syncWith) and reuses th
...
+    // option to also be set. The semantic of the watch request is as following:
+    // - ResourceVersionMatch = NotOlderThan
```
p0lyn0mial, line 464: okay, so this is what we called previously `SendInitialEvent and RV>0`

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -447,17 +447,35 @@ It replaces its internal store with the collected items (syncWith) and reuses th
...
+    //   It starts with sending initial events for all objects (at some resource
+    //   version), potentially followed by an event stream until the state
+    //   becomes synced to a resource version as fresh as the one provided by
+    //   the ResourceVersion option. At this point, a synthetic bookmark event
+    //   is send and watch stream is continued to be send.
+    //   If RV is unset, this is interpreted as "consistent read" and the
```
p0lyn0mial, line 470: I think this is be strange, setting `NotOlderThan without an RV` would give me a consistent read. At least for me, this would be unintuitive. I think in that case I would prefer to set `MostRecent with an RV`.

I think I'd forbid `NotOlderThan with an empty RV or an RV set to 0`

keps/sig-api-machinery/3157-watch-list/README.md
wojtek-t: I don't agree. What I wrote below is that effectively only `NotOlderThan` is supported. So you have to pass that specific option.
What we get from it, is that we make it very explicit, which forces people (at least in theory) to understand what they actually get.

keps/sig-api-machinery/3157-watch-list/README.md
wojtek-t: I don't fully agree here.

1. RV="".
Unset RV for List operations is already used as "consistent read". So people should already be familiar with it.
However, here what we're saying is that you will get the state that is consistent, but the bookmark event that you will get will actually only satisfy the "NotOlderThan" semantic.

If you look at your implementation (and as we were discussing in the past, I don't think we can do anything better) what we're doing is:
- list from etcd to get the current RV (from which the consistent read is happening)
- once cache is synced to that value, start sending state

But given everything here is asynchronous, nothing really can't prevent you from the situation thatn when you get the RV from etcd, in the meantime something else happened and watchcache is already synced to the RV higher than what you get from etcd.
So we will effectively return bookmark at RV higher that what we though we should do.

So I think that this actually is inline with the semantic that we want to sell.


2. RV=0

This is potentially debatable. But note that my description doesn't mention RV=0 anywhere. It's treated as every other value of resource version - it's just that any RV is greater than 0, so state from any point in time satisfies that.
So forbidding RV=0, doesn't buy us anything here  - it neither won't change anything in the explanation above (so it will take the same effort for users to understand it whether we support it or not) nor it will change the implementation.

If we want to deprecate RV=0, we should approach that more holistically and do that across everywhere, not attack in in some places imho.

keps/sig-api-machinery/3157-watch-list/README.md
lavalamp: I agree w/ Wojtek

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -447,17 +447,35 @@ It replaces its internal store with the collected items (syncWith) and reuses th
...
+    //   bookmark event is send when the state is synced to the moment when
+    //   request started being processed.
+    // - ResourceVersionMath = Exact
```
lavalamp, line 473:
```suggestion
    // - ResourceVersionMatch = Exact
```

keps/sig-api-machinery/3157-watch-list/README.md
lavalamp: What happens to the existing way of requesting a watch which sends the existing entire collection first?

keps/sig-api-machinery/3157-watch-list/README.md
lavalamp: I agree w/ Wojtek here also

keps/sig-api-machinery/3157-watch-list/README.md
lavalamp: I was going to say option 2, but since people have to start doing something extra to pay attention to the bookmarks anyway, maybe it's OK to deprecate it?

liggitt: thinking about how the current behavior of watching from rv="" or rv="0" interacts with the behavior controlled by the existing `resourceVersionMatch` and `allowWatchBookmarks` options, and the proposed sendInitialEvents option...

it would be _really_ nice if rv="" / rv="0" was less of a special case

* It always sends initial events, which implicitly matches the behavior of `sendInitialEvents=true`.
  * It feels silly to not allow `?watch=true&resourceVersion=0&sendInitialEvents=true&...` when we're actually sending initial events anyway.
  * it also feels silly to not allow `?watch=true&resourceVersion=0&sendInitialEvents=false&...`, when that would let us do less work and would let someone avoid receiving all existing items as `Added` events if they really just wanted to start a watch stream and didn't care what version they started at
* > It doesn't send any bookmark.
  * It could if `allowWatchBookmarks=true` was sent... that seems useful for someone wanting to know when the initial set of Added events was done
* > It's partially consistent across etcd3 and watchcache
  * Since we don't advertise whether watch cache is on for a given resource type, and we don't allow use of `resourceVersionMatch` with watch currently, I don't see how callers could rely on the "current state" etcd3 semantic as opposed to the "last synced state" watch cache semantic.
  * weirdly, we [document different semantics](https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch) for watching from rv=0 and rv="", but the watch cache doesn't appear to distinguish them: https://github.com/kubernetes/kubernetes/blob/07801383afe46fde2a62086da1f46266cd393462/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L482-L488
  * I think it would be reasonable for `resourceVersion=0` and `resourceVersion=` to imply `resourceVersionMatch=NotOlderThan` and `sendInitialEvents=true`

I don't see much benefit to deprecating and trying to remove watching with rv=0 (not that we could realistically remove that anyway as a GA behavior), especially if we're contemplating letting watchers explicitly request to stream all initial events anyway.

keps/sig-api-machinery/3157-watch-list/README.md
liggitt: > So forbidding RV=0, doesn't buy us anything here

I agree. As I commented below, I'd actually go in the opposite direction and try to make rv=0 less of a special case... why not let someone watching from rv="" or rv=0 opt into getting bookmarks, or opt out of receiving initial events?

keps/sig-api-machinery/3157-watch-list/README.md
wojtek-t: Thanks for your thoughts Jordan.
I think I agree with most of the above - just adding something thoughts here:

> It always sends initial events, which implicitly matches the behavior of sendInitialEvents=true

That makes perfect sense to me.
My only concern here is that effectively mean that:
- if RV="" or RV="0", sendInitialEvents is somewhat defaulted to true
- otherwise sendInitialEvents is default to false

OTOH, it's what it is now, so if you're ok with that - I'm fine with it too.

> It could if allowWatchBookmarks=true was sent...

Agree.

> weirdly, we document different semantics for watching from rv=0 and rv="", but the watch cache doesn't appear to distinguish them:

Hmm - I must admit that I completely missed that.
It's a bug that is independent from this proposal - I just filed: https://github.com/kubernetes/kubernetes/issues/114845


> I think it would be reasonable for resourceVersion=0 and resourceVersion= to imply resourceVersionMatch=NotOlderThan and sendInitialEvents=true

SGTM


Let me update this proposal to reflect that better (will try to do that later today or early next week).

wojtek-t:
@lavalamp @deads2k @liggitt - thanks a lot for your comments, I incorporated them into this PR. PTAL

keps/sig-api-machinery/3157-watch-list/README.md
wojtek-t: I update the proposal to reflect that.

deads2k:
The API described here makes sense to me.  I'm not entirely clear on how the synthetic is identifiable as "initial events have been sent" versus other normal bookmarks.

wojtek-t:
> The API described here makes sense to me. I'm not entirely clear on how the synthetic is identifiable as "initial events have been sent" versus other normal bookmarks.

@deads2k 
Thanks for the comment David!

My initial idea was that it's simply "the first bookmark within a given watch request".

If you think it's not enough, given that bookmarks are actually the regular API objects of the same type that is being watched (with just ResourceVersion field being set and everything else null-ified), I was also thinking about setting some predefined annotation on those, e.g. sth like
"k8s.io/initial-events-sent": "true"
[we would certainly need a better name - ideas welcome :) ]

I decided not to put it as I didn't feel the strong need for it, but if you think it's necessary - would the above satisfy you?

deads2k:
> My initial idea was that it's simply "the first bookmark within a given watch request".

> I decided not to put it as I didn't feel the strong need for it, but if you think it's necessary - would the above satisfy you?

I think we need something more specific than "the first bookmark" because we've talked about adding bookmarks to things like shutting down the kube-apiserver.  I'd be ok with an annotation["k8s.io/initial-events-sent"]="true.  I'd also be ok with extended the watch.Event with more metadata.

wojtek-t:
Thanks @deads2k  - I updated the PR mentioning the annotation aspect.

lavalamp:
It looks like @deads2k's concern is addressed now.

/lgtm
/approve

wojtek-t:
Thanks @deads2k and @lavalamp !

I also opted-in the corresponding issue for 1.27.
@lavalamp  - I will work with @p0lyn0mial on the code and you may expect API review some time soon :)

PR #3708, opened 2022-12-31
set kep 3070 to implemented

aojea:
set kep 3070 to implemented

Fixes: #3070


thockin (commented):
Thanks!

/lgtm
/approve

PR #3771, opened 2023-01-23
KEP-1040: update GA graduation criteria

MikeSpreitzer:
- One-line PR description:
Update GA graduation criteria: sort, make check-list, refine tuning item.

- Other comments:
- 1. Convert to check list.
- 2. Re-order so that completed items come first.
- 3. Update tuning item to refer to the whole interface (e.g, including logs and metrics) rather than just the API.
- 4. Update the tuning item to replace "auto-tuning" with simply "tuning".  The thinking is that the proper focus is on observability and controllabiility; automation of the tuning was always intended to be a matter for later consideration.

/cc @wojtek-t 
/cc @lavalamp 
/cc @deads2k 
/cc @tkashem 
/cc @marseel

### DIFF
--- a/keps/sig-api-machinery/1040-priority-and-fairness/README.md
+++ b/keps/sig-api-machinery/1040-priority-and-fairness/README.md
@@ -2834,11 +2834,11 @@ Beta:
 
 GA:
 
-- Satisfaction with LIST and WATCH support.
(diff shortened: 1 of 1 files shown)


deads2k:
lgtm as well.

Issue #3784, opened 2023-01-26
A database of Kubernetes-owned names


thockin:
### Enhancement Description

Publish a human- and machine-friendly database of names (labels, annotations, taints, etc) that the Kubernetes project uses, with some information about what they mean and who is responsible for them.

- Kubernetes Enhancement Proposal: TBD
- Discussion Link: https://github.com/kubernetes/enhancements/pull/3727/files#r1087336209
- Primary contact (assignee): @thockin
- Responsible SIGs: sig-architecture
- Enhancement target: N/A (no code)

thockin:
I still hope to do this, but if anyone has time and interest, please let me know!

PR #3787, opened 2023-01-26
KEP-3784: Names DB first draft

thockin:
First draft of a database for kube-owned names.

Issue link: https://github.com/kubernetes/enhancements/issues/3784


keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+// Name describes a single reserved name.
+type Name struct {
+    // Key is the full name being claimed, including the prefix.
+    // Required
+    Key string
+
+    // SIG specifies which SIG owns this name.  Multiple SIGs might use a name,
+    // but only one can own it.
+    // Required
+    SIG SIG
```
aojea, line 248: is this not redundant? in your example you put this file inside a folder with the sig name

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+fact, the only way to find them is to read human-oriented documentation or to
+grep the codebase(s).
+
+This KEP proposes to store these names in a way that is searchable, machine
+parseable, but also relatively human-friendly.
+
+## Motivation
+
+We have too many KEPs and PRs that add annotation- or label-keys without any
+sort of collision detection or governance or documentation.
```
aojea, line 130: and this solve it by creating some kind of git-database based on generating the secondary indexes?

keps/sig-architecture/3784-database-of-names/README.md
thockin: I wrote the record so the same type could be used under multiple indices.  But there are also top-level names and type-centric names that are not clear which SIG owns.

Who owns `k8s.io/hostname` or `k8s.io/metadata.name` ?

keps/sig-architecture/3784-database-of-names/README.md
thockin: We can at least grow a culture of "hey, you should register that name"

aojea: Yeah, I'm all in on that, but I think the key is to make it easy ... Having to PR another repo is always painful, at least to me ... How can make this simpler?

thockin: Yeah, I don't know how to do that except putting it in k/k/staging, which I would be fine with.

wojtek-t: What if this would be autogenerated based on special from of godoc?

It would require additional machinery, but with the existing stuff in codegen maybe it's not ridiculously big thing to do.

thockin: That could be - it seems like the *first* step is to define a place to store the results and a schema.  Will add this idea to the KEP.

thockin: Added a note about staging, too

sftim (commented):
Some thoughts

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,445 @@
...
+  valueType: "String",
+  resources: [ "core/nodes", core/persistentvolumes ],
+  description: "The availability region of the resource.",
+  docsLink: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
+}]
+```
+
+<<[UNRESOLVED key collisions ]>>
+Should we have a YAML per key - easier lookups and collision avoidance, but
+lots of files.  Would maybe need Context to be a list.
```
sftim, line 383: Another potential home for one-file-per-key is k/website. Code that wants to could pull from the website repo and turn [front matter](https://gohugo.io/content-management/front-matter/) YAML into something that other tools want to consume.

keps/sig-architecture/3784-database-of-names/README.md
aojea: yeah

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,445 @@
...
+OWNERS files to delegate ownership.
+
+Given that primary index, we can programmatically produce secondary indices
+which allow easy searching by other criteria (e.g. by SIG or by API resource).
+
+We can also use this data to generate docs, if we want to make the
+"description" parts of this proposal fairly robust.
+
+This repo should probably live under k/k/staging to make PRs easier to add
+names and code which uses them at the same time.
```
aojea, line 169: nit, this paragraph may be reworded to avoid some confusion if you don't know staging syncs to the external repo from k/k

> The "database" should live in a new git repo (proposed name:
"kubernetes/namesdb").

> This repo should probably live under k/k/staging

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+We have too many KEPs and PRs that add annotation- or label-keys without any
+sort of collision detection or governance or documentation.
+
+### Goals
+
+1) Publish a list of known values.
+2) Have a place where end-users can go to find information.
+3) Have a place we can send developers to get some modicum of review of these
+names.
```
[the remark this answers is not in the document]
thockin: I get that this is a feature in markdown, but I don't really get why.  It just makes the source hard to read.  Will apply.

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+        deployment/
+            names.yaml
+        endpoints/
+            names.yaml
+        service/
+            names.yaml
+```
+
+These secondary indices can be created by a simple program which we validate
+has been run before merging changes to the primary index.
```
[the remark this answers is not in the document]
aojea: we kind of do something similar today with all the generated and api docs, is it worth the complexity?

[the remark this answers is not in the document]
thockin: It's possible, but

a) I don't know if we do that anywhere else
b) We already demand users run myriad updaters, and we have CI patterns to enforce it

Will note.

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+  description: "The OS of the node.",
+}]
+```
+
+
+/by-name/io/kubernetes/topology/names.yaml:
+```yaml
+[{
+  key: "kubernetes.io/hostname",
+  key: "topology.kubernetes.io/zone",
```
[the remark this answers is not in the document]
thockin: I had an edit error, and I may be sloppy on some of the YAML - I did not validate it all

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+  sig: "Architecture",
+  context: "Label",
+  valueType: "String",
+  resources: [ "core/nodes", core/persistentvolumes ],
+  description: "The availability zone of the resource.",
+  docsLink: https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone
```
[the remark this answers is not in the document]
thockin: If we think all docs go to the same page, sure.  This is a little self-referential right now - do we generate this docs page from metadata, or do we generate this metadata from the docs, or do we demand contributors update 2 things (we know how well that works)?

I really want to hear your thougts on this specifically.  The docs that exist are much more verbose than the examples I am providing here and include cross-refs and other things which might be awkward in the database text.  But having 2 sources of truth is always suspect.

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,441 @@
...
+}, {
+  key: "topology.kubernetes.io/region",
+  sig: "Architecture",
+  context: "Label",
+  valueType: "String",
```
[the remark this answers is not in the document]
thockin: The transport type is string, but the actual value can be something else.  E.g. annotation values are sometimes JSON-encoded structs.  I meant this as a place to describe the "inner" type, but I see it's not clear.

Will update some text

keps/sig-architecture/3784-database-of-names/README.md
thockin: I'd like ONE source of truth and the rest derived from that, if possible.  Secondary concern is ease of adding/updating (which is why I suggest k/k/staging).  But I am open...

keps/sig-architecture/3784-database-of-names/README.md
thockin: ACK

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,449 @@
...
+// Context describes where a name might be used.
+type Context string
+
+const (
+    ContextLabel = Context("Label")             // key+value
+    ContextAnnotation = Context("Annotation")   // key+value
+    ContextTaint = Context("Taint")             // key
+    ContextCondition = Context("Condition")     // key
+    ContextAppProtocol = Context("AppProtocol") // key
+    ContextAppProtocol = Context("ApiGroup")    // key
```
danwinship, line 305:
```suggestion
    ContextApiGroup = Context("ApiGroup")    // key
```

(except really `API` not `Api` right?)

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+<<[UNRESOLVED enumerate resources ]>>
+Should we enumerate resources?  Good for validation, but a bit tedious.
+<<[/UNRESOLVED]>>
+
+#### Why YAML
+
+YAML is the lingua franca of Kubernetes.  We propose to use a subset of YAML
+which is closer to JSON and easier to avoid whitespace errors, while still
+allowing comments.
```
danwinship, line 333: Why comments? Shouldn't every piece of relevant information be stored in a field?

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+1. Completeness.  We know that getting 100% of the worms back into the can may
+   never happen.  That doesn't mean we should not try.
+1. Enforcement.  We do not intend to inject this into code-review processes.
+   We do intend to make people **aware** of it, so they can voluntarily contribute.
+
+## Proposal
+
+The "database" should live in a new git repo (proposed name:
+"kubernetes/namesdb").  In this repo we will establish a files-and-directories
```
danwinship, line 156: We have reserved things besides names too. eg https://github.com/kubernetes/kubernetes/blob/master/pkg/cluster/ports/ports.go. Would it be useful to step back one level and just have a generic "kubernetes reserved things" repository ("Kubernetes Assigned Numbers Authority"?), with names being one subdir and others potentially added later?

keps/sig-architecture/3784-database-of-names/README.md
thockin: ACK

keps/sig-architecture/3784-database-of-names/README.md
thockin: Good insight, but doesn't change my preference that comments be POSSIBLE :)

keps/sig-architecture/3784-database-of-names/README.md
thockin: I don't know that ports need all the same tooling and metadata, but I agree we don't do a great job of documenting them.  Would it be enough to just convert that .go file into markdown for k/website?

sftim: > Would it be enough to just convert that .go file into markdown for k/website?

We (SIG Docs) can also work with JSON or YAML. JSON is handy because it has a defined media type and we can do HTTP fetches.

keps/sig-architecture/3784-database-of-names/README.md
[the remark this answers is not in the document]
sftim: Watch out for whether you're defining the Kubernetes API (ie, as seen for conformance), or a particular implementation of that API.

If we change the internal implementation, could this drift?

keps/sig-architecture/3784-database-of-names/README.md
sftim: Website is the current source of truth - but not machine readable.

keps/sig-architecture/3784-database-of-names/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+N/A
+
+## Implementation History
+
+Jan 26, 2023: First draft.
+
+## Drawbacks
+
+It's YET ANOTHER thing to update.
```
sftim, line 440: Does this make it harder to localize the label / annotation definitions?

(I think: probably it does, but it's still do-able)

keps/sig-architecture/3784-database-of-names/README.md
danwinship: I was not concerned _specifically_ with ports being documented/parseable, I was just wondering if we're going to need a bigger registry later, and thinking it would be nicer to have it all in one repo rather than having both a "namesdb" and a "portsdb" and who knows what else (Kubernetes-specific IPFIX elements?)

(Though the fact that ports was the first thing that came to mind is because [OpenShift _does_ have a centralized port registry](https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md) because random operators kept trying to claim the same metrics ports. But this is a bigger management problem in OpenShift because it's spread across so many repos.)

keps/sig-architecture/3784-database-of-names/README.md
thockin: If we put enough human-friendly text into this DB, would you want to generate the doc from it?

keps/sig-architecture/3784-database-of-names/README.md
thockin: Great insight - yes.  Perhaps we need to split the machine-data from the human-data?  E.g. one of the fields in the record is a pointer the human-friendly strings which are localized and can feed docs?  Ideally it would be one PR to update those - does that mean all the localized strings live in the same repo?  Or do we put EN in this DB and non-EN translations are a secondary DB which is used by docs?

keps/sig-architecture/3784-database-of-names/README.md
thockin: ACK - I don't object to leaving room at the root for an alternate hierarchy of of data (e.g.  .../names/... vs .../ports/...) with different schema.

thockin: That changes "namesdb" to ... "iddb" or "nndb" (names and numbers) or "metadb" or something?

keps/sig-architecture/3784-database-of-names/README.md
sftim: For the website, a common pattern is to generate [the relevant part of] the English docs using automation, and then have each localized version be generated and maintained by hand.

So the localized text would live only in k/website. I think that would work for this.

PR #3800, opened 2023-01-30
1847: 1.27 beta update

mattcary:
- Update KEP #1847 status for beta graduation

- Issue link: #1847 

- Other comments: No changes are necessary for beta. This KEP was almost ready for beta in 1.26, but needs more e2e test soak time as the statefulset e2e tests were not being run in alpha clusters since cloud provider extraction / CSI migration (see also https://github.com/kubernetes/kubernetes/issues/114955).

/cc @msau42 
/cc @johnbelamaric


soltysh (approved):
/lgtm
/approve

PR #3877, opened 2023-02-14
KEP-1487: Mark AWS EBS CSI migration as done

jsafrane:
- One-line PR description: Mark the feature as `implemented`

- Issue link: https://github.com/kubernetes/enhancements/issues/1487

- Other comments:


sftim:
I think we should hold this until that change actually merges.

PR #3885, opened 2023-02-22
Update SIG network to include Shane Utt as a chair

caseydavenport:


kikisdeliveryservice:
Thanks for updating, will let the SIG-networkers do the final lgtm. 😄 

/approve
/assign @dcbw @thockin
cc: @shaneutt

thockin (commented):
Thanks!

/lgtm
/approve

PR #3894, opened 2023-02-28
KEP-2831: add reviewers

sallyom:
KEP-2831: adding reviewers

Adding @saschagrunert and @vrutkovs to kubelet tracing KEP. They are leading the effort for beta graduation.


saschagrunert (approved):
Thanks!

PR #3964, opened 2023-04-25
update kep-2799

zshihang:
- One-line PR description: update outdated text.
- Issue link: https://github.com/kubernetes/enhancements/issues/2799


keps/sig-auth/2799-reduction-of-secret-based-service-account-token/README.md
```diff
@@ -227,10 +227,10 @@ legacy tokens for security practices.
 
 #### Beta -> GA Graduation
 
-- [ ] In use by multiple distributions
-- [ ] Approved by PRR and scalability
-- [ ] Any known bugs fixed
-- [ ] Tests passing
+- [x] In use by multiple distributions
```
liggitt, line 230: GKE is deploying 1.27 with this enabled, check with a couple other distributions to make sure we're actually satisfying this requirement

keps/sig-auth/2799-reduction-of-secret-based-service-account-token/README.md
```diff
@@ -227,10 +227,10 @@ legacy tokens for security practices.
...
+- [x] Approved by PRR and scalability
```
liggitt, line 231: get an ack from PRR reviewer on the tracking portion of the KEP (no changes since beta, so I don't expect any issues)

liggitt:
/lgtm

get an lgtm from PRR and add details about the providers using this in 1.27

keps/sig-auth/2799-reduction-of-secret-based-service-account-token/README.md
zshihang: /cc @ritazh for Microsoft and @deads2k for RedHat

by default this feature is enabled, do you know if this feature is ever disabled?

deads2k:
PRR lgtm modulo the naming nit.  I think it's worth updating.

/approve
/hold

holding for the configmap name update so people can find this on their clusters.

PR #4081, opened 2023-06-12
add missing dependency question to KMSv2 KEP

aramase:
- Add missing dependency question for KMSv2 KEP

### DIFF
--- a/keps/sig-auth/3299-kms-v2-improvements/README.md
+++ b/keps/sig-auth/3299-kms-v2-improvements/README.md
@@ -562,6 +562,10 @@ No.
 
 No.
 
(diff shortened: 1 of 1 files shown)


keps/sig-auth/3299-kms-v2-improvements/README.md
```diff
@@ -562,6 +562,10 @@ No.
 
 No.
 
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No.
```
enj, line 567: I would mention that one socket is used per KMS plugin, and that old connections are closed after new connections have been created/validated for health during an automatic reload flow.

PR #4084, opened 2023-06-13
KEP-127: add support for stateful pods

giuseppe:
now that we use idmapped mounts, there is no real difference between stateful and stateless pods, and they can be handled in the same way.

Refers to https://github.com/kubernetes/enhancements/issues/127


saschagrunert (approved):

mrunalp (approved):
/lgtm

PR #4165, opened 2023-08-29
Addressed gRPC metadata and id format

carlbraganza:
- Send the auth token in the metadata
- Removed namespace and security token from the RPC calls


carlbraganza:
Opened in error

PR #4296, opened 2023-10-16
KEP-1027: fix markdown

mandre:
- One-line PR description: Add missing backtick that caused parts of the text to disappear when rendered. Check https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md#go-markers to see how it currently renders.

- Issue link:

- Other comments:


apelisse:
Thanks. I think we should also, beyond the messed-up marker, mention that the feature is actually not a thing at all and hasn't ever been implemented properly. I still see some occurences of people trying to use these markers when they indeed do nothing.
@liggitt

PR #4344, opened 2023-11-22
Update OWNERS_ALIASES from community repo

cblecker:


SergeyKanzhelev:
This PR broke the SIG Node approvership. Why those alias files must match? TLs has different permissions than all leads in SIG Node

SergeyKanzhelev:
created https://github.com/kubernetes/enhancements/pull/4458 to fix up sig node permissions

PR #4437, opened 2024-01-24
Update mdtoc, fix TOC entries with backticks

danwinship:
The version of mdtoc currently used by `update-toc.sh` / `verify-toc.sh` does not handle headings with backticks in them, resulting in broken links in some KEPs. eg, in [the nftables kube-proxy KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3866-nftables-proxy/README.md), the table of contents entry for the ``##### Forcing traffic on `HealthCheckNodePort`s to be accepted`` heading links to `#forcing-traffic-on-s-to-be-accepted` but the actual anchor generated by GitHub is `#forcing-traffic-on-healthchecknodeports-to-be-accepted`.

This has actually been fixed in https://github.com/kubernetes-sigs/mdtoc for a very long time, we're just still using a very old version of mdtoc...

cc @saschagrunert


mrbobbytables (commented):
Thanks!
/lgtm
/approve

PR #4439, opened 2024-01-25
KEP-127: graduate to Beta for 1.30

giuseppe:
- Issue link: https://github.com/kubernetes/enhancements/issues/127

### DIFF
--- a/keps/prod-readiness/sig-node/127.yaml
+++ b/keps/prod-readiness/sig-node/127.yaml
@@ -4,3 +4,5 @@
 kep-number: 127
 alpha:
   approver: "wojtek-t"
+beta:
+  approver: "wojtek-t"
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -288,6 +288,43 @@ message Mount {
 }
 ```
 
+The CRI runtime reports what runtime handlers have support for user
+namespaces through the `StatusResponse` message, that gains a new
+field `runtime_handlers`:
+
+```
+message StatusResponse {
+    // Status of the Runtime.
+    RuntimeStatus status = 1;
+    // Info is extra information of the Runtime. The key could be arbitrary string, and
+    // value should be in json format. The information could include anything useful for
+    // debug, e.g. plugins used by the container runtime.
+    // It should only be returned non-empty when Verbose is true.
+    map<string, string> info = 2;
+
+    // Runtime handlers.
+    repeated RuntimeHandler runtime_handlers = 3;
+}
+```
+
+Where RuntimeHandler is defined as below:
+
+```
+message RuntimeHandlerFeatures {
+    // supports_user_namespaces is set to true if the runtime handler supports
+    // user namespaces.
+    bool supports_user_namespaces = 1;
+}
+
+message RuntimeHandler {
+    // Name must be unique in StatusResponse.
+    // An empty string denotes the default handler.
+    string name = 1;
+    // Supported features.
+    RuntimeHandlerFeatures features = 2;
+}
+```
+
 ### Support for pods
 
 Make pods work with user namespaces. This is activated via the
@@ -593,6 +630,7 @@ use container runtime versions that have the needed changes.
 
 - Gather and address feedback from the community
 - Be able to configure UID/GID ranges to use for pods
+- This feature is not supported on Windows.
 - Get review from VM container runtimes maintainers (not blocker, as VM runtimes should just ignore
   the field, but nice to have)
 
@@ -602,6 +640,20 @@ use container runtime versions that have the needed changes.
 - Fully integrate into [Pod Security Standards (PSS)](#pod-security-standards-pss-integration)
 
 ### Upgrade / Downgrade Strategy
+
+Existing pods will still work as intended, as the new field is missing there.
+
+Upgrade will not change any current behaviors.
+
+When the new functionality wasn't yet used, downgrade will not be affected.
+
+On downgrade, when the functionality was used, the pods created with
+user namespaces that are running will continue to run with user
+namespaces. Pods will need to be re-created to stop using the user
+namespace.
+
+Versions of Kubernetes that doesn't have this feature implemented will
+ignore the new field `pod.spec.hostUsers`.
 
 ### Version Skew Strategy
 
@@ -635,11 +687,13 @@ doesn't create them. The runtime can detect this situation as the `user` field
 in the `NamespaceOption` will be seen as nil, [thanks to
 protobuf][proto3-defaults]. We already tested this with real code.
 
-Old runtime and new kubelet: containers are created without userns. As the
-`user` field of the `NamespaceOption` message is not part of the runtime
-protofiles, that part is ignored by the runtime and pods are created using the
-host userns.
+Old runtime and new kubelet: the runtime won't report that it supports
+user namespaces through the `StatusResponse` message, so the kubelet
+will detect it and return an error if a pod with user namespaces is
+created.
 
+We added unit tests for the feature gate disabled, and integration
+tests for the feature gate enabled and disabled.
 
 [proto3-defaults]: https://developers.google.com/protocol-buffers/docs/proto3#default
 
@@ -763,15 +817,14 @@ This section must be completed when targeting beta to a release.
 
 The rollout is just a feature flag on the kubelet and the kube-apiserver.
 
-If one API server is upgraded while others aren't, the pod will be accepted (if the apiserver is >=
-1.25). If it is scheduled to a node that the kubelet has the feature flag activated and the node
-meets the requirements to use user namespaces, then the pod will be created with the namespace. If
-it is scheduled to a node that has the feature disabled, it will be created without the user
-namespace.
+If one APIserver is upgraded while other's aren't and you are talking to a not upgraded the pod
+will be accepted (if the apiserver is >= 1.25). If it is scheduled to a node that the kubelet has
+the feature flag activated and the node meets the requirements to use user namespaces, then the
+pod will be created with the namespace. If it is scheduled to a node that has the feature disabled,
+it will be created without the user namespace.
 
 On a rollback, pods created while the feature was active (created with user namespaces) will have to
-be restarted to be re-created without user namespaces. Just a re-creation of the pod will do the
-trick.
+be re-created without user namespaces.
 
 <!--
 Try to be as paranoid as possible - e.g., what if some components will restart
@@ -788,7 +841,9 @@ will rollout across nodes.
 On Kubernetes side, the kubelet should start correctly.
 
 On the node runtime side, a pod created with pod.spec.hostUsers=false should be on RUNNING state if
-all node requirements are met.
+all node requirements are met. If the CRI runtime or the handler do not support the feature, the kubelet
+returns an error.
+
 <!--
 What signals should users be paying attention to when the feature is young
 that might indicate a serious problem?
@@ -796,16 +851,21 @@ that might indicate a serious problem?
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-Yes.
+Yes, we tested it locally using `./hack/local-up-cluster.sh`.
 
-We tested to enable the feature flag, create a deployment with pod.spec.hostUsers=false, and then disable
-the feature flag and restart the kubelet and kube-apiserver.
+We tested enabling the feature flag, created a deployment with pod.spec.hostUsers=false, and then disabled
+the feature flag and restarted the kubelet and kube-apiserver.
 
 After that, we deleted the deployment pods (not the deployment object), the pods were re-created
 without user namespaces just fine, without any modification needed on the deployment yaml.
 
 We then enabled the feature flag on the kubelet and kube-apiserver, and deleted the deployment pod.
 This re-created caused the pod to be re-created, this time with user namespaces enabled again.
+
+To validate it, it is necessary to exec into a container in the pod and run the command `cat /proc/self/uid_map`.
+When running in a user namespace the output is different than `0 0 4294967295` as it happens when running without
+a user namespace.
+
 <!--
 Describe manual testing that was done and the outcomes.
 Longer term, we may want to require automated upgrade/rollback tests, but we
@@ -839,10 +899,9 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
-Check if any pod has the pod.spec.hostUsers field set to false and is on RUNNING state on a node
-that meets all the requirements.
+If the runtime doesn't support user namespaces an error is returned by the kubelet.
 
-There are step-by-step examples in the Kubernetes documentation too.
+There are step-by-step examples in the Kubernetes documentation too: https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/
 
 <!--
 For instance, if this is a pod-related feature, it should be possible to determine if the feature is functioning properly
@@ -860,7 +919,8 @@ Recall that end users cannot usually observe component logs or access metrics.
   - Other field:
 - [x] Other (treat as last resort)
   - Details: check pods with pod.spec.hostUsers field set to false, and see if they are in RUNNING
-    state.
+    state. Exec into a container and run `cat /proc/self/uid_map` to verify that the mappings are different
+    than the mappings on the host.
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
@@ -1373,9 +1433,9 @@ instead of cluster-wide/cluster-scoped.
 
 Some use cases for longer mappings include:
 
-- running a container tool inside a Pod, where that container tool wants to use a UID range.
-- running an application inside a Pod where the application uses UIDs
-  above 65535 by default.
+There are no known use cases for longer mappings that we know of. The 16bit range (0-65535) is what
+is assumed by all POSIX tools that we are aware of. If the need arises, longer mapping can be
+considered in a future KEP.
 
 ### Allow runtimes to pick the mapping
 
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -591,14 +591,10 @@ use container runtime versions that have the needed changes.
 
 ##### Beta
 
-- Make plans on whether, when, and how to enable by default
-
-###### Open Questions
-
-- Should we reconsider making the mappings smaller by default?
-- Should we allow any way for users to for "more" IDs mapped? If yes, how many more and how?
-- Should we allow the user to ask for specific mappings?
-- Get review from VM container runtimes maintainers
+- Gather and address feedback from the community
+- Be able to configure UID/GID ranges to use for pods
+- Get review from VM container runtimes maintainers (not blocker, as VM runtimes should just ignore
+  the field, but nice to have)
 
 ##### GA
 
@@ -1361,15 +1357,23 @@ KEPs can explore this path if they so want to.
 
 ### 64k mappings?
 
-We will start with mappings of 64K. Tim Hockin, however, has expressed
-concerns. See more info on [this Github discussion](https://github.com/kubernetes/enhancements/pull/3065#discussion_r781676224)
-SergeyKanzhelev [suggested a nice alternative](https://github.com/kubernetes/enhancements/pull/3065#discussion_r807408134),
-to limit the number of pods so we guarantee enough spare host UIDs in case we
-need them for the future. There is no final decision yet on how to handle this.
-For now we will limit the number of pods, so the wide mapping is not
-problematic, but [there are downsides to this too](https://github.com/kubernetes/enhancements/pull/3065#discussion_r812806223)
+We discussed using shorter or even allowing for longer mappings in the past. The decision is to use
+64k mappings (IDs from 0-65535 are mapped/valid in the pod).
 
-For stateless pods this is of course not an issue.
+The reasons to consider smaller mappings were valid only before idmap mounts was merged into the
+kernel. However, idmap mounts is merged for some years now and we require it, making those reasons
+void.
+
+The issues without idmap mounts in previous iterations of this KEP, is that the IDs assigned to a
+pod had to be unique for every pod in the cluster, easily reaching a limit when the cluster is "big
+enough" and the UID space runs out.  However, with idmap mounts the IDs assigned to a pod just needs
+to be unique within the node (and with 64k ranges we have 64k pods possible in the node, so not
+really an issue). IOW, by using idmap mounts, we changed the IDs limit to be node-scoped instead of
+cluster-wide/cluster-scoped.
+
+There are no known use cases for longer mappings that we know of. The 16bit range (0-65535) is what
+is assumed by all POSIX tools that we are aware of. If the need arises, longer mapping can be
+considered in a future KEP.
 
 ### Allow runtimes to pick the mapping?
 
@@ -1378,7 +1382,11 @@ mapping and have different runtimes pick different mappings. While KEP authors
 disagree on this, we still need to discuss it and settle on something.  This was
 [raised here](https://github.com/kubernetes/enhancements/pull/3065#discussion_r798760382)
 
-This is not a blocker for the KEP, but it is something that can be changed later on.
+Furthermore, the reasons mentioned by Tim (some nodes having CRIO, some others having containerd,
+etc.) are handled correctly now. Different nodes can use different container runtimes, if a custom
+range needs to be used by the kubelet, that can be configured per-node.
+
+Therefore, this old concerned is now resolved.
 
 <!--
 What other approaches did you consider, and why did you rule them out? These do
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -19,6 +19,7 @@
   - [Pod.spec changes](#podspec-changes)
   - [CRI changes](#cri-changes)
   - [Support for pods](#support-for-pods)
+    - [Configuration of ranges](#configuration-of-ranges)
   - [Handling of volumes](#handling-of-volumes)
     - [Example of how idmap mounts work](#example-of-how-idmap-mounts-work)
       - [Example without idmap mounts](#example-without-idmap-mounts)
@@ -315,10 +316,21 @@ The picked range will be stored under a file named `userns` in the pod folder
 (by default it is usually located in `/var/lib/kubelet/pods/$POD/userns`). This
 way, the Kubelet can read all the allocated mappings if it restarts.
 
-During alpha, to make sure we don't exhaust the host UID namespace, we will
-limit the number of pods using user namespaces to `min(maxPods, 1024)`. This
-leaves us plenty of host UID space free and this limits is probably never hit in
-practice. See the [Unresolved section](#unresolved) for more details on this.
+#### Configuration of ranges
+
+If no configuration is present, the kubelet will use a sane default:
+
+ * `0-65535`: reserved for the host processes and files
+ * `65536-<65536 * maxPods>`: reserved for pods
+
+The kubelet will detect if a configuration is present by using the `getsubids` binary. It will query
+for ranges allocated to the "kubelet" user.
+
+If the kubelet user doesn't exist or `getsubids` is not installed, the kubelet will use the sane
+default aforementioned. On any other cases, it will return an error.
+
+By using `getsubids` we make sure the kubelet interacts fine with other programs in the host that
+also allocate ranges for user namespaces (be it /etc/subuid or centralized systems as freeIPA).
 
 ### Handling of volumes
 
@@ -1017,9 +1029,8 @@ and validate the declared limits?
 
 The kubelet is spliting the host UID/GID space for different pods, to use for
 their user namespace mapping. The design allows for 65k pods per node, and the
-resource is limited in the alpha phase to the min between maxPods per node
-kubelet setting and 1024. This guarantees we are not inadvertly exhausting the
-resource.
+resource is limited to maxPods per node (currently maxPods defaults to 110, it is unlikely we will
+reach the host limit).
 
 For container runtimes, they might use more disk space or inodes to chown the
 rootfs. This is if they chose to support this feature without relying on new
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -690,7 +690,7 @@ well as the [existing list] of feature gates.
 -->
 
 - [x] Feature gate (also fill in values in `kep.yaml`)
-  - Feature gate name: UserNamespacesStatelessPodsSupport
+  - Feature gate name: UserNamespacesPodsSupport
   - Components depending on the feature gate: kubelet, kube-apiserver
 
 ###### Does enabling the feature change any default behavior?
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -39,6 +39,8 @@
       - [GA](#ga)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
+    - [Kubelet and Kube-apiserver skew](#kubelet-and-kube-apiserver-skew)
+    - [Kubelet and container runtime skews](#kubelet-and-container-runtime-skews)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
   - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
   - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
@@ -612,7 +614,7 @@ use container runtime versions that have the needed changes.
 
 ##### critests
 
-- For Alpha, the feature is tested for containerd and CRI-O in cri-tools repo using critest to
+- For Beta, the feature is tested for containerd and CRI-O in cri-tools repo using critest to
   make sure the specified user namespace configuration is honored.
 
 - <test>: <link to test coverage>
@@ -630,6 +632,9 @@ use container runtime versions that have the needed changes.
 
 - Gather and address feedback from the community
 - Be able to configure UID/GID ranges to use for pods
+- Add unit tests that exercise the feature gate switch (see section "Are there
+  any tests for feature enablement/disablement?")
+- Add cri-tools test
 - This feature is not supported on Windows.
 - Get review from VM container runtimes maintainers (not blocker, as VM runtimes should just ignore
   the field, but nice to have)
@@ -669,6 +674,26 @@ enhancement:
 - Will any other components on the node change? For example, changes to CSI,
   CRI or CNI may require updating that component before the kubelet.
 -->
+
+#### Kubelet and Kube-apiserver skew
+
+The apiserver and kubelet feature gate enablement work fine in any combination:
+
+1. If the apiserver has the feature gate enabled and the kubelet doesn't, then the pod will show
+   that field and the kubelet will ignore it. Then, the pod is created without user namespaces.
+2. If the apiserver has the feature gate disabled and the kubelet enabled, the pod won't show this
+   field and therefore the kubelet won't act on a field that isn't shown. The pod is created without
+   user namespaces.
+
+The kubelet can still create pods with user namespaces if static-pods are configured with
+pod.spec.hostUsers and has the feature gate enabled.
+
+If the kube-apiserver doesn't support the feature at all (< 1.25), a pod with userns will be
+rejected.
+
+If the kubelet doesn't support the feature (< 1.25), it will ignore the pod.spec.hostUsers field.
+
+#### Kubelet and container runtime skews
 
 Some definitions first:
 - New kubelet: kubelet with CRI proto files that includes the changes proposed in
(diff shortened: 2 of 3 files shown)


keps/sig-node/127-user-namespaces/kep.yaml
```diff
@@ -16,7 +16,7 @@ approvers:
   - "@derekwaynecarr"
 
 stage: alpha
-latest-milestone: "v1.29"
+latest-milestone: "v1.30"
 milestone:
   alpha: "v1.25"
```
sftim, line 21: Did you mean:
```suggestion
stage: beta
latest-milestone: "v1.30"
milestone:
  alpha: "v1.25"
  beta: "v1.30"
```

giuseppe: thanks, fixed now.

keps/prod-readiness/sig-node/127.yaml
```diff
@@ -4,3 +4,5 @@
 kep-number: 127
 alpha:
   approver: "wojtek-t"
+beta:
+  approver: "wojtek-t"
```
rata, line 8: @wojtek-t can you please take a look? :)

thockin:
FTR - Prs with no assignee often fall thru cracks. :(

rata:
/assign @mrunalp @wojtek-t 

@thockin thanks, let's see if this command works, then :)

keps/prod-readiness/sig-node/127.yaml
wojtek-t: Will take a look later this week.
In the meantime, can you please work with SIG node on SIG approval?

rata: @wojtek-t Thanks! Yes, we are working with them since a few weeks already. I'll ping again :)

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -579,14 +591,10 @@ use container runtime versions that have the needed changes.
...
-###### Open Questions
-
-- Should we reconsider making the mappings smaller by default?
-- Should we allow any way for users to for "more" IDs mapped? If yes, how many more and how?
-- Should we allow the user to ask for specific mappings?
-- Get review from VM container runtimes maintainers
+- Gather and address feedback from the community
+- Be able to configure UID/GID ranges to use for pods
+- Get review from VM container runtimes maintainers (not blocker, as VM runtimes should just ignore
+  the field, but nice to have)
```
haircommander, line 597: @fidencio @littlejawa have you or anyone from kata community taken a look?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1046,8 +1118,144 @@ details). For now, we leave it here.
...
+        There are extensive unit tests on the function that parses the file content.
+        There are no tests for failures to read or write the file, the code-paths just return the errors
+        in those cases.
+
+
+- Error getting the kubelet IDs range configuration
+  - Detection: How can it be detected via metrics? Stated another way:
+    how can an operator troubleshoot without logging into a master or worker node?
+
+        In this case the Kubelet will fail to start with a clear error message.
```
mrunalp, line 1241: Assuming this is through subuids binary don't we fallback to default range?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +761,18 @@ This section must be completed when targeting beta to a release.
 
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
+The rollout is just a feature flag on the kubelet and the kube-apiserver.
+
+If one API server is upgraded while others aren't, the pod will be accepted (if the apiserver is >=
+1.25). If it is scheduled to a node that the kubelet has the feature flag activated and the node
+meets the requirements to use user namespaces, then the pod will be created with the namespace. If
+it is scheduled to a node that has the feature disabled, it will be scheduled without the user
```
mrunalp, line 769: scheduled or created?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -765,13 +785,27 @@ will rollout across nodes.
 
 ###### What specific metrics should inform a rollback?
 
+On Kubernetes side, the kubelet should start correctly.
+
+On the node runtime side, a pod created with pod.spec.HostUsers=false should be on RUNNING state if
+all node requirements are met.
```
wojtek-t, line 791: I'm not a fan of this answer - if I'm managing 1000s of clusters, it's not something I can reasonably monitor. Can we rely on easier to monitor conditions - ideally some metrics?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -803,6 +839,11 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
+Check if any pod has the pod.spec.HostUsers field set to false and is on RUNNING state on a node
+that meets all the requirements.
+
+There are step-by-step examples in the Kubernetes documentation too.
```
wojtek-t, line 845: Link?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1046,8 +1118,144 @@ details). For now, we leave it here.
...
+  - Mitigations: What can be done to stop the bleeding, especially for already
+    running user workloads?
+
+        Remove the pod.spec.HostUsers field or disable the feature gate.
+
+  - Diagnostics: What are the useful log messages and their required logging
+    levels that could help debug the issue?
+    Not required until feature graduated to beta.
+
+        The error is returned on pod-creation, no need to search for logs.
```
wojtek-t, line 1144: Rather on an attempt to start the pod?

IIUC, Kubelet will get an error from the runtime and will propagate it to the pod, right?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1166,7 +1383,11 @@ mapping and have different runtimes pick different mappings. While KEP authors
 disagree on this, we still need to discuss it and settle on something.  This was
 [raised here](https://github.com/kubernetes/enhancements/pull/3065#discussion_r798760382)
 
-This is not a blocker for the KEP, but it is something that can be changed later on.
+Furthermore, the reasons mentioned by Tim (some nodes having CRIO, some others having containerd,
```
sftim, line 1386: Let's specify which Tim. The Kubernetes project has multiple Tims :wink:

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -795,6 +830,7 @@ previous answers based on experience in the field.
 
 ###### How can an operator determine if the feature is in use by workloads?
 
+Check if any pod has the pod.spec.HostUsers field set to false.
```
sftim, line 833: (nit)
```suggestion
Check if any pod has the `.spec.hostUsers` field set to false.
```

In the API, the field is `.spec.hostUsers`.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -803,6 +839,11 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
+Check if any pod has the pod.spec.HostUsers field set to false and is on RUNNING state on a node
```
sftim, line 842:
```suggestion
Check if any pod has the `.spec.ghstUsers` field set to false and is on RUNNING state on a node
```

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -813,15 +854,22 @@ Recall that end users cannot usually observe component logs or access metrics.
...
+  - Event Reason:
 - [ ] API .status
-  - Condition name: 
-  - Other field: 
-- [ ] Other (treat as last resort)
-  - Details:
+  - Condition name:
+  - Other field:
+- [x] Other (treat as last resort)
+  - Details: check pods with pod.spec.HostUsers field set to false, and see if they are in RUNNING
```
sftim, line 862:
```suggestion
  - Details: check pods with `.spec.hostUsers` field set to false, and see if they are in RUNNING
```

keps/sig-node/127-user-namespaces/README.md
sftim, line 1241: (nit)
```suggestion
        In this case the kubelet will fail to start with a clear error message.
```

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1149,15 +1358,23 @@ KEPs can explore this path if they so want to.
...
+The reasons to consider smaller mappings were valid only before idmap mounts was merged into the
+kernel. However, idmap mounts is merged for some years now and we require it, making those reasons
+void.
 
-For stateless pods this is of course not an issue.
+The issues without idmap mounts in previous iterations of this KEP, is that the IDs assigned to a
+pod had to be unique for every pod in the cluster, easily reaching a limit when the cluster is "big
+enough" and the UID space runs out.  However, with idmap mounts the IDs assigned to a pod just needs
+to be unique within the node (and with 64k ranges we have 64k pods possible in the node, so not
+really an issue). IOW, by using idmap mounts, we changed the IDs limit to be node-scoped instead of
```
sftim, line 1372: (nit)
```suggestion
really an issue). In other words: by using idmap mounts, we changed the IDs limit to be node-scoped instead of
```

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1149,15 +1358,23 @@ KEPs can explore this path if they so want to.
...
+cluster-wide/cluster-scoped.
+
+There are no known use cases for longer mappings that we know of. The 16bit range (0-65535) is what
+is assumed by all POSIX tools that we are aware of. If the need arises, longer mapping can be
+considered in a future KEP.
```
sftim, line 1377: _no_ known use cases?

Two examples come to mind:
- running a container tool inside a Pod, where that container tool wants to use a UID range
- running an application inside a Pod where the application uses UIDs above 65535 by default; I've worked with some of these, usually avoiding the 0-65535 range to avoid the risk of conflict with a local system user.

keps/sig-node/127-user-namespaces/README.md
giuseppe: I think we should fallback to default ranges only if there is no `kubelet` user configured, or there is no `getsubids` binary present.

If the `kubelet` user is defined and we fail to read its configuration, an error should be reported.

keps/sig-node/127-user-namespaces/README.md
giuseppe: created

rata:
@wojtek-t thanks for the detailed review! I'm on my train to FOSDEM now, so I'm not sure I'll be able to answer all now

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +761,18 @@ This section must be completed when targeting beta to a release.
...
+The rollout is just a feature flag on the kubelet and the kube-apiserver.
+
+If one API server is upgraded while others aren't, the pod will be accepted (if the apiserver is >=
+1.25). If it is scheduled to a node that the kubelet has the feature flag activated and the node
+meets the requirements to use user namespaces, then the pod will be created with the namespace. If
+it is scheduled to a node that has the feature disabled, it will be scheduled without the user
+namespace.
+
+On a rollback, pods created while the feature was active (created with user namespaces) will have to
+be restarted to be re-created without user namespaces. Just a re-creation of the pod will do the
```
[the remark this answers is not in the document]
rata: Good point!

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -803,6 +839,11 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
+Check if any pod has the pod.spec.HostUsers field set to false and is on RUNNING state on a node
+that meets all the requirements.
+
```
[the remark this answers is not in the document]
rata: @wojtek-t Why do we need that? I'm honestly not sure I understand, let me explain why:

* We are not doing enforcement of any kind based on the feature gate `UserNamespacesSupport`, so there are no security concerns for requesting this and then not being enabled.
* AFAIK we don't have that metric in the kubelet for any other namespace (nor share the PID namespace, that was added in like 1.17, quite "recently"). **Why is this namespace special?** I honestly don't see it.
* Even if we do want a metric, isn't the container runtime a better place for this metric?
* The CRI runtime doesn't really know if this was honored or not. For runc-compatible runtimes, _maybe_ we can manage to get it (using the features subcommand if idmap is supported, then it must error-out if its not honored). But there is no real feedback from the OCI runtime on what was honored or not, let alone non-OCI runtimes. 

So, I really want to push back a little here and understand why we need this and why we need this in the kubelet metrics.

keps/sig-node/127-user-namespaces/README.md
rata: We can add metrics to signal on the errors listed here in the KEP (in the section "What are other known failure modes?"). 

This can potentially cover a lot at the kubelet layer. But things like one of the the filesystems used by the pod doesn't support idmap mounts on the kernel running will not be possible to detect from the kubelet. The runtime will return a clear error on that case, that is shown to the user, but I don't think we can have a metric for that.

Does doing that sounds like a good plan for you?

I'm not convinced about this, though. All errors are returned to the user, so it's very visible if it fails. And having a proliferation of metrics can also harm, especially long-term. Do you think this is definitely a "yes, this metrics are useful" or do you have mixed feelings about it too?

Also, do you think this is needed for beta graduation or can be added while the feature is in beta and still disabled by default?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -579,14 +591,10 @@ use container runtime versions that have the needed changes.
...
 
-- Make plans on whether, when, and how to enable by default
-
-###### Open Questions
-
-- Should we reconsider making the mappings smaller by default?
-- Should we allow any way for users to for "more" IDs mapped? If yes, how many more and how?
-- Should we allow the user to ask for specific mappings?
-- Get review from VM container runtimes maintainers
+- Gather and address feedback from the community
```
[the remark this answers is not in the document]
rata: Windows support was resolved several months ago. Windows maintainers commented that they don't want this feature in windows and it looked fine anyways from their point of view.

We will try to gather feedback from VM runtimes, yes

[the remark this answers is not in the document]
rata: I'll try to update with the links later

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -803,6 +839,11 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
+Check if any pod has the pod.spec.HostUsers field set to false and is on RUNNING state on a node
+that meets all the requirements.
```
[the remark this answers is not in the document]
rata: I can add the link to the k8s doc if that helps: https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/#before-you-begin

keps/sig-node/127-user-namespaces/README.md
rata: I can add a link to this: https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/

I didn't do it because links here can easily become stale over time.

keps/sig-node/127-user-namespaces/README.md
rata: Yes, exactly

keps/sig-node/127-user-namespaces/README.md
rata: 1. The first example doesn't necessarily need bigger mappings.
1. We didn't hear of any such application that wasn't possible to configure it. The point stands: that is out of scope for this KEP, it can be worked on in a future KEP.

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
thockin: "don't want this feature in windows"

Love it.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -678,7 +686,7 @@ well as the [existing list] of feature gates.
 -->
 
 - [x] Feature gate (also fill in values in `kep.yaml`)
-  - Feature gate name: UserNamespacesStatelessPodsSupport
+  - Feature gate name: UserNamespacesPodsSupport
```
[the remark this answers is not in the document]
thockin: The gate seems to actually be named "UserNamespacesSupport"

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
thockin: > Why is this namespace special? 

It's not special, but past mistakes don't justify future mistakes.

I think we have a real plumbing problem, though.  If you have a CRI which does not understand this new field, it can't report that it doesn't understand it...because it doesn't know what it doesn't know.  If we don't have a way to interrogate the runtime, how can we get this info (except to try it and look at the results)?

IOW, ideally it would be something like:

kubelet: Dear CRI, do you support userns?
runtime: No
kubelet: Dear PodStatus, this pod asked for userns but did not get it

But if there isn't a way to ask the question, it's a lot more complicated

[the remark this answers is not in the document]
mrunalp: We can use RuntimeConfig CRI API for this - https://github.com/kubernetes/kubernetes/commit/0f05a92ce68ec8ae5b65aa4db746300aba5121ca

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
wojtek-t: That SGTM - just please update the KEP to reflect it :)

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
wojtek-t: Thanks @thockin for making this explicit - this is exactly my concern.
I agree that this feature isn't special, but exactly - the fact that we did something bad wrong in the past doesn't mean we can repeat mistakes.

If as a user I'm asking for X (in this case user namespace) and my pod gets to RUNNING state, I really expect it to be running in user namespace. If it's not (and it silently dropped that intent), the actual impact doesn't matter that much - it's just bad because we didn't something else than the user asked.
I would expect the pod to be rejected, and if not, at the very very least see a easy to digest for the operator signal that shows that (e.g. a metric on kubelet and a status on the pod)....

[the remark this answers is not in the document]
giuseppe: would you be ok with something along the lines of the following check?

```diff
diff --git a/pkg/kubelet/kuberuntime/util/util.go b/pkg/kubelet/kuberuntime/util/util.go
index af61e17567e..a9f0a33c5d5 100644
--- a/pkg/kubelet/kuberuntime/util/util.go
+++ b/pkg/kubelet/kuberuntime/util/util.go
@@ -18,6 +18,7 @@ package util
 
 import (
        v1 "k8s.io/api/core/v1"
+       internalapi "k8s.io/cri-api/pkg/apis"
        runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
        "k8s.io/klog/v2"
        kubecontainer "k8s.io/kubernetes/pkg/kubelet/container"
@@ -99,11 +100,20 @@ func PidNamespaceForPod(pod *v1.Pod) runtimeapi.NamespaceMode {
 
 // namespacesForPod returns the runtimeapi.NamespaceOption for a given pod.
 // An empty or nil pod can be used to get the namespace defaults for v1.Pod.
-func NamespacesForPod(pod *v1.Pod, runtimeHelper kubecontainer.RuntimeHelper) (*runtimeapi.NamespaceOption, error) {
+func NamespacesForPod(ctx context.Context, pod *v1.Pod, runtimeHelper kubecontainer.RuntimeHelper, runtimeService internalapi.RuntimeService) (*runtimeapi.NamespaceOption, error) {
        userNs, err := runtimeHelper.GetOrCreateUserNamespaceMappings(pod)
        if err != nil {
                return nil, err
        }
+       if userNs != nil || userNs.Mode != runtimeapi.NamespaceMode_NODE {
+               cfg, err := runtimeService.RuntimeConfig(ctx.Background())
+               if err != nil {
+                       return nil, err
+               }
+               if !runtime.Linux.UserNamespaceSupported {
+                       return nil, fmt.Errorf("The runtime doesn't support user namespaces")
+               }
+       }
```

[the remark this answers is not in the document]
rata: Note: this modulo the feature gate checking :)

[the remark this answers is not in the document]
giuseppe: instead of using a single bool for the RuntimeConfig, we can use the same RuntimeHandlerFeatures mechanism described for KEP-3857 (Recursive Read-Only mounts): https://github.com/kubernetes/enhancements/pull/3858/files#diff-057b8627f24bc6a0742b51b4fea113938a3440eafe255edecc834f3301c008fcR478 so we can specify the UserNamespace feature support for each separate handler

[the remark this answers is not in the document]
mrunalp: Yes, that approach will work for both the cases and also handle runtime class specific features.

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
haircommander: @wojtek-t for 1 and 2: https://github.com/kubernetes/enhancements/pull/4439/files#diff-a9ca9fbce1538447b92f03125ca2b8474e2d875071f1172d2afa0b1e8cadeabaL613-L633 What are you looking for from this section that isn't covered already?

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
wojtek-t: > Just to be clear: this will only create the pod if the CRI runtime knows about userns, and return the error otherwise.

Yes - that's exactly what I would like to have.
Can you please update the PRR (and probably mention it in the core of the KEP too) to reflect that?

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
wojtek-t: 1. Update/Downgrade is not filled in at all
2. For version skew I explicit wrote  - please describe the case what happens if feature is enabled in kube-apiserver and not in kubelet and vice-versa.

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
rata: Sure! Now that we agreed on a direction, we will update the KEP :)

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
rata: 1. Oh, that is one of the few section that doesn't have any comments, it is such a small text that I didn't see it! We will add it now, sorry!

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
rata: We updated the kep @wojtek-t , PTAL!

keps/sig-node/127-user-namespaces/README.md
rata: Added

rata:
@wojtek-t PTAL, updated the KEP

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -595,6 +641,20 @@ use container runtime versions that have the needed changes.
...
+When the new functionality wasn't yet used, downgrade will not be affected.
+
+On downgrade, when the functionality was used, the pods created with
+user namespaces that are running will continue to run with user
+namespaces. Pods will need to be re-created to stop using the user
+namespace.
+
+Versions of Kubernetes that doesn't have this feature implemented will
+ignore the new field `pod.spec.hostUsers`.
+
```
wojtek-t, line 657: I already commented on that in https://github.com/kubernetes/enhancements/pull/4439/#discussion_r1475820339
but it got lost.

Can you please update verskiew skew strategy to cover also kube-apiserver (that may support the new feature whereas kubelet don't have it or have FG disabled)?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -725,7 +787,7 @@ Pods will have to be re-created to use the feature.
 
 We will add.
 
-We will test for when the field pod.spec.HostUsers is set to true, false
+We will test for when the field pod.spec.hostUsers is set to true, false
```
wojtek-t, line 790: See also the comment from the KEP teplate:
```
Additionally, for features that are introducing a new API field, unit tests that
are exercising the `switch` of feature gate itself (what happens if I disable a
feature gate after having objects written with the new field) are also critical.
You can take a look at one potential example of such test in:
https://github.com/kubernetes/kubernetes/pull/97058/files#diff-7826f7adbc1996a05ab52e3f5f02429e94b68ce6bce0dc534d1be636154fded3R246-R282
```

that is relevant here and extend this answer.


Also - can you please add enablement/disablement tests eplicitly as beta graduation criteria?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +815,17 @@ This section must be completed when targeting beta to a release.
 
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
+The rollout is just a feature flag on the kubelet and the kube-apiserver.
+
+If one APIserver is upgraded while other's aren't and you are talking to a not upgraded the pod
+will be accepted (if the apiserver is >= 1.25). If it is scheduled to a node that the kubelet has
+the feature flag activated and the node meets the requirements to use user namespaces, then the
+pod will be created with the namespace. If it is scheduled to a node that has the feature disabled,
+it will be created without the user namespace.
```
wojtek-t, line 824: I hope it's no longer true after our discussion - the pod should fail to start in the second case, right?

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +815,17 @@ This section must be completed when targeting beta to a release.
 
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
+The rollout is just a feature flag on the kubelet and the kube-apiserver.
+
+If one APIserver is upgraded while other's aren't and you are talking to a not upgraded the pod
```
wojtek-t, line 820: I would remove this first sentence - I don't think it matters really and it doesn't bring value.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +815,17 @@ This section must be completed when targeting beta to a release.
...
+will be accepted (if the apiserver is >= 1.25). If it is scheduled to a node that the kubelet has
+the feature flag activated and the node meets the requirements to use user namespaces, then the
+pod will be created with the namespace. If it is scheduled to a node that has the feature disabled,
+it will be created without the user namespace.
+
+On a rollback, pods created while the feature was active (created with user namespaces) will have to
+be re-created without user namespaces.
```
wojtek-t, line 827: ...  If those weren't recreated, they will continue to run in user namespace.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -765,13 +838,34 @@ will rollout across nodes.
 
 ###### What specific metrics should inform a rollback?
 
+On Kubernetes side, the kubelet should start correctly.
+
+On the node runtime side, a pod created with pod.spec.hostUsers=false should be on RUNNING state if
+all node requirements are met. If the CRI runtime or the handler do not support the feature, the kubelet
+returns an error.
```
wojtek-t, line 845: This puts the pod into Failed state, right?

So the number of Failed pods is effectively what we can look at, right?

[Not a native metric, but something measurable]

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -803,6 +899,10 @@ logs or events for this purpose.
 
 ###### How can someone using this feature know that it is working for their instance?
 
+If the runtime doesn't support user namespaces an error is returned by the kubelet.
```
wojtek-t, line 902: So the pod will effectively transition to Failed state, right?

keps/sig-node/127-user-namespaces/README.md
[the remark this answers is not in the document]
wojtek-t: Thank you - resolving and will add more targetted comments.

keps/sig-node/127-user-namespaces/README.md
wojtek-t: Can you clarify in the text?

keps/sig-node/127-user-namespaces/README.md
rata: Added now, sorry! It will be in the next push

keps/sig-node/127-user-namespaces/README.md
rata: Oh, great call. Added that to this section and to the beta criteria!

keps/sig-node/127-user-namespaces/README.md
rata: Okay. Sorry, I thought it would help as context, as there are lot of keps that added a field already and we all rely on existing plumbing to handle this correctly. But removed :)

keps/sig-node/127-user-namespaces/README.md
rata: @wojtek-t No, this is still true. If the feature is disabled in the kubelet, the kubelet completely ignores that field. The pod will only fail if the kubelet has the feature activated and the container runtime doesn't support it.

I don't think the kubelet can act on a field that should only act if the feature is enabled. We use this a lot on the error cases mentioned in detail in some other section, where you can just disable the feature gate to stop any bleeding that there might be related to this.

keps/sig-node/127-user-namespaces/README.md
rata: Oh, sorry. Added it now! Here and in the rest of the errors, that this is visible on the pod events.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -579,14 +628,11 @@ use container runtime versions that have the needed changes.
 
 ##### Beta
```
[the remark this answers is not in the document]
rata: Added, thanks!

keps/sig-node/127-user-namespaces/README.md
rata: I've added the following to clarify. 

---

When a pod hits this error returned by the kubelet, the status in `kubectl` is shown as
`ContainerCreating` and the pod events shows:

```
  Warning  FailedCreatePodSandBox  12s (x23 over 5m6s)  kubelet            Failed to create pod sandbox: user namespaces is not supported by the runtime
```

The following kubelet metrics are useful to check:
 - `kubelet_running_pods`: Shows the actual number of pods running
 - `kubelet_desired_pods`: The number of pods the kubelet is _trying_ to run

If these metrics are very different, it means there are desired pods that can't be set to running.
If that is the case, checking the pod events to see if they are failing for user namespaces reasons
(like the errors shown in this KEP) is advised, in which case it is recommended to rollback or
disable the feature gate.

keps/sig-node/127-user-namespaces/README.md
rata: Yes, there is a FailedCreatePodSandBox event. kubectl is shown as
ContainerCreating and the pod events shows:

```
  Warning  FailedCreatePodSandBox  12s (x23 over 5m6s)  kubelet            Failed to create pod sandbox: user namespaces is not supported by the runtime
```

rata:
@wojtek-t Fixed the latest review commets. PTAL! I've added it as a separate commit, so you can easily see the diff

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -610,6 +675,26 @@ enhancement:
...
+   that field and the kubelet will ignore it. Then, the pod is created without user namespaces.
+2. If the apiserver has the feature gate disabled and the kubelet enabled, the pod won't show this
+   field and therefore the kubelet won't act on a field that isn't shown. The pod is created without
+   user namespaces.
+
+The kubelet can still create pods with user namespaces if static-pods are configured with
+pod.spec.hostUsers and has the feature gate enabled.
+
+If the kube-apiserver doesn't support the feature at all (< 1.25), a pod with userns will be
+rejected.
```
wojtek-t, line 692: This is not true - kube-apiserver is not rejecting the pod, but rather is silently dropping this field.

Also - this isn't only about the release <1.25, but rather also when the feature gate is disabled.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -610,6 +675,26 @@ enhancement:
...
+
+If the kubelet doesn't support the feature (< 1.25), it will ignore the pod.spec.hostUsers field.
```
wojtek-t, line 694: If Kubelet is on old version not knowing the field - this is fine.

I'm more interested what happens if Kubelet is new enough and understands this field (say even 1.30), but it has feature gate disabled?
Will it ignore the field completely? [this isn't expected]
Or rather, will it reject the pod because it can't fulfill it? - which is what I would expect.

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -753,6 +843,19 @@ This section must be completed when targeting beta to a release.
...
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
+If one APIserver is upgraded while other's aren't and you are talking to a not
+upgraded one, the pod will be accepted (if the apiserver is >= 1.25, rejected if
+< 1.25).
+
+If it is scheduled to a node where the kubelet has the feature flag activated
+and the node meets the requirements to use user namespaces, then the pod will be
+created with the namespace. If it is scheduled to a node that has the feature
+disabled, it will be created without the user namespace.
```
wojtek-t, line 853: Same comment as for version skew.

keps/sig-node/127-user-namespaces/README.md
mrunalp: We should reject the pod in that case.

keps/sig-node/127-user-namespaces/README.md
rata: No, it is rejecting the pod. This is what happens when a field is known, for example:

```
$ kubectl apply -f mypod2.yaml 
Error from server (BadRequest): error when creating "mypod2.yaml": Pod in version "v1" cannot be handled as a Pod: strict decoding error: unknown field "spec.nonExistingField"
```

I added the `pod.spec.nonExistingField: true` to the pod.

Am I missing something? Or are we talking about different things?


Regarding the feature gate disabled, that is mentioned a few lines above: https://github.com/kubernetes/enhancements/pull/4439/commits/3766329491a71f33501ed417c8d34640258fffeb#diff-a9ca9fbce1538447b92f03125ca2b8474e2d875071f1172d2afa0b1e8cadeabaR684-R686

keps/sig-node/127-user-namespaces/README.md
rata: The current proposal is that it will ignore it completely. So feature disabled and an old kubelet that doesn't even know about this field, behave the same way.

We can change it to reject the pod in that case too (it is trivial). I proposed it this way, so by disabling the feature gate you can really stop any bleeding related to user namespaces. It seems more risky in case there is some bug, IMHO.

We can change it to that if you prefer, of course. As I said, it is trivial to do that too.

keps/sig-node/127-user-namespaces/README.md
rata: Idem.

keps/sig-node/127-user-namespaces/README.md
wojtek-t: I believe that we should reject (as I wrote above).

keps/sig-node/127-user-namespaces/README.md
wojtek-t: This is strange - the whole point of having the dropDisabledFields pattern is to handle and drop the field if the FG is disabled.
@liggitt - can you please shed some light on it?

keps/sig-node/127-user-namespaces/README.md
rata: Fixed to that!

keps/sig-node/127-user-namespaces/README.md
rata: Perfect, changed the text to do that instead. Thanks! :)

keps/sig-node/127-user-namespaces/README.md
rata: For context, this is the apiserver as started with the local-up-cluster.sh

liggitt: 1. kubectl opts into strict mode, where unknown fields are flagged / rejected. `kubectl apply ... --validate=false` shows what normal API clients (like controllers) would experience, where unknown fields are just dropped.
2. emulating the behavior where unknown fields are dropped on create when a feature is disabled is the standard approach taken.
3. *rarely*, for a security-related feature, we will reject attempts to use the new security-related field if the feature is disabled rather than dropping the field. The only one I can remember off the top of my head was the windows `hostProcess` field

haircommander: do we even need this sentence? 1.25 has been out of support for a while, and we'd also not support a 1.30->1.25 skew between kubelet and KAS

liggitt: > We aren't doing anything special in the apiserver for this field (pod.spec.hostUsers) on the apiserver.

Once you add the field, you need to start doing things with it in the apiserver when the feature is disabled.

haircommander: Updated to specify the kube-apiserver will drop the field if it's unknown

haircommander: how does it look @liggitt @wojtek-t

wojtek-t:
OK - with the recent changes, I'm fine with it from PRR pov.

/lgtm
/approve PRR

PR #4508, opened 2024-02-09
Mark client-go context as stable to match GA criteria in KEP

jpbetz:
Given https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1601-client-go-context#graduation-criteria:

"This enhancement will be GA immediately."

Let's track in in kep.yaml as such.

/sig api-machinery


deads2k:
A change that will be remembered :)

/lgtm

PR #4510, opened 2024-02-09
KEP-2887: modify GA criteria

jiahuif:
- depends on Declarative Validation.

  
- One-line PR description:
This PR modifies GA criteria of KEP-2887 to delegate the validation of tagged enum types to #4153

- Issue link: #2887

- Other comments:


jpbetz (approved):
/approve

I agree that, with the Declarative validation KEP being better suited for addressing this validation change, that we shouldn't block this KEP on it, especially since it's not a user visible change.

Holding LTGM briefly in case @deads2k has any concerns

PR #4557, opened 2024-03-21
KEP 4568: Resilient watchcache initialization

wojtek-t:
/sig api-machinery

@deads2k @jpbetz

### DIFF
--- a/keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
+++ b/keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
@@ -1,0 +1,479 @@
+# KEP-4568: Resilient watchcache initialization
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [Reduce the number of requests during initialization](#reduce-the-number-of-requests-during-initialization)
+    - [Reject hanging watches](#reject-hanging-watches)
+    - [Delegate get requests to etcd](#delegate-get-requests-to-etcd)
+    - [Adjust what lists are delegated to etcd](#adjust-what-lists-are-delegated-to-etcd)
+    - [Reject the rest of list requests](#reject-the-rest-of-list-requests)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+There are several issues that can lead to significant overload of kube-apiserver and
+etcd (or worst case even bringing it down) during its initialization or reinitialization
+of the watchcache layer. The KEP proposes a solution for mitigating these issues.
+
+## Motivation
+
+Watchcache is a caching layer inside kube-apiserver that is eventually consistent
+cache of cluster state stored in etcd - both the state itself as well as the most
+recent history of transaction log. It is critical to achieve the control plane
+scalability thanks to its built-in indexing and deduplicating of the same work
+needed e.g. to distribute the same watch even to multiple watches.
+
+However, there are several issues when watchcache is either not yet initialized
+on kube-apiserver startup or requires re-initialization due to not keeping up with
+load later on. Both of these cases may lead to significant overload of the whole
+control plane (including etcd), in worst case even bringing it down.
+
+The goal of this KEP is to describe the current issues and provide the path for
+addressing or at least mitigating them to avoid those control plane outages.
+
+### Goals
+
+- Describe issues that can be faced during watchcache initialization
+- Provide and analyze tradeoffs of potential mitigations and decide for path forward
+
+### Non-Goals
+
+- Redesigning watchcache or the storage layer itself
+
+
+## Proposal
+
+Watchcache is mostly delegating all the write operations to the next layer (being
+etcd3 implementation of storage interface). The only thing it is doing (for Update
+and Delete calls) is trying to get the current state of the object from its cache
+to avoid the need of reading it from etcd (in case of conflict it may still be
+needed, but in huge amount of situations it allows to avoid this roundtrip).
+What happens during initialization is that the cache is empty, so we're not able
+to retrieve the object from cache, in which case we simply avoid passing the
+current object down and it's then retrieved from etcd. We didn't observe nor
+heard about this causing real issues and we're not going to change this behavior
+as part of this KEP.
+
+For watch requests, unless it's a consistent watch request (setting RV="") - which
+isn't really used by none of our core components, it is always handled by the
+watchcache. For all such requests, watch is hanging until watchcache finishes
+initialization. The problem here is that for clusters with a large amount of data
+of a given type, initialization (being effectively a list from etcd) can take
+tens of seconds or even above a minute and in the meantime this watch is consuming
+seats of API Priority&Fairness. In many cases at most tens of such watches (which
+is a relatively small number in clusters with thousands of nodes) can effectively
+completely starve a given PriorityLevel and block any load on it (both write and
+read requests).
+
+For get/list requests that explicitly opt-in for being served from watchcache
+(namely those marked as `NotOlderThan` or `Any` in [the documentation]) - which
+amongst others is what Reflector/Informers framework is using in most of cases,
+if watchcache is not initialized there are two cases:
+- if ResourceVersion=0, these are blindly delegated to etcd
+- if ResourceVersion>0, they hang waiting for initialization similarly as watch
+requests
+
+The first case is particularly problematic for lists that hugely benefit from
+watchcache indexing (e.g. listing pods assigned to a given node). When served
+from watchcache, these requests are really efficient, but serving them from etcd
+requires reading a large amount of data, deserializing it, filtering out objects
+not matching selector and sending only those to the client. Passing too many of
+such requests down to etcd, can easily bring down the whole control plane.
+Additionally, in this case, when served from watchcache a LIMIT parameter is
+ignored and the whole result is returned in one go. If served from etcd, the
+LIMIT parameter is honored, so for large collections we end-up paging the
+response (which for large collections and limit=500 of reflector/informer
+frameworks can result in hundreds of consecutive requests for next pages).
+
+The second case is suffering from the same problem as watches, as those requests
+may effectively starve the whole PriorityLevel.
+
+In order to mitigate the above problem, we propose a couple different changes.
+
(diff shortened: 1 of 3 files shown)


keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,766 @@
...
+returned, so we're protecting from requests very selective list requests that
+are likely order(s) of magnitude more expensive when served from etcd.
+The second condition effectively bounds the cost of the request.
+
+We may decide to adjust it based on further feedback and/or experiments.
+
+#### Reject the rest of list requests
+
+To mitigate the problem of starving certain PriorityLevels, all other list
+requests will be rejected with 429 http code similarly to watch requests.
```
wojtek-t, line 195: TODO: Same for Get() requests.

wojtek-t: Alternatively, we can just pass them to etcd, given we already got to this point and those are much cheaper - might be better idea.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,766 @@
...
+Given we're changing the existing behavior, there is a risk of breaking some
+clients. To mitigate it, we will introduce all the logic behind a feature
+gate, to allow for disabling them if needed.
+
+However, given we're not changing the API itself, we propose to go directly
+to Beta, as Alpha will not give us any feedback.
+
+## Design Details
+
+<!--
```
wojtek-t, line 209: The POC: https://github.com/kubernetes/kubernetes/pull/124029

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+[properly handled]: https://github.com/kubernetes/kubernetes/blob/41e706600aea7468f486150d951d3b8948ce89d5/staging/src/k8s.io/client-go/tools/cache/reflector.go#L910-L920
+
+#### Delegate get requests to etcd
+
+Since GET requests have bounded cost, we always estimate their weight the same
+no matter if they are served from etcd or from cache (the latency would be
+different in those cases though, so the actual seat-seconds cost too) and
+finally given we already did some work to process this request, we would
+simply delegate those requests to etcd.
```
jpbetz, line 181: I really like how this KEP, combined with _list-watch_ and _consistent reads from cache_ all seem to move the system in a direction where the actual cost of GET requests will have less expected variation than they do today.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+#### Adjust what lists are delegated to etcd
+
+It's tempting to reject all list requests with 429 the same way as watches.
+However, that would also have a negative consequences, e.g. by slowing down
+kube-apiserver initialization (kube-apiserver on startup need to initialize
+its own informers by loop-client, and by rejecting all lists until watchcache
+is initialized, we would effectively block that initialization).
```
jpbetz, line 189: List-Watch would return 429?  Do we benefit from getting list-watch enabled-by-default before enabling this enhancement by default?

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+
+### Risks and Mitigations
+
+Given we're changing the existing behavior, there is a risk of breaking some
+clients. To mitigate it, we will introduce all the logic behind a feature
+gate, to allow for disabling them if needed.
+
+However, given we're not changing the API itself, we propose to go directly
+to Beta, as Alpha will not give us any feedback.
```
jpbetz, line 214: cc @deads2k

I haven't put much thought into what the criteria should be for skipping alpha.  This would be on-by-default I'm assuming?

Clearly all Beta PRR criteria would need to be met.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+#### Adjust what lists are delegated to etcd
+
+It's tempting to reject all list requests with 429 the same way as watches.
+However, that would also have a negative consequences, e.g. by slowing down
+kube-apiserver initialization (kube-apiserver on startup need to initialize
+its own informers by loop-client, and by rejecting all lists until watchcache
+is initialized, we would effectively block that initialization).
+
+While we don't may decide to adjust the approach here, we will start with
+the approach that we will delegate to etcd the requests that (a) are not
```
jpbetz, line 192:
```suggestion
While we didn't decide to reject _all_ list requests with 429 here, we will start with
an approach where we only delegate to etcd the requests that (a) are not
```

(wording suggestion only.. please only accept if this matches what was intended)

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+PostStartHook that will wait for watchcache of all *builtin* resources to be initialized.
+
+Why this should help with reducing the number of requests? The whole idea behind
+kube-apiserver readiness is that Kubernetes installations should be configured so that
+requests are not sent to it until kube-apiserver becomes ready. The above will allow us
+to delay this moment until watchcache is initialized.
+
+Why PostStartHook instead of /readyz check? Watchcache is per-type layer and we want
+to avoid marking whole kube-apiserver as not-ready if watchcache for one of resources
+requires reinitialization. We will handle those cases differently.
```
jpbetz, line 157: Just making sure I understand.  Would the first initialization of each builtin type be /readyz gated but any re-initialization be post- /readyz (but still result in 429 responses)?

wojtek-t: The way poststarthook works is just it effectively:
1) is starteded once:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/hooks.go#L194-L206
2) it's reported as healthy as soon as it succeeds for the first time:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/hooks.go#L238-L245
[and it never changes back to false later]

So what will happen is effectively:
- first initialization of each builtin type will prevent readyz from succeeding [initialization of kube-apiserver]
- even if watchcache needs to reinitialize later, it will not bring readyz/ back to failing, but as you wrote, requests [which exactly described below] will be returning 429

tried to clarify in the text

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+Finally, we suggest starting just with builtin resources, as CRDs can be created at
+any time, making it harder to implement with unclear gain. If it appears to not be
+enough, this decision can easily be revisited in the future.
```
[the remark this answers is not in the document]
wojtek-t: No - that seems easy to adjust later too. Added to the text.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
wojtek-t: Yes - ListWatch is effectively a regular Watch call underneathy (with some specific params, but handled by Watch()) methods. So yes - it will also return 429.

I don't think enabling list-watch first effectively gives us anything. Given that we want the behavior to be the same for watches and lists, the switch to list-watch should be no-op from that perspective.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
wojtek-t: it matches - thanks for clarifying

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+
+It's tempting to reject all list requests with 429 the same way as watches.
+However, that would also have a negative consequences, e.g. by slowing down
+kube-apiserver initialization (kube-apiserver on startup need to initialize
+its own informers by loop-client, and by rejecting all lists until watchcache
+is initialized, we would effectively block that initialization).
+
+While we don't may decide to adjust the approach here, we will start with
+the approach that we will delegate to etcd the requests that (a) are not
+setting any selectors (b) are setting a limit.
```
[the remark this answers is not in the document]
wojtek-t: AND - added to the text

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
wojtek-t: > This would be on-by-default I'm assuming?

Yes

> Clearly all Beta PRR criteria would need to be met.

of course - whole PRR still requires filling in


Regarding the main comment - for me Alpha stage makes sense for things that change API [e.g. to enable rollback]. For purely in-memory changes (like this one), alpha is no-op and noone uses it, so we effectively have zero benefit from it.

wojtek-t (commented):
@jpbetz - PTAL

[still requires filling in PRR, but the rest should be more-or-less ready]

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,764 @@
...
+As a first step, we will try to reduce the number of requests that are even reaching
+kube-apiserver with unitialized watchcache. To achieve it, we will introduce a new
+PostStartHook that will wait for watchcache of all *builtin* resources to be initialized.
+
+Why this should help with reducing the number of requests? The whole idea behind
+kube-apiserver readiness is that Kubernetes installations should be configured so that
+requests are not sent to it until kube-apiserver becomes ready. The above will allow us
+to delay this moment until watchcache is initialized.
+
+Why PostStartHook instead of /readyz check? Watchcache is per-type layer and we want
```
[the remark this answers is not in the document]
logicalhan: Yes, I realize that, I was (in a roundabout way) agreeing with your reasoning here.

keps/sig-api-machinery/xxx-resilient-watchcache-initialization/README.md
jpbetz: I support going straight to beta for this. I consider startup readiness something we should be able to refine w/o going through all the stability levels.

jpbetz:
This LGTM

wojtek-t:
> This LGTM

Thanks - I switched to implementable and added prod-readiness file.

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,462 @@
...
+
+In case of bugs:
+- kube-apiserver may not initialize correct due to buggy implementation of the
+  introduced post-start hook
+- kube-apiserver may incorrectly reject some requests with 429
+
+###### What specific metrics should inform a rollback?
+
+- kubeapiserver /readyz behaving visibly different than before
+- `apiserver_request_total` (especially watch for requests finished with 429 code)
```
jpbetz, line 344: Be more prescriptive here to make it extra obvious to a cluster admin what to watch for?  E.g. "If an apiserver continues to respond to /readyz requests with X significantly after startup should have completed ..."

jpbetz (commented):
PRR looks straightforward. One suggestion then LGTM.

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
wojtek-t: PTAL

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+In order to mitigate the above problem, we propose a couple different changes.
+
+[the documentation]: https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list
+
+#### Reduce the number of requests during initialization
+
+As a first step, we will try to reduce the number of requests that are even reaching
+kube-apiserver with unitialized watchcache. To achieve it, we will introduce a new
+PostStartHook that will wait for watchcache of all *builtin* resources to be initialized.
```
deads2k, line 150: Haven't finished yet, but this will also include KMS handling on the path for Secrets for instance.  While I'd like to be more efficient, I think we should  have a max-time that we're willing to wait.

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+#### Adjust what lists are delegated to etcd
+
+It's tempting to reject all list requests with 429 the same way as watches.
+However, that would also have a negative consequences, e.g. by slowing down
+kube-apiserver initialization (kube-apiserver on startup need to initialize
+its own informers by loop-client, and by rejecting all lists until watchcache
+is initialized, we would effectively block that initialization).
+
+While we didn't decide to reject _all_ list requests with 429, we will start with
```
deads2k, line 203: could this be a "stick" to have people start using streaming lists?  Streaming lists always allowed, but regular lists rejected?

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,464 @@
...
+
+
+### Risks and Mitigations
+
+Given we're changing the existing behavior, there is a risk of breaking some
+clients. To mitigate it, we will introduce all the logic behind a feature
+gate, to allow for disabling them if needed.
+
+However, given we're not changing the API itself, we propose to go directly
+to Beta, as Alpha will not give us any feedback.
```
deads2k, line 226: Alpha gives the interested parties a chance to test this without impacting default installs (still often used in production).

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,464 @@
...
+Finally, we suggest starting just with builtin resources, as CRDs can be created at
+any time, making it harder to implement with unclear gain. If it appears to not be
+enough (or alternatively we realize that only a subset of builtin resource types is
+needed), this decision can easily be revisited in the future.
+
+#### Reject hanging watches
+
+To mitigate the problem of starving certain PriorityLevels, instead of having the
+watch hang and wait for watchcache initialization, we will simply reject it with
+`Too Many Requests` 429 http code.
```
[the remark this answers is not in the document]
wojtek-t: That comment was for get/list requests.
This subsections is for watches.

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
wojtek-t: No. Streaming lists will never be allowed here as they are served from watchcache - so if the watchcache is not initialized, we will not be able to serve them at all.

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
wojtek-t: Realistically, providers are not enabling Alpha. There are users who enable Alpha, because they want a given feature, but for changes like this (effectively not being a feature), I doubt anyone would enable it.

That said - are you against the decision of going directly to beta, or was it more of a wording issue with the current text?

deads2k: > That said - are you against the decision of going directly to beta, or was it more of a wording issue with the current text?

I'm against going directly to beta.  I think we should produce an alpha that we can use to see if the feature achieves its goals before we enable such a change on all current default clusters (commonly used in production).

keps/sig-api-machinery/4568-resilient-watchcache-initialization/README.md
```diff
@@ -0,0 +1,479 @@
...
+implement them with desired quality and coverage, we propose to go directly
+to Beta.
+However, we admit that the risk in both cases is different:
+- for the new post-start hook, there is a risk of kube-apiserver not initializing,
+   thus we will start with Beta disabled by-default
+- for the changes to handing requests when watchcache is uninitialized, we
+   believe the risk is visibly lower (and 429 errors were already returned
+   before, even in small clusters where priority-levels are small enough
+   to often admit only 1-2 inflight requests at once anyway) and we will
+   enable it by-default from the very beginning
```
deads2k, line 241: lgtm, thanks.

deads2k:
Thanks for the enablement updates.  This lgtm from the sig perspective.  I'll leave PRR and approve with @jpbetz who has also reviewed already.

/lgtm
/assign @jpbetz

PR #4575, opened 2024-04-11
 KEP-3962: rewrite CEL library section.

jiahuif:
- One-line PR description: This PR adds information about CEL library changes to MutatingAdmissionPolicy feature.

- Issue link: https://github.com/kubernetes/enhancements/issues/3962

- Other comments:
This PR removes the question mark from removal operations.
Before
```yaml
mutations:
  - mutaton: >
        Object{
            spec: Object.spec{
                containers: object.spec.containers{
                    object.spec.containers.filter(c, c.name == "sidecar")
                    .map(c, Object.spec.containers.item{
                        ?env: optional.none()
                    })
                }
            }
        }
```
After

```yaml
mutations:
  - mutaton: >
        Object{
            spec: Object.spec{
                containers: object.spec.containers{
                    object.spec.containers.filter(c, c.name == "sidecar")
                    .map(c, Object.spec.containers.item{
                        env: optional.none()
                    })
                }
            }
        }
```

See content for details.

### DIFF
--- a/keps/sig-api-machinery/3962-mutating-admission-policies/README.md
+++ b/keps/sig-api-machinery/3962-mutating-admission-policies/README.md
@@ -447,7 +447,7 @@ feature to make it possible to express that a value should be unset. For example
 ```cel
 Object{
   spec: Object.spec{
-    ?fieldToRemoveIfPresent: optional.none()
+    fieldToRemoveIfPresent: optional.none()
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


cici37:
/lgtm
I am ok with not using question mark to avoid confusion. Previously we add it mainly for the consideration of field not exist case. With schema check in place, it shouldn't be an concern. Thank you!

PR #4588, opened 2024-04-22
KEP-3983: update the beta requirements

sohankunkerkar:
We already have a method to query the Kubelet's full effective configuration, so there's no need to implement that API. We can simply use this query: `kubectl get --raw "/api/v1/nodes/192.168.1.16/proxy/configz"`

  
- One-line PR description:

- Issue link: https://github.com/kubernetes/enhancements/issues/3983

- Other comments:


mrunalp (approved):
/lgtm

PR #4608, opened 2024-04-30
Consistent reads mitigation

wojtek-t:
Ref #2340

/cc @serathius @deads2k

### DIFF
--- a/keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
+++ b/keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
@@ -253,6 +253,33 @@ Since falling back to etcd won't work, we should fail the requests and rely on
 rate limiting to prevent cascading failure.  I.e. `Retry-After` HTTP header (for
 well-behaved clients) and [Priority and Fairness](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190228-priority-and-fairness.md).
 
+In order to mitigate such problems, let's present how the system currently works
+in different cases.
+
+| ResourceVersion | ResourceVersionMatch | Continuation      | Limit         | etcd implementation                     | watchcache implementation                          |
+|-----------------|----------------------|-------------------|---------------|-----------------------------------------|----------------------------------------------------|
+| _unset_         | _unset_              | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  |
+| _unset_         | _unset_              | _token_           | _unset_ / _N_ | Read request from RV encoded in _token_ | Delegated to etcd                                  |
+| _unset_         | _Exact_              | _unset_ / _token_ | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |
+| _unset_         | _NotOlderThan_       | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  |
+| _unset_         | _NotOlderThan_       | _token_           | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |
+| _0_             | _unset_              | _unset_           | _unset_ / _N_ | Quorum read request                     | List from cache ignoring _limit_                   |
+| _0_             | _unset_              | _token_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  |
+| _0_             | _Exact_              | _unset_ / _token_ | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |
+| _0_             | _NotOlderThan_       | _unset_           | _unset_ / _N_ | Quorum read request                     | List from cache ignoring _limit_                   |
+| _0_             | _NotOlderThan_       | _token_           | _unset_ / _N_ | Read request from RV encoded in _token_ | Delegated to etcd                                  |
+| _RV_            | _unset_              | _unset_           | _unset_       | Quorum read request                     | Wait for cache synced to _RV_+ and list from cache |
+| _RV_            | _unset_              | _unset_           | _N_           | Read request from RV=_RV_               | Delegated to etcd                                  |
+| _RV_            | _unset_              | _token_           | _unset_ / _N_ | Read request from RV encoded in _token_ | Delegated to etcd                                  |
+| _RV_            | _Exact_              | _unset_           | _unset_ / _N_ | Read request from RV=_RV_               | Delegated to etcd                                  |
+| _RV_            | _Exact_              | _token_           | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |
+| _RV_            | _NotOlderThan_       | _unset_           | _unset_       | Quorum read request + check for _RV_    | Wait for cache synced to _RV_+ and list from cache |
+| _RV_            | _NotOlderThan_       | _unset_           | _N_           | Quorum read request + check for _RV_    | Delegated to etcd                                  |
+| _RV_            | _NotOlderThan_       | _token_           | _unset_/ _N_  | Fails [validation]                      | Fails [validation]                                 |
+
+[validation]: https://github.com/kubernetes/kubernetes/blob/release-1.30/staging/src/k8s.io/apimachinery/pkg/apis/meta/internalversion/validation/validation.go#L28
+[etcd resolution]: https://github.com/kubernetes/kubernetes/blob/release-1.30/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L589-L627
+
 For such situations we will provide users with following tools:
 * a dedicated `apiserver_watch_cache_read_wait` metric to detect a problem with
   watch cache.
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
```diff
@@ -253,6 +253,34 @@ Since falling back to etcd won't work, we should fail the requests and rely on
...
+case will change how watchcache implementation will be handling the request.
+
+| ResourceVersion | ResourceVersionMatch | Continuation      | Limit         | etcd implementation                     | watchcache implementation                          | changed  |
+|-----------------|----------------------|-------------------|---------------|-----------------------------------------|----------------------------------------------------|----------|
+| _unset_         | _unset_              | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  | Yes      |
+| _unset_         | _unset_              | _token_           | _unset_ / _N_ | Read request from RV encoded in _token_ | Delegated to etcd                                  |          |
+| _unset_         | _Exact_              | _unset_ / _token_ | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |          |
+| _unset_         | _NotOlderThan_       | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  | Yes      |
```
deads2k, line 265: So this one of the rows in question.  after this kep, this request will be handled by the watch cache.

The question then becomes how does a user request `unset, NotOlderThan, unset, unset` from etcd.  

One proposal is that in order to request `ResourceVersionMatch=NotOlderThan` served from etcd, the user should request `ResourceVersionMatch=Exact` instead.  Thus `unset, NotOlderThan, unset, unset` becomes `unset, Exact, unset, unset`.

Another proposal is to have the user specify another option and indicate `ServeFromEtcd` or `ServeFromWatchCache`.  Thus to request request `ResourceVersionMatch=NotOlderThan` served from etcd, an additional parameter is added `unset, NotOlderThan, unset, unset, ServeFromEtcd`.

I see key decision points as
1. Is secondary authorization for `ServeFromEtcd` necessary? 
2. Is there value in separately selecting `ServeFromEtcd` or `ServeFromWatchCache` for every combination of options?
3. Is it easier to expand the set of allowable tuples for ResourceVersionMatch than adding a new option?  Consider both clients and servers.
4. Does choosing an obfuscated name for `ServeFromEtcd` make the usage of said option less likely?
5. Does adding a new option provide clearer semantics

My current thoughts:
1. I think past-Wojtek argued convincingly for this to be necessary.
2. I think there is value in doing this for all watch-cache back APIs, so List, Get, and Watch. And I think keeping the "do this" aspect of the query consistent and changing only the `ServerFromX` option.
3. I don't think the ResourceVersionMatch is notably easier server-side.  It doesn't achieve any secondary authorization benefit (or if it did, would be more complex).  Client-side I think it's harder to explain.  I'm not even sure that `unset, NotOlderThan, unset, unset`  is the same as `unset, Exact, unset, unset` or if it would mean something different.  Naively I would expect it to mean something different.
4. I don't think dissuading by naming oddly is effective
5. I do think that adding a new option provides clearer semantics due to the ability to universally apply to all requests and exact parity with all other options.  So questions about "does unset mean the same as Exact here?" and "does NotOlderThan mean unset here?".

keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
```diff
@@ -253,6 +253,34 @@ Since falling back to etcd won't work, we should fail the requests and rely on
...
+| _RV_            | _Exact_              | _token_           | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |          |
+| _RV_            | _NotOlderThan_       | _unset_           | _unset_       | Quorum read request + check for _RV_    | Wait for cache synced to _RV_+ and list from cache |          |
+| _RV_            | _NotOlderThan_       | _unset_           | _N_           | Quorum read request + check for _RV_    | Delegated to etcd                                  | Deferred |
+| _RV_            | _NotOlderThan_       | _token_           | _unset_/ _N_  | Fails [validation]                      | Fails [validation]                                 |          |
+
+[validation]: https://github.com/kubernetes/kubernetes/blob/release-1.30/staging/src/k8s.io/apimachinery/pkg/apis/meta/internalversion/validation/validation.go#L28
+[etcd resolution]: https://github.com/kubernetes/kubernetes/blob/release-1.30/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L589-L627
+
```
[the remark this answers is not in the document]
liggitt: wouldn't we would continue to look for ways to shift _more_ rows in the table with "Delegated to etcd" to be able to be fulfilled via the cache in the future?

I don't think using `RVM=Exact` only in combinations that currently fail validation as the only way to signal bypassing the cache actually lets us debug / troubleshoot watch cache behavior for all the rows that we're changing now, or will likely change in the future.

keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
liggitt: My thoughts on two of those questions:

1. I agree *some* check that limits explicitly bypassing the cache is needed.
2. Since the reason for the explicit "bypass cache" option is to detect / debug / troubleshoot watch cache behavior, keeping all other options identical and changing only the "bypass cache" option is the only way I see to effectively do that. Otherwise, we don't have the ability to debug / troubleshoot watch cache behavior that only manifests for the `ResourceVersionMatch=NotOlderThan` case, for example.

wojtek-t (commented):
I also propagated table for watch requests.

wojtek-t:
@p0lyn0mial @serathius

serathius:
I largely agree with @wojtek-t. After further consideration, I don't think a dedicated "SkipCache" API option is necessary in this scenario. We shouldn't view watchcache as a traditional database cache like memcache or varnish. Skipping cache for debugging purposes is suitable for periodically refreshed and invalidated caches to identify stale data. It operates under the expectation of consistent results. However, in distributed systems, it's normal to receive different results from concurrent read requests.

In distributed systems like etcd, two concurrent read requests can target different instances, leading to different pathing and arbitrary network delays, allowing for interleaved transactions and reordering. The only way to determine their "real" execution order is by checking the revision returned by etcd. Responses from two different reads should only match when they have the same revision. However, even in medium-sized clusters, there's enough traffic that two concurrent consistent read requests will often return different revisions. To better understand that I recommend reading   https://jepsen.io/consistency/models/linearizable

The only way to debug watch cache corruption is to verify if the cached data matches the data in etcd. If you want to use the K8s API for this, the only way is to make two requests that share the same ResourceVersion (RV). This means the correct flow to compare watchcache state to etcd state is:

1. Make a request without an RV (which will be served from the watch cache).
2. Follow it with a request that includes the exact RV from the first request and uses ResourceVersionMatch=Exact. (which is served from etcd)

p0lyn0mial:
My view on this case is that instead of adding a new list option that allows bypassing the cache, we should better test the feature before enabling it globally.

Mainly, my concern is that today, when the feature `ConsistentListFromCache` is enabled (or any other), it becomes available to all clients simultaneously. In my opinion, we should have a graduation mechanism and initially enable the feature for a specific client, a client with some attribute, or for a percentage of traffic coming to the server. Subsequently, we should gradually increase the number of clients or the percentage of traffic directed to the server until finally enabling the feature for all clients.

Additionally, as @serathius mentioned, we should (especially in CI) have a mechanism for detecting inconsistencies in data by comparing data returned from the cache and directly from etcd. For that we could use the [data_consistency_detector](https://github.com/kubernetes/kubernetes/blob/e798b9c269627a10291f36a88b4fa13f034ab57f/staging/src/k8s.io/client-go/tools/cache/reflector_data_consistency_detector.go) by issuing two request to the server.

deads2k:
Thank you @wojtek-t and @serathius for taking the time to put together the list, describe the potential debugging flows, and answer questions live.  I think I've reached where you are: *we should not create a new option for bypassing the watch cache.*

While it seems congruent to add a parameter for "bypass watch cache" and it could be useful if the watch cache is misbehaving, such an option is not useful for debugging answers from the watch cache.  This is because the wache cache cannot support `Exact` and etcd cannot support `NotOlderThan` and that prevents any congruent etcd and watch cache access.

This means that the only debugging flow available is make a watch cache request and then use the existing option `RV 	Exact 	unset 	unset`.  I like the idea of making this an opt-in self-check available in our client as referenced above.
> Additionally, as @serathius mentioned, we should (especially in CI) have a mechanism for detecting inconsistencies in data by comparing data returned from the cache and directly from etcd. For that we could use the [data_consistency_detector](https://github.com/kubernetes/kubernetes/blob/e798b9c269627a10291f36a88b4fa13f034ab57f/staging/src/k8s.io/client-go/tools/cache/reflector_data_consistency_detector.go) by issuing two request to the server.

wojtek-t:
@deads2k - I updated the PR to reflect our discussion today, PTAL

PR #4616, opened 2024-05-03
Add GA criteria for informer support

jpbetz:


deads2k:
I agree we should add this.  It will make it possible to select from listers for fields as do for labels today.

/lgtm

PR #4623, opened 2024-05-08
KEP-1040: Prescribe less extreme borrowing by exempt priority levels

MikeSpreitzer:
- One-line PR description: This revises the borrowing between priority levels from something that scares me a lot (and is not what is implemented) to something that scares me less.

- Issue link: 

- Other comments: There is an issue in k/k about the discrepancy between KEP and code: https://github.com/kubernetes/kubernetes/issues/124125


deads2k:
discussed long ago in a p&f sync.  This matches my expectations from that call.

/lgtm
/approve

PR #4630, opened 2024-05-13
Adds Kat Cosgrove as SIG Docs tech lead

katcosgrove:
Adding myself as SIG Docs tech lead!

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -45,6 +45,7 @@ aliases:
     - palnabarun
   sig-docs-leads:
     - divya-mohan0209
+    - katcosgrove
     - kbhawkey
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -45,6 +45,7 @@ aliases:
     - palnabarun
   sig-docs-leads:
     - divya-mohan0209
+    - katcosgrove
     - kbhawkey
     - natalisucks
     - onlydole
```
mrbobbytables, line 51: Can be done in a follow up - but should sync thing with current state kbhawkey and onlydole should be removed and everyone else added^^;;

/lgtm
/approve

PR #4714, opened 2024-06-11
Correction to the Job pod failure policy update to GA

mimowo:
- One-line PR description: Correction to the GA graduation plan

- Issue link: https://github.com/kubernetes/enhancements/issues/3329

- Other comments: I have realized the current graduation criteria are not consistent. We cannot add the test if we lock the feature gates JobPodFailurePolicy and PodDisruptionConditions. It is too late. Alternatively, we could move the lock to GA+1, and removal to GA+3. Since the disablement and re-enablement was tested manually for upgrade tests I propose to just drop the automated test: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/3329-retriable-and-non-retriable-failures#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested.


soltysh (approved):
/lgtm
/approve

PR #4721, opened 2024-06-12
list semantics: resourceVersionMatch is forbidden unless resourceVersion is provided

p0lyn0mial:


keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md
```diff
@@ -263,7 +263,7 @@ case will change how watchcache implementation will be handling the request.
 | _unset_         | _unset_              | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  | Yes      |
 | _unset_         | _unset_              | _token_           | _unset_ / _N_ | Read request from RV encoded in _token_ | Delegated to etcd                                  |          |
 | _unset_         | _Exact_              | _unset_ / _token_ | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |          |
-| _unset_         | _NotOlderThan_       | _unset_           | _unset_ / _N_ | Quorum read request                     | Delegated to etcd                                  | Yes      |
+| _unset_         | _NotOlderThan_       | _unset_           | _unset_ / _N_ | Fails [validation]                      | Fails [validation]                                 |          |
```
wojtek-t, line 266: Good catch:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/internalversion/validation/validation.go#L33-L36

PR #4741, opened 2024-06-19
KEP 1880 stay two releases on beta, one off and other on

aojea:
stay two releases on beta, one off and other on


thockin (commented):
Thanks!

/lgtm
/approve

PR #4863, opened 2024-09-19
remove total resource calculation proposal from the sidecar KEP

SergeyKanzhelev:
/sig node
KEP:  #753

We tried once but we always get in debates with the alpha features that are coming up. https://github.com/kubernetes/kubernetes/pull/115643

Last one is that PodLevel resource KEP will need to pick up this work.


mrunalp (approved):
/lgtm

PR #4936, opened 2024-10-30
KEP-740: Update with decision on max token lifetime

HarshalNeelkamal:
- One-line PR description:
Choice of max token lifetime was reconsidered during implementation. Documenting it accordingly. 

- Issue link:
https://github.com/kubernetes/enhancements/issues/740

- Other comments:
For ref: https://github.com/kubernetes/kubernetes/pull/128190#discussion_r1821412614


HarshalNeelkamal:
cc/ @liggitt

PR #4937, opened 2024-10-30
KEP-2558: mark as "withdrawn"

thockin:
- One-line PR description:

Nobody is working on this idea, and it has drifted from the original goal.

- Issue link: https://github.com/kubernetes/enhancements/issues/2558


jpbetz:
An interesting idea, but yes, without a champion, I agree that marking the KEP this way is appropriate.

/lgtm
/approve

PR #5017, opened 2024-12-30
Pagination from cache KEP

serathius:
Create first draft of https://github.com/kubernetes/enhancements/issues/4988 as provisional.

/cc @wojtek-t @deads2k @MadhavJivrajani @jpbetz

### DIFF
--- a/keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
+++ b/keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
@@ -1,0 +1,395 @@
+# KEP-4988 Serve pagination from cache
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [Client setting limit while not supporting pagination](#client-setting-limit-while-not-supporting-pagination)
+    - [Memory overhead](#memory-overhead)
+    - [Pagination request hitting another apiserver](#pagination-request-hitting-another-apiserver)
+    - [Delegating slow pagination to etcd](#delegating-slow-pagination-to-etcd)
+    - [Increased watch contention](#increased-watch-contention)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Enabling pagination typically improves performance by reducing response sizes.
+However, the kube-apiserver's caching mechanism (watchcache) does not currently
+support pagination. When a client requests a paginated list, the limit parameter
+might be ignored, or the request might bypass the cache and be served directly
+from etcd, leading to performance issues. This document proposes an enhancement
+to the kube-apiserver's caching layer to enable efficient serving of paginated
+lists from the cache, improving performance and predictability.
+
+## Motivation
+
+The kube-apiserver's watchcache is an eventually consistent cache of the cluster
+state. It allows the API server to serve read requests without constantly hitting
+etcd. However, watchcache is not designed for pagination.
+
+Pagination allows clients to request large lists of resources in smaller chunks.
+The client specifies a limit (maximum number of objects per page) and receives
+a continue token in the response. This token is then used in subsequent requests
+to fetch the next page.
+
+To serve paginated requests correctly, the API server needs access to a consistent
+snapshot of the data from which the initial page was generated. Watchcache does
+not store historical data. Therefore, currently, the API server does the following:
+
+- For `resourceVersion=""`: Serve paginated requests directly from etcd, bypassing the cache.
+- For `resourceVersion="0"`: Ignore the limit parameter and return the entire list in a single response.
+
+Both options are problematic:
+
+- Serving from etcd negates the benefits of the cache,
+  leading to increased load on both the API server and etcd.
+- Ignoring the limit parameter defeats the purpose of pagination,
+  potentially returning massive responses that can overwhelm clients.
+
+This situation leads to unexpected behavior for users,
+where enabling pagination can degrade performance instead of improving it.
+
+### Goals
+
+- Improve the performance of paginated list requests to be comparable to non-paginated lists.
+- Make pagination behavior predictable and consistent.
+- Reduce load on etcd caused by paginated requests.
+
+### Non-Goals
+
+- Serve `resourceVersion="N"` request from watch cache
+- Support indices when paginating.
+- Eliminate all paginated list requests to etcd.
+
+## Proposal
+
+Leveraging the recent rewrite of the watchcache storage layer to use a B-tree
+(https://github.com/kubernetes/kubernetes/pull/126754), we propose to utilize
+B-tree snapshots to serve paginated lists.
+
+Mechanism:
+1. **Snapshot Creation:** When a paginated list request (with a limit parameter
+   set) is received, the API server will create a snapshot using the efficient
+   [Clone()](https://pkg.go.dev/github.com/google/btree#BTree.Clone) method.
+   This clone is a lazy copy, only duplicating necessary nodes, resulting in
+   minimal overhead.
+2. **Snapshot Storage:** The snapshot will be stored in memory, keyed by
+   resourceVersion.
+3. **Serving Subsequent Pages:** When a subsequent request with a continue token
+   arrives, the API server will:
+   - Extract the resourceVersion from the continue token.
+   - Retrieve the corresponding snapshot using the resourceVersion as the key.
+   - Use the snapshot to serve the requested page, or pass the request to etcd is snapshot is not available.
+4. **Snapshot Cleanup:** Snapshots will be subject to a Time-To-Live (TTL)
+   mechanism. We will reuse the existing watch event cleanup logic, which has a
+   75s TTL. This ensures that snapshots don't accumulate indefinitely.
+
+As still some pagination requests will be delegated to etcd, we will monitor the
+success rate by measuring the pagination cache hit vs miss ratio.
+
+Consideration: Should we start respecting the limit parameter?
+
+### Risks and Mitigations
+
+#### Client setting limit while not supporting pagination
+
+#### Memory overhead
+
+No, B-tree only store pointers the actual objects, not the object themselves.
+The objects are already cached to serve watch, so it should only add a small
+overhead for the B-tree structure itself, which is negligible compared to the
+size of the cached objects.
+
+#### Pagination request hitting another apiserver
+
+When connecting directly expectation is that a client will stay connected to the
+same apiserver. What happens when apiserver is behind loadbalancer?
+
+For setups with L4 loadbalancer apiserver can be configured with Goaway, which
+requests client reconnects periodically, however per request probability should
+be configured around 0.1%.
+
+For L7 loadbalancer the default algorithm usually is round-robin. For most LBs
+it should be possible to switch the algorithm to be based on source IP hash.
+Even if that is not possible, stored snapshots will never be used and user will
+not be able to benefit from the feature.
+
+#### Delegating slow pagination to etcd
+
+To avoid breaking users the proposal still allows pagination requests older than
+75s to pass to etcd. This can have a huge performance impact if the resource is
+large. However, this seems still safer than:
+* Increasing the watch cache size 4 times to match etcd.
+* Block requests older than 75s
+
+#### Increased watch contention
+
+As pagination is not served from the main watch cache storage, but from separately
+stored snapshots, pagination request can be served without taking the main lock,
+and just depend on watch cache
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+- Ensure the pagination is well tested
+
+##### Unit tests
+
+- `k8s/apiserver/pkg/storage/cache`: `2024-12-12` - `<test coverage>`
+
+##### Integration tests
+
+<!--
+Integration tests are contained in k8s.io/kubernetes/test/integration.
+Integration tests allow control of the configuration parameters used to start the binaries under test.
+This is different from e2e tests which do not allow configuration of parameters.
+Doing this allows testing non-default options and multiple different and potentially conflicting command line options.
+-->
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, add links to added tests together with links to k8s-triage for those tests:
+https://storage.googleapis.com/k8s-triage/index.html
+-->
+
+- <test>: <link to test coverage>
+
+##### e2e tests
+
+Given we're only modifying kube-apiserver, integration tests are sufficient.
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature implemented behind a feature gate
+- Feature is covered with unit and integration tests
+
+#### Beta
+
+- Feature is enabled by default
+
+#### GA
+
+TODO
+
+### Upgrade / Downgrade Strategy
+
+The feature is purely in-memory so update/downgrade doesn't require any
+specific considerations.
+
+### Version Skew Strategy
+
+Feature touches only kube-apiserver and coordination between individual
+instances is not needed.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [X] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: PaginationFromCache
+  - Components depending on the feature gate: kube-apiserver
+- [ ] Other
+  - Describe the mechanism:
+  - Will enabling / disabling the feature require downtime of the control
+    plane?
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node?
+
+###### Does enabling the feature change any default behavior?
+
+Yes, kube-apiserver paginating LIST requests will no longer require request to etcd.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes, via disabling the feature-gate in kube-apiserver.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+The feature is purely in-memory so it will just work as enabled for the first time.
+
+###### Are there any tests for feature enablement/disablement?
+
+The feature is purely in-memory so feature enablement/disablement will not provide
+additional value on top of feature tests themselves.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+
+###### What specific metrics should inform a rollback?
+
+<!--
+What signals should users be paying attention to when the feature is young
+that might indicate a serious problem?
+-->
+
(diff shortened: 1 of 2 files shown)


dims:
cc @mengqiy  @chaochn47

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,395 @@
...
+size of the cached objects.
+
+#### Pagination request hitting another apiserver
+
+When connecting directly expectation is that a client will stay connected to the
+same apiserver. What happens when apiserver is behind loadbalancer?
+
+For setups with L4 loadbalancer apiserver can be configured with Goaway, which
+requests client reconnects periodically, however per request probability should
+be configured around 0.1%.
```
MadhavJivrajani, line 160: Is there a reason for 0.1% here?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,395 @@
...
+
+For L7 loadbalancer the default algorithm usually is round-robin. For most LBs
```
MadhavJivrajani, line 162: Trying to better understand this, is the worst case here as follows (assume 3 API Servers A, B, C):
1. Client hits API Server A: assuming the rv is cached, snapshot is created on receiving a LIST request with a limit parameter set
2. Client hits API Server B: no snapshot present, we delegate to etcd
3. Client hits API Server C: no snapshot present, we delegate to etcd 

so the performance degenerates to the current situation without cached pagination?

MadhavJivrajani: This ^ is assuming all are on a version that has support for pagination from the cache. If there is one server that is on a minor version which does not have support, my understanding is that that would again be delegated to etcd

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,395 @@
...
+   - Retrieve the corresponding snapshot using the resourceVersion as the key.
+   - Use the snapshot to serve the requested page, or pass the request to etcd is snapshot is not available.
+4. **Snapshot Cleanup:** Snapshots will be subject to a Time-To-Live (TTL)
+   mechanism. We will reuse the existing watch event cleanup logic, which has a
+   75s TTL. This ensures that snapshots don't accumulate indefinitely.
+
+As still some pagination requests will be delegated to etcd, we will monitor the
+success rate by measuring the pagination cache hit vs miss ratio.
+
+Consideration: Should we start respecting the limit parameter?
```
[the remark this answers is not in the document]
serathius: Currently API server doesn't respect limit when serving RV="0". https://github.com/kubernetes/kubernetes/blob/6746df77f2376c6bc1fd0de767d2a94e6bd6cec1/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L806-L818

I think we should consider re-enabling limit for consistency, however we need to better understand consequences. Impact on client/server when we cannot serve pagination from cache, like with L7 LB or pagination taking more than 75s. 

I'm not worried about clients, user setting limit should be already prepared to handle pagination as it is required when not setting RV.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: 0.1% is the recommended to configuration for GOAWAY https://github.com/kubernetes/kubernetes/pull/88567

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Right, there is no regression, assuming that:
* We will delegate continue requests, that we don't have cached responses for, to etcd.
* We will not change how API server doesn't respect limit for RV="0".

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+2. **Snapshot Storage:** The snapshot will be stored in memory, keyed by
+   resourceVersion the watchCache has propagated upto when the snapshot was created. It is also this
+   resourceVersion that is returned as part of the continuation token in the response. 
+3. **Serving Subsequent Pages:** When a subsequent request with a continue token
+   arrives, the API server will:
+  - Extract the resourceVersion from the continue token. 
+  - Since resourceVersions provide a global logical clock sequencing all events in the cluster, a snapshot
+    of the watchCache for this resourceVersion is retrieved using the resourceVersion as the key.
+  - The corresponding snapshot may not be present in the following 2 scenarios at an API Server:
+    - Snapshot has been cleaned up due to the 75s TTL (see below).
```
wojtek-t, line 137: nit: we switched that recently, with 75s still being default, but it now depends on request timeouts

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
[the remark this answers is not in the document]
wojtek-t: I'm actually also worried about clients - I think I've seen cases of people doing that and relying on the baviour of lack of pagination for RV=0.
I'm not saying it's a hard-no, but we need to figure out the story here.

That said, I would put it explicitly out-of-scope for this KEP and add that explicitly as future work.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+- Make pagination behavior predictable and consistent.
+- Reduce load on etcd caused by paginated requests.
+
+### Non-Goals
+
+- Serve `resourceVersion="N"` request from watch cache
+- Support indices when paginating.
+- Eliminate all paginated list requests to etcd.
+
+## Proposal
```
wojtek-t, line 114: I just very briefly skimmed through it and I don't see alternatives.
The main alternative that I have is that:

1) we do btree.Clone() on every watch event (and store it until TTL)  - i.e. independently if there was a request about it or not
2) we can now serve every request (e.g. different kube-apiserver risk goes away) up to TTL from cache from any kube-apiserver (we just use largest RV not greater then the provided one)

Unless the overhead of it is substantial (which I don't know) - that would be my preferred option. But I think we need some experiment to estimate the overhead.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+Consideration: Should we start respecting the limit parameter?
+
+### Risks and Mitigations
+
+#### Client setting limit while not supporting pagination
+
+#### Memory overhead
+
+No, B-tree only store pointers the actual objects, not the object themselves.
```
wojtek-t, line 154: Well - there is some overhead (as you also write below) - you just claim it's small (can we somehow quantify small?)

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+same apiserver. What happens when apiserver is behind loadbalancer?
+
+For setups with L4 loadbalancer apiserver can be configured with Goaway, which
+requests client reconnects periodically, however per request probability should
+be configured around 0.1%.
+
+For L7 loadbalancer the default algorithm usually is round-robin. For most LBs
+it should be possible to switch the algorithm to be based on source IP hash.
+Even if that is not possible, stored snapshots will never be used and user will
+not be able to benefit from the feature.
```
wojtek-t, line 171: I'm not sure we can expect providers to change their LB configuration...

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Right, I have implemented something similar for serving exact revision and was planning to as next step, however you are right that we should have a better understanding of the overhead to know the tradeoffs.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Right

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: My main point here was that we expect only minimal overhead, while users of L7 LB can can opt in via a common configuration option.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: FWIW - it might be worth referencing https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/365-paginated-lists#potential-future-extensions because it's mentioning a lot of that and the corresponding PR has some historical context on it

serathius: I run the scalability tests, the results show 7-18% CPU and 4-6 Memory regression and regression in 99%ile of single object operations. Current scalability tests don't utilize pagination or exact requests at all, so this should be the closes to measuring the overhead. However it's not great at showing us tradeoff, before deciding next step we would need to update scalability tests to be more realistic depiction of workloads using pagination.

serathius: > FWIW - it might be worth referencing https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/365-paginated-lists#potential-future-extensions because it's mentioning a lot of that and the corresponding PR has some historical context on it

Makes sense

serathius: Repeated the test looks like the overhead is small. Small enough that it might be acceptable option.

serathius:
I run the scalability tests to measure overhead of clone. Scalability tests are a good as they don't use pagination nor exact request. I used https://github.com/kubernetes/kubernetes/pull/126855 which clones the storage on each request. The results are good:

Overhead based on profiles collected during scalability tests:
* Additional 7GB of object allocations, which accounts for 0.2% of allocations.
  image
* Additional 300MB of memory used, which accounts for 1.3% of memory used in scalability test.
  image

The overhead is small enough that is within normal variance of memory usage during the test. The are some noticeable increases in request latency however I they are still far from SLO and could be due to high variance in results.

* LIST pods with namespace 99%ile increased from 1s to 1.2s (within variance https://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=APIServer&metricname=LoadResponsiveness_Prometheus&Resource=pods&Scope=namespace&Subresource=&Verb=LIST) 
* DELETE pods 99%ile increased from 170ms to 300ms https://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=APIServer&metricname=LoadResponsiveness_Prometheus&Resource=pods&Scope=resource&Subresource=&Verb=DELETE
* and some other single object operation have seen latency increase. 

If we account for high variance of latency in scalability tests and look at profile differences only, we can estimate the expected overhead of keeping all store snapshots in the watchcache to be below 2% of memory.

wojtek-t:
Are you looking at LoadResponsiveness_Prometheus or LoadResponsiveness_PrometheusSimple for latencies?
If you got 170ms for delete pods in base, it's probably the former, but it also has much higher variance.
What are is the comparison for the later?

https://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=APIServer&metricname=LoadResponsiveness_Prometheus&Resource=pods&Scope=resource&Subresource=&Verb=DELETE
https://perf-dash.k8s.io/#/?jobname=gce-5000Nodes&metriccategoryname=APIServer&metricname=LoadResponsiveness_PrometheusSimple&Resource=pods&Scope=resource&Subresource=&Verb=DELETE

wojtek-t:
I would focus on PrometheusSimple as something that is much more predictible/repeatable.
If those match, and the overhead as you wrote is fairly small (I would be interested in observing how it looks also on small scale), then this solution is much preferable to me (even if in the first step we will only support pagination and nothing else).

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+- Eliminate all paginated list requests to etcd.
+
+## Proposal
+
+Leveraging the recent rewrite of the watchcache storage layer to use a B-tree
+(https://github.com/kubernetes/kubernetes/pull/126754), we propose to utilize
+B-tree snapshots to serve paginated lists.
+
+Mechanism:
+1. **Snapshot Creation:** When a paginated list request (with a limit parameter
```
jpbetz, line 121: In a HA configuration where there are multiple apiservers and client requests are load balanced across those apiservers, is the idea that each apiserver creates the snapshot on the first paginated request it receives, even if the request is for a subsequent page?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+   arrives, the API server will:
+  - Extract the resourceVersion from the continue token. 
+  - Since resourceVersions provide a global logical clock sequencing all events in the cluster, a snapshot
+    of the watchCache for this resourceVersion is retrieved using the resourceVersion as the key.
+  - The corresponding snapshot may not be present in the following 2 scenarios at an API Server:
+    - Snapshot has been cleaned up due to the 75s TTL (see below).
+    - In a HA setup, the request goes to an API Server that did not get the initial LIST request (and hence does not have the in-memory snapshot created as described in (1)).
+4. **Snapshot Cleanup:** Snapshots will be subject to a Time-To-Live (TTL)
+   mechanism. We will reuse the existing watch event cleanup logic, which has a
+   75s TTL. This ensures that snapshots don't accumulate indefinitely.
```
jpbetz, line 141: If a snapshot is missing for a request (either cleaned up, or otherwise), is it recreated or does the request fail?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Goals
+
+- Improve the performance of paginated list requests to be comparable to non-paginated lists.
+- Make pagination behavior predictable and consistent.
+- Reduce load on etcd caused by paginated requests.
+
+### Non-Goals
+
+- Serve `resourceVersion="N"` request from watch cache
```
deads2k, line 110: So I'm clear, this means that a paginated list from RV=N (which is valid I think based on docs: https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-get-and-list )  will not be supported?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+   of the B-tree nodes. Since we do not edit event histories in the watchCache, a snapshot
+   is treated as a read-only entity and its creation is cheap (creating only pointers in memory).
+2. **Snapshot Storage:** The snapshot will be stored in memory, keyed by
+   resourceVersion the watchCache has propagated upto when the snapshot was created. It is also this
+   resourceVersion that is returned as part of the continuation token in the response. 
+3. **Serving Subsequent Pages:** When a subsequent request with a continue token
+   arrives, the API server will:
+  - Extract the resourceVersion from the continue token. 
+  - Since resourceVersions provide a global logical clock sequencing all events in the cluster, a snapshot
+    of the watchCache for this resourceVersion is retrieved using the resourceVersion as the key.
```
deads2k, line 135: if the request with the continue token goes to a different kube-apiserver than the initial list, how will this lookup succeed?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
deads2k: Based on the sentence below, perhaps it falls through to etcd and we get whatever etcd does with it?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+[API call latency SLO](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md)
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+[API call latency SLI](https://github.com/kubernetes/community/blob/master/sig-scalability/slos/api_call_latency.md)
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
```
deads2k, line 324: Given the importance of cache misses for continue tokens, I'd like to have those metrics available in alpha to inform going to beta.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+Overall we expect that cost of serving pagination will go down, however caching
+might increase RAM usage, if the client reads the first page, but never
+paginates. We expect that most controllers will read all pages.
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No
+
+### Troubleshooting
```
deads2k, line 365: How can we check in the field whether the response from the cache exactly matches the response from etcd?

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: Not initially - depending on the path we take here, if we take https://github.com/kubernetes/enhancements/pull/5017#discussion_r1909953609 it will be easily extendable to support it later.

We just wanted to reduce the scope initially, but if you prefer to have it supported from the beginning, we can change it.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: Not really - if I get only a subsequent request for n-th page, I simply forward it to etcd....

... unless we go with the approach I'm proposing instead: https://github.com/kubernetes/enhancements/pull/5017#discussion_r1909953609

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: it's forwarded to etcd... unless we go with the approach that I'm proposing: https://github.com/kubernetes/enhancements/pull/5017#discussion_r1909953609

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: As David wrote.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+##### Prerequisite testing updates
+
+- Ensure the pagination is well tested
+
+##### Unit tests
+
+- `k8s/apiserver/pkg/storage/cache`: `2024-12-12` - `<test coverage>`
+
+##### Integration tests
```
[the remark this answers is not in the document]
serathius: Agree that depending on per-apiserver state would require extensive tests. That's why I think it's better to drop idea of snapshots based on pagination requests. Experiments shown low overhead even if we snapshot every event.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: With changes in design cache misses should not occur.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: If we enable just pagination requests, they could be checked by making an exact request. However the question is why should you care about this at all? Do you want to check if cache was corrupted? For that we should have an automated mechanism.

serathius:
ping @deads2k @wojtek-t

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: ping @deads2k

wojtek-t:
> What small scale you have in mind. For me the scalability tests seem like a worst case scenario. They include large number of small objects with frequent updates. In this situation the overhead from B-tree structure should dominate the size of database.

You're roughly right with the exception that I would like to see the impact for "high-throughput scalability test". Basically I would like to understand if we're not regressing on the throughput that you can achieve now.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,410 @@
...
+   (configurable via request timeout). Additionally, snapshots are purged during
+   cache re-initialization.
+
+[Clone()]: https://pkg.go.dev/github.com/google/btree#BTree.Clone
+
+### Cache Inconsistency detection
+
+For the first iteration of the mechanism we will calculate the hash based on LIST response
+at some specific RV of each resource and expose it as a metric. Hash will be calculated for
+both cache and etcd, allowing users to setup an alert to detect if they don't match.
```
deads2k, line 169: name the metric please.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,410 @@
...
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature implemented behind a feature gate
+- Feature is covered with unit and integration tests
+
+
+#### Beta
```
deads2k, line 238: please include that the mismatch metric has not failed in alpha testing.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
deads2k, line 169: If this ever fails, I'd like the kube-apiserver to automatically revert to sending these requests to etcd and not handling them in the watch cache.  That leaves us a marker of "something broke" and avoids serving incorrect data.  I'm open a second gate (alpha and always off) that allows us to skip this behavior for diagnostic purposes, but we should not knowing continue to use a broken feature in production clusters.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,410 @@
...
+
+##### Prerequisite testing updates
+
+- Ensure the pagination is well tested
+
+##### Unit tests
+
+- `k8s/apiserver/pkg/storage/cache`: `2024-12-12` - `<test coverage>`
+
+##### Integration tests
```
deads2k, line 207: I'd like an integration test covering the "fallback to etcd on mismatch" case to be sure it works well.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
```diff
@@ -0,0 +1,410 @@
...
+
+The feature is purely in-memory so feature enablement/disablement will not provide
+additional value on top of feature tests themselves.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+
+###### What specific metrics should inform a rollback?
```
deads2k, line 295: add the cache mismatch metric here.

deads2k (changes requested):
Don't forget to add the PRR metadata.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Done

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Done

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: I'm worried about introducing an untested automatic switch mechanism that would cause severe performance impact. For Alpha I was planning to make the inconsistency detection default (assuming overhead is acceptable) to qualify it.

As for what action we should take if "something broke" I was thinking more about purging cache. It's more risky as it would make resource unavailable, that's why we should first qualify the inconsistency detection.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: Done.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
serathius: I was thinking more about purging the cache, but agree we need a test for it. Added the proposal for a test, we can discuss what action we should  in above comment.

keps/sig-api-machinery/4988-serve-pagination-from-cache/README.md
wojtek-t: +1 to Marek
We definitely need concistency-checker and I would like it to be enabled in all our CI pretty much from the beginning.
As for what to do in prod - given how significant the different between serving from cache vs etcd is, automatic fallback to etcd may have very bad consequences. Purging the cache and recreating it from scratch sounds much less risky to me.

deads2k: If the mechanism is producing an inconsistent cache, I don't agree that simply restarting that inconsistent cache is a good way to resolve the problem.

serathius: Discussed offline, we are in agreement that fallback to etcd would be better here, at least until GA.

PR #5050, opened 2025-01-18
KEP-4247: correct the milestone at kep.yaml

sanposhiho:
- One-line PR description: Change the beta milestone to v1.32

- Issue link: https://github.com/kubernetes/enhancements/issues/4247

- Other comments: So, the context is that we once graduate it as beta at v1.28, but we soon reverted it because we found a memory leak bug. v1.32 fixes the bug, and we tried again to graduate it to beta. So, I believe we should note the graduation of beta was v1.32.


Huang-Wei (commented):
/lgtm
/approve

PR #5054, opened 2025-01-20
KEP-5053: Fallback for HPA on failure to retrieve metrics

be0x74a:
Add KEP-5053: Fallback for HPA on failure to retrieve metrics

Issue link: https://github.com/kubernetes/enhancements/issues/5053

### DIFF
--- a/keps/sig-autoscaling/5053-hpa-fallback/README.md
+++ b/keps/sig-autoscaling/5053-hpa-fallback/README.md
@@ -1,0 +1,849 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+To get started with this template:
+
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-5053: Fallback for HPA on failure to retrieve metrics
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The [Horizontal Pod Autoscaler (HPA)][] relies on the controller manager to 
+fetch metrics from either the resource metrics API (for per-pod resource metrics) 
+or the custom metrics API (for other types of metrics). When these APIs experience 
+downtime, the HPA becomes unable to make scaling decisions, potentially leaving 
+workloads unmanaged.
+
+This proposal introduces a new configuration parameter for the HPA, enabling 
+users to define behavior in the event of metric retrieval failures. For example, 
+users can opt to scale the target resource to the maximum number of replicas 
+specified in the HPA, ensuring safer operation during metrics unavailability.
+
+[Horizontal Pod Autoscaler (HPA)]: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
+
+## Motivation
+
+The Horizontal Pod Autoscaler (HPA) is a critical component for scaling Kubernetes 
+workloads based on resource utilization or custom metrics. However, the current 
+implementation depends entirely on the availability of the resource metrics API 
+or custom metrics API to make scaling decisions. If these APIs experience 
+downtime or degradation, the HPA cannot take any scaling actions, leaving 
+workloads potentially overprovisioned, underprovisioned, or entirely unmanaged.
+
+In contrast, other autoscalers like [KEDA][] already provide mechanisms to define 
+fallback strategies in the event of metric retrieval failures. These strategies 
+mitigate the impact of API unavailability, enabling the autoscaler to maintain 
+a functional scaling strategy even when metrics are temporarily inaccessible.
+
+By allowing users to configure fallback behavior in HPA, this proposal aims to 
+reduce the criticality of the metrics APIs and improve the overall robustness 
+of the autoscaling system. This change allows users to define safe scaling 
+actions, both as scaling to a predefined maximum or holding the current scale 
+(current behaviour), ensuring workloads remain operational and better aligned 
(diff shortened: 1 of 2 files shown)


keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,849 @@
...
+### Non-Goals
+
+- N/A
+
+## Proposal
+
+Heavily inspired by [KEDA][] propose to add a new field to the existing [`HorizontalPodAutoscalerSpec`][] object:
+
+- `fallback`: an optional new object containing the following fields:
+  - `failureThreshold`: (integer) the number of failures fetching metrics to trigger the fallback behaviour. Must be greater than 0 and is option with a default of 3.
```
adrianmoisey, line 189: What does this mean:

> Must be greater than 0 and is option with a default of 3.

Is that saying that it's an optional field and that it defaults to "3" ?

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,849 @@
...
+- Address feedback on usage/changed behavior, provided on GitHub issues
+- Deprecate the flag
+-->
+
+#### Alpha
+
+- Feature implemented behind a `HPAFallback` feature flag
+- Initial e2e tests completed and enabled
+
+### Upgrade / Downgrade Strategy
```
adrianmoisey, line 471: This is possibly worth filling in.
ie: If someone enables the alpha feature, and does a rollback, what happens?

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Yes exactly, clarified the explanation. Thanks

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Filled it with a brief overview of what would happen on feature flag enable and disable

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+}
+```
+
+The `HorizontalPodAutoscaler` API is updated to add a new `fallback` field to the `HorizontalPodAutoscalerStatus` object:
+
+```golang
+type HorizontalPodAutoscalerStatus struct {
+    // metricRetrievalFailureCount tracks the number of consecutive failures in retrieving metrics. 
+    //+optional
+    MetricRetrievalFailureCount int32
```
raywainman, line 245: Should we maybe be explicit in the name that this is consecutive failures?

`ConsecutiveMetricRetrievalFailureCount`?

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+
+The `HorizontalPodAutoscaler` API is updated to add a new `fallback` field to the `HorizontalPodAutoscalerSpec` object:
+
+```golang
+type HorizontalPodAutoscalerSpec struct {
+    // fallback specifies the number of replicas to scale the object to during a 
+    // fallback state and defines the threshold for errors required to enter the 
+    // fallback state.
+    //+optional
+    Fallback *HPAFallback
```
raywainman, line 228: Any thoughts on adding this to the Behavior field?

(I don't feel strongly, just curious what you think)

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+            metricDesiredReplicas = hpa.Spec.Fallback.Replicas
+            a.eventRecorder.Event(hpa, v1.EventTypeWarning, "FallbackThresholdReached", err.Error())
+        } else {
+            inFallback = false
+        }
+    } else {
+        inFallback = false
+    }
+    
+    if !inFallback {
```
raywainman, line 301: I think we'll probably want to set a status when the HPA is in fallback mode?

raywainman (commented):
Thanks a lot for putting this together :) I think this will be tremendously useful for users.

Would love for the community to provide more feedback on the API, I think we are definitely on the right track here!

keps/sig-autoscaling/5053-hpa-fallback/README.md
adrianmoisey: May be an event too?
```
Events:
  Type     Reason                        Age                From                       Message
  ----     ------                        ----               ----                       -------
  Normal   SuccessfulRescale             2m19s              horizontal-pod-autoscaler  New size: 1; reason: All metrics below target
```

That's an event for a regular scale, so I think a similar message but with a different reason would be useful

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+    //+optional
+    MetricRetrievalFailureCount int32
+    
+    // Existing fields.
+    ObservedGeneration *int64
+    LastScaleTime *metav1.Time
+    CurrentReplicas int32
+    DesiredReplicas int32
+    CurrentMetrics []MetricStatus
+    Conditions []HorizontalPodAutoscalerCondition
```
adrianmoisey, line 253: I think a new condition may be needed too?
This is what we currently have:
```
Conditions:
  Type            Status  Reason            Message
  ----            ------  ------            -------
  AbleToScale     True    ReadyForNewScale  recommended size matches current size
  ScalingActive   True    ValidMetricFound  the HPA was able to successfully calculate a replica count from cpu resource utilization (percentage of request)
  ScalingLimited  True    TooFewReplicas    the desired replica count is less than the minimum replica count
```

I imagine a "Fallback" Type may be useful

keps/sig-autoscaling/5053-hpa-fallback/README.md
adrianmoisey: > I think we'll probably want to set a status when the HPA is in fallback mode?

I commented higher up about a condition (see https://github.com/kubernetes/enhancements/pull/5054/files#r1924247581).

I imagine that covers the status update you're talking about?

raywainman: Right - yes a condition flagging that the HPA is in this fallback state. Good callout.

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+### Non-Goals
+
+- N/A
+
+## Proposal
+
+Heavily inspired by [KEDA][] propose to add a new field to the existing [`HorizontalPodAutoscalerSpec`][] object:
+
+- `fallback`: an optional new object containing the following fields:
+  - `failureThreshold`: (integer) the number of failures fetching metrics to trigger the fallback behaviour. Must be a value greater than 0. This field is optional and defaults to 3 if not specified.
```
jm-franc, line 186: `failureThreshold` specifies a maximum number of retries. Isn't a duration what users are looking for (i.e. trigger fallback if scraping metrics fails for x seconds)?

(Note that for KEDA it's equivalent, as KEDA specifies the scraping frequency.)

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,856 @@
...
+  - `replicas`: (integer) the number of replicas to scale to in case of fallback. Must be greater than 0 and it's mandatory.
+
+To allow for tracking of failures to fetch metrics a new field should be added to the existing [`HorizontalPodAutoscalerStatus`][] object:
+- `metricRetrievalFailureCount`: (integer) tracks the number of consecutive failures in retrieving metrics.
```
jm-franc, line 190: It's a single integer, not an integer per metric? If two metrics start failing, don't we want to keep a separate counter for each?

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Haven't though of it but I can see how it would make sense to define the fallback **behavior** on the Behavior filed

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Altough maybe a big name, I agree it's better to be explicit they're consecutive failures

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: > That's an event for a regular scale, so I think a similar message but with a different reason would be useful

I added an event [here](https://github.com/kubernetes/enhancements/pull/5054/files#diff-b1b3ee22fed3ff485c89cf69181e8800d3694a97a4defd57cb7df44a94d7e2ebR293) but I guess it's better to have a specific message for fallback and not the error responsible for it.

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Hmmm, that's a good question. As a user, I would naturally go for the retry count but as I don't control scrape interval, that doesn't translate into much.

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: The way I was thinking was to integrate on the exit code of `computeReplicasForMetrics` which handles partial metric failure. That way we could track a single integer

adrianmoisey: I wonder if it's worth writing that down in the KEP? Specifically what happens when:
1. There is 1 metric configured, and it fails
2. There are multiple metrics configured, and a single metric fails
3. There are multiple metrics configured, and they take turns to fail (unlikely, but possible?)

keps/sig-autoscaling/5053-hpa-fallback/README.md
adrianmoisey: I semi-agree with this.
Behaviour feels right, but the description of Behaviour feels wrong: https://github.com/kubernetes/kubernetes/blob/0798325ba13643358aa3ebb7c6ddc3006ac26a7c/pkg/apis/autoscaling/types.go#L111-L113
> // HorizontalPodAutoscalerBehavior configures a scaling behavior for Up and Down direction
// (scaleUp and scaleDown fields respectively).


It feels to me like "Scaling behaviour" and "Fallback behaviour" are two different things. 
May be if the comment I linked to was reworded, then "Behaviour" is the right place?

keps/sig-autoscaling/5053-hpa-fallback/README.md
jm-franc: If `failureThreshold` is set to 10, and at a given moment metric `foo` failed 9 times while metric `bar` failed only once, I can't see how we could summarize the current state with one integer.

Do you mind elaborating on how you think this could be handled? I guess you'd take the max (9 in my example)? Is this desirable?

adrianmoisey: Taking a look at `computeReplicasForMetrics`, I assume the counter can only be incremented once per cycle, and only if at least 1 metric has failed. 

In the case you describe, if the first 9 cycles had a failure on metric `foo` and the 10th cycle had a failure on metric `bar`, then there would be 10 consecutive failures. 

But if the first cycle had both metrics fail, then the next 8 cycles had metric `foo` and the final 10th cycle had no metrics fail, then the count would max at 9 consecutive failures. 

This is just my understanding of what @be0x74a is saying though.
It's worth clarifying in the KEP though.

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Updated

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Added a section for the comment update for the behavior field and moved the fallback into it

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Added the `FallbackActive` condition

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: Done

keps/sig-autoscaling/5053-hpa-fallback/README.md
be0x74a: > Taking a look at computeReplicasForMetrics, I assume the counter can only be incremented once per cycle, and only if at least 1 metric has failed.
Yeah, that's exactly what I was thinking. I can add it to the KEP as the current state, but I don't mind discussing whether this is the correct approach. 
Btw @adrianmoisey do you have a section in mind where I can put this? Maybe in the Design Details section?

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,893 @@
...
+The `HorizontalPodAutoscaler` API is updated to introduce a new FallbackActive condition to the `HorizontalPodAutoscalerConditionType`:
+
+```golang
+const (
+    // FallbackActive indicates that the HPA has entered the fallback state due to repeated
+    // metric retrieval failures and is applying the configured fallback behavior.
+    FallbackActive HorizontalPodAutoscalerConditionType = "FallbackActive"
+
+	// Existing conditions
+	ScalingActive HorizontalPodAutoscalerConditionType = "ScalingActive"
```
adrianmoisey, line 284: Spacing issue: 

```suggestion
    // Existing conditions
    ScalingActive HorizontalPodAutoscalerConditionType = "ScalingActive"
```

keps/sig-autoscaling/5053-hpa-fallback/README.md
adrianmoisey: The Design Details section sounds good. I think a section describing what a failure is should suffice?

jm-franc: My advise may be idiosyncratic, but I'd like to know what others think:

1. KEDA implements the external metric API only, which I think is where a fallback makes  sense, as APIs running out of cluster are not under its control. How do we justify a special mechanism for (resource) metrics API failures, but we don't do the same for other server API failures?
2. KEDA fallback is per-metric, and allows to set a default value for this metric (e.g. 0 to "ignore"). This is less risky than the proposal here, as I can't really see what users could set `HPAFallback.replicas` to other than `maxReplicas`.
3. Because the fallback is not per-metric, users can't see what metric failed (iiuc). Is this acceptable?

omerap12: My two cents on this: I think we should use a per-metric approach. It's much safer and gives more control to the users so this is definitely something we should reconsider in the design. It can also help with visibility - when we fallback, users will be able to see why we fallback (e.g. event/status on the HPA etc).
So agree with @jm-franc

be0x74a: Hey @jm-franc & @omerap12, in response to your comments:
1. I get the point for consistency, and the only justification I can manage for this special mechanism is the higher fragility of metric sources + its potential impact when failing
2 & 3. Time granted me a new perspective on this discussion. Although I still appreciate the simplicity of the initial design, I now believe that allowing for better configurability of the fallbacks is preferable.

I'll try to get a bit of time this week to work on adding this to the proposal

be0x74a:
> Since KEDA provides a variety of fallback behavior options (https://keda.sh/docs/2.17/reference/scaledobject-spec/#fallbackbehavior), is there any reason not to support all of them?

Honestly, I missed these options altogether. But looking at them now, they look like a great addition

omerap12:
> > Since KEDA provides a variety of fallback behavior options (https://keda.sh/docs/2.17/reference/scaledobject-spec/#fallbackbehavior), is there any reason not to support all of them?
> 
> Honestly, I missed these options altogether. But looking at them now, they look like a great addition

I guess we can start with a basic fallback and add all other functionalities later

keps/sig-autoscaling/5053-hpa-fallback/README.md
```diff
@@ -0,0 +1,893 @@
...
+    var inFallback bool
+    
+    if hpa.Spec.Fallback != nil {
+        var failureThreshold int32
+        
+        if hpa.Spec.Fallback.FailureThreshold != nil {
+            failureThreshold = *hpa.Spec.Fallback.FailureThreshold
+        } else {
+            // Default value
+            failureThreshold = 3
```
omerap12, line 320: IMHO, we should align with KEDA on this, since the feature was originally designed by them and we want to avoid confusing users with differing behavior. So, this field should be mandatory if the fallback section is included.
Ref: https://keda.sh/docs/2.17/reference/scaledobject-spec/#fallback

keps/sig-autoscaling/5053-hpa-fallback/kep.yaml
```diff
@@ -0,0 +1,38 @@
...
+kep-number: 5053
+authors:
+  - "@be0x74a"
+owning-sig: sig-autoscaling
+status: provisional
+creation-date: 2025-01-20
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
omerap12, line 11: You can add @adrianmoisey and me as reviewers (we're the current HPA reviewers), and add one of the SIG leads as an approver - @gjtempleton, if he’s okay with that.

omerap12 (commented):
I added some comments, and you forgot the production-ready review :)

omerap12:
Hey @be0x74a,
Are you still planning to work on this? We’re hoping to push it forward, but if you’re not available, @adrianmoisey and I can take over.

be0x74a:
> Hey @be0x74a, Are you still planning to work on this? We’re hoping to push it forward, but if you’re not available, @adrianmoisey and I can take over.

Hey @omerap12, unfortunately, I don't think I'll be able to put a lot of work into it in the next few weeks. I may be able to return to it at a later date, but it wouldn't be fair to hold the project in the meantime.

adrianmoisey:
I don't believe there is time to get it in for Kubernetes 1.35, so you do have some time if you want to aim for 1.36

omerap12:
Since we’ve got some users who need this, @adrianmoisey and I will take it from here :)

omerap12:
Opened: https://github.com/kubernetes/enhancements/issues/5679
So we can close this, thanks for pushing it forward @be0x74a!

PR #5060, opened 2025-01-21
Fix typo in HPA scale velocity document

phihos:
A YAML example has 600 as value for the `periodSeconds` attribute, but in the text below it is referenced as a minute. The value should therefore be 60 to avoid confusion for the reader.


gjtempleton:
Thanks
/approve

PR #5063, opened 2025-01-21
KEP-4205: reporting experimental results and concerns on phase 2

tiraboschi:
- One-line PR description: PSI is currently not able to tell us if the pressure is caused by the contention of a scarce resource or it's due to CPU throttling according to the limits and this is reported also at node level.

- Issue link: https://github.com/kubernetes/enhancements/issues/5062

- Other comments:


fabiand:
cc @mrunalp @kannon92

PR #5119, opened 2025-01-31
Propose KEP-5116: Streaming response encoding

serathius:
https://github.com/kubernetes/enhancements/issues/5116

/assign @deads2k @liggitt


deads2k:
@benluddy can you ptal?

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,353 @@
...
+decision to enable compression until we have observed enough bytes to hit the threshold
+or we received whole response and we can write it without compressing.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
```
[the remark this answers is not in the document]
serathius: Structure of data encoded in List and WatchList is totally different. List encodes element within an array object, while WatchList is compatible to Watch, so it encodes elements within a new type of watch events. Don't think we can do a parity check.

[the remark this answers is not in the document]
benluddy: Sorry, I was unclear. I mentioned the list/watchlist detector only as an example of a similar test-only mechanism. I'm suggesting serializing lists using _both_ the optimized path and the standard path and checking those for parity (to be enabled for test jobs).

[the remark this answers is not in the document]
serathius: In unit tests we cover byte to byte parity with standard library encoder. Is this what you have in mind?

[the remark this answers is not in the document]
liggitt: I think @benluddy is suggesting that there is *way* more variety of outputs in a full e2e run than just the unit test cases constructed for the feature.

Given this only impacts JSON list requests, which no built-in typed clients make use of, I'm actually skeptical that "run both and compare" will get used that much in e2e runs (~mostly just on custom resources, I think).

I think unit tests can actually sufficiently cover the byte-for-byte comparison against standard serialization, since this only kicks in when we have exactly 4 fields to deal with, and can fully test every possible top-level value, beyond which we delegate to standard JSON serialization:
* ObjectMeta (we always fully delegate to standard ObjectMeta JSON serialization)
* TypeMeta.APIVersion (possible values: `""`, non-empty string)
* TypeMeta.Kind (possible values: `""`, non-empty string)
* Items []... (possible values: `null`, `[]`, non-empty list delegating to individual item JSON serialization)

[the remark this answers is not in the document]
benluddy: The KEP also proposes to implement streaming list encoding for Protobuf, but I also hadn't considered that JSON streaming wouldn't be exercised very well by something like this. I think I'm convinced that it's not worth doing. Constraining the supported types to ultra-conventional list types is a great feature of this proposal.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+
+## Summary
+
+This KEP proposes implementing streaming encoding for LIST responses served by the Kubernetes API server.
```
sftim, line 57:
```suggestion
This KEP proposes implementing streaming encoding for collections (responses to **list**) that are served by the Kubernetes API server.
```

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+## Implementation History
+
+## Drawbacks
+
+Maintaining around 500 lines of custom encoder code.
+
+## Alternatives
+
```
sftim, line 351: I guess we ought to mention https://github.com/kubernetes/enhancements/issues/3157 (watch lists) - but we should rule out relying on that because the **list** verb is generally available. We can't rely on clients using our preferred API, much less would-be attackers.

sftim:
BTW if we want to ship JSON and ProtoBuf as beta for v1.33, I don't see anything stopping us doing that and still adding streaming YAML collections before we close the KEP as Done.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+checking just first write size to determine if the response is large enough for compression.
+This is a bad assumption about internal encoder implementation details and should be fixed regardless.
+
+To ensure gzip compression works well with streaming,
+we will preempt all encoder changes by fixing the gzip compression.
+First will add unit tests that will prevent subsequent changes from impacting results,
+especially around the compression threshold.
+Then, we will rewrite the gzip compression to buffer the response and delay the
+decision to enable compression until we have observed enough bytes to hit the threshold
+or we received whole response and we can write it without compressing.
```
liggitt, line 155: just an implementation note, by adding an intermediate buffer to the gzip writer wrapper, be careful not to reintroduce the issue where we don't flush headers until a non-zero number of bytes are written

https://github.com/kubernetes/kubernetes/pull/31446/files#diff-67dd9e8f3ee257072765326cb4f242852554a2c0753563fa51e292c0a63a7b94R2488

maybe the code path that hits deferredResponseWriter isn't vulnerable to that issue if it already has the single object it is serializing, but watch out for buffering a generic gzip writer wrapper if we're not sure whether or not we'll have immediate output

liggitt:
a couple suggestions but lgtm overall

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+* Preserving the distinction between integers and floating-point numbers.
+* Handling structs with duplicate field names (json tag names) without producing duplicate keys in the encoded output ([https://go.dev/issue/17913](https://go.dev/issue/17913)).
+* Encoding Go strings containing invalid UTF-8 sequences without error.
+* Preserving the distinction between absent, present-but-null, and present-and-empty states for slices and maps.
+* Handling structs implementing `MarshallJSON` method, especially LIST types.
+* Handling raw bytes.
+
+Fuzz tests will cover the custom streaming encoders for the types with overwritten encoders:
+* testingapigroup.CarpList as surigate for builtin types
```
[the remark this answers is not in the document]
sftim: or see https://github.com/kubernetes/enhancements/pull/5119#discussion_r1942994412

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Goals
+
+* Implement streaming encoders for JSON and Protocol Buffer for LIST responses.
+* Significantly reduce and make more predictable the API server's memory consumption when serving large LIST responses.
+
+### Non-Goals
+
+* Implementing streaming decoders in clients. This KEP focuses on protecting the API server's memory usage. Clients can utilize existing mechanisms like pagination or WatchList to manage large datasets.
+* Implementing streaming encoders for all content types (e.g., "as=Table"). This KEP focuses on the most commonly used and resource-intensive content types to address the most impactful cases first.
```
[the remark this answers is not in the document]
serathius: Sure,

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+##### Unit tests
```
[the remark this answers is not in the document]
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,354 @@
...
+* Encoding Go strings containing invalid UTF-8 sequences without error.
+* Preserving the distinction between absent, present-but-null, and present-and-empty states for slices and maps.
+* Handling structs implementing `MarshallJSON` method, especially LIST types.
+* Handling raw bytes.
+
+Fuzz tests will cover the custom streaming encoders for the types with overwritten encoders:
+* testingapigroup.CarpList as surigate for builtin types
+* UnstructuredList 
+
+The skipped tests are primarily related to decoding or CBOR-specific features, which are not relevant to the streaming encoding of JSON and Proto addressed by this KEP.
```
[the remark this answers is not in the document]
serathius: Yes, we don't plan to remove YAML tests.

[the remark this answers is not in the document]
sftim: (resolved)

keps/sig-api-machinery/5116-streaming-response-encoding/kep.yaml
```diff
@@ -0,0 +1,25 @@
+title: Streaming JSON Encoding for LIST Responses
+kep-number: 5116
+authors:
+  - serathius
+owning-sig: sig-api-machinery
+participating-sigs:
+  - sig-aaa
+  - sig-bbb
```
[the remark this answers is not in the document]
sftim: Are there any other SIGs that should participate? Clients won't know the difference (and we'll test that).

fuweid (approved):
LGTM(non-binding)!

liggitt (commented):
this lgtm, will let @deads2k (for apimachinery) and @jpbetz (for PRR) have approval

sftim:
/retitle KEP-5116: Add KEP (Streaming response encoding)

wojtek-t:
> this lgtm, will let @deads2k (for apimachinery) and @jpbetz (for PRR) have approval

I'm fine with this as well.

@deads2k @jpbetz - friendly ping for review

benluddy:
LGTM. Strictly constraining the supported types keeps the special-case implementation simple, and client behavior remains unchanged.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+###### Does enabling the feature change any default behavior?
+
+No, we provide byte to byte compatibility.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes, without problem.
+
+###### What happens if we reenable the feature if it was previously rolled back?
```
deads2k, line 246: missing answer

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+N/A
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+N/A
+
+###### What specific metrics should inform a rollback?
+
+N/A
```
deads2k, line 262: this doesn't seem correct.  Latency and failed list request metrics should be checked.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+N/A
```
deads2k, line 266: Explain why.  Nothing is stored.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No
+
+### Monitoring Requirements
+
+
+###### How can an operator determine if the feature is in use by workloads?
+
+N/A
```
deads2k, line 277: inaccurate, if the featuregate is on and lists are used, you know it's used.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+###### How can someone using this feature know that it is working for their instance?
+
+N/A
```
deads2k, line 281: explain why. It's not a user-face feature. Users get byte-for-byte equivalence

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+N/A
```
deads2k, line 285: the existing lists shouldn't get worse.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+N/A
```
deads2k, line 289: again list latency and error rate metrics here.

keps/prod-readiness/sig-api-machinery/5116.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 5116
+beta:
+  approver: "@jpbetz"
```
deads2k, line 3: update for deads2k

deads2k:
needs updates for PRR enable/disable and monitoring.  The list of tests looks good.  lgtm otherwise.

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/sig-api-machinery/5116-streaming-response-encoding/README.md
serathius: Done

keps/prod-readiness/sig-api-machinery/5116.yaml
serathius: Done

serathius:
> needs updates for PRR enable/disable and monitoring. The list of tests looks good. lgtm otherwise.

Fixed, thanks for review!

PR #5201, opened 2025-03-12
KEP-4603: Correct conflict resolution table

lauralorenz:
- One-line PR description: Fix conflict resolution table, during implementation realized a line was incorrect and the prose had  conflated initial and max values in a case

- Issue link: https://github.com/kubernetes/enhancements/issues/4603

- Other comments: N/A

### DIFF
--- a/keps/sig-node/4603-tune-crashloopbackoff/README.md
+++ b/keps/sig-node/4603-tune-crashloopbackoff/README.md
@@ -1341,9 +1341,9 @@ kubelet will crash/be unable to start, like it does with other invalid kubelet
 configuration today.
 
 If on a given node at a given time, the per-node configured maximum backoff is
-higher than the current initial value, but within validation limits as it is
+higher than the default maximum backoff, but within validation limits as it is
 lower than 300s, it will be honored. In other words, operator-invoked
-configuration will have precedence over the default, even if it is slower, as
+configuration for the maximum backoff will have precedence over the default, even if it is slower, as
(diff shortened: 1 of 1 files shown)


keps/sig-node/4603-tune-crashloopbackoff/README.md
```diff
@@ -1341,9 +1341,9 @@ kubelet will crash/be unable to start, like it does with other invalid kubelet
 configuration today.
 
 If on a given node at a given time, the per-node configured maximum backoff is
-higher than the current initial value, but within validation limits as it is
+higher than the default maximum backoff, but within validation limits as it is
 lower than 300s, it will be honored. In other words, operator-invoked
-configuration will have precedence over the default, even if it is slower, as
+configuration for the maximum backoff will have precedence over the default, even if it is slower, as
 long as it is valid.
```
tallclair, line 1347: nit: is this paragraph necessary? Isn't this just stating what a default means? Or am I missing something?

lauralorenz: It is, but then so is the first paragraph (ending in "In other words, operator-invoked configuration will have
precedence over the default if it is faster."). I think both paragraphs were to make that explicit, and separately in the event there was appetite for different conflict resolution depending on if a node's default was slower or faster than what the user's crashloopbackoff.maxSeconds requested (though in the end the behavior ended up being the same).

PR #5256, opened 2025-04-22
SIG Scheduling: Leadership Change

dom4ha:
https://github.com/kubernetes/community/issues/8428

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -94,9 +94,9 @@ aliases:
     - shyamjvs
     - wojtek-t
   sig-scheduling-leads:
-    - Huang-Wei
-    - ahg-g
-    - alculquicondor
+    - dom4ha
+    - macsko
+    - sanposhiho
   sig-security-leads:
     - IanColdwater
     - tabbysable


dom4ha (commented):
cc @sanposhiho, @macsko, @alculquicondor, @Huang-Wei, @ahg-g

sanposhiho (approved):
/lgtm

kikisdeliveryservice (commented):
@dom4ha based on the CL: https://github.com/kubernetes/community/issues/8428#issue-3003154618

You're onboarding 3 leads (and off boarding 3). However you only have 2 being added. Is this correct?

dom4ha:
> Dominik Marciński based on the CL: [kubernetes/community#8428 (comment)](https://github.com/kubernetes/community/issues/8428#issue-3003154618)
> 
> You're onboarding 3 leads (and off boarding 3). However you only have 2 being added (macsko is missing). Is this correct?

The number of TLs decrease from 3 to 2. IIUC only TLs can approve enhancements, is my understanding correct?

kikisdeliveryservice:
If you want only tech leads to approve and not leads then you can do what SIG node did here: https://github.com/kubernetes/enhancements/commit/707641fdd25bb73ce39a4849ff91c41f315365a5

However, not all SIGs do it that way and unless I'm unaware, it's not a requirement that only TLs can approve enhancements.

cc: @johnbelamaric @dims

sanposhiho:
We can put @macsko as well then

dom4ha:
> We can put Maciej Skoczeń as well then

Yes, that would be much better. I faced some issues with my laptop, so could not update the PR earlier, but it should be fixed now :)

dom4ha:
> Dominik Marciński can you squash your second commit, there seems to be an issue with the commit message and this is not mergable. Thanks!

Done.

PR #5272, opened 2025-05-05
githooks: run all checks in pre-commit hook

marquiz:
Previous usage of exec caused only the first script to be run.


justaugustus (approved):
Thanks @marquiz!
/lgtm
/approve

PR #5318, opened 2025-05-19
Misc SIG Network KEP updates (especially "implementable" → "implemented")

danwinship:
Started with trying to update all implemented KEPs to correctly indicate that they were implemented, then had to update milestones as well to make `verify-kep-metadata.sh` happy, then did a bunch of other drive-by kep.yaml fixes, then did some README.md readability fixes too...

/assign @aojea


aojea:
/lgtm

great

PR #5324, opened 2025-05-21
Move KEP-5116 to GA

serathius:
Ref https://github.com/kubernetes/enhancements/issues/5116

Proposing to move the KEP-5116 to GA. 

There were no issues discovered in the implementation, so the only pending work is preventing regressions.
Added some information about regression we have already caught by just eyeballing the graph.

Don't think there is a need to block the GA on implementation of other serialization formats. While they are supported by apiserver, they are not used in Kubernetes core projects (`kubectl get pod -o yaml` will request JSON and generate yaml on client side) and should not be recommended. 

/assign @deads2k @liggitt


deads2k:
lgtm.

/approve

PR #5353, opened 2025-05-29
Target GA for recover from Volume Expansion failure feature

gnufied:
https://github.com/kubernetes/enhancements/issues/1790

### DIFF
--- a/keps/sig-storage/1790-recover-resize-failure/README.md
+++ b/keps/sig-storage/1790-recover-resize-failure/README.md
@@ -28,7 +28,7 @@
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
     - [Beta](#beta)
-  - [GA](#ga)
+    - [GA](#ga)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
   - [Monitoring](#monitoring)
@@ -343,14 +343,15 @@ There are already e2e tests running that verify correctness of this feature - ht
   -  There are already e2e tests running that verify correctness of this feature -
   https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-cos-alpha-features
   
-### GA  
+#### GA  
 
 - The feature has been extensively tested in CI and deployed with drivers in real world. For example - https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L166
 - The test grid already has tests for the feature.
 
 ### Upgrade / Downgrade Strategy
 
-Not Applicable
+- For the case of older kubelet running with newer resizer, the kubelet must handle the newer fields introduced by this feature even if feature-gate is not enabled. Having said that, if kubelet is older and has no notion of this feature at all and `api-server` and `external-resizer` is newer, it is possible that kubelet will not properly update `allocatedResourceStatus` after expansion is finished on the node. This is a known limitation and if user wants to keep running older version of kubelet (older than v1.31 as of this writing and hence unsupported), then they MUST not update external-resizer.
+- In general `external-resizer` should only be upgraded(and `RecoverVolumeExpansionFailure` enabled), after `kubelet`, `api-server` and `kube-controller-manager` already has this feature enabled.
 
 ### Version Skew Strategy
 
--- a/keps/sig-storage/1790-recover-resize-failure/README.md
+++ b/keps/sig-storage/1790-recover-resize-failure/README.md
@@ -449,13 +449,17 @@ if this feature-gate is enabled.
 
   - [X] Metrics
     - controller expansion operation duration:
-        - Metric name: storage_operation_duration_seconds{operation_name=expand_volume}
+        - Metric name: csi_sidecar_operations_seconds{method_name="/csi.v1.Controller/ControllerExpandVolume}
         - [Optional] Aggregation method: percentile
-        - Components exposing the metric: kube-controller-manager
+        - Components exposing the metric: external-resizer
     - controller expansion operation errors:
-        - Metric name: storage_operation_errors_total{operation_name=expand_volume}
+        - Metric name: csi_sidecar_operations_seconds{method_name="/csi.v1.Controller/ControllerExpandVolume, grpc_status_code!="OK"}
         - [Optional] Aggregation method: cumulative counter
-        - Components exposing the metric: kube-controller-manager
+        - Components exposing the metric: external-resizer
+    - CSI node expansion operation durations:
+        - Metric name: csi_operations_seconds{method_name="/csi.v1.Controller/NodeExpandVolume}
+        - [Optional] Aggregation method: cumulative counter
+        - Components exposing the metric: kubelet
     - node expansion operation duration:
         - Metric name: storage_operation_duration_seconds{operation_name=volume_fs_resize}
         - [Optional] Aggregation method: percentile
@@ -470,8 +474,9 @@ if this feature-gate is enabled.
 ###### What are the reasonable SLOs (Service Level Objectives) for the above SLIs?
 
   After this feature is rolled out, there should not be any increase in 95-99 percentile of
-  both `expand_volume` and `volume_fs_resize` durations. Also the error rate should not increase for 
-  `storage_operation_errors_total` metric.
+  both `csi_sidecar_operations_seconds{method_name="/csi.v1.Controller/ControllerExpandVolume"}` and `storage_operation_duration_seconds{operation_name=volume_fs_resize}` durations. 
+  
+  Also the error rate should not increase for `csi_sidecar_operations_seconds{method_name="/csi.v1.Controller/ControllerExpandVolume"}` metric.
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
--- a/keps/sig-storage/1790-recover-resize-failure/README.md
+++ b/keps/sig-storage/1790-recover-resize-failure/README.md
@@ -10,6 +10,7 @@
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
   - [Making allocatedResourceStatus not change unnecessarily for every error in 1.31](#making-allocatedresourcestatus-not-change-unnecessarily-for-every-error-in-131)
+  - [Handling of RWX volumes that don't require node expansion](#handling-of-rwx-volumes-that-dont-require-node-expansion)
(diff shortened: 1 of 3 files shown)


kannon92:
Shadowing @deads2k on this. I had one item that jumped at me reading the PRR.

Was this metric added?

> Are there any missing metrics that would be useful to have to improve observability if this feature?
We are planning to add new counter metrics that will record success and failure of recovery operations. In cases where recovery fails, the counter will forever be increasing until an admin action resolves the error.

> Tentative name of metric is - operation_operation_volume_recovery_total{state='success', volume_name='pvc-abce'}

> The reason of using PV name as a label is - we do not expect this feature to be used in a cluster very often and hence it should be okay to use name of PVs that were recovered this way.

kannon92:
It would be worth expanding on https://github.com/kubernetes/enhancements/blob/d52c7d43a67a3f337e9c8a1869b820e8204af5dd/keps/sig-storage/1790-recover-resize-failure/README.md#scalability as part of the GA graduation.

The answers were a bit vague so we should revisit that section and comment on the scalability questions with more information.

gnufied:
>Was this metric added?

No, we didn't add the metric. I was just discussing with storage folks and while the idea sounds nice on paper, there is very little practical benefit from this particular metric. From operational point of view, I can't see a reason why will an admin be interested in this metric. There are already other resizing related metrics, which work well enough. 

The metric item was tentative, so not sure if blocker for GA.

kannon92:
> > It would be worth expanding on https://github.com/kubernetes/enhancements/blob/d52c7d43a67a3f337e9c8a1869b820e8204af5dd/keps/sig-storage/1790-recover-resize-failure/README.md#scalability as part of the GA graduation.
> 
> I am adding some more information for GA graduation criteria but scalability requirements as such has not changed between beta and GA. Which bits you found vague? I answered couple of questions with - "Potentially Yes", because answer depends on whether a recovery was attempted.

The potentially yes jumped out at me because it sounded like we have not yet investigated if this would occur. 

Maybe you can expand on what you mean by recovery.

deads2k:
> No, we didn't add the metric. I was just discussing with storage folks and while the idea sounds nice on paper, there is very little practical benefit from this particular metric. From operational point of view, I can't see a reason why will an admin be interested in this metric. There are already other resizing related metrics, which work well enough.

What is the alternative method we recommend to a cluster-admin to determine when resize operations are failing on a particular node and/or failing across all nodes.

keps/sig-storage/1790-recover-resize-failure/kep.yaml
```diff
@@ -22,11 +22,12 @@ replaces:
 superseded-by:
 
 latest-milestone: "v1.32"
```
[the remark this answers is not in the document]
gnufied: fixed.

gnufied:
> What is the alternative method we recommend to a cluster-admin to determine when resize operations are failing on a particular node and/or failing across all nodes.

So there is `csi_sidecar_operations_seconds{method_name="/csi.v1.Controller/ControllerExpandVolume}`,  `csi_operations_seconds{method_name="/csi.v1.Node/NodeExpandVolume"}` and `storage_operation_duration_seconds{operation_name="volume_fs_resize"}`, which track these operations already on both controller and node side.  The `csi_xxx` metrics track grpc error codes, so if either Controller or node side expansion is failing, it will be recorded appropriately and a metric emitted.

Some of these metrics have evolved since KEP (This KEP is now 5 years old) was originally written, I will update the KEP with newer metric names.

gnufied:
>The potentially yes jumped out at me because it sounded like we have not yet investigated if this would occur.

I have reworded those, ptal.

keps/sig-storage/1790-recover-resize-failure/kep.yaml
```diff
@@ -22,11 +22,12 @@ replaces:
...
-stage: "alpha"
```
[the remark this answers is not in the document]
gnufied: I added a new section about annotation stuff.

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -330,6 +343,11 @@ There are already e2e tests running that verify correctness of this feature - ht
   -  There are already e2e tests running that verify correctness of this feature -
   https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-cos-alpha-features
   
+### GA  
+
+- The feature has been extensively tested in CI and deployed with drivers in real world. For example - https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L166
+- The test grid already has tests for the feature.
```
kannon92, line 349: Do we have both presubmits and periodics executing this feature?

kannon92:
> * **Were upgrade and rollback tested? Was upgrade->downgrade->upgrade path tested?**
  We have not fully tested upgrade and rollback but as part of beta process we will have it tested.
  
  For stable, this needs to be explained and documented how this was done.

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -330,6 +343,11 @@ There are already e2e tests running that verify correctness of this feature - ht
...
+
 ### Upgrade / Downgrade Strategy
```
msau42, line 351: Can you update the version skew strategy section with the fixes we made?

kannon92:
Reading the Version skew section, this feature seems to have a really complicated rollout plan. Is this documented anywhere? What order do you recommend turning on these features gates? Should someone enable this in external-resizer first? And then enable it in kas, kubelet, kcm?

What is the recommendation for rolling this feature out for operators?

external-resizer must have the feature gate enabled for this to work. The feature gate is still [beta](https://github.com/kubernetes-csi/external-resizer/blob/master/pkg/features/features.go#L33). Could we consider GAing this feature first and then promote the other feature gates?

edit: NVM. You call this out that this is actually bad..

So the recommendation here would be to turn this feature gate on in all kubernetes components first and then you promote the feature gate in external-resizer after skew is satisfied?

keps/sig-storage/1790-recover-resize-failure/README.md
gnufied: Added.

gnufied:
>So the recommendation here would be to turn this feature gate on in all kubernetes components first and then you promote the feature gate in external-resizer after skew is satisfied?

Yes that is generally recommended and I have added this in the latest commit. https://github.com/kubernetes/enhancements/pull/5353/commits/d6f5b11efce3d251f6865287e428b9c43562cc66 

Having said that, there is some fail-safety builtinto this. Older kubelets can work with newer resizer upto a certain version and external-resizer can work with newer api-server etc.  Please see version skew section for more details.

kannon92:
LGTM from PRR shadow standpoint. Thank you for the updates!

/assign @deads2k

deads2k:
PRR lgtm

/approve

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -27,6 +28,7 @@
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
     - [Beta](#beta)
+  - [GA](#ga)
```
jmickey:
```suggestion
    - [Alpha](#alpha)
    - [Beta](#beta)
    - [GA](#ga)
```

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -330,9 +343,15 @@ There are already e2e tests running that verify correctness of this feature - ht
   -  There are already e2e tests running that verify correctness of this feature -
   https://testgrid.k8s.io/presubmits-kubernetes-nonblocking#pull-kubernetes-e2e-gce-cos-alpha-features
   
+### GA
```
jmickey:
```suggestion
#### GA  
```

keps/sig-storage/1790-recover-resize-failure/README.md
```diff
@@ -330,9 +343,15 @@ There are already e2e tests running that verify correctness of this feature - ht
...
+#### GA  
+
+- The feature has been extensively tested in CI and deployed with drivers in real world. For example - https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/master/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L166
+- The test grid already has tests for the feature.
+
 ### Upgrade / Downgrade Strategy
 
-Not Applicable
+- For the case of older kubelet running with newer resizer, the kubelet must handle the newer fields introduced by this feature even if feature-gate is not enabled. Having said that, if kubelet is older and has no notion of this feature at all and `api-server` and `external-resizer` is newer, it is possible that kubelet will not properly update `allocatedResourceStatus` after expansion is finished on the node. This is a known limitation and if user wants to keep running older version of kubelet (older than v1.31 as of this writing and hence unsupported), then they MUST not update external-resizer.
+- In general `external-resizer` should only be upgraded(and `RecoverVolumeExpansionFailure` enabled), after `kubelet`, `api-server` and `kube-controller-manager` already has this feature enabled.
```
msau42, line 354: Now that we support 3 version node skew, does that mean we realistically cannot enable the feature until 3 releases past GA?

gnufied: Doesn't versions upto 1.31 (which have kubelet fix) cover this?  Even if version skew with node is 3 versions?  This does mean - users must upgrade their older nodes to latest z-streams, if they want to upgrade control-plane and use newer resizer.

msau42: I guess for the kubelet case, it's not about enabling the feature. It's about being a node version that has the fix.

gnufied: We can leave external-resizer beta in this release while moving k/k components GA? Or are you thinking we should move everything GA in 1.35 ?

msau42: I think we can go ahead and GA external-resizer. My comment is just a wording nitpick because it currently says don't enable resizer until kubelet has the feature enabled, which would imply kubelets have to be on 1.32 for default feature gate settings.

keps/sig-storage/1790-recover-resize-failure/README.md
jsafrane: The tests are still running only in alpha jobs, despite the feature is already in beta. I am going to LGTM the KEP, but I will block feature gate graduation until the test run in regular e2e jobs.

gnufied: err I misread the text. I will move the test for normal non-featuregated default.

PR #5383, opened 2025-06-06
Update watch-list KEP for next Beta targeted 1.34

wojtek-t:
/assign @jpbetz @deads2k 

/cc @p0lyn0mial

### DIFF
--- a/keps/sig-api-machinery/3157-watch-list/README.md
+++ b/keps/sig-api-machinery/3157-watch-list/README.md
@@ -102,6 +102,7 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Alpha](#alpha)
     - [Beta](#beta)
     - [Beta2](#beta2)
+    - [Beta3](#beta3)
     - [GA](#ga)
     - [Post-GA](#post-ga)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -763,16 +764,28 @@ We expect no non-infra related flakes in the last month as a GA graduation crite
...
+The discussion happened in [this document] and resulted in the following update for the criteria:
+
+- Revert the client-go changes that use watchList to implement List.
+  This inclues removing the API annotation that made this possible, because it doesn't
+  serve another purpose.
+- Ensure we don't break the "latestRV" usecase for StorageVersionMigrator reusing the
+  informer cache from the kube-controller-manager
+- Ensure that the feature is usable by external projects by validating it works with
+  controller-runtime out-of-the-box via simple enablement
+- Add support for AcceptContentType header with the value application/json;as=Table.
```
jpbetz: We also have `meta.k8s.io/v1 PartialObjectMetadata`.  If the intent of this criteria to be that we want to support all our "special" content types, should we call out PartialObjectMetadata? (or do we already support it?)

wojtek-t: This should already work - but I'm fine with adding it explicitly.

jpbetz (approved):
/lgtm
/approve

Thanks for merging the consensus from the doc back into the KEP!

PR #5399, opened 2025-06-11
kep-4317: (Pod Certificates): bump milestones for 1.34

stlaz:
- One-line PR description: bump milestones

- Issue link: https://github.com/kubernetes/enhancements/issues/4317

- Other comments:

/cc enj ahmedtd

@soltysh I see you did the PRR - there shouldn't be any changes necessary, right? We're only moving the alpha of the feature, the implementation hasn't merged yet.


soltysh:
> @soltysh I see you did the PRR - there shouldn't be any changes necessary, right? We're only moving the alpha of the feature, the implementation hasn't merged yet.

Yup, no need for re-applying for PRR since you're keeping it at the same level as before.

PR #5480, opened 2025-08-14
scale approvers for Env Vars since it passed an alpha stage

SergeyKanzhelev:
/sig node

Once KEP passed alpha, we can scale owners. I am marked as an approver for this KEP and I plan to see it thru Beta and GA


mrunalp (approved):
/lgtm

PR #5485, opened 2025-08-17
KEP-3751: add error handling strict mode

huww98:
- One-line PR description: add error handling strict mode

- Issue link: https://github.com/kubernetes/enhancements/issues/3751

- Other comments: Written on top of #5482 . I made a separate PR to keep discussions more focused. Please review the last commit.


keps/sig-storage/3751-volume-attributes-class/README.md
```diff
@@ -685,11 +718,111 @@ spec:
...
+
+Strict mode aims at making sure the quota set by admin is effective at any situation.
+It would help admin to better control the budget and defense malicious users in the situations where SP somehow made partial modification to the volume and failed, user switched target to another VAC which also failed.
+
+It employs a different reconcile logic to handle errors.
+Assuming we are modifying from VAC A to B. The main difference includes:
+* When there is an error, we will only retry with parameters of A or B, not allowing any other target.
+  User can only change the target after either the modify or the rollback succeeded.
+* if the `currentVolumeAttributesClass` is nil, we don't allow rollback.
+  Instead, we allow change target if the error is infeasible in this case, to allow user to correct some typos easily.
```
huww98, line 754: To make sure this one will not break the quota integrity. We will need further restriction in CSI spec:
```diff
- The SP MUST NOT have applied any modification to the volume as part of this specific call
+ The SP MUST NOT have applied any modification to the volume as part of this specific call or any consecutive previous calls with the same parameters.
```
But this also troubles me because it means whenever SP updates it logic, it will need to consider capability to any previous version, since there may be some volumes stuck at error state for years. I know this is a very edge case, but without that, we cannot make a 100% promise even in strict mode.

keps/sig-storage/3751-volume-attributes-class/README.md
```diff
@@ -685,11 +718,111 @@ spec:
...
+* We don't update rollback failed status to PVC, to avoid confusing it with normal operation.
+
+Flow diagram:
+![ModifyVolume Flow Diagram-strict mode](./VolumeAttributesClass-Modify-strict.jpg)
+```go
+func goForward(spec, target, cur, status string) bool {
+    switch spec {
+        case target, "": return true
+        case cur: return false
+        default: return status != "Infeasible"
```
huww98, line 822: This logic may requires more discussion. When current is A, target is B, spec is C, where should we go? Going to C is not allowed because it may break quota.
My current proposal is going to A if ModifyVolume(B) returns infeasible, otherwise continue to retry B, so that it will be more likely to automatically converge to the user specified state.
But maybe we should always retry B to make the system more predictable.

keps/sig-storage/3751-volume-attributes-class/README.md
```diff
@@ -685,11 +718,111 @@ spec:
...
+
+Letter A in the action column means call ControllerModifyVolume with parameters of VAC A.
+
+Number 1 in the last five columns means we should transit to state 1 in this case.
+If action equals to self, we have reached the final state (1, 3, 4, 9) and should stop reconcile until user modifies the spec again.
+
+One can verify this contains all states by arbitrarily changing the spec and verify it will still hit a listed state.
+
+For [strict mode](#strict-mode), the implementation is different:
+* We only leave the state `cur == A && target == B` when either modify to A or modify to B success.
```
huww98, line 808: The major difference to #5462 . I propose NOT to set target to A before rolling back, to keep tracking B if rollback still fails and maintain quota integrity.

PR #5528, opened 2025-09-11
KEP-5004: added rationale for deterministically picking a device class for an extended resource when there are multiple device classes matching

yliaog:
- One-line PR description:
added rationale for deterministically picking a device class for an extended resource when there are multiple device classes matching

- Issue link:
https://github.com/kubernetes/enhancements/issues/5004

- Other comments:
https://github.com/kubernetes/kubernetes/issues/133693#issuecomment-3250346939


yliaog:
friendly ping @macsko

PR #5539, opened 2025-09-17
KEP-4601 / KEP-4633: Mark completed 1.34 sig-auth features as implemented

liggitt:
Mark completed 1.34 sig-auth features as implemented

Closes #4601 #4633

/sig auth
/assign @enj @deads2k

### DIFF
--- a/keps/sig-auth/4601-authorize-with-selectors/kep.yaml
+++ b/keps/sig-auth/4601-authorize-with-selectors/kep.yaml
@@ -4,7 +4,7 @@ authors:
   - "@deads2k"
 owning-sig: sig-auth
 participating-sigs:
-status: implementable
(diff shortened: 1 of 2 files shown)


keps/sig-auth/4940-psa-block-host-field-in-probes/kep.yaml
```diff
@@ -5,7 +5,7 @@ authors:
 owning-sig: sig-auth
 participating-sigs:
   - sig-network
-status: provisional
```
liggitt: > enhancements/keps/prod-readiness/sig-auth/4940.yaml: no such file or directory

this didn't get marked as implementable, so never got a PRR review 🤦

will deal with that in a separate PR

PR #5597, opened 2025-10-01
KEP-4192: Remove obsolete fields from the API

stlaz:
- One-line PR description: The API described in the KEP contained several fields that were unused by the implementation. These should not appear in the Beta version of the API.

- Issue link: #4192 

- Other comments: There are some minor changes to the KEP structure. You may want to go commit-by-commit to see the actual changes to the API, which are isolated in a separate commit.


keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -121,118 +120,93 @@ As an end user using encryption at rest, whenever the key change is detected we
...
+  // +listType=map
+  // +listMapKey=type
+  // +optional
+  Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+  // ResourceVersion to compare with the GC cache for performing the migration.
+  // This is the current resource version of given group, version and resource when
+  // kube-controller-manager first observes this StorageVersionMigration resource.
+  ResourceVersion string `json:"resourceVersion,omitempty"`
+  // LastMigratedResource is used to pick up migration from where it left off in case of failure.
+  LastMigratedResource string `json:"lastMigratedResourceNameHash,omitempty"`
```
michaelasp: Yeah I don't think we need this at all, it doesn't give us anything and adds a lot of overhead since we have to update this every time we update an object.

michaelasp:
Overall makes sense to me, I agree we don't need the last migrated resource at all since we can just use the RV.

michaelasp:
API changes overall make sense, mostly just pruning extra unnecessary fields we don't want to bring into beta. 
/lgtm
/cc @jpbetz @enj

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -121,118 +120,89 @@ As an end user using encryption at rest, whenever the key change is detected we
...
+  // +optional
+  Status StorageVersionMigrationStatus `json:"status,omitempty"`
+}
+
+// Spec of the storage version migration.
+type StorageVersionMigrationSpec struct {
+  // The resource that is being migrated. The migrator sends requests to
+  // the endpoint serving the resource.
+  // Immutable.
+  Resource GroupVersionResource `json:"resource"`
```
enj: IMO this is wrong and should be `GroupResource` as the `GetMonitor` call will ignore the version.  In general, I think we should use whatever version the GC cache is using, and should not take version as input from the user.

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -121,118 +120,89 @@ As an end user using encryption at rest, whenever the key change is detected we
...
+  // Items is the list of StorageVersionMigration
+  // +patchMergeKey=type
+  // +patchStrategy=merge
+  // +listType=map
+  // +listMapKey=type
+  Items []StorageVersionMigration `json:"items" listType:"map" listMapKey:"type" patchStrategy:"merge" patchMergeKey:"type"`
+}
+```
 
 #### Changes while we move above APIs in-tree:
```
enj, line 188: Expand this section to describe the differences.

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -112,124 +111,101 @@ As an end user using encryption at rest, whenever the key change is detected we
...
-- `v1alpha1` of `storageversionmigrations.storagemigration.k8s.io`
+The migrated API is not a mirror of the original. These are the changes:
+- **Dropped fields**
+  - `.spec.continueToken` - unused in the in-tree implementation
+  - `.status.lastMigratedResourceNameHash` - unused in the in-tree implementation
+- **API modifications**
+  - `.spec.groupVersionResource` is replaced by `.spec.groupResource` and the original
+    `.spec.groupVersionResource.version` field is dropped as unused by the in-tree
+    implementation.
+  - `.status.conditions` changes type to `metav1.Condition` and inherits its validation.
```
enj, line 199: `.status.resourceVersion` is new as well.

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -112,124 +111,101 @@ As an end user using encryption at rest, whenever the key change is detected we
...
-
-    ```go
-    // StorageVersionMigration represents a migration of stored data to the latest
-    // storage version.
-    type StorageVersionMigration struct {
-      metav1.TypeMeta `json:",inline"`
-      // Standard object metadata.
-      // More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
-      // +optional
-      metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
```
enj, line 126: Uh why did we drop the proto tags?  Since this will be a built-in API, it will support proto...

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -112,124 +111,101 @@ As an end user using encryption at rest, whenever the key change is detected we
...
+  // +optional
+  Status StorageVersionMigrationStatus `json:"status,omitempty"`
+}
+
+// Spec of the storage version migration.
+type StorageVersionMigrationSpec struct {
+  // The resource that is being migrated. The migrator sends requests to
+  // the endpoint serving the resource.
+  // Immutable.
+  Resource GroupResource `json:"resource"`
```
enj:
```suggestion
  Resource metav1.GroupResource `json:"resource"`
```

(and drop the struct below)

keps/sig-api-machinery/4192-svm-in-tree/README.md
stlaz: They are just inconvenient, I don't feel like they belong to the design. Since it's a built-in API, they should be implied - aren't they also autogenerated?

deads2k:
The result is still serialization compatible for the client requests making it easy to switch, so this lgtm.

/lgtm
/approve
/hold

holding in case @enj  wants another look and for a squash

enj:
/lgtm
/approve
/hold cancel

(labeled so the bot will squash since Standa is OOO)

keps/sig-api-machinery/4192-svm-in-tree/README.md
```diff
@@ -112,124 +111,96 @@ As an end user using encryption at rest, whenever the key change is detected we
...
 To avoid any conflicts with the Storage Version Migrators running out of tree, we will change the _`group`_ from `migration.k8s.io` to `storagemigration.k8s.io`.
 
-The final APIs that will be moved in-tree are:
-- `v1alpha1` of `storageversionmigrations.storagemigration.k8s.io`
+The migrated API is not a mirror of the original. These are the changes:
+- **Dropped fields**
+  - `.spec.continueToken` - unused in the in-tree implementation
+  - `.status.lastMigratedResourceNameHash` - unused in the in-tree implementation
+- **API modifications**
+  - `.spec.groupVersionResource` is replaced by `.spec.groupResource` and changes type
```
michaelasp, line 196: Looking through this again to get https://github.com/kubernetes/kubernetes/pull/134480 into reviewable, I don't think this is possible unless I'm misunderstanding. We need the version to know what resource to migrate, and it is used here https://github.com/kubernetes/kubernetes/pull/134480/files#diff-71ee1fc95d120f0800d32369be68174335815068609f09e5feafc0fb24e68260L42

Without it, we can't tell which version of a resource to migrate, I think we need to have groupVersionResource. cc @enj

Issue #4963, opened 2024-11-15
Kube-proxy Services Acceleration


aojea:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Service acceleration for Kube-proxy using netfilter’s flowtable infrastructure
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/4964
- Discussion Link: https://github.com/kubernetes/kubernetes/pull/128392
- Primary contact (assignee): aojea
- Responsible SIGs: SIG-Network
- Enhancement target (which target equals to which milestone):
  - Alpha release target: 1.33
  - Beta release target: 1.34
  - Stable release target: 1.35
- [ ] Alpha
  - [ ] KEP (`k/enhancements`) update PR(s):  https://github.com/kubernetes/enhancements/pull/4964
  - [ ] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/128392
  - [ ] Docs (`k/website`) update PR(s):

aroradaman:
/reopen
resuming work on this.

PR #5629, opened 2025-10-07
KEP-4963: Kube-proxy Services Acceleration

aroradaman:
- One-line PR description: Service acceleration for Kube-proxy using netfilter’s flowtable infrastructure

- Issue link: https://github.com/kubernetes/enhancements/issues/4963

- Other comments:


danwinship:
So I still have the same concern I had before: there is no reason given for limiting this to only Service traffic. Of course, the reason the earlier KEP proposed it that way is because kube-proxy was the only place where SIG Network could insert this code (even though it [barely interacted with the rest of kube-proxy](https://github.com/kubernetes/kubernetes/pull/128392#issuecomment-2450082487)). But, as Calico already demonstrated, a network plugin could implement the same functionality for both Service and non-Service traffic, without needing anything from kube-proxy. (In fact, it would want to disable kube-proxy's own offloading.)

Given that we are now planning to [make kindnet a SIG Network project](https://github.com/kubernetes/org/issues/5801), I think the right answer is to implement offload _there_. But apparently we already do implement it there, so mission accomplished?

(The other possibility is to add a "usage hint" field to Service, which kube-proxy could then take advantage of to know that certain kinds of optimizations, like offload, were useful, https://github.com/kubernetes/kubernetes/issues/127259#issuecomment-2436201258. _Then_ it would make sense to implement offload in kube-proxy, for Services that were hinted in a way that indicated they'd benefit from it.)

adrianmoisey:
> (The other possibility is to add a "usage hint" field to Service, which kube-proxy could then take advantage of to know that certain kinds of optimizations, like offload, were useful, [kubernetes/kubernetes#127259 (comment)](https://github.com/kubernetes/kubernetes/issues/127259#issuecomment-2436201258). _Then_ it would make sense to implement offload in kube-proxy, for Services that were hinted in a way that indicated they'd benefit from it.)

I like this idea.
My concern with the current proposal, is that if I want to change how this feature works, it requires me configuring kube-proxy, something we normally don't do often. 
Accessing this feature via the Services API would be nicer, and give more control.

keps/sig-network/4963-kube-proxy-services-acceleration/README.md
```diff
@@ -0,0 +1,524 @@
...
+the option to disable the feature entirely.
+
+Flowtables netfilter infrastructure is not well documented and we need to validate assumptions to avoid unsupported or suboptimal configurations. Establishing good relations
+and involve netfilter maintainers in the design will mitigate these possible problems.
+
+## Design Details
+
+### Service API Extension
+
+Users can set the `TrafficHint` field to opt in to fastpath acceleration for a Service. In the future, this field can also be leveraged to enable other dataplane optimizations
```
adrianmoisey, line 130: I very much like the opt-in nature of this. If users ignore the feature, nothing changes.

keps/sig-network/4963-kube-proxy-services-acceleration/README.md
```diff
@@ -0,0 +1,524 @@
...
+such as GRO-based UDP packet aggregation, reduced conntrack timeouts, or traffic distribution towards local endpoints. These optimizations are workload-dependent and only
+beneficial for certain traffic patterns. The `TrafficHint` field provides kube-proxy with contextual information about the expected traffic semantics, allowing it to program
+the dataplane with suitable optimizations for the Service.
+
+```go
+// +enum
+// TrafficHint indicates the expected traffic behavior of a Service.
+// Service implementations may use this hint to apply dataplane optimizations, which can vary
+// depending on the type of optimization and the technology used to program the dataplane.
+type TrafficHint string
```
adrianmoisey, line 140: Is there any need to leave space in the API to allow for configuration?
I don't know what the future holds with other expansions here, but what if other modes require settings?
For example: with LongLived, may be we need to configure how long before it takes the fast path?
Or with a reduced conntrack timeout, it may need a configuration as to how long


My question here may come with a lot of inexperience, so it may not be valid.

aroradaman: I thought about that while writing this, but realised it will make the Service API too complicated. 

I think we can use gateway-api later to have different types of services maybe separate kinds for LondLivedService, or ShortLivedService and specs for each kind can have configurable attributes.

aroradaman: I will defer to @danwinship @aojea @thockin for this.

danwinship: My intent was that the hint field that would cover a lot of possibilities, and would basically be a full KEP on its own. And I'm not sure it would end up being a good idea in the end.

PR #5630, opened 2025-10-07
KEP-5284: update specifics based on implementation and feedback

enj:
- One-line PR description: KEP-5284: update specifics based on implementation and feedback
- Issue link: #5284

/assign liggitt deads2k


liggitt:
/lgtm

matches updates discussed in in-person review

PR #5640, opened 2025-10-09
promote kep-3673 to GA

pacoxu:
- One-line PR description: kep 3673 GA

- Issue link: https://github.com/kubernetes/enhancements/issues/3673

- Other comments: for GA, we  change default behavior to  `serializeImagePulls: false` and `maxParallelImagePulls: 2`.

### DIFF
--- a/keps/sig-node/3673-kubelet-parallel-image-pull-limit/README.md
+++ b/keps/sig-node/3673-kubelet-parallel-image-pull-limit/README.md
@@ -383,21 +383,25 @@ We expect no non-infra related flakes in the last month as a GA graduation crite
 -->
 
 A new node_e2e test with  `serialize-image-pulls==false` will be added test parallel image pull limits.
+
 1. When maxParallelImagePulls is reached, all further image pulls will be blocked.
 2. Verify the behavior when the same image is pulled in parallel, which will happen when image pull policy is `Always`.
 
-- <test>: <link to test coverage>
+- pull image parallel test cases: https://github.com/kubernetes/kubernetes/blob/6c258fa74b2f0644a6b31a7ce3e613dda41effd4/test/e2e_node/image_pull_test.go
 
 ### Graduation Criteria
 
 #### Alpha
+
 - Initial e2e tests completed and enabled
 
 #### Beta
+
 - Gather feedback from developers and surveys
 - Add e2e test to cover the parallel image pull case
 
 #### GA
+
 - Gather feedback from real-world usage from kubernetes vendors.
 - Allowing time for feedback.
 
@@ -566,8 +570,8 @@ feature.
 NOTE: Also set `disable-supported` to `true` or `false` in `kep.yaml`.
 -->
 
-The code change itself cannot be rolled back. But a user can roll back to the existing default behavior by setting `maxParallelImagePulls` to 0, or not setting it and letting it default to 0.
-
+A user can roll back to the previous default behavior by setting `serialize-image-pulls` to true and restarting kubelet.
+Similarly, setting `maxParallelImagePulls` to 1 will be equivalent to setting `serialize-image-pulls` to true.
 
 ###### What happens if we reenable the feature if it was previously rolled back?
 
@@ -575,7 +579,7 @@ Nothing will happen.
 
 ###### Are there any tests for feature enablement/disablement?
 
-This feature is purely in-memory, so such test isn't really needed.
+Yes, see e2e tests section above.
 
 <!--
 The e2e framework does not currently support enabling or disabling feature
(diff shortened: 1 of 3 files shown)


keps/sig-node/3673-kubelet-parallel-image-pull-limit/README.md
```diff
@@ -383,21 +383,26 @@ We expect no non-infra related flakes in the last month as a GA graduation crite
...
 - Initial e2e tests completed and enabled
 
 #### Beta
+
 - Gather feedback from developers and surveys
 - Add e2e test to cover the parallel image pull case
 
 #### GA
+
+- Change the default value of `serialize-image-pulls` to false and set the default value of `maxParallelImagePulls` to 2.
```
wojtek-t: Hmm - do we really want to change the default for "GA"?

With https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/5241-beta-featuregate-promotion-requirements, we generally want the GA to effectively be kind of "no-op". Changing the default might be a bit unexpected here.

I know that it doesn't explicitly affect the user (it may affect them implicitly because some pods startup (due to image pulling) may be slower/faster), but still it might not be intuitive.

Let me ping other PRR approvers about it for their thoughts.

pacoxu: If so, should we make this step as beta-2 for this KEP to change the default?

wojtek-t: So I thought about it and we discussed that a bit with other PRR approvers. Based on that discussion I have few questions before we make a final decision:

1) My understanding was that before this KEP, the behavior was that the default behavior was "unbounded parallel image pulls".
But now looking into this proposal I actually see this sentence: "Before this proposal, serialize-image-pulls is by default true", which suggests that this isn't true.
So either my understanding was incorrect or this sentence is not true or I don't understand the semantics of `serialize-image-pull` or something different.

Can you please clarify what is the current (before this proposal) default semantics?


2) If indeed I was wrong and by default we were serializing image pulls by default, then switching the default seems like a "no-go" (no matter if it would be ga, another beta or anything else) - that just seems like a potentially breaking change to users.

SergeyKanzhelev: serialized pull was the default. We are making the default marginally better by allowing 2 parallel pulls to improve reliability (one bad image will not break the whole node).

This is not strictly required. Many installations override this default anyways as it is not reliable anyways. It is just nice to do if somebody just trying it out.

If PRR is blocked on this, let's just keep old defaults and forget about it.

pacoxu: 1. Previous default behavior: serialized pull. One image pulling may block a node pulling new images for new pods.
2. Previous serialize-image-pull=false without maxParallelImagePulls set: unlimited parallel pull, even setting registryPullQPS and registryBurst(See https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/3673-kubelet-parallel-image-pull-limit/README.md#qpsburst-limits-on-kubelet-are-confusing). unlimited parallel pull may have too much IO pressure to disk.
3. After this default value change, we want to allow 2 parallel pulls to improve reliability (one bad image will not break the whole node), as Sergey said. 

I think 1 or 2 are not ideal default behavior and 3 would be a better default behavior, and this is a `breaking` change in some extend, but acceptable.

PS. 2 is a very conservative and cautious approach.

wojtek-t: So I would like to distinguish two things:
1) Changing the default from serialized to parallel
2) Changing the default for number of ParallelImagePulls if serialize is false from unbounded to 2.

In (1) we change the default behavior even for administrators that aren't aware of how image pulls work. I agree that it helps in some cases (blocking image pulls by a single pull), but otoh it may negatively affect existing usecases (if I have two large pulls at the same time, it's actually better to download the first one and only then start the second, rather than having the first one to last 2x longer).
So I don't think we should really change that default - the mitigation for administrator is to configure their setup with parallel pulls and they can do that with this feature.

In (2) - given that "serialize-image-pulls=false" is not a default, then administrators are already aware of it and configuring that. I agree that "unbounded" is bad and switching to MaxImagePulls=2 in such case (if someone didn't set it explicitly) seems helpful. I'm fine with this change, but the consensus among prr approvers is that it would have to be a second beta then.
[And I would actually go with this option.]

SergeyKanzhelev: I would prefer to just GA this KEP then. Dragging this longer for the default configuration update that most environments needs to fine tune anyways doesn't sound attractive.

pacoxu: There seems to be 3 proposals:
1.  Change both default values: serializeImagePulls: false and maxParallelImagePulls: 2
2. Changing the default for number of ParallelImagePulls if serialize is false from unbounded to 2.
3. do nothing and GA.

@wojtek-t +1 for 2, and @SergeyKanzhelev +1 for 3. 

IIRC, we have discussed on whether we should set a default value for maxParallelImagePulls during KEP initialization. 

https://github.com/kubernetes/kubernetes/blob/42ee6dafd55e411a258f6e5d947b1fb95f38a6b5/pkg/kubelet/apis/config/v1beta1/defaults.go#L226-L235

```
	if obj.SerializeImagePulls == nil {
		// SerializeImagePulls is default to true when MaxParallelImagePulls
		// is not set, and false when MaxParallelImagePulls is set.
		// This is to save users from having to set both configs.
		if obj.MaxParallelImagePulls == nil || *obj.MaxParallelImagePulls < 2 {
			obj.SerializeImagePulls = ptr.To(true)
		} else {
			obj.SerializeImagePulls = ptr.To(false)
		}
	}
```

Currently, setting `MaxParallelImagePulls=2` will enable parallel image pulling if SerializeImagePulls is not set.
So solution 2 may be something like below.
```
	if obj.SerializeImagePulls == nil {
		// SerializeImagePulls is default to true when MaxParallelImagePulls
		// is not set, and false when MaxParallelImagePulls is set.
		// This is to save users from having to set both configs.
		if obj.MaxParallelImagePulls == nil || *obj.MaxParallelImagePulls < 2 {
			obj.SerializeImagePulls = ptr.To(true)
		} else {
			obj.SerializeImagePulls = ptr.To(false)
		}
+	} else if !*obj.SerializeImagePulls && obj.MaxParallelImagePulls == nil  {
+		obj.MaxParallelImagePulls = ptr.To[int32](2)
+	}
```

I think I would +1 for do nothing and GA this.

pacoxu: Updated to do nothing and GA this KEP(adding the new maxParallelImagePulls configuration in kubelet, without a FG).

wojtek-t: > Currently, setting MaxParallelImagePulls=2 will enable parallel image pulling if SerializeImagePulls is not set.

That makes sense. And I think that helps with what we need.
I'm ok with GA-ing as is given the above.

PR #5646, opened 2025-10-09
[KEP-5573]: follows up from review

kannon92:
Address followups such as oom.group risk, calling out kubeadm as out of scope and renaming the top level folder to reflect the true name of the KEP.


keps/sig-node/5573-remove-cgroup-v1/README.md
```diff
@@ -158,7 +158,9 @@ the codebase.
 
 ### Non-Goals
 
-NA.
+In this KEP, we will focus on the kubelet related work to remove cgroup v1.
+Projects like minikube, kubeadm, kubespray and others may need work to support this flag going to false
```
BenTheElder, line 164: kubeadm already supports customizing kubelet config so I'm not sure if anything needs to be done here, updates to the preflight checks? docs?

kubeadm is a little bit weird because it's in-tree and used for release-blocking signal (similarly if the node_e2e or kube-up scripts didn't work we'd have a problem), I think calling out out-of-tree projects as out of scope is at least reasonable.

It's not super great to release k/k with the release binaries out of sync.

BenTheElder: The rest of the diff LGTM, not sure about this part.

kannon92: > It's not super great to release k/k with the release binaries out of sync.

I don't follow what you mean here.

kannon92: > kubeadm already supports customizing kubelet config so I'm not sure if anything needs to be done here, updates to the preflight checks? docs?

https://github.com/kubernetes/enhancements/pull/5574#discussion_r2386875178

BenTheElder: > I don't follow what you mean here.

Saying "this kep won't wait for another core release binary to be ready" is a little different from "we're not going to block on out of tree projects updating", kubeadm is part of the core kubernetes release. We shouldn't release a broken state.

> https://github.com/kubernetes/enhancements/pull/5574#discussion_r2386875178

I see ... commented

pacoxu: > kubeadm is a little bit weird because it's in-tree and used for release-blocking signal (similarly if the node_e2e or kube-up scripts didn't work we'd have a problem), I think calling out out-of-tree projects as out of scope is at least reasonable.

That's true and we also have an issue to move it out which is not updated for a long time. https://github.com/kubernetes/enhancements/issues/1424


https://github.com/kubernetes/system-validators/blob/a7fadf67da1228984480d7e8dd9c8ae39803a442/validators/cgroup_validator_linux.go#L117
```
		warns = append(warns, errors.New("cgroups v1 support is in maintenance mode, please migrate to cgroups v2"))
```

We have a warning message when cgroup v1 support went to maintenance mode. 

After this change happened in kubelet, we will update it to an error IIUC. This will be tracked in https://github.com/kubernetes/system-validators/issues/58.

kannon92: I took a stab at addressing this. I'm not entirely sure what the goal is for system-validators but I call that out as a goal for this KEP.

SergeyKanzhelev:
re-lgtm the change

/lgtm

PR #5692, opened 2025-11-17
KEP-4817: rbac permissions to update the claim status

aojea:
Update the KEP with the new logic to restrict changes to the claim status

Implementation in https://github.com/kubernetes/kubernetes/pull/134947

Discussed during kubecon

/assign @johnbelamaric @pohly @liggitt @enj 
/cc @klueska @LionelJouin

### DIFF
--- a/keps/sig-node/4817-resource-claim-device-status/README.md
+++ b/keps/sig-node/4817-resource-claim-device-status/README.md
@@ -298,29 +298,31 @@ to be reported only once in the slice, a device is being identified  by
 
 ### Write Permission
 
-To prevent unauthorized or accidental modifications by entities that do not 
-have access to a particular resource, a `ValidatingAdmissionPolicy` will be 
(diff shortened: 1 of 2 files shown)


pohly:
@aojea: do you want to discuss the proposal here in the KEP update PR or in the implementation PR?

As the implementation PR already has some comments, I prefer to continue there.

aojea:
> @aojea: do you want to discuss the proposal here in the KEP update PR or in the implementation PR?

I do not mind, one or other place, but after talking with all the different stakeholders during kubecon I had the sensation that everyone was ok with this proposal, is there any conflicting point that needs clarification or that is controversial?

johnbelamaric (commented):
LGTM - needs SIG approval

johnbelamaric:
/assign @mrunalp 

For SIG Node approval

PR #5704, opened 2025-11-21
KEP-3983: mark as implemented

SergeyKanzhelev:
KEP #3983 

/sig node

Closes #3983

### DIFF
--- a/keps/sig-node/3983-drop-in-configuration/kep.yaml
+++ b/keps/sig-node/3983-drop-in-configuration/kep.yaml
@@ -5,7 +5,7 @@ authors:
   - "@yuqi-zhang"
   - "@sohankunkerkar"
 owning-sig: sig-node
-status: implementable
+status: implemented
 creation-date: 2023-05-04
 last-updated: 2024-09-30
 reviewers:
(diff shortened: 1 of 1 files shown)


keps/sig-node/3983-drop-in-configuration/kep.yaml
lmktfy: This seems wrong @SergeyKanzhelev.

keps/sig-node/3983-drop-in-configuration/kep.yaml
```diff
@@ -5,7 +5,7 @@ authors:
   - "@yuqi-zhang"
   - "@sohankunkerkar"
 owning-sig: sig-node
-status: implementable
+status: implemented
 creation-date: 2023-05-04
 last-updated: 2024-09-30
```
lmktfy, line 10: Bump this too.

SergeyKanzhelev: I do not think we follow this pattern in other places. Mostly updating when the KEP text was updated or milestones changed

keps/sig-node/3983-drop-in-configuration/kep.yaml
kannon92: I agree. The feature is stable. Could you update this also with this update?

SergeyKanzhelev: fixed. Not sure how it was missed during the review.

SergeyKanzhelev:
@kannon92 @mrunalp can you lgtm pls - I fixed the latest stage

PR #5776, opened 2026-01-09
KEP-5776: Configurable cAdvisor Metrics Collection

NahumLitvin:
## Summary

This KEP proposes adding a `cadvisor` configuration section to KubeletConfiguration that allows operators to selectively disable expensive cAdvisor metric collectors. The primary use case is disabling `ProcessMetrics` collection, which scans `/proc` for every thread in every container and causes extreme CPU overhead on high-density nodes.

## Production Evidence

Testing on EKS 1.31 clusters with ~200 pods/node showed:
- **Kubelet CPU: 1332% → 8% (99.4% reduction)**
- Node Total CPU: 44% → 13% (70% reduction)
- ~10 cores freed per node

## Proposal

- New alpha feature gate: `ConfigurableCAdvisorMetrics` (target 1.33)
- New `KubeletConfiguration.cadvisor.includedMetrics.processMetrics` (default `true`)
- Maintains backward compatibility (all metrics enabled by default)

## Related

- kubernetes/kubernetes#123340 (original issue)
- Implementation PR (pending): kubernetes/kubernetes

/sig node
/kind kep

### DIFF
--- a/keps/sig-node/5000-configurable-cadvisor-metrics/README.md
+++ b/keps/sig-node/5000-configurable-cadvisor-metrics/README.md
@@ -1,0 +1,281 @@
+# KEP-5000: Configurable cAdvisor Metrics Collection
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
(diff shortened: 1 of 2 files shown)


SergeyKanzhelev:
you need to have a KEP issue first

/sig instrumentation

rexagod:
Per-component metric manipulations should ideally happen through a common interface: https://github.com/kubernetes/kubernetes/pull/131572, for multiple reasons (common underlying DX expectations, ability to ship faster as most changes can be made in component-base and trickle down to components, etc.).

For component-specific behaviors, it'd be good to have those gates and their configurations exposed through the same interface, albeit the implementation currently does not account for that, and will need to be updated to accomodate for per-component behaviors.

This is also something that should be discussed first because this effectively goes in the other direction, as currently all options exposed in the Metrics API as well as their behaviors all live in component-base.

PR #5784, opened 2026-01-13
kep-4317: add an in-cluster serving signer

stlaz:
- One-line PR description: proposes an in-tree serving certificates signer to the PodCertificates KEP

- Issue link: https://github.com/kubernetes/enhancements/issues/4317

- Other comments: The proposed feature is to be gated by a feature gate that has its own maturity that is independent of the other feature gates mentioned in the KEP.


stlaz:
https://github.com/kubernetes/kubernetes/pull/139409 is a PoC for the signer

PR #5801, opened 2026-01-16
KEP-4622: mark as implemented

ffromani:
- One-line PR description: mark KEP as implemented

- Issue link: https://github.com/kubernetes/enhancements/issues/4622

- Other comments: GA in 1.35


keps/sig-node/4622-topologymanager-max-allowable-numa-nodes/kep.yaml
lmktfy, line 1: If this is implemented, I expect to see `disable-supported: false` for the feature gate(s).

Does that sound right?

ffromani: I think you are right, thanks for pointing this out. This prompted me to do https://github.com/kubernetes/enhancements/pull/5824

PR #5802, opened 2026-01-16
KEP-4210: mark as implemented

haircommander:
- One-line PR description:

- Issue link:
closes https://github.com/kubernetes/enhancements/issues/4210

- Other comments:


keps/sig-node/4210-max-image-gc-age/kep.yaml
lmktfy, line 1: If this is implemented, I expect to see `disable-supported: false` for the feature gate(s).

Does that sound right?

haircommander: good point, done

PR #5819, opened 2026-01-20
KEP 5067: Mark Pod Generation as implemented

natasha41575:
Closes https://github.com/kubernetes/enhancements/issues/5067

Mark Pod Generation KEP as implemented as it has graduated to stable as of v1.35. 

/assign @tallclair @dchen1107 
/cc @haircommander @SergeyKanzhelev

### DIFF
--- a/keps/sig-node/5067-pod-generation/kep.yaml
+++ b/keps/sig-node/5067-pod-generation/kep.yaml
@@ -5,7 +5,7 @@ authors:
   - "@tallclair"
 owning-sig: sig-node
 participating-sigs:
-status: implementable
+status: implemented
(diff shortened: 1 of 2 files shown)


keps/sig-node/5067-pod-generation/kep.yaml
lmktfy, line 1: If this is implemented, I expect to see `disable-supported: false` for the feature gate(s).

Does that sound right?

natasha41575: done

natasha41575:
@dchen1107 or @mrunalp can you approve?

PR #5837, opened 2026-01-26
KEP-5328: widen approvers

SergeyKanzhelev:
This is a proposal on how we can mark and in future automate OWNER files:

https://github.com/kubernetes/community/pull/8794/

We can mark approvers and reviewers with the special comment so they will appear explicitly in OWNERS file for this KEP.

chars: @haircommander @mrunalp for discussion

### DIFF
--- a/keps/sig-node/5328-node-declared-features/OWNERS
+++ b/keps/sig-node/5328-node-declared-features/OWNERS
@@ -1,0 +1,10 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+# Scaling up SIG Node KEP approvers: https://github.com/kubernetes/community/blob/master/sig-node/CONTRIBUTING.md#scaling-up-kep-approvers
+
+reviewers:
+  - SergeyKanzhelev # sig-node-assigned-reviewer
+  - haircommander # sig-node-assigned-reviewer
+
+approvers:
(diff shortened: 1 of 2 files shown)


haircommander:
I find myself wondering if we keep a list of reviews/approvers/tls, could we just infer which roles each person is taking for each KEP? we can ignore names that aren't in those lists (e.g. for other SIGs) but auto populate based on inclusion on reviewers/approvers, instead of adding a tag. WDYT

SergeyKanzhelev:
> I find myself wondering if we keep a list of reviews/approvers/tls, could we just infer which roles each person is taking for each KEP? we can ignore names that aren't in those lists (e.g. for other SIGs) but auto populate based on inclusion on reviewers/approvers, instead of adding a tag. WDYT

I thought about it, but the list like this will be a new ladder step in maintainers progression that I do not want to discuss or deal with nominating/approvalship. It is easier to mark the proper reviewers and approvers using comments like in this PR. Wdyt?

haircommander:
/lgtm

let's try it out! we can always revise later

SergeyKanzhelev:
Delegating approvership to @tallclair. Discussed at SIG Node meeting KEP planning 1/27/2026

/assign @dchen1107

PR #5847, opened 2026-01-28
KEP-127: Let's go GA in 1.36!

rata:
- One-line PR description: Move UsernamespaceSupport to GA in 1.36

- Issue link: #127 

- Other comments:

cc @giuseppe @haircommander @saschagrunert @mrunalp

### DIFF
--- a/keps/sig-node/127-user-namespaces/README.md
+++ b/keps/sig-node/127-user-namespaces/README.md
@@ -27,7 +27,6 @@
     - [Regarding the previous implementation for volumes](#regarding-the-previous-implementation-for-volumes)
     - [Non-conformant volume types](#non-conformant-volume-types)
   - [Pod Security Standards (PSS) integration](#pod-security-standards-pss-integration)
-  - [Unresolved](#unresolved)
   - [Test Plan](#test-plan)
       - [Prerequisite testing updates](#prerequisite-testing-updates)
       - [Unit tests](#unit-tests)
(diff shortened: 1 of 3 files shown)


rata:
@wojtek-t the PRR was already completed and is unchanged. Not sure if you want to have another check before we go GA.

wojtek-t:
You will need GA block and my approval for it in:
https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-node/127.yaml

so I will take another look, but as you mentioned it should already be in a good shape

rata:
@wojtek-t oh, the beta migration was so long ago that I completely forgot that. Added now, thanks! Thanks for taking the time to review, all the way from alpha too! :)

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -1322,6 +1304,7 @@ be the cause of the problem.
 - Kubernetes 1.30: Feature went off-by-default beta
 - Kubernetes 1.33: Feature goes on-by-default beta
 - Kubernetes 1.34: Feature adds metrics
+- Kubernetes 1.36: Feature goes GA
 
 ## Drawbacks
```
lmktfy, line 1307: nit: could actually remove this section

keps/sig-node/127-user-namespaces/kep.yaml
```diff
@@ -16,10 +16,11 @@ approvers:
   - "@derekwaynecarr"
 
 stage: beta
```
wendy-ha18:
```suggestion
stage: stable
```

wendy-ha18 (changes requested):

keps/sig-node/127-user-namespaces/kep.yaml
rata: Fixed, thanks!

keps/sig-node/127-user-namespaces/README.md
```diff
@@ -82,6 +81,10 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 [kubernetes/kubernetes]: https://git.k8s.io/kubernetes
 [kubernetes/website]: https://git.k8s.io/website
 
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
```
wojtek-t, line 87: From PRR perspective:

> Are there any tests for feature enablement/disablement?

The answer there requires some update, though it's probably too late for adding those now (especially given we're close to locking the gate anyway).
So I will not block on it at this point.

> What specific metrics should inform a rollback?

"The following kubelet metrics will be added" - I see they were already added, so it would be good to update it to reflect it.

> Are there any missing metrics that would be useful to have to improve observability of this feature?

Same comment about reflecting these were added.

rata: Updated to reflect metrics were added. And regarding the tests, we had them already, added the links to them :)

See commit  "KEP-127: Reflect metrics and tests we already added" for these changes :)

wojtek-t:
/lgtm
/approve PRR

Thanks!

PR #5848, opened 2026-01-28
KEP-5073: Validation Lifecycle Mechanism and Updated Rollout Plan

yongruilin:
- One-line PR description:
Update KEP-5073 with Shadow Validation Semantic

- Issue link:
#5073 

- Other comments:

### DIFF
--- a/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
+++ b/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
@@ -3,6 +3,9 @@
 - [Release Signoff Checklist](#release-signoff-checklist)
 - [Summary](#summary)
 - [Motivation](#motivation)
+  - [The &quot;Implicit Shadow&quot; Reality](#the-implicit-shadow-reality)
+  - [Problem: Global vs. Local Control](#problem-global-vs-local-control)
+  - [Solution: Lifecycle Tags](#solution-lifecycle-tags)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
```diff
@@ -754,6 +763,82 @@ Once validation is generated, it will be easy to opt-in (see below snippet).
...
+*   **Legacy APIs**: May remain in “Implicit Shadow” mode without strategy changes. They transition to Explicit mode only when ready.
+*   **Safe Promotion**: To ensure safe rollout, a new or tightened validation rule must be shadowed for at least 2 releases before being promoted.
+*   **Linting**: Tooling should block the removal of the shadow prefix if the recorded version is less than 2 releases old.
+
+##### Stability Constraints (Alpha/Beta Tags)
+*   **Standard Tags**: MUST be Stable when using `ExplicitShadowing`.
+*   **Legacy Exemption**: Legacy types in "Implicit Shadow" mode are exempt.
+*   **Migration Rule**: Before enabling `WithExplicitShadowing()` for a resource, developers MUST ensure that all existing standard Alpha/Beta tags are converted to `+k8s:shadow=` tags.
+
+#### Example Walkthrough: Two-Field Migration
```
deads2k: This example helps a lot, thank you.

If I'm reading correctly, we go from phase 2 (not enforcing) to phase 3 (enforcing and cannot be disabled) without a state of "enforcing and can be disabled"?

If we have such a state, then I think the progression ends up looking like

1. release/N add validation to replace hand written validation with shadow.  Not enforcing, but collecting data.
2. release/N+1 (near the end) no failures found.  Enforcing, but can be disabled
3. release/N+2 (near the end) no failures found. Enforcing and cannot be disabled.

This would allow enforcement faster, but allow field correction if we have a problem.  It would only be for migrated fields, for new fields I think we're ready to enforce from the beginning.

deads2k:
Thanks, I think this looks good.

/lgtm
/hold

holding so @jpbetz can have a look too
/assign @jpbetz

PR #5886, opened 2026-02-03
KEP-5284: beta

enj:
- One-line PR description: KEP-5284: beta
- Issue link: #5284

/assign deads2k liggitt qiujian16

### DIFF
--- a/keps/sig-auth/5284-constrained-impersonation/README.md
+++ b/keps/sig-auth/5284-constrained-impersonation/README.md
@@ -1178,14 +1178,52 @@ There are existing metrics to record authz latency and request number:
 These will be expanded to include metrics that cover the impersonation handler
 to give an overall sense for the cost of impersonation:
 
-- apiserver_impersonation_attempts_total (success per mode or failure)
-- apiserver_impersonation_duration_seconds (success per mode or failure)
+- `apiserver_impersonation_attempts_total{status}`
+
+This will be incremented whenever impersonation is attempted.  On success, the
+`status` label will capture which mode was successful.  On failure, the `status`
+label will be set to `failed`.
+
+Labels and possible values:
+
+`status`: `associated-node`, `arbitrary-node`, `serviceaccount`, `user-info`, `legacy`, `failed`
+
+- `apiserver_impersonation_duration_seconds{status}`
+
+This is a Histogram metric that will track the time impersonation took to resolve
+the impersonated user whenever impersonation is attempted.  On success, the
+`status` label will capture which mode was successful.  On failure, the `status`
+label will be set to `failed`.
+
+Note that because of the caching performed within the handler, we expect this metric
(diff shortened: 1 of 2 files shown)


keps/sig-auth/5284-constrained-impersonation/README.md
```diff
@@ -1177,6 +1175,18 @@ There are existing metrics to record authz latency and request number:
 - webhook authorizer match condition latency
 - webhook authorizer match condition success
 
+These will be expanded to include metrics that cover the impersonation handler
+to give an overall sense for the cost of impersonation:
+
+- apiserver_impersonation_attempts_total (success per mode or failure)
```
liggitt: because we're planning to check the new modes first, existing users of legacy impersonation would show up as metric failures for the constrained modes, right?

liggitt: can we answer the question "how many requests successfully impersonated" and "how many requests attempted impersonation and failed impersonation" without regard for the constrained-mode → legacy fallback? because we do fallbacks (associated-node → any-node → legacy), I'm trying to figure out how we would interpret failure counts per mode

for example, if failed impersonations prior to this feature are at X% and failed impersonations with the feature enabled are at 2X%, that would seem significant

liggitt: or are you suggesting we would only record success-per-mode, or failure (not failure-per-mode)?

keps/sig-auth/5284-constrained-impersonation/kep.yaml
```diff
@@ -29,3 +29,7 @@ disable-supported: true
 
 # The following PRR answers are required at beta release
 metrics:
+  - apiserver_impersonation_attempts_total
+  - apiserver_impersonation_duration_seconds
+  - apiserver_impersonation_authorization_attempts_total
+  - apiserver_impersonation_authorization_duration_seconds
```
liggitt: it's helpful to include the labels and possible values here if we know them (here and in the markdown)

for example https://github.com/kubernetes/enhancements/tree/master/keps/sig-auth/3221-structured-authorization-configuration#monitoring and https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/3221-structured-authorization-configuration/kep.yaml#L31

keps/sig-auth/5284-constrained-impersonation/README.md
liggitt: > or are you suggesting we would only record success-per-mode, or failure (not failure-per-mode)?

talked offline, this was the intent

update this to make that clearer, then lgtm

liggitt: talked offline, failure metrics weren't per-mode... update this to make it clearer (explicitly listing the labels and values will help), then lgtm

deads2k:
Once updates already noted are made, lgtm.

Issue #5894, opened 2026-02-05
Node system partition


SergeyKanzhelev:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Node Partitions is an way to split a Node into partitions, each of them hosting a set of Pods and define a set of resources allocated for the partition.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5894-node-system-partition/
- Discussion Link: https://docs.google.com/document/d/1_nu8mTTOKASMnhB73DZlxynA9j9MK79U-H501TBjiuM/edit?tab=t.0#heading=h.2dndymo89y0f

/sig node

- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [X] KEP (`k/enhancements`) update PR(s): 
        - [ ] https://github.com/kubernetes/enhancements/pull/5895
        - [ ] https://github.com/kubernetes/enhancements/pull/6387
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s): https://github.com/kubernetes/website/pull/56311

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

pacoxu:
/milestone clear
since the KEP was not merged yet https://github.com/kubernetes/enhancements/pull/5895.

whtssub:
Hi @SergeyKanzhelev @pacoxu  👋, v1.37 Enhancements Lead here.

We are already at the start of v1.37 release. As part of cleanup, we are removing the `lead-opted-in` and `tracked/*` labels and milestone to prepare for the next cycle.

If you would like to work on this enhancement in v1.37, please have the SIG lead opt in by adding the `lead-opted-in` label and setting the milestone using `/milestone v1.37`, which will ensure it gets added to the tracking board for next release.

Thanks!

/remove-label lead-opted-in
/remove-label tracked/yes
/remove-label tracked/no
/milestone clear

tico88612:
Hello @SergeyKanzhelev 👋, 1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.37 (correct me, if otherwise)

[...]

PR #5895, opened 2026-02-05
KEP-5894: node system partitions KEP

SergeyKanzhelev:
KEP: https://github.com/kubernetes/enhancements/issues/5894

### DIFF
--- a/keps/sig-node/5894-node-system-partition/README.md
+++ b/keps/sig-node/5894-node-system-partition/README.md
@@ -1,0 +1,736 @@
+# KEP-5894: Node System Partition
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1: System Daemon Isolation](#story-1-system-daemon-isolation)
+    - [Story 2: HPC Workloads](#story-2-hpc-workloads)
+    - [Story 3: Telco Low-Latency Workloads](#story-3-telco-low-latency-workloads)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Cgroup Hierarchy](#cgroup-hierarchy)
+    - [Creation and Hosting](#creation-and-hosting)
+    - [Resource Limiting](#resource-limiting)
+  - [Configuration](#configuration)
+    - [Relationship to existing kubelet resource reservation](#relationship-to-existing-kubelet-resource-reservation)
+  - [Eviction](#eviction)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative cgroup hierarchies](#alternative-cgroup-hierarchies)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [X] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Node System Partition introduces a dedicated partition on a Node for
+system Pods (e.g., kube-system workloads), isolating them from user
+workloads. The system partition has its own cgroup hierarchy with
+dedicated CPU set and memory limits, ensuring system Pods cannot
+interfere with user Pods and vice versa.
+
+There are a few DIY solutions for system daemon isolation, but this
+KEP is needed because enforcing memory limits requires a separate
+cgroup hierarchy, and integrating that with kubelet functions like
+metrics collection and eviction is impossible to implement as a
+plugin.
+
+This KEP is scoped to a single system partition. Supporting arbitrary
+user-defined partitions is a non-goal.
+
+## Motivation
+
+Isolating system daemonsets from user workloads is a longstanding
+problem with numerous DIY solutions and no obvious winner. Today,
+system Pods and user Pods share the same resource boundaries — system
+Pods can burst into user resources and vice versa. This makes it
+impossible to guarantee that critical system components have the
+resources they need, or that user workloads are free from system
+interference.
+
+This problem is increasingly important as Kubernetes targets new
+workload types:
+
+1. **Traditional workloads**: Benefit from overcommit, but need basic
+   separation so a misbehaving system daemon doesn’t destabilize user
+   Pods.
+2. **HPC workloads**: Require minimal system interference and strict
+   resource isolation. These workloads need system components
+   constrained to a small, bounded resource footprint.
+3. **Telco workloads**: Optimized for low-latency and especially
+   sensitive to interference from CPU-stealing workloads. A stronger
+   isolation boundary between system and user workloads minimizes
+   noise and is critical to minimize latency.
+4. **AI/ML workloads**: Use specialized devices and need a responsive
+   management layer that is sandboxed and guaranteed its own
+   resources, without competing with the user workload.
+
+A dedicated system partition solves these problems by giving system
+Pods their own resource-limited cgroup hierarchy, eliminating
+interference between the management layer and user workloads.
+
+### Goals
+
+Alpha stage:
+
+- Introduce a system partition with a dedicated cgroup hierarchy for
+  system Pods (e.g., kube-system namespace).
+- Support memory limiting via the system partition's cgroup root.
+- Support setting a dedicated CPU set for system partition Pods.
+- Kubelet treats system and default partitions independently for
+  resource allocation and overcommit logic.
+- System partition is statically defined via kubelet configuration.
+- System partition shares resources with kubelet, container runtime,
+  and other host processes.
+
+After the first alpha:
+
+- Scheduling integration to target Pods to the system partition.
+- Additional resource isolation between system and default partitions.
+
+### Non-Goals
+
+- Supporting multiple arbitrary partitions. This KEP is scoped to a single system partition only.
+- Implementing this isolation purely via external plugins (NRI, DRA) without kubelet changes, as metrics collection and eviction require deep integration.
+- Partitioning DRA resources. DRA resources are not partitioned and remain available to Pods in both the system and default partitions.
+- Changing the fundamental QoS levels or resource accounting logic within a partition.
+
+## Proposal
+
+The Node System Partition introduces a system partition — a
+resource-bounded area of the Node dedicated to running system Pods
+(e.g., kube-system workloads). The system partition has dedicated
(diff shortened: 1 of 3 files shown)


keps/sig-node/5894-node-partitions/README.md
```diff
@@ -0,0 +1,948 @@
...
+How will UX be reviewed, and by whom?
+
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+- **Unexpected Evictions**: Since memory limits are enforced at the partition root, Pods that were previously able to consume unused node memory may now be evicted when the partition limit is reached. Mitigation: Clear documentation and monitoring for partition-level resource usage.
+- **Kubelet Complexity**: Adding separate cgroup hierarchies and partition-aware eviction logic increases Kubelet complexity. Mitigation: Extensive unit and integration testing of the new `container manager` logic.
+- **External tools integration**: Various external tools may be confused by the updated cgroup hierarchy. The KEP will explore potential issues and mitigations.
+
+## Design Details
```
44past4, line 429: Apart from kubelet changes described below this KEP will also require implementation of changes in kube-scheduler so that kube-scheduler will not be scheduling pods on nodes which latter on would be rejected by kubelet admission logic. In order for the kube-scheduler to be able to take into account the system and default partition sizes when scheduling pods it needs to the able to do two things:

- identify the partition in which a given pod will be running - this will require duplicating the logic  mentioned below and based on namespaces or some pod attributes both in kubelet and kube-scheduler

- identify the size of each partition - this will most likely require exposing the kubelet config mentioned below for configuration of the system partition on the Node obeject

keps/sig-node/5894-node-partitions/README.md
```diff
@@ -0,0 +1,948 @@
...
+- Allow kubelet to treat each partition as a whole node for resource allocation and overcommit logic.
+- Partition is statically defined using a kubelet config.
+- Design system partition to be able to share resources with kubelet, container runtime and other processes.
+
+After alpha:
+
+- Scheduling integration to allow target Pods to partitions.
+- More resources isolation between partitions.
+
+### Non-Goals
```
44past4, line 224: Proposal focuses on the CPU and memory resources partitioning. It does not talk about the DRA resources. As I understand those would not be partitioned and would be available for pods running in both system and default partitions. If this is the case it might be good to mention that DRA resources partitioning is a non-goal for this KEP.

lmktfy:
/retitle KEP-5894: Add draft for node partitionas KEP

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+- Supporting multiple arbitrary partitions. This KEP is scoped to a single system partition only.
+- Implementing this isolation purely via external plugins (NRI, DRA) without kubelet changes, as metrics collection and eviction require deep integration.
+- Changing the fundamental QoS levels or resource accounting logic within a partition.
+
+## Proposal
+
+The Node System Partition introduces a system partition — a
+resource-bounded area of the Node dedicated to running system Pods
+(e.g., kube-system workloads). The system partition has dedicated
```
mrunalp, line 152: It will be nice to make this namespace configurable.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+As a cluster administrator, I want to sandbox system daemonsets into a dedicated partition so that they do not interfere with user workloads and have a guaranteed amount of resources (CPU, Memory) regardless of user Pod activity.
+
+#### Story 2: HPC Workloads
+
+As an HPC user, I want to run my performance-critical applications in a partition consisting of high-performance cores with dedicated memory. This ensures my workloads have uninterrupted performance and eliminates noisy neighbor issues from system Pods.
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
```
mrunalp: Nit, drop the comments from the sections already filled.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+To enforce resource isolation at the partition level while adhering to the **Minimal Impact** principle, the Kubelet will introduce a targeted sub-hierarchy for system workloads while preserving the legacy structure for user workloads.
+
+**Existing Hierarchy:**
+`kubepods` -> `burstable` / `besteffort` / `pod<UID>`
+
+**Proposed Hierarchy:**
+`kubepods`
+├── `system` (New partition root for system workloads)
+│   ├── `burstable`
```
mrunalp: nit:  missing guaranteed in the diagram

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+│   ├── `besteffort`
+│   └── `pod<UID>`
```
mrunalp: pod should be nest one level under

mrunalp (approved):
I left some nits but this proposal looks good for alpha to solve this important issue.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+This problem is increasingly important as Kubernetes targets new
+workload types:
+
+1. **Traditional workloads**: Benefit from overcommit, but need basic
+   separation so a misbehaving system daemon doesn’t destabilize user
+   Pods.
+2. **HPC workloads**: Require minimal system interference and strict
+   resource isolation. These workloads need system components
+   constrained to a small, bounded resource footprint.
```
ffromani, line 109: telco workloads also can benefit from this split because it creates a stronger isolation boundary which is beneficial to minimize noise, critical to minimize latency.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+bogged down.
+-->
+
+#### Story 1: System Daemon Isolation
+
+As a cluster administrator, I want to sandbox system daemonsets into a dedicated partition so that they do not interfere with user workloads and have a guaranteed amount of resources (CPU, Memory) regardless of user Pod activity.
+
+#### Story 2: HPC Workloads
+
+As an HPC user, I want to run my performance-critical applications in a partition consisting of high-performance cores with dedicated memory. This ensures my workloads have uninterrupted performance and eliminates noisy neighbor issues from system Pods.
```
ffromani, line 183: you may want to mention telco workloads as well, they have a very similar set of constraints and benefit; telco workload tend to be optimized for low-latency (vs throughput) and they are especially sensitive to interference from other cpu-stealing workloads.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+The recommendation is to make Node Partitions as a dedicated CPU set as well as a separate cgroup hierarchy so the memory limit and other properties can be applied to the whole partition.
+
+### Cgroup Hierarchy
+
+To enforce resource isolation at the partition level while adhering to the **Minimal Impact** principle, the Kubelet will introduce a targeted sub-hierarchy for system workloads while preserving the legacy structure for user workloads.
+
+**Existing Hierarchy:**
+`kubepods` -> `burstable` / `besteffort` / `pod<UID>`
+
+**Proposed Hierarchy:**
```
ffromani, line 218: this renders poorly

Image

perhaps a multi-line block

```
like this
```
may help

keps/sig-node/5894-node-system-partition/README.md
ffromani: question: since we are adding a hierarchy existing tools don't know about, do we want to take the chance to make it more regular and add a `guaranteed` subdirectory?
I'm really just thinking aloud.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+│   ├── `burstable`
+│   ├── `besteffort`
+│   └── `pod<UID>`
+├── `burstable` (Legacy location, remains for non-system workloads)
+├── `besteffort` (Legacy location, remains for non-system workloads)
+└── `pod<UID>` (Legacy location, remains for non-system workloads)
+
+#### Creation and Hosting
+
+- **Pod Placement**: During Pod admission and sync, the Kubelet will determine if a Pod belongs to the `system` partition (e.g., via namespace check or explicit configuration). If it does, the sandbox and containers will be placed under the `kubepods/system` hierarchy. All other Pods will continue to use the legacy paths directly under `kubepods`.
```
ffromani, line 237: namespace check seems to be a nice compromise. Ideally we should have a pod field I reckon. Or perhaps just a namespace field.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+- **`kubeReservedCgroup` / `systemReservedCgroup`**: Enforce the
+  above reservations via cgroups. These cgroups are for host
+  processes (kubelet, containerd, sshd, etc.), not for Pods. The
+  system partition cgroup (`kubepods/system/`) is a separate
+  hierarchy under `kubepods` for Pod workloads.
+- **`reservedSystemCPUs`**: Pins specific CPUs for system use via
+  CPU Manager. The system partition's `cpuset` should typically match
+  `reservedSystemCPUs` so that system Pods and system services share
+  the same cores, keeping user workload CPUs free from system
+  interference.
```
ffromani, line 324: do we expect compatible configuration then? I think we want to add validation that the settings are compatible.
I'd also make sure that the partition works as expected if `--reserved-cpus` is not set.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+partition is configured and active.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] Other (treat as last resort)
+  - Details: Verify that system Pods (e.g., kube-system Pods) are
+    placed under the `kubepods/system/` cgroup hierarchy by
+    inspecting `/sys/fs/cgroup/kubepods/system/`. Verify memory
+    limits are set by reading `memory.max` on the system partition
+    cgroup.
```
ffromani, line 548: perhaps add gauges to report the count of pods in each partition?

ffromani (commented):
Took a review pass. I'm in general supportive of the feature and it addresses real user needs. I'd like to iterate on few design details (will post followup comments) but everything I want to doubleclick is improvement rather than blocker, so the proposal seems already good for alpha stage. Some inline comments.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+Kubelet will recreate the system partition cgroup hierarchy and, on
+the next Pod sync, move system Pods into the `kubepods/system/`
+hierarchy. This involves container restarts for affected Pods.
+
+###### Are there any tests for feature enablement/disablement?
+
+Unit tests will verify that the container manager correctly creates
+or skips the system partition cgroup hierarchy based on the feature
+gate state. Node e2e tests will verify system Pods are placed in the
+correct cgroup location with the feature enabled and disabled.
```
wojtek-t, line 489: These are good, but the primary missing thing is the test that verifies the behavior descried above:
if system pods are running on the wrong partition, they are restarted and moved to the right one

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+This is a node-local feature with no control plane component. Rollout
+is per-node via kubelet restart. If the system partition
+configuration is invalid (e.g., references CPUs that don't exist),
+kubelet will fail to start. Running user workloads are not affected
+by enabling the feature since they remain in the default cgroup
+locations.
```
wojtek-t, line 500: Please repeat that system workloads may be restarted.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+###### What specific metrics should inform a rollback?
+
+- Unexpected Pod evictions in the system partition (system Pods being
+  OOM-killed due to undersized memory limit).
+- Kubelet restart failures.
+- System Pod startup latency increases.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Will be tested manually before alpha release. The upgrade path
```
wojtek-t: Not a blocker for Alpha, but please describe the scenario in more detail (see https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5710-workload-aware-preemption#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested as one of the examples)

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+Check if the `kubepods/system` cgroup hierarchy exists on the node.
```
wojtek-t, line 537: Again not alpha thing - but checking hierarchy is not managable if I own thousands or millions of nodes. The metric is really needed.

keps/sig-node/5894-node-system-partition/README.md
```diff
@@ -0,0 +1,805 @@
...
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `NodeSystemPartition`
+  - Components depending on the feature gate: `kubelet`
+
+###### Does enabling the feature change any default behavior?
+
+No. The feature requires both the feature gate to be enabled and
+explicit kubelet configuration defining the system partition. Without
+configuration, kubelet behaves identically to today.
```
wojtek-t, line 457: I know it's technically not default behavior, but I think the resoure budget for user pods described above needs to be reflected somewhere in the PRR section and it seems like the most suitable place.

wojtek-t (commented):
Few small comments from the PRR perspective, but overall this LGTM.

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: it is, see later:

```
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
systemPartition:
  memoryLimit: "4Gi"
  cpuset: "0-3"
  namespaces:
    - kube-system
```

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: fixed

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: fixed

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: done

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: oh, never looked at render. Fixed

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: I have namespace in config. Not overcomplicating it here

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added, good idea. I think in beta I will explore which partition Pod is in, but for now added a metric

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: done

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: done

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added a count metric. I think for beta I will explore Pod status.

keps/sig-node/5894-node-system-partition/README.md
SergeyKanzhelev: added a new note, makes sense!

PR #5911, opened 2026-02-11
Update Topology-aware workload scheduling KEP with Workload alpha2 API

44past4:
- One-line PR description: Update Topology-aware workload scheduling KEP with Workload alpha2 API changes

- Issue link: #5732

- Other comments: See other KEPs

  - #5893
  - #5833

/sig scheduling

### DIFF
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -160,7 +160,7 @@ need for external dependencies.
 - To define the required changes to the Workload API (KEP-4671) to support
   ResourceClaims for DRA-aware workload scheduling. These changes will be
   proposed in a separate KEP:
-  [KEP-5729: DRA: ResourceClaim Support for Workloads](https://github.com/kubernetes/enhancements/pull/5736)
+  [KEP-5729: DRA: ResourceClaim Support for Workloads](https://kep.k8s.io/5729)
 - To replace the functionality of external workload queueing and admission
   control systems like Kueue. This proposal focuses on the in-scheduler
   placement decision for a single Workload at a time.
@@ -235,11 +235,11 @@ workload's pods to them.
 
 The Workload API (KEP-4671) will be extended to allow specifying group-level
 scheduling constraints. An optional `ScheduleConstraints` field is added to the
-`PodGroup` spec.
+`PodGroupTemplate` spec.
 
 ```go
 // PodGroup (definition from KEP-4671, with additions)
-type PodGroup struct {
+type PodGroupTemplate struct {
     Name *string
 
     // SchedulingConstraints defines group-level scheduling requirements,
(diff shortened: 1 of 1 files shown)


keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -352,19 +352,11 @@ Placement represents a candidate domain (nodes and resources) for a PodGroup.
...
-    WorkloadRef *workloadv1alpha1.Workload
-
-    // PodGroupName is the name of the PodGroup.
-    PodGroupName string
-
-    // PodGroupReplicaIndex is the index of the PodGroup replica, as defined in KEP-4671.
-    // This is relevant for PodGroups that have more than one replica.
-    PodGroupReplicaIndex int
+    // SchedulingGroup is a reference to the parent PodGroup object.
+    SchedulingGroup *workloadv1alpha2.PodSchedulingGroup
```
wojtek-t: Something doesn't work for me (and I'm not entirely sure what is the purpose of this field):

1) Do we want to link to the PodGroup itself (so that we e.g. have the exact policy "in hand")?
 Then it should be:  schedulingv1alpha2.PodGroup
  [BTW, we should probably reference the internal type representation rather than v1alpha2]

2) Or do we just want a reference to it - and then it should be corev1.PodSchedulingGroup

I think it's the former, but want to confirm.

wojtek-t (commented):
One clarifying comment, other than that it LGTM.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -235,11 +235,11 @@ workload's pods to them.
 
 The Workload API (KEP-4671) will be extended to allow specifying group-level
 scheduling constraints. An optional `ScheduleConstraints` field is added to the
-`PodGroup` spec.
+`PodGroupTemplate` spec.
 
 ```go
 // PodGroup (definition from KEP-4671, with additions)
-type PodGroup struct {
+type PodGroupTemplate struct {
```
dom4ha, line 242: We need to add SchedulingConstraints to PodGroupSpec as well

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -235,11 +235,11 @@ workload's pods to them.
...
     Name *string
```
dom4ha: Is this Name needed?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Actually this is the latter. I will fix the type used.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have extended the PR to update PodGroupSpec as well.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: This is just a placeholder for existing fields. I have clarified this.

wojtek-t:
For posterity - still LGTM :)

Issue #5917, opened 2026-02-12
Authenticator Provided Authorization Constraints


everettraven:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Adds a way for authenticators to specify constraints that should be enforced during authorization evaluation.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/5918
- Discussion Link:
    - Jan. 28th 2026 SIG auth bi-weekly: [agenda/notes](https://docs.google.com/document/d/1woLGRoONE3EBVx-wTb4pvp4CI7tmLZ6lS26VTbosLKM/edit?tab=t.0#heading=h.nbcsz8v4kqpq)
    - Jan. 14th 2026 SIG Auth bi-weekly: [agenda/notes](https://docs.google.com/document/d/1woLGRoONE3EBVx-wTb4pvp4CI7tmLZ6lS26VTbosLKM/edit?tab=t.0#heading=h.9fugbwx4ukvh)
    - Dec. 3rd 2025 SIG Auth bi-weekly: [agenda/notes](https://docs.google.com/document/d/1woLGRoONE3EBVx-wTb4pvp4CI7tmLZ6lS26VTbosLKM/edit?tab=t.0#heading=h.88wzfywxgod1)
- PRs by stage and milestone:
  - [x] Alpha - v1.xx (TBD)
    - ❓ we need to figure out what we want to achieve with this KEP in 1.38
    - [x] KEP (`k/enhancements`) update PR(s):
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

kannon92:
@everettraven 
Please update the description to reflect latest changes.

AFAIK you have a [KEP](https://github.com/kubernetes/enhancements/pull/5918). Please add that to description for alpha to show this work is ongoing.

everettraven:
@enj Since discussions on this aren't going to progress further during the 1.37 time-frame, I've been asked that we remove the `lead-opted-in` label. Would you mind doing so?

whtssub:
@everettraven Enhancements team here, since as I perceive from the above comment this KEP is not going forward in the v1.37 release cycle, can you please confirm if this KEP can be removed from the milestone?

everettraven:
@whtssub Yes, it can be removed from the milestone.

PR #5918, opened 2026-02-12
KEP-5917: Add KEP for Authenticator Provided Authorization Constraints

everettraven:
- One-line PR description: Adds KEP for adding support for authenticator provided authorization constraints

- Issue link: #5917

### DIFF
--- a/keps/sig-auth/5917-authenticator-conditions/README.md
+++ b/keps/sig-auth/5917-authenticator-conditions/README.md
@@ -1,0 +1,751 @@
+# KEP-5917: Authenticator Provided Authorization Constraints
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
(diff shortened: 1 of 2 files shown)


keps/sig-auth/5917-authenticator-conditions/README.md
```diff
@@ -0,0 +1,751 @@
...
+
+- Allow authenticators, both in-tree and out-of-tree, to specify constraints to be evaluated as part of  authorization decisions.
+
+### Non-Goals
+
+- Introducing a new in-tree token format that makes use of authentication constraints.
+
+## Proposal
+
+- Add a new `user.Info` field, `constraints` that contains a list of constraints, provided by the authenticator to be enforced during authorization.
```
liggitt, line 119: let's namespace any user info extra keys we add for kubernetes, xref other examples:

```
	IssuedCredentialIDAuditAnnotationKey = "authentication.kubernetes.io/issued-credential-id"
	// PodNameKey is the key used in a user's "extra" to specify the pod name of
	// the authenticating request.
	PodNameKey = "authentication.kubernetes.io/pod-name"
	// PodUIDKey is the key used in a user's "extra" to specify the pod UID of
	// the authenticating request.
	PodUIDKey = "authentication.kubernetes.io/pod-uid"
	// NodeNameKey is the key used in a user's "extra" to specify the node name of
	// the authenticating request.
	NodeNameKey = "authentication.kubernetes.io/node-name"
	// NodeUIDKey is the key used in a user's "extra" to specify the node UID of
	// the authenticating request.
	NodeUIDKey = "authentication.kubernetes.io/node-uid"
```

I think this was actually namespaced later on, just reflect that up here as well

liggitt, line 119:
```suggestion
- Add a new `user.Info` extra key, `...kubernetes.io/constraints` that contains a list of constraints, provided by the authenticator to be enforced during authorization. 
```

ahmedtd, line 119: I think we should build on top of KEP 5681 as much as possible:
* Same condition representation
* Same evaluation model: an authorizer can return conditions, which are then evaluating during the validating admission stage?

Basically, copy the design of KEP 5681, but sub in authenticators instead of authorizers.

If we wanted to think through a concrete case for supporting DPoP from the JWT or service account token authenticators:

1) The authenticator sees that the input JWT carries a cnf.jkt claim.
2) The authenticator gets the DPoP JWT from the DPoP header.
3) The authenticator validates the DPoP JWT against the cnf.jkt claim.  If the DPoP JWT does not come from the correct key, the request is rejected.
4) The authenticator checks any statically-checkable properties of the DPoP JWT, like expiration time.  If they fail, the request is rejected.
5) The authenticator returns an authentication response, with CEL conditions on the HTTP method and request path.
6) The CEL conditions are evaluated by the conditional authorization machinery during validating admission control.  We will need to make sure that the HTTP method and request path are available in the CEL evaluation context.

keps/sig-auth/5917-authenticator-conditions/README.md
```diff
@@ -0,0 +1,751 @@
...
+- Allow authenticators, both in-tree and out-of-tree, to specify constraints to be evaluated as part of  authorization decisions.
+
+### Non-Goals
+
+- Introducing a new in-tree token format that makes use of authentication constraints.
+
+## Proposal
+
+- Add a new `user.Info` field, `constraints` that contains a list of constraints, provided by the authenticator to be enforced during authorization. 
+- Add a new built-in authorizer, added first in the authorizer chain, that evaluates authenticator constraints.
```
ahmedtd, line 120: I think we should do this during validating admission control -- there is no reason a credential like a biscuit, macaron, or future JWT equivalent could not encode restrictions on the request body as well.

"This derived token is only allowed to create Pods that are labeled with foo=bar" or similar.

everettraven: >"This derived token is only allowed to create Pods that are labeled with foo=bar" or similar.

I agree that this specific statement would be best handled by a conditional authorization check that happens at admission time, but what about "This derived token is only allowed to get Pods in Namespace 'foo'"? 

Why should that operation be deferred all the way to validating admission control if we have the information available to use earlier in the request lifecycle to deny it?

keps/sig-auth/5917-authenticator-conditions/README.md
everettraven: >Same condition representation

I'm not strongly opinionated on the condition representation format, so if folks strongly prefer CEL expressions as that representation, I'm OK with that. It feels a bit heavy to need to compile CEL expressions and then run them on the fly per request though, so that is why I went with a more RBAC rule-style format for the simple path of something like "This token can only be used to get Pods in Namespace foo".

>Same evaluation model: an authorizer can return conditions, which are then evaluating during the validating admission stage?

I think I see two paths here:
1. If we are able to concretely enforce the conditions during the authorization stage we should short-circuit the rest of the request flow
2. If we are not able to concretely enforce the conditions during the authorization stage, we can defer evaluation of those conditions to the conditional authorization logic of KEP 5681

ahmedtd: > so if folks strongly prefer CEL expressions as that representation

Specifically, I think we should build on exactly the same machinery that's already been / being built for KEP 5681.  So that has a CEL (potentially in pre-parsed binary format) for evaluation within kube-apiserver, plus opaque formats that are only understood by the specific authenticator that issued them.

Let's see what others think as well, but I would be wary of introducing a third representation for conditions.

> If we are able to concretely enforce the conditions during the authorization stage we should short-circuit the rest of the request flow

I think the gain from this is likely to be small.

There's value in rejecting during authentication (to prevent someone from DoS-ing kube-apiserver with a completely invalid credential), and there's value in rejecting during validating admission control (that's when we have the maximum amount of information about the request available as input to conditions).

Rejecting during authorization saves some compute, but only if someone is spamming you with a valid but downscoped credential that is invalid for the verb/resource they are trying to act upon.

everettraven: What about non-resource URLs and resource requests that do not go through admission (like a get/list)?

kannon92:
If you want this to merge in 1.37, please request a PRR review. You can add a 5917.yaml file to sig-auth in prod-readiness folder.

keps/sig-auth/5917-authenticator-conditions/README.md
liggitt, line 119: In the places we persist/propagate/read user.Info fields (TokenReview status, API objects that store user.Info like CSR, etc), what are the implications of dropping or being unaware of this new field?

If a new field in user info is attenuating, dropping it expands the permissions, right? How do we make that safe for existing TokenReview / CSR / ... readers?

liggitt: Above the KEP proposed a new field (which it sounded like @enj was in favor of, but which brings skew / field-dropping complications). Here describes adding a well-known extra key. That's a detail to hammer out

Issue #5925, opened 2026-02-19
Enhanced Storage Capacity Tracking


bachmanity1:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Enhance storage capacity tracking to support pods with multiple PVCs and enable storage-aware pod rescheduling.
- Kubernetes Enhancement Proposal:  https://github.com/kubernetes/enhancements/pull/5926
- Discussion Link: 
- PRs by stage and milestone:
  - [ ] Alpha - v1.xx
    - [ ] KEP (`k/enhancements`) update PR(s):
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

bachmanity1:
Hi @xing-yang. Is there any plan to review this, or should I close it?

xing-yang:
I'll take a look.

PR #5926, opened 2026-02-20
KEP-5925: enhanced storage capacity tracking

bachmanity1:
- One-line PR description: Enhance storage capacity tracking to support pods with multiple PVCs and enable storage-aware pod rescheduling.

- Issue link: https://github.com/kubernetes/enhancements/issues/5925

- Other comments:

### DIFF
--- a/keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
+++ b/keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
@@ -1,0 +1,796 @@
+# KEP-5925: Enhanced Storage Capacity Tracking
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Multiple homogeneous storage pools and joint PVC evaluation](#story-1-multiple-homogeneous-storage-pools-and-joint-pvc-evaluation)
+    - [Story 2: Stale capacity data and overprovisioning](#story-2-stale-capacity-data-and-overprovisioning)
+    - [Story 3: Node drain with local storage](#story-3-node-drain-with-local-storage)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+- [Design Details](#design-details)
+  - [Enhancement 1: Multi-Pool Capacity Reporting](#enhancement-1-multi-pool-capacity-reporting)
+    - [CSI Spec Changes](#csi-spec-changes)
+    - [Kubernetes API Changes](#kubernetes-api-changes)
+    - [External-Provisioner Changes](#external-provisioner-changes)
+    - [Scheduler Changes](#scheduler-changes)
+  - [Enhancement 2: Capacity Reservation](#enhancement-2-capacity-reservation)
+    - [Data Structure](#data-structure)
+    - [Filter Phase](#filter-phase)
+    - [Reserve Phase](#reserve-phase)
+    - [Unreserve](#unreserve)
+    - [Interaction with Enhancement 1](#interaction-with-enhancement-1)
+  - [Enhancement 3: Capacity-Aware Rescheduling](#enhancement-3-capacity-aware-rescheduling)
+    - [Kubernetes API Changes](#kubernetes-api-changes-1)
+    - [CSI Driver Contract](#csi-driver-contract)
+    - [Scheduler Changes: checkBoundClaims](#scheduler-changes-checkboundclaims)
+    - [Scheduler Changes: Filter Phase](#scheduler-changes-filter-phase)
+    - [Scheduler Changes: PreBind Phase](#scheduler-changes-prebind-phase)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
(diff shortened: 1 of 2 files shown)


bachmanity1:
Hi @xing-yang @msau42 @pohly @jsafrane, could you please review this?

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
```diff
@@ -0,0 +1,796 @@
...
+The pod is scheduled on this node, but there is only one disk — the first PVC provisions successfully but the second
+fails.
+
+**Problem B — inaccurate aggregate capacity:** I have another node with three identical 100GB NVMe disks, each managed
+as a separate storage pool by the CSI driver. All three disks are the same type and same type. If the node reports
+`Capacity: 300Gi` (sum of all disks), a pod requesting a single 120Gi PVC passes the filter. But no individual disk can
+hold 120Gi, so provisioning fails.
+
+**Solution:** With multi-pool capacity reporting and joint PVC evaluation, the single-disk node reports
+`AvailableCapacities: [100Gi]` and the three-disk node reports `AvailableCapacities: [100Gi, 100Gi, 100Gi]`. The
```
xing-yang, line 177: In this example, all 3 disks report identical capacity 100Gi. Does this proposal support different capacities reported by different disks?

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
```diff
@@ -0,0 +1,796 @@
...
+directed to node-A. The first 5 pods provision successfully, but pods 6–10 fail because only 100GB was available total.
+With capacity reservation, the scheduler marks the `CSIStorageCapacity` object as reserved after scheduling the first
+pod. Subsequent pods see no available capacity on that object and are directed to other nodes or wait until the object
+is refreshed.
+
+#### Story 3: Node drain with local storage
+
+As a platform engineer running StatefulSet workloads with local storage on a CSI driver that supports volume rebuilding
+(e.g., Longhorn), I need to drain node-A for maintenance. I run `kubectl drain node-A`. The node becomes unschedulable
+and the pods are evicted. The CSI driver detects the node is unschedulable and unsets PV nodeAffinity for volumes on
```
xing-yang, line 200: CSI driver does not modify PV directly.

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
```diff
@@ -0,0 +1,796 @@
...
+that node (via KEP-5381). The StatefulSet controller recreates the pods. The scheduler processes each bound PVC: it sees
+the CSI driver supports volume rebuilding and the original node is unschedulable, so it reclassifies the PVC as a
+delayed-binding unbound PVC and checks `CSIStorageCapacity` to find a node with enough space. After selecting a new
```
xing-yang, line 203: What happens to the original PV after nodeAffinity is unset?  Who unbinds PVC and PV?  Is the original PV deleted?

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
```diff
@@ -0,0 +1,796 @@
...
+    // against the pool list rather than independently.
+    //
+    // This field is OPTIONAL. If the SP does not report per-pool
+    // capacity, this field MUST be empty, and the CO falls back to
+    // available_capacity.
+    //
+    // Example: a node with three 100GiB SSDs, each a separate pool,
+    // would report: available_capacities: [107374182400,
+    // 107374182400, 107374182400]
+    repeated int64 available_capacities = 4;
```
xing-yang, line 252: Are the sizes of all disks identical?

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
bachmanity1: Yes, it is possible for capacities to be different. I'll update the example to make this explicit.

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
bachmanity1: No, sizes can be different. I'll update the example to make this explicit.

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
bachmanity1: I meant to say that the underlying storage backend detects the node is unschedulable and unsets PV nodeAffinity.

keps/sig-storage/5925-enhanced-storage-capacity-tracking/README.md
bachmanity1: No, PV is not deleted. The only change made to PV is that its nodeAffinity is unset. 

> Who unbinds PVC and PV?

There is no actual **unbind** operation between a PVC and PV, since Kubernetes does not support unbinding an already bound PVC from its PV.

Instead, the volume binding plugin in kube-scheduler treats the PVC as effectively unbound when the originally selected node becomes unschedulable and the CSI driver supports volume rebuilding. This allows `CSIStorageCapacity` information to be considered again during node selection.

xing-yang: You meant the PVC/PV will stay in Bound state during the reschduling?  So the original volume is not really deleted. If it is local storage, it is bound to the node. If the node goes away, the storage on that node also goes away. When the workload moves to a different node, a new volume needs to be created on that node. I think the node affinity on local PV can't be changed because the volume is bound to the node. The mutable PV node affinity feature is for PV whose topology can be changed.

bachmanity1: The original volume backing the PV/PVC can be moved to another node. It follows the workload pod, so when the pod is rescheduled to a new node, the volume is asynchronously rebuilt there and the old volume is subsequently removed.

The problem is that storage capacity tracking is not considered when a PVC is already bound. As a result, a pod may be moved onto a node that does not have enough available storage to rebuild the volume.

This proposal enables kube-scheduler to take a pod's storage requirements into account even when its PVC is already bound, ensuring that the selected node has sufficient capacity for volume rebuilding.

xing-yang: So this is different from what we usually considered a local storage. Typically the nodeAffinity does not change for a local PV.

If the volume is asynchronously rebuilt on a new node, does the volume handle stay the same?  I assume the PVC is not attached when the rebuild is happening?  I see that it requires a CSI driver opt-in by setting CSIDriver.spec.volumeRebuilding: true and also the driver will be watching the node.  Do you have a CSI driver that has implemented this as a POC?

PR #5929, opened 2026-02-23
KEP-5491: update semantics of distinctAttribute with pairwise disjoint instead of empty intersection

everpeace:
- One-line PR description: update semantics of distinctAttribute with pairwise disjoint instead of empty intersection

- Issue link: https://github.com/kubernetes/enhancements/issues/5491

- Other comments:

### DIFF
--- a/keps/sig-scheduling/5491-dra-list-types-for-attributes/README.md
+++ b/keps/sig-scheduling/5491-dra-list-types-for-attributes/README.md
@@ -213,7 +213,7 @@ know that this has succeeded?
 - Support typed-list in device attribute values.
 - Extends(redefine) the semantics of `ResourceClaim`'s `constraints[].{matchAttribute,distinctAttribute}` fields as below so that it can work with list-type attribute values
   - `matchAttribute`: it is defined as non-empty intersection
-  - `distinctAttribute`: it is defined as empty intersection
+  - `distinctAttribute`: it is defined as pairwise disjoint
   - note: scalar values are treated as single-element lists
 - Keep _monotonicity_ in constraint.
   - Currently `Allocator`'s algorithm assumes [_monotonic_ constraints](https://github.com/kubernetes/kubernetes/blob/v1.34.2/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/experimental/allocator_experimental.go#L274-L276) only. Monotonic means that once a constraint returns false, adding more devices will never cause it to return true. This allows to bound the computational complexity for searching device combinations which satisfies the specified constraints. This KEP focuses to keep monotonicity of `matchAttribute/distinctAttribute` semantics.
@@ -252,7 +252,7 @@ The proposal has mainly two parts:
     - This KEP: it matches when the intersection (as a set) of all the list values among candidate devices is non-empty(i.e. `(∩ v_k != ∅)`)
   - For `DistinctAttribute`
     - Previously: it matches when all the attribute values among candidate devices are distinct (i.e. `∀i,j, s.t. i != j, v_i != v_j`)
-    - This KEP: it matches when the intersection (as a set) of all the list values among candidate devices is empty (i.e. `∩ v_k = ∅`)
+    - This KEP: it matches when all the list values among candidate devices are pairwise disjoint (i.e. `∀i,j, s.t. i != j, v_j ∩ v_k = ∅`)
 
 ### API Changes
 


keps/sig-scheduling/5491-dra-list-types-for-attributes/README.md
```diff
@@ -213,7 +213,7 @@ know that this has succeeded?
 - Support typed-list in device attribute values.
 - Extends(redefine) the semantics of `ResourceClaim`'s `constraints[].{matchAttribute,distinctAttribute}` fields as below so that it can work with list-type attribute values
   - `matchAttribute`: it is defined as non-empty intersection
-  - `distinctAttribute`: it is defined as empty intersection
+  - `distinctAttribute`: it is defined as pairwise disjoint
```
everpeace, line 216: I noticed that the empty intersection property violates [the definition of _monotonicity_](https://github.com/kubernetes/kubernetes/blob/v1.35.0/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/experimental/allocator_experimental.go#L265-L267):
> Constraints are assumed to be monotonic: once a constraint returns false, adding more devices will not cause it to return true.

Like this:
- `empty_intersection({{1, 2}, {2,3}}) ==> false`
- But, `empty_intersection({{1, 2}, {2,3}, {4,5}}) ==> true`

_note: the current definition of monotonicity could say "monotinically decreasing", but the empty intersection could say "monotonically increasing"._

But, pairwise disjoint (`∀i,j, s.t. i != j, v_j ∩ v_k = ∅`) is _monotonic_. Thus, I think pariwise disjoint could fit as a natural enhancement for this KEP.

johnbelamaric: Oh, good catch. Yes, I think this makes more sense. The use cases for distinct with lists are unclear, but we can't really validate or prevent these situations. Having a clear, deterministic behavior that is monotonic is the critical thing IMO.

/lgtm

sunya-ch: @everpeace Thank you. It makes more sense to me too. 

/lgtm

everpeace:
> Once this PR has been reviewed and has the lgtm label, please assign [sanposhiho](https://github.com/sanposhiho) for approval. For more information see [the Code Review Process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process).

/assign @sanposhiho 

This PR now has enough lgtm labels. Would you mind reviewing/approving this PR??

sanposhiho:
/unassign
/assign @dom4ha 
(the approver of this KEP from sig scheduling)

keps/sig-scheduling/5491-dra-list-types-for-attributes/README.md
dom4ha: Doesn't it mean we need to compare each pair separately going toward square complexity with each new device?

everpeace: DRA's [`consraint`](https://github.com/kubernetes/kubernetes/blob/v1.35.0/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/experimental/allocator_experimental.go#L732-L741) interface assumes devices to be added one by one from [`Allocator`](https://github.com/kubernetes/kubernetes/blob/v1.35.0/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/types.go#L39-L44). So, each `add` complexity is `O(k)` where `k` is the number of devices in Resource Request and bounded with [`k<=32` in API](https://github.com/kubernetes/kubernetes/blob/v1.35.1/staging/src/k8s.io/api/resource/v1/types.go#L773-L783) (not the number of devices in ResoruceSlices). So, the worst complexity could be `O(k^2)`. But, this constraint is monotonic. So, counter examples (device combinations that have non-empty intersection among specified attribute) can give-up early for searching, that can reduce total search complexity in the Allocator.


This KEP just extends the non-equality of scalars to the disjointness of sets. Inherent algorithm won't change at all. The original implementation(sclar version) is [here](https://github.com/kubernetes/kubernetes/blob/v1.35.0/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/experimental/constraint.go#L95-L131). And, my working implementation(set version) is [here](https://github.com/everpeace/kubernetes/blob/b656a1023e79e295ba7e5945dff1b469d89e7ef4/staging/src/k8s.io/dynamic-resource-allocation/structured/internal/experimental/constraint.go#L97-L125).


Thus, although this KEP actually increases the cost of evaluating `distinctAttribute` constraint (from scalar non-equality to set disjointness), because each size of set(list attribute) is [limited by API (max size is 64)](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5491-dra-list-types-for-attributes/README.md#deviceattribute), I think this can be admitted.

And, please note that the disctussion is the worst case. I believe the typical/practical size of device request(`k`) could be really small (at most 5 or so).

everpeace: The implementation PR is ready if this semantics update(empty intersection --> pairwise disjoint) is ok.

https://github.com/kubernetes/kubernetes/pull/137190

dom4ha:
/approve

@ania-borowiec @macsko

PR #5268, opened 2025-05-02
Graduate `external.metrics.k8s.io` to stable

rexagod:
- One-line PR description: "external.metrics.k8s.io" has been in beta longer than we'd hoped it to be, but this gradutes it to stable as it's been used extensively in production environments in the ecosystem.

- Issue link: https://github.com/kubernetes/enhancements/issues/5271

PR #5931, opened 2026-02-23
KEP-5207: metrics.k8s.io API definition

tico88612:
- One-line PR description: Graduate metrics.k8s.io to Stable

- Issue link: #5207 

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-instrumentation/5207.yaml
+++ b/keps/prod-readiness/sig-instrumentation/5207.yaml
@@ -1,0 +1,3 @@
+kep-number: 5207
+stable:
+  approver: "@soltysh"
--- a/keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
+++ b/keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
@@ -71,6 +71,8 @@ Given the API's long track record and unchanged schema, graduating to stable is 
 
 - Graduate `metrics.k8s.io` to stable, by introducing the same API under
   `v1` as exists currently under `v1beta1`.
+- Migrate the in-tree consumers (e.g., `pkg/controller/podautoscaler`,
+  `kubectl top`) to the `v1` API.
 - The graduation (and thus the migration to the newer version) should be
   non-breaking.
 
@@ -92,7 +94,7 @@ The proposal broadly entails two efforts:
 - Introducing `metrics.k8s.io/v1` for the `k8s.io/metrics` package, and,
 - Migrating in-house components and sub-projects to utilize the stable version.
 
-The former will more or less be a cosmetic change, essentially bumping
+The former will be a cosmetic change, essentially bumping
 `v1beta1` to `v1`. The `v1` API surface will be identical to `v1beta1`,
 with the only difference being the version name itself.
 
@@ -105,13 +107,13 @@ The latter will focus on utilizing the stable version in the
 For the purpose of explaining the risks, we'll assume two entities here.
 
 First, a requester (e.g., HPAs), which utilizes the `metrics.k8s.io`
-client, and second, a Metric Server which
+client, and second, a server (for e.g., [metrics-server](https://sigs.k8s.io/metrics-server)) which
 utilizes the type definitions exposed by
 `k8s.io/metrics/pkg/apis/metrics` to wrap over and expose the metrics
 received from the metrics backend in a manner that is consumable by the
 requester. 
 
-The requester and the Metric Server are linked through an `APIService` object, which
+The requester and the server are linked through an `APIService` object, which
 allows the latter to act as addon APIServer through the aggregation layer for
 all requests from the requester towards that `GroupVersion`. Note that while the
 APIServer will promote or fallback between versions that support conversions,
@@ -121,7 +123,7 @@ Both `v1.metrics.k8s.io` and `v1beta1.metrics.k8s.io` APIServices can be
 registered at the same time. A request is served only if an APIService is
 registered and available for the exact version requested; otherwise that
 request gets `404 NotFound`, while requests to other registered versions are
-unaffected. The metric server's internal type-definition version has no
+unaffected. The server's internal type-definition version has no
 effect, since the `v1` and `v1beta1` surfaces are identical.
 
 | Registered APIServices | `v1` client request | `v1beta1` client request |
@@ -177,7 +179,30 @@ spec:
   version: v1beta1
 ```
 
-* The implementation collects resource metrics from each kubelet and returns `NodeMetrics` or `PodMetrics` objects. `PodMetrics` includes per-container breakdown via `ContainerMetrics`.
+* The implementation collects resource metrics from each kubelet and returns `NodeMetrics` or `PodMetrics` objects. `PodMetrics` includes per-container breakdown via `ContainerMetrics`. The `v1` types are identical to `v1beta1` except for the version name:
+
+```go
+type NodeMetrics struct {
+    metav1.TypeMeta
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    Timestamp metav1.Time     `json:"timestamp"`
+    Window    metav1.Duration `json:"window"`
+    Usage     v1.ResourceList `json:"usage"`
+}
+
+type PodMetrics struct {
+    metav1.TypeMeta
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+    Timestamp  metav1.Time        `json:"timestamp"`
+    Window     metav1.Duration    `json:"window"`
+    Containers []ContainerMetrics `json:"containers"`
+}
+
+type ContainerMetrics struct {
+    Name  string          `json:"name"`
+    Usage v1.ResourceList `json:"usage"`
+}
+```
 
 ### Test Plan
 
@@ -230,11 +255,16 @@ lives out of tree.
 - HPA e2e tests
   ([`test/e2e/autoscaling/horizontal_pod_autoscaling*.go`](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/autoscaling))
   exercise the `metrics.k8s.io` API end-to-end through metrics-server in e2e
-  clusters: [sig-autoscaling-hpa testgrid](https://testgrid.k8s.io/sig-autoscaling-hpa).
+  clusters: [sig-autoscaling-hpa-periodics testgrid](https://testgrid.k8s.io/sig-autoscaling-hpa-periodics).
 - metrics-server's own e2e suite
   ([`test/e2e_test.go`](https://github.com/kubernetes-sigs/metrics-server/blob/master/test/e2e_test.go))
   directly verifies serving the API on a kind cluster, and will be extended
   to cover the `v1` APIService.
+- A `kubectl top node` / `kubectl top pod` e2e test is planned, to be added
+  under `test/e2e/kubectl`. The `--provider=gce` jobs that already run the HPA
+  e2e tests ship metrics-server (and the `metrics.k8s.io` APIService) by
+  default, so the path is viable; the exact placement will be coordinated with
+  SIG CLI, which owns `kubectl top`.
 
 ### Graduation Criteria
 
@@ -386,7 +416,7 @@ N/A
   ```
   kubectl get apiservice v1.metrics.k8s.io
   ```
-- [x] Other (treat as last resort)
+- [x] Kubectl
   - Details: Registered versions can be listed by querying `/apis`, and the
     API can be queried directly to confirm it serves data:
   ```
@@ -431,7 +461,7 @@ No.
 
 Yes. Serving the `metrics.k8s.io` API requires an implementation running in the
 cluster — most commonly
-[metrics-server](https://github.com/kubernetes-sigs/metrics-server) — that
+[metrics-server](https://sigs.k8s.io/metrics-server) — that
 registers and backs the corresponding APIService. Currently, `controller-manager`
 relies on this API, and in order to establish an overall workflow, needs the
 aggregation layer to route the requests accordingly.
--- a/keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
+++ b/keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
@@ -1,0 +1,374 @@
+# KEP-5207: metrics.k8s.io API definition
+
+<!-- TOC -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /TOC -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Promote the `metrics.k8s.io` API to stable. The API has been in beta since v1.8 (approximately 7 years ago) without a formal graduation effort. Given this lengthy soak period, the API has proven stable in production environments worldwide — consumers such as the Horizontal Pod Autoscaler and `kubectl top` have relied on it for years. We believe it is now appropriate to formally graduate the API to stable.
+
+## Motivation
+
+The `metrics.k8s.io` API was introduced in v1.6 (alpha) and promoted to beta in v1.8, predating the KEP process. Since then, the API has remained unchanged and has been widely used in production by components such as the Horizontal Pod Autoscaler and `kubectl top`.
+
+However, the API has never been formally graduated to stable. Beta APIs carry an implicit instability signal to users, as they may be subject to removal per the [Kubernetes deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/). Furthermore, the [KEP-1635 prevent-permabeta](https://kep.k8s.io/1635) initiative requires that APIs either graduate or be formally deprecated.
+
+Given the API's long track record and unchanged schema, graduating to stable is the appropriate path forward.
+
+### Goals
+
+- Graduate `metrics.k8s.io` to stable, by introducing the same API under
+  `v1` as exists currently under `v1beta1`.
+- The graduation (and thus the migration to the newer version) should be
+  non-breaking.
+
+### Non-Goals
+
+- Introduce `v1` version of the API that differs from `v1beta1` in any way,
+  except naming.
+- Drop support for any of the existing `v1beta1` version(s).
+
+## Proposal
+
+The proposal broadly entails two efforts:
+
+- Introducing `metrics.k8s.io/v1` for the `k8s.io/metrics` package, and,
+- Migrating in-house components and sub-projects to utilize the stable version.
+
+The former will more or less be a cosmetic change, essentially bumping
+`v1beta1` to `v1`. The `v1` API surface will be identical to `v1beta1`,
+with the only difference being the version name itself.
+
+The latter will focus on utilizing the stable version in the
+`metrics.k8s.io`'s client, as well as `pkg/controller/podautoscaler`,
+which is currently on the latest beta version.
+
+### Risks and Mitigations
+
+For the purpose of explaining the risks, we'll assume two entities here.
+
+First, a requester (e.g., HPAs), which utilizes the `metrics.k8s.io`
+client, and second, a Metric Server which
+utilizes the type definitions exposed by
+`k8s.io/metrics/pkg/apis/metrics` to wrap over and expose the metrics
+received from the metrics backend in a manner that is consumable by the
+requester. 
+
+The requester and the Metric Server are linked through an `APIService` object, which
+allows the latter to act as addon APIServer through the aggregation layer for
+all requests from the requester towards that `GroupVersion`. Note that while the
+APIServer will promote or fallback between versions that support conversions,
+that is not the case within the aggregation layer.
+
+Now, let's consider the different scenarios that can arise from varying
+configurations of the two:
+
+| Requester's `metrics.k8s.io` client version | APIService's `GroupVersion` | Metric Server's `k8s.io/metrics/pkg/apis/metrics` type definitions' version | Working version       |
+| -------------------------------------------------- | --------------------------- | ---------------------------------------------------------------------------- | --------------------- |
+| Beta                                               | Stable                      | No effect (as both API surfaces are interchangeable)                         | 404 for client's verb |
+| Stable                                             | Beta                        | No effect (as both API surfaces are interchangeable)                         | 404 for client's verb |
+| Stable                                             | Stable                      | No effect (as both API surfaces are interchangeable)                         | Stable                |
+| Beta                                               | Beta                        | No effect (as both API surfaces are interchangeable)                         | Beta                  |
+
+Therefore, users need to ensure that the same version is being used between the
+requester and the metric-server in order for the migration to be successful.
+
+## Design Details
+
+The `metrics.k8s.io` API provides resource metrics (CPU and memory)
+for nodes and pods in the cluster, in an implementation-agnostic manner.
+As a part of the aggregation layer, implementations instrumenting the
+API benefit from APIServer's authentication and authorization measures
+out-of-the-box.
+
+The diagram below gives a general idea on the flow:
+
+```mermaid
+graph TD
+    A[Consumer<br>HPA / kubectl top] -- "GET /apis/metrics.k8s.io/v1beta1/..." --> B[API Server]
+    B -- "v1beta1.metrics.k8s.io<br>APIService" --> C[metrics-server]
+    C -- "resource metrics endpoint" --> D[kubelet]
+    D -- "raw metrics" --> C
+    C -- "v1beta1.NodeMetrics OR v1beta1.PodMetrics" --> B
+```
+
+* The consumer, through the `metrics.k8s.io` client, sends a request. The format is dictated by `k8s.io/metrics/pkg/apis/metrics`:
+
+```
+GET /apis/metrics.k8s.io/v1beta1/nodes
+GET /apis/metrics.k8s.io/v1beta1/nodes/{name}
+GET /apis/metrics.k8s.io/v1beta1/namespaces/{ns}/pods
+GET /apis/metrics.k8s.io/v1beta1/namespaces/{ns}/pods/{name}
+```
+
+* The responsible addon APIServer is identified based on the registered APIService object. For e.g., for metrics-server:
+
+```yaml
+apiVersion: apiregistration.k8s.io/v1
+kind: APIService
+metadata:
+  name: v1beta1.metrics.k8s.io
+spec:
+  group: metrics.k8s.io
+  service:
+    name: metrics-server
+    namespace: kube-system
+  version: v1beta1
+```
+
+* The implementation collects resource metrics from each kubelet and returns `NodeMetrics` or `PodMetrics` objects. `PodMetrics` includes per-container breakdown via `ContainerMetrics`.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+Since the change only encompasses the API surface which remains unchanged
+going into stable, we believe that the existing testing should suffice.
+
+### Graduation Criteria
+
+#### Alpha
+
+Initial implementation of the resource metrics API under `v1alpha1`.
+
+#### Beta
+
+Promoted to `v1beta1`. The API surface has remained stable since then.
+
+#### GA
+
+- e2e tests have been stable with no flakes for a minimum two-week window.
+- All in-tree consumers (e.g., `pkg/controller/podautoscaler`) have been
+  migrated to use the `v1` API.
+
+### Upgrade / Downgrade Strategy
+
+The primary area of focus when upgrading or downgrading is ensuring that a
+registered APIService exists for the same version that the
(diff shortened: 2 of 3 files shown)


rexagod:
@tico88612 Could you PTAL at the failing jobs here?

tico88612:
@rexagod The failing job is about the PRR Reviewer. It will succeed if we have a PRR reviewer in here.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
```
rexagod: FYI

tico88612: I've looked through a lot of documentation, and conformance tests don't apply to optional features. This is the same reason why HPA doesn't have conformance tests (since HPA is based on `metrics-server`).

https://github.com/kubernetes/kubernetes/pull/106595#issuecomment-1006015756
FYI: https://github.com/kubernetes/kubernetes/pull/107349

keps/prod-readiness/sig-instrumentation/5207.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 5207
+stable:
+  approver: "" #TODO
```
soltysh: As mentioned on slack, put my name here.

tico88612: Added, thank you!

soltysh:
@tico88612 please sync with sig-instrumentation leads to opt-in this KEP into 1.37 release (ref https://github.com/kubernetes/enhancements/issues/5207#issuecomment-4497110783) so that I can track the PRR work for this KEP.

lmktfy:
> * One-line PR description: Graduate metrics.k8s.io to Stable

Ah. Mini concern, I hope we can fix it without big disagreement.

---

Picture this in a list, eg:
> The following enhancements are currently beta:
>
> * Dynamic frobnication
> * PhippyPerNamespace
> * Graduate metrics.k8s.io to Stable
> * PodGroup API

or:
> The v1.42 release is expected to graduate the following to stable:
>
> * Dynamic frobnication
> * PhippyPerNamespace
> * Graduate metrics.k8s.io to Stable
> * PodGroup API

Could we call the enhancement "metrics.k8s.io API definition" instead? Then it fits easily in those lists, and in others.

tico88612:
> Could we call the enhancement "metrics.k8s.io API definition" instead? Then it fits easily in those lists, and in others.

Agree, GA is a stage, this shouldn't be in the KEP name. `metrics.k8s.io API definition` would be better.

tico88612:
@soltysh the `lead-opt-in` label has been added in the KEP-5207.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+The requester and the Metric Server are linked through an `APIService` object, which
+allows the latter to act as addon APIServer through the aggregation layer for
+all requests from the requester towards that `GroupVersion`. Note that while the
+APIServer will promote or fallback between versions that support conversions,
+that is not the case within the aggregation layer.
+
+Now, let's consider the different scenarios that can arise from varying
+configurations of the two:
+
+| Requester's `metrics.k8s.io` client version | APIService's `GroupVersion` | Metric Server's `k8s.io/metrics/pkg/apis/metrics` type definitions' version | Working version       |
```
stlaz: You can have both APIServices present at the same time, as long as both API versions are being served.

What do you mean by "404 for client's verb" in the table? Just 404 to any request?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
```
soltysh, line 51: Please make sure to check the appropriate boxes above.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+* The implementation collects resource metrics from each kubelet and returns `NodeMetrics` or `PodMetrics` objects. `PodMetrics` includes per-container breakdown via `ContainerMetrics`.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+Since the change only encompasses the API surface which remains unchanged
+going into stable, we believe that the existing testing should suffice.
```
soltysh, line 214: This is fine, but still, please make sure to list existing unit, integration and e2e tests as required by [the template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md).

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+and vice versa.
+
+Do note that the version used in the implementation (assuming it's one
+of `v1beta1` or `v1`) does not affect its interoperability with the version
+defined in the APIService or the requester.
+
+### Version Skew Strategy
+
+Within the scope of k/k, an upgraded `pkg/controller/podautoscaler`
+(the only affected component) using the `v1` API will fail to fetch
```
soltysh: What about `kubectl top pod` and `kubectl top node` commands? I believe those should be updated as well.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? No. 
+
+###### Does enabling the feature change any default behavior?
+
+No.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. To roll back to the previous behavior, remove the `v1.metrics.k8s.io` APIService registration from the cluster and ensure that all callers (e.g., HPA controller, `kubectl top`) use the `v1beta1` version of the API.
```
soltysh, line 338: I believe both the hpa controller and kubectl should handle fallback gracefully. This is required to support the official https://kubernetes.io/releases/version-skew-policy/.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+Both of the above are true for a subset of errors for which the controller fails
+to fetch metrics. See
+[pkg/controller/podautoscaler/horizontal.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podautoscaler/horizontal.go)
+for more information.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+The API surface is identical between `v1beta1` and `v1`. Rollback
+amounts to re-registering the `v1beta1` APIService and ensuring
+consumers use the corresponding version.
```
soltysh: Since the implementation will target v1 directly, it's hard to perform the upgrade->downgrade->upgrade testing scenario at this point in time. But ideally, you should describe how the process looks like, and what should be verified.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+N/A
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] API .status
+  - Condition name: AbleToScale
+  - Other field: `.status == True`
+- [x] API .status
+  - Condition name: ScalingActive
+  - Other field: `.reason == ValidMetricFound`
```
soltysh: Those are HPA-related conditions, not something that is related with metrics. You should update this to look at apiservices, for example.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+  kubectl get --raw /apis/metrics.k8s.io/ | jq .
+  ```
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+N/A
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+N/A
```
soltysh: It's a bit of odd situation that a metrics server does not provide any reasonable metrics (for SLI) and then SLO. Don't you have anything defined for the metrics server regarding API availability?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+Yes, `v1.metrics.k8s.io` (w.r.t. the registered APIService object).
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+By one. An APIService object needs to be created to register the newer API version.
```
soltysh: This question covers the built-in types, and since `metrics.k8s.io` is served by an external apiserver I think you can just write no here.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+Initial implementation of the resource metrics API under `v1alpha1`.
+
+#### Beta
+
+Promoted to `v1beta1`. The API surface has remained stable since then.
+
+#### GA
+
+- e2e tests have been stable with no flakes for a minimum two-week window.
+- All in-tree consumers (e.g., `pkg/controller/podautoscaler`) have been
```
stlaz: What's going to be the deprecation strategy for the beta API?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+  migrated to use the `v1` API.
+
+### Upgrade / Downgrade Strategy
+
+The primary area of focus when upgrading or downgrading is ensuring that a
+registered APIService exists for the same version that the
+`metrics.k8s.io`'s client is using, i.e., if the latter is on an older
```
stlaz: > ensuring that a registered APIService exists for the same version that the `metrics.k8s.io`'s client is using

Which metrics client is going to be driving upgrades/downgrades?

Is this meant to say "ensuring an APIService exists for the version of the `metrics.k8s.io` API that is being served?"

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+version ensure that the APIServer knows where to route requests based on that,
+and vice versa.
+
+Do note that the version used in the implementation (assuming it's one
+of `v1beta1` or `v1`) does not affect its interoperability with the version
+defined in the APIService or the requester.
```
stlaz: It might should you decide to extend the API in the future.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+The API surface is identical between `v1beta1` and `v1`. Rollback
+amounts to re-registering the `v1beta1` APIService and ensuring
+consumers use the corresponding version.
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No deprecations or removals, only the introduction of a newer API version.
```
stlaz: What about the old "stable" API version?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,374 @@
...
+### Troubleshooting
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+Initially, when establishing trust, APIServer being down impacts the ability to
+register the addon APIServer, and thus disrupts the HPA controller.
+
+###### What are other known failure modes?
+
+N/A
```
stlaz: authn/network broken between the metrics-server and kube-apiserver or kubelet, or both?

stlaz (commented):
PRR shadow review

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Yes, we should update this.

soltysh:
@tico88612 with the deadline quickly approaching, are you planning to update this document?

tico88612:
> @tico88612 with the deadline quickly approaching, are you planning to update this document? 

I will update this quickly

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: I've checked off the information I know.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Got it, I've already added it.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Agreed. Updated Version Skew Strategy: the in-tree consumers (HPA controller, `kubectl top` etc.) negotiate the served version through discovery and fall back to v1beta1 when v1 isn't available.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Replaced the HPA conditions with the `Available` condition on the `v1.metrics.k8s.io` APIService.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Right, exercising this end-to-end needs an implementation release that serves `v1`, which isn't available yet. 

Described the manual verification process (upgrade → downgrade → upgrade) with what to check at each step. The API is read-only with no persisted state, so no data migration is involved.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: `metrics.k8s.io` uses the aggregation layer; I'm writing the aggregator-related metrics here.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Done, changed to No.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Reworked the table around which APIServices are registered (both can be present at once).

Clarified that 404 NotFound is returned only for a request to a version whose APIService isn't registered — requests to other registered versions are unaffected; the aggregation layer does no version conversion.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Added to Non-Goals: v1beta1 isn't dropped as part of this graduation; it's deprecated once v1 is GA and removed in a later release following the deprecation policy (≥ 3 releases / 9 months).

I will confirm the exact removal release with SIG Instrumentation leads.

cc @rexagod @dgrisonnet @richabanker

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Reworded. The available versions are determined by which APIServices are registered/available, not by the client. Clients negotiate via discovery, so no client "drives" the upgrade/downgrade.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Reworded. 

The claim is now conditioned on the `v1` and `v1beta1` surfaces being identical.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: This rollout introduces `v1` and removes nothing, but graduating `v1` deprecates `v1beta1` (which is served until removed later per the deprecation policy). Fixed the "no deprecations" wording.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Added as known failure modes: a broken authn/network hop between kube-apiserver and the implementation makes the APIService unavailable; between the implementation and the kubelets, metrics for affected nodes/pods go missing/stale.

soltysh (approved):
/approve
the PRR section

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,504 @@
...
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+No.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+Currently, `controller-manager` relies on this API, and in order to establish an
+overall workflow, needs the aggregation layer to route the requests accordingly.
```
soltysh: You should mention metrics-server here explicitly.

tico88612: Added

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+  months before removal).
+
+## Proposal
+
+The proposal broadly entails two efforts:
+
+- Introducing `metrics.k8s.io/v1` for the `k8s.io/metrics` package, and,
+- Migrating in-house components and sub-projects to utilize the stable version.
+
+The former will more or less be a cosmetic change, essentially bumping
```
rexagod:
```suggestion
The former will be a cosmetic change, essentially bumping
```

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+  [deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/)
+  (a deprecated beta API version is supported for at least 3 releases / 9
+  months before removal).
+
+## Proposal
+
+The proposal broadly entails two efforts:
+
+- Introducing `metrics.k8s.io/v1` for the `k8s.io/metrics` package, and,
+- Migrating in-house components and sub-projects to utilize the stable version.
```
rexagod, line 95: This should be reflected in goals.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+The latter will focus on utilizing the stable version in the
+`metrics.k8s.io`'s client, as well as `pkg/controller/podautoscaler` and
+`kubectl top`, which are currently on the latest beta version.
+
+### Risks and Mitigations
+
+For the purpose of explaining the risks, we'll assume two entities here.
+
+First, a requester (e.g., HPAs), which utilizes the `metrics.k8s.io`
+client, and second, a Metric Server which
```
rexagod:
```suggestion
client, and second, a server (for e.g., <link to metrics-server>) which
```

I understand that "Metric Server" here is a neutral term, referencing any metric servers built over custom api server, and not just the one under the SIG. However, since we already talk about this exposing metrics, the "Metric" part seems redundant. Additionally, a link to k-sigs/metrics-server itself should clear the ambiguity that we are not talking about a particular reference implementation here.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+  name: v1beta1.metrics.k8s.io
+spec:
+  group: metrics.k8s.io
+  service:
+    name: metrics-server
+    namespace: kube-system
+  version: v1beta1
+```
+
+* The implementation collects resource metrics from each kubelet and returns `NodeMetrics` or `PodMetrics` objects. `PodMetrics` includes per-container breakdown via `ContainerMetrics`.
```
rexagod: Ideally, we should have snippets with these depictions here, like so: https://github.com/kubernetes/enhancements/pull/5268/changes#diff-63b54916877fc7c9d86c6ed5fcd0dffc7541cf8a9f641c6cdb8bc93be5640fa3R343-R378

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+
+- HPA e2e tests
+  ([`test/e2e/autoscaling/horizontal_pod_autoscaling*.go`](https://github.com/kubernetes/kubernetes/tree/master/test/e2e/autoscaling))
+  exercise the `metrics.k8s.io` API end-to-end through metrics-server in e2e
+  clusters: [sig-autoscaling-hpa testgrid](https://testgrid.k8s.io/sig-autoscaling-hpa).
+- metrics-server's own e2e suite
+  ([`test/e2e_test.go`](https://github.com/kubernetes-sigs/metrics-server/blob/master/test/e2e_test.go))
+  directly verifies serving the API on a kind cluster, and will be extended
+  to cover the `v1` APIService.
+
```
rexagod, line 268: + `kubectl top`, as well as any other downstream consumers within k/ or k-sigs/

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+### Graduation Criteria
+
+#### Alpha
+
+Initial implementation of the resource metrics API under `v1alpha1`.
+
+#### Beta
+
+Promoted to `v1beta1`. The API surface has remained stable since then.
```
rexagod, line 277: If possible, can we link to when these happened? For instance, https://github.com/kubernetes/metrics/blob/49995afe3ec2a625457074abe6ef3c002c40ce31/pkg/apis/custom_metrics/v1alpha1/types.go are the type definitions that surfaced in k/k for custom metrics.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+[pkg/controller/podautoscaler/horizontal.go](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/podautoscaler/horizontal.go)
+for more information.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Not yet — exercising this path end-to-end requires an implementation
+release (e.g., metrics-server) that registers and serves the `v1`
+APIService, which will only become available after the `v1` API lands in
+`k8s.io/metrics`. The path will be verified manually as part of the
+graduation, following the process below.
```
rexagod, line 373: Ideally, we should test all the encompassed behaviors here (fallback to older versions, etc.) before landing this in GA. I'd suggest specifying here that we intend to do this in ALPHA/BETA, before instrumenting dependencies to use the latest version.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+N/A
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] API .status
+  - Condition name: `Available` on the `v1.metrics.k8s.io` APIService:
+  ```
+  kubectl get apiservice v1.metrics.k8s.io
+  ```
+- [x] Other (treat as last resort)
```
rexagod:
```suggestion
- [x] Kubectl
```

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+
+N/A
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] API .status
+  - Condition name: `Available` on the `v1.metrics.k8s.io` APIService:
+  ```
+  kubectl get apiservice v1.metrics.k8s.io
+  ```
```
rexagod, line 418: Can we denote all relevant fields that are worth a look, like so https://github.com/kubernetes/enhancements/pull/5269/changes#diff-d4c5dc6b1bb2d5e1d90d25e03f003a732f7cab0a3285ec37873106da9b47f456R724-R729?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
rexagod: Nevermind, I see these links are referenced in the footer.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
```diff
@@ -0,0 +1,508 @@
...
+  - Will enabling / disabling the feature require downtime or reprovisioning
+    of a node? No. 
+
+###### Does enabling the feature change any default behavior?
+
+No.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. To roll back to the previous behavior, remove the `v1.metrics.k8s.io` APIService registration from the cluster and ensure that all callers (e.g., HPA controller, `kubectl top`) use the `v1beta1` version of the API.
```
rexagod, line 338: It's important to note that:

(A) downstream will pin to v1,
(B) native metrics do **not** have fallbacks, and the autogenerated code needs to be present for each version, and as such,
(C) "rolling back" would also entail deploying the older version of the component, which was pinned to the older API version, so it works with that.

Of course, we could implement fallbacks for external and native metric clients (custom metric client is the only one that has that), but that may be out of scope for this KEP.

rexagod (changes requested):
Thank you for driving this.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Fixed.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Fixed.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Agree, fixed.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Added.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: `kubectl top` currently has unit tests in `k8s.io/kubectl/pkg/cmd/top` but no e2e; the `metrics.k8s.io` serving path it uses is already exercised e2e via the HPA suite and metrics-server's own e2e.

Adding a dedicated `kubectl top` e2e looks feasible (check the TestGrid `sig-cli-master` e2e, it has `metrics-server`) — but since `kubectl` testing is owned by SIG CLI, I'll check this with them.

cc @soltysh as SIG CLI Lead

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: > (A) downstream will pin to v1,

If downstream needs to be pinned to v1, this will be a breaking change.
As I recall, `metrics-server` itself needs to import Kubernetes code (like `k8s.io/metrics/pkg/apis/metrics/v1beta1`). Since `v1` hasn't been merged, `metrics-server` can't be updated either.

> (B) native metrics do not have fallbacks, and the autogenerated code needs to be present for each version, and as such,

Yeah, for my current `kubectl top` testing, I'm handling `v1` and `v1beta1` separately. (It's a bit of a hassle, but not a huge deal.)

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
richabanker: Is this a blocker for the KEP to move forward in 1.37?

tico88612: However, it seems that ALPHA (v1alpha1) and BETA (v1beta1) don't have a fallback, and metrics-server currently only has v1beta1. Is there any way for us to test this part?

richabanker:
/lgtm
/approve

/hold
for other comments from @rexagod to be addressed

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: I originally included it, but PRR advised me against including the HPA information.

https://github.com/kubernetes/enhancements/pull/5931/commits/dafa479e7d2d0731513661cd4b7b18dc989adea3

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
tico88612: Fixed.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
richabanker: request for v1 falling back to v1beta1 can be tested via unit tests as well right?

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
dgrisonnet: > (A) downstream will pin to v1,

Not necessarily. The beta version will still be served for a few releases after v1 is out. During that period downstream consumers could very well have two APIService in place, one for v1 and another one for v1beta1.

> (C) "rolling back" would also entail deploying the older version of the component, which was pinned to the older API version, so it works with that.

Rolling back can be done by removing the v1 APIService

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
dgrisonnet: I wouldn't be too hard on this personally. There is no API changes between v1beta1 and v1, so we shouldn't have compatibility issues and adding a fallback test would essentially be testing aggregated api code.

Having a test in metrics-server to verify that it serves both v1 and v1beta should be enough IMO.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
rexagod: s/pin/import.

If at a given point of time we import both, there will need to be logic in the importing component that implements fallback between the two, because that trait is not defined in the mative metrics client. This makes the change breaking.

Hypothetically, if they don't support both at any given point of time, and only move to v1 once its GA, they won't have to implement a fallback on their own (since the native metrics client lacks it), and it'll be drop in, and non-breaking, in that case.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
soltysh: I'm supportive.

keps/sig-instrumentation/5207-metrics-k8s-io-api-definition/README.md
rexagod: > This makes the change breaking.

Rethinking, this might not be the case, but an explicit fallback needs to be [implemented on the client side](https://github.com/kubernetes/kubernetes/pull/139726) when it should be done on our end since the earlier versions moved fast and never needed this, which may not be the case in the future, and we'd want an implementation to live in /metrics that does the conversion for downstream.

tico88612: I agree that the coordinating API should be done by us (not handled by the consumer side)

But because HPA directly writes the clientset of v1beta1, kubectl top has its own coordination API version function.

Under comprehensive consideration, I think it is less risky to maintain the original design.

If we want to change (or push v2) in the future, I think we should write KEP to deal with it separately.

rexagod: That's reasonable.

Just wanted to make sure we've got this at the back of our minds. I should've mentioned earlier but at the end of the day this is more of a "good to have" and "priority/important-soon" thing rather than a blocker to the graduation itself.

Issue #5933, opened 2026-02-24
CRD Sensitive Data Protection


trofimovdals:
### Enhancement Description

- One-line enhancement description (can be used as a release note):
Add support for marking CRD fields as sensitive with RBAC-gated access, and audit log masking
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/5937
- Discussion Link: 
  Related issue: https://github.com/kubernetes/kubernetes/issues/115592
- PRs by stage and milestone:
  - [ ] Alpha - v1.xx
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/5937
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #5937, opened 2026-02-25
KEP 5933: Add x-kubernetes-sensitive-data marker for CRD fields

trofimovdals:
Introduces `x-kubernetes-sensitive-data` vendor extension that enables field stripping on get/list/watch for unauthorized callers and unconditional masking of sensitive values in audit log entries.
  
- One-line PR description: Add support for marking CRD fields as sensitive with RBAC-gated access, and audit log masking

- Issue link: https://github.com/kubernetes/enhancements/issues/5933

### DIFF
--- a/keps/sig-api-machinery/5933-crd-sensitive-data/README.md
+++ b/keps/sig-api-machinery/5933-crd-sensitive-data/README.md
@@ -185,6 +185,8 @@ Callers **without** this permission receive the resource with sensitive fields r
 
 Write access to sensitive fields is controlled by standard `create`/`update`/`patch` verbs on the main resource — no additional write-side subresource is required.
 
+For resources with sensitive fields, full-object updates require authorization to read `<resource>/sensitive`. If a caller can update the main resource but cannot read its sensitive content, the API server rejects the request. This prevents accidental removal of sensitive fields during read-modify-write flows based on redacted responses.
+
 ### Field Stripping on get/list/watch
 
 In the CRD handler (`customresource_handler.go`), after deserializing the response object:
@@ -231,6 +233,10 @@ Every get/list/watch for a CRD with sensitive fields triggers an additional auth
 **Risk: Complexity in CRD handler and interaction with audit.**
 
 The CRD handler gains new responsibilities. Thorough unit and integration tests, as well as a clear separation into utility packages (`sensitive/sensitive.go`), will mitigate correctness risks.
+
+**Risk: Data loss from full-object updates by callers without sensitive read access.**  
+
+A caller that has permission to update the resource, but is not authorized to read `<resource>/sensitive`, receives a redacted form of the object. If that caller performs a full-object update (`PUT`) using the returned representation, sensitive fields omitted from the response may be unintentionally removed from the persisted object. To mitigate this risk, for resources with sensitive fields, the API server rejects full-object updates from callers that are not authorized to read the sensitive content.
 
 ## Design Details
 
@@ -461,3 +467,6 @@ This is a valid complementary approach but does not address the need for in-clus
 
 **Create a separate resource type for each "secret" object.**
 This leads to schema duplication and API proliferation. The per-field marker approach is more flexible and does not require additional resources.
+
+**Store sensitive values in a native Kubernetes Secret and reference it from the CRD.**  
+This is the primary alternative used today and remains broadly viable. However, it requires splitting the API across multiple objects and adds controller and lifecycle management complexity. This KEP aims to provide Secret-like protection for selected CRD fields without requiring an additional resource.
--- a/keps/sig-api-machinery/5933-crd-sensitive-data/README.md
+++ b/keps/sig-api-machinery/5933-crd-sensitive-data/README.md
@@ -1,0 +1,463 @@
+# KEP-5933: Sensitive Data Marker for CRD Fields
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [API Changes](#api-changes)
+    - [OpenAPI Extension](#openapi-extension)
+    - [Structural Schema](#structural-schema)
+    - [Validation Rules](#validation-rules)
+  - [RBAC Model](#rbac-model)
+  - [Field Stripping on get/list/watch](#field-stripping-on-getlistwatch)
+  - [Audit Log Masking](#audit-log-masking)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Feature Gate](#feature-gate)
+  - [Schema Layer](#schema-layer)
+  - [Sensitive Field Utilities](#sensitive-field-utilities)
+  - [CRD Handler Integration](#crd-handler-integration)
+  - [Watch Filtering](#watch-filtering)
+  - [Audit Integration](#audit-integration)
+  - [CRD Update and Deletion](#crd-update-and-deletion)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/5933-crd-sensitive-data/kep.yaml
```diff
@@ -0,0 +1,27 @@
+title: Sensitive Data Marker for CRD Fields
+kep-number: 5933
+authors:
+  - "@trofimovdals"
+  - "@sprait"
+owning-sig: sig-api-machinery
+participating-sigs:
+  - sig-auth
```
lmktfy, line 8: SIG Security too?

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,463 @@
...
+- 2025-02-19: Initial KEP draft.
+
+## Drawbacks
+
+- Requires upstream Kubernetes changes in apiextensions-apiserver and kube-apiserver, which implies a long review and acceptance cycle.
+- Adds complexity to the CRD handler, which must now integrate with RBAC subresources and the audit subsystem.
+- PATCH request audit masking is technically challenging, requiring JSON path analysis of patch bodies.
+
+## Alternatives
+
```
lmktfy, line 464: We should note that people could use a Secret instead (even if we rule out that alternative); it is a broadly viable option here.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,463 @@
...
+This brings Secret-grade protection to arbitrary CRD fields without requiring CRD authors to split sensitive values into separate Secret objects.
+
+## Motivation
+
+Operators and platform teams frequently store sensitive values — passwords, TLS private keys, API tokens — directly in CRD specs because the alternative (referencing a separate Secret and wiring it through controllers) adds significant complexity. Today, those values are:
+
+- Returned **in full** to every authenticated caller that can `get` the resource, even if the caller only needs non-sensitive fields.
+- Logged **in cleartext** in audit log entries, creating a compliance risk.
+
+There is no first-class Kubernetes primitive that lets a CRD author say "this field is sensitive — treat it accordingly".
```
lmktfy, line 92: That's true, and I note the non goals, but having a built in API with these properties would also be really convenient.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,463 @@
...
+
+As a platform team, I want to propose a standardized mechanism for marking sensitive CRD fields to the Kubernetes community, so that the ecosystem converges on a single approach instead of ad-hoc solutions in each operator.
+
+### Notes/Constraints/Caveats (Optional)
+
+- The marker applies at the API server level only. It does not prevent a controller with full RBAC from reading and leaking sensitive values through its own logs or status fields. Operators should apply the principle of least privilege.
+- PATCH masking in audit is more complex than GET/PUT masking because the patch body must be parsed and individual values replaced by path.
+
+### Risks and Mitigations
+
```
lmktfy, line 228: If I get an object and write it back with a PUT, do I lose data? How's that risk managed?

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
sprait: thanks, added to alternatives

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
sprait: Good catch. Yes, this is a potential data-loss scenario for callers that can update the resource but cannot read <resource>/sensitive, since they would receive a redacted representation and could overwrite hidden fields via PUT. I updated the KEP to explicitly document this risk and to mitigate it by rejecting full-object updates from such callers.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
sprait: this KEP is intentionally scoped to CRDs for now, while aiming to keep the design compatible with a more general built-in solution later

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,472 @@
...
+This brings Secret-grade protection to arbitrary CRD fields without requiring CRD authors to split sensitive values into separate Secret objects.
+
+## Motivation
+
+Operators and platform teams frequently store sensitive values — passwords, TLS private keys, API tokens — directly in CRD specs because the alternative (referencing a separate Secret and wiring it through controllers) adds significant complexity. Today, those values are:
+
+- Returned **in full** to every authenticated caller that can `get` the resource, even if the caller only needs non-sensitive fields.
+- Logged **in cleartext** in audit log entries, creating a compliance risk.
+
+There is no first-class Kubernetes primitive that lets a CRD author say "this field is sensitive — treat it accordingly".
```
sttts, line 92: Wasn't it always considered an anti-pattern in core types (in k/k) to have secret data outside of secrets in any type? Why should we handle CRDs differently? 

I remember in OpenShift we had a couple of resources that as a whole were considered sensitive (because mistakes were made not to follow the rule long ago).

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,472 @@
...
+kind: ClusterRole
+metadata:
+  name: database-sensitive-reader
+rules:
+  - apiGroups: ["example.com"]
+    resources: ["databases/sensitive"]
+    verbs: ["get", "list", "watch"]
+```
+
+Callers **without** this permission receive the resource with sensitive fields removed (set to `null` / omitted). Callers **with** this permission receive the full resource.
```
sttts, line 184: different behaviour depending on authz is not a good pattern. Assuming we want a feature like described, we would want very explicit list/get options for this, or a different resource path (subresource).

Also note that there are more dragons deep in the storage stack: watchlist caches events in marshalled form, and fans out to all consumers. We cannot differentiate between trusted and non-trusted there.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,472 @@
...
+
+Write access to sensitive fields is controlled by standard `create`/`update`/`patch` verbs on the main resource — no additional write-side subresource is required.
```
sttts, line 186: isn't this contradicting the direction of the role above? Being allowed to write because of `verb: *` permissions, but able to read sensitive data is very odd. Note that every update is also a read.

sttts: Thinking out loud: `databases/non-sensitive-thing` would give much cleaner behaviour. And then the client would have to ask for the sanitized versions. It would be a normal subresource with normal subresource semantics.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
sttts: So my gut feeling here is that we are building a feature to mitigate "wrong" usage. By design there are secrets in Kube as the primary (or only) resource that this senstivie.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
```diff
@@ -0,0 +1,472 @@
...
+
+For `watch`, a `watch.Filter` wrapper strips sensitive fields from every event object delivered to callers without the `sensitive` subresource permission.
+
+### Audit Log Masking
+
+Sensitive fields are **never** written to the audit log in cleartext. Before serialization:
+
+- `audit.LogRequestObject()` and `audit.LogResponseObject()` invoke an `ObjectTransformer` registered by the CRD handler.
+- The transformer deep-copies the object and calls `MaskSensitiveFields()`, replacing each sensitive value with `"******"`.
+- PATCH request bodies containing sensitive paths are masked in the same way.
```
sttts, line 206: how? patch could be json patch where it is not that clear how to mask.

keps/sig-api-machinery/5933-crd-sensitive-data/README.md
sttts: This case does not even make sense. A patch with `{}` payload is a read. Everybody with write permissions has inherently (sensitive) read too.

sttts: And it highlights another problem: authz can fail without the SAR caller to notice (think of authz webhooks). So a read would degrade to a masked read. And then every write is a potential data loss without noticing and without a way to know.

trofimovdals:
After discussing this KEP on Slack, we decided to close it, as there are still several controversial points.

Issue #5939, opened 2026-03-01
Cluster Selection for Multi-Cluster Services


MrFreezeex:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add cluster selection to Multi-Cluster Services, allowing users to steer traffic to preferred clusters based on ClusterProperty values
- Kubernetes Enhancement Proposal: 
- Discussion Link: https://docs.google.com/document/d/14Ts1PhkSd-ouiGLhMA6uZN6RfoObqKvqpxnf7qLko5Y
- PRs by stage and milestone:
  - [ ] Alpha - v1.xx
    - [ ] KEP (`k/enhancements`) update PR(s):
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

lauralorenz:
Triage note:
 - There was some preamble in SIG-MC meetings on 3/3, 2/17, 1/20, 1/13, 1/6
 - Planned follow up with SIG-Network (re: the general idea) and SIG-Api-Machinery (re: CEL) but want a SIG-MC review round on the KEP as written here first

PR #5940, opened 2026-03-01
KEP-5939: add Cluster Selection for Multi-Cluster Services

MrFreezeex:
- One-line PR description: Cluster Selection for Multi-Cluster Services

- Issue link: https://github.com/kubernetes/enhancements/issues/5939

- Other comments:

Adds a cluster selection mechanism to cover a standard way to let user express a preference over certain clusters (for instance routing to the local cluster first if endpoints are available then to clusters in the same region etc).

This KEP was built using AI mainly for research purposes and to some extent helped writing this document (it is still heavily human written/reviewed though!)

### DIFF
--- a/keps/sig-multicluster/5939-mcs-cluster-selection/README.md
+++ b/keps/sig-multicluster/5939-mcs-cluster-selection/README.md
@@ -1,0 +1,1117 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+Follow the guidelines of the [documentation style guide].
+In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+
+To get started with this template:
+
+- [x] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [x] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [x] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [x] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [x] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [x] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-5939: Cluster Selection for Multi-Cluster Services
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Locality-based failover](#story-1-locality-based-failover)
+    - [Story 2: Multi-dimension selection with arbitrary properties](#story-2-multi-dimension-selection-with-arbitrary-properties)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Overview](#api-overview)
+    - [Label Selector](#label-selector)
+    - [Validation](#validation)
+  - [Cluster Selection Algorithm](#cluster-selection-algorithm)
+    - [Resolving Cluster Selectors](#resolving-cluster-selectors)
+    - [Consuming Cluster Selector Results](#consuming-cluster-selector-results)
+  - [Interaction with Service Traffic Distribution](#interaction-with-service-traffic-distribution)
+  - [Reading ClusterProperties from Constituent Clusters](#reading-clusterproperties-from-constituent-clusters)
+  - [Reference Implementation in kubernetes-sigs/mcs-api](#reference-implementation-in-kubernetes-sigsmcs-api)
+  - [Conflict Resolution](#conflict-resolution)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha -&gt; Beta graduation](#alpha---beta-graduation)
+    - [Beta -&gt; GA graduation](#beta---ga-graduation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [EndpointSlice Active Label](#endpointslice-active-label)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+[MCS-API]: /keps/sig-multicluster/1645-multi-cluster-services-api
+[ClusterProperty]: /keps/sig-multicluster/2149-clusterid
+[PlacementDecision]: /keps/sig-multicluster/5313-placement-decision-api
+[ClusterProfile]: /keps/sig-multicluster/4322-cluster-inventory
+[mcs-api-repo]: https://github.com/kubernetes-sigs/mcs-api
+
+## Summary
+
+The [Multi-Cluster Services API (MCS-API)][MCS-API] lets users export a
+Kubernetes `Service` with `ServiceExport` so that it becomes consumable across
+a clusterset. When the same `Service` is exported from multiple clusters, each
+importing cluster gets a single automatically generated `ServiceImport`.
+
+Today there is no standard way to express which clusters should be preferred. A
+consumer cannot say "prefer my own cluster, then my region, then anywhere else".
+
+This KEP introduces a **cluster selection** concept for directing traffic. It
+adds an ordered `clusterSelectors` field to `ServiceExport` that lets users
+select subsets of constituent clusters based on
+[`ClusterProperty`][ClusterProperty] values.
+
+Each entry uses a standard Kubernetes `metav1.LabelSelector`, which may contain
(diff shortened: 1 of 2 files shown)


lauralorenz:
Cross post from issue:
 - There was some preamble in SIG-MC meetings on 3/3, 2/17, 1/20, 1/13, 1/6
 - Planned follow up with SIG-Network (re: the general idea) and SIG-Api-Machinery (re: CEL) but want a SIG-MC review round on the KEP as written here first

keps/sig-multicluster/5939-mcs-cluster-selection/README.md
```diff
@@ -0,0 +1,1233 @@
...
+    // +listType=atomic
+    // +kubebuilder:validation:items:MaxItems:=16
+    ClusterSelectors []ClusterSelector `json:"clusterSelectors,omitempty"`
+
+    // ClusterSelectorsFallbackPolicy is the reconciled fallback
+    // policy, derived from the constituent ServiceExports.
+    // +optional
+    // +kubebuilder:default="AllClusters"
+    // +kubebuilder:validation:Enum=AllClusters;None
+    ClusterSelectorsFallbackPolicy *FallbackPolicy `json:"clusterSelectorsFallbackPolicy,omitempty"`
```
aojea: My experience with all the iterations with traffic distribution:
- https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/2433-topology-aware-hints/README.md?plain=1
- https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3015-prefer-same-nod
- https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/4444-service-traffic-distribution

1. Is a scheduling problem,  if you schedule your workloads correctly the network logic to handle overflows or failover is much simpler and easier , no need to do complicated algorithms to solve weird situations and edge cases
2. Developers/network products want complicated, users want simple ... kind of builds on 1.

Based on that, is not `ClusterSelectors []ClusterSelector ` enough with `ClusterSelectorsFallbackPolicy` implicit to AllClusters? Why should I schedule workloads behind a service in a cluster and later use the network properties to not use those workloads?

MrFreezeex: I am not sure I fully got your last paragraph, do you mean that like for instance to get the user story 3 with gdpr, users should just export an entirely different service like `my-svc-gdpr` and that we should basically just remove `ClusterSelectorsFallbackPolicy` entirely (and make the only possible behavior `AllClusters`)?

If that's the case/I understood this correctly, I think it's a reasonable take thanks! Also even if we do have some users requests for that later and it turns out to be much more convenient, there's nothing preventing us to add back something like that in the future so it's most likely much better to not start with that :eyes:.

MrFreezeex: I just removed `ClusterSelectorsFallbackPolicy` from the KEP, thanks for the feedback!

MrFreezeex:
FTR I did a general pass on the KEP to improve the prose/phrasing everywhere and clarify a few things that could be ambiguous. And also while I was brainstorming about how to implement it in Cilium I noticed that there the API that I was proposing with an EndpointSlice label was hard to implement correctly so I changed to present it in the ServiceImport status to let consumers (which may include the own MCS implementation datapath!) knows about the selected clusters.

Will talk about this more in the next SIG-MC meeting!

kahirokunn:
I think source cluster selection involves two separate decisions:

1. Which importing clusters the Service owner makes the Service available to
2. How each importing cluster prioritizes the available source clusters

For example, an importing cluster in Japan might prefer the following order:

1. The local cluster
2. Other clusters in Japan
3. Clusters in Asia
4. Any other available clusters

An importing cluster in the EU might instead need:

1. The local cluster or clusters in the same country
2. Clusters within the EU
3. No fallback outside the EU

Some parts of this may overlap with ideas already discussed, but I tried to sketch one possible end state.

#### Provider-side availability

`ServiceExportGrant` represents which importing clusters the Service owner allows to use the export.

```yaml
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceExportGrant
metadata:
  name: api-eu
spec:
  targetRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceExport
      name: api
  toClusters:
    propertySelector:
      matchLabels:
        jurisdiction.example.io: eu
```

#### Importing-cluster-side routing

`ServiceImportRoutingPolicy` allows each importing cluster to specify the priority and fallback behavior for the source clusters available to it. The `ServiceImport` itself remains managed by the MCS implementation, and users refer to it from this separate resource.

```yaml
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceImportRoutingPolicy
metadata:
  name: api-eu-routing
spec:
  targetRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: api
  sourceClusterSelectors:
    - propertySelector:
        matchLabels:
          country.example.io: "@SameAsImporter@"
    - propertySelector:
        matchLabels:
          jurisdiction.example.io: eu
  fallbackPolicy: None
```

For an importing cluster in Japan, a different order can be specified for the same logical Service:

```yaml
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ServiceImportRoutingPolicy
metadata:
  name: api-japan-routing
spec:
  targetRefs:
    - group: multicluster.x-k8s.io
      kind: ServiceImport
      name: api
  sourceClusterSelectors:
    - propertySelector:
        matchLabels:
          cluster.clusterset.k8s.io: "@SameAsImporter@"
    - propertySelector:
        matchLabels:
          region.topology.k8s.io: "@SameAsImporter@"
    - propertySelector:
        matchLabels:
          continent.topology.k8s.io: "@SameAsImporter@"
  # Fall back only within the source clusters made available by ServiceExportGrant
  fallbackPolicy: AllAllowedClusters
```

What do you think about this direction of treating availability and per-importing-cluster routing as separate responsibilities?

kahirokunn:
Minor thing, but matchLabels inside something called propertySelector reads a little inconsistent to me. Wondering if that'll be intuitive for users.

MrFreezeex:
> [...]
> What do you think about this direction of treating availability and per-importing-cluster routing as separate responsibilities?

Hmm I think there are some ideas that could be interesting to explore. We didn't explore a lot making separate CRD TBH.

On your proposal the ability to have the routing decision on each cluster means that it doesn't come bundled with the usual ServiceImport so users would have to themselves apply it on each clusters. It means that on a cluster only importing the service users would have to do extra steps/the resource is no longer standalone (if the routing decision is critical in an env and this sounds like it could become critical for at least some people).

It could be a bit weird RBAC wise too, like you might want to give some consumers the ability to create this resource in another namespace or give producer access/the ability to configure that resources in other clusters that you would otherwise not have access to. I know that this is also the case with namespace but namespaces are a bit simpler than that and you can't really have in between state where the ServiceImport would work but not with the correct settings.

I reckon that configuring this might be useful in some situations but otherwise even if it's driven by producers you could handle those special cases by creating multiple services and export them as long as the selection mechanism is flexible enough so that you don't create too many of those ~.

The producer side restrictions seems interesting but this might be worthy to have its own KEP as it's a bit different than routing decision even if it might solve some adjacent issues 🤔. Something like that could be adding export and import restrictions at the same time too ~.

kahirokunn:
That makes sense. I agree that having a ServiceImport without the corresponding routing policy could leave the system in a partially configured state, which could be risky when the routing constraints are important.

Given that, using separate Services for different use cases and creating a ServiceExport for each seems like a reasonable and simpler approach for now.

Thanks for the explanation. I understand the trade-off now!

MrFreezeex:
> Minor thing, but matchLabels inside something called propertySelector reads a little inconsistent to me. Wondering if that'll be intuitive for users.

That's true also having the same go type could be problematic if this kubernetes type start adding restrictions that makes sense for labels that doesn't work for cluster properties (I.e.: any characters that you could not have in label values) 🤔. Will think about it and do some research, thanks!

BTW @kahirokunn if this wasn't clear do not hesitate if you have other ideas and/or want to participate more in this KEP, you are more than welcome to do so!

kahirokunn:
Thanks!

The OCM Placement API may be a useful reference here. It separates label-based selection and ClusterClaim-based selection into different fields:

https://open-cluster-management.io/docs/concepts/content-placement/placement/#label-or-clusterclaim-selection

Could we use a similar structure with `labelSelector` and `propertySelector`?

For example:

```yaml
clusterSelectors:
  - labelSelector:
      matchLabels:
        environment: production
        region: eu-west-1

  - propertySelector:
      matchExpressions:
        - name: infrastructure.example.io/cpu-architecture
          operator: In
          values:
            - amd64
            - arm64
```

This would make it clear whether each selector targets `metadata.labels` or `status.properties`, while avoiding `matchLabels` inside something named `propertySelector`.

I also raised a related question about the distinction between `metadata.labels` and `status.properties` here:

https://github.com/kubernetes-sigs/cluster-inventory-api/issues/76

MrFreezeex:
Hmmm I am not entirely sure we could do something about clusterprofile labels here :thinking:. This KEP is more About API based than ClusterProfile based and I am not sure how we would gather the ClusterProfile info in practice for implementations. It's something we might be able to add later if some implementation can actually do something with it and/or we find a way to make it more accessible somehow (but as it's something tied to ClusterProfile which could be represented multiple time with different labels I am not sure how possible it is)

MrFreezeex:
To explain a bit more, for instance in Cilium we have reasonably access to all clusters but we don't know which one is supposed to be the hub cluster (and it could not even be connected via our cluster mesh thing even). It's comfortable to rely on about API because we can expect that if implementations are able to export services they probably can also export the ClusterProperties. If someone wants to use the labels I have no problem with it though, but not all implementations might be able to support that somehow ~.

kahirokunn:
Thanks for the explanation. I think I may have misunderstood the intended architecture.

I had assumed that KEP-5939 would propagate `ClusterProfile` objects to the member clusters, and that each member cluster would read that information and act on it locally.

With that model, I thought this could be a first step toward going beyond simply publishing a Service within the boundary of a `ClusterSet`. For example, it could eventually enable mesh-style connectivity, PrivateLink-like connectivity between selected clusters, or network segmentation patterns similar to public, private, and database subnets.

That was also why I had assumed the architecture would not depend on a hub cluster at runtime.

MrFreezeex:
> I had assumed that KEP-5939 would propagate ClusterProfile objects to the member clusters, and that each member cluster would read that information and act on it locally.

Ah yeah nop that could be nice but would probably be it's own KEP to have something like that if it's actually possible. Like I am assuming it would be some KEP to be able to propagate some info to member cluster so that they can then use it (which could be in turn used by MCS-API or anything else). If that's true then yeah we could for sure extend that KEP to include some info available from there!

kahirokunn:
That makes sense!

I think that if MCS API could evolve from only publishing Services within the boundary of a `ClusterSet` into something that can represent a much broader range of network topologies, it could become useful to many more projects and potentially see wider adoption.

kahirokunn:
I've been thinking a bit more broadly about the MCS API and multi-cluster routing since our previous discussion. I'd like to share a few additional thoughts around cluster selection and routing.

These are still fairly rough ideas, and I'm not suggesting that everything needs to be addressed by this KEP. I hope some of this can be useful input for the discussion.

First, I think **cluster selection itself has clear value**, especially for locality, cost, and policy.

For example, even if another region has slightly better latency or more available capacity, users may still want to keep traffic within the same region under normal conditions to avoid cross-region network costs.

So I see cluster selection as answering something like:

> **Where should traffic preferably stay?**

However, I wonder what should happen when the clusters selected by the current `clusterSelectorResult` are still technically available, but their performance has significantly degraded.

As I understand the current proposal, consumers use the first `clusterSelectorResult` that has at least one eligible Ready or Serving endpoint. This means that a cluster can continue receiving traffic even when it is heavily degraded, as long as its endpoints remain Ready.

For example:

```mermaid
flowchart LR
    C[Client]

    subgraph A["Cluster A - same region"]
        A1["Pod A1<br/>250 ms<br/>Ready"]
        A2["Pod A2<br/>300 ms<br/>Ready"]
        A3["Pod A3<br/>400 ms<br/>Ready"]
    end

    subgraph B["Cluster B - next clusterSelectorResult"]
        B1["Pod B1<br/>25 ms<br/>Ready"]
        B2["Pod B2<br/>27 ms<br/>Ready"]
        B3["Pod B3<br/>24 ms<br/>Ready"]
    end

    C -->|"current clusterSelectorResult"| A1
    C --> A2
    C --> A3

    C -.->|"not considered while the current result<br/>has Ready endpoints"| B1
    C -.-> B2
    C -.-> B3
```

I think this can happen quite naturally in multi-cluster environments. Clusters can differ in geographic distance, hardware performance, available capacity, or simply experience partial degradation independently.

For runtime optimization in this kind of situation, one algorithm I particularly like is **Peak EWMA**.

Envoy has some interesting performance validation here:

https://github.com/envoyproxy/envoy/pull/40653

In their test, one of 10 upstream servers was 10x slower than the others. Peak EWMA was able to avoid the slow server very effectively and significantly reduce tail latency.

The Envoy community is also continuing work around Peak EWMA and latency-aware load balancing, including how per-response latency signals should be supported in the core LB infrastructure:

https://github.com/envoyproxy/envoy/issues/44686

So I'm not suggesting that Peak EWMA itself should become part of the MCS API, or that every MCS implementation can simply use it today. I see it more as a useful example of adaptive routing based on actual runtime conditions.

What I'm particularly interested in is how **cluster selection and adaptive load balancing are expected to work together**.

Conceptually, I imagine something like this:

```mermaid
flowchart LR
    A["clusterSelectorResults"]
    --> B["Clusters from the<br/>current result"]

    B --> C["Adaptive LB<br/>e.g. Peak EWMA"]

    C --> D["Endpoint selected using<br/>runtime conditions"]
```

This seems like a useful separation of responsibilities.

* **Cluster selection** decides which clusters should currently be considered, based on locality, cost, or policy.
* **Adaptive load balancing** decides which endpoint should receive the request based on runtime conditions.

For example, cluster selection could keep traffic within the same region to control network costs, while Peak EWMA could automatically avoid a slow Pod or Node within the selected clusters.

But this leaves another interesting question.

**What happens when all clusters in the current `clusterSelectorResult` become overloaded or degraded, while their endpoints are still technically Ready?**

Peak EWMA could choose the least bad endpoint among them, but it cannot help much if all of them are performing poorly and the next `clusterSelectorResult` is not being considered.

This feels similar to a problem Kubernetes has already encountered with **Topology Aware Routing**.

Topology Aware Routing also tries to keep traffic within a preferred topology for latency and network cost reasons, but locality is not applied unconditionally. It has several safeguards that fall back to cluster-wide routing when topology-aware routing cannot be considered safe.

https://kubernetes.io/docs/concepts/services-networking/topology-aware-routing/#safeguards

I find the `expected overload` safeguard particularly interesting here.

At the same time, that safeguard is not based on real-time traffic feedback, so it cannot completely prevent individual endpoints from becoming overloaded. I think this shows the value of combining topology or cluster information with runtime signals.

Maybe a similar idea could apply here:

```mermaid
flowchart LR
    A["clusterSelectorResults"]
    --> B["Clusters from the<br/>current result"]

    B --> S{"Safeguard<br/>can these clusters safely<br/>handle traffic?"}

    S -->|"yes"| C["Adaptive LB<br/>e.g. Peak EWMA"]
    S -->|"no"| N["Try next<br/>clusterSelectorResult"]

    N --> B
```

So I think there may be three separate concerns here:

1. **Cluster selection**: which clusters should currently be considered.
2. **Adaptive load balancing**: which endpoint should receive the request right now.
3. **Safeguard / fallback**: when we should stop using the current `clusterSelectorResult` and try the next one.

I understand that complex routing rules are a non-goal of this KEP, so I'm not suggesting that this KEP should define a full load-aware routing mechanism.

But Kubernetes has already accumulated some experience around the trade-off between locality and capacity through Topology Aware Routing. I wonder if its safeguards and fallback behavior could also be useful input for this KEP.

Maybe the general principle is:

> **Prefer locality, but allow that preference to be relaxed when locality becomes unsafe.**

I'm especially curious where we expect that safeguard and fallback decision to live. Would it be the MCS implementation, Gateway API / service mesh, the datapath implementation, or somewhere else?

WDYT?

MrFreezeex:
Hi @kahirokunn :wave:,

Thanks for proposing this new idea and useful context around Peak EWMA!

I pushed two new commits to address both of your comments/suggestions!

> First, I think cluster selection itself has clear value, especially for locality, cost, and policy.

:heart:

> So I think there may be three separate concerns here:
> 
> 1. **Cluster selection**: which clusters should currently be considered.
> 2. **Adaptive load balancing**: which endpoint should receive the request right now.
> 3. **Safeguard / fallback**: when we should stop using the current `clusterSelectorResult` and try the next one.

I like this way of listing the different concerns! I think my plan would be to do the following about each one:
1) Cluster Selection: at first selection based on `ClusterProperty`,  then most likely with a CEL expression, maybe a selection over `ClusterProfile` labels if we can have those in every clusters. And eventually some level of external integration where some other third party controllers could influence the selection result.

It's not clear right now if this external integration should be that the controllers give the full list of clusters selected (like `PlacementDecision`) or simply provide some additional parameters that could be fed to the CEL expressions. I think this external integration could potentially have some use case with your latest idea but we would need to be careful here since having no runtime dependencies makes cluster selection significantly easier to implement (if the logic dependent on runtime metrics is fully in a external thing that could still be doable though I guess).

2) Load balancing algorithm (I changed the title but it's about the same thing): This is driven by the consumer of a ServiceImport and can be anything that would work with Service too (so AFAIK any algorithm the consumer would like, potentially driven by some extra implementation specific configurations).

3) Safeguard / fallback: I think the default behavior should always be to choose the first set of clusters with available/eligible endpoints but the logic may change if there's any opt-in and consumer specific mechanism to configure the logic in another way.

> I'm especially curious where we expect that safeguard and fallback decision to live. Would it be the MCS implementation, Gateway API / service mesh, the datapath implementation, or somewhere else?

IMO it's much easier to push this on the consumer of the ServiceImport than getting it at the cluster selection stage. "ServiceImport" consumer could be any of the things that you mentioned though :sweat_smile:.

> But Kubernetes has already accumulated some experience around the trade-off between locality and capacity through Topology Aware Routing.

That's true that Kubernetes had some experience with something similar but... SIG network eventually went in the other direction of removing this entirely in the following KEP: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3015-prefer-same-node/README.md#risks-and-mitigations. So I went ahead and applied some similar language to justify that but I also added an additional paragraph to recognize that there may be some additional implementation specific config (like annotations or labels for instance) to have a different logic.

Given the language used in KEP-3015 I don't think that having implementation specific configuration for other fallback would be invalid for `trafficDistribution` too. Especially with the following sentence:
> **In general**, the proxy should always route to a same-zone endpoint if any is available.

In our case I think it's a bit more relevant to go a bit beyond that and recognize that implementation specific configuration may exist since we are not geared toward an official implementation (unlike KEP-3015 with kube-proxy) ~.

Does that makes sense?

kahirokunn:
Yes, this makes sense to me. Thanks for the detailed explanation and for updating the KEP!

I like the separation where:

1. cluster selection defines the ordered sets of preferred clusters,
2. the `ServiceImport` consumer decides how to load balance within the selected set, and
3. any smarter fallback to the next `clusterSelectorResult` is consumer-specific and explicitly opt-in.

I also agree that keeping runtime dependencies out of the cluster selection stage makes the base mechanism much simpler and more predictable.

One distinction that might be worth making explicit in the KEP is **preference vs. hard constraint**.

For example, selecting EU clusters because they are preferred for locality or cost is different from saying that traffic **must never leave the EU** for compliance or security reasons. Since the default behavior eventually falls back to all constituent clusters, I think `clusterSelectors` should probably be described as expressing routing preferences rather than an authorization / isolation / compliance boundary.

Maybe this could be mentioned briefly in the Non-Goals or Risks section.

Other than that, the current direction looks good to me 👍

MrFreezeex:
Ah yep that's a good point, thanks! The traffic distribution KEP also stated it as a non goal here: https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/4444-service-traffic-distribution/README.md#non-goals.

kahirokunn:
Thank you! In that case, I don't have any concerns for now. 🙏

tpantelis:
Overall the proposal looks reasonable.

PR #5942, opened 2026-03-03
KEP 5941: Shared Consumable Capacity

SchSeba:
- One-line PR description:
KEP to allow DRA drivers to define parent-scoped shared capacities that are consumed per allocation by related child devices, with scheduler accounting that prevents aggregate over-allocation. 

- Issue link:
https://github.com/kubernetes/enhancements/issues/5941

- Other comments:

### DIFF
--- a/keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
+++ b/keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
@@ -1,0 +1,560 @@
+# KEP-NNNN: DRA Shared Consumable Capacities Across Related Devices
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: SR-IOV bandwidth as a shared parent resource](#story-1-sr-iov-bandwidth-as-a-shared-parent-resource)
+    - [Story 2: CPU/Memory resource alignment via PCIE root grouping](#story-2-cpumemory-resource-alignment-via-pcie-root-grouping)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API additions](#api-additions)
+  - [Allocation behavior](#allocation-behavior)
+  - [Feature gate](#feature-gate)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+    - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Recommended Rollout](#recommended-rollout)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e tests for all Beta API operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests]
+  - [ ] (R) Minimum two week window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) all GA endpoints must be hit by [Conformance Tests] within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] Implementation history section is up to date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation, such as design docs, SIG meeting notes, relevant PRs/issues, release notes
+
+## Summary
+
+Dynamic Resource Allocation (DRA) currently supports:
+
+- parent-scoped shared integer counters via partitionable devices, and
+- per-device request-driven consumable capacity via `allowMultipleAllocations`.
+
+However, there is no upstream mechanism that combines both capabilities so that:
+
+- multiple related devices can consume from one shared parent-scoped capacity pool, and
+- the consumed amount is determined by the request at allocation time.
+
+This KEP proposes extending the existing DRA `sharedCounters` with `requestPolicy` support and adding a `valueFrom` mapping in `consumesCounters`, so that multiple related devices can consume request-driven amounts from shared counter sets during allocation, with scheduler accounting that prevents aggregate over-allocation.
+
+The proposal is generic and applies to multiple hardware models. SR-IOV VF bandwidth accounting is one motivating use case, but not the only target.
+
+## Motivation
+
+Some resources are naturally parent-scoped and shared across multiple allocatable child devices. Examples include link bandwidth, shared memory channels, or scheduler-governed throughput budgets.
+
+In these cases, publishing only child devices is desirable for user experience and hardware mapping, but scheduler-side accounting must still guard the aggregate shared parent budget.
+
+Existing DRA capabilities leave a gap:
+
+- Partitionable counters are shared across devices, but consumption is static in `ResourceSlice` data.
+- Consumable capacity is request-driven, but accounting is device-local.
+
+This causes either over-advertisement, rigid static partitioning, or custom out-of-band admission behavior, all of which reduce correctness and portability.
+
+### Goals
+
+- Introduce a generic DRA model for parent-scoped shared capacities consumed by related child devices.
+- Keep request-driven consumption semantics aligned with consumable capacity (`capacity.requests` + request policy).
+- Ensure scheduler allocator prevents aggregate over-allocation for each shared capacity set.
+- Preserve compatibility with existing DRA features and patterns where possible.
+
+### Non-Goals
+
+- Defining device- or vendor-specific runtime enforcement mechanisms (for example, tc, firmware limits, NIC policy engines).
+- Defining standard capacity names across vendors.
+- Solving cross-node or cross-pool global capacity aggregation.
+- Replacing existing partitionable counters or consumable capacity; this extends existing API types.
+
+## Proposal
+
+Extend the existing `sharedCounters` in `ResourceSlice` with `requestPolicy` support, and add a `valueFrom` mapping to `consumesCounters` so that devices can map inbound capacity requests to shared counters. Reuse existing `ResourceClaim` request syntax for capacity requests, with scheduler accounting performed against referenced shared counter sets.
+
+At a high level:
+
+1. A driver publishes `sharedCounters` with counters that include `requestPolicy` (default, valid range, step).
+2. Each allocatable device references the shared counter set(s) it draws from via `consumesCounters`, using `valueFrom` to map capacity request keys to counters.
+3. A claim requests capacity using existing `capacity.requests` fields.
+4. During allocation, the scheduler resolves the `valueFrom` mappings, checks and consumes the requested amount from all relevant counter sets (device-specific and/or shared), rejecting candidates that would exceed any counter limit.
+
+### User Stories
+
+#### Story 1: SR-IOV bandwidth as a shared parent resource
+
+As a cluster user, I request one or more VFs and a bandwidth amount per request.
+As an operator, I want scheduler admission to ensure that total bandwidth promised across VFs on the same PF does not exceed PF capacity.
+
+This allows:
+
+- keeping VF as the allocatable unit, and
+- preventing over-allocation of PF bandwidth without static pre-partitioning.
+
+#### Story 2: CPU/Memory resource alignment via PCIE root grouping
+
+Co-locating compute resources (cores, devices, memory) is essential for low-latency workloads, and there is work to allocate CPU and memory resources using DRA drivers. There is a growing consensus that NUMA or Socket IDs do not correctly represent modern CPU hardware; the community is shifting towards using PCIE roots as the alignment attribute. This is demonstrated by the fact that `pcieRoot` was the first DRA standard attribute. In modern CPUs, more than one PCIE root can be attached to the same NUMA zone, which further reinforces the case for using PCIE roots as the alignment mechanism.
+
+Therefore, a non-directly-accessible (parent) set of resources (a CPUSet and a NUMA node) is shared across accessible (child) devices, represented by PCIE roots.
+
+Example from real hardware (dual XEON Gold 6320R):
+
+```
+"root"="pci0000:00" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:17" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:3a" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:85" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+"root"="pci0000:ae" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+"root"="pci0000:d7" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+```
+
+A CPU/memory DRA driver can publish one shared capacity set per NUMA node (the parent budget of available CPUs and memory) and map each PCIE root device to the appropriate set. Scheduler accounting then prevents aggregate over-allocation of CPUs or memory across PCIE roots that share the same NUMA zone.
+
+See also: [kubernetes/enhancements#5491](https://github.com/kubernetes/enhancements/issues/5491) for related work on PCIE-root-based resource alignment.
+
+### Notes/Constraints/Caveats
+
+- This proposal adds API and allocator behavior and therefore requires cross-SIG review (at minimum SIG Scheduling and SIG API Machinery).
+- Request policy semantics should stay as close as possible to existing consumable-capacity behavior to reduce user surprise.
+- Drivers must publish consistent relationships between devices and shared counter sets; invalid `valueFrom` references must be rejected or treated as unsatisfiable.
+
+### Risks and Mitigations
+
+- **Risk: scheduler overhead increases with extra accounting.**  
+  Mitigation: follow existing allocator cache/aggregation patterns used by DRA allocation logic and add perf coverage.
+
+- **Risk: API complexity for users and driver authors.**  
+  Mitigation: keep user request syntax unchanged (`capacity.requests`) and keep shared-set definitions driver-facing.
+
+- **Risk: invalid or inconsistent pool references.**  
+  Mitigation: API validation plus allocator safeguards that reject incomplete or invalid candidate mappings.
+
+## Design Details
+
+### API additions
+
+This KEP proposes extending existing DRA API types by adding **optional fields** to the existing `Counter` struct. No existing types are removed, renamed, or split. All existing `ResourceSlice` objects remain valid without modification.
+
+Both `sharedCounters` and `consumesCounters` (and the `Counter` type they use) are alpha APIs (`+k8s:alpha(since: "1.36")`), gated behind the `DRAPartitionableDevices` feature gate. Alpha APIs carry no backward compatibility guarantees, but this proposal is designed to be purely additive regardless.
+
+The changes to the `Counter` struct are:
+
+1. **Add an optional `requestPolicy` field** to `Counter`. When a counter is defined in a `CounterSet` (inside `sharedCounters`), this field specifies default values, valid ranges, and step sizes for request-driven consumption. In other contexts this field is ignored.
+
+2. **Add an optional `valueFrom` field** to `Counter`. When a counter is referenced in `DeviceCounterConsumption` (inside `consumesCounters`), this field maps an inbound `capacity.requests` key to the counter, making consumption request-driven rather than static.
+
+3. **Relax the `value` field** from required to conditionally required: in `consumesCounters`, either `value` (static consumption, existing behavior) or `valueFrom` (request-driven consumption, new behavior) must be specified. Existing objects that set `value` continue to work unchanged.
+
+Example `ResourceSlice` with shared counter request policy:
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceSlice
+metadata:
+  name: counter-slice
+spec:
+  driver: "resource-driver.example.com"
+  pool:
+    generation: 1
+    name: "my-pool"
+    resourceSliceCount: 2
+  sharedCounters:
+  - name: pf-0-counter-set
+    counters:
+      bandwidth:
+        value: "100G"
+        requestPolicy:
+          default: "1G"
+          validRange:
+            min: "1M"
+            max: "100G"
+            step: "1M"
+```
+
+Example `ResourceSlice` with device `valueFrom` mapping:
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceSlice
+metadata:
+  name: device-slice
+spec:
+  driver: "resource-driver.example.com"
+  pool:
+    generation: 1
+    name: "my-pool"
+    resourceSliceCount: 2
+  nodeName: "my-node"
+  devices:
+  - name: vf-0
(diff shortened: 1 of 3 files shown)


ffromani:
/ok-to-test

you miss KEP.yaml and PRR data. Random example: https://github.com/kubernetes/enhancements/pull/5319/changes

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+As a cluster user, I request one or more VFs and a bandwidth amount per request.
+As an operator, I want scheduler admission to ensure that total bandwidth promised across VFs on the same PF does not exceed PF capacity.
+
+This allows:
+
+- keeping VF as the allocatable unit, and
+- preventing over-allocation of PF bandwidth without static pre-partitioning.
+
+#### Story 2: Generic parent-scoped shared resource
```
ffromani: Story 2: modelling CPU/Memory native resources using PCIE root as grouping attribute

Co-locating compute resources (core, devices,memory) is essential for low-latency workloa, and there is work to allocate  CPU and memory resources using DRA drivers. There's a growing consensus that NUMA or Socket IDs don't represent correctly modern CPU hardware, therefore the community is shifting towards using PCIE roots as aligning attribute. This is demonstrated by the fact pcieRoot was the first DRA standard attribute. In modern CPUs, more than a PCIE root can be attached to the same NUMA zone - and this further reinforces the point or using it as aligning mechanism. The linux Kernel will therefore report multiple times the same NUMA zone attached to different PCIE roots, and the sameset of CPUs is likewise affine to the same PCIE root.

Therefore, we have a non-directly-accessible (parent) set of resources (a CPUSet and a NUMA node) which is shared across accessile (child) devices, represented by PCIE roots.

Example from real hardware (dual XEON Gold 6320R)
```
"root"="pci0000:00" "localCPUs"="0,2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44,46,48,50,52,54,56,58,60,62,64,66,68,70,72,74,76,78,80,82,84,86,88,90,92,94,96,98,100,102" "NUMANode"=0
"root"="pci0000:17" "localCPUs"="0,2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44,46,48,50,52,54,56,58,60,62,64,66,68,70,72,74,76,78,80,82,84,86,88,90,92,94,96,98,100,102" "NUMANode"=0
"root"="pci0000:3a" "localCPUs"="0,2,4,6,8,10,12,14,16,18,20,22,24,26,28,30,32,34,36,38,40,42,44,46,48,50,52,54,56,58,60,62,64,66,68,70,72,74,76,78,80,82,84,86,88,90,92,94,96,98,100,102" "NUMANode"=0
"root"="pci0000:85" "localCPUs"="1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,45,47,49,51,53,55,57,59,61,63,65,67,69,71,73,75,77,79,81,83,85,87,89,91,93,95,97,99,101,103" "NUMANode"=1
"root"="pci0000:ae" "localCPUs"="1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,45,47,49,51,53,55,57,59,61,63,65,67,69,71,73,75,77,79,81,83,85,87,89,91,93,95,97,99,101,103" "NUMANode"=1
"root"="pci0000:d7" "localCPUs"="1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31,33,35,37,39,41,43,45,47,49,51,53,55,57,59,61,63,65,67,69,71,73,75,77,79,81,83,85,87,89,91,93,95,97,99,101,103" "NUMANode"=1
```

johnbelamaric: PTAL at https://github.com/kubernetes/enhancements/issues/5491 and see if that's useful.

dom4ha:
What about making the shared counters part of the scheduler core?

This way we could use shared capacity across different plugins and solve a few problems like native resource accounting across DRA and ResourceFit plugins: https://github.com/kubernetes/enhancements/issues/5517

@pravk03

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,448 @@
...
+- Adds API and scheduler complexity.
+- Introduces another DRA concept to learn for driver authors.
+- Requires careful policy semantics to avoid ambiguity across capacity sources.
+
+## Alternatives
+
+- Static parent capacity partitioning into child-specific fixed slices (simple but inflexible).
+- Model parent capacity as a separate allocatable device and require multi-request claims tied by constraints (works today, but higher user complexity).
+- Continue with device-local consumable capacity only (cannot model cross-device parent-scoped budgets correctly).
+
```
sunya-ch, line 552: I believe that another alternative is to extend Partitionable counters with the request policy and allow default value in consumesCounters[*].counters (to be filled by resource capacity requests). Could you please further explain why that would not cover the case?

For example,

```yaml
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  name: counter-slice
spec:
  driver: "resource-driver.example.com"
  pool:
    generation: 1
    name: "my-pool"
    resourceSliceCount: 2
  sharedCounters:
  - name: gpu-0-counter-set
     counters:
       memory: 
         value: 40Gi
         requestPolicy:
           default: "1G"
            validRange:
               min: "1M"
               max: "40Gi"
               step: "1M"
---
apiVersion: resource.k8s.io/v1
kind: ResourceSlice
metadata:
  name: device-slice
spec:
  driver: "resource-driver.example.com"
  pool:
    generation: 1
    name: "my-pool"
    resourceSliceCount: 2
  nodeName: "my-node"
  devices:
  - name: gpu-0
    consumesCounters:
    - counterSet: gpu-0-counter-set
```

One issue with the above extension is that it would require the change to partitionabledevice where Value of Counter is a required field.

sunya-ch:
> What about making the shared counters part of the scheduler core?
> 
> 
> 
> This way we could use shared capacity across different plugins and solve a few problems like native resource accounting across DRA and ResourceFit plugins: https://github.com/kubernetes/enhancements/issues/5517
> 
> 
> 
> @pravk03 

I think this could probably solve the issue of multiple drivers for different network device configuration as well. @LionelJouin

johnbelamaric:
> > What about making the shared counters part of the scheduler core?
> > 
> > 
> > 
> > This way we could use shared capacity across different plugins and solve a few problems like native resource accounting across DRA and ResourceFit plugins: https://github.com/kubernetes/enhancements/issues/5517
> > 
> > 
> > 
> > @pravk03 
> 
> I think this could probably solve the issue of multiple drivers for different network device configuration as well. @LionelJouin 

Something has to populate the global shared counters. A single owner for populating those for a given node would be critical. Multiple consumers could draw from them, that seems feasible. So we would have to figure out how to manage that. But pushing this to the framework seems possible.

An alternative that could solve the networking problem would be some way to link devices across drivers, like having the driver declare a dependency on the existence of a device from a different driver. That sounds hard though. But it mirrors the interface-on-top-of-interface structure of networking abstractions.

pravk03:
The source for populating shared counter could vary based on use-case. For the native resource case, it would be based on `node.status.allocatable`. For the multiple network drivers case, I assume it would based on information in the `ResourceSlice` ?.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
johnbelamaric: Yes, I agree. I don't see why we need a new `sharedCapacities` - we can just at `requestPolicy` to the `sharedCounters`.

We would need to split the [`Counter` struct](https://github.com/kubernetes/kubernetes/blob/8ed0e009c4fd5d5fd218239319f79bc046c362b2/staging/src/k8s.io/api/resource/v1/types.go#L567) into two separate structs: one for the `CounterSet` and one for the `DeviceCounterConsumption`. 

On the `CounterSet` side it would add `requestPolicy`. On the consumption side, it would add a way to map inbound `requests` to the counters. Something like:

```yaml
 devices:
  - name: gpu-0
    consumesCounters:
    - counterSet: gpu-0-counter-set
        memory:
          valueFrom:
            capacityKey: "dra.example.com/memory"
```

Making `valueFrom` a struct means if we ever need it, we can add a computation in there, in case 1:1 mapping is not the right thing. For example, a multiplier or something.

> One issue with the above extension is that it would require the change to partitionabledevice where Value of Counter is a required field.

We should be able to relax that requirement, and make either `value` or `valueFrom` required.

johnbelamaric (commented):
You are missing `kep.yaml` and the PRR files.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: @johnbelamaric @sunya-ch thanks for the comment!

can you please make another review and let me know if I manage to nail your intentions in the KEP now?

sunya-ch: @SchSeba Thanks for the changes! It looks good me.

johnbelamaric: yes, this captures it - I think this works for the use case and is quite flexible.

@nmn3m you should look at this - not sure you need to address it this cycle, but it will complicate your KEP

keps/sig-scheduling/5941-dra-shared-consumable-capacity/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+kep-number: 5941
+authors:
+  - "@SchSeba"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-node
+  - sig-network
+status: implementable
+creation-date: 2026-05-20
+reviewers: []
```
kannon92: We should get this filled out. @johnbelamaric anyone from sig-scheduling side that can review/approve?

keps/sig-scheduling/5941-dra-shared-consumable-capacity/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+  beta: "v1.38"
+  stable: "v1.39"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: DRASharedConsumableCapacity
+    components:
+      - kube-scheduler
+      - kubelet
```
dom4ha: Is kubelet involved in any way here?

keps/sig-scheduling/5941-dra-shared-consumable-capacity/kep.yaml
dom4ha: I can take it.

dom4ha (commented):
/approve

keps/sig-scheduling/5941-dra-shared-consumable-capacity/kep.yaml
SchSeba: you are right thanks!

SchSeba:
Hi @dom4ha thanks for the review I added you to the list :)

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+  - valid/invalid `requestPolicy` on shared counter definitions
+  - valid/invalid `valueFrom` references in device counter consumption
+  - feature-gate transition behavior for new fields
+- Scheduler allocator:
+  - aggregate accounting across multiple claims
+  - candidate rejection when shared counter set would be exceeded
+  - `requestPolicy` handling (default/range/step/values)
+  - mixed accounting when same counter has both static `value` and `valueFrom`-driven consumption
+
+- `<k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources>`: `<TBD date>` - `<TBD coverage>`
```
johnbelamaric: please add these

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+- Feature implemented behind `DRASharedConsumableCapacity`
+- Core unit and integration tests in place
+- Initial e2e signal for shared counter accounting correctness
+
+#### Beta
+
+- API shape and policy semantics settled
+- Stable periodic tests and no unresolved critical correctness bugs
+- Monitoring and troubleshooting guidance documented
```
johnbelamaric, line 345: scheduler_perf tests

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+### Upgrade / Downgrade Strategy
+
+- Upgrade with feature gate disabled preserves existing behavior (new fields ignored/dropped according to API conventions).
+- Enabling feature gate activates shared-pool validation and accounting for compatible objects.
+- Downgrade/disable support requires explicit compatibility behavior for objects containing gated fields; exact behavior to be documented with final API review.
+
+### Version Skew Strategy
+
+- Primary logic resides in apiserver and scheduler.
+- During control-plane skew, schedulers without feature support must treat gated fields as unsupported and avoid partial accounting behavior.
```
johnbelamaric: Please detail this more - it will also apply to enablement below in the PRR section.

We have three components: apiserver, the driver that publishes the slice, and the scheduler. Recommended enablement/upgrade would be apiserver first, then scheduler, and last the driver. Reverse order for downgrade / disablement.

What happens if someone screws up and upgrades/enables the driver before the scheduler? Will incorrect things schedule? Or will they fail to schedule? For an example see https://github.com/kubernetes/enhancements/pull/5987/changes#diff-b1920b064900047ebf14d76821b73b1d0224a077985348444df643ba83e6860bR1797

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+- API inspection of `ResourceSlice` objects containing `sharedCounters` with `requestPolicy` and devices with `valueFrom` in `consumesCounters`.
```
johnbelamaric, line 442: If you have many thousands of clusters this is rather painful. But I guess a metric is probably overkill, generally we don't want to have metrics for every field. We do have a metric that says if it's enabled, that doesn't mean it's in use.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+- API inspection of `ResourceClaim` allocations consuming named capacities.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [ ] API .status
+- [x] Events
+  - Event Reason: `<TBD allocation rejected due to shared capacity exhaustion>`
```
johnbelamaric: not sure this would be true. I think if we can't schedule at all, there will be an event, but it will just say that it doesn't fit on any node.

Is there anything that would appear in the status? I guess same as consumable capacity - something in the allocations with ShareID for a device with counter request policies.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+- No material regression to scheduler allocation success latency at supported scale.
+- No correctness regressions leading to persistent over-allocation or systematic false rejection under normal operation.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- [ ] Metrics
+  - Metric name: `<TBD scheduler dynamic resources allocation metrics>`
+  - Components exposing the metric: kube-scheduler
+- [x] Other
```
johnbelamaric: please lookup the appropriate metric names and add them here - you can probably check other DRA KEPs for them

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+  - Details: pending claim counts and allocation failure events for targeted classes
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+Likely yes. Candidate metrics:
+
+- shared counter allocation attempts/success/failure by reason
```
johnbelamaric: if we add metrics for these things, they also tell us it is in use. But in general this may be too granular; I would follow what we have done for example for consumable.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,534 @@
...
+- consumed vs available shared counter capacity per set (cardinality constraints permitting)
```
johnbelamaric: i don't think that we want a metric for that, but we do have #5677 which you may want to integrate with (if it's not already covered)

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: Added, thanks!

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: Updated the unit test coverage section with the actual packages from the implementation branch:

k8s.io/dynamic-resource-allocation/structured/internal/experimental
k8s.io/dynamic-resource-allocation/structured/internal/incubating
k8s.io/kubernetes/pkg/apis/resource/validation
k8s.io/kubernetes/pkg/registry/resource/resourceslice
k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources
Coverage numbers are still TBD and will be filled in once the implementation lands.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: I have reworked both the Upgrade / Downgrade Strategy and Version Skew Strategy sections following the same structure. The key points now covered:

Recommended rollout order: apiserver + scheduler gate first, then driver update
Reverse order for downgrade
Why this order matters (alpha-field stripping behavior)
Per-component skew analysis (apiserver, scheduler, kubelet, driver)
Failure mode: an older scheduler ignores valueFrom and may be overly permissive until upgraded — operationally safe as a transient state
PTAL and let me know if anything is missing.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: I have switched this to use API .status instead (following the consumable capacity KEP pattern), referencing ResourceClaim.Status.Allocation which reflects the resolved valueFrom consumption against shared counter sets. Events checkbox is now unchecked.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: Replaced the placeholder with the existing DRA scheduler metrics, following KEP-5075:

`scheduler_unschedulable_pods` with plugin="DynamicResources"
`scheduler_plugin_execution_duration_seconds` with plugin="DynamicResources"
`apiserver_request` with resource="resourceclaims"

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: Agreed. Removed the granular per-counter metrics proposals and aligned with the consumable capacity KEP approach (which answers "No" to this question). The existing DRA scheduler metrics cover our needs.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
SchSeba: I looked into KEP-5677 — it provides pool-level device count visibility via `ResourcePoolStatusRequest`, but doesn't currently track shared counter consumption (it counts devices, not counter budgets). I've updated the missing metrics answer to note that per-counter utilization metrics would be too granular for the scheduler, and referenced KEP-5677 as the right place to extend with shared counter consumption data in the future if needed.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,559 @@
...
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+- Incorrect accounting logic could cause false negatives (pending claims) or false positives (over-allocation).  
+- Already running workloads should remain running; primary impact is on new allocations.
+
+###### What specific metrics should inform a rollback?
+
+- Increased allocation failure rate attributable to shared counter checks
```
johnbelamaric, line 426: For beta, these will need to be specific metric names

johnbelamaric (commented):
/approve
/lgtm

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,560 @@
...
+```
+"root"="pci0000:00" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:17" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:3a" "localCPUs"="0,2,4,...,102" "NUMANode"=0
+"root"="pci0000:85" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+"root"="pci0000:ae" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+"root"="pci0000:d7" "localCPUs"="1,3,5,...,103" "NUMANode"=1
+```
+
+A CPU/memory DRA driver can publish one shared capacity set per NUMA node (the parent budget of available CPUs and memory) and map each PCIE root device to the appropriate set. Scheduler accounting then prevents aggregate over-allocation of CPUs or memory across PCIE roots that share the same NUMA zone.
```
kad, line 144: Well, do be precise here, it is more about `consuming from single physical package with multiple PCIe roots` rather than `share the same NUMA zone`, as NUMA is only about memory controller operational mode, and it can be multiple within physical CPU package.

kad:
With minor wording changes about NUMA and PCIe roots, it is also /lgtm from our side. Will be useful for many use cases.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,560 @@
+# KEP-NNNN: DRA Shared Consumable Capacities Across Related Devices
```
LionelJouin, line 1:
```suggestion
# KEP-5941: DRA Shared Consumable Capacities Across Related Devices
```

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,560 @@
...
+At a high level:
+
+1. A driver publishes `sharedCounters` with counters that include `requestPolicy` (default, valid range, step).
+2. Each allocatable device references the shared counter set(s) it draws from via `consumesCounters`, using `valueFrom` to map capacity request keys to counters.
+3. A claim requests capacity using existing `capacity.requests` fields.
+4. During allocation, the scheduler resolves the `valueFrom` mappings, checks and consumes the requested amount from all relevant counter sets (device-specific and/or shared), rejecting candidates that would exceed any counter limit.
+
+### User Stories
+
+#### Story 1: SR-IOV bandwidth as a shared parent resource
```
LionelJouin, line 117: This applies not only to SR-IOV VFs but also to any type of virtual network interfaces.

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,560 @@
...
+
+As a cluster user, I request one or more VFs and a bandwidth amount per request.
+As an operator, I want scheduler admission to ensure that total bandwidth promised across VFs on the same PF does not exceed PF capacity.
+
+This allows:
+
+- keeping VF as the allocatable unit, and
+- preventing over-allocation of PF bandwidth without static pre-partitioning.
+
+#### Story 2: CPU/Memory resource alignment via PCIE root grouping
```
LionelJouin, line 127: I discussed with some Telco people during Kubecon EU 2026 about the ability to allocate Cache shared between cores (Like L3 Cache) and Memory Bandwidth per CPU via, for example, Intel RDT. I don't know if story 2 is covering this?
I think @ffromani and @johnbelamaric you heard about similar use-case?

keps/sig-scheduling/5941-dra-shared-consumable-capacity/README.md
```diff
@@ -0,0 +1,560 @@
...
+
+- Compare allocation latency/error rates before and after enablement.
+- Inspect pending claims and failure reasons tied to shared capacity checks.
+- Disable feature gate as rollback if correctness or latency regression is severe.
+
+## Implementation History
+
+- 2026-03-03: Initial draft created in local design docs, generalized from SR-IOV-specific proposal.
+
+## Drawbacks
```
LionelJouin, line 539: I feel there would be more confusion with `capacity` and `counters` now as counter's behavior is changing a bit to be similar to capacity.

LionelJouin:
This /lgtm, great job @SchSeba!

> > > What about making the shared counters part of the scheduler core?
> > > This way we could use shared capacity across different plugins and solve a few problems like native resource accounting across DRA and ResourceFit plugins: #5517
> > > @pravk03
> > 
> > 
> > I think this could probably solve the issue of multiple drivers for different network device configuration as well. @LionelJouin
> 
> Something has to populate the global shared counters. A single owner for populating those for a given node would be critical. Multiple consumers could draw from them, that seems feasible. So we would have to figure out how to manage that. But pushing this to the framework seems possible.
> 
> An alternative that could solve the networking problem would be some way to link devices across drivers, like having the driver declare a dependency on the existence of a device from a different driver. That sounds hard though. But it mirrors the interface-on-top-of-interface structure of networking abstractions.

We discussed around this with @sunya-ch and @SchSeba, I think this is more a complete new KEP. Yes, it seemed quite hard to model and to understand who would be the first owner driver and how these resources would be exposed.

johnbelamaric:
@SchSeba given that it is late in your time zone and the deadline is in < 6 hours, I am going to lgtm this. The remaining comments look non-blocking to me.

Please follow up with a PR that addresses @kad's comments and any other outstanding minor issues.

/lgtm

Issue #4322, opened 2023-10-31
Cluster Inventory API for multicluster management


qiujian16:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Introduce the concept of cluster inventory to define a kubernetes cluster for multiple cluster management
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/4323
- Discussion Link: https://docs.google.com/document/d/1sUWbe81BTclQ4Uax3flnCoKtEWngH-JA9MyCqljJCBM/edit#heading=h.wyy5e36qm908
- Primary contact (assignee): @qiujian16
- Responsible SIGs: sig-multicluster
- Enhancement target (which target equals to which milestone):
  - Alpha release target (x.y): TK
  - Beta release target (x.y):
  - Stable release target (x.y):
- [ ] Alpha
  - [ ] KEP (`k/enhancements`) update PR(s): TK
  - [ ] Code (`k/k`) update PR(s):
  - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #5947, opened 2026-03-05
KEP-4322: Bump implemented for Cluster Inventory API parent KEP

lauralorenz:
- One-line PR description: Trying to update against reality since the alpha project is implemented in sigs.k8s.io/cluster-inventory-api

- Issue link: #4322

- Other comments: N/A

### DIFF
--- a/keps/sig-multicluster/4322-cluster-inventory/kep.yaml
+++ b/keps/sig-multicluster/4322-cluster-inventory/kep.yaml
@@ -5,7 +5,7 @@ authors:
   - "@dixudx"
   - "@mikeshng"
 owning-sig: sig-multicluster
-status: provisional
+status: implemented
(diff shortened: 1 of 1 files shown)


lauralorenz:
I see that changing this, though more representative to reality, does force some prow checks against PRR that we will have to handle to get it up to date (PRR itself being a special case for this since it's out of tree). Will need to discuss with authors and SIG

lauralorenz:
It looks like several of the SIG-MC KEPs have this pattern where they have not set any changes to `status` beyond `implementable`. In the grand scheme of things since this kep.yaml file doesn't seem to be consumed by anything except prow for checking this type of fact and we don't think it's an important discoverability mechanism for humans right now, I'm going to close this and leave the kep.yaml as is.

PR #5949, opened 2026-03-09
KEP-5732: Remove desiredCount from the TAS KEP

44past4:
- One-line PR description: Removes adding desiredCount field to Basic and Gang PodGroup scheduling policies from the scope of TAS KEP.

- Issue link: https://github.com/kubernetes/enhancements/issues/5732

- Other comments:

  - After some discussions there is no clarity when it comes to expected semantics of the desiredCount field which was planned to be added to Basic and Gang scheduling policies on PodGroup. Because of this desiredCount will be removed from the alpha scope of the TAS KEP.

/sig scheduling

### DIFF
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -445,6 +445,14 @@ The algorithm proceeds in three main phases for a given PodGroup.
 - **Potential Optimization:** Pre-filtering can check aggregate resources
   requested by PodGroup Pods before running the full simulation.
 
+- **Basic Scheduling Policy Handling:** The current algorithm may exhibit
+  inconsistent behavior when used with the PodGroup Basic Scheduling Policy.
+  Because the scheduler may only observe a subset of pods when scheduling
+  a PodGroup, placement feasibility is only validated for those specific
+  pods rather than the entire group. This limitation may be addressed in
+  future releases; currently, scheduling gates may be used as a partial
+  mitigation.
+
 - **Heterogeneous PodGroup Handling**: Sequential processing will be used
   initially. Pods are processed sequentially; if any fail, the placement is
   rejected.


keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -281,78 +281,6 @@ will be defined in a separate KEP:
...
-
-**1. Basic Policy Update**
-
-We introduce `desiredCount` to the `Basic` policy to allow users to signal the
-expected group size for optimization purposes.
-
-```go
-// BasicSchedulingPolicy indicates that standard Kubernetes
-// scheduling behavior should be used.
-type BasicSchedulingPolicy struct {
```
mm4tt, line 303: We know that TAS for non-gangs is tricky due to edge cases (e.g., the scheduler making placement decisions based on a partial view of the pods). I treat this as an (unsuccessful) attempt to fix the problem. While removing it now makes sense until we have a more thorough design,  I think we shouldn't just drop the context entirely.

Instead of simply removing it, should we add a short paragraph to the KEP outlining the current TAS limitations for basic policies, noting that this will be addressed in future releases? We could also mention scheduling gates as a currently available (albeit imperfect) mitigation.

The 'Phase 2' algorithm section might be a good place for this. We could expand on point '3. If all pods fit, the Placement is marked Feasible' to explain why this works perfectly for gangs (waiting for MinCount in pre-enqueue), but might behave inconsistently for basic policies (where the scheduler's decision depends on the number of observed pods).

dom4ha: > Instead of simply removing it, should we add a short paragraph to the KEP outlining the current TAS limitations for basic policies, noting that this will be addressed in future releases? We could also mention scheduling gates as a currently available (albeit imperfect) mitigation.

+1, but I think only the problem of "not-ready" pods is worth addressing as a part of TAS as it makes the feature working sub-optimally.

Addressing the desired future number of pods is rather an independent extension to TAS and IMHO deserves a separate KEP.

dom4ha:
FTR, dropped PR with the discussion: https://github.com/kubernetes/kubernetes/pull/137137

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have updated Phase 2 algorithm description with a point explaining the limitations of using TAS with Basic Scheduling Policy. Please take a look.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -445,6 +445,14 @@ The algorithm proceeds in three main phases for a given PodGroup.
 - **Potential Optimization:** Pre-filtering can check aggregate resources
   requested by PodGroup Pods before running the full simulation.
 
+- **Basic Scheduling Policy Handling:** The current algorithm may exhibit
```
dom4ha, line 448: The same applies to Gang Policy if more pods than minCount are created.

44past4: Yes, this is true that we have a similar problem with Gang Policy but in case of Gang Policy we have explicit information on minCount coming from the user and it is their conscious decision if they are setting minCount to lower number compared to total number of pods. So in the context of Gang Policy TAS adheres to explicit user intent and makes sure that it is fulfilled. For Basic scheduling policy user may expect some best effort behaviour from TAS which in many cases will simply not be true so this disclaimer makes sense for me for Basic Policy but not necessary for Gang Policy.

PR #5956, opened 2026-03-10
KEP-4671: Add performance tests to the beta graduation criteria

tosi3k:
#### One-line PR description
Add performance tests as a beta graduation criterion for the gang scheduling support in Kubernetes.

#### Issue link
https://github.com/kubernetes/enhancements/issues/4671

#### Other comments
Follow-up from a discussion in https://github.com/kubernetes/kubernetes/pull/136976#discussion_r2833884236.


dom4ha:
/approve

Performance tests for gangs are definitely needed. I agree with @tosi3k in https://github.com/kubernetes/kubernetes/pull/136976#discussion_r2833884236 that it does not mean we need to have a limit of minCount

PR #5968, opened 2026-03-19
KEP-127: Mark as implemented

rata:
Code PR moving to GA is merged already:
	https://github.com/kubernetes/kubernetes/pull/136792

Let's mark this as implemented.

  
- One-line PR description: KEP is implemented already, let's mark it as such here too.

- Issue link: #127 

- Other comments:

### DIFF
--- a/keps/sig-node/127-user-namespaces/kep.yaml
+++ b/keps/sig-node/127-user-namespaces/kep.yaml
@@ -6,7 +6,7 @@ authors:
   - "@saschagrunert"
 owning-sig: sig-node
 participating-sigs: []
-status: implementable
+status: implemented
 creation-date: 2021-11-02
 reviewers:
   - "@SergeyKanzhelev" # sig-node-assigned-reviewer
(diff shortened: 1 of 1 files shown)


rata:
cc @giuseppe

keps/sig-node/127-user-namespaces/kep.yaml
SergeyKanzhelev: make it false

SergeyKanzhelev:
if you can also update imlpementation history in readme, it will be great

keps/sig-node/127-user-namespaces/kep.yaml
rata: Oh, I doubted because most other PRs switching to implemented didn't change this. Updated now! :)

rata:
@SergeyKanzhelev Fixed that. PTAL.

The implementation history is already up to date (1.36 marked as the release it went GA): https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/127-user-namespaces/README.md#implementation-history

Issue #5970, opened 2026-03-23
DRA Driver Conformance Test Suite


asahay19:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add an upstream conformance test suite for validating DRA device drivers against the minimum DRA plugin contract
- Kubernetes Enhancement Proposal: (PR will be linked once filed)
- Discussion Links:
  - https://github.com/kubernetes/kubernetes/tree/master/test/e2e/dra (existing DRA e2e tests)
  - https://github.com/kubernetes-sigs/dra-example-driver (reference driver implementation)
  - https://github.com/kubernetes-csi/csi-test (prior art: CSI conformance testing)

/sig node
/sig testing
/wg device-management

- PRs by stage and milestone:

- [ ] Alpha - v1.37
  - [ ] KEP (`k/enhancements`) update PR(s):
  - [ ] Code (`k/k`) update PR(s):
  - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

### Motivation

DRA graduated to GA in Kubernetes v1.35 (`resource.k8s.io/v1`), and device vendors (NVIDIA, Intel, AMD, etc.) are actively building DRA drivers. However, there is no standardized way for a vendor to validate that their driver correctly implements the DRA plugin contract.

The existing 4 conformance tests (added in v1.35) only validate API-level CRUD operations for DRA resource types. They do not exercise the driver plugin interface at all. Meanwhile, vendors are writing ad-hoc validation tests independently (e.g., [NVIDIA/k8s-dra-driver-gpu#782](https://github.com/NVIDIA/k8s-dra-driver-gpu/issues/782)).

This enhancement proposes an upstream, driver-agnostic conformance test suite that validates the minimum DRA plugin contract: driver registration, ResourceSlice publishing, device allocation, NodePrepareResources, pod startup with device access, NodeUnprepareResources, and resource reuse.

### Scope

**In scope:**
- Minimum conformance behaviors (registration, allocation, prepare/unprepare, pod lifecycle, cleanup)
- Driver-agnostic test suite accepting driver name and DeviceClass as input
- Validation against the upstream fake test driver and dra-example-driver

cc @sairameshv @haircommander

pohly:
[...]

PR #5971, opened 2026-03-23
KEP-5970: Add DRA driver conformance test suite proposal

asahay19:
- One-line PR description: Add KEP-5970 for DRA driver conformance test suite
- Issue link: https://github.com/kubernetes/enhancements/issues/5970
- Other comments:

@sairameshv @haircommander

### DIFF
--- a/keps/sig-node/5970-dra-driver-conformance-test-suite/README.md
+++ b/keps/sig-node/5970-dra-driver-conformance-test-suite/README.md
@@ -1,0 +1,525 @@
+# KEP-5970: DRA Driver Conformance Test Suite
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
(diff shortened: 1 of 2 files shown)


dims:
@asahay19 please avoid the word `conformance`, please use `validation` or something else. Conformance has a special meaning in our community/ecosystem having to do with trademarks etc.

- https://github.com/cncf/k8s-conformance
- https://github.com/cncf/k8s-ai-conformance

haircommander:
I find myself wondering if this needs a KEP? we definitely want a forum to find consensus but I don't imagine this suite will follow the typical KEP lifecycle.

@dims FWIW: there have been discussions with wg-ai-conformance about including a suite to verify the drivers a distribution has are "conformant" so this potentially could evolve into a proper conformance program, but it needn't to start.

BTW @asahay19 we should probably discuss with wg-ai-conformance and wg-device-management as well to make sure we're all on the same page about what the direction of this suite is. for instance, at maintainer summit yesterday, there was a handwavy proposal about having drivers expose capabilities, and then the conformance suite would test capabilities. if a driver says "I support partitionable devices" then we could test that, but not if not. That could be a good way to branch on the various features of drivers. that all said, I think the capabilities you've described are a great start.

asahay19:
Thanks both of you for your feedback. Really helpful. I am closing this KEP now, agreed this doesn't need a KEP lifecycle.

PR #5974, opened 2026-03-23
Introduce KEP 5975: Declarative API Definitions

jpbetz:
- Issue link: https://github.com/kubernetes/enhancements/issues/5975


jpbetz:
@thockin @deads2k @liggitt Looking for early feedback.  If there is support I'd like to merge early and work on this in the off cycle.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,413 @@
...
+- **Feature-gate field dropping.** A `DropDisabledFields` hook called at the
+  correct lifecycle point for both create and update.
+- **Behavioral overrides.** Function-valued fields for custom normalization,
+  warnings, generation tracking, and status-specific behavior.
+
+### Spec/Status Accessors
+
+The strategy needs to copy, clear, and compare `Spec` and `Status` fields
+without knowing their concrete types. Rather than relying on reflection, we would like to
+introduce two interfaces:
```
jpbetz: I'm not fully convinced this is how I'd like to generically implement wiping.  One alternative is reflection.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,634 @@
...
+be introduced in a fully backward-compatible way.
+
+Strategy configs provide three categories of configuration:
+
+**Resource properties**:
+
+| Field | Default | Purpose |
+|-------|---------|---------|
+| `Namespaced` | `false` | Whether the resource is namespace-scoped |
+| `AllowCreateOnUpdate` | `false` | Whether PUT can create a non-existent object |
```
thockin: do we actually want anyone to set this?

jpbetz:
Note to self: Tests that verify DeepCopy, Conversion, and all other generators are run for each type would eliminate another aspect of manual API Review.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,601 @@
...
+```yaml
+apiVersion: apidefinitions.k8s.io/v1alpha1
+kind: GroupVersionRegistration
+
+group: admission.k8s.io
+version: v1
+externalTypes: k8s.io/api/admission/v1
+peers:
+- k8s.io/kubernetes/pkg/apis/admission
+```
```
jpbetz: @liggitt @thockin To simplify code generator management, I'm looking a way of enabling all code-generators by-default for in-tree APIs.

Initially I just tried adding a enforcement check that complained if an API author doesn't add all the `+k8s:deepcopy-gen=true` type tags, but @thockin mentioned look into driving the generators of small declarations.  So I'm trying it out.  This section shows what it would roughly look like.

Questions:
- Do we have an appetite for this sort of change? It will consume review bandwidth. The alternative would be to just merge something like https://github.com/kubernetes/kubernetes/pull/138577.
- Are we okay with something where each directory that does API definition stuff has a file? Or do we want to take this in another direction (single file for an entire API resource?!)

jpbetz:
@alvaroaleman does kubebuilder depend on any of the `+k8s:deepcopy=package` or other generator enablement tags in `doc.go`?

My intuition is that this propose would not impact kubebuilder, but want to double check.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
deads2k: I'm ok with a file that describes which generator to run and how. Sounds similar to https://github.com/openshift/api/blob/master/example/.codegen.yaml in OpenShift.  Some issues we've had:
1. opt-in or opt-out or both.  As the generators grow, it is sort of like admission. We have new items that contain default-on or default-off preferences and each API can have an "explicit off" or "explicit on" preference.
2. Configuring at group level, across all versions made for fewer files, but more painful changes when we needed to start bringing new generators online. 
3. This became a natural place for validation and linting exceptions too.  Not sure about your plan for growth there.

@JoelSpeed migrated us to the file and has been more hands on dealing with it recently.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,597 @@
...
+In the future, validation errors and warnings may be output in a single
+validation pass, but that is not a goal for this KEP.
+
+### Field Wiping and Fields Resetting
+
+Default behavior will follow best practices:
+
+- **Main strategy**: clears status on create (`obj.Status = TypeStatus{}`),
+  and clears status changes on update (`new.Status = old.Status`).
+- **Status substrategy**: clears spec, labels, and annotations changes
```
deads2k: Would we phrase this to clear all new fields as they're added unless we take active effort?  I'd like that so we're forced to think about it.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,597 @@
...
+This requires [Spec/Status accessor interfaces](#specstatus-accessor-interfaces).
+
+### Generation Management
+
+Default behavior will follow best practice: Generation is set to 1 on create
+and bumped on update when Spec changes.
+
+This requires [Spec/Status accessor interfaces](#specstatus-accessor-interfaces).
+
+### Feature-Gate Field Dropping
```
deads2k, line 247: This probably the most valuable part for me personally.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+When `StrategyConfig` is set, the provided configuration customizes strategy.
+
+### strategy.Config
+
+`strategy.Config` provides fields and hooks for custom strategy behavior.
+
+**Hooks:**
+
+| Hook | Purpose |
```
deads2k, line 325: no Hook for PrepareForCreate or PrepareForUpdate?

deads2k:
A few questions and comments.  I like the idea.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: The idea we'd like to start with is:  

- Here's a file that says everything code generators need to know about the API declaring files in in this directory
- When present, the generators run in a way that conforms to our API guidelines

I've got a PoC that shows this working against all in-tree APIs correctly and removes the need for a bunch of generator enablement tags in each doc.go file.

I do expect this to grow. I don't want it to accumulate validation/lint exceptions here.

For ecosystem adoption, I think it's fine to use outside of k/k. I would like it to focus on opinions we have for in-tree APIs. So if 3rd parties use it, they need to buy-in to our opinions.

alvaroaleman:
> @alvaroaleman does kubebuilder depend on any of the +k8s:deepcopy=package or other generator enablement tags in doc.go?

It depends on what you mean by "depends on", it does interpret them but that probably doesn't matter for in-tree resources as I don't think there is a good reason to use it to generate DeepCopies for in-tree resources: https://github.com/kubernetes-sigs/controller-tools/blob/e17f046d11b6e9e508d11abc37891b66831aefa4/pkg/deepcopy/gen.go#L45-L47

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: Yes, good point. I'll rephrase this.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: Yes, I'll call it out.

jpbetz:
> > @alvaroaleman does kubebuilder depend on any of the +k8s:deepcopy=package or other generator enablement tags in doc.go?
> 
> It depends on what you mean by "depends on", it does interpret them but that probably doesn't matter for in-tree resources as I don't think there is a good reason to use it to generate DeepCopies for in-tree resources: [kubernetes-sigs/controller-tools@`e17f046`/pkg/deepcopy/gen.go#L45-L47](https://github.com/kubernetes-sigs/controller-tools/blob/e17f046d11b6e9e508d11abc37891b66831aefa4/pkg/deepcopy/gen.go#L45-L47)

:+1:

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
JoelSpeed: > So if 3rd parties use it, they need to buy-in to our opinions.

There's already a vast ecosystem of 3rd parties using the existing generators, I assume a goal here should be that those existing 3rd parties should be able to continue using the generators as is?

jpbetz: Correct. Existing generator usage is unchanged. All flags/args/tags will continue to be supported.  So anyone using generators on their own types, or even against Kubernetes staging types (e.g. importing core/v1) will be able to continue to do so.

But are adding a [new opt-in way to configure the generators](https://github.com/kubernetes/kubernetes/pull/138657) (apigroup.yaml files).  And k/k will opt-in.

We're going to phase the change in. But this does mean that the doc.go files in the kubernetes project will eventually change. Any tooling outside of kubernetes that depends on tags like `+k8s:defaulter-gen=true` and similar generator enablement declarations in doc.go will be impacted in a future release when we remove the enablement tags.  We're going to keep the tags around for at least 1 release to allow for a transition.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,610 @@
...
+cross-cutting blackbox test to ensure that all APIs adhere to field wiping
+conventions. We will expand approach with testing, linting, and exception
+lists to cover:
+
+- ResetObjectMetaForStatus was used to reset metadata
+  - https://github.com/kubernetes/kubernetes/pull/137689 tests metadata wiping but does not ensure that ResetObjectMetaForStatus was used
+- Generation management (set to 1 on create and monotomically increased when spec is changed by an update)
+- Field dropping of feature gated fields
+  - It should be impossible to add a new field to an existing API without a feature gate 
+  - Iff a field is feature gated, field dropping must be implemented
```
sttts: not sure whether the iff = "if and only if" should tell me something?

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,610 @@
...
+
+### Goals
+
+- Make it impossible to author APIs that violate best practices without
+  being granted an exception by an API reviewer.
+- Eliminate the need to hand-write strategy boilerplate for resources that
+  follow standard conventions while preserving the customization that is
+  available today.
+
+### Non-Goals
```
sttts, line 109: is a goal to externalize the tooling? Other projects use our code generators. Would be good to be clear here.

I think it would be good if we simplified kube-core-like API definitions elsewhere. But even if not, we cannot compromise how our code generators are used. So in either case, we have to have the later as a goal, i.e. do not break our users.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
sttts: better put that into the goals, had commented there the same: don't break external consumers.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,610 @@
...
+  new metadata fields, are cleared unless we take active effort to allow status writes to them.
+
+Managed fields are reset to match the fields that are wiped.
+
+This requires [Spec/Status accessor interfaces](#specstatus-accessor-interfaces).
+
+### Generation Management
+
+Default behavior will follow best practice: Generation is set to 1 on create
+and bumped on update when Spec changes.
```
sttts: which means we align to CRD behaviour aka label+annotations etc. have no influence:

```suggestion
and bumped on update exactly when Spec changes.
```

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
```diff
@@ -0,0 +1,610 @@
...
+    // +k8s:minimum=0
+    // +k8s:maximum=1000
+    Priority *int32 `json:"priority,omitempty"`
+}
+```
+
+The appropriate field dropping is generated automatically when the feature gate is provided.
+
+Declarative validation also automatically handles proper validation of the field
+when the feature gate is off (i.e. in-use detection and ratcheting).
```
sttts, line 287: have not looked deeper: does this mean we enabling declarative validation where we (or external consumers) didn't in the past?

deads2k:
/lgtm
/hold

Looks like Stefan has some outstanding questions.  I'm expecting them to be easy to tidy up in the doc.  Let me know when it's ready again.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
deads2k: Wait, I skated over this.  Wasn't this going to be removed?

deads2k:
/lgtm cancel

one section I expected removed as still there.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: We're going to retain full backward compatibility and treat all changes as external. I'll clarify.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: I'll drop the IFF and clarify what is expected for non-feature gated code.

keps/sig-api-machinery/5975-declarative-api-definitions/README.md
jpbetz: Yes, in 1.36 and earlier, we were being careful in how we rolled out declarative validation and only wired it in and enabled it on select APIs. In the last few weeks we've turned it on for all APIs such that if a `+k8s:{validation-rule}` tag is used anywhere in the built-in types, it is actuated.

jpbetz:
Thanks @deads2k for the pass. I've removed the reference to yaml files and responded to all feedback.

PR #5976, opened 2026-03-23
KEP-5729: Update ResourceClaim Support for Workloads for beta in 1.37

nojnhuh:
- One-line PR description: Update KEP-5729 (DRA: ResourceClaim Support for Workloads) for beta promotion in Kubernetes 1.37.

- Issue link: https://github.com/kubernetes/enhancements/issues/5729

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-scheduling/5729.yaml
+++ b/keps/prod-readiness/sig-scheduling/5729.yaml
@@ -1,3 +1,5 @@
 kep-number: 5729
 alpha:
   approver: "@wojtek-t"
+beta:
+  approver: "@wojtek-t"
--- a/keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
+++ b/keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
@@ -1202,6 +1202,19 @@ PodGroup. It will do this even if there is already a PodGroup reference in the
 `status.reservedFor` list. This leads to the challenge described in the previous
 section.
 
+If the API server is on a version that supports the feature, but
+kube-controller-manager is not, then the ResourceClaim controller may observe
+PodGroups that define `spec.resourceClaims`. When Pods contain matching claims,
+the intent is that those claims are generated for the PodGroup instead of each
+Pod. Even when this feature is disabled, the ResourceClaim controller will check
+a Pod's claims against its PodGroup. If the controller _would have_ created a
+ResourceClaim for the PodGroup _if_ the feature _was_ enabled, then it will
+return an error. Users are expected to restart kube-controller-manager with the
+feature enabled to generate a ResourceClaim for the PodGroup. If the user
+intended to generate a ResourceClaim for that Pod, then the user has to recreate
+the PodGroup without the resource claim and all of its member Pods with the
+resource claim.
+
 ## Production Readiness Review Questionnaire
 
 <!--
--- a/keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
+++ b/keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
@@ -137,7 +137,6 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Alternatives](#alternatives)
   - [Increase the size limit on the <code>status.reservedFor</code> field](#increase-the-size-limit-on-the-statusreservedfor-field)
   - [Allow ResourceClaims to be reserved for any object](#allow-resourceclaims-to-be-reserved-for-any-object)
-- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
 <!-- /toc -->
 
 ## Release Signoff Checklist
@@ -419,8 +418,8 @@ The following API changes will be made:
 
 - The Workload and PodGroup APIs will be updated to include references to
   ResourceClaims and ResourceClaimTemplates, like Pods.
-- The Pod API will be updated to include references to claims listed in its
-  PodGroup.
+- The Pod API will include new semantics for the existing API fields to refer to
+  claims made by its PodGroup.
 
 #### Workload
 
@@ -432,21 +431,27 @@ type PodGroupTemplate struct {
 	...
 
 	// ResourceClaims defines which ResourceClaims may be shared among Pods in
-	// the group. Pods must reference these claims in order to consume the
-	// allocated devices.
+	// the group. Pods consume the devices allocated to a PodGroup's claim by
+	// defining a claim in its own Spec.ResourceClaims that matches the
+	// PodGroup's claim exactly. The claim must have the same name and refer to
+	// the same ResourceClaim or ResourceClaimTemplate.
 	//
 	// This is an alpha-level field and requires that the
-	// WorkloadPodGroupResourceClaimTemplate feature gate is enabled.
+	// DRAWorkloadResourceClaims feature gate is enabled.
 	//
 	// This field is immutable.
 	//
+	// +optional
 	// +patchMergeKey=name
 	// +patchStrategy=merge,retainKeys
 	// +listType=map
 	// +listMapKey=name
-	// +featureGate=WorkloadPodGroupResourceClaimTemplate
-	// +optional
-	ResourceClaims []PodGroupResourceClaim `json:"resourceClaims,omitempty"`
+	// +k8s:optional
+	// +k8s:listType=map
+	// +k8s:listMapKey=name
+	// +k8s:maxItems=4
+	// +featureGate=DRAWorkloadResourceClaims
+	ResourceClaims []PodGroupResourceClaim `json:"resourceClaims,omitempty" patchStrategy:"merge,retainKeys" patchMergeKey:"name"`
 }
 
 // PodGroupResourceClaim references exactly one ResourceClaim, either directly
@@ -454,11 +459,17 @@ type PodGroupTemplate struct {
 // for the PodGroup.
 //
 // It adds a name to it that uniquely identifies the ResourceClaim inside the PodGroup.
-// Pods that need access to the ResourceClaim reference it with this name.
+// Pods that need access to the ResourceClaim define a matching reference in its
+// own Spec.ResourceClaims. The Pod's claim must match all fields of the
+// PodGroup's claim exactly.
 type PodGroupResourceClaim struct {
 	// Name uniquely identifies this resource claim inside the PodGroup.
 	// This must be a DNS_LABEL.
-	Name string `json:"name"`
+	//
+	// +required
+	// +k8s:required
+	// +k8s:format=k8s-short-name
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
 
 	// ResourceClaimName is the name of a ResourceClaim object in the same
 	// namespace as this PodGroup. The ResourceClaim will be reserved for the
@@ -466,6 +477,11 @@ type PodGroupResourceClaim struct {
 	//
 	// Exactly one of ResourceClaimName and ResourceClaimTemplateName must
 	// be set.
+	//
+	// +optional
+	// +k8s:optional
+	// +k8s:unionMember
+	// +k8s:format=k8s-long-name
 	ResourceClaimName *string `json:"resourceClaimName,omitempty"`
 
 	// ResourceClaimTemplateName is the name of a ResourceClaimTemplate
@@ -475,7 +491,7 @@ type PodGroupResourceClaim struct {
 	// be bound to this PodGroup. When this PodGroup is deleted, the ResourceClaim
 	// will also be deleted. The PodGroup name and resource name, along with a
 	// generated component, will be used to form a unique name for the
-	// ResourceClaim, which will be recorded in pod.status.resourceClaimStatuses.
+	// ResourceClaim, which will be recorded in podgroup.status.resourceClaimStatuses.
 	//
 	// This field is immutable and no changes will be made to the
 	// corresponding ResourceClaim by the control plane after creating the
@@ -483,6 +499,11 @@ type PodGroupResourceClaim struct {
 	//
 	// Exactly one of ResourceClaimName and ResourceClaimTemplateName must
 	// be set.
+	//
+	// +optional
+	// +k8s:optional
+	// +k8s:unionMember
+	// +k8s:format=k8s-long-name
 	ResourceClaimTemplateName *string `json:"resourceClaimTemplateName,omitempty"`
 }
 ```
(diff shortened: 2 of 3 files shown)


keps/sig-scheduling/5729-resourceclaim-support-for-workloads/kep.yaml
```diff
@@ -7,7 +7,7 @@ authors:
 owning-sig: sig-scheduling
 participating-sigs:
   - sig-node # ResourceClaim controller
-status: implementable
+status: implemented
```
pohly: I think "implemented" is reserved for features that are completely implemented, i.e. GA.

pohly (commented):
As documentation for what was in 1.36 this is ready for merging, with the caveat about "implemented".

We were discussing changes the behavior when kube-controller-manager has the feature off (default in 1.36) and is running with an apiserver where it's on (say, on the transition to default enabled). We could include that change in this PR in a second commit, then we only need to review one PR - up to you.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/kep.yaml
nojnhuh: Got it, updated.

nojnhuh:
> We were discussing changes the behavior when kube-controller-manager has the feature off (default in 1.36) and is running with an apiserver where it's on (say, on the transition to default enabled). We could include that change in this PR in a second commit, then we only need to review one PR - up to you.

Added as a second commit here.

nojnhuh:
I updated this to include the required beta criteria in hopes of promoting this for 1.37.

/retitle KEP-5729: Update for beta in 1.37

nojnhuh:
/retitle KEP-5729: Update ResourceClaim Support for Workloads for beta in 1.37

keps/prod-readiness/sig-scheduling/5729.yaml
```diff
@@ -1,3 +1,5 @@
 kep-number: 5729
 alpha:
   approver: "@wojtek-t"
+beta:
+  approver: "@wojtek-t"
```
nojnhuh, line 5: FYI @wojtek

nojnhuh: Sorry, wrong Wojtek.

@wojtek-t

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -742,27 +843,28 @@ deleting the ResourceClaim once its owning PodGroup is deleted.
 Generated and standalone ResourceClaims referenced by a PodGroup remain
 unallocated until kube-scheduler allocates the ResourceClaim by setting
 `status.allocation` for the first Pod in the PodGroup that references the
-PodGroup's claim. When a Pod's claim is requested through
-`podGroupResourceClaim`, the ResourceClaim's `status.reservedFor` list will
+PodGroup's claim. When a Pod's claim matches a claim made by its PodGroup,
```
nojnhuh, line 843: @mm4tt @wojtek-t Is the PodGroup scheduling cycle ready to be plugged into with things like DRA for 1.37? Should we reconsider changing the ResourceClaim from being allocated with the first Pod to being allocated with the PodGroup?

Pros of allocating with the first Pod:
- Resources stay unallocated until a Pod is created which uses the claim. Useful if PodGroups are created with 0 replicas and stay that way for a while.
- That's what we already do :)

Pros of allocating with the PodGroup:
- Eliminate contention between every Pod in the group from making the same update to the same ResourceClaim during binding. Almost certainly more performant and reliable.
  - This contention already surfaces a bug being addressed in https://github.com/kubernetes/kubernetes/pull/138775.

cc @johnbelamaric

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -432,40 +431,58 @@ type PodGroupTemplate struct {
...
 	// +listType=map
 	// +listMapKey=name
-	// +featureGate=WorkloadPodGroupResourceClaimTemplate
-	// +optional
-	ResourceClaims []PodGroupResourceClaim `json:"resourceClaims,omitempty"`
+	// +k8s:optional
+	// +k8s:listType=map
+	// +k8s:listMapKey=name
+	// +k8s:maxItems=4
+	// +k8s:alpha(since:"1.36")=+k8s:immutable
```
yongruilin: `+k8s:alpha` is used for shadowing purpose, mostly used when migrating from handwritten validation to declarative validation.
I don't think we need this.
```suggestion
	// +k8s:immutable
```

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -497,50 +519,132 @@ type PodGroupSpec struct {
...
```
yongruilin: ditto

nojnhuh: Updated, thanks!

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: Actually I see this was never added to the actual type, so I'll remove it here: https://github.com/kubernetes/kubernetes/blob/5ffa4a018b34d0c0a438891cfe71834b10be4e0a/pkg/apis/scheduling/types.go#L208-L214

I'd like to be able to mark this as `+k8s:immutable`, but I see the DV unit tests fail:
```
% go test ./test/declarative_validation/scheduling/workload 
PASS
1 uncovered declarative-validation rules:
scheduling.k8s.io/v1alpha3, Kind=Workload:
  spec.podGroupTemplates[*].resourceClaims  FieldValueInvalid origin="immutable"

FAIL	k8s.io/kubernetes/test/declarative_validation/scheduling/workload	0.241s
FAIL
```
Existing test cases modify the `spec.podGroupTemplates[].resourceClaims` field and expect errors pointing to `spec.podGroupTemplates` being immutable. Marking the `resourceClaims` field as immutable now would be good to prevent the field from accidentally becoming mutable if `spec.podGroupTemplates` ever becomes mutable. Is this a case where it would be appropriate to add an exception in [`coverage-allowlist.yaml`](https://github.com/kubernetes/kubernetes/blob/5ffa4a018b34d0c0a438891cfe71834b10be4e0a/test/declarative_validation/coverage-allowlist.yaml) to allow that exact rule not to be triggered by a test?

yongruilin: `+k8s:alpha` is used for migration like this https://github.com/kubernetes/kubernetes/blob/b2346bcea5e823a57d22c7db60df28bb4738695c/staging/src/k8s.io/api/rbac/v1/types.go#L53

I didn't look carefully, I think we can drop the `+k8s:immutable` here because the parent field `WorkloadSpec.PodGroupTemplates` already have a +k8s:immutable tag. It already cover the immutability for all the underlying fields.

For more context: +k8s:immutable is a "short-circuit" validation, which means it won't exercise the following validation if the immutability check fails on the parent `spec.podGroupTemplates`.

nojnhuh: The idea is to add `+k8s:immutable` now (even though it's redundant) to protect the `resourceClaims` field from unintentionally becoming mutable if that marker is dropped from the parent field in the future.

yongruilin: Dropping DV tag on types.go is changing the validation behavior which would require API a strict and comprehensive API review.

It could be an exception if the `PodTemplate` is also imported and used in another mutable type, but currently it is only imported by the immutable `WorkloadSpec.PodGroupTemplates`. So I don't think we should keep the redundant +k8s:immutable here.

nojnhuh: Makes sense, thanks!

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -1241,12 +1399,19 @@ Longer term, we may want to require automated upgrade/rollback tests, but we
 are missing a bunch of machinery and tooling and can't do that now.
 -->
 
+New automated upgrade/downgrade tests will verify:
```
nojnhuh: I started working on a test for this but I'm not sure if it's feasible to implement now since 1.36 serves only the `v1alpha2` version of the PodGroup API where 1.37 instead serves `v1alpha3` (and eventually `v1beta1`). Do we have to wait until both the old and new versions serve the same API version?

Is there some other functionality we could test here that doesn't depend on the PodGroup persisting across upgrade/downgrade?

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
dom4ha: > Should we reconsider changing the ResourceClaim from being allocated with the first Pod to being allocated with the PodGroup?

I'd treat it as an implementation detail. I assume that we won't change the behavior even after introducing PodGroup level extension points.

In 1.38 we may consider the migration, which should give us support for ResourceClaim in CompositePodGroup for free. 

@wojtek-t @mm4tt

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -793,18 +895,18 @@ reference a PodGroup that has been allocated a claim, but haven't
...
-   run the pod.
--->
-
 1. If the DRA scheduler plugin is trying to find candidates for deallocation in
    the `PostFilter` function and sees a ResourceClaim with a non-Pod reference,
    it will not attempt to deallocate. The plugin has no way to know how many
    Pods are actually consuming the ResourceClaim without the explicit list in
    `status.reservedFor` list and therefore it will not be safe to deallocate.
 
+     - Now the DRA plugin reuses the scheduler's internal view of PodGroups to
```
dom4ha: We are introducing PodGroup level PostFilter in Workload Aware Preemption KEP (beta) and disable pod-by-pod one (See https://github.com/kubernetes/enhancements/pull/6106#discussion_r3348849539).

Shall we make the migration and support deallocation a part of this KEP?

nojnhuh: Yes, I'll update this.

To make sure I'm understanding, the pod-by-pod PostFilter will still run for pods outside PodGroups, but only the PodGroup PostFilter will run for pods in a group? And the DRA plugin's PodGroup PostFilter will do mostly the same things as its per-pod PostFilter, but will go further than only unreserving the claim for the PodGroup and deallocate the ResourceClaim?

nojnhuh: Pushed an update. I moved everything about `PostFilter` into the "Deallocate" section and added implementing the PodGroup-level `PostFilter` with deallocation for PodGroup claims.

Argh4k: > To make sure I'm understanding, the pod-by-pod PostFilter will still run for pods outside PodGroups, but only the PodGroup PostFilter will run for pods in a group?

Yes

> And the DRA plugin's PodGroup PostFilter will do mostly the same things as its per-pod PostFilter, but will go further than only unreserving the claim for the PodGroup and deallocate the ResourceClaim

I thought that for the initial implementation of DRA PodGroupPostFilter, that we need to do when introducing PodGroupPostFilter extension point, we would do basically the same thing as we do in current post filters. We would gather the unavailableClaims for pods from PodGroup in podGroupCycleState and in PodGroupPostFilter we would randomly deallocate one of them. 

My understanding is that as part of this KEP you would also like to extend this logic and deallocate pod group level claims, right?

dom4ha: > I thought that for the initial implementation of DRA PodGroupPostFilter, that we need to do when introducing PodGroupPostFilter extension point, we would do basically the same thing as we do in current post filters. We would gather the unavailableClaims for pods from PodGroup in podGroupCycleState and in PodGroupPostFilter we would randomly deallocate one of them.

We probably need to do it as a part of KEP-5710: Workload-Aware Preemption KEP as the changes needs to be under GenericWorkload feature gate (as it's a beta blocker for Gang Scheduling and WAP), but we can accept your contribution @nojnhuh if you are willing to do so.

> My understanding is that as part of this KEP you would also like to extend this logic and deallocate pod group level claims, right?

Yes, deallocation would be in turn part of this KEP beta promotion, as we should have all data on the PodGroup level at the time the PostFilter is called.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
wojtek-t: +1 - there are so many moving bits now that I wouldn't attach it now.
Let's get back to this in 1.38 timeframe an reconsider this decision.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
wojtek-t: Let's not bother about automated test for now - manual one with described scenario is probably good enough for now.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -581,55 +684,55 @@ spec:
...
+apiVersion: scheduling.k8s.io/v1beta1
 kind: PodGroup
 metadata:
   name: my-podgroup-1
   namespace: default
 spec:
   podGroupTemplateRef:
-    workloadName: my-workload
-    podGroupTemplateName: group-1
+    workload:
```
wojtek-t: nit: this will change (as part of migration to beta):
https://github.com/tosi3k/enhancements/blob/4ebe31656a230b92dc95077390a3e3a5d26b65c9/keps/sig-scheduling/6012-composite-podgroup-api/README.md#changes-to-the-podgroup-api

But that's a detail - this is not blocking this KEP for sure :)

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -742,27 +842,42 @@ deleting the ResourceClaim once its owning PodGroup is deleted.
...
+expected to run, the creator of the PodGroup is responsible for deleting it to
+free up the devices allocated by its ResourceClaims.
+
+Claims reserved for a PodGroup can also be deallocated by the scheduler in the
+DynamicResource plugin's `PostFilter` phase. In `PostFilter`, the
+DynamicResources plugin uses the scheduler's internal view of PodGroups to
+determine if any of the group's Pods are scheduled. When no Pods in the group
+are scheduled, the scheduler removes the PodGroup from the claim's
+`status.reservedFor`. A future invocation of `PostFilter` will deallocate the
+ResourceClaim if its `status.reservedFor` list is still empty.
```
wojtek-t: Why do we need PostFilter to be called twice:
- once for cleaning reservedFor
- second time for deallocation
?

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -1135,7 +1264,7 @@ well as the [existing list] of feature gates.
 -->
 
 - [X] Feature gate (also fill in values in `kep.yaml`)
-  - Feature gate name: WorkloadPodGroupResourceClaimTemplate
+  - Feature gate name: DRAWorkloadResourceClaims
```
wojtek-t, line 1265: Since I can't comment on the unchanged line:

1.
> Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?

> If the kubelet restarts with the feature disabled, existing containers continue to run with all of their allocated devices, including those from claims made by their PodGroup when the feature was enabled.

Can you clarify how that works? Isn't kubelet performing a regular admission of pods after restart, and then (given it no longer understands that the pod is referencing the claim), rejects it?

2.
> Are there any tests for feature enablement/disablement?

Were those added? Can you please add links?

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
wojtek-t: Let's focus on manual for now, but let's describe the scenario you're planning to test in more details, e.g.:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5710-workload-aware-preemption#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested

wojtek-t (commented):
Just some minor comments from PRR perspective - other than that LGTM.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
johnbelamaric: Does pushing this change to 1.38 delay the beta? ie, should this integration be a beta criteria?

I obviously would like to take this to beta in 1.37 but this is really up to SIG scheduling approvers and I defer to you.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: Thanks, updated.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: PostFilter stops and marks a Pod as unschedulable as soon as it unreserves any of its PodGroup resource claims. IIRC we don't deallocate there since unreserving the claim may be enough to schedule that Pod or another Pod. I'm not sure that's right though, so maybe we can unreserve and deallocate at the same time.

cc @dom4ha In case I'm forgetting pieces of this.

PR thread for context: https://github.com/kubernetes/kubernetes/pull/136989#discussion_r2947932486

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: > > Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
> 
> > If the kubelet restarts with the feature disabled, existing containers continue to run with all of their allocated devices, including those from claims made by their PodGroup when the feature was enabled.
> 
> Can you clarify how that works? Isn't kubelet performing a regular admission of pods after restart, and then (given it no longer understands that the pod is referencing the claim), rejects it?

I can't find any indication that DRA is involved in (re-)admission. And as long as the Pod sandbox stays intact, the kubelet doesn't check the Pod's ResourceClaims again.

> > Are there any tests for feature enablement/disablement?
> 
> Were those added? Can you please add links?

I have an open PR to add this test in https://github.com/kubernetes/kubernetes/pull/138363: https://github.com/kubernetes/kubernetes/pull/138363/changes#diff-d7f29364819f5cc467bdc62b18a49db19c3f02fdffb8f3d202e7978a2ce03c0d

I'll follow up to add a link here once it's merged.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: Sounds good, I'll add more detail here.

nojnhuh: Does the v1alpha2 -> v1alpha3 transition prevent even testing this manually though if a PodGroup can't survive transitions between 1.36 and 1.37?

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
wojtek-t: I'm not sure it has to be. For me this is an implementation detail.
Such change should be gated on its own (so we shouldn't make it as part of GA graduation), but I can imagine this happening as a second beta or even a follow-up.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
wojtek-t: > IIRC we don't deallocate there since unreserving the claim may be enough to schedule that Pod or another Pod.

I'm not sure I see a real-life usecase for it, but at least in theory it's possible I guess.

wojtek-t:
@nojnhuh - the robot is complaining about commit messages, please fix

nojnhuh:
> @nojnhuh - the robot is complaining about commit messages, please fix

Done

/hold cancel

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
dom4ha: There's no guarantee we won't encounter something unexpected the next release, but I can't come up with any obvious problem at the moment, so I'd say that beta promotion is fine.

My current understanding is that PodGroup scheduling is tightly coupled with its pods scheduling. Even a PodGroup with `Basic` scheduling policy won't be scheduled until the first pod is scheduled. Switching to PodGroup extension points may even not give us any reliability gain, as the PodGroup may be rescheduled when new pods appear, so we may need to face exactly the same edge cases.

So from the user POV, there should be no visible changes.

wojtek-t:
I'm fine with that from PRR perspective, but it still requires SIG approval.

/approve PRR

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
dom4ha: I refreshed my memory and what we did last time, we skipped removing a PodGroup from `reservedFor` on scheduling failure (in `Unreserve`):

https://github.com/kubernetes/kubernetes/blob/d88c48520ecd5a3bd9d0c9e90a48bcfca6f2ae11/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L1282-L1283

> // Ignore claims reserved for the PodGroup because other Pods in the
		// group may still need this claim. The PodGroup may be unreserved and
		// dellocated (in api-server) in the next scheduling cycle if the
		// PodGroup is not schedulable (in PostFilter) and not used yet.

We may need to fix that for beta and let scheduler implement the proper Unreserve for PodGroups, so that the next scheduling cycle can allocate different resources.

> Why do we need PostFilter to be called twice:

The first invocation should be about `Unreserve`, not `PostFilter`.  The updated `Unreserve` should handle most of the scenarios of the PodGroup scheduling failure. The tricky part is that `Unreserve` can be called synchronously as a part of the scheduling cycle (after `Reserve`) and asynchronously if binding fails.

The implementation in PostFilter only needs to handle the corner cases where scheduler failed to unregister (for instance after crash). The implementation should be already there, we only need to switch to PodGroup level PostFilter extension point, but it will be handled by Workload Aware Preemption (under the GenericWorkload gate).

dom4ha:
@nojnhuh please fix the wording about `Unreserve` vs `PostFilter`, but the rest looks good.

/approve

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: Updated this to clarify that `Unreserve` will remove the PodGroup from a claim's `status.reservedFor` and that `PostFilter` will deallocate the claim.

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -742,27 +840,42 @@ deleting the ResourceClaim once its owning PodGroup is deleted.
...
+deletion is gated by a finalizer managed by kube-controller-manager
 to prevent the PodGroup from being removed from `status.reservedFor` before all
 of its Pods are done using the ResourceClaim. When no more Pods in the group are
-expected to run, the creator of the PodGroup is responsible for removing the
-finalizer and deleting the PodGroup.
+expected to run, the creator of the PodGroup is responsible for deleting it to
+free up the devices allocated by its ResourceClaims.
+
+Claims reserved for a PodGroup can also be deallocated by the scheduler in the
+DynamicResource plugin's `PostFilter` phase when they are not reserved for any
```
dom4ha:
```suggestion
DynamicResource plugin's `Unreserve` phase when scheduling of a PodGroup failed and they are not reserved for any other resources (like other PodGroups).
```

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
```diff
@@ -742,27 +840,42 @@ deleting the ResourceClaim once its owning PodGroup is deleted.
...
+other resources. In the `Unreserve` phase, the
```
dom4ha:
```suggestion
In that  phase, the
```

dom4ha (commented):
Suggested to further change the wording, WDYT?

keps/sig-scheduling/5729-resourceclaim-support-for-workloads/README.md
nojnhuh: Sounds good, thanks! Updated.

dom4ha:
Thanks @nojnhuh !
/lgtm

PR #5989, opened 2026-03-31
KEP-5989: Preserve UDP flows for terminating endpoints

kairosci:
- One-line PR description: Adding new KEP for preservation of UDP flows for terminating endpoints

- Issue link: [kubernetes/kubernetes#134143](https://github.com/kubernetes/kubernetes/issues/134143)

- Other comments: This KEP adds a new field `Processing` to `EndpointConditions` in the EndpointSlice API and updates Kube-proxy logic to respect this field during UDP conntrack cleanup.

### DIFF
--- a/keps/sig-network/5989-preserve-udp-flows-terminating-endpoints/README.md
+++ b/keps/sig-network/5989-preserve-udp-flows-terminating-endpoints/README.md
@@ -1,0 +1,247 @@
+# KEP-5989: Preserve UDP flows for stateful UDP protocols during endpoint termination
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Implementation Approach](#implementation-approach)
+  - [Conntrack Cleanup Strategy](#conntrack-cleanup-strategy)
+  - [Kube-proxy Changes](#kube-proxy-changes)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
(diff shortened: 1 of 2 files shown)


aojea:
how is this different than terminating?
what is the state machine of the lifecycle of the endpoint with this new state?
adding a new state just for a particular protocol in an API does not look right to me, also this is a particular behavior of kube-proxy, that makes me be more defensive about justifying an API change 

IIRC the original issue is that there are same UDP protocols that are stateful, and the user wanted to avoid conntrack on those stateful udp protocl ... I will try to focus on solving the problem by having a way to indicate to kube-proxy what kind of udp traffic is stateful and should not be cleared during the terminating state

kairosci:
Hm, I hadn't considered it that way. I'm moving away from the API modification. Instead, I'll implement a way for kube-proxy to distinguish stateful UDP traffic so it can bypass conntrack flushing during the standard termination phase.

keps/sig-network/5989-preserve-udp-flows-terminating-endpoints/README.md
```diff
@@ -0,0 +1,202 @@
...
+
+A new annotation on Services will be introduced:
+
+```yaml
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+  annotations:
+    service.kubernetes.io/stateful-udp-ports: "3478,8080,9000"
```
aojea: annotations are always a poor api, I will try to piggyback on existing netfilter operations and extract lessons learned from them, they already created these extensions helpers https://github.com/regit/secure-conntrack-helpers/blob/master/secure-conntrack-helpers.rst

keps/sig-network/5989-preserve-udp-flows-terminating-endpoints/README.md
```diff
@@ -0,0 +1,202 @@
...
+```
+
+The annotation `service.kubernetes.io/stateful-udp-ports` contains a comma-separated list of port numbers (or port names) that are running stateful UDP protocols. Kube-proxy will read this annotation and, when an endpoint for this Service enters the terminating state, it will skip clearing conntrack entries for the specified UDP ports.
```
aojea: at one point they need to force clean it up, right? otherwise it will blackhole later entries, I think here we just need to define when we do the cleanup

kairosci: Right, force cleanup is needed.

When: after a grace period (default 30 seconds) following the transition to the terminating state.
How: kube-proxy tracks the termination timestamp and periodically cleans up the marked conntrack entries, or relies on conntrack's natural timeout.

I've included them in the README.

kairosci:
@aojea Could you take another look at it? That way, if it looks promising, I can start implementing it.

keps/sig-network/5989-preserve-udp-flows-terminating-endpoints/README.md
```diff
@@ -0,0 +1,236 @@
...
+### Risks and Mitigations
+
+- **Risk**: Stale conntrack entries might persist if the deferred cleanup fails.
+- **Mitigation**: The grace period ensures cleanup will happen eventually. Additionally, conntrack entries will still time out based on conntrack timeout settings.
+
+- **Risk**: Complexity in tracking terminating endpoints and their grace periods.
+- **Mitigation**: Reuse existing endpoint tracking infrastructure in kube-proxy's EndpointSliceCache.
+
+- **Risk**: Users may not properly mark their stateful UDP flows.
+- **Mitigation**: Documentation will explain how to use conntrack marks for stateful UDP protocols. Consider providing examples or helper scripts.
```
danwinship: I'm not sure there's really any particularly good way to do this.

The program itself isn't going to mark its packets, because it's (presumably) some general-purpose program that doesn't know Kubernetes-specific details.

That means doing it via something outside the program, but that will inherently be racy (sometimes kube-proxy will end up killing a connection before the helper gets around to marking it), and just seems inefficient as compared to having kube-proxy be able to figure it out by itself somehow.

However, I think there is an important insight in this KEP: in the original issue, I was thinking in terms of pod probes because that's how the reporter first described it. But this isn't really a property of the pods, it's a property of the service: for some kinds of services, we want the traditional "delete UDP conntrack records right away" semantics, and for others, we want "only delete UDP conntrack records after the pod exits" semantics.

So this sounds more like [the "connection type hint" idea](https://github.com/kubernetes/kubernetes/issues/127259#issuecomment-2436201258) again...

kairosci: I have removed the use of conntrack marks from the project to introduce a more straightforward solution. The new proposal adds the `udpStreamStrategy` field to `ServicePort`, allowing users to explicitly define how to handle UDP traffic when an endpoint is terminated.

Available options:
Default: Maintains the current behavior, immediately clearing the conntrack table as soon as the endpoint is marked as terminating.

OnPodExit: Preserves existing UDP streams until the Pod is actually removed from the EndpointSlice.

Goal of the change:
This change resolves the issue at its root by providing kube-proxy with the necessary context to distinguish between stateless and stateful UDP traffic. As a result, it is no longer necessary to rely on external helpers or indirect traffic marking.

PR #5992, opened 2026-04-02
KEP-5726: TopologyManager CPU-attached NUMA filter option

fanzhangio:
#### What type of PR is this?

/kind documentation
/kind cleanup
/kind kep

#### What this PR does / why we need it:

Adds a draft KEP for an alpha TopologyManager policy option:

- `restrict-to-cpu-numa-nodes`

This option is intended to help kubelet handle large NUMA systems that expose many CPU-less NUMA nodes, such as NVIDIA GraceBlackwell / VeraRubin-class platforms.

When enabled, TopologyManager computes an effective NUMA-node set containing only NUMA nodes with CPUs attached and propagates that same view to topology-aware hint providers.

The proposal is intended to address the root issue discussed in:
- kubernetes/kubernetes#135541

This proposal also reworks and supersedes the earlier code-only approach discussed in:
- kubernetes/kubernetes#135581

#### Which issue(s) this PR fixes:

Part of kubernetes/kubernetes#135541

#### Special notes for your reviewer:

Main points of the proposal:
- behavior is unchanged by default
- the option is alpha and opt-in
- the effective NUMA-node set is shared across TopologyManager, CPUManager, MemoryManager, and DeviceManager
- DeviceManager projects raw device NUMA topology onto the same effective CPU-attached NUMA-node set using NUMA distance information

There is also related broader enhancement discussion in:
- kubernetes/enhancements#5726

This KEP is intentionally scoped more narrowly around a kubelet TopologyManager option.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

### DIFF
--- a/keps/sig-node/5726-topologymanager-cpu-numa-filter/README.md
+++ b/keps/sig-node/5726-topologymanager-cpu-numa-filter/README.md
@@ -1,0 +1,172 @@
+# KEP-5726: TopologyManager CPU-attached NUMA filter option
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [User Stories](#user-stories)
+- [Design Details](#design-details)
+  - [API](#api)
+  - [TopologyManager Behavior](#topologymanager-behavior)
+  - [CPUManager and MemoryManager Behavior](#cpumanager-and-memorymanager-behavior)
+  - [DeviceManager Behavior](#devicemanager-behavior)
+  - [Failure Modes](#failure-modes)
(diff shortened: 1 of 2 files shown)


fanzhangio:
@dims @klueska

fanzhangio:
@ffromani Could we add this KEP to the SIG-Node agenda for discussion? I’d be happy to present it in the meeting (or follow up asynchronously over email if that’s easier) Thanks.

ffromani:
> @ffromani Could we add this KEP to the SIG-Node agenda for discussion? I’d be happy to present it in the meeting (or follow up asynchronously over email if that’s easier) Thanks.

Yes, it is 1.37 material, but the 1.37 cycle did not begun yet (we didn't finish 1.36). That said, you can present earlier to sig-node if you wish to.

Issue #6003, opened 2026-04-08
Configurable sync period for HPA


Fedosin:
### Enhancement Description

- One-line enhancement description (can be used as a release note):
  Add a new optional `syncPeriodSeconds` field to `HorizontalPodAutoscalerSpec`, allowing a per-HPA override of the global sync period (`--horizontal-pod-autoscaler-sync-period`). When unset, the global default is used.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/tree/master/keps/sig-autoscaling/6003-configurable-hpa-sync-period (KEP PR: https://github.com/kubernetes/enhancements/pull/6004)
- Discussion Link: https://github.com/kubernetes/kubernetes/issues/110317 (original request, pre-KEP)
- Primary contact (assignee): @Fedosin
- Responsible SIGs: sig-autoscaling
- Enhancement target (which target equals to which milestone):
  - Alpha release target (x.y): v1.38
  - Beta release target (x.y):
  - Stable release target (x.y):
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6004
    - [ ] Code (`k/k`) update PR(s): prototype https://github.com/kubernetes/kubernetes/pull/138222 (closed; to be reopened against the design in the KEP once it merges). Merged prerequisites: https://github.com/kubernetes/kubernetes/pull/138228, https://github.com/kubernetes/kubernetes/pull/138294
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

Fedosin:
/remove-lifecycle stale
/assign

The KEP is up for review in #6004 (PRR file in place, approver @soltysh), targeting alpha in v1.38. Updated the description with the KEP link and milestone.

Fedosin:
/stage alpha

@adrianmoisey @omerap12 -- this KEP is targeting alpha in v1.38, and the PRR approver (@soltysh) is ready to review once it shows up on the release tracking board. Could one of you opt it in with `/label lead-opted-in` and `/milestone v1.38`?

`kep.yaml` already carries `stage: alpha` and `latest-milestone: v1.38`, and the PRR file names soltysh as approver. Thanks!

adrianmoisey:
/label lead-opted-in
/milestone v1.38

[...]

PR #6004, opened 2026-04-08
KEP-6003: Add KEP for configurable HPA sync period

Fedosin:
- One-line PR description: Introduce KEP proposing an optional syncPeriodSeconds field on HorizontalPodAutoscalerSpec to allow per-HPA override of the global --horizontal-pod-autoscaler-sync-period.

- Issue link: https://github.com/kubernetes/enhancements/issues/6003

- Other comments:
  - The original request predates the KEP process: https://github.com/kubernetes/kubernetes/issues/110317
  - Prototype implementation, now closed, to be reopened against the design this KEP settles on: https://github.com/kubernetes/kubernetes/pull/138222
  - Merged prerequisites: https://github.com/kubernetes/kubernetes/pull/138228 (HPA generation tracking) and https://github.com/kubernetes/kubernetes/pull/138294 (immediate enqueue on create and spec change)
  - Targeting alpha in v1.38. The field sits at the top level of `spec` rather than under `spec.behavior`, and is added to both served API versions (`autoscaling/v1` and `autoscaling/v2`).

### DIFF
--- a/keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
+++ b/keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
@@ -1,0 +1,565 @@
+# KEP-6003: Configurable sync period for HPA
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Upgrade](#upgrade)
+    - [Downgrade](#downgrade)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+[Horizontal Pod Autoscaler][] (HPA) periodically reconciles the desired replica
+count for a given Deployment (or other resource with a `/scale` subresource)
+based on observed metrics. The frequency at which this reconciliation happens
+is governed by a single global flag,
+`--horizontal-pod-autoscaler-sync-period`, which defaults to 15 seconds
+and applies to every HPA in the cluster.
+
+This proposal adds an optional `syncPeriodSeconds` field to
+`HorizontalPodAutoscalerBehavior` that allows users to override the global sync
+period on a per-HPA basis. When the field is unset, the global default
+continues to apply.
+
+[Horizontal Pod Autoscaler]: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/
+
+## Motivation
+
+Today the HPA sync period is a cluster-wide parameter set using the
+[Kube Controller Manager][]
+`--horizontal-pod-autoscaler-sync-period` flag (default 15s). While this
+value is reasonable for many workloads, it is a one-size-fits-all setting
+that forces every HPA in the cluster to reconcile at the same frequency.
+
+This creates a tension between two competing goals:
+
+1. **Latency-sensitive workloads** need faster scaling decisions (e.g. every
+   1-5 seconds) to respond to rapid traffic spikes.
+2. **Stable workloads** are fine with the default or even longer periods, and
+   increasing the global frequency for all HPAs unnecessarily increases load
+   on the API server and metrics backends.
+
+Since appropriate sync periods are workload-dependent, this KEP proposes to
+let users set a custom sync period per `HorizontalPodAutoscaler` resource,
+overriding the global default when present.
+
+[Kube Controller Manager]: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/
+
+### Goals
+
+- Allow users to optionally override the default HPA reconciliation frequency
+  on a per-HPA basis.
+- Maintain full backward compatibility: existing HPAs without the new field
+  continue to use the global sync period.
+
+### Non-Goals
+
+- Change the default value of the global
+  `--horizontal-pod-autoscaler-sync-period` flag.
+- Allow sub-second sync periods.
+
+## Proposal
+
+We propose to add a new field to the existing [`HorizontalPodAutoscalerBehavior`][] object:
+
+- `syncPeriodSeconds`: *(int32)* the period in seconds between each
+  reconciliation of this HPA. Must be greater than 0 and less than or equal
+  to 3600 (one hour).
+
+The `syncPeriodSeconds` field is optional, and when not specified the HPA will
+continue to use the value of the global
+`--horizontal-pod-autoscaler-sync-period` flag.
+
+**Interaction with metrics sources**: The effectiveness of reducing the sync
+period depends on the freshness of the underlying metrics. For resource metrics
+served by [metrics-server][], values are typically collected at a fixed interval
+(default 60s) and cached, so syncing the HPA faster than the metrics collection
+interval will not yield fresher data -- the HPA will simply re-evaluate the same
+cached values. Reducing the sync period is most impactful when used with
+custom or external metrics providers that expose rapidly updating values. Users
+should consider their metrics pipeline latency when choosing a
+`syncPeriodSeconds` value.
+
+This field is placed under `Behavior` rather than at the top-level spec because
+it is an operational tuning parameter that controls *how* the HPA operates,
+consistent with the existing `scaleUp`/`scaleDown` rules and the `tolerance`
+field introduced in [KEP-4951][].
+
+[ValidatingAdmissionPolicy]: https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/
+[`HorizontalPodAutoscalerBehavior`]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#horizontalpodautoscalerbehavior-v2-autoscaling
+[KEP-4951]: /keps/sig-autoscaling/4951-configurable-hpa-tolerance
+[metrics-server]: https://github.com/kubernetes-sigs/metrics-server
+
+### User Stories
+
+#### Story 1
+
+As an operator of a latency-sensitive web application, I want my HPA to
+reconcile every 2 seconds so that scale-up decisions are made quickly during
+sudden traffic spikes, without forcing every other HPA in the cluster to also
+reconcile that frequently.
+
+#### Story 2
+
+As a platform engineer managing a multi-tenant cluster, I want different teams
+to be able to tune HPA responsiveness for their own workloads independently,
+without requiring cluster-admin involvement to change the global flag.
+
+### Risks and Mitigations
+
+There should be minimal risk introduced by the proposed changes:
+
+- **Increased API server and metrics load**: Users could set very low sync
+  periods for many HPAs, increasing the rate of metrics queries and scale
+  sub-resource calls. This is mitigated by multiple layers:
+  - **Validation bounds**: `syncPeriodSeconds` must be >= 1 and <= 3600. We
+    may raise the lower bound (e.g. to 5s) before Beta if real-world usage
+    data suggests that very short periods cause issues in practice.
+  - **Feature gate**: In Alpha, the feature is gated behind
+    `HPAConfigurableSyncPeriod`, giving cluster administrators explicit
+    control over whether the field is accepted at all.
+  - **Best-effort semantics**: The field specifies a *target* interval, not a
+    hard guarantee. If a reconciliation cycle takes longer than the configured
+    period, the controller will not queue additional work -- it will simply
+    start the next cycle once the current one completes. This prevents
+    workqueue saturation (see [Design Details](#design-details)).
+  - **Policy enforcement**: Cluster administrators can use [ValidatingAdmissionPolicy][]
+    (or a webhook) to set a cluster-specific floor. For example, a simple CEL
+    rule can reject HPAs with `syncPeriodSeconds` below a chosen threshold
+    without needing to reason about aggregate frequency:
+    ```
+    rule: "!has(object.spec.behavior) ||
+           !has(object.spec.behavior.syncPeriodSeconds) ||
+           object.spec.behavior.syncPeriodSeconds >= 10"
+    ```
+- The new field is optional, and its absence results in no changes to the
+  current autoscaling behavior.
+- If a change to the new field results in undesirable behavior, the change
+  can be reverted by removing the `syncPeriodSeconds` field from the HPA,
+  restoring the global default.
+
+## Design Details
+
+The `HorizontalPodAutoscaler` API is updated to add a new `syncPeriodSeconds`
+field to the `HorizontalPodAutoscalerBehavior` object:
+
+```golang
+type HorizontalPodAutoscalerBehavior struct {
+  // syncPeriodSeconds is the period in seconds between each reconciliation
+  // of this HPA. When unset, the global
+  // --horizontal-pod-autoscaler-sync-period value is used (default: 15s).
+  // Must be greater than 0 and less than or equal to 3600.
+  // +optional
+  SyncPeriodSeconds *int32
+
+  // Existing fields.
+  ScaleUp *HPAScalingRules
+  ScaleDown *HPAScalingRules
+}
+```
+
+The `syncPeriodSeconds` field specifies a best-effort target interval: the
+controller will attempt to reconcile the HPA at least this often, but does not
+guarantee a hard real-time deadline. If a reconciliation cycle (including
+metrics queries and scale sub-resource calls) takes longer than the configured
+period, the controller will start the next cycle immediately after the current
+one completes rather than queuing up additional work. This prevents workqueue
+saturation even when the metrics backend is slow or the configured period is
+shorter than the end-to-end reconciliation latency.
+
+The per-HPA sync frequency is implemented via a new `PerItemIntervalRateLimiter`
+in the HPA controller's workqueue. This rate limiter supports per-key interval
+overrides with a fallback to the global default. The controller updates the
+per-item interval during reconciliation and cleans it up on HPA deletion.
+
+The informer event handlers are updated so that:
+- Newly created HPAs and spec changes (detected via `Generation` comparison)
+  are enqueued immediately via `queue.Add`, so they are processed without
+  waiting for the rate limiter delay. This relies on the HPA `Generation`
+  field being properly incremented on spec changes, which was fixed in
+  [kubernetes#138228][].
+- Status-only or metadata-only updates continue to use `AddRateLimited` to
+  preserve the hot-loop prevention introduced in [#42715][].
+- The periodic resync cadence is maintained by the existing `AddRateLimited`
+  call in `processNextWorkItem`, which re-enqueues each HPA with its
+  configured (or default) sync period delay after every reconciliation.
+
+The informer's `AddEventHandlerWithResyncPeriod` continues to use the global
+`resyncPeriod` as a background safety net; the per-HPA sync frequency is
+driven entirely by the workqueue rate limiter.
+
+Since the added field is optional and its omission does not change the existing
+autoscaling behavior, this feature will only be added to the latest stable API
+version `pkg/apis/autoscaling/v2`. Older versions (i.e. `v1`, `v2beta1`,
+`v2beta2`) will not include the new field, but converters will be updated where
+needed to comply with [round-trip requirements][].
+
+The validation logic will be updated to ensure that the `syncPeriodSeconds`
+field is greater than 0 and does not exceed 3600.
+
+**Interaction with stabilization windows**: A shorter `syncPeriodSeconds`
+increases the number of scaling recommendations produced within a given
+`stabilizationWindowSeconds` window. For example, with a 5s sync period and a
+300s stabilization window the controller accumulates ~60 recommendations,
+compared to ~20 with the default 15s period. Because the stabilization logic
+selects the highest (scale-up) or lowest (scale-down) recommendation within
+the window, more samples do not change the eventual scaling decision -- they
+only make the window more granular. The semantics of `stabilizationWindowSeconds`
+remain unchanged.
+
+[#42715]: https://github.com/kubernetes/kubernetes/pull/42715
+[kubernetes#138228]: https://github.com/kubernetes/kubernetes/pull/138228
+[kubernetes#138294]: https://github.com/kubernetes/kubernetes/pull/138294
+[round-trip requirements]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+##### Unit tests
+
+- `pkg/apis/autoscaling/validation`: `<date>` - `<test coverage>`
+- `pkg/controller/podautoscaler`: `<date>` - `<test coverage>`
+
+##### Integration tests
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+-->
+
+An integration test will be added to verify the autoscaling behavior when
+custom sync periods are set on an HPA and that HPAs without the field
+continue to use the global default.
+
+##### e2e tests
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+-->
+
+Existing e2e tests ensure the autoscaling behavior uses the default sync period
+when no configurable sync period is specified.
+
+New e2e tests will be added to verify:
+- An HPA with a short `syncPeriodSeconds` reconciles more frequently than the
+  global default.
+- An HPA without `syncPeriodSeconds` continues to reconcile at the global
+  default rate.
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature implemented behind a `HPAConfigurableSyncPeriod` feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- All tests described in the [`e2e tests` section](#e2e-tests) are implemented
+  and linked in this KEP.
+- We have monitored for negative user feedback and addressed relevant concerns.
+- Integration test verifying behavior with and without the field.
+- Per-HPA observability metric exposing the effective sync period is available.
+
+#### GA
+
+- The feature has been Beta for at least 2 releases with no major issues
+  reported.
+- Conformance tests are in place.
+- Documentation is published on kubernetes.io.
+
+### Upgrade / Downgrade Strategy
+
+#### Upgrade
+Existing HPAs will continue to work as they do today, using the global
+`--horizontal-pod-autoscaler-sync-period` value from the
+`kube-controller-manager`. Users can use the new feature by enabling the
+Feature Gate (alpha only) and setting the new `syncPeriodSeconds` field in
+an HPA's `spec.behavior`.
+
+#### Downgrade
+On downgrade, all HPAs will revert to using the global
+`--horizontal-pod-autoscaler-sync-period` value from the
+`kube-controller-manager`, regardless of any configured `syncPeriodSeconds`
+value on the HPA itself.
+
+### Version Skew Strategy
+
+1. `kube-apiserver`: More recent instances will accept the new
+   `syncPeriodSeconds` field, while older instances will ignore it.
+2. `kube-controller-manager`: An older version could receive an HPA containing
+   the new `syncPeriodSeconds` field from a more recent API server, in which
+   case it would ignore it (i.e. reconcile at the global default rate).
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: HPAConfigurableSyncPeriod
+  - Components depending on the feature gate: `kube-controller-manager` and
+    `kube-apiserver`.
+
+###### Does enabling the feature change any default behavior?
+
+No.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+The feature can be disabled by restarting the `kube-controller-manager` with
+the feature gate set to `false`.
+
+Any `syncPeriodSeconds` values set on existing HPAs will be ignored by the
+`kube-controller-manager` and `kube-apiserver` when the feature gate is off.
+All HPAs will revert to using the global sync period.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+When the feature is re-enabled, any HPAs with configured `syncPeriodSeconds`
+values will use those as their reconciliation interval, rather than the global
+sync period from the `kube-controller-manager`.
+
+###### Are there any tests for feature enablement/disablement?
+
+Unit tests will be added to verify that HPAs with and without the new field
+are properly validated, both when the feature gate is enabled or not.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+This feature does not introduce new failure modes: during rollout/rollback,
+some API servers will allow or disallow setting the new `syncPeriodSeconds`
+field. The new field is possibly ignored until the controller manager is fully
+updated.
+
+###### What specific metrics should inform a rollback?
+
+A high `horizontal_pod_autoscaler_controller_metric_computation_duration_seconds`
+metric can indicate a problem related to this feature. Additionally, unexpected
+increases in API server request rates from the HPA controller may indicate that
+sync periods are set too aggressively.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+<!--
+Describe manual testing that was done and the outcomes.
+Longer term, we may want to require automated upgrade/rollback tests, but we
+are missing a bunch of machinery and tooling and can't do that now.
+-->
+
+TBD
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+The presence of the new `syncPeriodSeconds` field in the HPA's
+`spec.behavior` indicates that the feature is in use.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [X] Events
+  - Event Reason: `SuccessfulRescale`
+- [X] Other
+  - Controller logs at verbosity >= 4 include the reconciliation interval
+    for each HPA, allowing operators to verify the configured period is
+    being respected.
+
+Users can observe the frequency of `SuccessfulRescale` events when scaling is
+active. Note that these events only fire when a scaling decision is made, not
+on every reconciliation cycle. For periods where the HPA is stable (no scaling
+needed), operators can increase the kube-controller-manager log verbosity to
+confirm the reconciliation cadence. For Beta, we plan to introduce a per-HPA
+metric (e.g. `horizontal_pod_autoscaler_reconciliation_interval_seconds`) that
+exposes the observed interval between reconciliations, providing a direct signal
+regardless of whether scaling occurs.
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+Although the absolute value of the
+`horizontal_pod_autoscaler_controller_metric_computation_duration_seconds`
+metric depends on HPAs configuration, it should be unimpacted by this feature.
+This metric should not vary by more than 5%.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+This KEP is not expected to have any impact on SLIs/SLOs as it doesn't
+introduce a new HPA behavior, but merely allows users to easily change the
+reconciliation frequency that is otherwise a global parameter.
+
+The standard HPA metric
+`horizontal_pod_autoscaler_controller_metric_computation_duration_seconds` can
+be used to verify the HPA controller health.
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+Yes. A per-HPA metric `horizontal_pod_autoscaler_reconciliation_interval_seconds`
+exposing the actual observed interval between reconciliations would allow
+operators to verify that the configured `syncPeriodSeconds` is achievable given
+their metrics pipeline latency. This metric is planned for Beta graduation.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+No, this feature does not depend on any specific service.
+
+### Scalability
+
+###### Will enabling / using this feature result in any new API calls?
+
+No new API call types are introduced. However, setting a lower
+`syncPeriodSeconds` on an HPA will increase the frequency of existing API
+calls (metrics queries, scale sub-resource reads/updates) for that specific
+HPA proportionally.
+
+###### Will enabling / using this feature result in introducing new API types?
+
+No.
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+- This feature adds one new optional int32 field to
+  `HorizontalPodAutoscaler` `v2` objects. Users should expect this object to
+  increase in size by approximately 4 bytes when the field is set.
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+No.
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any components?
+
+Setting lower `syncPeriodSeconds` values will proportionally increase
+CPU usage and API call volume on the `kube-controller-manager` and metrics
+backend for the affected HPAs. The validation bound (minimum 1s, maximum 3600s)
+limits the impact, but operators should monitor resource usage when configuring
+aggressive sync periods on many HPAs.
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No. The feature only affects the HPA controller's reconciliation frequency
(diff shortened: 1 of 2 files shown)


keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,492 @@
...
+}
+```
+
+The per-HPA sync frequency is implemented via a new `PerItemIntervalRateLimiter`
+in the HPA controller's workqueue. This rate limiter supports per-key interval
+overrides with a fallback to the global default. The controller updates the
+per-item interval during reconciliation and cleans it up on HPA deletion.
+
+The informer event handlers are updated so that:
+- Newly created HPAs and spec changes (detected via `Generation` comparison)
```
adrianmoisey: HPA's generation fields are currently not set.
I have https://github.com/kubernetes/kubernetes/pull/138228 to get this fixed

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,492 @@
...
+- Change the default value of the global
+  `--horizontal-pod-autoscaler-sync-period` flag.
+- Allow sub-second sync periods.
+
+## Proposal
+
+We propose to add a new field to the existing [`HorizontalPodAutoscalerBehavior`][] object:
+
+- `syncPeriodSeconds`: *(int32)* the period in seconds between each
+  reconciliation of this HPA. Must be greater than 0 and less than or equal
```
jm-franc, line 120: This will allow this HPA to query more frequently the metric source, which is usually metrics-server, which caches its metrics. The end result is that in most situations reducing this sync period won't have much impact. Conceivably you could add a field that only targets custom/external metrics, although my suspicion is that will result in a non-intuitive API.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,492 @@
...
+
+### Risks and Mitigations
+
+There should be minimal risk introduced by the proposed changes:
+
+- **Increased API server and metrics load**: Users could set very low sync
+  periods (e.g. 1s) for many HPAs, increasing the rate of metrics queries
+  and scale sub-resource calls. This is mitigated by:
+  - Validation bounds: `syncPeriodSeconds` must be >= 1 and <= 3600.
+  - Cluster administrators can use admission webhooks or policies to enforce
```
jm-franc, line 158: Wouldn't a meaningful webhook require to calculate the aggregate sync frequency across all HPAs in the cluster? For users, this looks difficult to reason about.

Do you see another way such a webhook could implemented? Perhaps it could be part of this KEP to ensure this feature is safe by default?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,492 @@
...
+  SyncPeriodSeconds *int32
+
+  // Existing fields.
+  ScaleUp *HPAScalingRules
+  ScaleDown *HPAScalingRules
+}
+```
+
+The per-HPA sync frequency is implemented via a new `PerItemIntervalRateLimiter`
+in the HPA controller's workqueue. This rate limiter supports per-key interval
```
jm-franc, line 187: Wouldn't metrics-server and custom/external metrics API need to respond in less than 1s to ensure the workqueue doesn't saturate? This looks like a very short timeout. Do you see how to prevent the workqueue from blocking? Perhaps this new field should only be a hint (i.e. a best-effort goal)?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Good catch, thank you! I've added a note in the Design Details section acknowledging the dependency on https://github.com/kubernetes/kubernetes/pull/138228 for properly incrementing the HPA `Generation` field on spec changes.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Great point. I've added a new "Interaction with metrics sources" paragraph to the Proposal section addressing this. It explains that for resource metrics served by metrics-server (which caches values at ~60s intervals), syncing the HPA faster than the metrics collection interval won't yield fresher data. Reducing the sync period is most impactful when used with custom or external metrics providers that expose rapidly updating values. Users are advised to consider their metrics pipeline latency when choosing a `syncPeriodSeconds` value.

I considered scoping the field to custom/external metrics only, but as you noted, that would result in a non-intuitive API -- users would need to reason about which metric types their HPA uses before setting the field. A single field that applies uniformly is simpler, and the documentation clarifies when it's actually beneficial.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: You're right that a webhook calculating aggregate frequency would be hard to reason about. I've reworked the Risks and Mitigations section to present a multi-layered safety story instead:

1. **Validation bounds** (>= 1, <= 3600) with a note that we may raise the lower bound before Beta based on real-world usage data.
2. **Feature gate** -- in Alpha, cluster admins have explicit opt-in control.
3. **Best-effort semantics** -- the field is a target interval, not a hard guarantee. The controller won't queue additional work if reconciliation takes longer than the configured period, preventing workqueue saturation.
4. **Policy enforcement** -- I've replaced the vague "admission webhooks" suggestion with a concrete `ValidatingAdmissionPolicy` example using a simple CEL rule that enforces a per-HPA floor (e.g. `syncPeriodSeconds >= 10`). This doesn't require reasoning about aggregate frequency -- it's a straightforward per-object check.

This makes the feature safe by default while still giving cluster admins a simple knob if they want tighter control.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Good concern. I've added a paragraph in Design Details clarifying that `syncPeriodSeconds` is a best-effort target interval, not a hard real-time guarantee. Specifically:

- If a reconciliation cycle (including metrics queries and scale sub-resource calls) takes longer than the configured period, the controller will start the next cycle immediately after the current one completes rather than queuing up additional work.
- This means the workqueue cannot saturate even if the metrics backend is slow or the configured period is shorter than the end-to-end reconciliation latency.
- The rate limiter only re-enqueues a key *after* the current reconciliation for that key finishes, so there is at most one pending item per HPA in the queue at any time.

So effectively, `syncPeriodSeconds` is already a hint/best-effort goal by design. I've also reflected this in the Risks and Mitigations section.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,533 @@
...
+If possible increase the log level for kube-controller-manager and check
+controller logs:
+1. Verify that the `syncPeriodSeconds` value is being respected by observing
+   the reconciliation frequency in logs and events.
+2. Check if the increased reconciliation frequency is causing excessive load
+   on the API server or metrics backend.
+3. Look for warnings and errors which might point where the problem lies.
+
+## Implementation History
+
```
wozniakjan, line 837: might be worth referencing the old issue tracker before KEPs existed :)
https://github.com/kubernetes/kubernetes/issues/110317

wozniakjan (commented):
<3 from KEDA

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+**Reverting a bad value.** The field is optional; removing it restores current
+behavior.
+
+[ValidatingAdmissionPolicy]: https://kubernetes.io/docs/reference/access-authn-authz/validating-admission-policy/
+
+## Design Details
+
+The `HorizontalPodAutoscaler` API is updated to add `syncPeriodSeconds` to
+`HorizontalPodAutoscalerSpec`:
```
adrianmoisey: The HPA API has 2 versions, and we decided to try ensure that new fields are configured in both APIs, rather than hosting them into annotations in v1.
Could this be indicated on the KEP somewhere?

May be it's worth showing the change in both v1 and v2? (even though they're identical) and explaining how round-tripping will work

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+Existing unit tests in `pkg/controller/podautoscaler` build the controller with a
+single global sync period and assert behavior without controlling time. They will
+be refactored to allow a per-HPA delay and a fake clock to be injected, so
+timing can be asserted deterministically. No change to existing test semantics is
+expected.
+
+##### Unit tests
+
+- `pkg/apis/autoscaling/validation`: `<date>` - `<test coverage>`
+- `pkg/controller/podautoscaler`: `<date>` - `<test coverage>`
```
adrianmoisey: Can these be filled in?
I believe they are supposed to show coverage before the KEP is created.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- Feature implemented behind a `HPAConfigurableSyncPeriod` feature flag
+- Unit and integration tests described above implemented and enabled
+- Per-HPA reconciliation scheduling-delay metric available, so operators can
+  tell whether configured delays are being met
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- All tests described in the [`e2e tests` section](#e2e-tests) are implemented
+  and linked in this KEP.
```
adrianmoisey: I'm not too sure what this means, since alpha has the e2e tests covered

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- We have monitored for negative user feedback and addressed relevant concerns.
+- Real-world usage data reviewed to decide whether the lower validation bound
+  should be relaxed below 3s, and whether cross-HPA interference warrants
+  workqueue fairness or a controller-level budget on aggregate reconcile rate.
```
adrianmoisey: Where will this real-world usage data come from?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- Documentation covering when the field is and is not effective, and the
+  interaction with `--concurrent-horizontal-pod-autoscaler-syncs`.
+
+#### GA
+
+- The feature has been Beta for at least 2 releases with no major issues
+  reported.
```
adrianmoisey: Any specific reason for this?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- Conformance tests are in place.
```
adrianmoisey: Do we need conformance? I don't believe we have any for HPA

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- Documentation is published on kubernetes.io.
```
adrianmoisey, line 535: Can documentation come along with alpha?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+  This predates the KEP process and is the original tracking issue.
+- 2026-04-06: [Implementation PR](https://github.com/kubernetes/kubernetes/pull/138222) opened.
+- 2026-04-08: Initial KEP created.
+- 2026-04-24: [kubernetes#138228][] merged -- HPA generation tracking.
+- 2026-05-07: [kubernetes#138294][] merged -- immediate enqueue on HPA creation
+  and spec changes.
+- 2026-08-19: Field moved from `spec.behavior` to top-level `spec`;
+  scheduling semantics specified as a fixed delay after reconciliation;
+  observability specified as a per-HPA delay ratio gauge. KEP marked
+  `implementable`, targeting alpha in v1.38.
```
adrianmoisey: I could be wrong, but I think this `Implementation History` doesn't need this sort of detail

adrianmoisey (commented):
Left some comments after a first pass

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+`horizontal_pod_autoscaler_controller_reconciliation_delay_ratio` metric, and
+is cleared on deletion. When `syncPeriodSeconds` changes, the pending
+eligibility deadline is recomputed; a stale delayed workqueue entry must not
+cause a periodic reconciliation before the new deadline.
+
+The informer event handlers are updated so that:
+- Creations and spec changes (detected via `Generation`) are enqueued
+  immediately with `queue.Add`, bypassing the rate limiter delay. This relies
+  on `Generation` being incremented on spec changes, fixed in
+  [kubernetes#138228][].
```
omerap12, line 358: The `HPAGeneration` feature gate is not yet locked to enabled. What happens if someone enables this feature gate while `HPAGeneration` is off?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+second; the actual steady-state rate is lower because each delay begins after
+reconciliation completes. Worker capacity is roughly
+`concurrentSyncs / meanReconcileDuration`. Once offered load exceeds capacity,
+every HPA's scheduling delay stretches, including HPAs that never set the field.
+A per-object admission floor does not help, because aggregate load depends on
+HPA count as much as on any individual delay.
+
+Mitigations and residual risk:
+
+- Administrators expecting short sync periods should raise
```
omerap12, line 268: How would an administrator know what value to raise this to? Since `--concurrent-horizontal-pod-autoscaler-syncs` is a kube-controller-manager argument, it requires a restart each time an administrator needs to test a new value.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+- [X] Metrics
+  - Metric name: `horizontal_pod_autoscaler_controller_reconciliation_delay_ratio`
+  - Components exposing the metric: `kube-controller-manager`
+- [ ] Events
+- [X] Other
+  - Controller logs at verbosity >= 4 include the configured and observed
+    scheduling delay for each HPA.
+
+A new gauge,
+`horizontal_pod_autoscaler_controller_reconciliation_delay_ratio{namespace,hpa_name}`,
```
omerap12, line 674: Is a gauge really the right metric type here? A gauge records only the latest value per HPA. If an HPA is consistently late but has one good cycle, the gauge won't show it.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,756 @@
...
+
+- Administrators expecting short sync periods should raise
+  `--concurrent-horizontal-pod-autoscaler-syncs`, and can bound HPA count per
+  namespace with a `ResourceQuota`. This will be documented.
+- The failure mode is graceful degradation -- longer scheduling delays --
+  rather than unbounded queue growth, memory growth, or dropped work.
+- The per-HPA delay ratio metric under
+  [Monitoring Requirements](#monitoring-requirements) makes the condition
+  observable.
+- Residual risk is accepted for Alpha, behind the gate. If Alpha feedback shows
```
omerap12, line 285: Is this acceptable? Why would we add a feature that has a high chance of causing API server problems? I think we should come up with a well-reasoned lower bound rather than just guessing. That said, sig-scalability or API review may have a different opinion here.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
adrianmoisey: The plan is to lock `HPAGeneration` to enabled in 1.38

omerap12: So, until `HPAGeneration` is locked to enabled, this is a blocker for us.

adrianmoisey: It is technically a blocker, but not one we can't solve

omerap12: Agree.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Makes sense, done. The field is now in both `autoscaling/v1` and
`autoscaling/v2` with the same semantics and validation, and Design Details
covers the round-tripping: every served version carries the field, so v1 <-> v2
conversion is just a copy, and a client that reads and writes back through
either version keeps the value. v2beta1 and v2beta2 aren't served, so they're
untouched.

I also wrote down how the validation actually works instead of just asserting
it's identical in both versions. The bounds go on the field as declarative
validation tags (`+k8s:minimum`, `+k8s:maximum`) in both external types, the
same way `minReplicas` and `maxReplicas` are already done, plus the hand-written
internal check marked `MarkCoveredByDeclarative()` while the migration is in
progress. For the gate, `dropDisabledFields` in
`pkg/registry/autoscaling/horizontalpodautoscaler` already does this exact
pattern for `HPAConfigurableTolerance`, so I point at it. It does need
restructuring though: today it returns early on the first enabled gate, which
doesn't work once there are two gated fields.

One thing I'd like your read on. `autoscaling/v1` is GA, so putting the field
there is a wider API review surface than a v2-only change would be. I think
it's the right shape long term, but it's probably worth flagging to API
reviewers up front instead of having it come up late.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Done. From master today: `pkg/apis/autoscaling/validation` is at 95.3% of
statements, `pkg/controller/podautoscaler` at 89.0%. I added
`pkg/registry/autoscaling/horizontalpodautoscaler` as well (56.9%), since the
disabled-field behavior in the test plan lives in the registry strategy rather
than in either of the other two packages.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Reworded. Alpha gets initial e2e coverage, and Beta is where it has to be
complete: both shorter and longer periods than the default, linked here, and
flake free for two weeks.

While I was in there I noticed the gate lifecycle was missing from both
criteria, so I added it. Enabled by default at Beta, locked to enabled at GA,
removed two releases after that.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Listed them out: the delay-ratio metric from clusters running with the gate on,
early adopters like the KEDA HTTP Add-on, and whatever comes back on the KEP
issue and in SIG meetings. SIG Scalability gets consulted before we relax the
lower bound.

Something more concrete on the first one. We plan to make the gate available in
OpenShift as soon as our feature-gate policy allows rather than waiting for
Beta, since customers have been asking for this for a while. It gets there
through the Custom Metrics Autoscaler (our downstream KEDA), which is what
generates the HPAs that want a short period, so the delay numbers should come
with real request-driven workloads attached instead of synthetic ones. I'll post
what we see on the KEP issue.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Added the reasoning: the lower-bound decision we make at Beta, plus any
workqueue fairness follow-ups, should have a release of production usage behind
them before the API gets locked at GA. That said, I'm not attached to two
releases. Happy to make it one if you'd rather.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: You're right, dropped it. GA now needs the two-release Beta bake, the gate
locked to enabled, and docs published.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Yep. Added an Alpha criterion for a kubernetes/website draft covering the
field, the best-effort semantics, and when it actually helps versus when it
doesn't.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Trimmed. It's down to the implementable milestone and the target release.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Written up in Design Details. I think it's less of a blocker than it looks, for
three reasons:

1. `HPAGeneration` has been Beta and on by default since 1.37, so the only
   clusters on the degraded path are ones that deliberately turned it off.
2. Even there it's bounded, and bounded by the resync period (15s by default)
   rather than by whatever `syncPeriodSeconds` used to be. The delaying
   workqueue only ever moves an entry's ready time earlier, so a re-enqueue
   carrying a shorter delay wins at the next resync, and a longer delay gets
   enforced by the eligibility deadline instead. Going from 3600s to 3s with the
   gate off still takes effect in about a resync period, not an hour.
3. If the lock to enabled slips in 1.38, we can drop the dependency entirely.
   Compare `spec` semantically in the controller's update handler, which is the
   same check the registry strategy already does in `PrepareForUpdate`, instead
   of reading `Generation`.

Periodic scheduling never touches `Generation` at all, so none of this is a
correctness question, just responsiveness.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Added sizing guidance to Risks and Mitigations. The starting point falls out of
the SLIs: to sustain an aggregate rate of `sum(1 / delay_i)` per second you
want roughly `concurrentSyncs >= aggregate rate * mean reconciliation
duration`, and the duration is already visible in
`horizontal_pod_autoscaler_controller_reconciliation_duration_seconds`.

Your point about the restart is fair, and it's now called out in the text: since
you can't iterate on this cheaply, size it with headroom before rolling out
short delays, then use the delay-ratio metric to check whether you got it right.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: You're right that reading a bare gauge hides this. The KEP now says SLO
evaluation and alerting should aggregate over a window (fraction of scrapes over
the threshold, or `avg_over_time`) rather than reading the instantaneous value,
so one good cycle can't cover for a consistently late HPA.

The reason I went with a per-HPA gauge instead of a histogram is that the
question operators actually ask here is which HPAs are late and how many, and a
histogram counts samples rather than HPAs, so short-delay HPAs would dominate
it. If windowed gauge queries turn out to be annoying in practice during Alpha,
adding an unlabelled histogram of the same ratio in Beta is cheap and we can
have both.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Fair, and I think the KEP was only answering half of this. Taking the halves
separately.

API server load first, where two things were missing.

Most of the additional traffic isn't storage traffic, which the Scalability
section now spells out. Per reconcile, two calls reach etcd: a `GET` on the
target's `scale` subresource, and an `UpdateStatus` on the HPA when the
recomputed status differs from the previous one. The rest is metrics queries,
which the apiserver authenticates and then proxies to whatever is registered for
`metrics.k8s.io`, `custom.metrics.k8s.io` or `external.metrics.k8s.io`. For the
request-driven workloads this feature is aimed at, those queries are most of the
increase, and the cost of serving them lands on metrics-server or the adapter
rather than on apiserver storage. Pod and HPA reads come out of informer caches
and don't change at all.

The second thing is that the controller's request path is already rate limited
at both ends. Every client is built from the kube-controller-manager client
config, so each one is capped client-side by `--kube-api-qps` /
`--kube-api-burst` (20 and 30 by default, one bucket per client), and APF caps
it server-side. So even if people do set aggressive delays widely, it can't turn
into apiserver saturation. It turns into longer scheduling delays inside the HPA
controller, which is the risk the KEP already spends most of its space on and
which the delay-ratio metric is there to expose.

Now the bound itself. You were right that it was a guess, so I reworked it to
derive 3s from controller safety and reversibility:

1. With fixed-delay semantics, a delay shorter than the reconciliation itself
   collapses into back-to-back reconciliation. At 3s, an HPA taking 1s to
   reconcile uses at most a quarter of one worker, so no single HPA can eat the
   pool.
2. The bound is relax-only. Lowering a minimum later makes previously invalid
   values valid, but raising it would invalidate values already persisted in
   etcd. 3s keeps the option of going lower in Beta once we have data, whereas
   1s would be a one-way door.

I also threw out the metrics-freshness argument I had in the first draft,
because it doesn't survive contact with the rest of the KEP. External and object
metrics can be as fresh as the query, which is the whole point of the feature,
and Knative's Pod Autoscaler ticks at 2s, which is below the floor rather than
above it. So the KEP now says the awkward part out loud: 3s plus a 1-2s
reconciliation is roughly 4-5s start to start, about twice Knative's cadence,
and that doesn't fully close the gap for the most latency-sensitive workloads.
It's a deliberate Alpha trade for a bound we can loosen later, not a claim that
nobody needs to go faster.

I added SIG Scalability to the Beta criterion for relaxing the bound too, so
that call isn't just ours.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Done, thanks. It's in Motivation as the original request, and as the first
Implementation History entry.

Fedosin:
Pushed an update covering this round. Replies are on the individual threads; a few changes don't map to any comment:

- The `TBD` under "Were upgrade and rollback tested?" is a real answer now. It can't be exercised until the field exists in a binary, so the section says that, then lists the three steps we'll run on a local cluster once the Alpha implementation merges (gate on, gate off, gate back on, checking persistence and dropping at each stage), with the results recorded before the Beta review.
- kubernetes/kubernetes#138222 is described as a closed prototype that will be reopened against whatever design lands here, rather than as an open implementation PR.
- Two metrics-server link definitions pointed at the same file, so I merged them.
- The PR description still said the field lives on `HorizontalPodAutoscalerBehavior`, which stopped being true when it moved to the top level of `spec`. Fixed.

Thanks for the detailed reviews.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: One clarification on the OpenShift side, since I was vague above. It lands as Tech Preview first, not as a generally supported option. Tech Preview clusters can't take minor upgrades, so nobody sensibly runs one in production, which means the early numbers come from preview and dev clusters rather than customer fleets. Production signal realistically arrives once this is Beta upstream and we can ship it supported.

Doesn't change the plan, just sets expectations on what kind of data we can bring back during Alpha.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
adrianmoisey: > One thing I'd like your read on. `autoscaling/v1` is GA, so putting the field
> there is a wider API review surface than a v2-only change would be.

Sorry, can you expand on this more, I'm not sure I understand.
Both v1 and v2 are GA.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
adrianmoisey: > While I was in there I noticed the gate lifecycle was missing from both
> criteria, so I added it. Enabled by default at Beta, locked to enabled at GA,
> removed two releases after that.

I believe it's removed 3 releases after being locked to default.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,887 @@
...
+- Documentation drafted on kubernetes/website describing the field, its
+  best-effort semantics, and when it is and is not effective
+
+#### Beta
+
+- The `HPAConfigurableSyncPeriod` feature gate is enabled by default.
+- The e2e tests described in the [`e2e tests` section](#e2e-tests), covering
+  both shorter and longer periods than the default, are implemented beyond the
+  initial Alpha coverage, linked in this KEP, and have run flake-free for at
+  least two weeks.
```
adrianmoisey: Is there a specific reason why this belongs in the beta stage? Why not add it to alpha?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: You're right, both are GA, so the concern doesn't apply. An alpha-gated field lands in a stable version either way, and there's nothing specific to v1 that widens the review. Disregard that part.

The field goes into both versions either way, as you asked.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: You're right, fixed. Rule #9 of the deprecation policy says a gate has to keep functioning for 6 months or 2 releases after the beta to GA transition, whichever is longer, and with three releases a year the two releases is what binds. So it goes non-operational at GA and comes out in the third release after. The GA criterion says that now and links the policy.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Fair point, moved. The only part that actually needs Beta is the soak time, so Alpha now covers the full e2e set from the test plan, both the shorter and the longer period, implemented and enabled. Beta keeps just linking them here and the two weeks flake free.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
adrianmoisey: Oh, it's possible that you are right.
See https://kubernetes.io/releases/version-skew-policy/#kube-apiserver

If the feature is locked in 1.35, then it can only be removed in 1.37 onwards, since that scenario ensures that all api-server instances have the gate enabled.

Fedosin: That lines up, and the skew angle is the clearer justification of the two. With one minor version of apiserver skew allowed in HA clusters, a 1.36 cluster can still be serving from 1.35 instances, so the flag has to stay accepted through 1.36 and 1.37 is the first release where dropping it is safe. That also settles how to read Rule #9's "no less than 2 releases": it counts the GA release plus the one after, so the two agree.

Put it back to two, worded as a floor rather than a fixed release: locked to enabled and non-operational at GA, removed no sooner than two releases later, with both policies linked.

elmiko (commented):
this is making good sense to me and is well written. i don't have any suggestions for changes but i do have a general question. if a user decreases the sync period for an HPA, will there be an associated increase in metrics production/storage?

if yes, is it large enough to warrant informing users about?

Fedosin:
> this is making good sense to me and is well written. i don't have any suggestions for changes but i do have a general question. if a user decreases the sync period for an HPA, will there be an associated increase in metrics production/storage?
> 
> if yes, is it large enough to warrant informing users about?

Thanks! Short answer: no real increase in metrics storage, though a few things tick faster.

The controller's own `horizontal_pod_autoscaler_controller_*` metrics don't grow. None of their labels depend on the period, so a shorter delay just bumps the same counters and histograms more often between scrapes; series
count and Prometheus storage are unchanged. The metrics pipeline (metrics-server, KEDA, etc.) gets queried more often, but that's read load, not storage, and it's what the Risks and Scalability sections already cover.

What does write more is HPA status: `status.currentMetrics` changes most cycles on an active workload, so the `UpdateStatus` rate scales with frequency. That's already in Scalability. Events don't add up either,
`SuccessfulRescale` only fires on an actual scale change, which `behavior` policies bound regardless of period, and repeated warnings are deduped and rate-limited by the client-go recorder.

So I don't think it needs a user-facing warning beyond what's there, but happy to add a sentence to Scalability making the fixed-cardinality point explicit if you think it'd help.

elmiko:
thanks @Fedosin , sounds good to me.

nothing more from my side, i'm lgtm

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
omerap12: > Most of the additional traffic isn't storage traffic, which the Scalability
section now spells out. 

Can you clarify what "storage traffic" means here? Are you distinguishing between calls that reach etcd and calls that the apiserver proxies to metrics extension servers?

> The rest is metrics queries, which the apiserver authenticates and then proxies to whatever is registered for
metrics.k8s.io, custom.metrics.k8s.io or external.metrics.k8s.io

So the apiserver still does real work for those queries right (authentication, proxying etc)? All consume resources even if nothing hits etcd. I'm not a sig-scalability or sig-api-machinery expert, so I think we need someone from those SIGs to weigh in on whether the proposed lower bound is safe.

> Every client is built from the kube-controller-manager client config, so each one is capped client-side by --kube-api-qps / --kube-api-burst

I don't think we should use rate limiting as justification for the bound being safe. If the feature is designed well, reasonable usage shouldn't be hitting those limits in the first place. 

> I added SIG Scalability to the Beta criterion for relaxing the bound too, so that call isn't just ours.

I think SIG Scalability should be involved at Alpha, not Beta.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,893 @@
...
+  whether cross-HPA interference warrants workqueue fairness or a
+  controller-level budget on aggregate reconcile rate.
+- Documentation covering when the field is and is not effective, and the
+  interaction with `--concurrent-horizontal-pod-autoscaler-syncs`.
+
+#### GA
+
+- The feature has been Beta for at least 2 releases with no major issues
+  reported, so the Beta lower-bound decision and any fairness follow-ups have
+  at least one release of production usage behind them before the API is
```
adrianmoisey: I believe once we change the API, it's there to stay, no matter the maturity of the feature.

I assume this should say "behind them before the feature-gate is locked" ?

Fedosin:
/cc @kubernetes/sig-scalability-pr-reviews

Requesting a SIG Scalability opinion on the API server cost model and the 3s lower bound (the "API server cost model" block under Risks and Mitigations, and the Scalability PRR section), as asked in the thread above. Context in #sig-scalability: https://kubernetes.slack.com/archives/C09QZTRH7/p1789392305438039. A note here on whether this is acceptable for Alpha -- or whether you'd prefer a higher floor, the status-write decoupling described there, or an aggregate control -- would be much appreciated.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Good pushback, and you're right on two of these. Taking them in order.

**"Storage traffic"** -- yes, that's the distinction I was drawing: calls that reach etcd (the `GET` on the scale subresource and `UpdateStatus` on the HPA) versus calls the apiserver authenticates and proxies to a metrics extension server. Poor choice of words on my side; the KEP now says "etcd-backed" and "aggregation-layer".

**Proxied queries still cost the apiserver** -- agreed. Authentication, authorization, APF classification and proxying all consume apiserver CPU and connections, even though nothing hits etcd. I've made the KEP say that explicitly instead of leaning on "not storage".

**Rate limiting as justification** -- agreed, and I've dropped it from the safety argument. A well-configured deployment should never approach `--kube-api-qps` or APF limits; they only describe the failure mode of a misconfigured one. The Risks section now makes the argument on the shape of the load instead: one reconciliation is ~1 etcd-backed read, at most 1 etcd-backed write, and one aggregation-layer query per metric spec, so the field multiplies an existing per-object cost by `15 / period` for HPAs that set it -- 3x at 5s, at most 5x at the 3s floor -- adding no new call type and no new scaling dimension. Two things keep that small in practice: the floor is a bound, not the expected value (5s or 10s cover most request-driven cases), and only the request-driven subset of HPAs -- serverless-style, scale-from-idle workloads on adapter metrics -- has any reason to set the field at all; the CPU/memory HPAs that make up most clusters gain nothing from it. A representative 1,000-HPA cluster with 50 such HPAs at 5s adds roughly 10% to the reconciliation rate. The upper bound is in the KEP for completeness: 1,000 HPAs all at 3s would be ~333 reconciliations/s, roughly 1,000 apiserver requests/s -- the same as 5,000 HPAs at today's default, which is valid right now with no bound at all. Aggregate load stays linear in HPA count, and admins control both factors. The bound's own job is narrower: keeping any single HPA from monopolizing a controller worker.

One more thing worth knowing while judging that model: the etcd *write* half of it is removable. It exists only because `status.currentMetrics` carries raw observed values that change almost every cycle, so the status write is a by-product of reporting, not of scaling. If the write rate at the floor turns out to be the concern, the mitigation held in reserve is to keep reconciling every `syncPeriodSeconds` but persist `currentMetrics` at most once per global period (and immediately on any replica/condition change) -- same shape as KEP-589 moving node heartbeats to `Lease`. That leaves the fast loop with one read plus the metrics queries. I've written it up in Risks as a reserve option rather than part of the Alpha design, since it would benefit every HPA and deserves its own change if needed.

**SIG Scalability at Alpha** -- fair, done. I've moved it from the Beta criteria to an Alpha graduation criterion and asked in #sig-scalability directly (link: https://kubernetes.slack.com/archives/C09QZTRH7/p1789392305438039). I've also /cc'd the sig-scalability-pr-reviews team in a separate comment on this PR so the review is on the record in this PR. wojtek-t sits on both sig-scalability and the PRR approvers list, so PRR will cover this angle as well, but I agree it should be visible before Alpha rather than assumed.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Right -- once the field is in a served version it's there regardless of maturity. Changed to "before the feature gate is locked to enabled". Thanks!

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+
+The per-HPA scheduling-delay gauge above does not exist today and is required by
+this feature; it is part of the Alpha graduation criteria rather than an
+outstanding gap. No additional metric is required for Alpha.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+No.
```
omerap12, line 807: I still didn't understand if `HPAGeneration` is mandatory for this KEP or not https://github.com/kubernetes/enhancements/pull/6004#discussion_r3879627767
If it is - it should be listed here.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+
+The user-visible consequence of a rollback is that HPAs relying on a short sync
+period revert to the global default and become less responsive. That is a
+latency regression rather than an outage: replica counts are still computed
+from the same metrics with the same algorithm.
+
+###### What specific metrics should inform a rollback?
+
+- `horizontal_pod_autoscaler_controller_reconciliation_delay_ratio` exceeding
+  1.2 across many HPAs, meaning configured scheduling delays are not being met.
```
omerap12, line 689: What's the basis for the 1.2 threshold?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+decision, so a stable HPA produces none. Access on a managed control plane
+depends on the provider exposing `kube-controller-manager` metrics.
+
+Because the gauge records only the most recent cycle, SLO evaluation and
+alerting should aggregate over a time window -- for example the fraction of
+scrapes above the threshold, or `avg_over_time` -- rather than reading
+instantaneous values, so that a single on-time cycle cannot mask a
+persistently late HPA. If Alpha usage shows windowed gauge queries are awkward
+in practice, an unlabelled histogram of the same ratio can be added in Beta to
+expose the distribution directly.
```
omerap12, line 769: I am not sure I understand. can you please elaborate shortly on this? previously you were saying the opposite about the metric type and here you are saying the opposite. I also asked about it in  https://github.com/kubernetes/enhancements/pull/6004/files#r3879699662

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+At least 99% of HPAs with a recorded periodic requeue should have
+`horizontal_pod_autoscaler_controller_reconciliation_delay_ratio` at or below
+1.2. Sustained violations are an operator-actionable capacity signal and should
+prompt rollback, longer configured delays, or an increase to
+`--concurrent-horizontal-pod-autoscaler-syncs`.
```
omerap12, line 777: Again, what's the basis for the 1.2/99% thresholds?

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+  the calls that reach etcd, and their rate is proportional to the configured
+  frequency.
+- **Aggregation-layer calls**: `metrics.k8s.io`, `custom.metrics.k8s.io`, and
+  `external.metrics.k8s.io` queries, which the `kube-apiserver` authenticates,
+  authorizes, and proxies to the extension API server registered for them.
+  These never reach etcd, but they are still API server work -- authentication,
+  authorization, APF classification, and proxying consume apiserver CPU and
+  connections -- and the cost of serving them lands on metrics-server or the
+  metrics adapter. For the request-driven workloads this feature targets, this
+  is where most of the additional traffic goes.
```
omerap12, line 832: Is that really necessary? the question is simply asks if any new API calls will be created. the first sentence is enough: "No new call types. A lower `syncPeriodSeconds` increases the frequency of
existing calls for that HPA."

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+Pod and HPA reads are served from informer caches and are unaffected by the
+configured period. The overall increase is approximately proportional when
+reconciliation duration and event-driven work are small relative to the
+configured delay: `15 / period` per HPA that sets the field, at most 5x at the
+3s floor. In expected usage only the request-driven subset of HPAs opts in, and
+often at 5s or 10s rather than the floor; 50 such HPAs at 5s in a cluster of
+1,000 add roughly 10% to the controller's reconciliation rate. As a worst-case
+anchor, 1,000 HPAs all at 3s issue about 333 reconciliations/s, or roughly
+1,000 API server requests/s -- equivalent to 5,000 HPAs at the default period.
+The full cost model is in [Risks and Mitigations](#risks-and-mitigations).
```
omerap12, line 843: Can you please elaborate where did you get those numbers? also same as above.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
```diff
@@ -0,0 +1,975 @@
...
+###### What are other known failure modes?
+
+- **Worker-pool exhaustion.** Short periods on many HPAs can drive the aggregate
+  requested reconcile rate above what the controller can sustain. The symptom is
+  observed scheduling delays exceeding configured ones for all HPAs, including
+  those that do not set the field. Detection is via the SLIs above. Mitigation
+  is to raise `--concurrent-horizontal-pod-autoscaler-syncs`, increase the
+  affected delays, or remove the field.
+- **No improvement on a resource-metric HPA.** Expected when the metrics source
+  is slower than the configured period; see [Proposal](#proposal).
```
omerap12, line 897: Is that a failure mode?

omerap12 (commented):
PRR shadow review comments.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Good catch, and thanks for pushing on it. I went back through the queue code and my earlier answer was wrong. The per-HPA delay was only set during reconciliation, so with `HPAGeneration` off, a change from 3600s to 3s could take up to an hour to apply, not 15s. So yes, it was effectively a dependency.

I've removed it instead of listing it. The update handler now refreshes the per-HPA delay from the new object on every update (resyncs included) and then enqueues the same way it does today. The workqueue only ever moves a waiting item earlier, so a shorter period kicks in after roughly the new delay, with the 15s resync as a backstop. A longer period costs at most one extra reconcile at the old delay. Nothing reads `Generation`, and status updates are still rate-limited, so the #42715 hot loop can't come back. The Dependencies section now says there's no dependency on the gate, and the tests cover it both on and off.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Honestly, there wasn't one. They were meant as a starting point, but written as an SLO they claim more than we actually know, especially since there's no HPA scale baseline to compare against. I've removed both numbers. The rollback signal is now the delay ratio getting worse over time compared to the cluster's own numbers before rollout, and picking a real threshold from Alpha data is now a Beta criterion.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Same as above, both numbers are gone. See the updated SLO section.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Fair point, that paragraph read like I was arguing both sides. I'm sticking with a per-HPA gauge, because the question operators ask is which HPAs are running late. A histogram without HPA labels records every cycle but can't tell you which HPA it was, and adding the labels multiplies the number of series by the bucket count. The downside is that the gauge only shows the latest cycle, so alerts should look at it over a time window, which also handles the "one good cycle hides a late HPA" case you raised earlier. I rewrote the section to say that plainly and dropped the "maybe a histogram in Beta" line, since it only made the choice look undecided.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: Agreed, it got bloated. It's down to those two sentences now, plus a link to the cost model in Risks and Mitigations.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: They're plain arithmetic, not measurements, and the KEP now shows the math. 1,000 HPAs at the 15s default is 1000/15 = 66.7 reconciliations/s. Moving 50 of them to 5s gives 950/15 + 50/5 = 73.3/s, so about 10% more. 1,000 HPAs at 3s is 1000/3 = 333/s, and the ~1,000 requests/s figure assumes one metric per HPA plus a status write every cycle (scale GET, metrics query, UpdateStatus). I also renamed "Upper bound" to "Stress anchor", because HPAs with several metrics make one extra query per metric, so it isn't really a ceiling. And the old text said 74/s, which was a rounding mistake. Fixed.

keps/sig-autoscaling/6003-configurable-hpa-sync-period/README.md
Fedosin: No, you're right. It's a configuration that doesn't help, not a failure. I removed it, since the Proposal already explains when a shorter period is useless.

Issue #6007, opened 2026-04-10
Add Topology Manager option to improve workload density for single-numa-node


jingczhang:
### Enhancement Description

- One-line enhancement description (can be used as a release note):
Add Topology Manager option to improve workload density for single-numa-node
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6008
- Discussion Link: https://github.com/nokia/kubernetes-enhancements/tree/single-numa-node-capacity/keps/sig-node/6007-prefer-most-allocated-numa-node
- PRs by stage and milestone:
  - [x] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6008
    - [ ] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/138321
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

tico88612:
Hello @jingczhang 👋, 1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [x] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

With all the PRR freeze requirements in place, this enhancement is now marked as `Tracked for PRR freeze`! Please keep the issue description up-to-date with appropriate stages as well.

[...]

PR #6008, opened 2026-04-10
KEP-6007: Add Topology Manager option to improve workload density for single-numa-node

jingczhang:
- One-line PR description:
Add Topology Manager option to improve workload density for single-numa-node

- Issue link:
https://github.com/kubernetes/enhancements/issues/6007

- Other comments:
HTML Link: https://github.com/nokia/kubernetes-enhancements/tree/single-numa-node-capacity/keps/sig-node/6007-prefer-most-allocated-numa-node
Kubelet PR: https://github.com/kubernetes/kubernetes/pull/138321


keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+
+## Summary
+
+Clusters that enforce strict NUMA locality with `topologyManagerPolicy: single-numa-node`
+often give up **workload density**: pods fail to schedule even when node has **enough CPUs
+in total**, because exclusive CPUs are fragmented across NUMAs.
+Among **tied** equally valid single-NUMA merged outcomes, Topology Manager today breaks ties
+with **bitmask / `Narrowest`** (equal width → **smaller mask / lower NUMA id**), **not**
```
pravk03: It would be good to clarify what `Narrowest` means in this context. I believe, its the number of numa nodes in the hint ?. And since its  `single-numa-node` the hint will always contain 1 NUMA node. We then break the tie by picking the node with the smallest ID ?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,267 @@
...
+where **allocatable** accounts for reserved resources so the ratio reflects true
+utilization.
+
+1. **CPU signal (static CPU manager):** Score each NUMA by exclusive-CPU utilization
+   (assigned / allocatable, where allocatable excludes reserved CPUs). Higher score wins.
+   Equal scores or non-static policy → undecided.
+2. **Memory signal (static memory manager):** Score each NUMA by regular-memory utilization
+   (assigned / allocatable, where allocatable excludes per-NUMA reserved memory). Higher score
+   wins. Equal scores or non-static policy → undecided.
+3. **Combine:** Neither decides → **`Narrowest`** fallback. One decides → use it.
```
pravk03: Lets consider the below scenario -

CPU manager allocation scores: NUMA 0 = 80%, NUMA 1 = 40% (NUMA 0 is better for CPU packing)
Memory manager allocation scores: NUMA 0 = 20%, NUMA 1 = 90% (NUMA 1 is better for memory packing)

In this case, I believe we would default to the `Narrowest Fallback` because the CPU and Memory managers do not agree on the most utilized NUMA node. 

My concern is that the number of cases where there is a agreement in CPU and Memory utilization percentages across NUMA nodes may be limited. Which means, we will frequently end up in the Narrowest fallback state (current behavior).

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,267 @@
...
+### Algorithm
+
+**Trigger:** `topologyManagerPolicy` is **`single-numa-node`**, **`prefer-most-allocated-numa-node`**
+is set in **`topologyManagerPolicyOptions`** (with required feature gates), and the hint merger is
+comparing **preferred** merged hints whose NUMA affinity is a **single** NUMA node (bit count 1).
+
+**Scoring (kubelet aggregator):**
+
+Each signal computes a **utilization score** per NUMA using the same formula as
+kube-scheduler's `MostAllocated` plugin: `score = (assigned × 100) / allocatable`,
```
pravk03, line 164: Can we include more details regarding how `assigned` is calculated for each resource type

For CPU, does it refer to currently allocated exclusive CPUS to other guaranteed pods ?

What about memory ?. Some questions that come to my mind are - 
* Do we only consider only `Guaranteed` pods ?. 
* If a guaranteed pod is assigned to multiple NUMA nodes, how do we attribute its memory request towards each node ?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+#### Integration / e2e
+
+- Multi-NUMA **static** CPU + `single-numa-node`: ordered admission of Guaranteed pods
+  (no devices) to validate **baseline** vs **option** NUMA choice when ties exist.
+- Scenarios where **one** NUMA already holds an exclusive / device-bound pod and pods could
+  admit to **either** NUMA—assert the option changes **which** NUMA wins vs low-index default.
+
+### Rollout and Documentation
+
```
pravk03, line 303: PRR sections needs to be added and we also need a PRR file.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,267 @@
...
+  prefer-most-allocated-numa-node: "true"
+topologyManagerPolicy: single-numa-node
+memoryManagerPolicy: Static
+cpuManagerPolicy: static
+...
+```
+
+### Algorithm
+
+**Trigger:** `topologyManagerPolicy` is **`single-numa-node`**, **`prefer-most-allocated-numa-node`**
```
pravk03, line 157: Have we considered allowing packing based only on CPU or only on Memory as separate configuration options ?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
pravk03: One option is to compute an aggregate score based on CPU and Memory Manager scores, similar to how its done in the scheduler.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: Yes, that's correct — and it's what the parenthetical in that sentence already describes: under single-numa-node all merged hints are single-bit (equal width), so Narrowest degenerates to "pick the lower NUMA id."

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: This is not an issue, it is the design intent. The intent is to pick the node with most CPU and memory usage, which is the most common pattern in Telco / single-numa-node deployments, and leave the rest unchanged.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The formula is already described in the KEP: score = (assigned × 100) / allocatable, same as kube-scheduler's MostAllocated plugin. "Assigned" comes directly from what static CPU/Memory managers track — exclusive CPUs and per-NUMA memory respectively — which by definition only covers Guaranteed pods. And under single-numa-node policy, a pod cannot span multiple NUMAs.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: You are right, strange I missed it, I will add them.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: No, there is no real-life need for separate options. The KEP intentionally keeps things simple and aligned with existing kube-scheduler behavior — the scheduler's MostAllocated scoring also combines resources together rather than offering per-resource knobs. Additionally, the algorithm already handles this naturally: if only static CPU manager is configured (memory manager is not static), only the CPU signal contributes and vice versa — no extra configuration needed.

jingczhang:
Hi @pravk03 thank you so much for the review. I have added the missing PRR section in the KEP. Hi @kannon92, I have added you as the approver in the PR file.

jingczhang:
Hi @ffromani, you are an expert on kubelet admission control, can you please help review the KEP, I appreciate it very much.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
pravk03: When I read `smaller mask`,  I initially thought you are referring to `CPUMask` until I looked into the code. You are referring to a Kubelet implementation detail here where NUMA nodes are represented as a bitmask when calculating hints and its worth calling that out IMO.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
pravk03: Sorry, I still fail to understand how this is not an issue. 

What be the behavior in the scenario in my comment above ? - Do we endup picking Node 0 because of fallback (smaller node ID)

To further clarify what I am referring to, consider another example -

CPU manager allocation scores: NUMA 0 = 40%, NUMA 1 = 80% (NUMA 1 is better for CPU packing)
Memory manager allocation scores: NUMA 0 = 80%, NUMA 1 = 79% (NUMA 0 is better for memory packing)

Here it is clearly better to pick NUMA 1 if the goal is to pick "most allocated", but my understanding of the described logic is, we endup picking NUMA 0 (because CPU and memory hints do not agree on the same NUMA node and we fall back to picking the smallest node id).

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: This is the current writing "Topology Manager today breaks ties with bitmask / Narrowest (equal width → smaller mask / lower NUMA id)", I will re-phrase it to "Topology Manager today breaks ties with Narrowest (fewest NUMA nodes in the affinity mask; under single-numa-node all masks have width 1, so this degenerates to lowest NUMA node ID)", hope this can address your concern.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: You raised a good point, the KEP intent is to align with kube-scheduler MostAllocated algorithm, the current KEP missed the aggregated step which aggregates "weighted average of all resource scores into one final number per node". I will make the correction.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,426 @@
...
+```yaml
+kind: KubeletConfiguration
+apiVersion: kubelet.config.k8s.io/v1beta1
+featureGates:
+  ...
+  TopologyManagerPolicyAlphaOptions: true
+topologyManagerPolicyOptions:
+  prefer-most-allocated-numa-node: "true"
+topologyManagerPolicy: single-numa-node
+memoryManagerPolicy: Static
```
pravk03, line 150: I believe its ok to have only one of static CPU or static Memory policy enabled ?. Would `prefer-most-allocated-numa-node` default to using only CPU signal, or memory signal in that case ? 

It would be good to cover that case.

jingczhang: Agreed, code already behaves this way, so I re-phrased the KEP accordingly: This KEP is targeted for both CPU Manager and Memory Manager using static policy. A non-static manager contributes zero scores for all NUMAs, so the feature falls back to a single-signal tie-break.

kannon92:
/assign @ShaanveerS 
for PRR shadow

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+### Goals
+
+- **Improve workload density** (better use of nodes)
+  for operators who use **`single-numa-node`** for strict locality (like in Telco).
+- Integrate via **`TopologyManagerPolicyOptions`**, using the same **feature gate /
+  graduation pattern** as for other topology policy options.
+
+### Non-Goals
+
+- Reimplementing **`NodeResourcesFit`** or other scheduler scoring. The NUMA-level scoring
```
ffromani: OK about reimplementing, but we can consume it. I mean: it's an option, not a requirement! Thing is the kubelet admission flow does call scheduler code already (and it's been doing so for a while):
considering k/k @ 7a1385a332e
```
  predicateAdmitHandler.Admit()                            # pkg/kubelet/lifecycle/predicate.go:119
  └── predicateAdmitHandler.generalFilter()                # pkg/kubelet/lifecycle/predicate.go:250
      │   (called at line 184)
      └── generalFilter()                                  # pkg/kubelet/lifecycle/predicate.go:432
          │   (called at line 253, retried at 269)
          └── scheduler.AdmissionCheck()                   # pkg/scheduler/eventhandlers.go:669
              │   (called at line 433)
              ├── noderesources.Fits()                     # .../plugins/noderesources/fit.go:643
              │   │   (called at line 671, draManager=nil)
              │   └── fitsRequest()                        # .../plugins/noderesources/fit.go:647
              │       │   (called at line 644)
              │       ├── computePodResourceRequest()      # .../plugins/noderesources/fit.go:317
              │       ├── check pod count vs AllowedPodNumber
              │       ├── check MilliCPU                   (line 668)
              │       ├── check Memory                     (line 678)
              │       ├── check EphemeralStorage            (line 688)
              │       └── check ScalarResources             (line 700)
              ├── nodeaffinity.Match()                      (line 685)
              ├── nodename.Fits()                           (line 691)
              └── nodeports.Fits()                          (line 697)
```

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+  **aligns with** the scheduler's `mostRequestedScore` formula but does not import or
+  duplicate scheduler code.
+
+**Side benefit:** Because the default kube-scheduler is **not NUMA-aware**, it may place a
```
ffromani, line 115: OTOH this may be the core reason not to consume the scheduler's NodeResourceFit code

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+## Proposal
+
+### User Stories
+
+1. **As a cluster operator** running `single-numa-node` with static CPU and Memory managers,
+   I want **better density** under strict single-NUMA locality—without switching to
+   **`best-effort`** (which results in cross-NUMA CPUs).
+2. **As a platform engineer**, I want an **opt-in** policy option so existing clusters keep
+   today’s behavior unless they enable the new option.
```
ffromani: this applies to a wide range of features so I'd not call it a user-story yet. Perhaps merge with the previous, flesh out or even drop it

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+### Proposed API
+
+Introduce a new Topology Manager policy option **`prefer-most-allocated-numa-node`**, configurable
+via kubelet configuration alongside existing options:
+- Only in effect when **`topologyManagerPolicy` is `single-numa-node`**.
```
ffromani, line 134: uhm, therefore maybe `prefer-most-allocated-single-numa-node`?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+This KEP is targeted for **both** CPU Manager and Memory Manager using **static** policy.
+A non-static manager contributes zero scores for all NUMAs, so the feature
+falls back to a single-signal tie-break.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Admission latency | Run scoring only on the tie path; reuse cached summaries where possible |
```
ffromani: more than cached, it seems to me the scoring is O(NUMA) meaning it needs to evaluate all the NUMA node candidates linearly. IOW, there's no combinatorial explosion risk and the scoring is scales linearly with the number of NUMA nodes.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+## Design Details
+
+### Kubelet wiring (container manager ↔ topology manager)
+
+Topology Manager is created **before** the CPU and Memory managers, so those dependencies are
+not available at `NewManager` time. Inside the **container manager**, after the CPU and Memory
+managers are constructed (and registered as topology hint providers), kubelet builds a **preferred
+NUMA tie-breaker** object that **holds references** to those two managers and registers it via
+**`TopologyManager.SetPreferredSingleNUMATieBreaker`**. The topology **Manager** forwards to
```
ffromani: Everything else equal, I'd adopt the Score-related terminology because ultimately this is what we are doing. Arguably, the `merge` lingo inside the topology manager is itself a scoring function we are now generalizing

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+from the respective managers, computes the **aggregate score** per NUMA as described under
+**Algorithm**, and returns the winner.
+
+### Test Plan
+
+[X] Owners of involved components may require updates to existing tests before merge.
+
+#### Unit tests
+
+Five test suites cover each layer of the feature:
```
ffromani, line 264: naming nit: I'd go for a simpler `cpu_scoring_test.go` - and likewise for everything else

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+5. **Policy option gating** (`pkg/kubelet/cm/topologymanager/policy_options_test.go`) —
+   `prefer-most-allocated-numa-node` accepted only with `TopologyManagerPolicyAlphaOptions`
+   enabled; rejected otherwise.
+
+#### Integration / e2e
+
+- Multi-NUMA **static** CPU + `single-numa-node`: ordered admission of Guaranteed pods
+  (no devices) to validate **baseline** vs **option** NUMA choice when ties exist.
+- Scenarios where **one** NUMA already holds an exclusive / device-bound pod and pods could
+  admit to **either** NUMA—assert the option changes **which** NUMA wins vs low-index default.
```
ffromani, line 300: I anticipate proper e2e testing would need to create specific node allocation states which are very trick to get right and reliable in the noisy CI environment. I recommend to plan large amount of time to write, tune and stabilize the e2e tests.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,428 @@
...
+Inspect the kubelet configuration on the node to verify `TopologyManagerPolicyAlphaOptions`
+is enabled and `prefer-most-allocated-numa-node: "true"` is set in
+`topologyManagerPolicyOptions`.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] Other (treat as last resort)
+  - Details: The operator observes increased workload density on the node — more
+    pods successfully admitted under `single-numa-node` policy compared to the
+    default tie-break behavior.
```
ffromani: I'd let the PRR folks comment here but we need to think extensively about how to have a decent observability. We start from a very poor base and I'd not sneak heavy refactoring in this PR (which would be welcome, though. Another story). But we totally need to ensure visibility for a change which is impacting, albeit opt-in, a behavior entrenched since day 1 of the topology manager.

Proper (contextual and structured) logs are a given. I'd actually block about metrics. We need to think about how we can surface the the scoring logic behavior through metrics.

ffromani (commented):
Initial review.

I'm in favor of the design, even though it's a relatively large albeit encapsulated resource manager change.
Designwise, except for naming nits, I don't see big problems. We can use a few round of polishing, but nothing major caught my eye.
The biggest single area on which we should spend some effort trying to improve is the observability, because the current baseline is not great.

ffromani:
Another important point I want to mention sooner than later: how the managers and the topology manager interact with each other in the context of this work is the next most-important question. Should we add a new interface the managers implement? should we extend HintProvider with new methods? or rather extend the _existing_ methods?
The posted implementation PR likely already answers these questions, but I think these should be detailed more in the KEP. I'll add my take in the next review round.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: Good point. I corrected the Non-goal accordingly: Reusing the scheduler's **`NodeResourcesFit`** scoring. Kubelet admission already  imports scheduler code—`predicateAdmitHandler` calls `scheduler.AdmissionCheck`,
  which runs `noderesources.Fits`, `nodeaffinity.Match`, etc.—but those checks operate at **node** granularity. This feature needs **per-NUMA** scoring using CPU and Memory manager state the scheduler does not have.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: yes, clarification is added when taking care of your previous comment.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: agreed, removed the 2nd user story.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: agreed, I updated the current statement for accuracy: Scoring runs only on the tie path and is O(NUMA nodes), typically 2–8 iterations of simple integer arithmetic

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The implementation actually uses both vocabularies, split by layer:
(1) Scoring layer (CPU manager, Memory manager, aggregator): uses score terminology throughout — singleNUMAScorer interface, GetNUMAUtilizationScores, getMostAllocatedNUMAScore.
(2) Topology manager integration point: uses tie-breaker terminology — PreferredSingleNUMATieBreaker interface, ComparePreferredSingleNUMAForTopology — because from the merger's perspective it's resolving a tie between two equally-preferred single-NUMA hints. The return type is (preferCandidate bool, ok bool), not a numeric score.
I think this split is accurate: the topology manager doesn't see scores, it just gets a comparison result. The scoring happens inside the tie-breaker implementation.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The option only applies to single-numa-node, so encoding that in the name makes the scope self-documenting. On the other hand, prefer-most-allocated-single-numa-node is 42 characters, and the validation already rejects it for other policies with a clear error. I'd lean toward keeping the shorter prefer-most-allocated-numa-node.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: agreed, will update accordingly.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: ok, the feature works in real life environment, so it is matter of carefully code those e2e tests for the CI environment. There is no e2e test in the kubelet PR yet, it will be managed in beta phase.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: I feel the need to push back a bit on this comment. The feature has a single impact: when there is a tie, it picks a different NUMA node than the lowest-ID one. It cannot change whether a pod is admitted or rejected — only which equally-valid NUMA wins, isn't it? I don't think new metrics are warranted for a tiebreaker but you are the guru in this area, please let me know what makes you worried here?

jingczhang:
> Another important point I want to mention sooner than later: how the managers and the topology manager interact with each other in the context of this work is the next most-important question. Should we add a new interface the managers implement? should we extend HintProvider with new methods? or rather extend the _existing_ methods? The posted implementation PR likely already answers these questions, but I think these should be detailed more in the KEP. I'll add my take in the next review round.

It is not as that complicated -) The feature simply introduces a callback in Topology Manager, and the callback is invoked when there is a NUMA node tie upon pod admission. I updated the KEP to make this clear.

jingczhang:
Hi @ffromani, thank you so much for the detailed review. I have taken care half of your comments and replied to the other half for you to review.

keps/sig-node/6007-prefer-most-allocated-numa-node/kep.yaml
```diff
@@ -0,0 +1,27 @@
...
+milestone: 
+  alpha: "v1.37"
+  beta: "v1.38"
+  stable: "v1.40"
+feature-gates:
+  - name: "TopologyManagerPolicyAlphaOptions"
+    components:
+      - kubelet
+disable-supported: true
+metrics: []
```
ShaanveerS: In the metric section of the main document you've used `topology_manager_admission_duration_ms`, please add it here as well.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,446 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements]
```
ShaanveerS, line 44: Please make sure to fill this out as you complete it.

ShaanveerS (commented):
PRR shadow review

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: ok, I don't have strong opinions here.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: I'm unsold about the tie-breaker naming, and I'll have another pass, but since this is internal API naming I don't think this is a blocker.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: The decision point needs to be observable, and for that logs are the minimum required tool. But that, albeit essential, gives us very poor obvservability at cluster scale. I'm trying to think if metrics make sense or if there are other solution available.

keps/sig-node/6007-prefer-most-allocated-numa-node/kep.yaml
jingczhang: done

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: ok, noted, will do.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: ok, thanks, resolved this item for now.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: ok, thanks, resolved this item.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: Would it be best to defer the deeper observability thinking to Beta, informed by real user feedback?

For the existing topology manager metrics: admission_duration_ms should be unaffected (the scoring adds negligible overhead), admission_requests_total should be unaffected (same pods go through admission), admission_errors_total we expect to trend lower over time for workloads that frequently hit single-NUMA capacity limits, since better packing preserves larger contiguous holes for subsequent admissions. I will add this in the KEP.

jingczhang:
Hi @dchen1107 @derekwaynecarr @mrunalp this KEP is targeting alpha for v1.37 and needs approval before enhancements freeze (June 16). Can you please review and approve? Thank you

jingczhang:
Hi @kannon92 would appreciate PRR review and approval when you get a chance, enhancement freeze is next Tuesday. Thank you.

kannon92:
Yea I'm aware and this is on my list.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: Fair enough. Gathering data makes sense. But we know already from past graduation and operational history that the topology manager observability is poor; adding metrics is _one_ vehicle to improve it, and one of the easiest. It doesn't necessarily makes sense.
Let's sketch out in the KEP which options we see feasible to improve the observability in this area we are touching, and let's sketch what logs need to make sure to expose.

ffromani:
> > Another important point I want to mention sooner than later: how the managers and the topology manager interact with each other in the context of this work is the next most-important question. Should we add a new interface the managers implement? should we extend HintProvider with new methods? or rather extend the _existing_ methods? The posted implementation PR likely already answers these questions, but I think these should be detailed more in the KEP. I'll add my take in the next review round.
> 
> It is not as that complicated -) The feature simply introduces a callback in Topology Manager, and the callback is invoked when there is a NUMA node tie upon pod admission. I updated the KEP to make this clear.

Sure, we can keep it simple. Let's sketch which implementation is recommended and why (e.g. why options have been discarded). As usual in KEPs, let's focus on the design level, no need to encode coding details in the design except for some possible `interface` sketches.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: I added the following in the KEP:
There is an existing `cpu_manager_allocation_per_numa` gauge that reports the
number of CPUs allocated per NUMA node, but no equivalent exists for memory.
In Beta, the following new metrics can be considered:
- `memory_manager_allocation_per_numa`: a per-NUMA gauge reporting memory bytes
  allocated per NUMA node, mirroring the existing CPU gauge.
- `topology_manager_utilization_per_numa`: a per-NUMA gauge reporting the
  aggregated utilization score that the tie-breaker uses to make decisions,
  giving operators direct visibility into NUMA packing density.
Real user feedback may dictate other needs, we will see.

jingczhang:
> > > Another important point I want to mention sooner than later: how the managers and the topology manager interact with each other in the context of this work is the next most-important question. Should we add a new interface the managers implement? should we extend HintProvider with new methods? or rather extend the _existing_ methods? The posted implementation PR likely already answers these questions, but I think these should be detailed more in the KEP. I'll add my take in the next review round.
> > 
> > 
> > It is not as that complicated -) The feature simply introduces a callback in Topology Manager, and the callback is invoked when there is a NUMA node tie upon pod admission. I updated the KEP to make this clear.
> 
> Sure, we can keep it simple. Let's sketch which implementation is recommended and why (e.g. why options have been discarded). As usual in KEPs, let's focus on the design level, no need to encode coding details in the design except for some possible `interface` sketches.

I polished the Design Details as such:
Under `prefer-most-allocated-numa-node` option, the topology manager needs to break
NUMA selection ties using utilization data it does not own. The CPU and Memory
managers hold this data.

**Design pattern: Strategy (via callback registration).** The topology manager
defines a tie-breaking interface. Any component with the knowledge to score
NUMA nodes registers a strategy. The topology manager only invokes it when a
tie needs breaking — it never reaches into manager internals.

**Interface sketch:**

```go
type PreferredSingleNUMATieBreaker interface {
    ComparePreferredSingleNUMAForTopology(current, candidate bitmask.BitMask) TieBreakResult
}
```

During kubelet startup, the container manager calls
`TopologyManager.SetPreferredSingleNUMATieBreaker(aggregator)` to register the
strategy. During pod admission, the topology manager calls
`aggregator.ComparePreferredSingleNUMAForTopology()` when a NUMA tie needs
breaking.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,480 @@
...
+
+**Scoring (kubelet aggregator):**
+
+Each signal computes a **utilization score** per NUMA using the same formula as
+kube-scheduler's `MostAllocated` plugin: `score = (assigned × 100) / allocatable`,
+where **allocatable** accounts for reserved resources so the ratio reflects true
+utilization.
+
+1. **CPU score (static CPU manager):** Score each NUMA by exclusive-CPU utilization
+   (assigned / allocatable, where allocatable excludes reserved CPUs).
```
pravk03, line 169: I think what you are referring to here is - cpu's reserved through `--reserved-cpus` flag on the numa node ? . 

Worth adding the detail if this is what you mean (or something else)

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,480 @@
...
+   Non-static policy → score 0 for all NUMAs.
+2. **Memory score (static memory manager):** Score each NUMA by regular-memory utilization
+   (assigned / allocatable, where allocatable excludes per-NUMA reserved memory).
```
pravk03, line 172: Could you clarify what you are referring to here ?.
I am not sure if there is a config for per-NUMA reserved memory

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,480 @@
...
+   Non-static policy → score 0 for all NUMAs.
+3. **Aggregate:** For each candidate NUMA, compute
+   `aggregateScore = (cpuScore + memoryScore) / 2`, mirroring the scheduler's
```
pravk03, line 175: I feel this simple aggregation might limit the usefulness of this policy. It's effective when either the static CPU policy OR the static Memory policy is enabled (which is still a win). But when BOTH are enabled, I suspect how useful it might be in a production environments with mixed workloads.

For Eg: The use case we have in `## Illustrative NUMA packing` section above would work only when memory utilization is also high in NUMA node 1 or static memory policy is disabled. In other cases, we would end up with the current behavior - scheduling on NUMA Node 0.

The way scheduler `MostAllocated` solves for this is by doing a weighted average and allowing for independent weights for resources. For a future milestone, do we want to consider something similar ?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: I purposely omitted those details since they are not changed by this KEP. I feel it's better we keep it this way so we can focus on the new feature design, especially as the KEP size already reaches XL now. Agree?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The reservedMemory field in KubeletConfiguration is per-NUMA ([reference](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/#kubelet-config-k8s-io-v1beta1-KubeletConfiguration)). You can see how the memory scorer consumes it in the implementation PR linked at the top of the page: https://github.com/kubernetes/kubernetes/pull/138321.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The equal-weight scoring mirrors kube-scheduler's default configuration for MostAllocated. Weighted scoring can be considered in Beta if real-world feedback demands it, but I'd rather not add knobs ahead of need (YAGNI). I'll add a note about this in the KEP.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: ok, this makes sense to me. Metricwise it's good.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,480 @@
...
+type PreferredSingleNUMATieBreaker interface {
+    ComparePreferredSingleNUMAForTopology(current, candidate bitmask.BitMask) TieBreakResult
+}
+```
+
+During kubelet startup, the container manager calls
+`TopologyManager.SetPreferredSingleNUMATieBreaker(aggregator)` to register the
+strategy. During pod admission, the topology manager calls
+`aggregator.ComparePreferredSingleNUMAForTopology()` when a NUMA tie needs
+breaking.
```
ffromani: I dislike the `TieBreaker` terminology significantly - I really believe this is a form of `Scoring`. I think the terminology makes some sense on 2-way NUMA machines but scales poorly to 4+ way NUMA machines.
But this is an internal detail we can change later.

It seems to me this is a must-have for resource managers so the only remaining question is why a new interface vs extending the `HintProvider` interface. The `why` should be elaborated.

ffromani (commented):
I think we are converging.
1. must-have: I'm happy with the identified metrics (solved!)
2. must-have: the logging portion is important and I'd (strongly) prefer to be mentioned in the KEP but we can get away with careful, possibly iterative implementation (solved)
3. must-have: we need a bit more discussion about the interface sketch. I don't have objections to add a new interface, but it deserves some rationale vs extending `HintProvider`.

I can have a final pass later but I'd expect that once item 3 is addressed (and possibly item 2 with some more fleshing?) it's LGTM.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: OK, thanks, I resolved this comment.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
pravk03: I am ok with starting simple for alpha and get feedback. However, I would still like to capture my concerns -  

> proven node-level algorithm to NUMA level — not to invent a new on

The scheduler already provides options to set scores  - https://kubernetes.io/docs/concepts/scheduling-eviction/resource-bin-packing/#enabling-bin-packing-using-mostallocated-strategy. 

I have confidence that this is a great addition and would work well when static CPU **OR** Memory policy is enabled. However, with BOTH policies enabled, a simple average score may dilute the score and lead more spreading compared to the current default of picking the lowest NUMA ID. This is a risk which is not documented.

jingczhang: You are right on both, kube-scheduler already supported weighted scoring, and it will only help it won't hurt. I added the following in KEP under "Notes / constraints / Caveats":
Kube-scheduler supports weighted scoring for `MostAllocated` and defaults to equal weight.
This KEP uses equal-weight aggregation between CPU and memory because CPU and memory usage
usually track together in production and rarely diverge.
If real-world feedback shows one resource dominates placement decisions, weighted scoring
can be added in Beta.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: I feel you :) You have a valid point; there is no good reason for using tieBreaker — it's not a better name — so I changed it to scorer. For why we don't extend HintProvider: simply because we are not adding a new topology hint, we are adding the capability to score already-merged hints. I made this clear in the KEP now. Please review.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] Other (treat as last resort)
+  - Details: When the scorer overrides the default NUMA selection, kubelet
+    emits an info-level log with the selected and default NUMA IDs plus CPU,
+    memory, and aggregate utilization scores for both. Operators can grep
+    kubelet logs for `"Prefer-most-allocated scorer overrode default NUMA selection"`
+    to confirm the feature is active and understand why each decision was made.
+
```
kannon92, line 408: Kubelet logs are really last resort.

You mention metrics for beta. I think these could be useful here to get an idea of placement.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+- Alpha: new option behind `TopologyManagerPolicyAlphaOptions`
+- User-facing docs: kubelet configuration reference, relationship to `single-numa-node` and
+  static managers
+- Release notes per phase.
+
+### Graduation Criteria
+
+- **Alpha:** Implementation + unit tests; documented semantics and known limitations.
+- **Beta:** e2e signal on multi-NUMA CI; no major semantic surprises from production feedback.
```
kannon92: Beta and GA need more details.

kannon92: - Beta:                                                                                                                                                                                                                                                                                  
     - e2e tests running in CI with multi-NUMA topology (specify which test jobs/boards)                                                                                                                                                                                                  
     - At least N users/operators providing feedback (or link to a feedback mechanism)                                                                                                                                                                                                    
     - New observability metrics added (the ones listed in the "missing metrics" section)                                                                                                                                                                                                 
     - Weighted scoring considered based on alpha feedback                                                                                                                                                                                                                                
 - GA:                                                                                                                                                                                                                                                                                    
     - e2e tests stable for at least 2 releases                                                                                                                                                                                                                                           
     - No open P0/P1 bugs related to the feature

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|------------|
+| Admission latency | Scoring runs only on the tie path and is O(NUMA nodes), typically 2–8 iterations of simple integer arithmetic |
+| Behavior surprise when enabled | Opt-in; metrics `topology_manager_admission_*` already exist |
+| Placement affects subsequent admissions | The feature cannot fail the current pod's admission (it only selects among already-valid NUMAs), but placement choices are designed to improve admission for subsequent pods |
+
+## Design Details
```
kannon92, line 211: This is maybe unrelated but I would be curious on how this feature could be supported in dra-driver-cpu.

https://github.com/kubernetes-sigs/dra-driver-cpu

kannon92, line 211: AI help here. Can you add some comments on what other policies are possible with this option? 

### 7. Consider documenting interaction with best-effort policy                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                          
 The KEP correctly states the option is only valid with single-numa-node policy. It would be good to explicitly document what happens if someone sets prefer-most-allocated-numa-node: "true" with a different policy (e.g., best-effort or restricted). From the implementation PR, it   
 appears it would be rejected at validation — documenting this in the KEP would be clearer.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+  admit to **either** NUMA—assert the option changes **which** NUMA wins vs low-index default.
+
+### Rollout and Documentation
+
+- Alpha: new option behind `TopologyManagerPolicyAlphaOptions`
+- User-facing docs: kubelet configuration reference, relationship to `single-numa-node` and
+  static managers
+- Release notes per phase.
+
+### Graduation Criteria
```
kannon92, line 309: Don't forget that for beta you will also use TopologyManagerPolicyBetaOptions.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+
+## Drawbacks
+
+- More logic and coupling between Topology Manager and CPU/Memory manager state.
+
+## Alternatives
+
+1. **Status quo:** Keep bitmask / low-NUMA-id selection; accept lower density on **asymmetric**
+   nodes when symmetric pods **always** stack on the lower-id NUMA first.
+
```
kannon92, line 509: There is a section usually on infrastructure needed? Do we have the CI necessary to test this feature?

Are you able to run tests with different number of numa nodes (2, 4, 8?)?

ffromani:
/lgtm
all my comments have been addressed. From my POV it's good for the first alpha.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: You mean adding new metrics already in Alpha? The three existing topology manager metrics (admission_duration_ms, admission_requests_total, admission_errors_total) already cover admission health. The new ones (memory_manager_allocation_per_numa, topology_manager_utilization_per_numa) are additive visibility — useful but not essential to validate the feature. The kubelet PR is already XL without them, so I'd prefer to defer to Beta as agreed with @ffromani. Acceptable?

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: Updated KEP per your comment, including the metrics in Beta

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
swatisehgal: This is a good point, we should explicitly document behavior when this policy option is set with a non `single-numa-node` policy.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
```diff
@@ -0,0 +1,495 @@
...
+###### What are other known failure modes?
+
+None. The feature is a scorer that only selects among already-valid NUMA outcomes.
+It cannot cause admission failures or affect running workloads.
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+This feature is not expected to impact any SLOs. If issues are observed, disable the
+option and restart kubelet to rule it out.
```
swatisehgal: Not expecting any impact on SLOs is too strong. At minimum, we should inspect metrics and logs to detect any latency or failure impact.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
ffromani: in a nutshell, supporting rebalancing and workload distributionis something we can do in the `dra-driver-cpu` and in genera in a DRA-powered scenario. The fine details change significantly though. For example, part of the allocation responsability would be transferred in the scheduler, which in turn will be heavily influenced by how the driver exposes the allocatable CPUs (grouped vs individual for example).

jingczhang: NUMA affinity for DRA devices is handled at the scheduler via device attributes, not by kubelet Topology Manager. So DRA is out of scope here.

jingczhang: The option is only consumed by the single-numa-node policy — other policies don't have a NUMA scorer, so the option is a no-op. I added a note to the KEP clarifying this.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: This is taken care of as part of your first comment.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: The existing topology manager e2e tests already run on multi-NUMA CI nodes (2 and 4 NUMA). E2e tests for this feature are planned for Beta and will use the same CI infrastructure. I updated the KEP to make this clear.

jingczhang:
Hi @kannon92, thank you for the review, I have addressed all your comments, please review and approve.

keps/sig-node/6007-prefer-most-allocated-numa-node/README.md
jingczhang: Good point. I updated the KEP to reference the existing metrics to inspect.

kannon92:
PRR looks good for alpha.

/approve

jingczhang:
Hi @dchen1107 @derekwaynecarr @mrunalp, sorry to bother again, this KEP is targeting alpha for v1.37 and needs approval before enhancements freeze tomorrow. Can you please review and approve? Thank you

pravk03:
/lgtm
Look good to me for alpha. Thanks for addressing the comments.

jingczhang:
Hi @haircommander @SergeyKanzhelev could one of you please review and approve? Enhancement freeze is today. Thank you.

kad:
After looking at this and thinking in more details, wouldn't it be better for each manager to return possible mask not as a single value, but as an ordered list based on the consumption? So, each manager (cpu / memory / device) are authoritative for making order of preference (e.g. based on utilization), but then TopologyManager can have policy to start processing those hints lists from start or from the end element?
That would lead to change in APIs between managers, but algorithm might become more flexible and more scalable in overall.

jingczhang:
> After looking at this and thinking in more details, wouldn't it be better for each manager to return possible mask not as a single value, but as an ordered list based on the consumption? So, each manager (cpu / memory / device) are authoritative for making order of preference (e.g. based on utilization), but then TopologyManager can have policy to start processing those hints lists from start or from the end element? That would lead to change in APIs between managers, but algorithm might become more flexible and more scalable in overall.

That's an interesting direction, but it would require refactoring all existing hint providers — each would need to manage most-allocated (or least-allocated, weighted, etc.) policy options internally. That's a significantly larger scope. The current design keeps that logic centralized in the scorer, leaving hint providers unchanged. In terms of flexibility and scalability, the centralized scorer approach seems preferable — new scoring strategies can be added without touching any hint provider.

kad:
Centralized scorer means that TopologyManager need to look into internal state data of the CPU manager, Memory Manager and Device Managers (the only 3 pleases that are providing hints to TopologyManager). This practically eliminates architectural responsibility splits between those components. Each manager (cpu, memory, device) has internal state and that state tracks current utilization. Returning ordered "mask" still can keep "spread" policy inside topology manager, but state tracking and score calculation more inside of individual component.

jingczhang:
> Centralized scorer means that TopologyManager need to look into internal state data of the CPU manager, Memory Manager and Device Managers (the only 3 pleases that are providing hints to TopologyManager). This practically eliminates architectural responsibility splits between those components. Each manager (cpu, memory, device) has internal state and that state tracks current utilization. Returning ordered "mask" still can keep "spread" policy inside topology manager, but state tracking and score calculation more inside of individual component.

Yes, but the Topology Manager doesn't reach into manager internals — each manager exposes a single method (GetNUMAUtilizationScores) that returns opaque scores. The internal state (CPU assignments, memory blocks, topology details) stays encapsulated. Managers are data sources, not decision makers. The scoring logic (aggregation, weights, comparison direction) is centralized in NUMAScorerAggregator. For illustration, adding a prefer-least-allocated-numa-node option would be ~10 lines — just flip the comparison in the aggregator. No manager changes needed.

kad:
`GetNUMAUtilizationScores` doesn't exist currently, so it is anyway new interface between topology manager and rest of managers. What I mentioned above, means improving hint mechanism (returned list can include ordered list which includes zone number and score(s)). Single call/request to managers, instead of potentially race of generating hints, then secondary call to get scores....

jingczhang:
> `GetNUMAUtilizationScores` doesn't exist currently, so it is anyway new interface between topology manager and rest of managers. What I mentioned above, means improving hint mechanism (returned list can include ordered list which includes zone number and score(s)). Single call/request to managers, instead of potentially race of generating hints, then secondary call to get scores....

There's no race — pod admission is serialized and manager state doesn't change between hint generation and scoring. The key difference with your suggestion is your requires changing the interface between topology manager and hint providers — the current TopologyHint bitmask can't carry per-NUMA utilization data. That change affects all providers and all policies, not only single-numa-node. Each manager would also need to know whether scoring is enabled to avoid spending cycles computing utilization and ordering when it's not needed.

kad:
Race potential can be due to stopping of some pods inbetween calls to collect secondary hints from providers.

pravk03:
> This practically eliminates architectural responsibility splits between those components. Each manager (cpu, memory, device) has internal state and that state tracks current utilization. Returning ordered "mask" still can keep "spread" policy inside topology manager, but state tracking and score calculation more inside of individual component.

I think this is a good point. Combining the hint generation and scoring stages into a single call maybe cleaner and avoids additional call (GetNUMAUtilizationScores) from topology manager to individual resource managers. 

> wouldn't it be better for each manager to return possible mask not as a single value, but as an ordered list based on the consumption? So, each manager (cpu / memory / device) are authoritative for making order of preference (e.g. based on utilization), but then TopologyManager can have policy to start processing those hints lists from start or from the end element?

Another option is to pass the score as an explicit field inside [TopologyHint struct](https://github.com/kubernetes/kubernetes/blob/c5c431b186af47a36e26bfd76bf22913f7570ac7/pkg/kubelet/cm/topologymanager/topology_manager.go#L134) to avoid implicitly encoding the score in the sort order ? 
 
Wondering if this can be considered as an implementation detail and addressed in the linked implementation PR ? 
@kad @ffromani @dchen1107

ffromani:
> > This practically eliminates architectural responsibility splits between those components. Each manager (cpu, memory, device) has internal state and that state tracks current utilization. Returning ordered "mask" still can keep "spread" policy inside topology manager, but state tracking and score calculation more inside of individual component.
> 
> I think this is a good point. Combining the hint generation and scoring stages into a single call maybe cleaner and avoids additional call (GetNUMAUtilizationScores) from topology manager to individual resource managers.

I'm very open to that. I asked about API and interface shape earlier in review and had a similar possible path in the back of my head. I was toying with the idea to pursue this path described here as followup after the work laid out in the KEP lands, so I'd surely supportive of experimentation like this in the context of this KEP.

> > wouldn't it be better for each manager to return possible mask not as a single value, but as an ordered list based on the consumption? So, each manager (cpu / memory / device) are authoritative for making order of preference (e.g. based on utilization), but then TopologyManager can have policy to start processing those hints lists from start or from the end element?

Arguably `Preferred` in the existing `TopologyHint` shape is already limiting. An ordered list is better in almost every dimension, the only possible exception being the struct size which is technically true but totally negligible IMO.

This is another API change I'd be super-favorable to see (or at least research) and likewise moving in this direction is fine by me.

> Another option is to pass the score as an explicit field inside [TopologyHint struct](https://github.com/kubernetes/kubernetes/blob/c5c431b186af47a36e26bfd76bf22913f7570ac7/pkg/kubelet/cm/topologymanager/topology_manager.go#L134) to avoid implicitly encoding the score in the sort order ?
> 
> Wondering if this can be considered as an implementation detail and addressed in the linked implementation PR ? @kad @ffromani @dchen1107

My 2c: it may be the case depending on how we do.

Note these are internal API, so it's AFAIK fair to drift a bit from the KEP sketch in the alpha/beta phases if
1. the internal implementation does NOT alter the user-visible API and contract
2. the changes bubble up in the KEP text on next update

kad:
> Arguably `Preferred` in the existing `TopologyHint` shape is already limiting. An ordered list is better in almost every dimension, the only possible exception being the struct size which is technically true but totally negligible IMO.


Some of the background reasons why I proposed that was to eliminate old algorithm that was comparing bitmasks one bit at a time and not really scalable above 4 numa nodes in the system. Fixing this internal hint api between set of managers will drastically improve performance of them and give a bit more flexibility to extend if we really want in some future.  It is indeed internal API that can be adjusted during implementation phase, but at least I wanted to have in the KEP description that utilization is based on logic inside individual manager (cpu, memory, device).

kad:
BTW, if I remember text of KEP correctly, the device manager is considered out-of-scope, however it is one of the important ones if we want to optimize placement of workloads that consume several devices at a time from device plugins. So, it should be in the scope of the algorithm.

jingczhang:
Hi @kad @ffromani @pravk03, sorry for the late reply. I believe what @kad is proposing is a different solution, not an implementation detail of this KEP — it changes the HintProvider interface and hence all Hint Providers. I understand the desire to restructure how managers and the Topology Manager interact more broadly, but that is a larger effort, better left to sig-node maintainers like you. If sig-node decides to pursue the HintProvider interface changes, a feature like this could be re-proposed on top of the new foundation. I will withdraw this KEP unless you want to discuss more. Thank you.

Issue #5982, opened 2026-03-27
KEP: ReplicaSet Consolidation-Aware Scale-In Strategy


nathangeology:
## Summary

This issue tracks the KEP for adding a consolidation-aware scale-in heuristic to the ReplicaSet controller's pod deletion algorithm.

When the `ConsolidatingScaleDown` feature gate is enabled, the ReplicaSet controller prefers deleting pods on nodes with fewer total active pods during scale-down, enabling workload consolidation onto fewer nodes. It also respects do-not-disrupt signals so that pods on protected nodes are deprioritized for deletion.

## Motivation

The current ReplicaSet scale-down algorithm prefers deleting pods on nodes with *more* colocated replicas (a spreading heuristic). While this promotes even distribution, it actively works against node consolidation. Node autoscalers (Karpenter, cluster-autoscaler) can only reclaim nodes when workloads consolidate during scale-down, but the spreading heuristic distributes deletions evenly, leaving every node partially occupied.

KEP-2255 (PodDeletionCost) provides a mechanism via annotations, but requires an external controller to continuously update annotations before scale-down events — operationally complex and API-server intensive.

## Key Design Points

- **Feature gate**: `ConsolidatingScaleDown` (kube-controller-manager)
- **Owning SIG**: sig-apps
- **Target milestone**: v1.37 (alpha)
- **Generic do-not-disrupt annotation** to avoid coupling with any specific node autoscaler
- **Complementary to KEP-2255** (PodDeletionCost) — both mechanisms coexist, with PodDeletionCost taking precedence in the existing sort order
- **Zero overhead when disabled** — conditional node informer initialization

## Related

- KEP-2255: PodDeletionCost
- kubernetes/kubernetes#107598
- kubernetes/kubernetes#123541

## KEP

KEP directory: `keps/sig-apps/NNNN-replicaset-consolidation-scale-in/`

/sig apps
/stage alpha
/milestone v1.37

BenTheElder:
There's a standard issue template for KEPs with required metadata, can you update the issue body to reflect the template?

You can view https://github.com/kubernetes/enhancements/issues => "Enhancement tracking issue"

Removed "KEP:" from title, by default all issues here are KEPs and the issue number is the KEP identifier.

To file a KEP you will need more than the issue, the summary/motivation/design etc should be in a PR following the standard enhancement proposal format where it will be reviewed, not in the KEP issue.

https://github.com/kubernetes/enhancements/blob/master/README.md has the process.

[...]

PR #6019, opened 2026-04-15
KEP-5982: ReplicaSet Consolidation-Aware Scale-In Strategy

nathangeology:
Provisional KEP for adding a consolidation-aware pod deletion heuristic to the ReplicaSet controller.

Tracking issue: https://github.com/kubernetes/enhancements/issues/5982

/sig apps

### DIFF
--- a/keps/sig-apps/5982-replicaset-consolidation-scale-in/README.md
+++ b/keps/sig-apps/5982-replicaset-consolidation-scale-in/README.md
@@ -1,0 +1,42 @@
+# KEP-5982: ReplicaSet Consolidation-Aware Scale-In Strategy
+
+<!-- toc -->
+<!-- /toc -->
+
+## Summary
+
+Add a `ConsolidatingScaleDown` feature gate to `kube-controller-manager` that changes the ReplicaSet controller's pod deletion sort order during scale-down. When enabled, the controller prefers deleting pods on nodes with fewer total active pods, enabling node autoscalers to reclaim nodes with less disruption.
+
+## Motivation
+
+The ReplicaSet controller's current scale-down heuristic spreads pod deletions evenly across nodes. This works against node autoscalers (Karpenter, cluster-autoscaler) which need nodes to become empty or underutilized before they can remove them. The spreading heuristic leaves every node partially occupied after a scale-down, requiring the autoscaler to evict additional pods to consolidate — increasing total disruption.
+
+### Goals
+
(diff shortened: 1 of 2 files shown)


keps/sig-apps/5982-replicaset-consolidation-scale-in/kep.yaml
```diff
@@ -0,0 +1,29 @@
+title: ReplicaSet Consolidation-Aware Scale-In Strategy
+kep-number: 5982
+authors:
+  - "@nathangeology"
+owning-sig: sig-apps
+participating-sigs:
+  - sig-autoscaling
```
DerekFrank, line 7: sig-apps too, no?

keps/sig-apps/5982-replicaset-consolidation-scale-in/README.md
```diff
@@ -0,0 +1,42 @@
+# KEP-5982: ReplicaSet Consolidation-Aware Scale-In Strategy
+
+<!-- toc -->
+<!-- /toc -->
+
+## Summary
+
+Add a `ConsolidatingScaleDown` feature gate to `kube-controller-manager` that changes the ReplicaSet controller's pod deletion sort order during scale-down. When enabled, the controller prefers deleting pods on nodes with fewer total active pods, enabling node autoscalers to reclaim nodes with less disruption.
```
atiratree, line 8: I think this is an interesting option to consider, but it conflicts slightly with https://github.com/kubernetes/kubernetes/pull/80004. This KEP should clarify where the `ConsolidatingScaleDown` feature will be placed in the sorting/ranking of the pods. We should also explore the consequences of the various sort/rank order options.

nathangeology: Yeah, this would be an alternate heuristic here. The current one is a good one if you do want to spread out pod replicas, but if someone has dynamic capacity and wants to control their compute costs then this proposed heuristic would be really useful. You could imagine other cases (e.g. load-aware or scheduling constraints aware), but then the question is where the 'brain' for parsing through these competing concerns should live. Once you move beyond a simple heuristic change, you probably don't want to do it in the replicaset controller.

keps/sig-apps/5982-replicaset-consolidation-scale-in/kep.yaml
nathangeology: Does the owning sig need to be in the participating sigs too?

PR #6025, opened 2026-04-21
KEP-2021: Promote to beta

johanneswuerbach:
- Promote KEP-2021 to beta.

- Issue link: https://github.com/kubernetes/enhancements/issues/2021

- Rendered version https://github.com/johanneswuerbach/enhancements/blob/kep-2021-beta/keps/sig-autoscaling/2021-scale-from-zero/README.md

### DIFF
--- a/keps/sig-autoscaling/2021-scale-from-zero/README.md
+++ b/keps/sig-autoscaling/2021-scale-from-zero/README.md
@@ -130,18 +130,18 @@ checklist items _must_ be updated for the enhancement to be released.
 
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
-  - [ ] e2e Tests for all Beta API Operations (endpoints)
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [x] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
+- [x] (R) Graduation criteria is in place
   - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
 - [ ] (R) Production readiness review completed
 - [ ] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
+- [x] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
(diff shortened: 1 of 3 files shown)


keps/sig-autoscaling/2021-scale-from-zero/README.md
```diff
@@ -472,12 +476,24 @@ in back-to-back releases.
...
+  (<https://github.com/kubernetes/kubernetes/pull/138464>) and a negative case
+  ensuring HPAs configured with resource (CPU) metrics are not scaled to zero
+- Production readiness review approved for beta
+- User-facing documentation updated in `kubernetes/website`
 
 #### GA
 
-- Allowing time for feedback
+- `HPAScaleToZero` feature gate has been enabled by default in beta for at
+  least two releases without blocking bugs
```
adrianmoisey: I don't think this is a requirement, I think we can just have a single release in beta before it goes GA... assuming no feedback that needs addressing

adrianmoisey (commented):
Just one small comment

keps/sig-autoscaling/2021-scale-from-zero/kep.yaml
```diff
@@ -16,17 +16,17 @@ see-also:
 replaces:
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: alpha
+stage: beta
```
kannon92, line 19: When changing stage (beta), please request a PRR review.

ie add a beta field [here](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-autoscaling/2021.yaml)

keps/sig-autoscaling/2021-scale-from-zero/README.md
johanneswuerbach: Removed.

keps/sig-autoscaling/2021-scale-from-zero/kep.yaml
johanneswuerbach: Added.

keps/sig-autoscaling/2021-scale-from-zero/README.md
```diff
@@ -493,9 +509,9 @@ enhancement:
...
+As this KEP changes the allowed values for `minReplicas`, special care is required for the downgrade case to not prevent any kind of updates for HPA objects using `minReplicas: 0`. API validation has accepted `minReplicas: 0` with the `HPAScaleToZero` feature gate enabled since Kubernetes 1.16, so downgrades to any version >= 1.16 will not reject existing HPA objects.
 
-Before downgrading all HPAs need to be set to `minReplicas: 1` to avoid any deployments being stuck at `replicas: 1`.
+Before downgrading to a version without the condition-based implementation, all HPAs using `minReplicas: 0` should be set to `minReplicas: 1` and their workloads scaled to at least one replica, otherwise workloads currently scaled to `replicas: 0` may remain stuck at `replicas: 0` (the old controller cannot distinguish "manually paused" from "HPA scaled to zero" without the `ScaledToZero` condition).
```
adrianmoisey, line 514: I assume this only applies when downgrading from 1.36 to 1.35?

Downgrading from 1.37 to 1.36 should be safe, right? (ie: the controller can scale up from zero, delete the condition and continue as if the feature gate is now disabled)

adrianmoisey:
/ok-to-test
/lgtm 
(I just left 1 minor comment)

keps/sig-autoscaling/2021-scale-from-zero/README.md
johanneswuerbach: This would only be safe if the gate was enabled, if the gate is disabled, the behaviour is the same as in 1.35 today.

keps/sig-autoscaling/2021-scale-from-zero/README.md
sohankunkerkar: The version skew section says the gate lives in both kube-apiserver and kube-controller-manager, but this only lists kube-apiserver. kep.yaml is also missing kcm. Should both be listed?

sohankunkerkar: This still says "not yet" but the condition-based implementation merged in 1.36 (kubernetes/kubernetes#135118). Can you update with manual test results?

sohankunkerkar, line 822: Scaling 0→1 is a new latency path that doesn't exist today. Have you measured how long it takes? Even a rough number tied to the HPA reconcile interval would help.

sohankunkerkar (commented):
I have added a few comments from the PRR shadow reviewer side. PTAL. Thanks!

keps/sig-autoscaling/2021-scale-from-zero/README.md
johanneswuerbach: Yes, updated.

johanneswuerbach: Added a note.

johanneswuerbach: Updated.

adrianmoisey:
From sig-autoscaling: 
/lgtm 
/approve 

For PRR:
/assign johnbelamaric

Issue #6035, opened 2026-04-28
Exec session identity propagation


nispriha:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Propagate the audit request ID from the API server through the kubelet to the CRI ExecRequest, enabling runtime security tooling to attribute exec session activity to the originating Kubernetes user.
- Kubernetes Enhancement Proposal:  https://github.com/kubernetes/enhancements/pull/6036
- Discussion Link:  
https://kubernetes.slack.com/archives/C019LFTGNQ3/p1774874170937829
  - https://blog.sigtrapd.dev/posts/k8s-exec-identity-gap/
  - https://blog.sigtrapd.dev/posts/exploring-the-exec-identity-gap/
  - https://blog.sigtrapd.dev/posts/bridging-k8s-exec-identity-gap/

- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6036
    - [ ] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/138752
    - [ ] Docs ( `k/website` ) update PR(s):
      -    Documentation updates:
            - [ ] https://github.com/kubernetes/website/pull/56339
      -    Release blog:
           - [ ] https://github.com/kubernetes/website/pull/56458

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._
cc: @haircommander @ngopalak-redhat @sigtrapd

ngopalak-redhat:
I have currently assigned this to sig-node as we know that kubelet is involved, this could impact another sigs too. Will update

kikisdeliveryservice:
Hi @nispriha 

The Discussion link section is for demonstrating that you have discussed this with a sponsoring SIG ahead of opening the issue/pr. You have instead included blog posts. Please update the issue description to include a discussion link to confirm that you've brought your idea to a SIG and have some support to move this forward.

ngopalak-redhat:
@kikisdeliveryservice  Thanks, @nispriha Can you add this link to the description: https://kubernetes.slack.com/archives/C019LFTGNQ3/p1774874170937829

nispriha:
Thanks @kikisdeliveryservice @ngopalak-redhat , I have updated the description to add the discussion link.

kikisdeliveryservice:
Great, thanks for the update @nispriha !

[...]

PR #6036, opened 2026-04-28
KEP-6035: Add KEP for exec session identity propagation

nispriha:
- One-line PR description: Add KEP for propagating audit request ID from the API server through the kubelet to the CRI ExecRequest.

- Issue link: https://github.com/kubernetes/enhancements/issues/6035
- Implementation PR: https://github.com/kubernetes/kubernetes/pull/138752

- Other comments:

### DIFF
--- a/keps/sig-node/6035-exec-session-identity/README.md
+++ b/keps/sig-node/6035-exec-session-identity/README.md
@@ -65,7 +65,7 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 - [ ] (R) Production readiness review approved
 - [ ] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [ ] Supporting documentation, e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 [kubernetes.io]: https://kubernetes.io/
 [kubernetes/enhancements]: https://git.k8s.io/enhancements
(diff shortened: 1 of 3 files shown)


ngopalak-redhat:
@nispriha @ngopalak-redhat @sigtrapd working on this

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,823 @@
...
+
+The W3C Baggage specification defines a standardised header format: `baggage: audit-id=abc123,other-key=value`. It is independent of any specific OTel SDK - any language or framework can read it with plain string parsing.
+
+This approach is chosen for the following reasons:
+
+- **No CRI protobuf change.** The `ExecRequest` message is unchanged. The audit ID travels via gRPC metadata (HTTP/2 headers), not a new protobuf field. This avoids a CRI API change and the coordination across CRI implementations that comes with it.
```
mikebrow: > avoids a CRI API change

sort of... it introduces a one/off env field injection to be done at the grpc client/server layer

> This avoids a CRI API change and the coordination across CRI implementations

still requires a CRI integration/implementation change... containerd ships in sync with kubernetes now.. so we should have less coordination entanglements. 

.. personally I prefer an optional set of extra envs passed from the kubelet on the exec api

mikebrow (commented):
hmm

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: @mikebrow We were about to swap them anyways. The additional `Env` field is the primary direction with baggage being an evaluated alternative. Thanks

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,818 @@
...
+
+###### How can an operator determine if the feature is in use by workloads?
+
+<!--
+Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+checking if there are objects with field X set) may be a last resort. Avoid
+logs or events for this purpose.
+-->
+
+###### How can someone using this feature know that it is working for their instance?
```
sohankunkerkar, line 425: Please answer this as part of the Alpha requirement for PRR

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,818 @@
...
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
```
sohankunkerkar, line 140: The audit ID is caller-settable, so the correlation key this KEP propagates to runtime agents is not tamper-proof. Worth documenting so consumers understand the trust boundary.

sohankunkerkar (commented):
I have two points from the PRR shadow side.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: Hi @sohankunkerkar 
The Audit ID is set by the API server. Even if the user were to provide a value for it, it would be overridden by the API server on its downstream call to the Kubelet. 

However, the tamper proof guarantees fail for the problem we are trying to solve at the runtime layer as the audit ID is injected as an environment variable. We have out of tree solutions using eBPF to overcome that part of the problem. You can read about it here - https://blog.sigtrapd.dev/posts/bridging-k8s-exec-identity-gap/

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: @nispriha  Can you please update update the README?
Users would be able to confirm that the feature is working successfully by checking for the AUDIT_ID (whatever the exact name is) environment variable inside the pod exec session.

nispriha: Done

keps/sig-node/6035-exec-session-identity/README.md
nispriha: I have updated the KEP to document this.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+Every layer of the chain is explicit and inspectable with this approach. It requires changes at every layer: a new CRI proto field (`repeated KeyValue envs` in `ExecRequest`), new Kubernetes API types (internal and versioned `PodExecOptions`), deepcopy/conversion/openapi regeneration, URL construction changes in the API server, query parameter parsing in the kubelet, and runtime updates in CRI-O and containerd.
+
+#### Implementation
+
+**CRI protobuf**
+
+Add `repeated KeyValue envs` to `ExecRequest` and `ExecSyncRequest` in `staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto`. This aligns exec with the existing `ContainerConfig.envs` pattern used for container creation.
+
+**API server**
```
SergeyKanzhelev: will this URL support ANY key/value and we will inject the AUDIT_ID specifically or there will be filtering somewhere that only allows AUDIT_ID for now? I suggest to support key/values from the beginning so we can discover override behaviors by different components

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+`pkg/registry/core/pod/rest/subresources.go:Connect` - read the audit ID from `audit.AuditContext(ctx).AuditID()` and add it as `?env=AUDIT_ID=<value>` to the kubelet request URL.
```
SergeyKanzhelev: is it ok to have Audit ID in logs? What about future key/values? I am thinking whether we should use headers instead so the values will not end up in logs by default.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+CRI-O and containerd each require a one-line change to pass the gRPC handler context through to `GetExec`, plus an interface change in the `cri-streaming` library.
+
+However, the contract - that the W3C Baggage key `audit-id` becomes the environment variable `AUDIT_ID` in the exec process - is implicit in the runtime's baggage-reading code rather than visible in the CRI API or Kubernetes API. The mechanism is specific to out-of-band metadata propagation; it does not generalise to arbitrary environment variable injection by API callers. It also introduces a one-off env field injection at the gRPC client/server layer rather than using a first-class CRI API field, and still requires CRI integration/implementation changes in each runtime.
```
SergeyKanzhelev: yes, this is fair. The propagation is there, but AUDIT_ID seems like a business logic rather than an ambient context.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: I was not planning to filter out the AUDIT_ID alone so that, even if some other use case wanted to utilise env variables, they would be able to.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: I feel that the audit id by itself is not sensitive. It is anyways present in the API server audit logs.

And the concern about future key value pairs containing secrets and they ending up in the logs is indeed something to think about. Which makes me think, if people were to really send secrets via this feature, should we also allow them to reference them from existing kubernetes secrets?

 If you think that might be a valid solution, do you think it can be another enhancement down the line or something you'd want to be included in the scope of this KEP? Because without this, the values would be exposed to logs inevitably either at the API server or the Kubelet

keps/sig-node/6035-exec-session-identity/kep.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - "@sigtrapd"
+  - "@ngopalak-redhat"
+  - "@nispriha"
+
+owning-sig: sig-node
+participating-sigs:
+  - sig-security
+  - sig-api-machinery
+
+status: provisional
```
soltysh: implementable?

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
```
soltysh: Nit: make sure to check appropriate boxes above.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+When a user runs `kubectl exec`, the API server authenticates the caller and records their identity in the audit log, but this identity is never forwarded to the kubelet or the container runtime. This makes it impossible for runtime security tooling to attribute in-container activity to the Kubernetes user who initiated the session. This KEP proposes passing the audit request ID from the API server through to the kubelet as part of the exec request flow, enabling runtime-level tooling to correlate process-level events with the originating user. Crucially, raw user identity data (such as usernames or tokens) will not be transmitted to the nodes to prevent potential credential theft by compromised or malicious workloads running on that node.
```
soltysh: You're mixing two things here:
1. is the user identity - think user name
2. is the audit ID - a UUID generated when the request first connects to kube-apiserver, which allows tracing the same connections as it goes through the layers in kube-apiserver. 

Which one do you want to use here? Be more explicit about that.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+This KEP proposes propagating identity information from the API server through the kubelet and into the CRI layer as part of the exec (and attach) request flow. Today, identity is present at the API server but is dropped before the request reaches the kubelet. The proposed change would carry the audit request ID through the entire chain, so that the container runtime has access to it.
```
soltysh: Again, you're confusing identity and audit ID, if you want both, say so explicitly. But don't use them interchangeably, since they are not the same.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+Every layer of the chain is explicit and inspectable with this approach. It requires changes at every layer: a new CRI proto field (`repeated KeyValue envs` in `ExecRequest`), new Kubernetes API types (internal and versioned `PodExecOptions`), deepcopy/conversion/openapi regeneration, URL construction changes in the API server, query parameter parsing in the kubelet, and runtime updates in CRI-O and containerd.
+
+#### Implementation
+
+**CRI protobuf**
+
+Add `repeated KeyValue envs` to `ExecRequest` and `ExecSyncRequest` in `staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto`. This aligns exec with the existing `ContainerConfig.envs` pattern used for container creation.
```
soltysh: This makes this KEP rely on https://github.com/kubernetes/enhancements/issues/6090 and that is not explicitly stated anywhere. Why?

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+##### Unit tests
+
+Unit tests will be added to verify that the API server injects the audit ID as an `env` query parameter on outbound exec requests, and that the kubelet forwards it to `ExecRequest.Env`.
+
+- `pkg/registry/core/pod/rest`: TBD
+
+##### Integration tests
+
+Not applicable for alpha. The propagation path crosses component boundaries (API server, kubelet, CRI runtime) which are better validated through e2e tests.
```
soltysh: This is only partially true. Our integration tests allow combining several components as required. They only don't run kubelet.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+##### Prerequisite testing updates
+
+None.
+
+##### Unit tests
+
+Unit tests will be added to verify that the API server injects the audit ID as an `env` query parameter on outbound exec requests, and that the kubelet forwards it to `ExecRequest.Env`.
+
+- `pkg/registry/core/pod/rest`: TBD
```
soltysh: I'd expect a list of proposed unit tests and a list of packages you'll be touching with their current coverage. See [template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md#unit-tests)

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+#### Alpha
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- No major bugs reported during alpha
+- Gather feedback from developers and surveys
```
soltysh: Whether this KEP will touch the CRI or not is still a question I don't know the answer to. So depending on the answer these requirements might change, requiring a 2 CRI implementations existing for beta promotion.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+    - kubelet
+
+###### Does enabling the feature change any default behavior?
+
+<!--
+Any change of default behavior may be surprising to users or break existing
+automations, so be extremely careful here.
+-->
+
+No, there is no change in the default behavior.
```
soltysh: No true, when you enable the FG, the new value is being sent to CRI, no?

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+No.
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+No. No existing API objects are modified in size. The audit ID is propagated via a query parameter on the exec proxy request and a new CRI protobuf field, not stored in any persistent API object.
```
soltysh: What about expanding `PodExecOptions` with the new `Env` array?

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+  - Details:
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+<!--
+Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+implementation difficulties, etc.).
+-->
+
+### Dependencies
```
soltysh, line 448: Since you're touching CRI, I'd expect this section be filled in sooner than later.

keps/sig-node/6035-exec-session-identity/README.md
SergeyKanzhelev: taking secrets values form k8s secrets objects make sense for the future improvements.

Here I am asking about "sensitive". How much precedence do we have with query arguments vs. headers?

I am ok to leave it as implementation detail during PR review. I generally have bad experience with urls and information disclosure.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,803 @@
...
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+This KEP proposes propagating identity information from the API server through the kubelet and into the CRI layer as part of the exec (and attach) request flow. Today, identity is present at the API server but is dropped before the request reaches the kubelet. The proposed change would carry the audit request ID through the entire chain, so that the container runtime has access to it.
+
+Once the ID reaches the container runtime, runtime security agents (such as Falco, Tetragon etc.) would be able to read it and correlate in-container activity with the API server audit log, closing the attribution gap.
```
haircommander: this is a little hand wavy here. how?

haircommander: https://github.com/kubernetes/enhancements/pull/6036/changes#diff-060e79068aa187e6f3b540df3006b6bf6e73e7602b2a5668579aa322354f2b88R308

keps/sig-node/6035-exec-session-identity/README.md
haircommander: yeah I think the two should be compressed

sigtrapd:
Hi @soltysh @SergeyKanzhelev @mikebrow 
The KEPs #6035 and #6090 Have been merged into one single KEP. I have addressed all the questions that were raised in the previous round of review. It would be great if you could review this once.
CC: @ngopalak-redhat @haircommander @nispriha

keps/sig-node/6035-exec-session-identity/kep.yaml
```diff
@@ -0,0 +1,62 @@
...
+kep-number: 6035
+authors:
+  - "@sigtrapd"
+  - "@ngopalak-redhat"
+  - "@nispriha"
+
+owning-sig: sig-node
+participating-sigs:
+  - sig-security
+  - sig-api-machinery
```
soltysh, line 11: Have you reached out to them for their acks on this KEP?

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+**API server** (`pkg/registry/core/pod/rest/subresources.go:Connect`)
+
+Read `audit.AuditContext(ctx).AuditID()` and append `?env=KUBERNETES_EXEC_AUDIT_ID=<value>` to the kubelet request URL before forwarding the exec request. If the audit context is absent or returns an empty ID (e.g., audit logging is disabled), inject a fixed placeholder value (e.g., `KUBERNETES_EXEC_AUDIT_ID=unknown`) so the variable is always present when the feature gate is enabled. Caller-supplied `KUBERNETES_EXEC_AUDIT_ID` entries in `PodExecOptions.Env` are rejected at validation time (see [Validation](#validation)).
+
+A Prometheus counter `apiserver_exec_audit_id_injected_total` is incremented on each exec request, with a `result` label distinguishing `injected` (real audit ID), `placeholder` (fallback value), and `skipped` (feature gate disabled). This metric is the primary signal for operators to confirm the feature is in use and to detect regressions.
```
soltysh, line 229: I'm not sure if that kind of metrics is valuable. It seems to narrow for a metric. But I'll leave it to apimachinery folks to decide.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+- [x] Metrics
+  - Metric name: `apiserver_exec_audit_id_injected_total{result="injected"}`
+  - Components exposing the metric: kube-apiserver
+- [x] Other
+  - Details: Run `kubectl exec <pod> -- env | grep KUBERNETES_EXEC_AUDIT_ID`. If the variable is present and contains a UUID, the feature is functioning correctly for that exec session. The UUID should match the `auditID` field in the API server audit log entry for that exec request.
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+When both feature gates are enabled, every exec request reaching the API server should result in either an `injected` or `placeholder` increment of `apiserver_exec_audit_id_injected_total`. The ratio of `placeholder` to `injected` should remain near zero in clusters with audit logging enabled; a sustained non-zero `placeholder` rate indicates audit context is missing and warrants investigation.
```
soltysh, line 435: Can you re-use existing metrics for running pods, eg. `kubelet_pod_start_sli_duration_seconds`?

soltysh (commented):
/approve
the PRR section for alpha

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+Solving the identity correlation gap requires addressing two distinct problems:
+
+- **The transport problem**: how does the audit ID travel from the API server to the container runtime through the Kubernetes stack? This KEP solves the transport problem.
+- **The attribution problem**: once it arrives, how do we ensure it cannot be tampered with by the caller? This is explicitly out of scope and deferred to follow-up work. See [Bridging the Kubernetes Exec Identity Gap](https://blog.sigtrapd.dev/posts/bridging-k8s-exec-identity-gap/) for background on out-of-tree approaches to this problem.
+
+### Goals
+
+- Add `repeated KeyValue envs` to `ExecRequest` and `ExecSyncRequest` in the CRI protobuf, establishing a clear runtime contract for environment variable injection at exec time.
+- Add `Env []EnvVar` to `PodExecOptions` as the Kubernetes API surface for this capability, allowing user agents (such as `kubectl exec --env`) to reliably inject environment variables into exec'd processes without depending on any binary or shell being present inside the container image.
```
haircommander, line 101: what is this going to be used for if apiserver needs to inject request id

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+- Propagate the API server audit request ID, not raw user identity, end-to-end as the first consumer of the CRI primitive.
+- Enable runtime security agents to correlate in-container process activity with the originating exec request, using the audit ID as a join key against the API server audit log.
+
+### Non-Goals
+
+- Solving the attribution problem: ensuring the propagated audit ID cannot be tampered with by the caller. Deferred to follow-up work (e.g., eBPF-based agents capturing the value at `execve` time into kernel-managed storage before any userspace code runs).
+- Identity propagation for lifecycle hooks and probes via `ExecSyncRequest`. The CRI field is added to `ExecSyncRequest` for API consistency, but no component populates it in this KEP. This is deferred as independent work.
```
haircommander: idk if we should add it if it won't be used. the two apis are different already

keps/sig-node/6035-exec-session-identity/README.md
haircommander: in fact, allowing a user to use this field opens the opportunity for them to attempt to clobber the one injected by apiserver

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+
+#### Story 3
+
+As a cluster operator, I want to inject environment variables into exec sessions without depending on any binary or shell being present in the container image. Today, the only option is to prepend `env KEY=VALUE` to the exec command, which fails silently on distroless images. With this KEP, `kubectl exec --env KEY=VALUE` provides a reliable mechanism to inject metadata at exec time without any in-container dependency.
+
+### Risks and Mitigations
+
+- **Environment variable tamperability.** Once injected into the exec'd process, userspace code inside the exec session can overwrite or unset the `KUBERNETES_EXEC_AUDIT_ID` value before a security agent reads it. Consumers should treat it as a correlation hint rather than a security assertion. Tamper-proofing via kernel-space eBPF capture is deferred to follow-up work.
```
haircommander, line 142: @ngopalak-redhat didn't you find using /proc/1/environ mitigated this because it's not updated during command execution or am i misremembering

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+```
+
+The contract established by this change: the runtime must inject all provided env vars into the exec'd process's environment. Values are injected literally; no expansion or interpolation is performed by the runtime, regardless of whether a value contains shell-style references such as `$VAR` or `${VAR}`. Callers that need dynamic values must resolve them before placing them in `ExecRequest.envs`.
+
+### Environment Variable Precedence
+
+The final environment of the exec'd process is constructed so that earlier entries take precedence on collision:
+
+```
+final_envp[] = runtime_envs + kubelet_envs + apiserver_envs + caller_envs + container_envs
```
haircommander, line 192: container_envs overriding nullifies this though right? also what about image envs?

haircommander: oh image envs are included there

haircommander: oh nevermind on this whole point, the precedence is not clear with this equation.

keps/sig-node/6035-exec-session-identity/README.md
```diff
@@ -0,0 +1,534 @@
...
+- **Maximum 256 entries per request.** Prevents unbounded URL/audit log growth.
+- **Maximum 32KB total size across all entries.** Prevents oversized requests that could exceed practical URL length limits or bloat audit logs.
+- **Reserved keys are rejected with a clear error.** `KUBERNETES_EXEC_AUDIT_ID` and any other reserved keys (defined in a constant list) cannot be supplied by callers; the API server returns a validation error such as `"KUBERNETES_EXEC_AUDIT_ID is reserved by the API server"`. This gives users immediate feedback rather than silently stripping the entry.
+
+### Silent Degradation
+
+If a runtime receives `ExecRequest.envs` populated but has not yet implemented forwarding, the field is silently ignored (standard protobuf zero-value behaviour for unrecognised fields). Exec sessions continue to work normally; the audit ID is simply not injected into the process environment. This is consistent with how other new CRI fields have been handled historically and is acceptable during the rollout period. Beta graduation explicitly requires at least two runtimes to have implemented the field.
```
haircommander: or you could use runtime features to branch on this in the kubelet, potentially denying exec requests that are run on a node that doesn't allow it?

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: Here are some use cases I thought about for exposing the env vars to users via `PodExecOptions` :
- exec'ing a specific binary in a distroless container with no shell or no env binary, while also having to inject some additional environment variables like LOG_LEVEL=DEBUG
- Injecting dynamic credentials for the exec'd process session based on the incoming user's identity.
- Can enforce custom runtime security policies based on the identity. For example, if a user is not allowed to read a certain file inside an exec session, you can inject the policy context into the session, from where eBPF tools can contextually and selectively block those calls for that user alone. (I feel that this can be very powerful)
- And of course, can be used to inject TraceID/ Audit ID for attribution of commands run inside


And what happens when the user inputs conflict the `KUBERNETES_EXEC_AUDIT_ID` env var?
As defined by the precedence equation below, API server injected environment variables take more precedence compared to user injected ones. This can be handled in two ways during the implementation:
- The user's value can be discarded and the original AUDIT_ID from the API server is sent down, or
- The API server can reject a exec request if it contains the `KUBERNETES_EXEC_AUDIT_ID` env var defined by the user with an error message saying it cannot be overridden.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: I wanted to introduce this feature for Pod lifecycle probes and hooks as well. But in that case, there wont be an AUDIT ID at all since it does not come from the API server. It would have to be something else coming from the kubelet. You would need the Env field in `ExecSyncRequest` then. Happy to remove this for now.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: Runtime features would give us a cleaner signal than relying on protobuf zero-value behaviour.

A two-stage approach could work: at alpha, the kubelet queries the runtime and logs a warning (or emits a metric) when ExecRequest.envs is unsupported, but still lets the exec proceed. At beta, we could add a kubelet flag for operators who want strict enforcement to deny exec on unsupported runtimes.

haircommander: there are two cases you need to worry about if it's on in the API but not in the kubelet:
- is it explicitly turned off in the kubelet
- is the kubelet too old to recognize it

Then there's the case where it's on in the kubelet, but not in the CRI.

If operators are to rely on this, we may want to fail in any case we can recognize if it's not supported by the entire stack

sigtrapd: In both the scenarios, it would fail silently but would not break anything since it is just going to be query params. 

If we want to identify incompatibility, then the API must identity the Kubelet version and not allow the feature if the kubelet does not support it. Similarly the kubelet will not accept entries in the env query param if it knows the CRI does not support it. Thoughts?

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: Have removed the addition.

keps/sig-node/6035-exec-session-identity/README.md
haircommander: I don't think apiserver recognizing kubelet version is the right move, we've had much more impactful features in the past that didn't do that. but other paths I think we can error to indicate to a cluster admin their envs won't propagate

sigtrapd: Agreed. I'm hoping we can address this better during the implementation.

keps/sig-node/6035-exec-session-identity/README.md
sigtrapd: https://github.com/kubernetes/enhancements/pull/6036/changes#diff-060e79068aa187e6f3b540df3006b6bf6e73e7602b2a5668579aa322354f2b88R122

I have slightly refined the text here to go into a little more detail without going into too much of it. I hope that makes it look a little more complete.

haircommander:
LGTM from node side

keps/sig-node/6035-exec-session-identity/README.md
mrunalp: If we do add this, it may be worth blocking certain prefixes entirely.

mrunalp (approved):
/approve

for node. Still need API machinery to review the proposed changes before final lgtm

mikebrow (approved):
/lgtm

mikebrow:
@mrunalp Appologies, I missed that there was no hold for the api server review, my lgtm was turned into an lgtm label by the bot, then it merged it. Will be more cautious around `/lgtm` responses from now on.

ngopalak-redhat:
@mikebrow Thanks, We'll reach out on slack and keep this updated for the API review.
cc: @mrunalp

ngopalak-redhat:
@mikebrow We have opened a PR to trigger the API review #6205

PR #6037, opened 2026-04-28
KEP 1645: add endpointSliceObjectsPresence field to ServiceImport

MrFreezeex:
- One-line PR description: add endpointSlicesAvailable field to ServiceImport

- Issue link: https://github.com/kubernetes/enhancements/issues/1645

- Other comments:

Add endpointSliceObjectsPresence to ServiceImport status so that consumer such as Gateway API implementations can decide if they should use EndpointSlice objects directly. Since surfacing EndpointSlices from remote clusters is optional without this fields consumers that wanted to directly reached the backends by reading the EndpointSlice objects had to fail in unclear ways or needed to have more complex probing logic to determine that info.

Cilium does not always sync back the EndpointSlice objects. We do that only if the ServiceImport is headless or if a certain annotation is present on the ServiceImport object (as the ServiceImport IPs load-balancing is not required
to have the EndpointSlice objects present in this implementation). Since EndpointSlice syncing is optional this is fully spec compliant. The addition of the `endpointSliceObjectsPresence` on ServiceImport would greatly benefit us if adopted by consumers, because  our users would probably see a clearer error that should led them to add the implementation specific annotation signaling that EndpointSlices should be synced.

Note that I decided to not propose a ServiceExport field to replace the Cilium annotation at this moment, since I am not sure that any other implementation have the same behavior (if at least one other implementation have something like that it too would probably be worth it though!).

### DIFF
--- a/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
+++ b/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
@@ -629,6 +629,11 @@ type ServicePort struct {
 
 // ServiceImportStatus describes derived state of an imported service.
 type ServiceImportStatus struct {
+  // EndpointSliceObjects indicates whether imported EndpointSlice objects are
+  // present for this ServiceImport.
+  // +kubebuilder:validation:Enum=Present;Absent
+  // +optional
+  EndpointSliceObjects EndpointSliceObjectsStatus `json:"endpointSliceObjects,omitempty"`
   // +optional
   // +patchStrategy=merge
   // +patchMergeKey=cluster
@@ -647,6 +652,20 @@ type ServiceImportStatus struct {
 type ClusterStatus struct {
  Cluster string `json:"cluster"`
 }
+
+// EndpointSliceObjectsStatus indicates whether imported EndpointSlice objects
+// are present for a ServiceImport.
+type EndpointSliceObjectsStatus string
+
+const (
+  // EndpointSliceObjectsPresent indicates that imported EndpointSlice objects
+  // are present for a ServiceImport.
+  EndpointSliceObjectsPresent EndpointSliceObjectsStatus = "Present"
+
+  // EndpointSliceObjectsAbsent indicates that imported EndpointSlice objects are
+  // absent for a ServiceImport.
+  EndpointSliceObjectsAbsent EndpointSliceObjectsStatus = "Absent"
+)
 ```
 
 ```yaml
@@ -667,6 +686,7 @@ spec:
     port: 80
   sessionAffinity: None
 status:
+  endpointSliceObjects: Present
   conditions:
   - type: Ready
     reason: Ready
@@ -901,6 +921,13 @@ conform to the specification in the following section.
 
 _Optional to create, but specification defined if present._
 
+Implementations must publish whether imported `EndpointSlice` objects are
+present through `ServiceImport.Status.EndpointSliceObjects`. API consumers
+that would normally always use `EndpointSlice` objects (such as Gateway API
+implementations) can use this to decide whether to consume `EndpointSlice`
+objects directly, degrade to `ServiceImport` IPs if they support doing so,
+or fail if `EndpointSlice` objects are required.
+
 If an implementation does create `discovery.k8s.io/v1 EndpointSlice`s, they must
 conform to the following structure. This structure was originally required as
 part of this specification in alpha, and are the structure on which other
@@ -951,6 +978,7 @@ spec:
     port: 80
   sessionAffinity: None
 status:
+  endpointSliceObjects: Present
   clusters:
   - cluster: us-west2-a-my-cluster
 ---


lauralorenz:
A few thoughts:
- On "must" fill out and whether the field is optional: Optional in the CRD makes upgrades easier, but what does the field being null mean? Is it the same as picking one of "present" or "absent"? If such a meaning is ascribed to null, then what is the default and what does that mean for implementations as they upgrade?
- On naming: Ideas "EndpointSliceObjectsPresence" -> 
  - "EndpointSliceSynced"
  - "EndpointSliceSupplied"
  - just "EndpointSliceObjects:[Present|Absent]"
  - or just "EndpointSlices:[Present|Absent]" 
    - but then it kinda feels like it should have all the endpointslices there directly
    - Would we ever actually include a ref to the EndpointSlices here directly instead? Probably not for performance reasons but bringing it up

MrFreezeex:
> * On "must" fill out and whether the field is optional: Optional in the CRD makes upgrades easier, but what does the field being null mean? Is it the same as picking one of "present" or "absent"? If such a meaning is ascribed to null, then what is the default and what does that mean for implementations as they upgrade?

Yeah the field being optional is fully for upgrade, if the field is not there yet it should only be because the implementation does not support that CRD version yet. As a consumer you can't yet expect anything in that case (you are in the same state as without that field). It might take a bit of time so that consumers can safely rely on that info (implementation should support it, users upgrade to a version of implementations supporting it etc).

The "must" is to say that you won't be passing core/required conformance test (whenever that will be added, meaning probably 0.6.0 at the earliest).

> On naming: Ideas [...]

I think I might prefer "EndpointSliceObjects:[Present|Absent]" among the one you suggested :eyes:. It presuppose less things and we might have an easier time extending that enum it in the future (not sure we will ever do but it still more future possibilities :man_shrugging:).

> or just "EndpointSlices:[Present|Absent]"
> but then it kinda feels like it should have all the endpointslices there directly

Yep we probably want more to convey to consumers the decision of the implementation that EndpointSlice should or should not be there at some point not necessarily that it's already there indeed!

> Would we ever actually include a ref to the EndpointSlices here directly instead? Probably not for performance reasons but bringing it up

This is indeed probably not necessary since there's already all the labels in the EndpointSlice objects which you can specifically watch for efficiently as a consumer/client and indeed it has some performance cost to maintain those. We probably want to do the same thing as the relation between Service and EndpointSlice and not introduce that IMO.

lauralorenz:
Was thinking that we could do a best-effort reach out to the [MCS implementations](https://multicluster.sigs.k8s.io/implementations/mcs-implementations/) notifying them of this PR and availability for comment/+1s. Since this is a non breaking change I think a maximum wait time of 2 weeks would be sufficient.

MrFreezeex:
> Was thinking that we could do a best-effort reach out to the [MCS implementations](https://multicluster.sigs.k8s.io/implementations/mcs-implementations/) notifying them of this PR and availability for comment/+1s. Since this is a non breaking change I think a maximum wait time of 2 weeks would be sufficient.

Hmm reaching out to implementation seems more something that we could do once the API is ready and released on kubernetes-sigs/mcs-api? There is no real guarantee that any implementations is actually conformant today without accounting this change (besides probably Submariner and Cilium that are known to execute the conformance suite).

Speaking of, we also would need to reach out most likely relatively with the conformance report program work from @AnupamGhosh once ready. We might be able to reach out with the announcement of the report program and encourage implementations to pick up from all the "recent" changes (including this one) at the same time.

MrFreezeex:
After some discussion in our weekly call, I will reach out to implementations to get their +1 early/before this is merged with a lazy consensus of 2 weeks if no feedbacks.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -629,6 +629,11 @@ type ServicePort struct {
 
 // ServiceImportStatus describes derived state of an imported service.
 type ServiceImportStatus struct {
+  // EndpointSliceObjects indicates whether imported EndpointSlice objects are
+  // present for this ServiceImport.
+  // +kubebuilder:validation:Enum=Present;Absent
+  // +optional
+  EndpointSliceObjects EndpointSliceObjectsStatus `json:"endpointSliceObjects,omitempty"`
```
tpantelis, line 636: What's the behavior if it's empty, seeing that only `Present` or `Absent` is valid but `omitempty` is allowed?

Would a simple boolean make more sense here?

MrFreezeex: This is to ease the CRD upgrade path but an empty field should not be valid in a conformant implementation (I made an answer with more details here fyi: https://github.com/kubernetes/enhancements/pull/6037#issuecomment-4345979589).

> Would a simple boolean make more sense here?

It originally started as a boolean but we changed it to a enum because it's a kubernetes API recommendation to favor enum over boolean for future extensivity (for instance I am not sure we would want to come to that but a possible future value could be `Partial`) and also because `Absent` and `Present` could be somewhat clearer than a boolean as to what it means.

MrFreezeex:
@tpantelis could you add your lgtm if it makes sense to you 🙏  (or let me know if you have further questions/comments/suggestions). It should be ready to be merged now that the 2 weeks delay is up!

PR #6039, opened 2026-04-29
KEP-5598: Extend opportunistic batching with rescoring

romanbaron:
- One-line PR description:
Add rescoring to handle multi-pod-per-node workloads: when the last chosen node remains feasible, rescore it in-place and continue batching rather than flushing the cache.

- Issue link: https://github.com/kubernetes/kubernetes/issues/137707

- Other comments:
AI tooling was used to assist in preparing this PR. All changes have been reviewed and verified by the author.

### DIFF
--- a/keps/sig-scheduling/5598-opportunistic-batching/README.md
+++ b/keps/sig-scheduling/5598-opportunistic-batching/README.md
@@ -94,12 +94,16 @@ tags, and then generate with `hack/update-toc.sh`.
     - [GetNodeHint](#getnodehint)
     - [StoreScheduleResults](#storescheduleresults)
     - [Integration with Scheduling Cycle](#integration-with-scheduling-cycle)
+  - [Rescoring](#rescoring)
+    - [Rescoring the Last Chosen Node](#rescoring-the-last-chosen-node)
+    - [NormalizeScore on a Subset](#normalizescore-on-a-subset)
   - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
   - [Risks and Mitigations](#risks-and-mitigations)
     - [Plugins need to keep signatures up to date](#plugins-need-to-keep-signatures-up-to-date)
     - [Limited workload coverage](#limited-workload-coverage)
     - [No prior production history](#no-prior-production-history)
     - [Cached nodes are not re-filtered each cycle](#cached-nodes-are-not-re-filtered-each-cycle)
+    - [Memory overhead from raw score caching](#memory-overhead-from-raw-score-caching)
 - [Design Details](#design-details)
   - [Pod signature](#pod-signature)
   - [Test Plan](#test-plan)
@@ -123,6 +127,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
   - [Comparison with Equivalence Cache (circa 2018)](#comparison-with-equivalence-cache-circa-2018)
+  - [New Rescore Extension Point](#new-rescore-extension-point)
 - [Future work](#future-work)
 - [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
 <!-- /toc -->
@@ -181,6 +186,9 @@ scheduling. To implement this mechanism we propose the following additions:
    hints for multiple subsequent pods with matching scheduling signatures.
  - **Opportunistic batching:** Transparent inclusion of the batching mechanism in the scheduler to
    improve the performance of targeted workloads that could benefit from it.
+ - **Rescoring:** An extension to the batching mechanism that removes the one-pod-per-node
+   limitation: when the last chosen node is still feasible, it is rescored in-place rather than
+   flushing the batch.
 
 ## Motivation
 
@@ -198,6 +206,27 @@ complement of daemonset pods. For these workloads, once a pod is placed on a nod
 and can be skipped for subsequent pods. This allows the scheduler to reuse not only filtering but
 also scoring results, reducing per-cycle cost.
 
+However, not all workloads follow the one-pod-per-node pattern. For multi-pod-per-node pods (e.g., a
+web server requesting 100m CPU / 128Mi memory) the previously chosen node remains feasible after
+each placement, so naively checking for infeasibility would invalidate the cache on every cycle.
+[kubernetes/kubernetes#137707] identifies three consequences if this case is unhandled:
+
+1. **Redundant synchronous overhead.** The `RunFilterPlugins` call in `batchStateCompatible`
+   executes on the critical scheduling path for every pod in a multi-pod-per-node batch without ever
+   producing a useful result.
+2. **Duplicate filtering.** The same node is filtered in `GetNodeHint` and then again in the
+   filtering pass that immediately follows after batch state is invalidated.
+3. **Cache invalidation every cycle.** The sorted node list is discarded on every cycle despite the
+   pod signatures being identical, so the O(N) `Filter` and `Score` pipeline runs in full for every
+   pod in the batch.
+
+Rescoring solves this: when the last chosen node is still feasible, the scheduler rescores it
+against the current pod and updates its rank in the cached list rather than discarding the batch
+state. Both workload patterns, one-pod-per-node and multi-pod-per-node, are handled efficiently
+within the same mechanism.
+
+[kubernetes/kubernetes#137707]: https://github.com/kubernetes/kubernetes/issues/137707
+
 ### Goals
 
  * Improve the performance of scheduling large jobs on large clusters where the constraints are
@@ -205,7 +234,7 @@ also scoring results, reducing per-cycle cost.
  * Begin building infrastructure to support gang scheduling and other "multi-pod" scheduling
    mechanisms.
  * Ensure that the infrastructure we build is maintainable as we update, add and remove plugins.
- * Reduce per-cycle scheduling cost for a targeted set of workloads in this release.
+ * Reduce per-cycle scheduling cost for both one-pod-per-node and multi-pod-per-node workloads.
  * Provide a path where we can expand batching to apply to most or all workloads over the next few
    releases.
  * Allow users to continue to use out-of-tree plugins. For this KEP we need to ensure that out-of-
@@ -219,14 +248,18 @@ also scoring results, reducing per-cycle cost.
  * Adding gang scheduling. This is purely a performance improvement without dependency on the
    Workload API [KEP-4671](https://github.com/kubernetes/enhancements/pull/5558), although this work
    is intended to build toward it.
+ * Supporting group-aware scoring plugins (`PodTopologySpread`, `InterPodAffinity`) for rescoring.
+   These cause pods to be unsignable and are excluded from batching entirely.
+ * Introducing a new `Rescore` plugin interface. That extension point is deferred to future work and
+   described under [Alternatives](#alternatives).
 
 ## Proposal
 
 The batching mechanism is applied transparently to simple cases in the scheduling cycle, reusing the
 cached sorted node list from a previous cycle to provide node hints for subsequent identical pods.
-More complex integrations are left to future work. 
+More complex integrations are left to future work.
 
-The proposal covers two components: pod scheduling signature and batching mechanism.
+The proposal covers three components: pod scheduling signature, batching mechanism and rescoring.
 
 ### Pod scheduling signature
 
@@ -303,10 +336,12 @@ current pod by checking:
 2. **Signature match:** The pod's signature must exactly match the cached signature.
 3. **Cache freshness:** The cached data must be sufficiently recent to avoid relying on stale
    scheduling decisions.
-4. **Last chosen node feasibility check:** The node chosen in the previous scheduling cycle must now 
-   be infeasible for the new pod. This is verified by running filter plugins against that node. This 
-   validation ensures the one-pod-per-node constraint that allows us to reuse scoring results without 
-   rescoring.
+4. **Last chosen node feasibility check:** Filter plugins are run against the node chosen in the
+   previous cycle. Two outcomes are possible:
+   - **Infeasible:** The node is full (one-pod-per-node case). The node is discarded and batch
+     proceeds without it.
+   - **Still feasible:** The node can host another pod (multi-pod-per-node case). This node is
+     rescored (see [Rescoring the Last Chosen Node](#rescoring-the-last-chosen-node)).
 
 If all checks pass, the operation pops the next best node from the cached sorted list and returns it
 as a hint (see [Integration with Scheduling Cycle](#integration-with-scheduling-cycle) below for how
@@ -356,6 +391,54 @@ nodes. This ensures correctness while providing significant performance benefits
 valid. The batching mechanism can be used in multiple places, including future gang scheduling
 implementations, without requiring changes to the Pod API.
 
+### Rescoring
+
+#### Rescoring the Last Chosen Node
+
+When `RunFilterPlugins` finds the last chosen node (node A) still feasible, rescoring runs instead
+of flushing the batch:
+
+1. Node A is re-inserted into the cached sorted list. Since it is still feasible, it can host
+   additional pods.
+2. `Score` is called for each scoring plugin against node A using the current `CycleState`. The
+   resulting score accurately reflects pod N's placement.
+3. `NormalizeScore` is called for each plugin over all nodes in the cached list (the re-inserted
+   node A plus the remaining nodes).
+4. The cached list is re-sorted by aggregate normalized score.
+5. The top-ranked node is popped and returned as the hint for the current pod.
+
+This is correct under the assumption that scoring is node-local: placing pod N on node A does not
+affect the scores of other nodes. This holds for all signable pods. Group-aware scoring plugins
+(which would violate it) already make pods unsignable and are excluded from batching entirely.
+
+**Cost comparison:**
+
+| Scenario | Batching disabled | Batching enabled |
+|---|---|---|
+| One-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter, 0 scoring |
+| Multi-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter + O(1) Score + O(M) NormalizeScore |
+
+Where M is the number of nodes in the cached list (M ≤ N). For multi-pod-per-node workloads M stays
+close to N since nodes are re-inserted each cycle. However, both the O(N) Filter pass and the O(N)
+Score pass are eliminated: only a single `RunFilterPlugins` call and a single `Score` call run per
+cycle. `NormalizeScore` over M cached nodes is the only remaining cost, and it is lightweight
+relative to `Score`: it is arithmetic normalization rather than plugin logic.
+
+#### NormalizeScore on a Subset
+
+To support rescoring, `batchState` also caches the raw score for each node in the cached list. When
+rescoring, only the rescored node's raw score is updated; the rest remain valid under the node-local
+scoring assumption.
+
+`NormalizeScore` is applied to the cached node subset rather than all feasible nodes. When the cache
+is fresh, meaning the cached list contains exactly the nodes a fresh full pipeline would filter to,
+rescoring node A and normalizing all cached nodes produces identical results to a full pipeline run.
+The raw scores of all other nodes are unchanged (node-local scoring), node A's raw score is updated
+by the explicit `Score` call, and `NormalizeScore` sees the same set with the same scores.
+
+The only source of divergence is cache staleness: if cluster state has changed since the initial
+full pipeline run, the cached raw scores for non-rescored nodes may be slightly out of date. This is
+bounded by the 500ms `maxBatchAge` expiry (see [Risks and Mitigations](#risks-and-mitigations)).
 
 ### Notes/Constraints/Caveats (Optional)
 
@@ -396,6 +479,19 @@ than 500ms is flushed and a full pipeline reruns. In a typical scheduling burst,
 processed every few milliseconds, the window for meaningful node state drift is very small.
 Operators can disable `OpportunisticBatching` to restore full-pipeline behavior if degradation is
 observed.
+
+#### Memory overhead from raw score caching
+
+`batchState` stores raw per-plugin scores for every node in the cached list to support rescoring.
+The overhead is:
+
+  `O(|cached list| × |scoring plugins|) × sizeof(int64)`
+
+For a cluster with 5000 nodes and 10 scoring plugins, this is on the order of a few hundred
+kilobytes per active batch. The cached list also shrinks over the lifetime of a batch as nodes are
+popped and not re-inserted (one-pod-per-node case). Each scheduler profile maintains its own
+`batchState`, so the total overhead scales with the number of active profiles - typically one or two
+in practice.
 
 ## Design Details
 
@@ -516,6 +612,14 @@ Will add an extra function and test for plugins we touch.
 - `batching_test.go` - Test cases for the batching mechanism, separate from the actual integration
   into the scheduling pipeline.
 
+Rescoring-specific unit tests (will be added to `batching_test.go` and `schedule_one_test.go`):
+- Raw score updated only for the rescored node; other nodes' cached scores unchanged.
+- `NormalizeScore` called once per plugin with the full cached node list (including re-inserted
+  `lastChosenNode`).
+- Returned hint is the node with the highest aggregate normalized score.
+- Sorted list after the hint pop contains the correct remaining nodes.
+- Raw scores are correctly populated by `StoreScheduleResults` after a full pipeline run.
+
 ##### Integration tests
 
 <!--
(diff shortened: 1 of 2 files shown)


keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -295,53 +318,98 @@ During the scheduling cycle, if a node was hinted by the batching mechanism, the
...
+
+#### NormalizeScore on a Subset
+
+To support rescoring, `batchState` also caches the raw score for each node in the cached list. When rescoring, only the rescored node's raw score is updated; the rest remain valid under the node-local scoring assumption.
+
+`NormalizeScore` is applied to the cached node subset rather than all feasible nodes. When the cache is fresh, meaning the cached list contains exactly the nodes a fresh full pipeline would filter to, rescoring node A and normalizing all cached nodes produces identical results to a full pipeline run. The raw scores of all other nodes are unchanged (node-local scoring), node A's raw score is updated by the explicit `Score` call, and `NormalizeScore` sees the same set with the same scores.
+
```
singh1203, line 439: I'm curious about this part: the raw score for the other nodes stays unchanged, but the normalized scores can still shift when lastChosenNode is rescored because NormalizeScore is applied across the cached node list. Is the correctness argument that this still matches the fresh, full-pipeline result as long as the cached set is complete and fresh?

singh1203 (approved):
Hey @romanbaron, thanks for including me. I went through the full diff carefully. Overall, the rescoring design is clean and well written. I just had one question, and thanks for answering it. 🙇
Overall, it looks good to me.

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: Yes, `NormalizeScore` scales everything relative to the min/max values, so if lastChosenNode's raw score changes, normalized scores can shift. It is fine as long as the cache is complete and fresh.

keps/sig-scheduling/5598-opportunistic-batching/kep.yaml
```diff
@@ -36,9 +36,14 @@ feature-gates:
   - name: OpportunisticBatching
     components:
       - kube-scheduler
+  - name: OpportunisticBatchingRescoring
+    components:
+      - kube-scheduler
```
macsko: We can use the same feature gate as for OpportunisticBatching. We don't want to graduate these features differently, and to preserve two code paths

keps/sig-scheduling/5598-opportunistic-batching/kep.yaml
macsko:
```yaml
latest-milestone: "v1.37"

milestone:
  beta: "v1.35"
  stable: "v1.38"
```

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -176,42 +182,57 @@ This KEP proposes an opportunistic batching mechanism in the scheduler to improv
...
+ * Applying this optimization to all pods. Batching is transparent but limited to signable workloads in this KEP.
+ * Adding gang scheduling. This is purely a performance improvement without dependency on the Workload API [KEP-4671](https://github.com/kubernetes/enhancements/pull/5558), although this work is intended to build toward it.
+ * Supporting group-aware scoring plugins (`PodTopologySpread`, `InterPodAffinity`) for rescoring. These cause pods to be unsignable and are excluded from batching entirely.
+ * Introducing a new `Rescore` plugin interface. That extension point is deferred to future work and described under [Alternatives](#alternatives).
 
 ## Proposal
 
-We discuss each of the added items: pod scheduling signature, batching mechanism and opportunistic batching in turn.
+The batching mechanism is applied transparently to simple cases in the scheduling cycle, sharing batch state (including the cluster snapshot and plugin CycleState) across multiple pods. More complex integrations, are left to future work.
```
macsko: I think both the cluster snapshot and cycle state are not shared between the cycles

keps/sig-scheduling/5598-opportunistic-batching/README.md
macsko, line 1: In this PR you propose both KEP refactoring (which is highly valuable) and describing the new restoring behavior. It would be easier to review this if you open a separate KEP just with the refactoring. It could be merged quickly and leave this PR with the most relevant change, which is restoring.

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -176,42 +182,57 @@ This KEP proposes an opportunistic batching mechanism in the scheduler to improv
...
 
-Another change is the shift towards 1-pod-per-node in batch and ML environments. Many of these environments (among others) only attempt to run a single user pod on each node, along with a complement of daemon set pods. This simplifies our scheduling needs significantly, as it allows to reuse not only filtering, but also scoring results.
+A common pattern in batch and ML environments is to run a single user pod on each node alongside a complement of daemonset pods. For these workloads, once a pod is placed on a node, that node is full and can be skipped for subsequent pods. This allows the scheduler to reuse not only filtering but also scoring results, reducing per-cycle cost to O(1).
+
+However, not all workloads follow the one-pod-per-node pattern. For multi-pod-per-node pods (e.g., a web server requesting 100m CPU / 128Mi memory) the previously chosen node remains feasible after each placement, so naively checking for infeasibility would invalidate the cache on every cycle. [kubernetes/kubernetes#137707] identifies three consequences if this case is unhandled:
```
macsko: nit: It would be also good to split the long lines of text into shorter ones. It simplifies the review process and further changes. I know the KEP was written without such line breaks, but let's keep the newer changes following this rule.

keps/sig-scheduling/5598-opportunistic-batching/README.md
singh1203: Makes sense. Thank you for clarifying here.

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: @macsko - I can definitely do that, but it will be easier if we split it into several PRs. 
Let's start with line wrapping: https://github.com/kubernetes/enhancements/pull/6049

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: I actually meant “between pods,” but I see now that this is confusing. I will rephrase the sentence to say that the cached list of nodes is reused to provide a node hint.

keps/sig-scheduling/5598-opportunistic-batching/README.md
macsko: Good idea, thanks

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: Sure thing.
The following PR includes the refactoring part: https://github.com/kubernetes/enhancements/pull/6057
After PR above is merged it should be much easier to review the rescoring related changes in this PR.

romanbaron:
The PR is now rebased on top of the refactored KEP and only includes Rescore-related changes.

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -833,6 +946,9 @@ that might indicate a serious problem?
   longer signable.
 - `scheduler_batch_cache_flushed_total` - unexpected spikes indicate the cache is being invalidated
   more than expected.
+- `scheduler_batch_rescore_attempts_total` - zero value for a multi-pod-per-node workload 
+  means those pods are not being rescored and the full pipeline runs for every cycle.
+- `scheduler_batch_rescore_duration_seconds` - unexpected latency in the rescore path.
```
utam0k, line 962: Should we also measure memory overhead using scheduler_perf?

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -356,6 +391,54 @@ nodes. This ensures correctness while providing significant performance benefits
...
+
+#### Rescoring the Last Chosen Node
+
+When `RunFilterPlugins` finds the last chosen node (node A) still feasible, rescoring runs instead
+of flushing the batch:
+
+1. Node A is re-inserted into the cached sorted list. Since it is still feasible, it can host
+   additional pods.
+2. `Score` is called for each scoring plugin against node A using the current `CycleState`. The
+   resulting score accurately reflects pod N's placement.
```
utam0k, line 405: I believe that node states can change over time. For example, I think the following cases could occur:

* Changes in ImageLocality scores due to image GC
* Taints are being added to a node
* Pods terminating

It is possible that these events occur on nodes other than Node A, causing their scores to change. Ideally, scoring should be performed on all nodes, but doing so would negate the benefits of this change.
Is this acceptable?

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -356,6 +391,54 @@ nodes. This ensures correctness while providing significant performance benefits
...
+This is correct under the assumption that scoring is node-local: placing pod N on node A does not
+affect the scores of other nodes. This holds for all signable pods. Group-aware scoring plugins
+(which would violate it) already make pods unsignable and are excluded from batching entirely.
+
+**Cost comparison:**
+
+| Scenario | Batching disabled | Batching enabled |
+|---|---|---|
+| One-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter, 0 scoring |
+| Multi-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter + O(1) Score + O(M) NormalizeScore |
```
utam0k: The cost table omits the cost of step 4 (re-sorting the cached list). A naive re-sort is `O(M log M)`. Worth surfacing in the table

```suggestion
| Multi-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter + O(1) Score + O(M) NormalizeScore + O(M log M) sort |
```

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: To spot cases where cached objects unexpectedly increase scheduler memory usage?

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: I agree with your observation, and I think this can affect not only scoring but also filtering. A taint applied to a node might make that node infeasible for the next pod. That said, this concern also applies to the current design, since cached nodes do not go through filtering on every cycle.

Cache staleness is bounded by `maxBatchAge` (500ms). This is explained in more detail in the risks and mitigations section [here](https://github.com/romanbaron/enhancements/blob/opportunistic-batching-rescore/keps/sig-scheduling/5598-opportunistic-batching/README.md?plain=1#L471).

In theory, the scheduler snapshot could also have been taken before such a taint was applied, so the same issue can occur without Opportunistic Batching as well, although the risk is probably much lower.

I think this is important to document this concern, but I don’t see how it can be eliminated without removing the Opportunistic Batching optimization entirely.

macsko: Still, the scheduler is only eventually consistent and it can observe any cluster change with some delay. Moreover, kube-scheduler doesn't pick the fully optimal score, because it subsample number of feasible nodes to score. Given than and the small age of batching information, this proposal is fine.

utam0k: I posted my comment because I felt this was a potential concern. If the decision has been made to accept the risks, then I think this policy is fine. Thank you.

keps/sig-scheduling/5598-opportunistic-batching/README.md
utam0k: This concern stems from "## Memory overhead from raw score caching."

romanbaron: ok, I added it to rollback signals and to integration tests.

keps/sig-scheduling/5598-opportunistic-batching/README.md
macsko: @romanbaron You can consider adding this concern to the Risks and Mitigations section

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -569,6 +680,7 @@ If e2e tests are not necessary or useful, explain why.
 - Run the existing scheduling e2e tests with batching enabled and disabled, to ensure they pass in
   both cases.
 - Add e2e tests ensuring that pod configurations we expect to be batched are in fact batched.
+- Add an e2e test with a multi-pod-per-node workload that asserts that rescoring is triggered.
```
macsko: I'm not sure whether we need any e2e test here. It can be well-modeled using integration tests. Do we even have any e2e tests for batching (I see we have some defined above, but I'm not sure we have them coded)?

macsko (commented):
The KEP looks good overall, thanks for doing this!

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: It is already [there](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5598-opportunistic-batching/README.md?plain=1#L388)

macsko: It's a bit different. The current risk mentions a case where the node becomes infeasible between the cycles, wasting the batched results. Here we have the risk, where the node is still feasible, but its score has changed, so that it's not correctly comparable with the actual node, because their scores base on two different cluster states. The mitigation is almost the same for both cases - small max batch age time.

utam0k:
I've read through it, and it looks good. This is a great update! Thanks!

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: That’s a good question. Besides unit tests, this is the only place I found:
https://github.com/kubernetes/kubernetes/blob/master/test/integration/scheduler/batch/batch_test.go 

It made sense to me that if there’s an e2e test for batching, it should also cover rescoring.

Both the one-pod-per-node and multi-pod-per-node scenarios are mentioned in the integration tests section. Should I remove the last two points from here?

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: Got, thanks! I've added it to the risks.

keps/sig-scheduling/5598-opportunistic-batching/README.md
romanbaron: @macsko - any thoughts about it?

macsko: Yeah, I think you can remove the last two points. We can cover them on a lower level

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -356,6 +392,54 @@ nodes. This ensures correctness while providing significant performance benefits
...
+### Rescoring
+
+#### Rescoring the Last Chosen Node
+
+When `RunFilterPlugins` finds the last chosen node (node A) still feasible, rescoring runs instead
+of flushing the batch:
+
+1. Node A is re-inserted into the cached sorted list. Since it is still feasible, it can host
+   additional pods.
+2. `Score` is called for each scoring plugin against node A using the current `CycleState`. The
```
dom4ha, line 404: I think we could cache the cycleState as well to avoid rerunning PreFilter. All we'd need to do is to run `PreFilterExtension.AddPod` on a pod assumed in the previous cycle.

dom4ha (commented):
The approach with rescoring LGTM. I think we could optimize it further with caching cycleState as well, but I'd treat it as optional implementation detail, not as the core element of the solution, since it assumes all plugins implement it.

Leaving the final approval on @sanposhiho

keps/sig-scheduling/5598-opportunistic-batching/README.md
macsko: I think not all plugins that have the `PreFilter ` implement the `PreFilterExtension.AddPod`. AddPod is expected to be used in the preemption, so plugins that know the preemption doesn't matter for them, can skip defining AddPod/RemovePod. At some point we could try to cache the cycleState where possible, but I wouldn't consider it for now

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -303,10 +337,12 @@ current pod by checking:
 2. **Signature match:** The pod's signature must exactly match the cached signature.
 3. **Cache freshness:** The cached data must be sufficiently recent to avoid relying on stale
    scheduling decisions.
-4. **Last chosen node feasibility check:** The node chosen in the previous scheduling cycle must now 
-   be infeasible for the new pod. This is verified by running filter plugins against that node. This 
-   validation ensures the one-pod-per-node constraint that allows us to reuse scoring results without 
-   rescoring.
+4. **Last chosen node feasibility check:** Filter plugins are run against the node chosen in the
+   previous cycle. Two outcomes are possible:
+   - **Infeasible:** The node is full (one-pod-per-node case). The node is discarded and batch
```
wojtek-t, line 342: nit: it doesn't necessary mean the node is full or really be pod-per-node - it may e.g. be result of using node-port where scheduling more pods *from our batch* is impossible

wojtek-t (commented):
The PRR part (unsurprisingly) still looks fine.

keps/sig-scheduling/5598-opportunistic-batching/README.md
dom4ha: Yeah, I already mentioned it as optional, so it's not a blocking comment for this KEP.

Anyway, we could make `PreFilterExtension` implementation either required in Signature or at least detect whether it's implemented before we can use it.

It should give the Opportunistic Batching a significant boost as PreFilter scales with a cluster size, but we treat it immutable anyway within opportunistic batching cache. So I think it's worth to consider it not in this KEP update, but in the future ones.

keps/sig-scheduling/5598-opportunistic-batching/README.md
sanposhiho, line 647: As we make this feature richer and richer, the overhead of the calculation is getting bigger.
We must make sure there's no degradation on scenarios where _various_ pods are being evaluated, otherwise the proposed change will make the perf worse in scenarios the opportunistic batching is not working and this feature is simply a perf overhead. (like highlighted [here](https://github.com/kubernetes/kubernetes/issues/137707))
Many scheduler-perf scenarios are using the same spec pods in a scenario, which cannot catch how big the degradation is.

...or, an alternative idea (as I raised in the issue) is enabling the opportunistic batching only in WAS scheduling cycle.

sanposhiho: btw, this "overhead for scenarios where the opportunistic batching doesn't work" should be mentioned on `Risks and Mitigations`

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -356,6 +392,54 @@ nodes. This ensures correctness while providing significant performance benefits
...
+
+| Scenario | Batching disabled | Batching enabled |
+|---|---|---|
+| One-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter, 0 scoring |
+| Multi-pod-per-node | O(N) filter + O(N) Score + O(N) NormalizeScore | O(1) filter + O(1) Score + O(M) NormalizeScore + O(M log M) sort |
+
+Where M is the number of nodes in the cached list (M ≤ N). For multi-pod-per-node workloads M stays
+close to N since nodes are re-inserted each cycle. However, both the O(N) Filter pass and the O(N)
+Score pass are eliminated: only a single `RunFilterPlugins` call and a single `Score` call run per
+cycle. `NormalizeScore` over M cached nodes is the only remaining cost, and it is lightweight
```
sanposhiho, line 425: > it is lightweight

...and we must make sure it is lightweight in scheduler-perf.
One reason I originally proposed a new `Rescore` ext point ([here](https://github.com/kubernetes/kubernetes/issues/137707#issuecomment-4067371239)) is actually to reduce the cost of rerunning NormalizeScore for all nodes. (The other reason was to support cross-node scoring, but that reason can be ignored for now)

sanposhiho: this cost comparison is only talking about the scenarios where opportunistic batching is working. if the next pod is not the same spec pod, all of those would simply be an overhead, as the issue mentioned.
https://github.com/kubernetes/kubernetes/issues/137707

keps/sig-scheduling/5598-opportunistic-batching/README.md
```diff
@@ -397,6 +481,29 @@ processed every few milliseconds, the window for meaningful node state drift is
...
+
+For a cluster with 5000 nodes and 10 scoring plugins, this is on the order of a few hundred
+kilobytes per active batch. The cached list also shrinks over the lifetime of a batch as nodes are
+popped and not re-inserted (one-pod-per-node case). Each scheduler profile maintains its own
+`batchState`, so the total overhead scales with the number of active profiles - typically one or two
+in practice.
+
 ## Design Details
 
 ### Pod signature
```
sanposhiho, line 509: we must make sure unsignable is returned when cross-node scoring scheduling is used on pods (preferred pod affinity, preferred topology spread, etc)

sanposhiho (commented):
~~perf is the biggest concern - we already get a report from user, but this change will increase the overhead of this feature more, without mentioning any solution about the overhead problem~~

keps/sig-scheduling/5598-opportunistic-batching/README.md
dom4ha, line 404: The way I read it is that we recompute PreFilter state as in each new cycle. I had a proposal to reuse (cache) the cycleState of the previous cycle in the comment https://github.com/kubernetes/enhancements/pull/6039#discussion_r3426669527, but only as a future extension

sanposhiho, line 404: > `Score` is called for each scoring plugin against node A using the current `CycleState`.

But, CycleState doesn't have any state calculated at PreScore at this point?

keps/sig-scheduling/5598-opportunistic-batching/README.md
dom4ha: Isn't the issue actually addressed by this update? The complain was about the case in which batch job was still able to schedule on the same node and this update is actually opening a fast path.

keps/sig-scheduling/5598-opportunistic-batching/README.md
dom4ha: Isn't computed only once we made sure the next pod has the correct signature?

keps/sig-scheduling/5598-opportunistic-batching/README.md
sanposhiho: Yes, we can resolve those my perf related threads

keps/sig-scheduling/5598-opportunistic-batching/README.md
sanposhiho: We discussed internally, and we agreed that we will run PreScore of all plugins before this. This is not the optimized way, but we can consider a further optimization in the next release cycle

/cc @dom4ha @macsko

macsko: > PreFilter

PreScore?

sanposhiho:
/approve
/lgtm

@romanbaron is on vacation and cannot modify the KEP for this comment-

https://github.com/kubernetes/enhancements/pull/6039#discussion_r3427586375

Approving the PR because all the leads are on the same page. 
We will ask him to edit it based on the agreement once he's back.

keps/sig-scheduling/5598-opportunistic-batching/README.md
sanposhiho: A bit of more summery of our internal discussion as a record (not only about this prescore stuff)

Running PreScore of all plugins is not very cheap, and we definitely need to discuss how we can make it efficient in the next release cycle. Very likely, we need to reuse the cyclestate from the last pod somehow.

Also, another discussion was whether we want to rerun Score() for all cached nodes. The current KEP says we just run Score() for the node that the last pod selected. But, that means we don't support the support of cross-node scoring. The current opportunistic batching doesn't support the default scheduling config because the default scheduling config has a preferred topology spread, which is a cross-node scoring. So, actually supporting such cross-node scoring is important to make this feature more widely usable.

We have two options to discuss in the next release cycle-
1. If we drop the support of cross-node scoring for now, we can implement `PreScoreExtension.AddPod` to modify the cyclestate of the last pod. Then, the flow will be PreScoreExtension.AddPod(last pod) -> Score(selected node) -> NormalizeScore(all cached nodes)
2. If we want to support cross-node scoring, there are two options further 
  a.  PreScoreExtension.AddPod(last pod) -> Score(all cached nodes) -> NormalizeScore(all cached nodes)
  b.  implement a new ext point `Rescore()`, which will do (a) essentially but more efficiently.

We don't decide it now, but will discuss in the next release KEP cycle

Issue #6038, opened 2026-04-29
HTTPS (TLS) for kube-proxy admin endpoints (metrics & healthz)


yashsingh74:
### Enhancement Description

- One-line enhancement description (can be used as a release note): **Optional TLS (HTTPS) for kube-proxy metrics and healthz admin listeners instead of cleartext HTTP only.**
- Kubernetes Enhancement Proposal: 
- Discussion Link: 
  - [ ] Alpha - v1.xx
  - [ ] KEP (`k/enhancements`) update PR(s): _(link PR here once opened)_
  - [ ] Code (`k/k`) update PR(s): _(none yet)_
  - [ ] Docs (`k/website`) update PR(s): _(none yet)_

**Authors:** Yash Kumar Singh, Daman Arora  

**SIG:** sig-network (participating: sig-instrumentation for metrics scraping docs; sig-auth if client auth scope grows)

**Scope (short):** TLS for kube-proxy’s existing HTTP servers: `MetricsBindAddress` (e.g. `/metrics`, mux-mounted diagnostics) and `HealthzBindAddress` (`/healthz`, `/livez`). Does **not** cover Service/NodePort dataplane encryption.

**Motivation:** Compliance and defense-in-depth when these ports are not loopback-only; alignment with TLS patterns used elsewhere in the project.

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #6041, opened 2026-04-29
KEP-6038: HTTPS (TLS) for kube-proxy admin endpoints

yashsingh74:
- One-line PR description: HTTPS (TLS) for kube-proxy admin endpoints

- Issue link: https://github.com/kubernetes/enhancements/issues/6038

- Other comments:

### DIFF
--- a/keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
+++ b/keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
@@ -1,0 +1,273 @@
+# KEP-6038: HTTPS (TLS) for kube-proxy admin endpoints
+
+## Table of Contents
+
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [User Stories](#user-stories)
+    - [Notes / Constraints / Caveats](#notes--constraints--caveats)
+    - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [Configuration API](#configuration-api)
+    - [Implementation Sketch](#implementation-sketch)
+    - [Security Model](#security-model)
+    - [Interaction with Cluster Lifecycle](#interaction-with-cluster-lifecycle)
+    - [Test Plan](#test-plan)
+    - [Graduation Criteria](#graduation-criteria)
+    - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements](https://github.com/kubernetes/enhancements)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation in [kubernetes/website](https://github.com/kubernetes/website)
+- [ ] Supporting documentation (issues, mailing list threads, relevant PRs)
+
+## Summary
+
+Today kube-proxy serves its **administrative** HTTP endpoints (health checks, Prometheus metrics, optional profiling and debug handlers, configz/flagz, etc.) over **cleartext HTTP** on:
+
+- the **metrics** listener (`MetricsBindAddress`, default `127.0.0.1:10249` on IPv4 nodes), and
+- the **health** listener (`HealthzBindAddress`, default `0.0.0.0:10256` / `[::]:10256` depending on configuration).
+
+This KEP proposes **optional TLS (HTTPS)** for those listeners so that operators who expose these ports beyond localhost (or who require encryption-in-transit for compliance) can do so using Kubernetes-consistent certificate configuration, without changing how kube-proxy implements **data-plane** forwarding (iptables/IPVS/nftables/kernelspace).
+
+## Motivation
+
+- **Confidentiality and integrity**: Metrics and health responses can reveal cluster topology, timing, and operational state. On shared networks or when bind addresses are widened, cleartext HTTP is vulnerable to passive observation and tampering.
+- **Alignment with other control-plane and node components**: Many Kubernetes components already expose privileged operational endpoints over TLS or are moving that direction; kube-proxy remains a notable exception for its metrics/health servers.
+- **Policy / compliance**: Some environments mandate TLS for any management interface, even on the node loopback, when traffic crosses certain boundaries (e.g., sidecars, host networking, or compliance frameworks).
+
+### Goals
+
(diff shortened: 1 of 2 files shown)


keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+  certFile: /var/lib/kube-proxy/pki/healthz.crt
+  keyFile: /var/lib/kube-proxy/pki/healthz.key
+```
+
+**Fields to consider** (exact names subject to API review):
+
+- `enable` (bool): when false, HTTP (current behavior).
+- `certFile`, `keyFile` (strings): PEM paths; required when `enable` is true.
+- Optional: `clientCAFile` for client verification; `cipherSuites`, `minVersion`, `curvePreferences` aligned with existing Kubernetes flags elsewhere.
+- Optional future: `certDirectory` / auto-generated self-signed certs for development only (mirrors some binaries); **not** required for MVP if it complicates security posture.
```
aojea, line 132: Can you please do the exercise of analying current kube-scheduler, kube-controller-manager and kubelet and base your proposal on that so we are consistent?

danwinship: Yeah, my impression is that we don't do TLS for metrics anywhere. I know in openshift, the standard pattern is that we just only serve metrics to `127.0.0.1`, and then we run a kube-rbac-proxy to handle authenticated access to it from off-node

yashsingh74: @aojea 
> Can you please do the exercise of analying current kube-scheduler, kube-controller-manager and kubelet and base your proposal on that so we are consistent?

I've analyzed the TLS patterns across control plane components:

kube-apiserver: Full TLS with client authentication for serving APIs
kube-scheduler: Has --secure-port for webhooks/metrics, uses TLS when enabled
kube-controller-manager: Similar secure serving options with TLS support
kubelet: Supports TLS for its API server (--tls-cert-file, --tls-private-key-file)

yashsingh74: > Yeah, my impression is that we don't do TLS for metrics anywhere. I know in openshift, the standard pattern is that we just only serve metrics to `127.0.0.1`, and then we run a kube-rbac-proxy to handle authenticated access to it from off-node

@danwinship 
You're correct that kube-rbac-proxy is a common solution, and this KEP doesn't replace that pattern. Instead, it addresses different use cases:

- Compliance environments: Some require TLS for all HTTP traffic, even on localhost
- Certificate-based authentication: Where client certs are preferred over RBAC proxy architecture
- Simplified deployments: Direct TLS without additional sidecar complexity
- Consistency: Aligning with other components that offer native TLS options


I will refine KEP based on the feedback.

keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+2. **Proxy health server** — today `ProxyHealthServer.Run` in `pkg/proxy/healthcheck/proxy_health.go`, serving `/healthz` and `/livez` on `HealthzBindAddress`.
+
+When TLS is **disabled** (default), behavior is unchanged: plain HTTP.
+
+When TLS is **enabled** for a given listener, kube-proxy wraps the TCP listener with `tls.NewListener` (or equivalent) using configured cert/key material and optional client authentication settings.
+
+### User Stories
+
+1. **As a cluster operator**, I bind metrics on a non-loopback interface for scraping; I want **HTTPS** with a cluster-issued or corporate CA so scrapers verify the server identity.
+2. **As a security reviewer**, I want kube-proxy’s **health** port on `10256` to use TLS so that JSON health payloads are not sent in cleartext on the node network segment.
```
danwinship, line 86: There is no actual security reason to have the health port be TLS:

1. The healthz response is not secret. If you prevent someone from reaching /healthz, they can reasonably accurately determine whether it would have claimed "healthy" or "not healthy" just by trying to connect to Services on that node.
2. The healthz response does not depend on the request in any way; every caller (at any given time) gets the same response. So if the Service is accessible to all users, then there is no point in worrying about "cleartext", because even if you can't sniff someone else's response, you can just get the same response yourself.
3. If healthz is configured to use client certificates so that it's _not_ accessible to all users, then it is only useful if your cloud load balancers support using client certificates with health checks, and it doesn't seem like popular cloud platforms support this.

You pointed out in another comment that other k8s components support https metrics, but do any support https health checks?

If there are situations where compliance says "literally everything has to be encrypted even if it's not secret", then we can add TLS here, but it's not actually going to help anything.

keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+**Fields to consider** (exact names subject to API review):
+
+- `enable` (bool): when false, HTTP (current behavior).
+- `certFile`, `keyFile` (strings): PEM paths; required when `enable` is true.
+- Optional: `clientCAFile` for client verification; `cipherSuites`, `minVersion`, `curvePreferences` aligned with existing Kubernetes flags elsewhere.
+- Optional future: `certDirectory` / auto-generated self-signed certs for development only (mirrors some binaries); **not** required for MVP if it complicates security posture.
+
+**Validation rules (draft):**
+
+- If `enable` is true, `certFile` and `keyFile` must both be set and readable at startup (or follow dynamic cert loading semantics if chosen).
```
danwinship, line 136: You don't explain "dynamic cert loading semantics" anywhere

keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+- TLS must not be silently “half-enabled”; invalid combinations fail `kube-proxy` validation with clear errors.
+- Document interaction with **`BindAddressHardFail`**: TLS handshake or cert load failures should respect existing fatal-bind semantics where applicable.
```
danwinship, line 138: should probably just require `bindAddressHardFail` to be set when using TLS.

keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+- Extend construction / `Run` to accept optional TLS config; apply to the listener used in `server.Serve(listener)` today.
+- Log line should say HTTPS when TLS is enabled (avoid claiming “HTTP server” when it is not).
+
+**Shared code**:
+
+- Prefer consolidating TLS construction in a small helper under `cmd/kube-proxy/app/` or `pkg/proxy` to avoid duplication between metrics and health paths.
+
+**Feature gate** (recommended):
+
+- e.g. `KubeProxyAdminTLS` (name subject to review) — gates new config fields and TLS code paths until beta/GA criteria are met.
```
danwinship, line 158: should clarify that it's about listening ports.

if we don't do healthz then it could be `KubeProxyHTTPSMetrics`. If we do, `KubeProxyHTTPSListeners`

keps/sig-network/6038-kube-proxy-https-admin-serving/README.md
```diff
@@ -0,0 +1,273 @@
...
+- Enabling TLS: rolling DaemonSet update with coordinated scraper/probe updates, or enable per-node with canaries.
+- Downgrading: disable TLS in config before rolling back to a binary that lacks the fields (or keep fields but gate prevents use—document clearly).
+
+### Version Skew Strategy
+
+- Skew between **kube-proxy versions** on nodes is independent; each node’s config controls TLS.
+- **API server** is unaffected.
+- **Monitoring agents** must understand HTTPS per node when enabled.
+
+## Production Readiness Review Questionnaire
```
danwinship, line 214: You'll need to fill out the full PRR questionnaire (at least the Alpha parts for now)

danwinship (commented):
Sorry, catching up with KEP PRs...

This is not currently tracked for 1.38. Are you interested in working on this in the 1.38 timeframe? If so, the [KEP Readiness Deadline](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) (formerly known as PRR Freeze) is September 23.

PR #5740, opened 2025-12-16
KEP-4222: Add requirement to support discovery, statusz, etc.

benluddy:
- One-line PR description: Add requirement to support discovery APIs.

- Issue link: https://kep.k8s.io/4222

- Other comments:

PR #6044, opened 2026-04-29
KEP-4222: Update beta requirements and target for 1.37.

benluddy:
- One-line PR description: Adding found beta requirements and updating beta target to 1.37.

- Issue link: https://github.com/kubernetes/enhancements/issues/4222

- Other comments: Consolidates https://github.com/kubernetes/enhancements/pull/5459 and https://github.com/kubernetes/enhancements/pull/5740.

### DIFF
--- a/keps/sig-api-machinery/4222-cbor-serializer/README.md
+++ b/keps/sig-api-machinery/4222-cbor-serializer/README.md
@@ -1084,6 +1084,8 @@ in back-to-back releases.
   custom CBOR behaviors.
 - All Kubernetes components have opted out of automatic transcoding to JSON for FieldsV1 and
   RawExtension.
+- Collection (i.e. List) object encoding supports "true" streaming (i.e. buffer size is not
+  proportional to output size).
 
 #### GA
 
--- a/keps/sig-api-machinery/4222-cbor-serializer/README.md
+++ b/keps/sig-api-machinery/4222-cbor-serializer/README.md
@@ -1086,6 +1086,7 @@ in back-to-back releases.
   RawExtension.
 - Collection (i.e. List) object encoding supports "true" streaming (i.e. buffer size is not
   proportional to output size).
+- Structured endpoints like discovery, statusz, flagz, etc., support CBOR.
 
 #### GA
 
--- a/keps/sig-api-machinery/4222-cbor-serializer/kep.yaml
+++ b/keps/sig-api-machinery/4222-cbor-serializer/kep.yaml
@@ -27,7 +27,7 @@ latest-milestone: "v1.32"
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
   alpha: "v1.32"
-  beta: ""
+  beta: "1.37"
   stable: ""
 
 # The following PRR answers are required at alpha release


keps/sig-api-machinery/4222-cbor-serializer/README.md
```diff
@@ -1084,6 +1084,9 @@ in back-to-back releases.
   custom CBOR behaviors.
 - All Kubernetes components have opted out of automatic transcoding to JSON for FieldsV1 and
   RawExtension.
+- Collection (i.e. List) object encoding supports "true" streaming (i.e. buffer size is not
```
jpbetz, line 1087: +1 this is essential

keps/sig-api-machinery/4222-cbor-serializer/README.md
```diff
@@ -1084,6 +1084,9 @@ in back-to-back releases.
...
+  proportional to output size).
+- Structured endpoints like discovery, statusz, flagz, etc., support CBOR.
```
jpbetz, line 1089: I just hit a CBOR issue today I wanted to mention.

For the field `GroupPriorityMinimum` in `staging/src/k8s.io/kube-aggregator/pkg/apis/apiregistration/v1/types.go I had added an entry to `test/integration/etcd/data.go` something like:  `"spec": {"group": "foo.com", "version": "as2", "groupPriorityMinimum":100, "versionPriority":10}`.  This works fine for JSON but breaks for CBOR due to a float vs int issue.

jpbetz: This might be a problem in common ecosystem workflows?  I'm imagining something like `dynClient.Resource(...).Create(...)` with a `yaml.Unmarshal` or `json.Unmarshal` of a type like APIService with int fields is going to end up with floats sent by the client and an error when CBOR is enabled

chenk008: @jpbetz I've tested the CBOR serializer for int32, and it appears to be working correctly on my end. Could you please provide a script that reproduces the issue you're seeing?

```
		{
			name: "test APIService",
			in: v1.APIService{
				ObjectMeta: metav1.ObjectMeta{
					Name: "as2.foo.com",
				},
				Spec: v1.APIServiceSpec{
					Group: "foo.com",
					Version: "as2",
					GroupPriorityMinimum: 100,
					VersionPriority:      10,

				},
			},
			want:          []byte{
				0xa3, 0x44, 0x73, 0x70, 0x65, 0x63, 0xa4, 0x45, 0x67, 0x72, 0x6f, 0x75,
				0x70, 0x47, 0x66, 0x6f, 0x6f, 0x2e, 0x63, 0x6f, 0x6d, 0x47, 0x76, 0x65,
				0x72, 0x73, 0x69, 0x6f, 0x6e, 0x43, 0x61, 0x73, 0x32, 0x4f, 0x76, 0x65,
				0x72, 0x73, 0x69, 0x6f, 0x6e, 0x50, 0x72, 0x69, 0x6f, 0x72, 0x69, 0x74,
				0x79, 0x0a, 0x54, 0x67, 0x72, 0x6f, 0x75, 0x70, 0x50, 0x72, 0x69, 0x6f,
				0x72, 0x69, 0x74, 0x79, 0x4d, 0x69, 0x6e, 0x69, 0x6d, 0x75, 0x6d, 0x18,
				0x64, 0x46, 0x73, 0x74, 0x61, 0x74, 0x75, 0x73, 0xa0, 0x48, 0x6d, 0x65,
				0x74, 0x61, 0x64, 0x61, 0x74, 0x61, 0xa1, 0x44, 0x6e, 0x61, 0x6d, 0x65,
				0x4b, 0x61, 0x73, 0x32, 0x2e, 0x66, 0x6f, 0x6f, 0x2e, 0x63, 0x6f, 0x6d,
			},
			// {
			// 	"spec": {
			// 		"group": "foo.com",
			// 		"version": "as2",
			// 		"versionPriority": 10,
			// 		"groupPriorityMinimum": 100
			// 	},
			// 	"status": {},
			// 	"metadata": {
			// 		"name": "as2.foo.com"
			// 	}
			// }
			assertOnError: assertNilError,
		},
```

jpbetz: EDIT: Looks like this is fine so long as `k8s.io/apimachinery/pkg/util/json` is used. It always converts whole numbers to int64.

Here's a reproduction: https://github.com/jpbetz/kubernetes/pull/new/cbor-dynamic-repro

This test passes if you use JSON instead of CBOR.

The problem with CBOR:

- Load data for an in-tree resource from JSON data into unstructured
- Enable CBOR on the client
- Use the dynamic client to create the resource
- Create results in error: "cbor: cannot unmarshal primitives into Go struct field v1.Pod.spec of type int64"

jpbetz: This is not blocking. We may get reports from the community if someone uses `encoding/json` to load unstructured like in my above example, but that's not the recommended approach.

benluddy: I missed the latest comments here and wrote my own test (https://github.com/kubernetes/kubernetes/pull/139149). I'm working on an upstream patch to support this case since it will be a point of friction for some users.

PR #6050, opened 2026-04-30
KEP 6030: Dynamic Resize of Memory-Backed Volumes [alpha]

natasha41575:
New KEP for Dynamic Resize of Memory-Backed Volumes (alpha targeting 1.37)

Issue link: #6030

This KEP proposes extending the In-Place Pod Vertical Scaling infrastructure to support dynamic resizing of memory-backed `emptyDir` volumes without pod disruption. 

A working proof-of-concept implementation based on the proposed design can be found at https://github.com/kubernetes/kubernetes/pull/139425. The goal of the PoC is to demonstrate the simplicity and feasibility.

Antigravity was used to help with research and writing. However:
- I thoroughly reviewed / fact-checked all claims made by AI to the best of my ability, including running manual tests, combing through the codebase, cross-checking documentation, and prototyping code changes to ensure technical feasibility.
- All design ideas came from humans.

### DIFF
--- a/keps/prod-readiness/sig-node/6030.yaml
+++ b/keps/prod-readiness/sig-node/6030.yaml
@@ -1,0 +1,3 @@
+kep-number: 6030
+alpha:
+  approver: "kannon92"
--- a/keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
+++ b/keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
@@ -1,0 +1,1051 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+Follow the guidelines of the [documentation style guide].
+In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-6030: Dynamic Resize of Memory-Backed Volumes
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+    - [OOM-kills](#oom-kills)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+    - [New Status Fields and State Mapping](#new-status-fields-and-state-mapping)
+    - [API Validation and Restrictions](#api-validation-and-restrictions)
+    - [Resize Restart Policy](#resize-restart-policy)
+    - [Atomic Resize Principle](#atomic-resize-principle)
+  - [Component Interaction &amp; Design Flow](#component-interaction--design-flow)
+  - [1. User/Control Plane](#1-usercontrol-plane)
+  - [2. Kubelet Pod Workers &amp; Allocation](#2-kubelet-pod-workers--allocation)
+  - [3. Coordination &amp; Actuation (KuberuntimeManager)](#3-coordination--actuation-kuberuntimemanager)
+  - [4. Observation &amp; Feedback Loop](#4-observation--feedback-loop)
+  - [Resource Coordination: Volume and Cgroup Ordering of Updates](#resource-coordination-volume-and-cgroup-ordering-of-updates)
+    - [Background: The Enforcement Divergence](#background-the-enforcement-divergence)
+      - [Memory Attribution &amp; cgroup Limits](#memory-attribution--cgroup-limits)
+      - [Sharing Volumes Between Multiple Containers](#sharing-volumes-between-multiple-containers)
+      - [Init Container Lifecycles &amp; Handoff](#init-container-lifecycles--handoff)
+    - [Priority of Enforcement](#priority-of-enforcement)
+    - [Order of Actuation](#order-of-actuation)
+      - [Asynchronous Updates or Opposite Directions](#asynchronous-updates-or-opposite-directions)
+      - [Actuation Priority Matrix](#actuation-priority-matrix)
+    - [Role of the Allocation Manager and Kuberuntime Manager](#role-of-the-allocation-manager-and-kuberuntime-manager)
+    - [Volume Manager interface](#volume-manager-interface)
+    - [The Role of the Container Runtime](#the-role-of-the-container-runtime)
+  - [Shrinkage Safety](#shrinkage-safety)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [emptyDir Size Limit Monitoring and Eviction](#emptydir-size-limit-monitoring-and-eviction)
+  - [Interaction with Ephemeral Storage: Resource Accounting and Eviction](#interaction-with-ephemeral-storage-resource-accounting-and-eviction)
+  - [Interaction with Secrets, Projected Volumes, and DownwardAPI](#interaction-with-secrets-projected-volumes-and-downwardapi)
+    - [ConfigMaps](#configmaps)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha-1)
+    - [Beta](#beta-1)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Upgrade](#upgrade)
+    - [Downgrade](#downgrade)
+  - [Version Skew Strategy](#version-skew-strategy)
+    - [Kubelet vs API Server](#kubelet-vs-api-server)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Volume Reconciler detects changes (Async approach)](#volume-reconciler-detects-changes-async-approach)
+  - [Moving <code>emptyDir</code> handling entirely to <code>KuberuntimeManager</code>](#moving-emptydir-handling-entirely-to-kuberuntimemanager)
+  - [Restricting <code>sizeLimit</code> to not exceed memory limits](#restricting-sizelimit-to-not-exceed-memory-limits)
+  - [Non-atomic handling of volume and resource resizes](#non-atomic-handling-of-volume-and-resource-resizes)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes a mechanism to dynamically resize `emptyDir` volumes with `medium: Memory` (tmpfs) without requiring Pod recreation or Pod restart, and while maintaining the files already written to the volume's filesystem. It leverages the `InPlacePodVerticalScaling` (IPPVS) infrastructure to manage the transition from the desired state to the actual state via an allocation and actuation loop.
+
+## Motivation
+
+Currently, increasing memory-backed storage (tmpfs `emptyDir` volumes) requires Pod recreation, which is disruptive for high-performance caches or in-memory databases that are otherwise capable of resizing their memory footprints dynamically.
+
+This enhancement builds directly on the foundation laid by **In-Place Pod Vertical Scaling (IPPVS)** (KEP-1287). Just as IPPVS allows scaling container CPU and memory resources without Pod disruption, high-performance workloads frequently require scaling their scratch space or in-memory caches simultaneously. Since memory-backed `emptyDir` volumes consume the node's memory (and can be bounded by the Pod's memory limits), integrating their resize lifecycle with the IPPVS allocation and actuation phases ensures consistent resource accounting and prevents node exhaustion.
+
+### Goals
+
+*   Allow updating `emptyDir.sizeLimit` for existing Pods when `medium` is `Memory`.
+*   Reflect resize progress in `Pod.Status` using an allocation/actuation model similar to IPPVS.
+*   Actuate the resize via `tmpfs` remount without Pod disruption and while maintaining the files already written to the volume's filesystem.
+*   Support dynamic resizing of memory-backed volumes on cgroup v2 nodes only (where memory charging and charge migration for `tmpfs` are properly supported by the kernel).
+
+### Non-Goals
+
+*   Resizing non-memory-backed `emptyDir` volumes (disk-backed).
+*   Resizing other volume types (such as PVCs, CSI ephemeral volumes etc.) through this mechanism.
+*   Supporting this feature on nodes running with cgroup v1 (due to limitations in memory accounting and tracking for shared `tmpfs` volumes).
+
+## Proposal
+
+We propose extending the `InPlacePodVerticalScaling` pattern to memory-backed `emptyDir` volumes.
+
+### User Stories
+
+#### Story 1
+A user runs an in-memory database in a Pod and needs to increase the available cache size. The user updates the `emptyDir.sizeLimit` in the Pod spec. The Kubelet remounts the `tmpfs` volume with the new size limit without restarting the container, allowing the database to expand its cache immediately.
+
+#### Story 2
+An autoscaling controller dynamically scales the memory limits of workload worker pods (e.g., for machine learning or data processing) to handle varying demands. These workers utilize memory-backed `emptyDir` volumes for shared memory and caching, typically sizing the volume to a fixed proportion of the container's total memory limit. 
+
+While the controller can use `InPlacePodVerticalScaling` to scale up a container's memory limit without restarting the pod, the inability to dynamically resize the associated memory-backed volume limits the worker's ability to utilize the expanded capacity for shared memory tasks without a disruptive restart. This feature allows the controller to update both the container memory limits and the `emptyDir.sizeLimit` simultaneously, providing seamless vertical scaling for distributed workloads.
+
+### Risks and Mitigations
+
+#### OOM-kills
+
+The user is allowed to set the volume `sizeLimit` higher than the pod or container memory limits. Attempting to add such a restriction is considered out of scope for this KEP (see [Restricting sizeLimit to not exceed memory limits](#restricting-sizelimit-to-not-exceed-memory-limits)). This means that the user is responsible for ensuring that the volume `sizeLimit` is set to be within the pod or container memory limits. Failure to do so may result in OOM-kills as applications within the pod fill up the volume.
+
+The mitigation for this risk is that the API server can emit a warning when it sees a resize request that results in the volume `sizeLimit` being set to higher than the pod-level memory limits. 
+
+## Design Details
+
+### API Changes
+
+#### New Status Fields and State Mapping
+
+To support dynamic resizing, the `Pod.Spec.Volumes[].EmptyDir.SizeLimit` field is made **mutable** when modified via the `/resize` subresource for existing Pods (when `medium` is `Memory`). This represents the **Desired State**.
+
+We introduce a new sub-structure under `corev1.VolumeStatus` (in `Pod.Status`) to track the actual resize state of the volume:
+
+*   **Desired State**: Represented by the mutable `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
+*   **Allocated State**: Checkpointed and maintained by the Kubelet's `AllocationManager` for local resource accounting, but **not exposed** in `Pod.Status`.
+    *   *Definition*: Represents the volume capacity acknowledged and admitted by the Kubelet on the node.
+    *   *Initial Admission*: Admitted and checkpointed locally by the Kubelet once the initial resource admission phase completes.
+    *   *Admission during Resize*: Upon admitting a resize update, the `AllocationManager` updates the Kubelet's local checkpoint with the newly acknowledged capacity.
+    *   *Scope*: Maintained only for memory-backed `emptyDir` volumes.
+*   **Actual State**: Represented by the new `VolumeStatus.EmptyDir.SizeLimit`.
+    *   *Definition*: Represents the actual mounted filesystem capacity reported by the mount/remount call.
+    *   *Pod Creation*: Remains unpopulated (`nil`) until Kubelet completes the initial volume mount setup and receives capacity confirmation from the mounter.
+    *   *Pod Resize*: During an active resize operation, this field retains the *old* mounted size until the Kubelet successfully actuates the `tmpfs` remount, at which point it is updated to the new actual capacity.
+    *   *Default/Fallback*: Remains unpopulated (`nil`) for all non-memory-backed volumes.
+
+The new status field `Pod.Status.ContainerStatuses[].VolumeMounts[].VolumeStatus.EmptyDir.SizeLimit` is only populated for memory-backed volumes, and remains unpopulated for other types of volumes. The `VolumeStatus` Go struct is updated as follows:
+
+```go
+type VolumeStatus struct {
+    ...
+    // emptyDir represents the status of an emptyDir volume.
+    // +optional
+    EmptyDir *EmptyDirVolumeStatus `json:"emptyDir,omitempty" protobuf:"bytes,2,opt,name=emptyDir"`
+}
+type EmptyDirVolumeStatus struct {
+    // sizeLimit represents the actual mounted capacity of the emptyDir volume.
+    // This is only populated for memory-backed emptyDir.
+    // +optional
+    SizeLimit *resource.Quantity `json:"sizeLimit,omitempty" protobuf:"bytes,1,opt,name=sizeLimit"`
+}
+```
+
+#### API Validation and Restrictions
+
+API validation logic ensures that `Pod.Spec.Volumes[].EmptyDir.SizeLimit` is only mutable when:
+* The volume is memory-backed (meaning the `medium` field is set to `Memory`).
+* Updates occur through the `resize` subresource.
+
+Only modifications to an existing `Pod.Spec.Volumes[].EmptyDir.SizeLimit` are permitted.
+* The addition or removal of a memory-backed volume in a pod spec is forbidden. 
+* For Alpha, adding or removing a `sizeLimit` to an existing memory-backed volume is not allowed. This boundary is re-evaluated for Beta.
+
+#### Resize Restart Policy 
+
+There are existing resize restart policies that control when a container is restarted during a resize operation, which are described in more detail in the [InPlacePodVerticalScaling KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#container-resize-policy).
+
+We do not introduce an analogous policy nor honor the existing restart policies when resizing memory-backed `emptyDir` volumes. The resizing of memory-backed `emptyDir` volumes is fundamentally non-disruptive to running processes and does not require application-level awareness beyond standard handling of filesystem capacity. If a volume resize is accompanied by a container or pod resource resize, the container's existing `RestartPolicy` governs container restarts.
+
+#### Atomic Resize Principle 
+
+The atomic resize principle is preserved. The Kubelet treats updates to volume `sizeLimit` and container or pod-level resources (both requests and limits) as a single atomic transaction. This applies to requests that modify multiple volumes, multiple containers, pod-level resources, or any combination thereof.
+
+The `AllocationManager` performs an All-or-Nothing Admission check, rejecting the entire update if the resize feasibility checks do not pass. The volume `sizeLimit` is also checkpointed atomically along with container and pod resources.
+* The volume `sizeLimit` is not taken into consideration during the resource-fit checks, but will fail the admission check atomically if the container/pod resize fails to be admitted.
+* On cgroups v1, volume resize is not supported, so any resize request containing a `sizeLimit` update will be atomically rejected as `Infeasible` during the admission phase.
+
+The Pod's `ResizeStatus` (tracked via `PodResizeInProgress` condition) only transitions to completed once all container and volume states reach their respective desired values. If any component of the update fails, the status remains in a `PodResizeInProgress` state (with an error) to reflect the incomplete transition. This ensures resource integrity by preventing "partial resizes" that could leave a volume expanded within a container that fails to scale its memory budget.
+
+See [Non-atomic handling of volume and resource resizes](#non-atomic-handling-of-volume-and-resource-resizes) for the alternative that was considered and rejected.
+
+### Component Interaction & Design Flow
+
+The following diagram illustrates the proposed end-to-end flow.
+
+```mermaid
+graph TD
+    User["User / Controller"]
+    API["API Server"]
+    Kubelet["Kubelet Pod Worker"]
+    Alloc["AllocationManager"]
+    KRT["KuberuntimeManager"]
+    VM["Volume Manager"]
+    CRI["CRI"]
+    Kernel["Linux Kernel tmpfs"]
+
+    User -->|1. Updates Spec| API
+    API -->|2. Syncs Pod| Kubelet
+    Kubelet -->|3. Requests Allocation| Alloc
+    Alloc -->|4. Checkpoints State Locally| Alloc
+    Kubelet -->|"5. Updates Status (Container Resources)"| API
+    KRT -->|6. Reads Checkpointed Allocation| Alloc
+    KRT -->|7. Evaluates Resize Direction| KRT
+    KRT -->|"8a. Remounts Volume (Downsize)"| VM
+    KRT -->|8b. Updates Cgroups| CRI
+    KRT -->|"8c. Remounts Volume (Upsize)"| VM
+    VM -->|9. Executes Remount System Call| Kernel
+    KRT -->|"10. Updates Status (Volume Mount Actual State)"| API
+```
+
+The following steps elaborate on this flow:
+
+### 1. User/Control Plane
+*   The user or an external controller updates `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
+*   The API server preserves volume updates during the `dropNonResizeUpdates` phase.
+
+### 2. Kubelet Pod Workers & Allocation
+*   The Kubelet detects the specification change in `HandlePodUpdates`.
+*   The `AllocationManager` validates the new resize request. The changing `sizeLimit` does not affect the feasibility of the resize on cgroup v2, as pod admission takes only pod and container requests into consideration. If the node is running with cgroup v1, the `AllocationManager` will reject the resize request as `Infeasible`.
+*   If allocation succeeds, the `AllocationManager` updates the Kubelet's internal allocated state. The `AllocatedSizeLimit` is checkpointed locally (along with container and pod requests/limits) to ensure that memory-backed volume size limits survive Kubelet restarts.
+
+### 3. Coordination & Actuation (KuberuntimeManager)
+*   The `KuberuntimeManager` reads the `AllocatedSizeLimit` from the local checkpointed state maintained by `AllocationManager`.
+*   It acts as the central orchestrator for the actuation phase to enforce strict ordering. See [Resource Coordination: Volume and Cgroup Ordering of Updates](#resource-coordination-volume-and-cgroup-ordering-of-updates) for the detailed rationale behind this ordering.
+*   The `KuberuntimeManager` checks the direction of the volume resize (upsize vs. downsize) by comparing the newly allocated size with the checkpointed **Actuated State** (the target size that the Kubelet last attempted to apply, which is maintained in Kuberuntime's internal checkpoint). Comparing against the *actuated* state rather than the live kernel state avoids unnecessary remount requests caused by kernel rounding differences.
+* The `KuberuntimeManager` directly calls a new interface on the Volume Manager (see [Volume Manager interface](#volume-manager-interface)) to perform the remount at the correct time, bypassing the async reconciler loop. The Volume Manager executes the remount: `mount -o remount,size=<limit> -t tmpfs tmpfs <path>`. Once this actuation completes, the `KuberuntimeManager` checkpoints this sizeLimit in its own **Actuated State**. This Actuated State represents what the Kubelet tries to actuate, which may diverge slightly from what is read directly from the kernel (`Actual State`).
+* **Note**: Bypassing the async reconciler loop is both safe and inconsequential. The `emptyDir` reconciler today never remounts because `RequiresRemount` is hardcoded to always return false; this means that the `KuberuntimeManager` is the only component that makes this change. We considered alternatives such as relying on the `emptyDir` reconciler to detect changes asynchronously or moving `emptyDir` handling entirely to `KuberuntimeManager`, but they are ruled out. See [Alternatives](#alternatives) for more details.
+
+### 4. Observation & Feedback Loop
+*   Upon successful remount and cgroup update, the `KuberuntimeManager` checkpoints the new target values as the current 'actuated resources' (the baseline for future resizes).
+*   The Kubelet updates `Pod.Status.ContainerStatuses[].VolumeMounts[].VolumeStatus.EmptyDir.SizeLimit` to reflect the successful resize.
+*   Once the actual size matches the allocated size checkpoint (and container resource updates are complete), the Kubelet clears the `PodResizeInProgress` condition.
+
+### Resource Coordination: Volume and Cgroup Ordering of Updates
+ 
+#### Background: The Enforcement Divergence
+
+Memory-backed volumes consume the same physical RAM as the container's processes. While all usage is enforced by the container and pod cgroups, it is governed by two different enforcement thresholds:
+
+*   **tmpfs `size` option**: An internal quota for the volume. When a write exceeds the volume's `sizeLimit`, the kernel returns a disk full (**ENOSPC**) error. This is a graceful I/O exception.
+*   **cgroup `memory.max`**: The external limit for the entire container. Since `tmpfs` usage counts toward the cgroup total, hitting this limit triggers an **OOM (Out of Memory) Kill**, resulting in container termination.
+
+##### Memory Attribution & cgroup Limits
+
+Because memory-backed volumes are backed by `tmpfs`, their memory usage is charged directly to the **cgroup of the process that writes the data**, rather than to the volume itself.
+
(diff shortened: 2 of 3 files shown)


keps/prod-readiness/sig-node/6030.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 6030
+alpha:
+  approver: "TBD"
```
kannon92:
```suggestion
  approver: "kannon92"
```

I'll go ahead and take this one.

natasha41575: thank you!!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,906 @@
...
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+-->
+
+### Scalability
+
+<!--
+For alpha, this section is encouraged: reviewers should consider these questions
```
kannon92, line 851: Since you have time before 1.37 freeze, can you add this section?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,906 @@
...
+
+While the controller can use `InPlacePodVerticalScaling` to scale up a container's memory limit without restarting the pod, the inability to dynamically resize the associated memory-backed volume limits the worker's ability to utilize the expanded capacity for shared memory tasks without a disruptive restart. This feature allows the controller to update both the container memory limits and the `emptyDir.sizeLimit` simultaneously, providing seamless vertical scaling for distributed workloads.
+
+### Risks and Mitigations
+
+#### OOM-kills
+
+The user is allowed to set the volume `sizeLimit` higher than the pod or container memory limits. Attempting to add such a restriction is considered out of scope for this KEP (see [Restricting sizeLimit to not exceed memory limits](#restricting-sizelimit-to-not-exceed-memory-limits)). This means that the user is reponsible for ensuring that the volume `sizeLimit` is set to be within the pod or container memory limits. Failure to do so may result in OOM-kills as applications within the pod fill up the volume.
```
kannon92: This reminds me of https://github.com/kubernetes/kubernetes/issues/128339.

I'm not saying we have to solve this issue but curious if this issue will be more of a problem of resizing.

natasha41575: It's definitely related. I don't think it's necessarily more likely with resizing though. Resizing might actually help as you could increase the pod/container memory limits so that it doesn't get oom-killed on restart. 

I talked with @ndixita about this problem a couple weeks back and we discussed that might be able to improve the situation in the context of the pod-level resources project. It could be a reasonable restriction to say that a volume's `sizeLimit` can't exceed pod-level memory limits, which could help but not necessarily eradicate the issue.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: done

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+
+We introduce two new fields to `corev1.VolumeStatus` (included within `VolumeMountStatus` in `Pod.Status`) to track the resize lifecycle:
+
+*   **Desired State**: Represented by the mutable `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
+*   **Allocated State**: Represented by the new `VolumeStatus.AllocatedSizeLimit`.
+    *   *Definition*: Represents the volume capacity acknowledged and reserved by the Kubelet on the node.
+    *   *Pod Creation*: Unpopulated (`nil`) until Kubelet's initial resource admission phase completes, at which point it is populated with the admitted size limit.
+    *   *Pod Resize*: Upon admitting a resize update, the `AllocationManager` updates the Kubelet's internal checkpoint, and the Kubelet updates this status field to reflect the newly acknowledged capacity.
+    *   *Default/Fallback*: Remains unpopulated (`nil`) for all non-memory-backed volumes.
+*   **Actual State**: Represented by the new `VolumeStatus.SizeLimit`.
```
tallclair: I don't think we should put EmptyDir-specific fields in the top-level VolumeStatus. Is there a more generic way we can represent this, or do we just need to start adding a per-volume-type status to capture additional data?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+### API Changes
+
+#### New Status Fields and State Mapping
+
+To support dynamic resizing, the `Pod.Spec.Volumes[].EmptyDir.SizeLimit` field is made **mutable** for existing Pods (when `medium` is `Memory`). This represents the **Desired State**.
+
+We introduce two new fields to `corev1.VolumeStatus` (included within `VolumeMountStatus` in `Pod.Status`) to track the resize lifecycle:
+
+*   **Desired State**: Represented by the mutable `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
+*   **Allocated State**: Represented by the new `VolumeStatus.AllocatedSizeLimit`.
```
tallclair: Do we really need to surface the allocated state? We don't currently surface allocated limits to the API.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+*   The `KuberuntimeManager` checks the direction of the volume resize (upsize vs. downsize) by comparing the newly allocated size with the checkpointed **Actuated State** (the target size that the Kubelet last attempted to apply, which is maintained in Kuberuntime's internal checkpoint). Comparing against the *actuated* state rather than the live kernel state avoids unnecessary remount requests caused by kernel rounding differences.
+* The `KuberuntimeManager` directly calls a new interface on the Volume Manager (see [Volume Manager interface](#volume-manager-interface)) to perform the remount at the correct time, bypassing the async reconciler loop. The Volume Manager executes the remount: `mount -o remount,size=<limit> -t tmpfs tmpfs <path>`. Once this actuation completes, the `KuberuntimeManager` checkpoints this sizeLimit in its own **Actuated State**. This Actuated State represents what the Kubelet tried to actuate, which may diverge slightly from what is read directly from the kernel (`Actual State`).
```
tallclair: This makes me nervous. Can it be built into the reconciler loop, rather than bypassing it? Similar to how we handle on-demand PLEG relist? https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/pleg/generic.go#L190-L229

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+### 4. Observation & Feedback Loop
+*   Upon successful remount and cgroup update, the `KuberuntimeManager` checkpoints the new target values as the current 'actuated resources' (the baseline for future resizes).
+*   The Kubelet updates `Pod.Status.VolumeMounts[].VolumeStatus.SizeLimit` to reflect the successful resize.
+*   Once the actual size matches the allocated size (and any container resource updates are done), the Kubelet clears the `PodResizeInProgress` condition.
+
+### Resource Coordination: Volume and Cgroup Ordering of Updates
+ 
+#### Technical Context: The Enforcement Divergence
+
+Memory-backed volumes consume the same physical RAM as the container's processes. While all usage is enforced by the container and pod cgroups, it is governed by two different enforcement thresholds:
```
tallclair, line 368: Can you add some background on how the memory accounting works with pods and containers? If there is a pod-level limit, I'd assume the memory just counted towards that. What if there is no pod-level limit? I'm confused about how container limits factor in.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+| **Expansion** | Increasing ⬆️ | Increasing ⬆️ | **Cgroup** | **Volume** | Expand envelope before expanding the internal quota. |
+| **Contraction** | Decreasing ⬇️ | Decreasing ⬇️ | **Volume** | **Cgroup** | Shrink the internal quota before tightening the envelope. |
+| **Volume Only** | Increasing / Decreasing | Unchanged | **Volume** | **N/A** | No cgroup update necessary. |
+| **Opposite (Vol Up)** | Increasing ⬆️ | Decreasing ⬇️ | **Cgroup** | **Volume** | If this configuration results in an OOM kill, it will do so regardless of ordering; follows "Vol Up" sequence for consistency. |
+| **Opposite (Vol Down)** | Decreasing ⬇️ | Increasing ⬆️ | **Volume** | **Cgroup** | If this configuration results in an OOM kill it will do so regardless of ordering; follows "Vol Down" sequence for consistency. |
+| **Cgroup Only** | Unchanged | Increasing / Decreasing | **Cgroup** | **N/A** | Standard container resize; no volume remount required. |
+
+#### Role of the Allocation Manager and Kuberuntime Manager
+
+The `AllocationManager`'s role is focused on maintaining the source of truth for allocated resources (checkpointing) and ensuring atomic admission of resource updates.
```
tallclair: This means that the volume manager should only ever see the allocated pod, right? I could see that mattering if the volume needs to be recreated, for example (e.g. after a node reboot).

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,952 @@
...
+
+In the case that there is an ongoing request to shrink the volume below usage, the kubelet will repeatedly attempt to resize the volume on every periodic sync until the resize is either successful, the resize request is cancelled, or the pod is terminated.
+
+#### Ephemeral Storage Monitoring and Eviction
+
+Kubelet is running an ephemeral storage monitoring system (the eviction manager) to ensure that the space usage of the ephemeral storage on the node does not exceed the volume size limits, and to trigger evictions when necessary. The eviction manager today looks at the `sizeLimit` defined in the pod spec, and compares it to the ephemeral storage usage for the pod.
+
+For memory-backed volumes, the eviction manager will be updated to instead look at the actual `sizeLimit` by reading the actual capacity from the pod volume stats. This ensures that eviction manager does not trigger evictions due to changes in the desired `sizeLimit`, but only when the actual `sizeLimit` is exceeded.
```
tallclair: Memory-backed volumes shouldn't affect ephemeral storage eviction, right? Just memory pressure eviction?

This section seems to conflict with the next one.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Let me think on whether it is worth the additional complexity; I think we can do something similar to your suggestion, but it would require some sort of lock so that the kuberuntime manager can block on the reconciliation completing before it proceeds with updating the cgroups after volume downsize

This would mostly be for future-proofing though, as the current emptyDir reconciler loop never remounts (`RequiresRemount` is hardcoded to return false)

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Sorry I think this is worded a bit incorrectly. When I tested it, the eviction manager did trigger eviction when memory-backed volumes exceeded the `sizeLimit`, so it appears to be monitoring not just ephemeral storage but also memory-backed volumes. I'll update this to be more accurate.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: yeah, added a note to state this explicitly

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: took it out!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Thanks for catching this. I changed it to be `VolumeStatus.EmptyDir.SizeLimit` (which is what I meant to do... sorry for the oversight)

I'm not sure that we want to 'generically' represent it... We could theoretically add a top-level VolumeStatus.Capacity field that represents something different depending on the volume type. But this seems confusing and potentially hard to understand.

I know you don't like how we are piece-by-piece duplicating every field in the pod spec into the pod status as we make them mutable (I don't like it either), but I don't have a simple solution for you... iirc you were in the past exploring the idea of having kubelet create an 'allocatedPod' object ?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: I looked into it a bit more and I think your suggestion is a little more 'kubernetes-esque' but it's more complicated if we want to maintain the strict actuation ordering between cgroups and the volume mount:

- The kuberuntime manager can actuate container upsizes immediately, but has to wait for the volume mount resizes to complete before actuating container downsizes.
- The volume manager can actuate volume mount downsizes immediately, but has to wait for container resizes to complete before actuating volume mount upsizes.

So the kuberuntime manager sync loop would need to do something like:
1. actuate cgroup upsize
2. request an on-demand volume manager reconcile and block on its completion
3. actuate cgroup downsize

Assuming that the suggestion is to allow the volume manager reconciler to be triggered both on-demand by the kuberuntime manager and also in its own periodic reconciliation, it would need to do something like:
1. actuate volume mount downsize
2. have a polling loop that queries the kuberuntime manager to see whether allocated resources == actuated resources
3. actuate volume mount upsize

It can be done, but I'm not so sure I'm convinced that this is better as we are trying to perform a series of synchronous operations with two asynchronous reconciliation loops. I think the state management will become extremely complex and hard to reason about.

Since the emptyDir reconciler doesn't actually ever trigger a remount, the current proposal isn't really "bypassing" the reconciler loop; the reconciler's remounting flow for emptyDir today is a no-op. 

https://github.com/kubernetes/kubernetes/blob/f176c7f2692e00503d39c84e06e9cf27b9294150/pkg/volume/emptydir/empty_dir.go#L99-L101

Having kuberuntime manager be the only one that triggers the remount should be safe... but perhaps there is a middle ground alternative?
1. Make it so that the on-demand reconcile is the _only_ time the emptyDir reconciler ever remounts... I'm also not 100% convinced that this is better either.
2. Have a mutex so that if this synchronous path is ever used for a different type of volume (one that supports remount), we are safely preventing reentry into the higher level reconciler code. I'm more amenable to this option.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Added based on some research. The memory for `tmpfs` is accounted for by the kernel directly by the cgroup of the process that writes the data. That means it is charged against the container-level limit (and also counts against the pod-level limits since it's nested under that). 

I had gemini do this research and cite its sources, including them here because I was impressed by it:

Source 1 https://kubernetes.io/docs/concepts/storage/volumes/#emptydir:
> While tmpfs is very fast, be aware that, unlike disks, files you write count against the memory limit of the container that wrote them.

Source 2 https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v2.rst:

> A memory area is charged to the cgroup which instantiated it and stays charged to the cgroup until the area is released. Migrating a process to a different cgroup doesn't move the memory usages that it instantiated while in the previous cgroup to the new cgroup.

> A memory area may be used by processes belonging to different cgroups. To which cgroup the area will be charged is in-deterministic...

Source 3 (linux source code!): 

https://github.com/torvalds/linux/blob/v7.1-rc4/mm/memcontrol.c#L3835-L3837

which explains that terminated cgroup memory charges get reparented. 

I know we discussed offline possibly wanting to change the resource accounting, but I don't think it's feasible as it is done at the kernel-level.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
tallclair: > When I tested it, the eviction manager did trigger eviction when memory-backed volumes exceeded the `sizeLimit`

Is `sizeLimit` a hard limit, or a soft limit? I mean, if I try to write a file that goes over the limit, does the write fail or succeed? I assumed it was kernel-enforced, but this makes it sound like it's a soft-limit enforced through eviction?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
tallclair: OK, I'm convinced. I still need to review the implementation, but I think we can defer the rest of this discussion to that PR.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: It is a limit hard-enforced by the kernel and also soft-enforced by the eviction mgr. If you try to go over the limit, you get an out-of-space error from the kernel.

The eviction manager _also_ enforces the size limit. I discovered that it does this because when I hadn't changed the eviction manager to look at actual rather than desired size limit, I saw the eviction manager erroneously evicting the pod when it had exceeded the desired size limit.

natasha41575:
/assign @kannon92 
(PRR)

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+
+The `AllocationManager`'s role is focused on maintaining the source of truth for allocated resources (checkpointing) and ensuring atomic admission of resource updates. The volume manager should only ever see the allocated size limits, rather than the desired size limits.
+
+The `KuberuntimeManager` serves as the coordination layer that understands the nested relationship between volume size and cgroup limits and enforces the strict ordering of updates.
+
+#### Volume Manager interface
+
+To support direct resizing from the `KuberuntimeManager` while maintaining separation of concerns, this proposal extends the Volume Manager and Volume Plugin interfaces:
+
+*   **VolumeManager Extension**: The `VolumeManager` interface is extended with `ResizeVolume(pod *v1.Pod, volumeName string, newSize *resource.Quantity) error` and `GetVolumeSize(pod *v1.Pod, volumeName string) (*resource.Quantity, error)`. This allows the `KuberuntimeManager` to query the current size and trigger volume operations synchronously at the correct time in its sync loop.
```
jsafrane: `ResizeEphemeralVolume` could fit the VolumeManager better. We already allow resize for PVs and tmpfs resize should be a clearly different operation.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+*   **DirectResizableVolumePlugin Interface**: A new optional interface `DirectResizableVolumePlugin` is introduced in `pkg/volume/plugins.go`. Volume plugins that support online, synchronous resizing (in this case, the `emptyDir` plugin for memory-backed volumes) can implement this interface. The `VolumeManager` checks if the plugin for the volume implements this interface and delegates the call to it.
```
jsafrane: `ResizableEphemeralVolumePlugin` ?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+### Goals
+
+*   Allow updating `emptyDir.sizeLimit` for existing Pods when `medium` is `Memory`.
+*   Reflect resize progress in `Pod.Status` using an allocation/actuation model similar to IPPVS.
+*   Actuate the resize via `tmpfs` remount without Pod disruption.
+
+### Non-Goals
+
+*   Resizing non-memory-backed `emptyDir` volumes (disk-backed).
+*   Resizing other volume types (e.g., PVCs) through this mechanism.
```
jsafrane: PVCs are obvious, but please explicitly add [in-line CSI ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#csi-ephemeral-volumes) to the list. 
They are less known and they allow for extending Kubernetes with e.g. Secrets-like volumes, where the actual secret is not in etcd. But they can do whatever they want, e.g. implement a better version of EmptyDir and mount tmpfs.

I don't see demand for such a feature and it would probably need a different mechanism to resize such a volume. Kubelet blindly passes [volumeAttributes](https://github.com/kubernetes/api/blob/25001c854943b552769cd32eacd0bef7467c0eea/core/v1/types.go#L2294) to the CSI driver and has no knowledge if there is any parameter with the volume size and how it's called. And we don't want to support any other parameter modification.

jsafrane (commented):
From storage perspective, the proposal looks quite good. I left a few small nits.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+
+Below are some examples to consider, in addition to the aforementioned [maturity levels][maturity-levels].
+
+-->
+
+#### Alpha
+- The feature is implemented behind the `InPlacePodVerticalScalingMemoryBackedVolumes` feature gate.
+- Unit test coverage is complete.
+- E2E tests verify successful upsize and downsize operations in a live cluster.
```
kannon92: I am not clear if you are saying you will have e2e tests running in CI or you will perform manual testing?

Can you clarify this?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+*   The API server once again permits mutations to `emptyDir.sizeLimit` via the `resize` subresource.
+*   The Kubelet resumes its reconciliation logic. If a Pod's actual mounted volume size differs from the target allocated capacity checkpointed locally (e.g., if a resize is pending or if the spec was updated but not actuated before rollback), the Kubelet proceeds to actuate the resize to match the desired state.
+
+###### Are there any tests for feature enablement/disablement?
+
+Yes. 
+*   **API Validation Tests**: Unit tests in `pkg/apis/core/validation` verify that mutating `emptyDir.sizeLimit` is forbidden when the feature gate is disabled and permitted when it is enabled.
+*   **Kubelet Tests**: Unit tests in the Kubelet (specifically in `KuberuntimeManager` and the `eviction` manager) verify that actuation and enhanced monitoring logic are skipped or fall back to default behavior when the feature gate is disabled.
+
+### Rollout, Upgrade and Rollback Planning
```
kannon92, line 675: This KEP is so good but it would be better if you filled this section out also.

I won't block on it though.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,990 @@
...
+are missing a bunch of machinery and tooling and can't do that now.
+-->
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+<!--
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
```
kannon92, line 752: This section would be nice espeically to see filled out.

kannon92 (commented):
/approve
/hold

Putting hold to avoid merging.

PRR is really good for alpha.

I have a few nits and ideally would like the questionaire filled out. But I won't block approval on it.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: SG, updated!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: updated!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Added CSI ephemeral volumes to the list under non-goals here

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: updated to clarify that E2E tests implemented in CI will verify the operations

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: SG, filled out the section - PTAL when you can!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: I started filling it out, but stopped because I felt that we need to figure out the metrics / events for this feature to properly answer the questions. I would prefer to leave that for beta if it is not a requirement now. 

For example, we already have a metric that records in-place pod resize operations; for beta of this KEP we will probably want to extend that metric to have a label i.e. `resource=memoryBackedEmptyDirVolume`, `operation=upsize/downsize`. Likewise we currently have an event that emits the allocation checkpoint for container/pod resources when a resize is started/completed/errors out; we'd probably want to update that to also include the empty dir volume size. Once those details are figured out, operators/users could use those to determine whether the feature is working.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
kannon92: Looks great. Thank you.

tallclair (commented):
/lgtm
/approve

Looks good for SIG-Node.

natasha41575:
@jsafrane @msau42 PTAL when you can, thanks!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+
+Only modifications to an existing `Pod.Spec.Volumes[].EmptyDir.SizeLimit` are permitted.
+* The addition or removal of a memory-backed volume in a pod spec is forbidden. 
+* For Alpha, adding or removing a `sizeLimit` to an existing memory-backed volume is not allowed. This boundary is re-evaluated for Beta.
+
+#### Resize Restart Policy 
+
+There are existing resize restart policies that control when a container is restarted during a resize operation, which are described in more detail in the [InPlacePodVerticalScaling KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#container-resize-policy).
+
+We do not introduce an analogous policy nor honor the existing restart policies when resizing memory-backed `emptyDir` volumes. The resizing of memory-backed `emptyDir` volumes is fundamentally non-disruptive to running processes and does not require application-level awareness beyond standard handling of filesystem capacity. If a volume resize is accompanied by a container or pod resource resize, the container's existing `RestartPolicy` governs container restarts.
```
SergeyKanzhelev, line 298: > If a volume resize is accompanied by a container or pod resource resize, the container's existing `RestartPolicy` governs container restarts.

Just to clarify, those cannot be combined together, those are two independent resizes, right?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+
+#### Atomic Resize Principle 
+
+The atomic resize principle is maintained. The Kubelet treats updates to volume `sizeLimit` and container / pod resources (both limits and requests) as a single, atomic transaction. The `AllocationManager` performs an All-or-Nothing Admission check, rejecting the entire update if the resize feasibility checks do not pass. The volume `sizeLimit` is not taken into consideration during the feasibility checks, but fails admission atomically if the container or pod resize fails to be admitted. The volume `sizeLimit` is also checkpointed atomically along with container and pod resources.
```
SergeyKanzhelev: clarify atomicity - is it on a level of individual volumes or I can resize two by increasing one and decreasing another for the same amount in a single operation?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+*   The `KuberuntimeManager` checks the direction of the volume resize (upsize vs. downsize) by comparing the newly allocated size with the checkpointed **Actuated State** (the target size that the Kubelet last attempted to apply, which is maintained in Kuberuntime's internal checkpoint). Comparing against the *actuated* state rather than the live kernel state avoids unnecessary remount requests caused by kernel rounding differences.
+* The `KuberuntimeManager` directly calls a new interface on the Volume Manager (see [Volume Manager interface](#volume-manager-interface)) to perform the remount at the correct time, bypassing the async reconciler loop. The Volume Manager executes the remount: `mount -o remount,size=<limit> -t tmpfs tmpfs <path>`. Once this actuation completes, the `KuberuntimeManager` checkpoints this sizeLimit in its own **Actuated State**. This Actuated State represents what the Kubelet tries to actuate, which may diverge slightly from what is read directly from the kernel (`Actual State`).
```
SergeyKanzhelev, line 356: @msau42 if the remount is generally supported, can it solve the in-place Pod restart problem when the certain mounts needs to be remounted (when sidecar actually back up this mount). Would you be open to add remount-on-restart for those so we can improve the gcfsd reliability?

SergeyKanzhelev: cc @yuanwang04

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes a mechanism to dynamically resize `emptyDir` volumes with `medium: Memory` (tmpfs) without requiring Pod recreation or Pod restart. It leverages the `InPlacePodVerticalScaling` (IPPVS) infrastructure to manage the transition from the desired state to the actual state via an allocation and actuation loop.
```
SergeyKanzhelev: please specify whether the expectation is to keep all existing files on resize. It is not clear from this paragraph

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+
+This design allows `KuberuntimeManager` to act as the central orchestrator for resource updates (ensuring strict ordering with cgroups) while leaving the actual filesystem manipulation logic encapsulated within the volume plugins.
+
+We considered alternatives such as relying on the Volume Reconciler to detect changes asynchronously or moving `emptyDir` handling entirely to `KuberuntimeManager`, but they are ruled out. See [Alternatives](#alternatives) for more details.
+
+### Shrinkage Safety 
+
+In the existing `InPlacePodVerticalScaling` feature, a best-effort safety logic ensures that container memory limits are not decreased below usage.
+
+For Alpha, handling the analogous case for memory-backed volumes is out of scope. If the user attempts to remount the volume with a smaller size than currently used, the mount update fails with an `EINVAL` error. In this case, the error is propagated back to the Kubelet and surfaced in the `PodResizeInProgress` condition. The error message from the kernel is `mount point not mounted or bad option`, which is surfaced to the user in the condition message.
```
SergeyKanzhelev: it is not clear from this whether kubelet will continue retrying this resize.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1029 @@
...
+* **Note**: Bypassing the async reconciler loop is both safe and inconsequential. The `emptyDir` reconciler today never remounts because `RequiresRemount` is hardcoded to always return false; this means that the `KuberuntimeManager` is the only component that makes this change. We considered alternatives such as relying on the `emptyDir` reconciler to detect changes asynchronously or moving `emptyDir` handling entirely to `KuberuntimeManager`, but they are ruled out. See [Alternatives](#alternatives) for more details.
+
+### 4. Observation & Feedback Loop
+*   Upon successful remount and cgroup update, the `KuberuntimeManager` checkpoints the new target values as the current 'actuated resources' (the baseline for future resizes).
+*   The Kubelet updates `Pod.Status.ContainerStatuses[].VolumeMounts[].VolumeStatus.EmptyDir.SizeLimit` to reflect the successful resize.
+*   Once the actual size matches the allocated size checkpoint (and container resource updates are complete), the Kubelet clears the `PodResizeInProgress` condition.
+
+### Resource Coordination: Volume and Cgroup Ordering of Updates
```
SergeyKanzhelev, line 364: do we need to report this change to the container runtime in any form? Please add a note

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: You may modify all the fields in the same resize request. In the same request, user can modify memory, cpu, and volume size.

A container will be restarted if either:
- the resize request modifies cpu reqs/limits, and the cpu restart policy is `RestartContainer`
- the resize request modifies memory reqs/limits, and the memory restart policy is `RestartContainer`

A volume resize on its own never triggers a container restart. This part of the KEP is trying to communicate that the existing `RestartPolicy` field does not interact with the volume resize change in any way.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: You may increase one and decrease another in a single operation. The kubelet will treat it atomically. 

I will add this clarification to the KEP

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Yeah the goal is to preserve all the existing files stored in the volume mount. I will update this (I'll also add it to the goals)

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: sorry, the sentence at the bottom of this section was meant to apply to both alpha-and-beta stages. I will update the wording to clarify that kubelet will continue to retry on every pod sync

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: I don't think container runtime notifications are applicable in this scenario. My understanding is that on startup, the container runtime helps to bind the Kubelet's host path into the container's mount namespace, but subsequent `remount` requests do not require intervention from the container runtime.

For in-place pod resize, we issue a CRI notification via UpdatePodSandboxResources because the container runtime (especially sandbox runtimes like gvisor or kata containers) might be interested to know how the pod sandbox is being resized, but I don't have a use case why the runtime would need to be made aware of changing volume sizes.

SergeyKanzhelev: add an explicit note about it in the KEP please.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
SergeyKanzhelev: The pod resize is done using subresource, while volume resize via the pod spec. Do you mean this will be recieved by the kubelet as a single "new pod definition"? From API perspective it cannot be atomic, correct?

natasha41575: The proposal is to allow mem-backed volume resize to be done via the /resize subresource. 

The initial use case for the feature request we received was that memory-backed volumes needed to be resized alongside memory resources so that the volumes can be set to a specific proportion of the pod's total memory. 

In general, I think it is desirable to allow users to tie memory-backed volume resize to memory req/limit changes. Usage of mem-backed volumes is charged against cgroup values, so there are likely many use cases to change both in a single update request.

ETA: I'm not proposing that all volume resizes be done via the /resize subresource, but for specifically mem-backed volumes, the volume is considered part of the pod's memory, so it makes sense here.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
yuanwang04: Agreed it seems a very useful extension for Volume Manager

yuanwang04: If remount is part of the volume resize operation, would that be disruptive for the application container? Or is the expectation for the application container to handle those remount gracefully?

natasha41575: Specifically for memory-backed volumes, the remount operation is non-disruptive. The application does not need to specifically handle it.

I can't speak for other volume types.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
SergeyKanzhelev: I just read in the beginning of the KEP that the size will be mutable. Perhaps it was a leftover then? I think extending the resize is better

natasha41575: Oh I see the misunderstanding. Ok, I will update that sentence to clarify the size will be mutable only under the resize subresource. 

Originally I had only put this detail under the api validation section but totally see how this could be easily missed.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: added!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+
+```go
+type VolumeStatus struct {
+    ...
+    // emptyDir represents the status of an emptyDir volume.
+    // +optional
+    EmptyDir *EmptyDirVolumeStatus `json:"emptyDir,omitempty" protobuf:"bytes,2,opt,name=emptyDir"`
+}
+type EmptyDirVolumeStatus struct {
+    // sizeLimit represents the actual mounted capacity of the emptyDir volume.
```
msau42, line 277: In the comments it would be good to say that it's restricted to memory-backed emptydir

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+
+Only modifications to an existing `Pod.Spec.Volumes[].EmptyDir.SizeLimit` are permitted.
+* The addition or removal of a memory-backed volume in a pod spec is forbidden. 
+* For Alpha, adding or removing a `sizeLimit` to an existing memory-backed volume is not allowed. This boundary is re-evaluated for Beta.
+
+#### Resize Restart Policy 
+
+There are existing resize restart policies that control when a container is restarted during a resize operation, which are described in more detail in the [InPlacePodVerticalScaling KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources#container-resize-policy).
+
+We do not introduce an analogous policy nor honor the existing restart policies when resizing memory-backed `emptyDir` volumes. The resizing of memory-backed `emptyDir` volumes is fundamentally non-disruptive to running processes and does not require application-level awareness beyond standard handling of filesystem capacity. If a volume resize is accompanied by a container or pod resource resize, the container's existing `RestartPolicy` governs container restarts.
```
msau42, line 298: Is there a way to notify the application that the resize is complete? Some applications may need to invoke additional steps to take advantage of the larger size.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+### API Changes
+
+#### New Status Fields and State Mapping
+
+To support dynamic resizing, the `Pod.Spec.Volumes[].EmptyDir.SizeLimit` field is made **mutable** when modified via the `/resize` subresource for existing Pods (when `medium` is `Memory`). This represents the **Desired State**.
+
+We introduce a new sub-structure under `corev1.VolumeStatus` (in `Pod.Status`) to track the actual resize state of the volume:
+
+*   **Desired State**: Represented by the mutable `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
+*   **Allocated State**: Checkpointed and maintained by the Kubelet's `AllocationManager` for local resource accounting, but **not exposed** in `Pod.Status`.
```
msau42, line 256: I'm not familiar with Allocation manager. Does this reflect in pod scheduling or quota decisions?

For example, resizing Pod A is in progress but hasn't finished yet. Now Pod B is also sends a resize request but it would put it over the allocatable if you also added the pending Pod A's request. When do we reject Pod B?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
msau42: It may help but it would be good to do a poc here for the fuse sidecar scenario.

This kep in particular avoids the concurrency/coordination problem by restricting this to only emptyDir types.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+1. **Init Execution**: During execution, files written by the Init container count against the **Init container's individual limit**.
+2. **Termination**: When the Init container exits and its cgroup is cleaned up, the files remain in the memory volume. The memory usage of these files continues to count against the overall **Pod cgroup memory limit**.
+3. **App Container Handoff**: When the primary application container starts, it can read the pre-populated data. Under standard memory accounting rules, simply reading pre-existing files does not charge that memory to the app container's individual limit. The app container can use the cache without it counting toward its own limit (it only consumes Pod-level cgroup headroom).
+ 
+#### Priority of Enforcement
+
+The volume `sizeLimit` acts as a safety "inner boundary" to ensure volume growth does not starve the container's processes of execution RAM. Ideally, `sizeLimit` is lower than the container's memory limit.
+
+Kubernetes does not strictly enforce this "lower-than" relationship, allowing users to set a `sizeLimit` exceeding the pod's memory limits or shrink pod memory limits below the volume `sizeLimit`. In these cases:
```
msau42, line 401: Is this to support the scenario where no Pod memory limit is set?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
msau42, line 256: Another question about Allocation manager, does it handle scenarios like "upsize then downsize" very quickly before the first operation has completed?

For the general volume expansion feature, we had a lot of challenges supporting the rollback scenario and ensuring that we properly charged for increases in progress despite the desired state being rolled back.

natasha41575: kubelet is running two loops, independent of each other, running in separate goroutines:

1. Allocation manager, which compares desired (spec) to allocated (from its allocation checkpoint). If they differ, it checks whether the resize is 'feasible' based on the node's allocatable resources; if so, it allocates the desired spec and updates the allocation checkpoint. For the feasibility check, kubelet assumes the resources of other pods on the node using max(allocated, actual); then it checks whether the desired resources of the new pod (pod B in this case) can fit against those.
2. The SyncPod loop, which calls out to the kuberuntime manager to compare allocated (from the allocation checkpoint) against the last actuated resources. SyncPod / kuberuntime manager are only ever aware of the allocated pod, and do not ever see the 'desired' resources from the podspec.

So in your scenario, Pod A, which presumably has already been 'allocated' by the allocation manager, is being processed by SyncPod, would continue and complete independently of the Pod B (infeasible) resize request. The allocation manager would independently evaluate and reject Pod B without SyncPod ever being aware of it.

Allocated resources are surfaced to scheduling in that the scheduler assumes the pod resource footprint as max(desired, allocated, actual). This is for normal cpu/mem resources; volume size is not taken into account there.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: > Kubernetes does not strictly enforce this "lower-than" relationship, allowing users to set a `sizeLimit` exceeding the pod's memory limits or shrink pod memory limits below the volume `sizeLimit`.

This is the behavior of kubernetes today. I assume it is a remnant of not having support for pod-level resource prior. We don't propose to change the existing sizeLimit restrictions in this KEP but we may want to consider it in the scope of the pod-level resources project.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: > does it handle scenarios like "upsize then downsize" very quickly before the first operation has completed?

I explained a little further how the allocation manager / sync pod interaction works in https://github.com/kubernetes/enhancements/pull/6050#discussion_r3363689206.

For "upsize then downsize" scenario, what would happen:

- Allocation manager allocates the 'upsize'; then adds an update to the channel that triggers a SyncPod iteration to occur.
- SyncPod calls out to kuberuntime manager to actuate the resize.
- If the 'downsize' request is then called while SyncPod is still in progress, the allocation manager will 'allocate' the downsize, but SyncPod will still continue and complete its in-progress iteration.
- Then SyncPod would be invoked again to actuate the 'downsize'. 

It would all happen very quickly - no delays between steps. But once SyncPod starts the upsize, it would try to complete it before seeing the new allocation.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: done.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: AFAIK kubelet and the container runtime don't have a mechanism to send direct notification signals to running applications. I don't know if this use case alone warrants it, but maybe we can consider it in the future if such requests come up?

Applications can read the volume size from standard filesystem queries (e.g. `statfs`) to dynamically adjust to filesystem changes. This is architecturally consistent with in-place pod resize; the applications are not notified with the cgroup values have changed, but they can read the cgroup values from the filesystem and dynamically adjust from that.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
msau42: > Allocated resources are surfaced to scheduling in that the scheduler assumes the pod resource footprint as max(desired, allocated, actual). This is for normal cpu/mem resources; volume size is not taken into account there.

If volume size is not accounted for by the scheduler, does that mean the node could have less memory available then what it reports? So the scheduler can schedule pods to nodes but then just get rejected by kubelet? Or if it doesn't get rejected by kubelet, then could there be unaccounted for memory pressure?

natasha41575: When a memory-backed volume gets written to, the memory gets charged against *both* the volume size limit _and also_ the pod/container memory limits. So taking into account memory limits alone covers it, and there's no need to take the volume size limit into account when determining memory fit.

Both scheduler and kubelet ignore it (they share the same resource fit code).

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Do you think there is a need to have a restart policy for volume resizes? i.e. an API field to tell kubelet to restart the container upon doing the resize?

We have this option for cpu and memory resources because there are some applications that cannot take advantage of the larger cgroups without a restart. I wasn't aware of analogous use cases for volumes, but we could add something like that here if there are any.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
msau42: Ok so memory usage is accounted for under the pod limit. If the user grows their emptydir usage beyond the pod limit but within emptydir limit they will get limited by cgroup limits.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
msau42: I don't have any concrete use cases that would need it but we can keep that idea open as we get more feedback on the feature.

msau42 (commented):
/approve

from sig-storage

SergeyKanzhelev:
/assign @dchen1107

(as per kep.yaml)

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+* **Pod-level charging**: Since container cgroups are nested under the parent Pod cgroup, all files written to the memory-backed volume also count against the **Pod-level cgroup memory limit**.
+
+##### Sharing Volumes Between Multiple Containers
+
+When a single memory-backed `emptyDir` volume is mounted and shared by multiple containers under the same Pod:
+* **Volume Quota (`ENOSPC`)**: The volume-level `sizeLimit` acts as a hard boundary on the *aggregate* space consumed. If the sum of bytes written across all sharing containers exceeds this limit, standard disk-full (`ENOSPC`) errors are returned to subsequent writes. The containers are not OOM-killed at this boundary.
+* **cgroup Enforcement (OOM-Kill)**: If a container's write to the shared volume pushes its total memory usage (processes + files written) past its cgroup memory limit, the container's processes are **OOM-killed** by the kernel—even if the shared volume usage is well below its `sizeLimit`.
```
dchen1107, line 387: AFAIK this statement is only true for the very first container writing data to shared volume. In Linux, tmpfs pages reside in the global shmem/page cache. In cgroup v2, when container A writes a 500MiB file to the shared volume, those pages are charged to container A's cgroup initially. However, if container A finishes writing and container B later triggers certain page-fault operations on that exact same file, the kernel's active memory controller can dynamically shift or split the page cache attribution between the two container cgroups depending on memory pressure.

I personally viewed this as one of the huge enhancements from cgroup v1 to v2, but such enhancement makes container level estimation less predictable, and could potentially create a blindspot for kubelet and K8s VPA controller due to the memory charge migration. Can we explicitly document this limitation caused by kernel behavior, and suggest AllocationManager or resizer must look at the overall pod level cgroup too when calculating if an in-place downsize or upsize is safe to execute?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+This enhancement builds directly on the foundation laid by **In-Place Pod Vertical Scaling (IPPVS)** (KEP-1287). Just as IPPVS allows scaling container CPU and memory resources without Pod disruption, high-performance workloads frequently require scaling their scratch space or in-memory caches simultaneously. Since memory-backed `emptyDir` volumes consume the node's memory (and can be bounded by the Pod's memory limits), integrating their resize lifecycle with the IPPVS allocation and actuation phases ensures consistent resource accounting and prevents node exhaustion.
+
+### Goals
+
+*   Allow updating `emptyDir.sizeLimit` for existing Pods when `medium` is `Memory`.
+*   Reflect resize progress in `Pod.Status` using an allocation/actuation model similar to IPPVS.
+*   Actuate the resize via `tmpfs` remount without Pod disruption and while maintaining the files already written to the volume's filesystem.
+
+### Non-Goals
+
```
dchen1107, line 218: Can explicitly call out cgroup v2 only for the feature? See my comment: https://github.com/kubernetes/enhancements/pull/6050/changes#r3383088518 for more details?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -0,0 +1,1036 @@
...
+* **Persistence of Orphaned Space**: If a container is terminated or OOM-killed, the files it wrote remain in host memory (tmpfs survives container lifecycles). This memory continues to consume headroom in the parent Pod cgroup.
+
+##### Init Container Lifecycles & Handoff
+
+Init containers are commonly used to pre-populate a memory-backed volume (e.g., pre-loading templates, caches, or datasets) before primary application containers start:
+1. **Init Execution**: During execution, files written by the Init container count against the **Init container's individual limit**.
+2. **Termination**: When the Init container exits and its cgroup is cleaned up, the files remain in the memory volume. The memory usage of these files continues to count against the overall **Pod cgroup memory limit**.
+3. **App Container Handoff**: When the primary application container starts, it can read the pre-populated data. Under standard memory accounting rules, simply reading pre-existing files does not charge that memory to the app container's individual limit. The app container can use the cache without it counting toward its own limit (it only consumes Pod-level cgroup headroom).
```
dchen1107, line 395: What about the user uses an in-place update to shrink both the app container's memory limit and the emptyDir.sizeLimit in this scenario?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
dchen1107: +1.

dchen1107:
I have some open questions but can live with them later. 

/lgtm
/approve

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Thank you for correcting me. I updated this section to document this limitation. 

> suggest AllocationManager or resizer must look at the overall pod level cgroup too when calculating if an in-place downsize or upsize is safe to execute?

kubelet already checks both container memory usage against container memory limits and also aggregate pod memory usage against pod limits. So I think we are covered on that end, but I added a note about that to the KEP too.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: only supporting cgroup v2 for this feature SGTM. I added to the KEP that if kubelet sees this resize request on a node with cgroup v1, it will mark the resize as `Infeasible`.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: This is handled by our actuation ordering (volume first). If the volume shrink fails (usage > limit), we abort the transaction before shrinking cgroups. 

I also noted above that Kubelet's existing checks will also validate that cgroup limits aren't shrunk below aggregate Pod usage. I updated the "Shrinkage Safety" section to clarify these. 

These two together I think should make the downsize safe. Let me know if it answers your questions.

Issue #6051, opened 2026-05-01
Add stream-time authorization hook to CRI streaming server


mattsu2020:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add an optional `AuthorizeStream` hook to the CRI streaming server config for stream-time authorization of exec, attach, and port-forward requests
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6052
- Discussion Link: https://github.com/kubernetes/kubernetes/pull/138616
- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6052
    - [ ] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/138616
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

PR #6052, opened 2026-05-01
KEP-6051: CRI Streaming Authorization Hook

mattsu2020:
## Summary

Adds a KEP proposing an optional `AuthorizeStream` hook to the CRI streaming
server configuration (`k8s.io/cri-streaming`). The hook is invoked after a
short-lived stream URL token is validated and before exec, attach, or
port-forward streams are served.

Enhancement issue: kubernetes/enhancements#6051
Code PR: kubernetes/kubernetes#138616

## Motivation

The CRI streaming package currently relies on short-lived, single-use stream
URLs issued after the original CRI request has been authorized. This KEP adds an
explicit stream-time authorization point for CRI runtimes that need additional
checks at the streaming server boundary.

CRI runtime implementers can use the hook for:

- Defense in depth at the CRI shim boundary
- Audit logging for compliance
- Custom runtime-specific authorization policies

## Key Design Decisions

- Add `AuthorizeStream func(*http.Request, any) error` to `streaming.Config`
- Pass the cached CRI stream request (`ExecRequest`, `AttachRequest`, or
  `PortForwardRequest`) to the hook so implementations can authorize the target
  stream
- Preserve existing behavior when `AuthorizeStream` is nil
- Keep the feature opt-in for CRI streaming embedders rather than adding a
  kubelet feature gate
- Call the hook after token validation and before stream serving
- Reject denied requests with `403 Forbidden`

## Validation

- `GOCACHE=/private/tmp/go-build-cache make verify-toc`
- `GOCACHE=/private/tmp/go-build-cache make verify-metadata`

## SIG Node Discussion

Initial feedback from @mikebrow on kubernetes/kubernetes#138616 requested a KEP
with solid use cases and comprehensive tests. This KEP documents the use cases,
the opt-in API contract, and the test plan for nil, allow, deny, invalid-token,
and cached request type coverage.

### DIFF
--- a/keps/sig-node/6051-cri-streaming-authorization/README.md
+++ b/keps/sig-node/6051-cri-streaming-authorization/README.md
@@ -1,0 +1,459 @@
+# KEP-6051: CRI Streaming Authorization Hook
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Defense in Depth](#story-1-defense-in-depth)
+    - [Story 2: Audit Logging for Compliance](#story-2-audit-logging-for-compliance)
+    - [Story 3: Custom Runtime Authorization](#story-3-custom-runtime-authorization)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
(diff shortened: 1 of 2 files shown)


aojea:
I'm unclear on how this works with the current kubernetes setup, IIRC the decision to split the streaming bits from cri was to simplify dependencies. 
This proposal seems that is trying to make this module generic and build a library or at least I miss how the kubelet will use this new hook

mattsu2020:
> I'm unclear on how this works with the current kubernetes setup, IIRC the decision to split the streaming bits from cri was to simplify dependencies. This proposal seems that is trying to make this module generic and build a library or at least I miss how the kubelet will use this new hook

Thanks, that is a fair point.

My intent was to add a stream-time authorization point for CRI streaming server embedders, but I agree the current proposal does not clearly show how this fits into the current kubelet/CRI flow. In the current Kubernetes setup, kubelet authorizes the original exec/attach/port-forward request and then proxies to the URL returned by the runtime; kubelet does not directly construct the cri-streaming server config in the usual path.

Given that, the proposal is currently underspecified. I need to either describe a concrete kubelet-facing integration, including what identity/authorization context would be propagated to the runtime stream server, or narrow/withdraw the proposal if this is only a generic library hook without an in-tree Kubernetes use case.

I will revisit the KEP with that in mind.

PR #6054, opened 2026-05-01
KEP-6053: Kubelet Image Garbage Collection Observability

mattsu2020:
### What changed

Adds a new SIG Node KEP for improving kubelet Image Garbage Collection observability.

The KEP proposes low-cardinality kubelet metrics for Image GC attempts, duration, reclaim targets, estimated reclaimed bytes, image deletion outcomes, skipped images, image filesystem usage, and failure/no-op outcomes.

### Why this is needed

Kubelet currently exposes `kubelet_image_garbage_collected_total{reason}`, which only counts successfully removed images. Operators cannot reliably tell whether Image GC is running, failing, doing no work, running slowly, or reclaiming less space than requested without inspecting kubelet logs node by node.

### Notes

This proposal is passive observability only. It does not change Image GC policy, kubelet configuration, Kubernetes APIs, CRI APIs, pod behavior, or eviction behavior.

The KEP preserves the existing `kubelet_image_garbage_collected_total{reason}` label values while adding richer Image GC deletion outcome metrics. It also calls out that skipped-image accounting needs to cover images filtered before deletion ordering, such as in-use or runtime-pinned images.

### Validation

- `make update-toc`
- `make verify-toc`
- `make verify-metadata`

### DIFF
--- a/keps/sig-node/6053-kubelet-image-gc-observability/README.md
+++ b/keps/sig-node/6053-kubelet-image-gc-observability/README.md
@@ -1,0 +1,562 @@
+# KEP-6053: Kubelet Image Garbage Collection Observability
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Metrics](#metrics)
+  - [Trigger Attribution](#trigger-attribution)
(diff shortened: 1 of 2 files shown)


mattsu2020:
Hi reviewers, this PR is ready for review.
Could someone please take a look and add /ok-to-test if the patch looks safe to test?

This is a KEP-only PR under sig-node, adding passive kubelet Image GC observability metrics.
It does not change kubelet behavior, APIs, CRI APIs, feature gates, or eviction behavior.

Issue #6058, opened 2026-05-05
CSI global mount fallback for volume reconstruction


cniackz:
### Enhancement Description

- One-line enhancement description (can be used as a release note): When CSI volume reconstruction at kubelet restart fails to load the pod-local `vol_data.json`, fall back to the equivalent file under the global mount directory to prevent orphaned global mounts that can lead to data corruption on RWO volumes.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6059 ; see implementation PR https://github.com/kubernetes/kubernetes/pull/138454 and issue https://github.com/kubernetes/kubernetes/issues/101791
- Discussion Link: tracked in https://github.com/kubernetes/kubernetes/issues/101791 since 2021
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6059
    - [x] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/138454
    - [ ] Docs (`k/website`) update PR(s):

This extends KEP-3756 (Robust VolumeManager reconstruction after kubelet restart) by replacing its documented manual cleanup procedure for orphaned global mounts with automatic recovery, gated behind a new `VolumeReconstructionFallback` alpha feature gate.

/sig storage
/kind feature

kannon92:
Please add #6059 to the description.

karimzakzouk:
Hello @cniackz 👋, v1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for v1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
[...]

PR #6059, opened 2026-05-05
KEP-6058: CSI volume reconstruction fallback

cniackz:
Adds KEP-6058, the design proposal backing the implementation in [kubernetes/kubernetes#138454](https://github.com/kubernetes/kubernetes/pull/138454). When CSI volume reconstruction at kubelet restart fails to load the pod-local vol_data.json, the new VolumeReconstructionFallback alpha gate (default off) lets kubelet fall back to the equivalent file under the global mount directory, preventing the orphaned global mounts that today require the manual operator cleanup documented in KEP-3756. Tracking issue: #6058. Targeting alpha in v1.38. /cc @jsafrane @msau42 @gnufied


cniackz:
Hi @msau42 @jsafrane, can one of you give /ok-to-test so the prow checks can run? thanks!

kannon92:
You need to request PRR review.

You can do so by creating a file for 6048 [here](https://github.com/kubernetes/enhancements/tree/master/keps/prod-readiness/sig-storage).

And you can add me as author.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,491 @@
...
+- Gate the new behavior behind an alpha feature gate, default off, so it can
+  be disabled at any release.
+- Cover the recovery path with a unit test (added in the implementation PR)
+  and a node e2e test (planned for beta).
+
+### Non-Goals
+
+- In-tree volume plugins (FibreChannel, iSCSI, NFS, etc.) that derive state
+  from `/proc/mounts` rather than a state file. Those plugins are being
+  migrated to CSI; addressing them in-tree is out of scope.
```
jsafrane: They're not migrated to CSI, they will stay in-tree until someone makes a compatible replacement.

NFS does not use MountDevice, so it is not affected.

Both iSCSI and FC could benefit from this KEP. I'd suggest rephrasing it that FC and iSCSI are out of scope, however, someone may implement support for the global mount reconstruction in a similar way as the CSI volume plugin.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,491 @@
...
+## Motivation
+
+Issue [#101791][] has tracked the orphaned-global-mount class of bugs since
+2021. The root cause is that pod-local `vol_data.json` is the only source of
+truth used for reconstruction; any condition that loses or corrupts it
+(disk full at write time, partial write during shutdown, manual deletion by an
+operator chasing a different problem, filesystem corruption) leaves the global
+mount live with no in-memory record. The attach/detach controller then sees a
+`volumesInUse` set that does not reflect reality and proceeds to attach
+elsewhere.
```
jsafrane, line 111: There is one root cause not covered above: `UnmountDevice()` / `UnstageVolume()` is in progress while kubelet restarts (or node reboots). If that happens, the Pod mount is already removed. A newly started kubelet does not see it and thus it won't call `ConstructVolumeSpec()` in the CSI volume plugin that this KEP plans to update.

See e.g. https://github.com/kubernetes/kubernetes/issues/121937

Simple use case: node drain + reboot when the storage network has some hiccup.
Node drain waits for all Pods to get deleted. Kubelet allows a Pod to get deleted after `TearDown()` / `NodeUnpublish()` succeeds on all volumes on the Pod. But kubelet does not wait for `UnmountDevice` / `Unstage` . If the storage network is unhealthy, kubelet may declare the node drained before the global mount is removed.

I am afraid any fix of this issue will need to work without the pod mount directory existing. There should be a scan of `/var/lib/kubelet/plugins/kubernetes.io/csi/*/*` to find all global mounts and their jsons, independent from scanning `/var/lib/kubelet/pods`.

cniackz: Hello @jsafrane

you are right, scanning only /var/lib/kubelet/pods misses global mounts where the pod directory is already gone, like when unstage is still in progress during a node drain and reboot, and I will update the design to scan /var/lib/kubelet/plugins/kubernetes.io/csi/*/* directly, independent of the pod directory. For alpha, do you want the fallback to also clean up orphaned global mounts that have no pod?

Thank you

jsafrane: > For alpha, do you want the fallback to also clean up orphaned global mounts that have no pod?

I am not sure I understand. At the time of reconstruction, I'd add the global mount to ASW as uncertain and let the volume manager figure out if the directory is needed or not. It should call NodeUnstage before removing the directory.

Maybe only when the directory is empty, then remove it directly, because there is no json to reconstruct the volume.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Hello @jsafrane,

I rephrased it as you suggested. The Non-Goals section now lists FC and iSCSI as out of scope, with a note that the same global mount reconstruction could be implemented for them too, similar to what we did for the CSI volume plugin.

Thank you

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Hello @jsafrane,

That approach makes sense to me. I updated the design to follow it: reconstruction now scans /var/lib/kubelet/plugins/kubernetes.io/csi/*/* independent of the pod directories, adds each global mount found to the ActualStateOfWorld as uncertain, and lets the volume manager call NodeUnstage before removing the directory. If a directory is empty, it gets removed directly, since there is no json to reconstruct the volume from. The changes are in the Proposal (new scan section), Design Details (item 4), User Stories (Story 3, the drain plus reboot case), and now the Test Plan also covers the orphaned global mount path.

Thank you

cniackz:
Hello @jsafrane,

I pushed the design update we discussed, the reconstruction pass now also scans the CSI plugins directory for global mounts whose pod directory is gone, registers them as uncertain in the ActualStateOfWorld, and lets the volume manager call NodeUnstage before removing the directory (empty directories are removed directly). I also rebased on master. Let me know when you have time to take another look.

Thank you

cniackz:
Hello @kannon92,

Done: I added keps/prod-readiness/sig-storage/6058.yaml with you as the PRR approver for alpha, and I also added you as an author in kep.yaml.

Thank you

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+The fallback is wrapped by `utilfeature.DefaultFeatureGate.Enabled(features.VolumeReconstructionFallback)`.
+With the gate off, `ConstructVolumeSpec` behaves exactly as before. With the
+gate on, only the failure path is altered: the success path is unchanged.
+
+The fallback above still starts from a pod directory. To also cover global
+mounts whose pod directory no longer exists, the reconstruction pass gains a
+second, independent source of candidates: a scan of
+`/var/lib/kubelet/plugins/kubernetes.io/csi/*/*`, the parent directories of
+every CSI global mount on the node. For each directory found that is not
+already tracked in the ActualStateOfWorld:
```
jsafrane: I would call this as the main point of the KEP and not a fallback. IMHO it's far more common than corrupted pod-local json because of OOM or cluster admin removing it.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+mount's `vol_data.json`, succeeds, and the volume goes through the normal
+unmount path on both layers. No data corruption.
+
+#### Story 2: operator deletes a pod-local vol_data.json by mistake
+
+An operator investigating a stuck pod deletes the pod's volume directory
+contents. Without the fallback, the symptom is the same as Story 1. With the
+fallback, kubelet recovers automatically.
+
+#### Story 3: node drain and reboot while unstage is in progress
```
jsafrane: I think this is far more common than OOM or cluster admin deleting the json file. And the main reason behind this KEP.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+The information needed to rebuild the spec is already available on disk: the
+global mount's `vol_data.json`, written by `csiAttacher.MountDevice`,
+already stores `volumeHandle` and `driverName`. With two more fields it is
+self-sufficient. Using it as a fallback turns a manual-recovery path into an
+automatic one without changing any contract with CSI drivers.
+
+[#101791]: https://github.com/kubernetes/kubernetes/issues/101791
+[#121937]: https://github.com/kubernetes/kubernetes/issues/121937
+
+### Goals
```
jsafrane, line 120: Please add support for block volumes. Or describe why it can't be done and add it to non-goals.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+mounts: a failed pod-local load now falls back to the global file, and a
+global mount left behind by an interrupted unstage is now unstaged cleanly
+instead of leaking.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Disabling the gate restores the previous behavior on the next kubelet
+restart. There is no on-disk state migration. Any global `vol_data.json`
+files written with the extra fields remain valid and are simply ignored by
+the disabled code path.
```
jsafrane: Note that they're deleted by kubelet when it cleans up the  global mount.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+###### How can an operator determine if the feature is in use by workloads?
+
+At alpha: kubelet logs at V(2) emit `plugin.ConstructVolumeSpec recovered
+vol_data from global mount %s` whenever the fallback fires,
+and a similar V(2) line when the plugins directory scan registers an
+orphaned global mount as uncertain and when its directory is removed after
+a successful unstage.
+
+At beta: a label on the existing `reconstruct_volume_operations_total` metric
+distinguishing `pod-local` vs `global-fallback`.
```
jsafrane: mention that the metric exists and is alpha. We plan to add a new label to it as part of this KEP.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,695 @@
...
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+Marginally, and only on the affected node. `node.status.volumesInUse` gains one
+entry per orphaned global mount the scan recovers, for as long as the reconciler
+takes to unstage it. The count is bounded by the global mounts that outlived a
+kubelet restart, which on a healthy node is zero.
```
jsafrane: We do want all volumes that are mounted in `volumesInUse`. Increased Node size is a bugfix, not some unfortunate side effect of this KEP. Can you rephrase it to be clearer?

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: You are right, and the document argued it the other way round. The scan is now what the KEP leads with: Summary, Motivation, Goals, Proposal and the first user story are the drain and reboot, Design Details puts the scan second right after the field it needs, and the gate is renamed to `CSIGlobalMountReconstruction`. What I have not solved is ownership. The scan's implementation is #136771, opened by @shivamwayal37 in February against #121937 and reviewed over four rounds, and I would rather not duplicate it, so alpha here currently scopes to the two paths that have code and tests and the scan sits in beta. I do not think that split is right if the scan is the point, and the gate ends up on the wrong half of the feature: the scan is what registers volumes nobody asked about and unstages them, while the pod-local fallback only changes an error path. Would you help bring @shivamwayal37 and #136771 under this KEP, as a co-author, with the gate added to his branch? That is a small diff and it settles ownership, gate placement and the beta soak in one move. If it cannot be arranged before the freeze I would rather move alpha to v1.39 with the full scope than ship the half that is not the point.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Agreed, and it is Story 1 now. It reads as the case that needs no corruption and no operator mistake: kubelet lets the pod go once `NodeUnpublishVolume` succeeds, the drain completes on that basis, and the global mount survives the reboot with its pod directory already gone.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Block volumes keep their `vol_data.json` at `plugins/kubernetes.io/csi/volumeDevices/<specVolID>/data`, already node-global, and both `ConstructBlockVolumeSpec` and `NewBlockVolumeUnmapper` read it from there, so the pod-local failure mode cannot happen for them; their pod-local artifact is a symlink rather than a bind mount, so there is no mount reference to follow back either. That is in Non-Goals now. The staging path they can still leak is the scan's territory, so block travels with the scan rather than separately. One related finding while I was in there: the glob as written matches `volumeDevices` too, whose volume data sits one level deeper, so it would report every block volume on the node as a directory with no readable volume data and ask an operator to look at it. That is written into the design.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Added, thank you. The answer now says kubelet deletes that file when it cleans up the global mount: `UnmountDevice` calls `removeMountDir` after a successful `NodeUnstageVolume`, which removes the `globalmount` directory, the `vol_data.json` beside it and the volume directory itself, so the two fields live exactly as long as the mount they describe.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Corrected. It now says the metric already exists in kubelet at ALPHA stability and is an unlabelled counter today, incremented once per pod volume directory reconstruction attempt, and that this KEP adds a label distinguishing `pod-local` from `global-mount`. Worth noting that this turns a `NewCounter` into a `NewCounterVec`, so it is a change across its callers rather than a new value on an existing label; its alpha stability allows that without a deprecation cycle.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Rephrased with your framing. It now says the growth is the fix rather than a cost of it: `volumesInUse` is what the attach/detach controller reads to know the node still holds a device, `processVolumesInUse` copies it in as `MountedByNode` and the detach reconciler skips anything carrying that flag unless a force detach or the `node.kubernetes.io/out-of-service` taint overrides it, so an orphaned global mount being mounted and not listed is the defect, and is exactly what lets the controller attach it elsewhere.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: You are right and I had it backwards: the non-goal justified itself by saying both were migrating to CSI. It now says they stay in-tree until someone writes a compatible replacement, that both could benefit from the same recovery and someone may implement global mount reconstruction for them in a similar way to the CSI plugin, and that NFS does not use `MountDevice` and is unaffected.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation, e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
```
jsafrane, line 63: Most of this summary is repeated in Motivation below.

Summary should show how the proposed feature is useful, here for a cluster admin. Don't dive into detals, mention the admin visible impact - right now, after kubelet restart or node reboot, kubelet may declare volumes as unused while they're still mounted. That may result in volume corruption. We're trying to fix that by better scanning of mounted volumes and finish volume volume unmount, preventing such corruptions (yay, everyone wants that, rigjht?)

How exactly will we improve the scan is for subsequent chapters.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+prescribed manual operator cleanup. This KEP closes it too, by making the
+global mount's own `vol_data.json` self-sufficient: `MountDevice` already
+writes `volumeHandle` and `driverName` there, and this KEP adds `specVolID` and
+`volumeLifecycleMode`, so reconstruction and unmounting can both fall back to
+that file when the pod-local copy cannot be loaded.
+
+Both recovery paths sit behind one new alpha feature gate,
+`CSIGlobalMountReconstruction`, default off.
+
+## Motivation
```
jsafrane, line 79: I think this chapter is good - shows exact problem that we have + linking bugs. It does not need to be that verbose, but I can live with that.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+- Gate the new behavior behind an alpha feature gate, default off.
+
+### Non-Goals
+
+- In-tree FibreChannel and iSCSI. They stay in-tree until someone writes a
+  compatible replacement, and they derive their state from `/proc/mounts`
+  rather than from a state file, so this failure mode takes a different shape
+  there. Both could benefit from the same recovery, and someone may implement
+  global mount reconstruction for them in a similar way to the CSI plugin, but
+  it is out of scope here. NFS does not use `MountDevice` and is unaffected.
```
jsafrane: Yes, iSCSI and FC should be a non-goal, however, the overall architecture in kublet should be IMO prepared for them. So when I want to implement reconstruction for FC, I just implement a volume plugin interface (or functions) and it will fit the existing kubelet code perfectly.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+The scan starts from the plugin directory, so it cannot help the case where the
+pod directory is still present and it is the pod-local `vol_data.json` that is
+unreadable. For that, `ConstructVolumeSpec` is changed to fall back to the
+global file when the pod-local load fails. It asks the mount table which global mount the pod-local
+mount is bound to: `NodePublishVolume` publishes the staged volume into the pod
+directory, which drivers implement as a bind mount from the staging path, so
+the global mount is a mount reference of `<mountPath>/mount` and the data
+directory is its parent. If no global mount is bind mounted there, or its
+`vol_data.json` names no driver and no volume handle, reconstruction fails as
+today.
```
jsafrane: Looking for a bind-mount works after kubelet restart. What about node reboot? The mount table is empty and the pod local json is corrupt. What happens in this case? What should happen?

`ConstructVolumeSpec()` will probably fail - there is not enough data to reconstruct. Global recustruction succeeds and kubelet calls NodeUnstage, without NodeUnpublish. That's much better than today! Still... From CSI spec perspective, kubelet should call NodeUpublish too.

The pod local directory has PV name in it. Can we add PV name to the global `vol_data.json` and match by this instead of the mount table?

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+
+The fallback is wrapped by `utilfeature.DefaultFeatureGate.Enabled(features.CSIGlobalMountReconstruction)`.
+With the gate off, `ConstructVolumeSpec` behaves exactly as before. With the
+gate on, only the failure path is altered: the success path is unchanged.
```
jsafrane: Not only the fallback, but also the global dir reconstruction is feature-gated. I am not sure we need to mention it here - the whole feature is behind the feature gate.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+
+1. `csi_attacher.go`: `MountDevice` adds `specVolID` (from `spec.Name()`) and
+   `volumeLifecycleMode` (constant `string(storage.VolumeLifecyclePersistent)`)
+   to the data map written to the global `vol_data.json`. Field additions are
+   unconditional: written regardless of the feature gate, so a downgrade does
+   not produce stale or partial files.
+
+2. `pkg/kubelet/volumemanager` (reconstruction): after the existing scan of
+   `/var/lib/kubelet/pods`, a new step scans
+   `/var/lib/kubelet/plugins/kubernetes.io/csi/*/*` and skips every
```
jsafrane: So you hardcode CSI volume plugin details into the volume manager? That's IMO bad.
I think there should be a new volume plugin interface / call (ListGlobalVolumes?) that returns whatever the volume manager needs. And CSi can scan its kubernetes.io/csi directories and iSCSI can later add scan of its global dir.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,832 @@
...
+   each remaining directory: an empty one is removed directly; one with a
+   readable `vol_data.json` is turned into a `volume.Spec` (from the same
+   four fields the fallback uses) and added to the ActualStateOfWorld with
+   its device mount state marked uncertain, so the reconciler either
+   re-verifies it (volume still in the desired state) or calls
+   `UnmountDevice` / `NodeUnstageVolume` and removes the directory after a
+   successful unstage; one that is neither empty nor readable is logged as
+   an error and left for operator intervention, as today.
+
+   The pattern needs one exclusion. `plugins/kubernetes.io/csi` holds one
```
jsafrane: See how the CSI details leak into the volume manager. It's bad. Knowledge about volumeDevices shoul be in the CSI volume plugin.

jsafrane (commented):
I think the factual content of the KEP is mostly OK, however, it's quite verbose. The proposed changes are not complicated and their description should not be that long. Real people are reading it! I write long KEPs too, and I struggle making them short, but I think it could be manageable here.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Rewritten and cut to a third: it opens with what an admin sees, and the
mechanics stayed in Motivation and the Proposal, where they already were.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Done. Reconstruction now calls `ListGlobalVolumes` on any plugin implementing
`GlobalVolumeListerPlugin`, so FC is one method rather than a change to the
volume manager. It should fit: FC already rebuilds its spec from the global
directory name with `parsePDName`, and iSCSI from its own path with
`extractDeviceAndPrefix`, `extractPortalAndIqn` and `extractIface`.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: On the spec point you are right and it is unresolved: nothing calls
`NodeUnpublishVolume` there, so the CO obligation in the `NodeUnstageVolume`
section is not met. Your suggestion is what would fix it, since a name match is
what makes `NewUnmounter` succeed, and most of it is already on the branch:
`MountDevice` writes `specVolID`, the PV name, into the global file on every
stage, and `ConstructVolumeSpec` compares it against the pod directory name, as
a check after the mount table rather than as the key. What promoting
it costs is what the mount table gives for free: `Spec.Name()` is a PV name for
a persistent volume and a pod-spec entry for an inline ephemeral one, so an
inline volume whose pod-local file is unreadable could pair with an unrelated
staged PV of the same name. Taking the name only where there is no mount
reference at all, and refusing when more than one global mount claims it, keeps
that out of the common path. If that works for you I will make the change;
otherwise I would rather record the missing unpublish as a known alpha gap than
leave it unsaid.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Deleted, along with one more like it above.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Agreed, and changed in the code too. `GlobalVolumeListerPlugin` in
`pkg/volume/plugins.go` embeds `DeviceMountableVolumePlugin` and adds one
method, `ListGlobalVolumes() ([]GlobalVolume, error)`; reconstruction calls it
for every plugin that implements it and holds no path, glob or file name any
more. It is an interface rather than functions because reconstruction needs the
plugin itself afterwards, for `GetPluginName`, `CanDeviceMount` and
`GetUniqueVolumeNameFromSpec`.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Moved. The plugin compares the full path from `GetVolumeDevicePluginDir`
instead of hardcoding the directory name, and a test stages a directory under
it that passes every other check, so only the exclusion keeps it out.

cniackz:
On length: 6721 words down to 5641, so about 1.7x the median sig-storage KEP
rather than 2x. The Summary lost two thirds, the User Stories stopped retelling
Motivation, Implementation History went back to being dates, and the PRR
answers are shorter. Design Details grew, since it carries the interface. Next
to go are the remaining Proposal and Design Details overlaps.

cniackz:
@jsafrane one process question: #6058 still has no `lead-opted-in` label or
milestone, and the enhancements opt-in for v1.38 closes this month. If v1.38
still looks right to you, that label is what unblocks it.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,739 @@
...
+
+Raw block volumes are excluded from alpha: `GlobalVolume` carries one path and
+no volume mode, and describing a block volume's staging, publish and device
+paths needs that field.
+
+#### Beta
+
+- Raw block volumes, which need a volume mode on `GlobalVolume` before their
+  staging path can be described, with unit tests in
+  `pkg/kubelet/volumemanager`.
```
jsafrane, line 467: Block volume reconstruction needs to be done during alpha. It can be implemented later than filesystem volumes, still, when the feature gate goes Beta, all functionality must be implemented and tested.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,739 @@
...
+
+- Two releases at beta with the gate default-on, no regressions.
+- Conformance test if SIG Architecture deems applicable.
+- Production usage documented (CSI driver vendors confirm no surprises).
+
+### Upgrade / Downgrade Strategy
+
+- Upgrade with gate disabled: no behavior change. New global
+  `vol_data.json` files include extra fields; older code ignores unknown
+  fields, so a future downgrade is safe.
```
jsafrane, line 484: Here (or in the bullet below) describe what happens if new kubelet with the feature gate enabled sees old vol_data.json. It does not see new fields there and thus `ListGlobalVolumes` returns XYZ and ... what?
Kubelet should be at least as good as without the feature gate, i.e. if there is pod-level vol_data.json, the volume is reconstructed and marked uncertain both as pod-mount and device-mount.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -0,0 +1,739 @@
...
+prevents: the reconciler re-verifies anything still in the desired state.
+
+###### What specific metrics should inform a rollback?
+
+`reconstruct_volume_operations_errors_total` should not increase post-rollout
+versus pre-rollout. If it does, disable the gate.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Will be exercised in the e2e test added at beta.
```
jsafrane, line 555: All e2e tests need to be available *before* the feature graduates to beta.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
jsafrane: > Taking the name only where there is no mount reference at all, and refusing when more than one global mount claims it, keeps that out of the common path

That sounds fairly complicated. Keep the KEP as it is, use the global json only if there is a mount reference from the Pod dir.

We can improve it later (as  a bugfix / separate KEP), it can happen only after node reboot + corrupted pod-local json.

jsafrane:
I added the milestone label to #6058.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Agreed, block moves into alpha. The unmap side already works without a pod: in a local test a staged block volume was torn down with no pod anywhere, including NodeUnstageVolume and the staging and publish directories. What is left is the listing. Block stages under volumeDevices with a layout of its own, so the CSI plugin needs a second walk there and reconstruction has to stop skipping it. The volume mode has to travel with it, or UnmountDevice takes the filesystem branch, finds no volume data, and reports success on a volume that is still staged.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Tested on the branch, both gate positions. With an old global `vol_data.json` (no `specVolID`), `ListGlobalVolumes` names the spec from the volume handle, and the unique name is driver plus handle, so it is the same volume the pod walk finds. In both runs the pod mount and the device mount came out uncertain, so the gate makes it no worse. One detail for the KEP: when both sources see a volume, `addVolume` keeps the first spec and each pod keeps its own. That does not move the unstage, which recomputes from driver and handle.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Understood. I moved the e2e tests into alpha so they land in CI before the gate goes beta, and fixed the PRR answer that promised them at beta.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
cniackz: Agreed, keeping it as it is. The KEP records it as a known alpha gap: after a reboot with a corrupt pod-local `vol_data.json`, `NodeUnpublishVolume` is never called, with your note that it can be fixed later as a bugfix or its own KEP. Thank you.

keps/sig-storage/6058-csi-global-mount-fallback/README.md
jsafrane: I mean, write it into the KEP. It's a special corner case that we need to have covered.

jsafrane:
@kannon92 there is a small detail to iron out, but overall the PR is ready for PRR.
edit: I lgtm-ed, the PR is good for alpha, IMO

keps/sig-storage/6058-csi-global-mount-fallback/README.md
jsafrane: Oh, I can see a bullet below with "*Upgrade with the gate enabled, over volumes an older kubelet staged*"

keps/sig-storage/6058-csi-global-mount-fallback/README.md
```diff
@@ -484,6 +515,15 @@ paths needs that field.
   fields, so a future downgrade is safe.
 - Upgrade with gate enabled: reconstruction uses the fallback when needed.
   No interaction with control plane.
+- Upgrade with the gate enabled, over volumes an older kubelet staged: those
+  global files carry no `specVolID`, so the listing names the spec from the
+  volume handle instead. The unique volume name comes from the driver and the
+  handle either way, so it is the same volume the pod directory walk finds, and
+  a volume whose pod directory survived is still marked uncertain as a pod mount
```
jsafrane, line 522: What if there is no pod-local `vol_data.json`? What will the `GlobalVolumeListerPlugin` return and how will the volume manager deal with it?

Again, it's file if the global mount does not reach the actual state of the world, becuase it's the same as today, but we should know what to expect.

jsafrane: Sorry, I can see it's answered below.

jsafrane:
/lgm
/approve 
from the sig-storage point of view

keps/prod-readiness/sig-storage/6058.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 6058
+alpha:
+  approver: "@kannon92"
```
soltysh, line 3: Kevin is already pretty much booked for 1.38, I'll take this one over.

PR #6064, opened 2026-05-06
KEP-6063: Configuration for Per-Pod PID Limit

BhargaviGudi:
- One-line PR description:
Add support for setting PID limits at the Pod level to override the node-level kubelet PID limit configuration. This allows individual Pods to specify their own PID limits using spec.resources.limits.pid in the Pod specification.

- Issue link: https://github.com/kubernetes/enhancements/issues/6063
- Implementation PR: https://github.com/kubernetes/kubernetes/pull/139277

### DIFF
--- a/keps/sig-node/6063-pod-pid-limit/README.md
+++ b/keps/sig-node/6063-pod-pid-limit/README.md
@@ -1,0 +1,684 @@
+# KEP-6063: Per-Pod PID Limit
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [API Changes](#api-changes)
+  - [Valid Values](#valid-values)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+    - [Story 3](#story-3)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [Feature Gate](#feature-gate)
+    - [API Validation](#api-validation)
+    - [Kubelet Implementation](#kubelet-implementation)
+    - [Pod Security Admission (PSA)](#pod-security-admission-psa)
+    - [Eviction Manager Interaction](#eviction-manager-interaction)
+    - [Node Declared Features Integration](#node-declared-features-integration)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha (target 1.37)](#alpha-target-137)
+    - [Beta (target 1.38)](#beta-target-138)
+    - [GA (target 1.40)](#ga-target-140)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [x] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [x] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Add support for setting PID limits at the Pod level to restrict below the node-level kubelet PID limit. This allows individual Pods to specify their own PID limits using `spec.resources.limits.pid` in the Pod specification. The effective limit is the lower of the Pod-specified value and the node's `podPidsLimit`.
+
+## Motivation
+
+Currently, PID limiting in Kubernetes is configured globally at the node level via the kubelet's `podPidsLimit` setting. This creates challenges when different workloads on the same node have different PID requirements:
+1. **Infrastructure pods affected by low limits**: If a cluster administrator sets a low PID limit to protect against runaway processes, this limit applies equally to all pods on the node, including critical infrastructure pods that may legitimately need more PIDs.
+
+2. **Agent and sandbox workloads**: Modern workloads like AI agent runtimes (Agent Sandbox, OpenShell) and service mesh sidecars (Istio) often have specific PID requirements that differ from typical application pods.
+
+3. **No granular control**: Administrators cannot apply different PID limits to different types of workloads without node-level segregation, which is operationally expensive.
+
+4. **Pod-level resource model**: With the introduction of Pod-level resources in Kubernetes, PID limits logically belong alongside other Pod-level resource constraints.
+
+
+### Goals
+
+- Allow Pods to specify PID limits via `spec.resources.limits.pid` in the Pod specification
+- Restrict below the kubelet's default `podPidsLimit` when a Pod-level limit is set
+- Use the lower value when both kubelet and Pod limits are specified
+- Implement PID limiting at the Pod cgroup level for cgroupsv2
+- Maintain backward compatibility with existing node-level PID limiting
+
+### Non-Goals
+
+- Support for non-Linux operating systems (PID cgroup limits are Linux-specific)
+- Support for cgroupsv1 (this enhancement targets cgroupsv2 only)
+- Container-level PID limits (this KEP focuses on Pod-level limits only)
+- Automatic PID limit calculation based on workload characteristics
+- Changes to the existing node-level PID limiting mechanism
+- In-place resize of PID limits (in-place resize currently supports only CPU and memory)
+
+## Proposal
+
+Add a new `pid` resource type that can be specified under `spec.resources.limits` in the Pod specification. When set, the effective PID limit is `min(podPidsLimit, pod.spec.resources.limits.pid)` — the Pod can only restrict further, not exceed the node limit.
+
+### API Changes
+Extend the Pod specification to support PID limits:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+spec:
+  resources:
+    limits:
+      pid: "2048"         # Pod-level PID limit
+  containers:
+  - name: app
+    image: my-app:latest
+    resources:
+      limits:
+        cpu: "2"
+        memory: "4Gi"
+  - name: sidecar
+    image: log-agent:latest
+    resources:
+      limits:
+        cpu: "100m"
+        memory: "64Mi"
+```
+
+- Only `spec.resources.limits.pid` is accepted; `spec.resources.requests.pid` is forbidden and will be rejected by the API server
+- When `spec.resources.limits.pid` is specified, the kubelet applies this limit to the Pod's cgroup
+- All containers in the Pod share the same PID pool enforced by the pod cgroup
+- If both kubelet `podPidsLimit` and Pod `spec.resources.limits.pid` are set, the kubelet enforces `min(podPidsLimit, spec.resources.limits.pid)`. The Pod may request a value higher than the node's `podPidsLimit` (up to the API maximum of 16384), but the kubelet always caps at the node's `podPidsLimit`, ensuring the node administrator retains control.
+- If `spec.resources.limits.pid` is not specified, the kubelet's `podPidsLimit` applies (existing behavior)
+
+
+### Valid Values
+Following the same constraints as node-level `podPidsLimit`:
+
+ - **Minimum**: 1024 (conservative lower bound)
+ - **Maximum**: 16384 (maximum supported in managed environments)
+ - **Default**: If not specified, inherits kubelet's `podPidsLimit`
+
+### User Stories
+
+#### Story 1
+Agent Sandbox Workloads
+
+As a platform engineer running AI agent workloads (such as Agent Sandbox or
+OpenShell), I need to restrict PID limits for agent pods because they run
+untrusted code, while allowing trusted application pods to use the full
+node default.
+
+Current state: I must set a low `podPidsLimit` at the node level to
+protect against runaway agent processes, but this unnecessarily restricts
+all pods on the node, including trusted application pods that may
+legitimately need more PIDs.
+
+With this enhancement: I can keep the node's `podPidsLimit` at a
+comfortable level (e.g., 4096) for trusted application pods. I can then
+set `spec.resources.limits.pid: "2048"` on agent pods to contain untrusted
+workloads. Because the kubelet enforces the lower of the two limits, agent
+pods are safely restricted to 2048, while trusted application pods can use
+the full node default of 4096.
+
+#### Story 2
+Infrastructure Pod Protection
+
+As a cluster administrator, I want to set a conservative PID limit (2048) for
+tenant workloads while allowing infrastructure pods (e.g., monitoring agents,
+log collectors, service mesh proxies) to use higher limits (4096) on the same
+nodes.
+
+Current state: I must choose between protecting the node (a low node limit
+that starves infrastructure pods) or allowing infrastructure pods to function
+(a high node limit that risks PID exhaustion from tenant pods).
+
+With this enhancement: I can set the kubelet's `podPidsLimit` to 4096 to
+support infrastructure pods. I can then enforce
+`spec.resources.limits.pid: "2048"` on tenant pods (via admission control or
+default LimitRanges). Since the lower value wins, tenant pods are strictly
+capped at 2048, while infrastructure pods can scale up to the node limit of
+4096.
+
+#### Story 3
+Service Mesh Sidecar Requirements
+
+As an application developer using Istio, my application pods with Envoy
+sidecars need more PIDs than standard applications because the sidecar creates
+multiple worker threads and connections.
+
+Current state: If the default namespace limit is too low, I have no way to
+increase my PID limit without the cluster administrator globally changing the
+node-level `podPidsLimit`, which reduces overall cluster safety.
+
+With this enhancement: The cluster administrator sets the node's
+`podPidsLimit` to 8192. Standard application pods set
+`spec.resources.limits.pid: "2048"` to stay conservative. For my Istio
+pods, I set `spec.resources.limits.pid: "4096"` to give the Envoy sidecar
+the headroom it needs. The kubelet enforces `min(8192, 4096) = 4096`,
+granting my pod more PIDs without affecting other workloads or requiring
+a node-level configuration change.
+
+### Notes/Constraints/Caveats
+
+cgroupsv2 only: This enhancement requires cgroupsv2. On cgroupsv1 nodes, the kubelet will reject pods that specify `spec.resources.limits.pid` during admission, rather than silently ignoring the field. This prevents a situation where the user expects PID enforcement but the node cannot deliver it.
+
+Lower value precedence: When both kubelet and Pod specify limits, the lower value is enforced. This ensures that node-level protections cannot be bypassed by Pod specifications.
+
+Hierarchical cgroup limits: PID limits are hierarchical in cgroups, so the most restrictive limit in the hierarchy applies.
+
+Resource quota and limits: PID limits should be considered in cluster resource planning, though they do not consume schedulable resources like CPU or memory.
+
+Pod Security Admission (PSA) compatibility: The `spec.resources.limits.pid` field has no PSA impact. Pods specifying PID limits are admitted under all PSA profiles (Privileged, Baseline, and Restricted) without triggering any policy violations. PSA does not inspect resource limits, so no special configuration or exemptions are needed.
+
+Node Declared Features: This feature integrates with the Node Declared Features framework ([KEP-5328](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5328-node-declared-features)) to handle version skew and mixed cgroupsv1/v2 clusters. Nodes that support per-pod PID limits declare `PerPodPIDLimit` in `node.status.declaredFeatures`, enabling the scheduler to avoid placing pods with `spec.resources.limits.pid` on incompatible nodes.
+
+`hostPID` compatibility: Pods with `hostPID: true` and `spec.resources.limits.pid` are fully supported. The PID namespace (process visibility) and PID cgroup controller (process creation accounting) are independent kernel subsystems. `hostPID` allows the container to see host processes, but those processes belong to different cgroups and are not counted against the pod's PID limit. The cgroup `pids.max` limit is enforced regardless of PID namespace configuration.
+
+### Risks and Mitigations
+
+1. **Users expect to raise PID limits above node default**
+   - Risk: The "lower value wins" design means Pods can only restrict, not raise. Users may expect `pid: 8192` to override a node `podPidsLimit` of 4096.
+   - Mitigation: Clear documentation. The node admin retains control — raise `podPidsLimit` if workloads need more.
+
+2. **Interaction with existing PID exhaustion protections**
+   - Risk: A Pod could set a very low PID limit, making itself non-functional.
+   - Mitigation: API validation enforces a minimum of 1024 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (1024), the pod has sufficient PIDs to function.
+
+3. **Version skew: apiserver accepts but kubelet ignores**
+   - Risk: If only apiserver enables the gate, users think their limit is enforced but kubelet applies the node default.
+   - Mitigation: Documented in Version Skew Strategy. Both components must enable the gate for enforcement.
+
+## Design Details
+
+#### Feature Gate
+The feature is controlled by a new feature gate: `PerPodPIDLimit`.
+
+This feature depends on the `PodLevelResources` feature gate (Beta, enabled by default
+since v1.34) because `pid` is specified under `pod.spec.resources.limits`. If
+`PodLevelResources` is disabled, `PerPodPIDLimit` cannot be enabled — the kubelet
+and API server will reject the configuration at startup.
+
+When the feature gate is disabled:
+
+- The kube-apiserver rejects Pods that specify `pid` under `spec.resources.limits`
+- Existing Pods that already contain `pid` are preserved and remain unchanged
+- The kubelet ignores the `pid` limit and continues using the node-level `podPidsLimit` setting instead
+
+#### API Validation
+Add `pid` to the list of valid resource types
+
+Validate that `pid` values are integers within the allowed range (1024-16384)
+
+#### Kubelet Implementation
+Parse Pod specification: The kubelet reads `spec.resources.limits.pid` from the Pod spec during Pod admission.
+
+Limit enforcement: When creating the Pod cgroup, the kubelet sets effective PID limits based on:
+
+- If Pod `spec.resources.limits.pid` is set: `min(podPidsLimit, pod.spec.resources.limits.pid)`
+- If Pod `spec.resources.limits.pid` is not set: `podPidsLimit` (current behavior)
+
+Validation: The kubelet validates that the PID limit is within acceptable bounds (1024 to 16384).
+
+cgroupsv2 requirement: The kubelet checks the cgroup version and only applies Pod-level limits on cgroupsv2 systems. On cgroupsv1 systems, the kubelet rejects pods that specify `spec.resources.limits.pid` during admission.
+
+Event on capping: When a pod's requested PID limit exceeds the node's `podPidsLimit`, the kubelet emits a `PIDLimitCapped` event on the pod, indicating the effective limit was capped at the node-defined maximum. This gives users visibility when their requested PID limit is not fully honored.
+
+#### Pod Security Admission (PSA)
+
+This feature has no PSA interaction. Pod Security Admission does not inspect
+resource limits (`spec.resources.limits`), so pods specifying
+`spec.resources.limits.pid` are admitted under all PSA profiles (Privileged,
+Baseline, and Restricted) without triggering any policy violations. No special
+configuration, exemptions, or PSA policy changes are required.
+
+#### Eviction Manager Interaction
+
+The kubelet eviction manager monitors node-level PID availability and can set
+the `PIDPressure` node condition when available PIDs fall below configured
+eviction thresholds. When PID pressure is detected, the eviction manager may
+evict pods to recover node stability.
+
+Per-pod PID limits do not modify the eviction manager's behavior, eviction
+thresholds, or pod ranking logic. PID pressure detection continues to be based
+on node-level PID availability (`pid.available`) and operates independently of
+pod-level PID limits.
+
+Per-pod PID limits act as a preventative control by limiting the maximum number
+of processes that an individual pod may create. This can reduce the likelihood
+of node-wide PID exhaustion and therefore reduce the frequency of
+`PIDPressure` conditions.
+
+#### Node Declared Features Integration
+
+This feature integrates with the Node Declared Features framework
+([KEP-5328](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5328-node-declared-features),
+GA since v1.36) to handle version skew and mixed cgroupsv1/v2 clusters
+gracefully.
+
+When the `PerPodPIDLimit` feature gate is enabled and the node supports
+cgroupsv2, the kubelet declares `PerPodPIDLimit` in
+`node.status.declaredFeatures` during bootstrap. This enables:
+
+- **Scheduler filtering**: The scheduler infers that a pod with
+  `spec.resources.limits.pid` requires `PerPodPIDLimit` and only places it on
+  nodes that declare the feature. This prevents pods from being scheduled to
+  cgroupsv1 nodes or nodes without the feature gate, avoiding kubelet admission
+  rejections that would put the pod into a non-retriable `Failed` state.
+- **Kubelet admission validation**: Before admitting a pod, the kubelet
+  validates the pod's feature requirements against its declared features as a
+  secondary check (handles node restart with a feature gate flip).
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+No prerequisite testing updates are required. The existing PID limiting tests
+in `test/e2e_node/pids_test.go` (`PodPidsLimit` suite) cover the node-level
+`podPidsLimit` behavior which this feature builds upon.
+
+##### Unit tests
+
+This feature touches API validation, field stripping, and kubelet cgroup enforcement. For Alpha, unit test coverage for the following packages is added:
+
+* `pkg/apis/core/validation` will be updated with validation rules for the `pid` resource (range 1024-16384, container-level rejection, `requests.pid` forbidden).
+* `pkg/api/pod` will be updated with field stripping logic when the `PerPodPIDLimit` gate is disabled (`requests.pid` is always stripped).
+* `pkg/kubelet/cm` will be updated with `getPodPIDLimit` logic (pod limit lower/higher than node, no pod limit fallback), cgroupsv1 rejection error message, `PIDLimitCapped` event emission when the effective limit is capped, and static pod PID limit enforcement (kubelet applies `getPodPIDLimit` uniformly for both regular and static pods).
+* `pkg/apis/core/helper` will be updated to exclude `ResourcePID` from `standardContainerResources`.
+* `pkg/kubelet/eviction` will be updated to verify that PID pressure detection, eviction ranking, and pod selection are unaffected by `spec.resources.limits.pid` when the `PerPodPIDLimit` gate is enabled.
+* `k8s.io/component-helpers/nodedeclaredfeatures` will be updated to verify that the kubelet declares `PerPodPIDLimit` in `node.status.declaredFeatures` when the feature gate is enabled and cgroupsv2 is available, and does not declare it when the gate is disabled or on cgroupsv1 nodes.
+
+`k8s.io/kubernetes/pkg/apis/core/validation`: API validation of `pid` resource
+`k8s.io/kubernetes/pkg/api/pod`: Field stripping when gate is disabled
+`k8s.io/kubernetes/pkg/kubelet/cm`: getPodPIDLimit enforcement logic
+`k8s.io/kubernetes/pkg/apis/core/helper`: ResourcePID helper functions
+`k8s.io/kubernetes/pkg/kubelet/eviction`: Eviction manager unaffected by per-pod PID limits
+`k8s.io/component-helpers/nodedeclaredfeatures`: Node Declared Features registration and inference
+
+##### Integration tests
+
+e2e tests provide good test coverage of the interaction between the API server (validation of `pid` resource) and the kubelet (cgroup enforcement). We may add integration tests before Beta to cover API admission edge cases that are not covered by the planned unit and e2e tests.
+
+##### e2e tests
+
+* Pod PID limit lower than node default is applied.
+* Pod PID limit higher than node default is capped at node default and a `PIDLimitCapped` event is emitted.
+* Multi-container pod shares a single pod-level PID limit.
+* No pod PID limit falls back to node default.
+* Pod with pid limit is admitted under Baseline PSA profile and limit is enforced.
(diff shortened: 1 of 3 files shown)


ngopalak-redhat:
I'm working with @BhargaviGudi on this KEP. 
[KEP-2837](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2837-pod-level-resource-spec/README.md) sets up the `resources` stanza at Pod level but limited scope to CPU, memory, and hugepages. This KEP extends that same API to add PID limits
This KEP has a dependency on the `PodLevelResources` featuregate which enabled by default now. This KEP introduces a new feature gate `PerPodPIDLimit` only for the Pid limits.
cc: @haircommander @ndixita

keps/sig-node/6063-pod-pid-limit/kep.yaml
```diff
@@ -0,0 +1,47 @@
...
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.37"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.37"
+  beta: "v1.38"
+  stable: "v1.39"
```
haircommander: usually features take two releases in beta

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+- Support for non-Linux operating systems (PID cgroup limits are Linux-specific)
+- Support for cgroupsv1 (this enhancement targets cgroupsv2 only)
+- Container-level PID limits (this KEP focuses on Pod-level limits only)
+- Automatic PID limit calculation based on workload characteristics
+- Changes to the existing node-level PID limiting mechanism
+
```
haircommander, line 105: in place resize in scope?

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+    image: log-agent:latest
+    resources:
+      limits:
+        cpu: "100m"
+        memory: "64Mi"
+```
+
+- When `spec.resources.limits.pid` is specified, the kubelet applies this limit to the Pod's cgroup
+- All containers in the Pod share the same PID pool enforced by the pod cgroup
+- If both kubelet `podPidsLimit` and Pod `spec.resources.limits.pid` are set, the **lower value wins**
```
haircommander: this implies a contradiction: `spec.resources.limits.pid` is apparently not able to be above podPidsLimit, so if it's set, then it always wins

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+- If `spec.resources.limits.pid` is not specified, the kubelet's `podPidsLimit` applies (existing behavior)
+- A value of `-1` or `max` means unlimited PIDs (up to node capacity)
```
haircommander: I think we should just allow to be unset rather than special casing values...

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+Include as much detail as possible so that people can understand the "how" of
+the system. The goal here is to make this feel real for users without getting
+bogged down.
+-->
+
+#### Story 1
+Agent Sandbox Workloads
+
+As a platform engineer running AI agent workloads (such as Agent Sandbox or
+OpenShell), I need to allow high PID limits for agent pods while enforcing
```
haircommander: I actually think the use case is the opposite. i  want agent pods to be more limited than application pods because they're running untrusted workloads

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+
+### Notes/Constraints/Caveats (Optional)
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+cgroupsv2 only: This enhancement requires cgroupsv2. Nodes running cgroupsv1 will ignore the Pod-level PID limit and fall back to kubelet configuration.
```
haircommander: I don't know if ignore is right. we could use node capabilities to branch on whether a node can handle the field.  i also think the kubelet should deny any pods that attempt to use it that get scheduled to its node if we're using v1, as the apiserver (or more importantly, policy engines) expects the pod be more limited than it will be

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+   - Risk: The "lower value wins" design means Pods can only restrict, not raise. Users may expect `pid: 8192` to override a node `podPidsLimit` of 4096.
+   - Mitigation: Clear documentation. The node admin retains control — raise `podPidsLimit` if workloads need more.
+
+2. **Interaction with existing PID exhaustion protections**
+   - Risk: A Pod setting `pid: 1` could make itself non-functional.
+   - Mitigation: This is user-initiated and no different from setting `memory: 1Mi`. The Pod fails on its own, not affecting other Pods.
+
+3. **Version skew: apiserver accepts but kubelet ignores**
+   - Risk: If only apiserver enables the gate, users think their limit is enforced but kubelet applies the node default.
+   - Mitigation: Documented in Version Skew Strategy. Both components must enable the gate for enforcement.
```
haircommander, line 239: node declared features, and kubelet should deny this if it's new enough

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+Parse Pod specification: The kubelet reads `spec.resources.limits.pid` from the Pod spec during Pod admission.
+
+Limit enforcement: When creating the Pod cgroup, the kubelet sets effective PID limits based on:
+
+- If Pod `spec.resources.limits.pid` is set: `min(podPidsLimit, pod.spec.resources.limits.pid)`
+- If Pod `spec.resources.limits.pid` is not set: `podPidsLimit` (current behavior)
+
+Validation: The kubelet validates that the PID limit is within acceptable bounds (1024 to 16384, or -1 for unlimited).
+
+cgroupsv2 requirement: The kubelet checks the cgroup version and only applies Pod-level limits on cgroupsv2 systems. On cgroupsv1 systems, a warning is logged and the node-level limit applies.
```
haircommander: for future reference: a log line is not visible enough to a sufficiently busy cluster admin. metrics at minimum, but really we should try to prevent this case

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+
+### Valid Values
+Following the same constraints as node-level `podPidsLimit`:
+
+ - **Minimum**: 1024 (conservative lower bound)
+ - **Maximum**: 16384 (maximum supported in managed environments)
+ - **Special values**: `-1` for unlimited (defaults to node allocatable PID
+      capacity)
+ - **Default**: If not specified, inherits kubelet's `podPidsLimit`
+
```
haircommander, line 150: who sets this value? kubelet with libcontainer cgroups package?

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+Limit enforcement: When creating the Pod cgroup, the kubelet sets effective PID limits based on:
+
+- If Pod `spec.resources.limits.pid` is set: `min(podPidsLimit, pod.spec.resources.limits.pid)`
+- If Pod `spec.resources.limits.pid` is not set: `podPidsLimit` (current behavior)
+
+Validation: The kubelet validates that the PID limit is within acceptable bounds (1024 to 16384, or -1 for unlimited).
+
+cgroupsv2 requirement: The kubelet checks the cgroup version and only applies Pod-level limits on cgroupsv2 systems. On cgroupsv1 systems, a warning is logged and the node-level limit applies.
+
+### Test Plan
```
haircommander: this section and below need some attention

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+-->
+
+#### Alpha (target 1.37)
+
+- Feature implemented behind `PerPodPIDLimit` feature gate (disabled by default)
+- Kubelet enforces `min(podPidsLimit, pod.spec.resources.limits.pid)` on Pod cgroup
+- Unit tests and initial node e2e tests completed
+- `kubelet_pod_pid_limit_applied_total` metric exposed
+
+#### Beta (target 1.39)
```
haircommander: these targets are different than what's in kep.yaml above

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,993 @@
...
+* **2026-05-06:** [Initial KEP PR created](https://github.com/kubernetes/enhancements/pull/6064)
+* **2026-05-25:** [Alpha implementation PR created](https://github.com/kubernetes/kubernetes/pull/139277) targeting v1.37
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
```
haircommander, line 668: nri could be an alternative, annotation based approach, we could base on priority class

haircommander (commented):
probably calling out somewhere PSA interaction (any PSA level can set this field) would be good

keps/sig-node/6063-pod-pid-limit/README.md
ngopalak-redhat: I think it is safe to add in-place pod resizing (explained [here](https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/)) to the Non-Goals section due to the following limitation of that feature:
```
Resource Types: Only CPU and memory resources can be resized.
```
Since we are exclusively dealing with PID limits, this will not affect our work.

keps/sig-node/6063-pod-pid-limit/kep.yaml
BhargaviGudi: updated the GA target from 1.39 to 1.40 to allow two releases in Beta (1.38, 1.39).

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated. The Pod may request a value higher than the node's podPidsLimit (up to the API maximum of 16384), but the kubelet always caps at the node's podPidsLimit. So the effective limit is min(podPidsLimit, spec.resources.limits.pid) — the lower value wins. Clarified this in the proposal section.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Agreed. Removed any mention of -1 or max as special values. If spec.resources.limits.pid is not set, the kubelet's podPidsLimit applies (existing behavior). No special-case values needed.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated the behavior: on cgroupsv1 nodes, the kubelet now rejects pods that specify spec.resources.limits.pid during admission with an explicit error, rather than silently ignoring the field.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: The cgroupsv1 case now returns an admission error.

BhargaviGudi:
> probably calling out somewhere PSA interaction (any PSA level can set this field) would be good

Added `PSA compatibility` note to the` Notes/Constraints/Caveats` section. Also added e2e tests validating that pods with spec.resources.limits.pid are admitted and enforced under Baseline and Restricted PSA profiles. The pid resource field has no PSA impact — it doesn't trigger any PSA policy violations under any profile (Privileged, Baseline, or Restricted).

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: added in-place resize to Non-Goals.

kannon92:
Please request a PRR review and you can add me as an approver.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+2. **Interaction with existing PID exhaustion protections**
+   - Risk: A Pod setting `pid: 1` could make itself non-functional.
+   - Mitigation: This is user-initiated and no different from setting `memory: 1Mi`. The Pod fails on its own, not affecting other Pods.
+
+3. **Version skew: apiserver accepts but kubelet ignores**
+   - Risk: If only apiserver enables the gate, users think their limit is enforced but kubelet applies the node default.
+   - Mitigation: Documented in Version Skew Strategy. Both components must enable the gate for enforcement.
+
+## Design Details
```
HirazawaUi, line 241: Could you please explain how this feature integrates with PSA in this section? While you mentioned PSA-related tests in the e2e testing chapter, the *Design Details* have not yet been described.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+Consider including folks who also work outside the SIG or subproject.
+-->
+
+1. **Users expect to raise PID limits above node default**
+   - Risk: The "lower value wins" design means Pods can only restrict, not raise. Users may expect `pid: 8192` to override a node `podPidsLimit` of 4096.
+   - Mitigation: Clear documentation. The node admin retains control — raise `podPidsLimit` if workloads need more.
+
+2. **Interaction with existing PID exhaustion protections**
+   - Risk: A Pod setting `pid: 1` could make itself non-functional.
+   - Mitigation: This is user-initiated and no different from setting `memory: 1Mi`. The Pod fails on its own, not affecting other Pods.
```
HirazawaUi: Since you've restricted the allowed PID range to 1024-16384, this should not be an issue. Alternatively, would it be possible to include this range limit in the Mitigation section?

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+Unit tests cover gate toggle: validation accepts/rejects `pid` based on gate state, and field stripping preserves `pid` on existing objects when gate is off.
+
+### Rollout, Upgrade and Rollback Planning
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
```
HirazawaUi, line 481: Restarting the kubelet will not affect running workloads.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+<!--
+Describe them, providing:
+  - Which API(s):
+  - Estimated increase:
+-->
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
```
HirazawaUi, line 600: Yes, this is because we have added new fields to the Pod API.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+For each of these, fill in the following—thinking about running existing user workloads
+and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+-->
+
+### Scalability
```
HirazawaUi, line 584: Please fill in this section.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+This feature only works on cgroupsv2 nodes. Pods specifying `spec.resources.limits.pid` will be rejected on cgroupsv1 nodes, which may cause confusion in mixed clusters where some nodes have been upgraded to cgroupsv2 and others have not. Workload authors need to be aware of the underlying cgroup version to avoid unexpected admission failures.
+
+Additionally, the "lower value wins" design means pods can only restrict PID limits below the node default, not raise them. Users who expect `pid: 8192` to override a node `podPidsLimit` of 4096 may find this counterintuitive, though this is consistent with how Kubernetes treats node-level resource limits as the administrator's ceiling.
+
+## Alternatives
```
HirazawaUi, line 668: A potential alternative would be to add the `nproc` option for `ulimit` in KEP-5758. However, since `ulimit` does not operate at the cgroup level, we decided not to pursue this implementation in KEP-5758. For more details, please refer to: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5758-per-container-ulimits-configuration

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+spec:
+  resources:
+    limits:
+      pid: "2048"         # Pod-level PID limit
```
SergeyKanzhelev, line 121: why not both - Pod and Container level?

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+
+With this enhancement: I can set the kubelet's `podPidsLimit` to 4096 to
+support infrastructure pods. I can then enforce
+`spec.resources.limits.pid: "2048"` on tenant pods (via admission control or
+default LimitRanges). Since the lower value wins, tenant pods are strictly
+capped at 2048, while infrastructure pods can scale up to the node limit of
+4096.
+
+#### Story 3
+Service Mesh Sidecar Requirements
```
SergeyKanzhelev, line 193: this user story will benefit by per-container pid limits

keps/sig-node/6063-pod-pid-limit/README.md
SergeyKanzhelev, line 121: Do we envision the request field for this in future iterations so sum or requests is not exceeding the node's  allocatable? If we are not doing it today, it may become the question later. Please add a note about it

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1025 @@
...
+    image: log-agent:latest
+    resources:
+      limits:
+        cpu: "100m"
+        memory: "64Mi"
+```
+
+- When `spec.resources.limits.pid` is specified, the kubelet applies this limit to the Pod's cgroup
+- All containers in the Pod share the same PID pool enforced by the pod cgroup
+- If both kubelet `podPidsLimit` and Pod `spec.resources.limits.pid` are set, the kubelet enforces `min(podPidsLimit, spec.resources.limits.pid)`. The Pod may request a value higher than the node's `podPidsLimit` (up to the API maximum of 16384), but the kubelet always caps at the node's `podPidsLimit`, ensuring the node administrator retains control.
```
SergeyKanzhelev, line 140: in this case (when min was chosen that doesn't match the requested value), I think we should either raise an event or have a pod status reflecting the "actual" value.

kannon92:
/ok-to-test

/hold

https://github.com/kubernetes/enhancements/pull/6064#issuecomment-4632883829

This needs to be addressed before this can continue.

BhargaviGudi:
Thanks for the detailed reviews. I'm actively working on addressing all the feedback and will push an updated revision shortly.

keps/sig-node/6063-pod-pid-limit/README.md
ngopalak-redhat: I'm working with @BhargaviGudi on this KEP.

We scoped this to a kubelet change and therefore restricted it to limits alone, which are enforced via cgroups.

Adding support for requests would mean the scheduler needs to check if the node can accommodate the pod. This gets complicated because the number of PIDs in use at the exact moment the scheduler evaluates the node is highly dynamic. To support requests, we would need a much more complex algorithm to reliably determine if a node can admit the pod. 

Can we open a new KEP for it? May be consider it for later release.

cc: @haircommander

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Removed the -1 special value per your earlier comment. To clarify the mechanism: we use libcontainercgroups.IsCgroup2UnifiedMode() only to check if the node is running cgroupsv2 ([line 101](https://github.com/kubernetes/kubernetes/pull/139277/files#diff-pod_container_manager_linux.go-L101)). The actual PID limit is set via containerConfig.ResourceParameters.PidsLimit ([line 112](https://github.com/kubernetes/kubernetes/pull/139277/files#diff-pod_container_manager_linux.go-L112)), which the kubelet passes to the container runtime through the CRI. The runtime then writes pids.max on the pod cgroup.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: It was discussed in the [sig-node weekly meeting](https://docs.google.com/document/d/1Ne57gvidMEWXR70OxxnRkYquAoMpt56o75oZtg-OeBg/edit?tab=t.0#heading=h.16dqthn53k2t) and the decision was to start with pod-level. I
n cgroupsv2, the PID controller is set at the pod cgroup (pids.max), and all containers in the pod share a single PID pool. Pod-level keeps the semantics simple and consistent with how cgroupsv2 PID isolation works.  Adding per-container would add complexity for Alpha.
Container-level PID limits are listed as a Non-Goal for this KEP and can be explored in a follow-up based on user feedback.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: The story is written for pod-level PID limits — the user sets a higher PID pool for the entire pod so all containers (including the Envoy sidecar) get the headroom they need.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Agreed. We've added a PIDLimitCapped warning event — the kubelet now emits it whenever the pod's requested PID limit exceeds the node's podPidsLimit, with the message: "Requested PID limit 8192 exceeds node podPidsLimit 4096; effective limit capped to 4096". The event is visible via kubectl describe pod. We also log it in kubelet logs. This is documented in the Design Details and Monitoring Requirements sections. made changes in implementation as well.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated Design Details.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated Risk #2 Mitigation to reference the API validation range.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Yes, filled in the "API object size" section

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Added to the Alternatives section.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Added NRI plugin (annotation-based) and PriorityClass-based approaches as alternatives in the Alternatives section.

BhargaviGudi:
> Please request a PRR review and you can add me as an approver.

@kannon92 PRR review has been requested and added you as an approver. Could you please help to review the KEP? Thanks

keps/sig-node/6063-pod-pid-limit/README.md
SergeyKanzhelev: makes sense. Pod level is a good place to start

keps/sig-node/6063-pod-pid-limit/kep.yaml
```diff
@@ -0,0 +1,47 @@
...
+feature-gates:
+  - name: PerPodPIDLimit
+    components:
+      - kubelet
+      - kube-apiserver
+disable-supported: true
+
+# The following PRR answers are required at beta release
+metrics:
+  - TBD
```
SergeyKanzhelev: is there a metric already on pid usage that user can watch today to understand what limit will work for them?

SergeyKanzhelev (approved):
/lgtm
/approve

keps/sig-node/6063-pod-pid-limit/kep.yaml
```diff
@@ -0,0 +1,47 @@
+title: Per-Pod PID Limit
+kep-number: 6063
+authors:
+  - "@BhargaviGudi"
+  - "@ngopalak-redhat"
+owning-sig: sig-node
+participating-sigs:
+  - sig-node
+status: provisional
```
kannon92:
```suggestion
status: implementable
```

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+have been executed regularly, and have been stable.
+This can be done with:
+- permalinks to the GitHub source code
+- links to the periodic job (typically https://testgrid.k8s.io/sig-release-master-blocking#integration-master), filtered by the test name
+- a search in the Kubernetes bug triage tool (https://storage.googleapis.com/k8s-triage/index.html)
+-->
+
+e2e tests provide good test coverage of the interaction between the API server (validation of `pid` resource) and the kubelet (cgroup enforcement). We may add integration tests before Beta to cover API admission edge cases that are not covered by the planned unit and e2e tests.
+
+##### e2e tests
```
kannon92, line 355: We should consider the interaction of this feature with eviction manager.

Eviction manager is aware of pids and does eviction based on running out of pids.

kannon92: https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/eviction_test.go#L660

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+When the feature gate is disabled:
+
+- The kube-apiserver rejects Pods that specify `pid` under `spec.resources.limits`
+- Existing Pods that already contain `pid` are preserved and remain unchanged
+- The kubelet ignores the `pid` limit and continues using the node-level `podPidsLimit` setting instead
+
+#### API Validation
+Add `pid` to the list of valid resource types
+
+Validate that `pid` values are integers within the allowed range (1024-16384)
```
kannon92, line 260: I think we need validation to forbid someone setting pids at the container level.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+
+2. **Interaction with existing PID exhaustion protections**
+   - Risk: A Pod could set a very low PID limit, making itself non-functional.
+   - Mitigation: API validation enforces a minimum of 1024 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (1024), the pod has sufficient PIDs to function.
+
+3. **Version skew: apiserver accepts but kubelet ignores**
+   - Risk: If only apiserver enables the gate, users think their limit is enforced but kubelet applies the node default.
+   - Mitigation: Documented in Version Skew Strategy. Both components must enable the gate for enforcement.
+
+## Design Details
```
kannon92, line 241: I would like to see a section on eviction manager.

https://github.com/kubernetes/enhancements/pull/6064/changes#r3382744067

https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-conditions

I don't think there is anything too concerning with this feature. In fact I think this may node pressure safer as there is more control over this.

But we should include testing and understand how this feature related to PID Pressure at the condition level.

keps/sig-node/6063-pod-pid-limit/kep.yaml
kannon92: I believe you mention a metric in the KEP that you are adding. You should add this here.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+Consider the following in developing a version skew strategy for this
+enhancement:
+- Does this enhancement involve coordinating behavior in the control plane and nodes?
+- How does an n-3 kubelet or kube-proxy without this feature available behave when this feature is used?
+- How does an n-1 kube-controller-manager or kube-scheduler without this feature available behave when this feature is used?
+- Will any other components on the node change? For example, changes to CSI,
+  CRI or CNI may require updating that component before the kubelet.
+-->
+
+A new field `spec.resources.limits.pid` is being added, hence version skew is not impacted.
```
kannon92: This is not true.

You can have skew between apiserver and kubelet. 

THink through validation and kubelet behavior in case of separate rollouts.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+<!--
+Describe them, providing:
+  - API type(s):
+  - Estimated increase in size: (e.g., new annotation of size 32B)
+  - Estimated amount of new objects: (e.g., new Object X for every existing Pod)
+-->
+
+- API type(s): Pod
+- Estimated increase in size: ~20-30 bytes per Pod when `spec.resources.limits.pid`
+  and `spec.resources.requests.pid` are set. Pods that do not use this field are unchanged.
```
kannon92: Are we allowing `spec.resources.requests.pids`?

requests are a scheduler concern so I think we should forbid the use of pids for requests.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+are missing a bunch of machinery and tooling and can't do that now.
+-->
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+<!--
+Even if applying deprecation policies, they may still surprise some users.
+-->
+
+### Monitoring Requirements
```
kannon92, line 520: Can you answer these questions in this section?

You partially answered some of them and didn't answer others.

I would choose answer all of them.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1091 @@
...
+
+<!--
+This section must be completed when targeting beta to a release.
+
+For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field.
+-->
+
+###### How can an operator determine if the feature is in use by workloads?
```
kannon92, line 522: One could use the metric you are adding which seems to be the purpose of this metric.

kannon92:
PRR review complete. Please address my comments but overall this is looking really close.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Yes, this is already implemented and tested. Setting pid at the container level is rejected by validation because pid is excluded from `standardContainerResources`.

Actual API error:
The Pod "neg-container-pid-test" is invalid: 
```
spec.containers[0].resources.limits[pid]: Invalid value: "pid": must be a standard resource for containers
spec.containers[0].resources.requests[pid]: Invalid value: "pid": must be a standard resource for containers
```

Unit test:`TestValidateContainerResourceName_PID/pid_rejected_at_container_level_even_when_gate_is_on`, added in implementation - passes regardless of gate state.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: No. This was discussed [here](https://github.com/kubernetes/enhancements/pull/6064#discussion_r3374768301) and agreed to keep the scope limited to PID limits only.

Supporting `spec.resources.requests.pid` would require scheduler changes to determine whether a node can accommodate the requested PIDs. Since PID usage on a node is highly dynamic, we felt this deserved a separate KEP and can be considered in a future release.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated monitoring requirements section

keps/sig-node/6063-pod-pid-limit/kep.yaml
BhargaviGudi: Updated

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Updated the Version Skew Strategy section with a full analysis of all four apiserver/kubelet skew scenarios:

- Apiserver ON, kubelet OFF: The apiserver accepts `spec.resources.limits.pid`, but the kubelet ignores it and applies the node-level `podPidsLimit`. Safe, but the per-pod limit is not enforced.

- Apiserver OFF, kubelet ON: Two scenarios:

1. Regular Pods: The apiserver silently rejects `spec.resources.limits.pid` from new pod specs, so the field never reaches the kubelet. The kubelet applies the node-level `podPidsLimit`.
2. Static Pods: Static pods are defined directly on the node and bypass the apiserver. Since the kubelet has the gate enabled, if `spec.resources.limits.pid` is present, the kubelet will enforce the configured validation and apply the pod-level PID limit logic accordingly.

- Both ON: Full enforcement. Apiserver validates, kubelet enforces `min(podPidsLimit, spec.resources.limits.pid)`.

- Both OFF: Feature disabled, existing behavior.

BhargaviGudi:
Working on the eviction manager part. I’ll update changes soon.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Now requests.pid is now explicitly forbidden in validation. Removed the `spec.resources.requests.pid` reference from the scalability section and updated the unit test descriptions.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,1137 @@
...
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+<!--
```
kannon92: nit: as you finish sections, can you remove the comments?

It really helps with reviewing.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: I've added a section describing the interaction with eviction manager and PIDPressure.

The feature does not modify PIDPressure calculation, eviction thresholds, or eviction behavior. Eviction manager continues to operate on node-level PID availability. Per-pod PID limits act as a preventative control that can reduce the likelihood of node-wide PID exhaustion.

I've also added e2e coverage to verify PIDPressure detection and eviction behavior remain unchanged when pod-level PID limits are enabled.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: I've added e2e coverage to verify PIDPressure detection and eviction behavior remain unchanged when pod-level PID limits are enabled.

BhargaviGudi:
@kannon92 I have addressed all the comments, added the Eviction Manager Interaction section under Design Details, and added e2e test coverage. Could you please take another look when you have a chance? Thanks!

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,639 @@
...
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
```
kannon92, line 53: Please fill this out as you complete this.

BhargaviGudi: Updated

kannon92 (commented):
PRR looks good for alpha.

/approve

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,639 @@
...
+- When `spec.resources.limits.pid` is specified, the kubelet applies this limit to the Pod's cgroup
+- All containers in the Pod share the same PID pool enforced by the pod cgroup
+- If both kubelet `podPidsLimit` and Pod `spec.resources.limits.pid` are set, the kubelet enforces `min(podPidsLimit, spec.resources.limits.pid)`. The Pod may request a value higher than the node's `podPidsLimit` (up to the API maximum of 16384), but the kubelet always caps at the node's `podPidsLimit`, ensuring the node administrator retains control.
+- If `spec.resources.limits.pid` is not specified, the kubelet's `podPidsLimit` applies (existing behavior)
+
+
+### Valid Values
+Following the same constraints as node-level `podPidsLimit`:
+
+ - **Minimum**: 1024 (conservative lower bound)
```
haircommander, line 147: why do we need a minimum? it could be even lower if we need it IMO

haircommander:
/lgtm

I feel the minimum could be dropped or lowered but that's not blocking

haircommander:
cc @SergeyKanzhelev

keps/sig-node/6063-pod-pid-limit/README.md
mrunalp: The minimum really goes down to the container runtime implementation. We shouldn't require 1024 which is quite high IMO. +1 to either dropping it or finding real world lowest limit that could be set on container runtimes.

SergeyKanzhelev: I would vote for some minimum to avoid people shoorting themselves in a foot. We picked `1024` in GKE just to be on a safe side: https://docs.cloud.google.com/kubernetes-engine/docs/how-to/node-system-config#pid_management but I am open for a smaller number. Just don't do it a single digit, it is always risky with exec probes suddenly not working and inability to run multiple containers

keps/sig-node/6063-pod-pid-limit/kep.yaml
```diff
@@ -1,26 +1,20 @@
...
-  - "@jsafrane"
-  - "@tallclair"
-  - "@ndixita"
+  - "@haircommander"
 approvers:
-  - "@dchen1107"
-  - "@msau42"
-  - "@jsafrane"
-  - "@tallclair"
+  - "@haircommander"
```
SergeyKanzhelev: should be @mrunalp

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,639 @@
...
+Extend the Pod specification to support PID limits:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+spec:
+  resources:
+    limits:
```
SergeyKanzhelev, line 120: should this be incompatible with `hostPID: true`?

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,639 @@
...
+### API Changes
+Extend the Pod specification to support PID limits:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: example-pod
+spec:
+  resources:
```
SergeyKanzhelev, line 119: Make a note that requests cannot be set for pid for now and reject those pods explicitly

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -0,0 +1,639 @@
...
+* Static pod without `spec.resources.limits.pid` falls back to node-level `podPidsLimit`.
+* Node enters `PIDPressure` when available PIDs fall below configured eviction thresholds, regardless of pod-level PID limits.
+* Eviction manager continues to evict pods under `PIDPressure`; pod-level PID limits do not alter eviction behavior or pod selection.
+* Pod-level PID limits reduce aggregate PID consumption and can prevent a workload from triggering `PIDPressure` compared to the node-level default limit.
+* PID pressure calculation is based on node-level PID availability and is unaffected by the configured value of `spec.resources.limits.pid`.
+
+- [PerPodPIDLimit](https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/pids_test.go): [SIG Node](https://testgrid.k8s.io/sig-node-kubelet-serial), [triage search](https://storage.googleapis.com/k8s-triage/index.html?test=PerPodPIDLimit)
+- [PerPodPIDLimit PSA Compatibility](https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/pids_test.go): validates admission and enforcement under Baseline and Restricted PSA profiles
+
+### Graduation Criteria
```
SergeyKanzhelev, line 377: will this be integrated with the Node Declared Features or we just rely on graceful degradation?

keps/sig-node/6063-pod-pid-limit/README.md
ngopalak-redhat: Thanks, we'll do some testing of workloads and update this minimum.

keps/sig-node/6063-pod-pid-limit/README.md
ngopalak-redhat: My understanding is that cgroup v2 will still enforce PID limits even if hostPID is set to true. We will try this scenario in the draft implementation and update the KEP.

keps/sig-node/6063-pod-pid-limit/README.md
ngopalak-redhat: NodeDeclaredFeatures feature gate is set to true in v1.36. We mostly referred to podlevelresources KEP while writing this KEP and hence didn't consider this. We should be considering the use of Node Declared features as the KEP for NodeDeclaredFeature particularly mentions this use case: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5328-node-declared-features/README.md#user-stories
Will update the KEP

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Added a note in the API Changes section: "Only spec.resources.limits.pid is accepted; spec.resources.requests.pid is forbidden and will be rejected by the API server."

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Addressed the feedback by integrating Node Declared Features (KEP-5328) into the KEP, including design details, version skew handling, dependency documentation, and unit/e2e test coverage for feature declaration and scheduler filtering.

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: I investigated the interaction between PerPodPIDLimit and hostPID: true on a cgroupsv2 node.

In testing, a pod with hostPID: true and spec.resources.limits.pid: 1024:

- Was admitted successfully.
- Had pids.max=1024 applied to the pod cgroup.
- Could see host processes via ps.
- Was unable to create more than ~1024 processes; additional forks failed with EAGAIN, and cgroup metrics confirmed the limit was enforced.

Host processes were visible but not counted against the pod's PID limit because they belong to different cgroups.
Detailed testing results can be found [here](https://docs.google.com/document/d/1ZF3FO9EKovu4DM2l3xf6gD8CRcj6PDNHg5xgRZJj7RE/edit?tab=t.0)

I've added a note in the KEP documenting hostPID compatibility and added an e2e test case covering this scenario.

kannon92:
https://github.com/kubernetes/enhancements/pull/6064#issuecomment-4651085033

This is done.

/hold cancel

BhargaviGudi:
@SergeyKanzhelev @kannon92 @mrunalp I have addressed the latest review comments.

I'm still working on determining an appropriate minimum PID limit value by running representative workloads and collecting data. Peter mentioned that this should not block the KEP for now, and we can continue refining the minimum value separately.

Could you please take another look at the KEP PR when you have a chance? Thanks

keps/sig-node/6063-pod-pid-limit/README.md
SergeyKanzhelev: /lgtm

thanks!

PR #6066, opened 2026-05-06
KEP-5073: document the tag lifecycle progression process, ownership, and lint rules

aaron-prindle:
- One-line PR description: This KEP update documents the maintainer process for progressing migrated declarative validation rules through Alpha -> Beta -> Stable — graduation requirements (~1 release soak + zero declarative validation mismatch/panic metrics), default ownership (PR author of the prior transition) with a backstop "DV rotation owner" role, exemption tag (`+k8s:validation-lifecycle-exempt`), and `validation-gen` lint rules for stale lifecycle tags.

- Issue link: https://github.com/kubernetes/enhancements/issues/5073

- Other comments: There are 2 open questions related to the tag lifecycle progression process outlined here (markeed w/ the <<[UNRESOLVED ..]>> wrapper in the PR currently).  They are: 
  - DV rotation owner selection: Starting point is one DV WG member per release owning this chosen ad-hoc in a oround robin. Seeking opinions on a more robust method for choosing this DV rotation member.
  - Addition of an`owner` field on alpha/beta tags:  Should we add an explicit `owner` field to the alpha/beta tags (similar to feature gates) so that ownership tracking does not depend on `git blame`?

**Summarized Important Points for the DV Tag Lifecycle Process:**
  - PR author who adds `+k8s:alpha(...)` owns the Alpha -> Beta graduation. The author who graduates to Beta owns Beta -> Stable. Ownership is not exclusive,  anyone can pick up cleanup.
  - New per-release role rotated across DV WG members to be backstop owner for alpha -> beta and beta -> stable transitions. They sweep lint warnings, file cleanup PRs, and route blocking errors when the original author is unavailable.
  - Each transition requires ~1 release of soak and zero `declarative_validation_mismatch_total` / `declarative_validation_panic_total` hits for that case during soak.
  - Beta -> Stable removes the handwritten code in the same PR.
  - Suppress stale-lifecycle warnings for a single field and must cite a tracking issue:
    ```go
    // +k8s:validation-lifecycle-exempt="tracking issue #12345: mismatch under investigation"
    // +k8s:alpha(since: "1.37")=+k8s:minimum=0
    ```
  - **Lint** in `validation-gen` errors on missing/malformed `since:` and warns (does not fail CI) on tags eligible for graduation.

### DIFF
--- a/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
+++ b/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
@@ -34,6 +34,9 @@
   - [Phase 2: v1.37 - v1.38 (Transition to Beta)](#phase-2-v137---v138-transition-to-beta)
   - [Phase 3: v1.39+ (Gate Removal &amp; GA)](#phase-3-v139-gate-removal--ga)
 - [Lifecycle: Promotion Process (Continuous)](#lifecycle-promotion-process-continuous)
+  - [Graduation Steps](#graduation-steps)
+  - [When Lifecycle Tags Apply](#when-lifecycle-tags-apply)
+  - [Lint Rules](#lint-rules)
   - [Example Walkthrough: Two-Field Migration](#example-walkthrough-two-field-migration)
     - [Phase 1: Legacy / Implicit Shadowing (v1.36 - v1.38)](#phase-1-legacy--implicit-shadowing-v136---v138)
     - [Phase 2: Explicit Onboarding (v1.39 / Release N)](#phase-2-explicit-onboarding-v139--release-n)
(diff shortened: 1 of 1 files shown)


keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
```diff
@@ -480,24 +484,121 @@ Our goal is to standardize on the Explicit Strategy and Lifecycle mechanism. The
...
+metrics):
+
+```diff
+-// +k8s:alpha(since: "1.37")=+k8s:minimum=0
++// +k8s:beta(since: "1.38")=+k8s:minimum=0
+ Replicas *int32 `json:"replicas,omitempty"`
+```
+
+DV is now authoritative; handwritten remains as the feature-gated
+fallback during Beta soak.
```
jpbetz: Post-GA, which feature gate is this?  Explaining the actual steps someone not deeply familiar with the system would take feels valuable here.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
```diff
@@ -480,24 +484,121 @@ Our goal is to standardize on the Explicit Strategy and Lifecycle mechanism. The
...
+
+**Step 3: Beta -> Stable**: (after >=1 release at Beta, zero mismatch/panic
+metrics, reviewers agree the safety switch is no longer needed):
+
+```diff
+-// +k8s:beta(since: "1.38")=+k8s:minimum=0
++// +k8s:minimum=0
+ Replicas *int32 `json:"replicas,omitempty"`
+```
```
jpbetz: This looks right.

How does this happen?  Does the PR that graduates the tag also drop all `k8s:beta` tags? Is there a checker that makes sure they get this right?

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
```diff
@@ -480,24 +484,121 @@ Our goal is to standardize on the Explicit Strategy and Lifecycle mechanism. The
...
+| API reviewer / approver | Check entry state, `since:`, tests, related versions; approve graduations and exemptions. |
+| SIG API Machinery approver/lead | Resolve exemptions and disputed graduations. |
+| **DV rotation owner (new)** | Per-release backstop: sweeps lint warnings via `hack/update-codegen.sh validation`, files Alpha -> Beta / Beta -> Stable cleanup PRs, and routes blocking lint errors to package/DV owners. |
+
+This mirrors prerelease lifecycle ownership: the person who adds or
+updates the metadata is the next planned editor, but the tag belongs to
+the API and anyone can move it forward. The DV rotation owner is a work
+assignment from the DV WG (one person named per release) and is the
+ultimate backstop for the workstream.
+
```
jpbetz: Supporting a list of owners for each tag that is similar in role expectations to feature gate ownership SGTM.  I don't think we need to say more here.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
```diff
@@ -480,24 +484,121 @@ Our goal is to standardize on the Explicit Strategy and Lifecycle mechanism. The
...
+  - `since: "X.YZ"` is required and must be a valid Kubernetes minor
+    version (error). Existing test usages without `since:` will be
+    addressed by either updating tests or accepting a sentinel like
+    `since: "dummy-version"`.
+  - Alpha/Beta tags one or more releases old emit per-tag warnings
+    (non-blocking) prompting graduation review.
+
+`hack/update-codegen.sh validation` uses `DefaultKubeBinaryVersion` as
+the current release for these checks.
```
jpbetz: I don't think we need to go this far.  How we implement validation (HW vs DV) does not feel like something we need to emulation version guard.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
aaron-prindle: Good point,  going to remove this section and then update the document (and the tags directly in the implementation PR) to have an "owners: <>" field and just explain the owners responsibility

jpbetz (approved):
/approve
/lgtm

This looks like a nice representation of what we've informally been doing.

/hold
For a few days to give others to weight in.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
aaron-prindle: Post-GA this is the `DeclarativeValidationBeta` feature gate. I've updated this section now to explain this in more detail and to explain the actual steps needed:

> DV is now authoritative. The handwritten code can be made authoritative (rolling back to +k8s:alpha functionality) by disabling the `DeclarativeValidationBeta` gate on kube-apiserver. This can be used to revert if a Beta migrated declarative validation rule misbehaves. After Stable, the wrapper is gone and there is no per-rule rollback (no hand-written code to rollback to, we confirmed the migration is safe at this point).

> **Rollback procedure** (applies during Beta soak only):
> - Disable `DeclarativeValidationBeta` the kube-apiserver: `--feature-gates=DeclarativeValidationBeta=false`
>   - With the gate off, every `+k8s:beta(...)`-wrapped rule in the cluster shadows rather than enforces. Handwritten validation becomes authoritative, restoring `+k8s:alpha`-equivalent behavior. The rollback using DeclarativeValidationBeta rolls back all instances of +k8s:beta, it is not per-rule
> - [optional] File an issue against the misbehaving rule. Keep it at `+k8s:beta(...)` (do not graduate to Stable). Add a `+k8s:validation-lifecycle-exempt` tag if the issue cannot be resolved within the next release.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
aaron-prindle: > Does the PR that graduates the tag also drop all k8s:beta tags?

No, currently the PR that graduates the tag would not drop all k8s:beta tags. If a tag (+k8s:newlyStable) is graduated to StabilityLevelStable, then this would then begin firing lint messages to upgrade all +k8s:beta=+k8s:newlyStable usages of that tag but ONLY for cases that have met the criteria of being soaked 1 release.  In theory it is possible that graduating a tag would effectively lead to a sweeping bump of all +k8s:beta=+k8s:newlyStable usages to Stable" if there was a number of existing +k8s:beta tag with soak but it is soft-enforced w/ lint rules that don't block CI and can be exempted with addition of the `+k8s:validation-lifecycle-exempt="<reason>" tag`.  

> Is there a checker that makes sure they get this right?
There is no static check that the handwritten removal is equivalent to the DV rule. The static lint does block one related failure mode: when the chained tag  to a removed wrapper (+k8s[alpha|beta]) is still below Stable (in a stable-versioned package) -> lint fails.  This enforces a rule cannot be silently graduated to Stable while its inner tag is still Beta-only.

For checking that the removed handwritten code is *semantically equivalent* to the DV rule, we rely on the mismatch checking and soak period: Migration Equivalency Tests run both HW and DV on every test case during Beta soak and fail on divergence; `declarative_validation_mismatch_total` increments on any production-traffic disagreement. Graduation requires zero mismatch hits over >=1 release of soak. There is no static check that the handwritten function body and the DV rule are semantically equivalent (would require symbolic analysis of arbitrary Go validation, intractable in general); behavioral evidence from soak is the load-bearing safety net.

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
aaron-prindle: Updated the doc to explain the +k8s:[alpha|beta] lifecycle tags now have an  "owners: <>" field which must be used when using the tags in k/k APIs

keps/sig-api-machinery/5073-declarative-validation-with-validation-gen/README.md
aaron-prindle: +1, dropping the line entirely as it relates to specific implementation details. Doc now just states the high level design that lint compares tag `since:` value against the current release.

jpbetz (commented):
/approve

PR #6068, opened 2026-05-10
 KEP-5677: second Alpha (1.37) for partitionable + consumable devices

nmn3m:
- One-line PR description: plan a second Alpha cycle in 1.37 to fix partitionable + consumable device accounting before graduating to Beta.

  - Issue link: https://github.com/kubernetes/enhancements/issues/5677


nmn3m:
/wg device-management
cc @johnbelamaric @kannon92

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,320 @@ in their namespace.
...
+| `pools[].unavailableDevices` | `*int32` (optional) | Count of physical devices with at least one `NoSchedule` or `NoExecute` taint. **0 in Alpha 1.36** (hard-coded); **computed from `ResourceSlice.Spec.Devices[].Taints` and matching `DeviceTaintRule`s in Alpha 1.37**. Unset when `validationError` is set. |
+| `pools[].validationError` | `*string` (optional, max 256 bytes) | Set when the pool's data could not be fully validated (e.g., incomplete slice publication). When set, count fields above may be unset. |
+| `pools[].counterSets[]` | atomic list of `CounterSetStatus`, max 32 (Alpha 1.37) | Per-`CounterSet` capacity / consumed / available, derived from `ResourceSlice.Spec.SharedCounters` and the `consumesCounters` of each non-AdminAccess allocation. Omitted when the pool has no shared counters. **Note:** this is a new `CounterSetStatus` type, not a reuse of the spec-side `CounterSet`. The spec's `Counter` only carries `Value` (inventory); a status-side type is needed to add the `consumed` and `available` fields without overloading the spec type. |
```
johnbelamaric: There are at most 32 counter sets per slice, but since a pool may contain many slices, there is no limit on actual countersets per pool. In the typical case of node-local resources, 32 is probably reasonable. I am considering using this for a broader, multi-node use case where I may need more.

I don't think we need to worry about it in alpha, we will know more later. We may eventually want to increase this, which I think we can do when we go to beta. We will need to put some cap on it, so no matter what the report will only contain "the first XX counter sets". I'm just not sure 32 is the right XX value.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,320 @@ in their namespace.
...
+| `pools[].counterSets[].name` | string (required) | Counter-set name as declared in `ResourceSlice.Spec.SharedCounters[].Name`. |
```
johnbelamaric: Reporting counter set information is for sure the most straightforward way to do this. But it's not the most user friendly. Generally that counterset concept is challenging and not very user-facing. An alternative would be to 1) let the driver specify an attribute name to count as the "device type"; 2) count up the number of each device type that is still allocatable based on the current counterset state.

Suppose then you had a device that could be consumed in 1/4, 1/2, or full sizes. If a single 1/4 partition of three full devices were taken, you could get a list like:

```console
Device Types Available:
  Full Device: 2 of 3
  Half Device: 5 of 6
  Quarter Device: 8 of 9
```

This may be expensive to calculate, though.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,320 @@ in their namespace.
...
+| `pools[].shareableDevices[]` | atomic list, max 256 (Alpha 1.37) | Per-device capacity / consumed / allocations for devices with `allowMultipleAllocations=true`. Omitted when the pool has no such devices. |
+| `pools[].shareableDevices[].name` | string (required) | Device name (matches `ResourceSlice.Spec.Devices[].Name`). |
+| `pools[].shareableDevices[].capacity` | `map[QualifiedName]DeviceCapacity` (required) | Mirrors the source `ResourceSlice.Spec.Devices[].Capacity` field; same `DeviceCapacity` wrapper as the spec so future fields (e.g. `RequestPolicy`) are forward-compatible. |
+| `pools[].shareableDevices[].consumed` | `map[QualifiedName]resource.Quantity` (required) | Sum of `DeviceRequestAllocationResult.ConsumedCapacity` across non-AdminAccess allocations on this device. Quantity rather than `DeviceCapacity` because the spec's per-result `consumedCapacity` is itself a plain `Quantity`. |
+| `pools[].shareableDevices[].allocations` | int32 (required) | Number of concurrent non-AdminAccess claims sharing this device. |
```
johnbelamaric: This results in listing all devices, that seems like it would get too big. Maybe we just list:
- Number of devices fully available (no consumption)
- Number of devices partially consumed

Then, maybe, for each defined capacity type: aggregate capacity total/available/consumed?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: The per-slice spec cap is `ResourceSliceMaxCounterSets = 8`, not 32, our 32 doesn't mirror anything upstream, it was a guess. Keeping the status cap at 32 for Alpha 1.37 as explicitly provisional with a note to revisit at Beta. Over-cap pools now produce a per-pool `validationError` instead of silent truncation, so operators see the loss.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: Adopted, replaced `shareableDevices[]` with `pools[].shareableSummary` (`fullyAvailable`, `partiallyConsumed`, plus `capacity[]` with per-key `total`/`consumed`/`available`). Verified the inputs are already-served fields (`Device.Capacity`, `ConsumedCapacity`), so the aggregate is a mechanical sum with no new driver dependencies. One caveat documented in the KEP: per-key totals are clean only for homogeneous schemas, for mixed pools, devices that don't carry a given key simply don't contribute to that key's total.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: Doable, the math is tractable and drivers already publish profile-like attributes (KEP-4815 uses `gpu.nvidia.com/profile`). The hard part is who declares the grouping axis, and the cleanest long-term answer (driver-declared at slice level) belongs in KEP-4815/5075 rather than here. Keeping raw `counterSets[]` for Alpha 1.37 as the durable primitive and treating the typed view as a follow-up, happy to file a tracking issue if useful.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,338 @@ in their namespace.
...
+   would misrepresent capacity.
+
+`availableDevices` keeps its existing definition
+(`totalDevices − allocatedDevices − unavailableDevices`). On plain
+pools it is the operationally useful "how many more claims fit"
+signal. **On pools with shared counters or shareable devices it is
+not.** Two cases the operator must understand:
+
+- **Partitionable pools.** When the bottleneck is a shared
+  `CounterSet`, all device entries can be unallocated yet no further
```
johnbelamaric, line 982: I know what you have here is what we discussed after the last review. But as I read it, I am considering that it still ends up with an entry for every physical device in the pool, in the MIG case at least. I suspect we might end up wanting to change it. I am curious if others have feedback for that.

The alternative is what we discussed briefly: 1) driver publishes a grouping attribute key (e.g., 'partitionType'); 2) summary counts number of that partition type allocatable in the pool. So the total entries per pool would be the cardinality of that grouping attribute.

cc @shivamerla @pohly @klueska

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -637,74 +878,239 @@ None required.
...
+   `allocatedDevices` per allocation result, which (a) overcounts
+   on devices with `allowMultipleAllocations=true` (consumable) and
+   (b) does not reflect shared-counter consumption on partitionable
+   devices. The visible symptom is `availableDevices=0` reported on
+   pools that actually have free capacity. Fixing this requires new
+   API fields (`counterSets[]`, `shareableSummary`), not just a
+   controller patch — and adding new API surface in Beta is exactly
+   what Alpha cycles exist to avoid. This is the load-bearing
+   reason; the points below are supporting evidence.
+2. **No production DRA driver has integrated yet.** The original
```
johnbelamaric: What integration is required? I don't think there is any?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,338 @@ in their namespace.
...
-    generation: 5                 # Pool generation observed
-
-  # Conditions indicating processing status
+    unavailableDevices: 0         # 0 in Alpha 1.36; computed from device taints in Alpha 1.37
+  - driver: example.com/gpu
+    poolName: node-2
+    generation: 5
+    # validationError is set when a pool is incomplete (observed < expected
+    # slice count). When set, count fields are unset. Max 256 bytes.
+    validationError: "pool example.com/gpu/node-2 is incomplete: observed 1/2 slices at generation 5"
```
troychiu, line 591: The above mentions that the controller requeues the request if a `validationError` is set. However, the section below states that once `status` is populated, the entire object (including `metadata` and `spec`) is immutable. Could you help me understand how this requeue process functions?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +536,338 @@ in their namespace.
...
-  # Timestamp when calculation was performed
-  observationTime: "2026-02-07T10:30:00Z"
+  # Total number of pools that matched the filter (even if the response is
+  # truncated by `limit`). If 0, no pools matched.
+  poolCount: 2
 
-  # List of pools matching the filter
+  # First `spec.limit` matching pools, sorted by driver then pool name.
+  # If len(pools) < poolCount, the response was truncated.
   pools:
```
troychiu, line 576: It would be great if we can include pools with counterSets and consumable devices here

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -339,76 +360,106 @@ spec:
...
+
+   - **Partitionable** (`DRAPartitionableDevices` feature gate): a
+     single physical device may appear as multiple mutually-exclusive
+     partitions that share a `CounterSet`. Counting devices ignores
+     the shared bottleneck.
+   - **Consumable** (`DRAConsumableCapacity` feature gate): a device
+     with `allowMultipleAllocations=true` may serve many claims
+     simultaneously. Counting each claim against `allocatedDevices`
+     drives `availableDevices` to 0 on pools that still have free
+     capacity (the `max(0, …)` floor in the controller hides the
```
sunya-ch, line 412: Is it possible to expand the device capacity field in the status report? We would also like to track the remaining capacity, potentially as an aggregated value.
`max(0, …)` may mislead in the case that we have two shared devices, one allocated by two resource claims half each. The available is 0 when the other shared device is fully free.

Instead, we may define allocatedDevices as unique allocated devices.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: The requeue happens *before* `status` is ever written: when any pool has a `validationError`, the controller returns early without calling `UpdateStatus`, and the workqueue retries (up to 5). Status stays nil through retries, and only the attempt that finds all pools complete (or the final attempt before the queue drops it) writes status. Immutability kicks in **after** that write, so the requeue path and the immutability rule never overlap on the same object state.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: That's exactly the behaviour Alpha 1.37 lands: `allocatedDevices` is capped at 1 per physical device (each device counted once regardless of how many non-AdminAccess claims it holds), and aggregated remaining capacity is exposed via `pools[].shareableSummary.capacity[].{total, consumed, available}`. In your "two shared devices, one half-allocated by two claims" case, the new behaviour reports `allocatedDevices=1`, `availableDevices=1`, plus the per-key headroom on the consumed device, see the "Partitionable & Consumable Device Accounting" subsection.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: Wrong word on my part, I meant "validated with at least one production DRA driver(out-of-tree testing)", which was the original Beta gate.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
sunya-ch: @nmn3m I'm sorry—I missed the earlier explanation and got confused by this note.
Would it be helpful to separate the addressed issues into subsections for clarity?

It might be also helpful to include an internal link to the relevant section like [Partitionable & Consumable Device Accounting](#partitionable--consumable-device-accounting).

sunya-ch:
@nmn3m Accounting of consumable capacity in `shareableSummary` and cap-at-1 rule for allocatedDevices make sense to me. 

---

Additional comments:

In my opinion, explicitly adding the word "Devices" to `fullyAvailable` and `partiallyConsumed` in the same way with `availableDevices` would make it clearer that these fields refer to the device objects rather than their capacity.

Additionally, I suggest using partiallyAvailable instead of partiallyConsumed to keep the naming convention consistent and focused on resource availability rather than consumption.

My suggestion is :
-`fullyAvailable` --> `fullAvailableDevices`
-`partiallyConsumed` --> `partiallyAvailableDevices`.

I wonder about the accounting when consumable capacity is used together with partitionable device. 
I believe that `availableDevices` will excludes the devices those cannot be allocated due to `counterSet` constraints and that should also be considered in shareableSummary section too but I cannot see it addressed in the Partitionable & Consumable Device Accounting section. 

For example, partitionable device allows one full device or two half devices. Full devices have consumable 80Gi, and half device have 40Gi each.
If 60Gi of full device has been allocated, two half devices should not be counted in availableDevices.
The available capacity should account only 20Gi not 100Gi.

cc: @johnbelamaric

nmn3m:
Thanks @sunya-ch 
Both renames adopted,  `fullyAvailableDevices` / `partiallyAvailableDevices`.
On the partitionable + consumable case: correct that a consumed counter can block sibling partitions, and Alpha 1.37 doesn't net that out of `availableDevices` / `shareableSummary` — `counterSets[].counters[].available` is the authoritative bound there (your 80/60→20Gi example). Documented as a known limitation in the new [Devices That Are Both Partitionable and Consumable](#devices-that-are-both-partitionable-and-consumable) subsection; folding both into one number is deferred for a soak cycle.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
johnbelamaric: This also avoids all the detailed counter information ending up in the summary. Right now, for MIG, I think we end up with a lot of data - each GPU has an entry with each counter in that explicitly listed. I think there are quite a few counters in the counterSet for a GPU, aren't there? Or is it just a couple?

If we wait on this, it means another alpha, I'd rather not wait and have to change it again, so it would be good to come to a decision. I think the sticking point is how/where we declare the grouping key. This is a per pool attribute. I think we could add it to ResourceSlice and just consider it an error if there is more than one value for a pool. In fact, there should be exactly one value for any pool that uses sharedCounters. The best practice could be to add it to the ResourceSlice that defines the counter sets. In fact we could add a validation such that it can only be set if sharedCounters is set. We would need to have some fallback though for how to handle it if the grouping key is not set.

nmn3m: Added ResourceSlice.Spec.PartitionTypeAttribute (*FullyQualifiedName) plus a typed partitionSummary[] (type / total / allocatable); allocatable[T] = min(freshDevices[T], min_c floor(counterAvailable[c] / cost[T][c])) with counterAvailable debited once per in-use device, matches the scheduler's skipCounterCheck for shareable partitions at allocator_incubating.go:1335, so the hybrid allowMultipleAllocations case nets out
correctly and @sunya-ch example resolves directly. 

MIG goes from one entry per GPU to one per profile; counterSets[] stays as a fallback when the attribute isn't declared. 

ping @pohly @klueska

kannon92:
/assign @Champbreed 
For PRR shadow

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+| `pools[].shareableSummary.capacity[].name` | string (`QualifiedName`, required) | Capacity key as it appears in `ResourceSlice.Spec.Devices[].Capacity`. |
+| `pools[].shareableSummary.capacity[].total` | `resource.Quantity` (required) | Sum of `Device.Capacity[name].Value` across all shareable devices in the pool that carry this key. Devices that do not carry the key contribute nothing (rather than zero), which is the correct behaviour for heterogeneous-schema pools. |
+| `pools[].shareableSummary.capacity[].consumed` | `resource.Quantity` (required) | Sum of `DeviceRequestAllocationResult.ConsumedCapacity[name]` across non-AdminAccess allocations on shareable devices that carry this key. |
+| `pools[].shareableSummary.capacity[].available` | `resource.Quantity` (required) | `total − consumed`, clamped at zero (never negative). |
```
troychiu: Would it be useful to display the maximum capacity that a single device provides for a specific capacity key? Users might need to know if they are requesting too much or if there is another reason why a claim cannot be allocated.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -339,76 +362,113 @@ spec:
...
 ### Notes/Constraints/Caveats
 
 1. **Asynchronous operation**: Unlike SubjectAccessReview (synchronous), this
    uses the CSR pattern where user must wait for controller processing.
 
-2. **One-time calculation**: Each request is processed once. To get updated
-   data, delete and recreate the request.
+2. **One-time calculation**: Each request is processed once. Once `status`
+   is set the entire object (metadata included) becomes immutable. To get
+   updated data, delete and recreate the request.
```
pohly: What's the rationale for locking the metadata? Does this really include things like labels? This seems a bit odd to me.

Is this a proposal or has it been implemented like that?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+// groups devices in this pool by partition type. When declared on
+// every slice in a pool (all with the same value), the resource
+// pool status controller emits a typed partitionSummary[] view
+// instead of the verbose counterSets[] dump. Leaving the field
+// unset on every slice in a pool keeps the counterSets[] fallback;
+// a mix of set and unset across slices in the same pool, or
+// different values, produces a per-pool validationError.
+//
+// Only meaningful when SharedCounters is set somewhere in the
+// pool. Declarative validation rejects this field only when the
```
pohly:
```suggestion
// pool. Validation rejects this field only when the
```

I doubt that this can be done with declarative validation; you'll have to write manual validation for it.

Either way, the user of the API (whom this text is written for) won't care about this distinction.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+#### Companion API Change: `ResourceSlice.Spec.PartitionTypeAttribute`
+
+Alpha 1.37 adds one optional field to `ResourceSliceSpec`
+(`staging/src/k8s.io/api/resource/v1/types.go`):
+
+```go
+// PartitionTypeAttribute names a device attribute whose value
+// groups devices in this pool by partition type. When declared on
+// every slice in a pool (all with the same value), the resource
+// pool status controller emits a typed partitionSummary[] view
```
pohly: This notation with trailing `[]` looks odd. I think you want to indicate that this is a slice. I think you can drop this everywhere, it's not a common notation.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+| `conditions[]` | map list by `type`, max 10 | `Complete` (True when processed) or `Failed` (True on error). |
+
+#### Companion API Change: `ResourceSlice.Spec.PartitionTypeAttribute`
+
+Alpha 1.37 adds one optional field to `ResourceSliceSpec`
+(`staging/src/k8s.io/api/resource/v1/types.go`):
+
+```go
+// PartitionTypeAttribute names a device attribute whose value
+// groups devices in this pool by partition type. When declared on
```
pohly: What is a "partition type"? Which devices in the slice have to have this attribute? What attribute types are valid for it? Is that validated?

As it stands right now I am not sure whether DRA driver authors will understand how they are supposed to use this. Perhaps include an example? The first paragraph should introduce this concept and explain it.

The second paragraph can then explain how this is used by the resource pool status controller, but only briefly: APIs typically don't document all the ways how they are used, only what they define.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+
+Feature-gate interaction: the field is gated behind
+`DRAResourcePoolStatus`, but it is only effective when
+`DRAPartitionableDevices` is also enabled (since the field is only
+meaningful for pools with `SharedCounters`, which is itself gated by
+`DRAPartitionableDevices`). When only one gate is enabled, the field
+is silently dropped at write time (gate off) or accepted but ignored
+by the status controller (`DRAResourcePoolStatus` off).
+
+Validation rules (declarative on the slice, plus controller-side
```
pohly: Validation constraints should be described in the API above, not the KEP.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+  by `PartitionTypeAttribute`. Devices missing the attribute produce
+  a per-pool `validationError` rather than being silently bucketed.
+- **Pool-level: homogeneous `ConsumesCounters` per partition type.**
+  Devices in the same partition-type group must share the same
+  `ConsumesCounters` cost; mixed costs in one group produce a
+  per-pool `validationError` since the `allocatable[T]` math
+  assumes a single canonical cost per type.
+
+Cross-KEP relationship: the field lives on the `ResourceSlice` shape
+owned by [KEP-4815](/keps/sig-scheduling/4815-dra-partitionable-devices),
```
pohly: What do you mean with "the `ResourceSlice` shape owned by KEP-4815"? What is this "shape"?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -485,148 +545,561 @@ in their namespace.
...
+| `pools[].validationError` | `*string` (optional, max 256 bytes) | Set when the pool's data could not be fully validated (e.g., incomplete slice publication). When set, count fields above may be unset. |
+| `pools[].partitionSummary[]` | atomic list of `PartitionTypeStatus`, max 32 (Alpha 1.37, **provisional** — revisit at Beta) | Per-partition-type aggregate, emitted when the pool's slices declare `ResourceSlice.Spec.PartitionTypeAttribute` (new in 1.37). Each entry reports `type` (the value of the declared attribute on devices in this group), `total` (count of devices in the pool with this value), and `allocatable` (how many more devices of this type can still be allocated given current shared-counter state). Devices in the pool whose `ConsumesCounters` cost differs from peers of the same type produce a per-pool `validationError` rather than silent skew. Cap of 32 is a provisional starting point that fits MIG-class pools (3–7 partition types typical); over-cap pools produce a per-pool `validationError` instead of silent truncation. **Mutually exclusive with `counterSets[]`** — a pool emits one view or the other depending on whether the driver declared a grouping attribute. |
```
johnbelamaric: probably a validation error if type is missing on any device, too

keps/sig-node/5677-dra-resource-availability-visibility/README.md
johnbelamaric: @nmn3m the object was introduced in the structured parameters KEP, but KEPs don't really "own" an object. There is nothing wrong (and it is very very common) with adding a field to and existing object, you don't need to call out the other KEP here. You can put it in the "see also" section, which I think you've already done.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: The implementation freezes only status, metadata uses standard `ValidateObjectMetaUpdate`, so labels/annotations remain mutable. The "(metadata included)" wording in the KEP is wrong,  I'll drop it.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: @johnbelamaric WDYT? Or should I defer it to BETA?

Champbreed:
PRR shadow

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -339,76 +362,114 @@ spec:
...
 
 1. **Asynchronous operation**: Unlike SubjectAccessReview (synchronous), this
    uses the CSR pattern where user must wait for controller processing.
 
-2. **One-time calculation**: Each request is processed once. To get updated
-   data, delete and recreate the request.
+2. **One-time calculation**: Each request is processed once. Once `status`
+   is set it becomes immutable; metadata (labels, annotations) follows
+   the standard object-meta update rules. To get updated data, delete and
+   recreate the request.
```
Champbreed, line 385: This statement says metadata (labels, annotations) follows standard object-meta update rules — meaning metadata remains mutable after status is set.
However, two other sections of this same document say the opposite:
- API Definition section (after the YAML example): "Once status is populated the entire object (including metadata and spec) is immutable; update requests are rejected by the API server."
- One-time Processing section (Controller Implementation): "Once status is written, the entire object is immutable (spec, metadata, and status all rejected for update by the registry strategy / validation)."
All three statements describe the same object behavior after status is set.

 This section says metadata stays mutable. The other two say it is immutable. They cannot all be correct. Which is the intended behavior?

keps/sig-node/5677-dra-resource-availability-visibility/README.md
johnbelamaric: Nice, thank you.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
```diff
@@ -875,10 +1676,10 @@ No impact on scheduling or pod startup.
 ###### Will enabling / using this feature result in non-negligible increase of resource usage?
 
 Minimal:
-- etcd: Small objects, users should clean up (TTL in Beta)
-- KCM: Reuses existing informers, adds small controller
+- etcd: Small objects, bounded by built-in TTL cleanup (Alpha: 1h completed / 24h pending)
+- KCM: Reuses existing `resource.k8s.io/v1` informers for ResourceSlice and ResourceClaim, adds a small controller with its own work queue
 - API server: Standard API operations
-- Response size: Bounded by required `driver` field (one driver's pools) and optional `limit` field (values determined at implementation based on size calculations)
+- Response size: Bounded by the required `driver` field (one driver's pools), the `limit` field (default 100, max 1000), the `+k8s:maxItems=1000` constraint on `status.pools`, and (for Alpha 1.37) `+k8s:maxItems=32` on each of `partitionSummary`, `counterSets`, and `shareableSummary.capacity` per pool
```
Champbreed, line 1689: @nmn3m, The transition from the unbounded `shareableDevices[]` array over to the flattened `shareableSummary` aggregate counters addresses the O(N) object bloat risks, and the strict validation limit of 32 entries on the `partitionSummary` array via `+k8s:maxItems=32` provides a clear schema-level ceiling. 

I just wanted to clarify one quick boundary condition regarding how the KCM controller behaves when aggregating these limits. Since the `+k8s:maxItems=32` cap protects individual objects on write, what happens internally within the KCM status controller if it processes multiple independent, valid `ResourceSlice` objects from a high-density hardware cluster that collectively exceed 32 distinct partition types when combined? Does the controller gracefully truncate its internal evaluation array to the provisional 32-item ceiling and log a specific warning, or will it attempt to pass an over-cap array back to the status subresource?

nmn3m: @Champbreed 
The controller never passes an over-cap array. As documented at L683/L687, when aggregation across slices in a pool yields more than 32 distinct partition types, the controller sets `pools[i].validationError` on that pool and emits neither `partitionSummary` nor `counterSets` for it.
This routes through the same per-pool `validationError` already used for incomplete-pool / inconsistent-cost cases, so operators see an explicit signal instead of silent truncation. The 32 cap itself is provisional for Alpha 1.37 and explicitly flagged for revisit at Beta.

keps/sig-node/5677-dra-resource-availability-visibility/README.md
nmn3m: The "entire object" wording at L648-650 and L819-820 was stale and contradicted L383-384.
 Fixed both

keps/sig-node/5677-dra-resource-availability-visibility/README.md
Champbreed: Thanks for the clear breakdown, @nmn3m. Since the over cap case routes through the same `validationError` path used for incomplete slice publication, will the controller populate a distinct error message signal or unique reason text within that field?
Operators diagnosing a pool in an error state will need to quickly know whether they are hitting a provisional scaling ceiling or a transient driver synchronization issue, since the remediation steps for those two states are entirely different.

nmn3m: Yes, the controller will use stable, machine-readable prefixes on the validationError string per condition, e.g. `PoolIncomplete:, PartitionSummaryOverCap:, CounterSetsOverCap:, PartitionTypeMissing:, PartitionCostMismatch:`. 
That keeps Alpha 1.37 schema-stable while still letting operators grep / alert on the specific case. 

@Champbreed Do you want me to add any specific points about that?

Champbreed: Yes, @nmn3m. It would be helpful to include a provisional note listing those current prefixes directly within the pools.validationError field description in the Status Fields table. 
Documenting them as reference behavior ensures operators testing the Alpha cycle know exactly what strings to expect for early alerting and grepping without having to dig through the controller implementation source code.

nmn3m: @Champbreed , Added.

Champbreed:
LGTM 
cc: @kannon92

johnbelamaric (commented):
@nmn3m this looks great. Main content LGTM. I am holding of on PRR section to give shadows a chance to review first.

kannon92:
> @nmn3m this looks great. Main content LGTM. I am holding of on PRR section to give shadows a chance to review first.

@johnbelamaric technically PRR approved this in alpha for 1.36. The shadow took a look but I don't see my approval needed here.

Right now, its really just waiting on sig approval.

nmn3m:
> > @nmn3m this looks great. Main content LGTM. I am holding of on PRR section to give shadows a chance to review first.
> 
> @johnbelamaric technically PRR approved this in alpha for 1.36. The shadow took a look but I don't see my approval needed here.
> 
> Right now, its really just waiting on sig approval.

Depending on that 
/assign @dchen1107

johnbelamaric:
/assign @mrunalp 

He has been doing the DRA SIG Node KEPs.

/unassign @dchen1107

mrunalp (approved):

petr-muller:

PR #6071, opened 2026-05-12
KEP-5966: RangeStream kubernetes integration

Jefftree:
KEP 5966: RangeStream kubernetes integration

etcd has already been implemented https://github.com/etcd-io/etcd/issues/21619

etcd benchmarks: https://github.com/etcd-io/etcd/pull/21701

This update documents the Kubernetes integration.

watch cache benchmarks can be found https://github.com/kubernetes/kubernetes/issues/138815.

### DIFF
--- a/keps/sig-etcd/5966-etcd-range-stream/README.md
+++ b/keps/sig-etcd/5966-etcd-range-stream/README.md
@@ -41,18 +41,18 @@
 
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+- [X] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [X] (R) KEP approvers have approved the KEP status as `implementable`
+- [X] (R) Design details are appropriately documented
+- [X] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
   - [ ] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
+- [X] (R) Graduation criteria is in place
   - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
-- [ ] (R) Production readiness review completed
-- [ ] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
+- [X] (R) Production readiness review completed
+- [X] (R) Production readiness review approved
+- [X] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
@@ -74,6 +74,9 @@ This KEP proposes a new server-streaming `RangeStream` RPC that reuses
 internally with adaptive chunk sizing and pins to a single MVCC revision for
 consistency. The total count is derived from the running tally of streamed
 keys, eliminating the separate index traversal required by paginated Range.
+
+The feature requires etcd 3.7+. With older etcd, kube-apiserver detects
+`Unimplemented` and continues using unary `Range` with no behavior change.
 
 ## Motivation
 
@@ -127,14 +130,16 @@ should retry the request.
 
 ### Risks and Mitigations
 
-- **etcd version skew.** Older etcd versions return gRPC `Unimplemented`.
-  kube-apiserver detects this at runtime, logs a warning, and falls back
-  to the existing paginated list path.
 - **Stream interrupted by compaction.** If the pinned revision is compacted
   mid-stream, the server returns `ErrCompacted`. kube-apiserver surfaces
   this as a watch cache initialization failure, and the cacher retries
   initialization. The behavior is identical to a paginated list that races
-  compaction.
+  compaction. A unary `Range` without pagination technically avoids this
+  problem, because a single non-streaming call does not pin a revision
+  across multiple reads. But at the scale where a stream takes longer than
+  kube-apiserver's compaction interval (5 minutes by default) to complete,
+  the unary response approaches protobuf's 2 GB message limit and is
+  itself unlikely to succeed.
 
 ## Design Details
 
@@ -217,6 +222,10 @@ The following components are modified:
 - **Client** (`client/v3/kv.go`): `RangeStreamToRangeResponse` — reassembles
   a stream into a single `RangeResponse` so callers can transparently switch
   between unary and streaming.
+- **etcdctl**: A `--stream` flag on `etcdctl get` issues a `RangeStream`
+  request and reassembles the chunks into a `RangeResponse` for display.
+  This provides a direct CLI entry point, useful for inspections and
+  comparisons against unary `get` on the same key range.
 
 ### Test Plan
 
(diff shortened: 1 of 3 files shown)


keps/sig-etcd/5966-etcd-range-stream/kep.yaml
```diff
@@ -15,3 +15,20 @@ approvers:
...
+
+stage: beta
+
+latest-milestone: "v1.37"
+
+milestone:
+  beta: "v1.37"
+
+feature-gates:
+  - name: RangeStream
```
serathius: There are a lot of feature gates done on every level of K8s stack, we should be more specific which RangeStream we mean here. Recently CRI introduced streaming, we don't want to be mixed up.

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +278,221 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
 
+### Version Skew Strategy
+
+Skew is between kube-apiserver and etcd. The supported combinations:
+
+- gate on + etcd 3.7 or newer: uses `RangeStream`.
+- gate on + etcd 3.6 or older: falls back to unary `Range`.
+- gate off + any etcd version: uses unary `Range`.
+
+kube-apiserver caches the result of feature discovery rather than
```
serathius: Clarify if we want to implement new mechanism or use the existing mechanism. 

I would recommend not to reinvent the wheel if not needed.
Context: https://github.com/kubernetes/kubernetes/pull/124483 from https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/2340-Consistent-reads-from-cache/README.md

Jefftree: Clarified to directly reference the feature checker.

keps/sig-etcd/5966-etcd-range-stream/kep.yaml
Jefftree: renamed to `EtcdRangeStream`.

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +287,229 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
+
+#### Beta
+
+- Feature implemented behind the `EtcdRangeStream` feature gate, default on.
+- Watch cache initialization routes through `RangeStream` with
+  automatic `Unimplemented` fallback for older etcd.
+- Feature is covered with unit, integration, and e2e tests.
+- Scalability test on a 5000-node cluster measuring latency on large
+  list responses.
+- `etcdctl get --stream` available for direct CLI access to `RangeStream`.
```
Jefftree, line 302: @deads2k Adding this as a beta blocker

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -86,6 +128,19 @@ should retry the request.
   avoids the bbolt caveat where long-running read transactions can block write
   transactions when the database needs to remap/allocate new pages.
 
+### Risks and Mitigations
+
+- **Stream interrupted by compaction.** If the pinned revision is compacted
+  mid-stream, the server returns `ErrCompacted`. kube-apiserver surfaces
+  this as a watch cache initialization failure, and the cacher retries
+  initialization. The behavior is identical to a paginated list that races
+  compaction. A unary `Range` without pagination technically avoids this
```
Jefftree, line 137: Specifically calling out this risk that a unary range could succeed in compaction cases, but it is likely rare given that we pretty much paginate everywhere in k8s list path.

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +287,229 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
+downgrade), the `Unimplemented` fallback will still fall back to
+`Range`.
+
+For HA kube-apiserver during a rolling upgrade, each apiserver
+negotiates with etcd on its own. An apiserver on the old version uses
+unary `Range` and one on the new version uses `RangeStream` when
+available. Watch cache state is per-apiserver, so they don't interact.
+For an etcd cluster rolling upgrade, requests may land on different
+members during the transition. Once a `RangeStream` call fails,
+kube-apiserver falls back to unary `Range`. For an etcd downgrade
```
jpbetz, line 340: If an kube-apiserver has the gate on and etcd is upgraded from 3.6 to 3.7 while the kube-apiserver is running? The apiserver continues to use `Range` until the next time it is restarted at which point it will start to use `RangeStream`?

EDIT: I ask because it cluster operators that want to ensure a upgrade is fully complete would need to restart their kube-apiserver as part of their upgrade in this case.

jpbetz: Nvm, looks like `storage.FeatureSupportChecker` is run periodically. Let's document that here briefly.

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +287,229 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
+Yes. Setting `EtcdRangeStream=false` and restarting kube-apiserver fully
+disables the feature. The feature is stateless.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+Same as enabling the feature for the first time.
+
+###### Are there any tests for feature enablement/disablement?
+
+Yes, testing will be added for parity of watch cache behavior with RangeStream on and off.
```
jpbetz: How will we test skew with versions of etcd that don't have the feature?

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +287,229 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+- **etcd does not support `RangeStream`.** kube-apiserver detects
+  `Unimplemented` and falls back to paginated list. No impact on
+  running workloads.
+- **etcd compacts the pinned revision mid-stream.** kube-apiserver
+  surfaces `ErrCompacted` to the cacher, which retries
+  initialization. This is identical to a paginated list racing
+  compaction. Watches that were already established are unaffected.
```
jpbetz: Let's move these elsewhere.  The first is not an observable failure, it's just a mechanism that manages the protocol switch. The second is a not rollout/rollback specific.  Both are good to document, but I think the answer to this specific PRR question is that there are no known failure modes if the KEP is implemented correctly (the protocol switches handle all cases).

keps/sig-etcd/5966-etcd-range-stream/README.md
```diff
@@ -205,14 +287,229 @@ decode. When the server returns `Unimplemented` or the feature gate is
...
+
+###### What specific metrics should inform a rollback?
+
+- `apiserver_watch_cache_initialization_duration_seconds{group, resource}`
+  increases significantly compared to average.
+- `etcd_request_duration_seconds{operation="listStream"}` p99 reports extremely high numbers.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+n/a.
```
jpbetz: This is needed. The skew testing and protocol switches (both `storage.FeatureSupportChecker` and `Unimplemented` responses) need some form of coverage.  We need some kind of assurance that the upgrade->downgrade->upgrade path has been shown to work.

keps/sig-etcd/5966-etcd-range-stream/README.md
Jefftree: Updated.

jpbetz (commented):
Did a PRR pass

keps/sig-etcd/5966-etcd-range-stream/README.md
Jefftree: Removed.

keps/sig-etcd/5966-etcd-range-stream/README.md
Jefftree: 1. manual testing with the etcd binaries
2. mocking responses in certain unit tests to ensure apiserver behaves correctly

Added these notes.

keps/sig-etcd/5966-etcd-range-stream/README.md
Jefftree: Added.

jpbetz:
/approve
For PRR

jpbetz:
We (API Machinery leads) discussed this and are supportive of implementing this KEP. There is an open question about if this should be a on-by-default beta, but we can start implementation under a off-by-default gate and discuss that async.

serathius:
/lgtm
/approve

The etcd side didn't change, this PR is just changing K8s integration details.

soltysh:
@jpbetz @Jefftree make sure to get that KEP opted-in for 1.37, it doesn't seem to be tracked currently.

jpbetz:
> @jpbetz @Jefftree make sure to get that KEP opted-in for 1.37, it doesn't seem to be tracked currently.

cc @serathius as this one is filed under sig-etcd

PR #6073, opened 2026-05-12
KEP-6072: DRA: Standard numaNode Device Attribute

johnahull:
## Summary

Standardize `resource.kubernetes.io/numaNode` as a well-known DRA device attribute alongside `pcieRoot` and `pciBusID`. Adds helper functions to the `deviceattribute` library.

- `pcieRoot` measures bus topology (which PCIe switch). `numaNode` measures memory topology (which memory controller). Orthogonal signals.
- Several DRA drivers, NVIDIA and AMD GPU in particular, publish numa already with different names. One standard name enables cross-driver `matchAttribute` co-placement. CPU, SRIOV, and memory do also. 
- Given today's HW configurations and topology, not all GPU/NIC matches with matchAttribute will be possible with pcieRoot alone. There is hardware where no pcie slots share the same pcieRoot. There are many configs where NIC VF's are shared across pcieRoots to other GPUs. 

Tracking issue: https://github.com/kubernetes/enhancements/issues/6072

/sig node
/wg device-management
/kind kep


keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,374 @@
...
+## Summary
+
+Standardize `resource.kubernetes.io/numaNode` as a well-known DRA device attribute alongside the existing `pcieRoot` and `pciBusID`. This attribute identifies which memory controller services a device — an orthogonal signal to `pcieRoot` (which identifies the PCIe switch tree). Adding a shared helper function in the `deviceattribute` library and having DRA drivers publish this attribute under a common name enables cross-driver NUMA co-placement via a single `matchAttribute` constraint.
+
+Today, six DRA drivers publish NUMA node information under five different vendor-specific attribute names. `matchAttribute` requires a common name. Users cannot write a cross-driver NUMA co-placement constraint without middleware. This proposal standardizes the name so that one constraint works across all drivers:
+
+```yaml
+constraints:
+- matchAttribute: resource.kubernetes.io/numaNode
+  requests: [gpu, nic, cpu, mem]
```
swatisehgal, line 83: Do we want to embed maturity level in the attribute name? Since this attribute is introduced as part of an alpha feature, it’s not entirely clear whether the attribute itself should be considered alpha as well, or whether it implicitly carries GA-like expectations once standardised. In practice, attributes like `numaNode` will likely evolve as we learn more from real deployments and implementation in diverse hardware. Clearly indicating that an attribute is experimental in semantics and expectations gives us the ability to evolve and iterate while still enabling cross-driver interoperability.

I had a discussion with @ffromani who suggested the idea of a staging registry for experimental attributes. As a lighter-weight alternative, we could consider embedding maturity metadata directly into the attribute definition to make its lifecycle and expectations clearer without adding too much process overhead.

swatisehgal (commented):
NUMA alignment is useful, but it only captures one dimension of hardware locality. Real systems increasingly have more complex topologies (multi-socket systems, heterogeneous accelerators, varying PCIe hierarchies), where same NUMA node is often necessary but not sufficient for optimal placement. In that sense, this proposal feels like it improves the current baseline but still stays within an equality-based model.

I have been thinking about longer-term expressiveness where equality on a single (or a small set of) attributes can be limiting. We were having some discussion on this topic here https://github.com/kubernetes/enhancements/issues/5213#issuecomment-4450881222.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,374 @@
...
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation — e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Standardize `resource.kubernetes.io/numaNode` as a well-known DRA device attribute alongside the existing `pcieRoot` and `pciBusID`. This attribute identifies which memory controller services a device — an orthogonal signal to `pcieRoot` (which identifies the PCIe switch tree). Adding a shared helper function in the `deviceattribute` library and having DRA drivers publish this attribute under a common name enables cross-driver NUMA co-placement via a single `matchAttribute` constraint.
```
kad: The problem is that in modern CPUs the relation between device and memory controller is not 1-to-1.
Modern CPU physical package have one or more I/O dies where several PCIe Root Ports are located.
Those I/O dies are connected to multiple memory controllers which can be operating as single logical unit (1 numa node = 1 socket) or multiple logical units (multiple numa nodes per socket). Bandwidth of memory controllers and interconnect between memory controllers and I/O dies are often way higher than speed of single PCIe port behind which device is located. So, this attribute to be meaningful for hardware sold in past 5+ years would need to be a list, denoting which group(s) of memory controllers are considered the closest to the PCIe root where device is connected.

johnahull: @kad I put this on the slack discussion as well.  I went through the kernel source to see what topology data is actually available for PCI devices. What the kernel exposes:
  * numa_node — a single integer from firmware _PXM, inherited by every device behind the root bridge
  * local_cpulist — derived from the same NUMA node, not an independent signal
  * PCIe bus hierarchy — root complex / bus / device / function
  * cpuSocketID — derived: numa_node → cpulist → physical_package_id
  * SLIT distances — firmware pass-through, NUMA-to-NUMA only. The kernel explicitly skips I/O proximity domains (srat.c: "I/O localities are not supported at this point")

I'd love to use the "richer" topology that you're describing, but the kernel doesn't expose it yet, AFAIK. e.g. IO die identity, per-device distance vectors, etc. CPU die_id exists in sysfs; PCI devices have no equivalent. If I'm missing something, I'm certainly open to suggestions/corrections. 
  
Maybe we should standardize CPU socket ID also? On shared I/O die HW (1 I/O die + 4 compute dies), GPU placement can use CPU socket ID, since the device is equidistant to all dies on the socket. CPU-memory affinity still uses numaNode. AFAICT (I don't have HW to test), on multi-I/O-die hardware (e.g., Intel Granite Rapids with 2 HSIO dies per socket), each HSIO die gets its own PCIe root complexes and different _PXM values. So pcieRoot distinguishes devices on different I/O dies, and numaNode distinguishes their closest compute dies.

None of the attributes claims to describe I/O die affinity or cross-component bandwidth. If the kernel gains richer interfaces going forward, then we can add more attributes to better describe the topology. Right now all we have is pcieRoot to match against. If we have numanode and cpu socket id, then we at least give users options for how they want to place devices.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: I agree that NUMA only captures one dimension. I expect that there would be others in the future, especially based on what @kad has outlined, and that numa node will be just one piece of it. 

I've also been developing/testing multi-level constraints patches for matchAttribute, where a pceiRoot match is "Preferred", but if not available, it can fall back to a Required match such as numaNode, CPUsocketID, etc. This might address some of the concern around equality-based models

kannon92 (commented):
Please request a PRR review.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: @kad I implemented your numaNode as a list suggestion. The updated proposal makes `numaNode` an integer list (using KEP-5491 `DRAListTypeAttributes`).

The list is built from two filters applied to the ACPI SLIT distance matrix:

1. **Minimum distance**: only NUMA nodes at the minimum non-self SLIT distance from the device's physical node are included
2. **Same socket*: nodes on a different socket (different `physical_package_id`) are excluded, even if their SLIT distance matches

The physical NUMA node (from sysfs `numa_node`) is always the first element:

- **I/O devices**: `[physical, same-socket nodes at min SLIT distance...]`
- **CPU/memory devices**: `[N]`  single element, since they ARE a NUMA node

On current AMD EPYC Turin hardware under NPS4, all intra-socket SLIT distances are equal (distance 12), so the list includes all same-socket nodes: a NIC on NUMA 6 → `[6, 4, 5, 7]`. On future multi-IOD hardware where intra-socket distances vary (e.g., Granite Rapids with 2 IODs per socket), the minimum distance filter would naturally narrow the list to same-IOD nodes only.

`matchAttribute: numaNode` uses KEP-5491 intersection matching: CPU `[4]` ∩ NIC `[6, 4, 5, 7]` = `{4}` ≠ ∅ → match.

I tested this on a Dell R7725 (2x EPYC 9825, NPS4, 8 NUMA nodes) with 4 DRA drivers (CPU, memory, NVMe, SR-IOV) all publishing `numaNode` as a list. 16 pods with `matchAttribute: numaNode` constraints allocated correctly, with CPU capacity spreading across NUMA nodes as each filled up.

During testing I also found that the `DRAListTypeAttributes` feature gate was not wired to the scheduler's allocator (the experimental allocator with intersection matching was never selected). Fix submitted as kubernetes/kubernetes#139332.

I'll update the KEP to reflect the list type approach.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: @swatisehgal My updated proposal makes numaNode an integer list (via KEP-5491 DRAListTypeAttributes), where matchAttribute uses non-empty set intersection rather than equality. This enables multi-dimensional co-placement that a scalar attribute can't express:

  - A CPU on NUMA 4 publishes [4]
  - A NIC on NUMA 6 (equidistant to NUMA 4-7 per ACPI SLIT) publishes [6, 4, 5, 7]
  - Intersection: {4} ≠ ∅ → co-placed, even though they're on different NUMA nodes

This can adapt to heterogeneous accelerator topologies and varying PCIe hierarchies: the SLIT distance matrix captures whatever the hardware exposes, and the intersection matching finds valid combinations without requiring exact topology knowledge in the claim.

  Re: the discussion in #5213 — I think numaNode as a list is complementary to richer topology expressiveness. It solves the immediate cross-driver NUMA co-placement gap with infrastructure that already exists (KEP-5491), while longer-term proposals for topology-aware scheduling can build on the same list attribute foundation.

johnahull:
@kannon92 PRR questionnaire and approval file added (keps/prod-readiness/sig-node/6072.yaml). Ready for PRR review.

keps/sig-node/6072-dra-standard-numanode/README.md
kad: Thanks, having as a list would be better solution from I/O complexes.
As for CPU: there are scenarios (SNC/NPS) when those might be still be valuable as a lists. E.g. with AMD memory controllers located on I/O die, it doesn't much matter from which CPU core tile access will be  towards memory region within socket. (`C` in acronym `NUMA` represents `CPU`)

Single NUMA number makes sense only if it describes distinct memory region (or mode of operation for controller). In most other cases, the list would be also applicable. In any case, list with single number is still a list :)

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,496 @@
...
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation — e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Standardize `resource.kubernetes.io/numaNode` as a well-known DRA device attribute alongside the existing `pcieRoot` and `pciBusID`. The value is an **integer list** (requires KEP-5491 `DRAListTypeAttributes` feature gate) where:
+
+- The first element is the device's physical NUMA node (from the kernel's `numa_node` sysfs entry)
+- Additional elements are same-socket NUMA nodes at the minimum ACPI SLIT distance
```
pohly: Would it be possible to say "if such same-socket NUMA nodes are not available or `DRAListTypeAttributes` is disabled, then a single value with the physical NUMA node can be used instead"?

It won't make a difference for `matchAttribute`, it will be able to match a single value from one driver with a list from another.

With CEL, users need to be more careful: if `DRAListTypeAttributes` is on, they can use the `includes` function. Otherwise they have to manually write their CEL so that it works with single value or a list. But this shouldn't come up often in practice.

With this tweak, this KEP becomes somewhat usable in a more limited way also without list attributes. Because it's documented already now as "can also be a list", it's not an API break if DRA drivers start doing that.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,496 @@
...
+
+This enables cross-driver NUMA co-placement via `matchAttribute` with KEP-5491 non-empty intersection matching:
+
+```yaml
+constraints:
+- matchAttribute: resource.kubernetes.io/numaNode
+  requests: [gpu, nic, cpu, mem]
+```
+
+CPU device `[4]` ∩ NIC `[6, 4, 5, 7]` = `{4}` ≠ ∅ → match.
```
pohly: In this example, the CPU DRA driver could publish `4` as a single value, not a list.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Great idea. This removes the hard gate on KEP-5491 and makes the attribute immediately useful for simple same-NUMA matching.

I'll update the KEP to say drivers MAY publish `numaNode` as a scalar int (physical NUMA node) when `DRAListTypeAttributes` is disabled, and SHOULD publish it as an integer list when enabled. The CPU/memory driver example would publish `4` (scalar) rather than `[4]` (single-element list) in that case.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Updated the KEP to reflect this. CPU/memory drivers MAY publish a multi-element list under NPS/SNC configurations where multiple NUMA nodes share symmetric memory controller access. What each driver publishes is up to the driver authors.

keps/sig-node/6072-dra-standard-numanode/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+kep-number: 6072
+authors:
+  - "@johnahull"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scheduling
+  - wg-device-management
+status: provisional
+creation-date: 2026-05-12
+reviewers:
```
kannon92, line 10: @haircommander 

do you know who we picked for reviews for this KEP?

keps/sig-node/6072-dra-standard-numanode/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - TBD
+approvers:
+  - TBD
```
kannon92:
```suggestion
  - @mrunalp 
```

keps/sig-node/6072-dra-standard-numanode/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+see-also:
+  - "/keps/sig-scheduling/5491-dra-list-types-for-attributes"
+stage: alpha
+latest-milestone: "v1.37"
+milestone:
+  alpha: "v1.37"
+  beta: "v1.38"
+  stable: "v1.39"
+feature-gates:
+  - name: DRAStandardNUMANode
```
kannon92: There is an inconsistency with the feature gate here and your KEP.

Are we just relying on feature gate for KEP-5491? Or are we introducing a new feature gate?

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+- Components for list form: kube-apiserver (accepts `IntValues` in ResourceSlice), kube-scheduler (intersection matching in experimental allocator)
+- The helper functions are library code used by drivers at build time.
+
+###### Does enabling the feature change any default behavior?
+
+No. The attribute is additive. Existing claims and drivers are unaffected.
+
+###### Can the feature be disabled once it has been enabled?
+
+Yes. Disabling `DRAListTypeAttributes` prevents new ResourceSlices with list attributes. Existing ResourceSlices retain their values but new ones must use scalar attributes. Claims with `matchAttribute: numaNode` fall back to equality matching.
```
kannon92, line 429: This concerns me.

This KEP is coupled with the lifecycle of KEP of DRAListTypeAttributes which may graduate at different times.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation — e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+## Summary
+
+Standardize `resource.kubernetes.io/numaNode` as a well-known DRA device attribute alongside the existing `pcieRoot` and `pciBusID`. The value can be:
```
kannon92, line 73: The KEP process overall seems really quite heavy to just standardize an attribute. cc @pohly @johnbelamaric @klueska 

Reading this KEP I am unclear what alpha, beta and stable mean for just adding a well-known DRA attribute.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+No new API types. The standard attribute name `resource.kubernetes.io/numaNode` is added to the documented set of well-known device attributes in the `deviceattribute` library, alongside `pcieRoot` and `pciBusID`. The attribute uses the `IntValues` (list) field from `DeviceAttribute`, which requires the `DRAListTypeAttributes` feature gate (KEP-5491).
+
+### Helper Implementation
+
+```go
+// In k8s.io/dynamic-resource-allocation/deviceattribute
+
+const StandardDeviceAttributeNUMANode = StandardDeviceAttributePrefix + "numaNode"
+
+// GetNUMANodeListByPCIBusID reads numa_node from sysfs, then applies
```
kannon92: Reading this psuedo code, I am curious if/how we plan to support Numa on windows.

@pohly @johnbelamaric @marosset 
do we include windows for any DRA driver or support?

kannon92: Not saying we have to solve this but this would def not work on windows.

keps/sig-node/6072-dra-standard-numanode/kep.yaml
kannon92: Reading this KEP over I think you want to leave this blank. You are not introducing a feature gate.

kannon92:
/assign
for PRR.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+This makes the attribute immediately useful for same-NUMA matching without requiring `DRAListTypeAttributes`, while enabling richer cross-quadrant co-placement when list attributes are available.
+
+This attribute measures **memory topology** — which memory controllers are closest to the device. It is orthogonal to `pcieRoot`, which measures **bus topology** — which PCIe switch tree the device sits behind.
+
+### List Construction Algorithm
+
+The `numaNode` list is built from two filters applied to the ACPI SLIT distance matrix:
+
+1. **Minimum distance** — only NUMA nodes at the minimum non-self SLIT distance from the device's physical node are included
+2. **Same socket** — nodes on a different socket (different `physical_package_id`) are excluded, even if their SLIT distance matches
```
ffromani, line 128: does the list encode an ordering or a priority? meaning, is the order of entries relevant, or should be?
I'm not sure ordering is supported in the list attributes (need to refresh the KEP); if not the semantic should be "any in this list" meaning all the entries carry equal weight, which would create issues already with "just" 4 NUMA nodes in some HW microtopologies

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+
+TBD for beta.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+TBD for beta.
+
+###### Are there any missing metrics that would be useful to have in this category?
+
+TBD for beta.
```
Champbreed: For the monitoring section, the questionnaire expects at least one candidate metric to be identified at Alpha, even if implementation is deferred to Beta. A blanket "TBD for Beta" does not satisfy this field. It would be useful to at least name a candidate here, such as tracking how often allocations succeed via numaNode intersection versus standard matching.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,525 @@
...
+
+### Upgrade / Downgrade Strategy
+
+**Upgrade:** Existing claims are unaffected. Drivers can start publishing `resource.kubernetes.io/numaNode` at any time — it's just a new attribute value.
+
+**Downgrade:** If `DRAListTypeAttributes` is disabled, the API server will reject ResourceSlices with `IntValues` fields. Drivers would need to fall back to publishing `numaNode` as a scalar int (losing the SLIT-based list). Claims using `matchAttribute: numaNode` would revert to equality matching.
+
+### Version Skew Strategy
+
+The attribute depends on KEP-5491 `DRAListTypeAttributes` for the list type. All components (API server, scheduler) must have this feature gate enabled. The scheduler must also have the allocator wiring fix (kubernetes/kubernetes#139332) to select the experimental allocator that implements intersection matching.
```
Champbreed, line 389: During a rolling upgrade where kube-apiserver already has DRAListTypeAttributes enabled but the kube-scheduler hasn't been updated yet to support intersection matching, what happens to ResourceClaims trying to use matchAttribute: numaNode with list attributes during that window?

Champbreed:
PRR shadow review

keps/sig-node/6072-dra-standard-numanode/README.md
pohly: > the questionnaire expects at least one candidate metric to be identified at Alpha

Thanks for caling this. However, is the existence of such a metric really a hard requirement for all features? Not all features have a meaningful metric.

In this case, IMHO an appropriate and acceptable answer is "none" because a) nothing in core Kubernetes is checking which match attributes are used (making the implementation a major and potentially performance-degrading effort) and b) who would really care?

We already have too many metrics that remain in alpha because no-one is using them and thus we lack confidence that they should become beta. Let's only add metrics if there are realistic use cases for them. I don't see that in this case.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: The first number in the list is the numa node reported by the kernel. The rest of the numbers are in numerical order, but the order doesn't really matter. they are all equidistant and are the minimum non-self SLIT distance

keps/sig-node/6072-dra-standard-numanode/kep.yaml
johnahull: I'll leave it blank

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,550 @@
...
+2. **Same socket** — nodes on a different socket (different `physical_package_id`) are excluded, even if their SLIT distance matches
+
+The physical NUMA node (from sysfs `numa_node`) is always the first element.
+
+**CPU and memory devices** typically publish their physical NUMA node as a scalar `N` or single-element list `[N]`. Under NPS/SNC configurations where multiple NUMA nodes share symmetric memory controller access (e.g., AMD EPYC with memory controllers on the I/O die), drivers MAY publish a multi-element list. What each driver publishes is up to the driver authors.
+
+**I/O devices** (GPUs, NICs, NVMe) publish `[physical, nodes at min SLIT distance on same socket...]`. On hardware where multiple NUMA nodes are equidistant to an I/O device (e.g., AMD EPYC chiplets under NPS4 where all same-socket NUMA nodes are at distance 12), the list includes all same-socket nodes. On future multi-IOD hardware with asymmetric intra-socket SLIT distances (e.g., Intel Granite Rapids with 2 IODs per socket), the minimum distance filter would naturally narrow the list to same-IOD nodes only.
```
kad: One minimal nit: instead of codenames, it is better to use product name "Xeon 6".

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,550 @@
...
+
+**Examples (AMD EPYC 9825, NPS4, 8 NUMA nodes):**
+
+| Device | Physical NUMA | SLIT distances | `numaNode` value |
+|--------|--------------|----------------|------------------|
+| CPU on NUMA 4 | 4 | (is NUMA node) | `4` (scalar) or `[4]` (list) |
```
kad, line 142: Depending on how complex we might want to have that example...:
- CPU core on EPYC might have practically equal distance to multople NUMA nodes on the same socket in NPS4 mode (as memory controllers are on I/O die).
- Cores might be associated with connected to same physical package "cpu-less" numa nodes, like PMEM, CXL, HBM memory regions. But probably that is overkill for this kep to be listed.

keps/sig-node/6072-dra-standard-numanode/README.md
Champbreed: Agreed @pohly "none" with reasoning is the right answer here.

kannon92:
@johnahull 
can you remove claude from your last commit? This is failing the reason your CLA checks are failing.

johnahull:
> kannon92

Resolved the CI issue with Claude co-author

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: I hadn't planned to implement this for windows. Let me know if that's something I should do at some point.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Done, thanks.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Done: updated the Summary example so the CPU driver publishes scalar `4` instead of `[4]`, and noted that `matchAttribute` compares a scalar from one driver against a list from another. Thanks!

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Done. I answered the metrics/SLI/SLO questions as "None" with reasoning rather than "TBD for beta", per the consensus with @pohly.

keps/sig-node/6072-dra-standard-numanode/kep.yaml
johnahull: Applied: `@mrunalp` is set as the approver in kep.yaml.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Added a "Decoupling from KEP-5491 graduation" subsection. The scalar form has no dependency on KEP-5491 and works at any stage; KEP-5491 graduating, stalling, or rolling back only affects availability of the richer list form. The two KEPs are layered, not bound to a shared graduation schedule.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Thanks, good points. The minimum-SLIT-distance + same-socket filter already covers the EPYC NPS4 case (cores/IO equidistant to multiple same-socket nodes => the list includes them all). I'm keeping the cpu-less CXL/PMEM/HBM memory regions out of the example to keep it focused, as you suggested.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: To follow up on the ordering question: the equal weight is on purpose, and I don't think it causes the problem you're worried about.

The list only includes nodes at the minimum SLIT distance on the same socket. Anything farther away gets filtered out, it doesn't get added with lower priority. So the list is just "the nodes that are equally close," not "any nearby node."

For the 4-NUMA-node case (NPS4, memory controllers on the I/O die), all 4 really are the same distance from the device, so matching any of them is correct. If they weren't equally close, the minimum-distance filter wouldn't have put them all in the list.

The one thing I'll call out: SLIT is a rough signal, so two nodes that look equidistant can still differ a little in practice. The list attribute type from KEP-5491 doesn't support ordering anyway, so real priority/ranking is out of scope here. If we want that later it can build on this same attribute.

I added a note to the KEP spelling this out. Does that cover it, or do you have a specific microtopology in mind where you'd still want to rank the nodes in the list?

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Good question. I dug into the allocator code to get this right.

The intersection matching lives in the "experimental" allocator, and that allocator only gets selected when `DRAListTypeAttributes` is enabled in the scheduler (it's the only allocator variant whose supported-feature set includes list-type attributes). A scheduler that doesn't have the gate on runs the older "stable" allocator, which has no handling for list attributes at all: it sees the list as an unknown value type and the `matchAttribute` constraint just fails.

So during that rolling-upgrade window (apiserver storing list attributes, scheduler not yet on the new allocator):

- The claim doesn't get allocated, the pod stays `Pending`.
- It never gets placed on the wrong NUMA node. The match fails closed; it doesn't fall back to some looser comparison.
- Once the scheduler is updated with the gate on, it switches to the experimental allocator, the pending claim is re-evaluated, and it schedules normally via intersection matching.

So it's fail-safe and self-healing, and the only effect during the skew is that these claims wait to be scheduled until the scheduler side catches up. No incorrect placement, nothing to clean up. (Drivers publishing only the scalar form are unaffected, since scalar matching works on all allocator variants.)

I've added this to the Version Skew section of the KEP.

johnahull:
**Scalar fallback detection: planning to reuse the resourceslice publisher's `DroppedFieldsError`**

Following up on the scalar-baseline direction from @pohly (publish a scalar `numaNode` when `DRAListTypeAttributes` is off, a list when it is on). The piece I wanted to settle was how a driver knows which form to publish, since it can't read the apiserver's gates directly and the list field is silently dropped (not rejected) when the gate is off.

Looking at the code, the resourceslice publisher (`k8s.io/dynamic-resource-allocation/resourceslice`) already has the mechanism for this. It compares what the driver asked to publish against the version the apiserver actually stored (received back through the informer), and when fields were dropped it hands the driver a `DroppedFieldsError` through the `ErrorHandler`, with `DisabledFeatures()` naming the responsible gate. Four features already use this path: `DRADeviceTaints`, `DRAPartitionableDevices`, `DRADeviceBindingConditions`, `DRAConsumableCapacity`.

So the plan would be:

1. Extend `DisabledFeatures()` to recognize a dropped list-valued attribute and return `DRAListTypeAttributes`.
2. In the driver's `ErrorHandler`, republish the `numaNode` attribute in scalar form when that feature comes back as dropped.
3. Add a scalar helper in `deviceattribute` alongside the existing list helpers.

This reuses the established pattern instead of adding a separate probe or a config flag. Does that approach sound right, or is there a reason the other DRA features wouldn't want list-attribute detection folded into the same `DroppedFieldsError` path?

cc @pohly @klueska @johnbelamaric

keps/sig-node/6072-dra-standard-numanode/README.md
kannon92: We should probably call it out as a non goal then.

johnahull:
Correction to my comment above. I looked at the apiserver behavior more carefully, and the `DroppedFieldsError` detection I described does not work for this attribute.

When `DRAListTypeAttributes` is off, the apiserver drops the list field, which leaves the `numaNode` attribute with no value, and the slice then fails validation (a device attribute must have exactly one value). So the publish is rejected with an `Invalid` error, not stored with the field dropped, and no `DroppedFieldsError` is produced. This matches the apiserver strategy test `drop-fields-list-type-attributes`, which expects a validation error when the gate is off.

For context, numaNode looks like the first in-tree producer of a list-valued attribute, so there is no existing driver pattern to follow here.

The plan is still to publish the list form when the gate is on and a scalar (the physical NUMA node) when it is off, so the attribute stays usable without the gate. The open question is how a driver decides which form to publish, since it cannot read the apiserver feature gates directly. Candidates:

- Dry-run probe at startup: the driver does a dry-run Create of a small ResourceSlice carrying a list attribute. Success means the gate is on, an `Invalid` error means it is off. Decided once, with no failed real writes.
- A driver config flag set by the operator.
- Publish the list and fall back to scalar on an `Invalid` error, though that costs a failed publish first and `Invalid` is less specific.

I'm leaning toward the dry-run probe. Is that a reasonable pattern here, or is there an established way for an out-of-tree driver to learn whether a gated field is enabled?

cc @pohly @klueska @johnbelamaric

johnbelamaric:
> I'm leaning toward the dry-run probe. Is that a reasonable pattern here, or is there an established way for an out-of-tree driver to learn whether a gated field is enabled?

Yes. In fact I think we do this somewhere? We talked about even creating a function in the driver framework that used this to do a feature gate discovery. It would create a fake ResourceSlice using all the features and read it back, to see what's enabled or disabled, and return that list. I don't think we ever implemented it though.

johnbelamaric:
Oh, I see you found that code already. It's not clear to me why the current mechanism doesn't suffice to discover the feature enablement? I think adding it there is a good thing, if it works for you.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,581 @@
...
+
+This KEP standardizes a device attribute and ships helper library code rather than a gated API or controller, so the stages track **adoption and validation maturity** rather than the enablement lifecycle of a feature gate:
+
+- **Alpha** — the attribute name, semantics, and helper functions exist and are tested, with at least one driver publishing the value. The name is documented as "can also be a list," so drivers adopting it early is forward-compatible and not an API break.
+- **Beta** — enough independent drivers have adopted it (target: three) to confirm the name and SLIT-based semantics work across real hardware, with feedback from topology-aware workloads.
+- **GA** — real-world cross-driver co-placement deployments demonstrate the standard is stable and sufficient.
+
+Because the scalar form carries no feature gate (see [Decoupling from KEP-5491 graduation](#decoupling-from-kep-5491-graduation)), these stages gate confidence in the *convention*, not the availability of a gated code path.
+
+#### Alpha
```
kannon92: My big concern on this feature is that without a feature gate there is no graduation process for this.

If the code merges, its essentially GA.

kannon92:
/hold

https://github.com/kubernetes/enhancements/pull/6073/changes#r3393051319

Honestly I think this is blocking for this feature.

If we have no feature gate, alpha, beta or stable mean nothing. This feature is going GA right when it merges.

I am not that concerned with standardizing an attribute but I am suspect on essentially skipping feature graduation process. Should we consider a feature gate with beta on (so at least there is fall back if there are issues) and then we consider graduation once there is prod usage?

cc @johnbelamaric @pohly @liggitt

pohly:
> > I'm leaning toward the dry-run probe. Is that a reasonable pattern here, or is there an established way for an out-of-tree driver to learn whether a gated field is enabled?
>
> Yes. In fact I think we do this somewhere? We talked about even creating a function in the driver framework that used this to do a feature gate discovery. It would create a fake ResourceSlice using all the features and read it back, to see what's enabled or disabled, and return that list. I don't think we ever implemented it though.

No, we don't do this. I proposed it and then you convinced me that such a dynamic feature discovery is not common and too fragile (feature could also be flipped on again later, so the driver would have to constantly check and re-check). Instead, admins must ensure that components only use features which are enabled in the cluster. If not, components should crash-loop to make the admin aware of a configuration mistake.

For DRA drivers that means that a cluster must first be updated *completely* (apiserver and control plane) to support list attributes, then a DRA driver can be configured or updated to use it. Downgrading must first revert that change in the DRA driver, then downgrade the cluster.

This also covers the intermittent state correctly where apiserver and the other components have different support for the feature. Checking only the apiserver would incorrectly report "enabled" during a rolling update when kube-scheduler is from a release where the feature was alpha and off by default and the apiserver is newer and has it on by default.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Done

johnahull:
Thanks both. Here is the approach I would like to propose, and the design question I want your read on.

**Detection (@pohly):** I would drop the dry-run probe. The driver would not discover the cluster's feature state at runtime. Instead the numaNode helper would take a `listEnabled` input that the operator sets on the driver to match what the cluster has enabled: the list form when on, the scalar physical node when off. If a driver is configured to publish the list while `DRAListTypeAttributes` is off in the cluster, the apiserver rejects the publish and the driver should treat that as fatal rather than adapting. That follows the "components only use features the admin enabled, and crash-loop otherwise" model. Does that match what you had in mind?

**Graduation (@kannon92):** I agree the feature needs a graduation and rollback lever, so I want to settle where a gate would live. The thing that makes this unusual: numaNode has no code path of its own in the apiserver or scheduler. They treat it like any other device attribute, and the list type plus the intersection matching it enables are already gated by `DRAListTypeAttributes` (KEP-5491). The only numaNode-specific behavior is the driver deciding to publish the standardized attribute.

Because that one gated capability (the list form) already rides `DRAListTypeAttributes`, a `DRAStandardNUMANode` gate would exist for graduation and rollback governance rather than to gate any new behavior. So the first question is whether that lever is worth adding, or whether riding 5491 is sufficient.

If we do add a gate, it has two possible shapes:

1. It gates the driver's emission of the standardized attribute, and the list-vs-scalar choice rides on `DRAListTypeAttributes`. Natural fit since that is where the behavior is, but it is a driver-side gate rather than an apiserver or scheduler one.
2. We add in-tree behavior to gate: the apiserver recognizes `resource.kubernetes.io/numaNode` as a reserved well-known attribute and validates it (must be an int or int list) under the gate. A conventional in-tree gate with normal graduation, but it adds validation semantics the apiserver does not have for any well-known attribute today.

I lean toward option 1 if we add a gate. Before I write anything into the KEP and kep.yaml I would like your read: is a dedicated gate warranted given the capability already rides 5491, and if so which shape do you prefer? That should also tell us what is needed to lift the hold.

cc @pohly @kannon92 @johnbelamaric @liggitt

pohly:
> Detection (@pohly): [...] Does that match what you had in mind?

Yes.

> Graduation (@kannon92): I agree the feature needs a graduation and rollback lever,  [... ] DRAStandardNUMANode gate would exist for graduation [...]

I prefer option 1: DRA drivers decide what they publish, Kubernetes remains unaware. But that then means that we *don't* define a `DRAStandardNUMANode` feature gate because nothing uses it. DRA drivers don't use the in-tree feature gates, or maybe not even have feature gates at all. It's entirely up to the DRA driver author which config options or CLI parameter they provide.

kannon92:
> I prefer option 1: DRA drivers decide what they publish, Kubernetes remains unaware. But that then means that we don't define a DRAStandardNUMANode feature gate because nothing uses it. DRA drivers don't use the in-tree feature gates, or maybe not even have feature gates at all. It's entirely up to the DRA driver author which config options or CLI parameter they provide.

So in this case, the feature would be as follows:

- Create library functions for NUMA for driver authors
- Document numaNode as a standardized attribute

Maybe there is no need for a KEP for this?

keps/sig-node/6072-dra-standard-numanode/kep.yaml
```diff
@@ -0,0 +1,25 @@
+title: "DRA: Standard numaNode Device Attribute"
+kep-number: 6072
+authors:
+  - "@johnahull"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scheduling
+  - wg-device-management
```
kannon92:
```suggestion
```

pohly:
> Maybe there is no need for a KEP for this?

It's not a "normal" one, but not the first where a KEP is used to document and discuss some change that impacts the Kubernetes ecosystem. I think a KEP is useful for numaNode.

johnahull:
Thanks @kannon92, @pohly. Sounds like we have converged. Confirming the final shape:

- No `DRAStandardNUMANode` feature gate. Drivers decide what they publish and Kubernetes stays unaware (option 1).
- The deliverable is the `deviceattribute` library helpers plus the documented standardized `numaNode` attribute and its matching semantics.
- The KEP stays as the place to document and discuss it, per @pohly.

The KEP already reflects this:

- `kep.yaml` has `feature-gates: []` and `disable-supported: false`.
- A driver chooses list vs scalar through its own configuration (`listEnabled`), with no runtime feature discovery. The "Choosing the list or scalar form" section describes this, and the implementation follows it.
- A new "Consumer Expectations" section covers how `matchAttribute` and direct readers handle both forms.
- Graduation tracks adoption (one driver, then several, then real deployments) rather than a gate lifecycle, which the Graduation Criteria already describe.

Given there is no feature gate to graduate and we have agreement on the shape, would you be okay lifting the hold?

cc @pohly @mrunalp

keps/sig-node/6072-dra-standard-numanode/kep.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - sig-scheduling
+status: implementable
+creation-date: 2026-05-12
+reviewers:
+  - "@ffromani"
+approvers:
+  - "@mrunalp"
+see-also:
+  - "/keps/sig-scheduling/5491-dra-list-types-for-attributes"
+stage: alpha
```
kannon92: So what is this KEP stage?

kannon92:
> Graduation tracks adoption (one driver, then several, then real deployments) rather than a gate lifecycle, which the Graduation Criteria already describe.

I guess this is still where I am confused. Should I treat this KEP as beta on (ready for prod use)? If so I want the entire PRR filled out.

johnahull:
> > Graduation tracks adoption (one driver, then several, then real deployments) rather than a gate lifecycle, which the Graduation Criteria already describe.
> 
> I guess this is still where I am confused. Should I treat this KEP as beta on (ready for prod use)? If so I want the entire PRR filled out.

I think we should call it alpha for 1.37, not beta. Even without its own gate, it's not on for everyone. The part that actually does something, the list form and the matchAttribute intersection, rides DRAListTypeAttributes, which is still alpha. So on a default cluster the list isn't accepted and no matching happens. 

The scalar form has no gate, so a driver could publish numaNode as a plain int today. I think the alpha is the standardization and the helper library, and the part with real behavior rides 5491's alpha gate.

So in my opinion I'd keep it alpha and treat the PRR at the alpha level, which is what I filled out (enablement and rollback, dependencies, upgrade/downgrade, version skew, monitoring). I'd fill out the rest for beta in 1.38.

  If there are specific PRR sections you want done now to be comfortable lifting the hold, I'll do them. Just let me know. 

  cc @pohly @mrunalp

kannon92:
https://kubernetes.slack.com/archives/C0409NGC1TK/p1781274618415639

Outcome from slack:

- KEP is going directly to stable.
- PRR is not really relevant here.

Feel free to mark NA for areas that you don't think are relevant.

Without a feature gate, this feature doesn't have a graduation process so its not "alpha" but that is fine.

kannon92:
> I think we should call it alpha for 1.37, not beta. Even without its own gate, it's not on for everyone. The part that actually does something, the list form and the matchAttribute intersection, rides DRAListTypeAttributes, which is still alpha. So on a default cluster the list isn't accepted and no matching happens.

This feature is on for everyone. There is no guard against this once the code merges.

This feature is basically staying we are standarizing numaNode device attribute and driver authors can use that. The scheduling part depends on DRAListTypeAttributes but I don't see that really blocking this feature.

johnahull:
Thanks @kannon92 for the discussion and for lifting the hold. I've updated the KEP to match the outcome (direct to stable, no feature gate):

- `kep.yaml` is now `stage: stable` with `milestone: stable: v1.37`, and the PRR approval file moved to the stable stage.
- Graduation Criteria rewritten as direct-to-stable: with no gate there is no phased graduation; it merges as stable once the helpers, the documented attribute and semantics, and at least one publishing driver are in place.
- PRR questionnaire: dropped the beta placeholders, answered what applies, and marked the rest N/A (no control loop or gated behavior to roll out).
- Also fixed the enablement test wording to match the rest of the KEP: a list-only `numaNode` published while `DRAListTypeAttributes` is off is rejected, not silently dropped.

The implementation matches this: helper functions plus operator-configured list/scalar selection, no runtime probe, no feature gate.

I think it's ready for another look. @kannon92 for PRR and @mrunalp for sig-node approval, with the 1.37 enhancements freeze in mind.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,577 @@
...
+
+**Risk:** On future multi-IOD-per-socket hardware, all intra-socket SLIT distances might be equal even though IOD affinity differs.
+
+**Mitigation:** If SLIT distances don't differentiate IODs, the list includes all same-socket nodes — which is a safe over-approximation. If SLIT does differentiate, the minimum distance filter naturally narrows to same-IOD nodes. No code changes needed.
+
+## Design Details
+
+### API Changes
+
+No new API types. The standard attribute name `resource.kubernetes.io/numaNode` is added to the documented set of well-known device attributes in the `deviceattribute` library, alongside `pcieRoot` and `pciBusID`. The attribute uses the `IntValues` (list) field from `DeviceAttribute`, which requires the `DRAListTypeAttributes` feature gate (KEP-5491).
```
kannon92, line 256: It may be worth also adding these "well-known" attributes to https://kubernetes.io/docs/reference/labels-annotations-taints/.

WDYT @lmktfy?

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,577 @@
...
+
+#### Integration tests
+
+- DRA allocator integration test: CPU `[4]` and NIC `[4, 5, 6, 7]` are co-placed by `matchAttribute: resource.kubernetes.io/numaNode` (intersection `{4}` ≠ ∅)
+- DRA allocator integration test: CPU `[0]` and NIC `[4, 5, 6, 7]` fail `matchAttribute` (intersection ∅)
+- Consumable capacity test: multiple pods share the same CPU device with `matchAttribute` constraint
+
+#### e2e tests
+
+- End-to-end test with at least two DRA drivers publishing `resource.kubernetes.io/numaNode` as a list and a ResourceClaim with `matchAttribute` across them
```
kannon92: I don't really know how you can do this considering we don't e2e test any real DRA drivers?

I guess you could use in-tree driver?

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,577 @@
...
+
+No. The attribute is additive. Existing claims and drivers are unaffected.
+
+###### Can the feature be disabled once it has been enabled?
+
+Yes. Disabling `DRAListTypeAttributes` prevents new ResourceSlices with list attributes. Existing ResourceSlices retain their values but new ones must use scalar attributes. Claims with `matchAttribute: numaNode` fall back to equality matching.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+Drivers will resume publishing `resource.kubernetes.io/numaNode` as a list attribute in their ResourceSlices. The scheduler will resume using intersection matching for `matchAttribute` constraints referencing it. No state migration is needed — ResourceSlices are recreated on driver startup, and claims are re-evaluated by the scheduler. Previously allocated claims are unaffected since allocation results are stored in the claim status, not derived from ResourceSlice attributes at runtime.
```
kannon92: Should this be NA? 

We can't roll back this feature.

keps/sig-node/6072-dra-standard-numanode/README.md
```diff
@@ -0,0 +1,577 @@
...
+
+###### Are there any tests for feature enablement/disablement?
+
+Tests verify that a `numaNode` list attribute is accepted when `DRAListTypeAttributes` is enabled, and that publishing it when the gate is disabled is rejected by the apiserver (the dropped list value leaves a value-less attribute that fails validation), so a driver must publish the scalar form on such clusters.
```
kannon92: Not sure this question is still true.

> that publishing it when the gate is disabled is rejected by the apiserver

Is this referring to DRAListTypeAttributes?

kannon92 (commented):
This is mostly a opt in feature with a publishing a well known attribute and providing libaries for driver authors.

/approve

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Agreed, marked it N/A. This feature has no gate of its own to roll back; the list form availability just follows KEP-5491. (c3ccbbae)

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Yes, it refers to `DRAListTypeAttributes` (KEP-5491). Since this KEP has no gate of its own, I clarified the wording: there is no enablement/disablement to test for this KEP directly, the relevant gate is `DRAListTypeAttributes`, and the tests there verify the list attribute is accepted when it is on and rejected when off. (c3ccbbae)

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Good point. Reworded to use the in-tree test driver, since CI cannot run real vendor drivers. (c3ccbbae)

pohly: Let's make this N/A?

The in-tree driver doesn't support actual devices and thus any NUMA information would be entirely artificial. What this would test is "we can publish attributes", which is already well-covered.

keps/sig-node/6072-dra-standard-numanode/README.md
pohly: Yes. We already discussed doing that for pciRoot and someone wanted to do it, but I don't think it really happened. I believe we got stuck on having to do a major revamp of that page to fit the new category of information.

keps/sig-node/6072-dra-standard-numanode/README.md
johnahull: Makes sense, marked it N/A.

mrunalp (approved):

johnahull:
@ffromani @pohly @kannon92 @mrunalp — this is approved and tracked for v1.37 enhancements freeze (deadline is today). Just needs a `/lgtm` to merge. Let me know if there's anything outstanding.

haircommander:
sorry @johnahull I think you need to file an exception but it'll be a very straightforward one because the approvals were already there, we just missed the final piece of paperwork

kannon92:
> sorry @johnahull I think you need to file an exception but it'll be a very straightforward one because the approvals were already there, we just missed the final piece of paperwork

https://github.com/kubernetes/enhancements/issues/6072#issuecomment-4709375920

Release team marked this as tracking. Let's see if this merges. I don't think we need an exception for this.

PR #6074, opened 2026-05-12
KEP-4563: update EvictionRequest API

atiratree:
- One-line PR description: Apply API design changes discovered during the implementation and resolve pending  discussions/suggestions.

- Issue link: https://github.com/kubernetes/enhancements/issues/4563

- Other comments:


keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,304 +706,552 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
+    // eviction in one of the following formats:
+    // 1. acme.io/foo: "requester"
+    // 2. acme.io/foo: "responder"
+    // 3. acme.io/foo: "requesterresponder"
+    // +optional
+    metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
+
+    // spec defines the eviction specification.
+    // https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
     // +required
```
yongruilin, line 1016:
```suggestion
    // +required
    // +k8s:required
```

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,304 +706,552 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
+    Target EvictionTarget `json:"target" protobuf:"bytes,1,opt,name=target"`
+
+    // requesterName allows you to identify the entity, that requested the eviction of the target.
+    //
+    // It must be a valid domain-prefixed path (such as "acme.io/foo").
+    // Domain names *.k8s.io and *.kubernetes.io are reserved.
+    // This field is required and immutable.
+    // +required
+    // +k8s:required
+    RequesterName string `json:"requesterName" protobuf:"bytes,2,opt,name=requesterName"`
```
yongruilin, line 924: Does this fit into any existing string format validation that can be used with `+k8s:format=<format>`? If not, should we have one, as it is used in many fields?

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Some of the specs in kube API use required and some use optional. I also haven't seen any declarative validation, but it shouldn't hurt I guess?  

Any other thoughts on what the best practice should be here?

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Yes, it would be great to have `k8s-scoped-label-key` format. we have briefly discussed this in https://github.com/kubernetes/kubernetes/pull/137050

atiratree:
Hi, could you please review the EvictionRequest KEP update?

@haircommander @dchen1107 

@JoelSpeed @thockin for API review

@wojtek-t for PRR


The KEP update includes
- the API changes that were identified during [implementation](https://github.com/kubernetes/kubernetes/pull/137050). Most notably renaming intercerptor to a responder, status consolidation and responder state definitions.
- This KEP now features two APIs: **EvictionRequest** and **Eviction**. This was done to solve a couple of issues, mainly concerning the EvictionRequest lifecycle and eviction cancelation/withdrawal/restarts and finalizers. Please see https://github.com/atiratree/kube-enhancements/blob/eviction-request-api/keps/sig-node/4563-eviction-request-api/README.md#evictionrequest-and-eviction-api for more details. Previous conversations: [1](https://github.com/kubernetes/kubernetes/pull/137050#discussion_r2970400197) and [2](https://github.com/kubernetes/kubernetes/pull/137050#discussion_r2955476586)
- Removed complicated admission and there are no difficult naming issues now. Lookup is done via labels.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -265,68 +279,119 @@ For completeness here is a complete list of open PDB issues. Most are relevant t
 
 ## Proposal
 
-We will introduce a new declarative API called EvictionRequest, whose purpose is to coordinate the
-eviction of a pod from a node. It creates a contract between the eviction requester, the pod, and
-the interceptor (described later). The contract is enforced by the eviction request controller, the
-API type and validation.
+We will introduce new declarative APIs called EvictionRequest and Eviction, whose purpose is to
+coordinate the eviction of a pod from a node. They createa a contract between the eviction
```
haircommander: nit: create a*

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1777,9 +2207,9 @@ No.
...
 ###### How does this feature react if the API server and/or etcd is unavailable?
 
-- It is not possible to communicate eviction intent via the EvictionRequest API.
+- It is not possible to communicate eviction intent via the EvictionRequest and Eviction API.
 - EvictionRequest pods are not evicted/terminated.
-- Migration operations done by interceptors may still happen in the background, but the
+- Migration operations done by responders may still happen in the background, but the
   observability will be limited.
 
 ###### What are other known failure modes?
```
sohankunkerkar, line 2224: For beta, a few scenarios worth documenting: (1) responder crashes mid-eviction and its heartbeat expires, (2) evictionrequest-controller restarts with in-flight Evictions, (3) PDB blocks the imperative eviction responder indefinitely. Flagging now so they're on the radar for beta.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1581,26 +2007,26 @@ No.
...
 
 Cluster admin should take the following precautions, before disabling this feature.
-- Ensure that no eviction requests are in progress and no new eviction requests will occur during
+- Ensure that no evictions are in progress and no new eviction requests will occur during
   the disablement.
-- Ensure that all dependent components (eviction requesters, interceptors) in the cluster support
+- Ensure that all dependent components (eviction requesters, responders) in the cluster support
   disabling the feature.
 
 ###### What happens if we reenable the feature if it was previously rolled back?
```
sohankunkerkar, line 2030: If the feature was disabled longer than 20 minutes, responder heartbeats would have expired. When re-enabled, does the controller skip expired responders and move to the next one, or restart the responder chain? Worth clarifying for operators who disable temporarily during incidents.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -265,68 +279,119 @@ For completeness here is a complete list of open PDB issues. Most are relevant t
...
+Alternatively, the EvictionRequest and Eviction could be combined into a single object. However,
+there are major advantages to decoupling the eviction request and the eviction invocation into two
+separate types.
+
+The eviction request lifecycle differs from the eviction lifecycle. There is a many-to-many
+relationship between the requests and evictions. Multiple requesters can request the eviction of the
+same pod, and they can register and withdraw their eviction intent over time. 
+- If there were only a single EvictionRequest type and an instance for a single pod, multiple actors
+  would run into conflicts when requesting or canceling an eviction. 
+- Some actors could even block an eviction by creating an EvictionRequest object with a finalizer
```
sohankunkerkar: I think apart from single-type design, the two-type design still allows any actor with create access to add finalizers on their EvictionRequest, preventing GC of the associated Eviction object. Is there RBAC guidance or admission validation planned to prevent leaked Evictions from accumulating?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1720,28 +2144,34 @@ No.
...
   pod when it is created, which could result in performance issues.
-- There are N interceptors per one EvictionRequest. Only one interceptor is instructed to interact
-  with it at a time. This interceptor is instructed to update the
-  `.status.interceptors[].heartbeatTime` and the rest of the interceptor status
-  `.status.interceptors[]`. We validate the `heartbeatTime` and only allow periodic updates that are
-  60+ seconds to prevent frequent interceptor status updates.
+- There are N responders per one Eviction. Only one responder is instructed to interact
+  with it at a time. This responder is instructed to update the
+  `.status.responders[].heartbeatTime` and the rest of the responder status
+  `.status.responders[]`. We validate the `heartbeatTime` and only allow periodic updates that are
```
sohankunkerkar, line 2173: If a responder is rate-limited and its updates are rejected, its heartbeat stops advancing and eventually times out. Is there a metric or event to distinguish `responder is down` from `responder is being rate-limited`?  Without that, operators can't tell why a responder timed out.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1634,7 +2060,7 @@ the feature.
 ###### What specific metrics should inform a rollback?
 
 - If there is an unreasonably large number of evictions reported by a
-  `evictionrequest_controller_imperative_evictions` metric.
+  `imperative_eviction_responder_controller_failed_evictions` metric.
```
sohankunkerkar, line 2071: kep.yaml lists `imperative_eviction_responder_controller_imperative_evictions`. Are these the same metric or different? If different, the failed_evictions metric should also be in kep.yaml

sohankunkerkar (commented):
Reviewing this from the PRR reviewer-side (first pass):
1. Any actor with create permission on EvictionRequests can trigger pod eviction through the responder chain. Is there recommended RBAC guidance for who should have access?

2. For visibility: [KEP-5986](https://github.com/kubernetes/enhancements/pull/6141) is a potential consumer of the EvictionRequest API at beta. The kubelet would create EvictionRequests when it detects sustained memory pressure, which is a different urgency profile than node drain or descheduling. The Alpha2 considerations around `.spec.heartbeatDeadlineSeconds` and `.spec.type` would be useful for this use case.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -469,75 +537,65 @@ we can see many administrators override these default settings and many componen
...
+We will introduce `evictionrequest.coordination.k8s.io` and `eviction.coordination.k8s.io` types to
+enforce the contract between the eviction requesters and the responders. This type is a bit similar
+to `leases.coordination.k8s.io` in that it requires multiple actors to synchronize the state. Which
+in our case is the progress of the eviction.
 
 ### Eviction Requester
 [Eviction Requester](#eviction-requester) section provides a general overview.
 
-There can be many eviction requesters for a single EvictionRequest.
+There can be many eviction requesters, each of whome manages their own EvictionRequest object.
```
haircommander: whom*

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Yes, finalizers can still appear on EvictionRequests and Evictions. But it is not possible to disable eviction of a pod, because a new EvictionRequest+Eviction can always be created. There is no UID constraint for the name and the name can be generated.

At least in the beginning, we will only rely on default GC (ownerReferences) and requesters cleaning up their requests after they observe the final status.

If we see that the leakage of Evictions is a common issue, we can try to invest in a custom garbage collector (e.g. with a TTL after success).

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Most likely we would fail the eviction request with `NoFurtherResponder`. Although it could happen that the heartbeat could make it though before the KCM and the controller fully starts again.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: The heartbeat timeout is considerably longer (20m) than the rate limiting (1m) and the responder should have a reasonable heartbeat much lower than 20m (e.g. 3 minutes). I think it is reasonable to conclude that the responder is down after 19m of inactivity.

atiratree:
> Any actor with create permission on EvictionRequests can trigger pod eviction through the responder chain. Is there recommended RBAC guidance for who should have access?

sig-auth opinion is that the namespaced RBAC should be a sufficient boundary for creating EvictionRequests resources and that no additional authorization checks are needed at this time.

And the EvictionRequest RBAC access should be given to components managing the disruption (e.g. node drainer).

> For visibility: https://github.com/kubernetes/enhancements/pull/6141 is a potential consumer of the EvictionRequest API at beta. The kubelet would create EvictionRequests when it detects sustained memory pressure, which is a different urgency profile than node drain or descheduling. The Alpha2 considerations around .spec.heartbeatDeadlineSeconds and .spec.type would be useful for this use case.

Thanks, I will mention it in the KEP.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Thanks, I added the first 2. The third should be an expected behavior that is part of the design and not a failure mode.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: They are the same - updated.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: updated

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,364 +837,614 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
-set only the index 0 interceptor from the interceptor list in the beginning. After that, it is
-possible to set only the next interceptor at a higher index and so on. We can also condition this
-transition according to the other fields. `.status.interceptors[].completionTime` is set or
-`.status.interceptors[].heartbeatTime` has exceeded the 20-minute deadline.
+`spec.requesterName`: the identity of the requester should not change during the lifecycle of the
+EvictionRequest.
 
-`.status.processedInterceptors` should include only interceptors that have previously been in
-`.status.activeInterceptors`.
+### EvictionRequest Process
```
mortent: There are two level 3 headings with this name. We should find a different name for one of them.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1053,29 +1453,33 @@ The following diagrams describe what the EvictionRequest process will look like
 ![eviction-request-process](eviction-request-process.svg)
```
mortent, line 1457: Should this be updated? It references fields that doesn't match the rest of the doc.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1053,29 +1453,33 @@ The following diagrams describe what the EvictionRequest process will look like
...
 
 
-#### EvictionRequest Completion and Deletion
+#### EvictionRequest and Eviction Completion and Deletion
```
mortent, line 1460: Maybe rephrase this heading. I assume if covers Completion and Deletion for both the EvictionRequest and Eviction APIs, but the language is ambigious.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1053,29 +1453,33 @@ The following diagrams describe what the EvictionRequest process will look like
...
-The eviction request is considered canceled if:
+The Eviction is considered canceled if:
+- All requesters' intents are set to `Withdrawn`.
+
+The EvictionRequest is considered canceled if:
 - Validation fails after eviction request is created, for example the target pod is not found. This
   additional validation is enforced by the eviction request controller. 
-- The length of the `.spec.requesters` is 0.
-
+- All requesters' intents are set to `Withdrawn`. This has been taken over from the Eviction
```
mortent: There is only a single requester for each EvictionRequest right?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1053,29 +1453,33 @@ The following diagrams describe what the EvictionRequest process will look like
...
+  cancellation reasons above.
 
 Eviction controller will signal each of these by setting  `Canceled` or `Evicted` conditions to
-`True`.
+`True` in the EvictionRequest and Eviction objects. See
+[Condition synchronization](#condition-synchronization) for more details.
 
 Requesters should be able to delete the EvictionRequest when the `Canceled` or `Evicted` condition
```
mortent: This statement doesn't quite make it clear whether there will be validation on this, or whether we just expect clients to adhere to this.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -265,68 +277,124 @@ For completeness here is a complete list of open PDB issues. Most are relevant t
...
+the object. Alternatively, they can also wait until the eviction completion, read the results,
+and garbage collect the request when desired. There are no naming restrictions and the name can be
+generated as required.
+
+The lifecycle management of the Eviction invocation is solely handled by the eviction request
+controller. This controller will translate EvictionRequests into an Eviction.
+- No validation is needed on admission. The controller examines the last state of the world
+  (EvictionRequests) to determine if the Eviction should be present and if it should be active or
+  canceled. This can later be enhanced to offer various cancellation policies.
+- We have the ability to translate a single EvictionRequest into multiple Evictions for new targets
```
mortent, line 335: Do we have any more thoughts on how this would work? It seems like each Eviction will be handled separately, at least by the responders. But would there be a need to coordinate between the Evictions for different pods within the name PodGroup?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -449,16 +520,24 @@ purposes after it has reached the terminal phase (`Succeeded` or `Failed`).
...
 ### Risks and Mitigations
 
-If there is no interceptor, and the application has insufficient availability and a blocking PDB or
-blocking validating admission webhook, then the eviction request controller will enter into an
-API-initiated eviction cold loop with a backoff. To mitigate this we will increment the
-`evictionrequest_controller_imperative_evictions` metric.
+If there is no responder, and the application has insufficient availability and a blocking PDB or
+blocking validating admission webhook, then the imperative eviction responder controller will enter
+into an API-initiated eviction cold loop with a backoff. To mitigate this we will increment the
+`imperative_eviction_responder_controller_failed_evictions` metric.
```
mortent: Is it correct to call this a mitigation? Adding the metric helps with observability, but doesn't really mitigate the issue?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -616,68 +697,137 @@ be decided solely by the user deploying the application and resolved by creating
...
+reason (`EvictionInvalid`) and a message (`Target Pod xyz was not found.`).
 
 The new Workload API groups a set of pods that may not handle individual pod evictions well. It may
 be preferable to target the Workload resource. In the alpha version of the EvictionRequest API,
-pods referencing Workloads will be ignored. Therefore, we will reject any EvictionRequest that has a
-`.spec.workloadRef`. We will consider [Workload API Support](#workload-api-support) later.
-The rejection will be done again by setting a `Canceled` condition with an appropriate
-reason (`ValidationFailed`) and a message
+pods referencing PodGroups will be ignored. Therefore, we will reject any EvictionRequest that has a
+`.spec.schedulingGroup`. We will consider [Workload API Support](#workload-api-support) later.
```
mortent: Is this correct? I think we want to reject any EvictionRequest that references a pod that has a `spec.schedulingGroup`.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -469,75 +548,65 @@ we can see many administrators override these default settings and many componen
...
-interceptors can be used if all other concrete ones fail or are unavailable.
+responders can be used if all other concrete ones fail or are unavailable.
 
-First, the interceptor should register itself with all the pods it is interested in
-evicting/intercepting (either partially or fully) by adding itself to the
-`.spec.evictionInterceptors` field of the pod. This list is then added to the
-[EvictionRequest](#pod-and-evictionrequest-api) after creation by the eviction request controller.
+First, the responder should register itself with all the pods it is interested in
+evicting/processing (either partially or fully) by adding itself to the
+`.spec.evictionResponders` field of the pod. This list is then added to the
```
mortent, line 601: Who will manage the order of the responders? Since some responders might choose to terminate the pod the order would be important. It seems difficult for each responder to know where it should be in the list without understanding the semantics of the other responders.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Ack, one of them was an editing mistake, removed.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: My bad, the diagram should be up to date now.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Yes, but there can be multiple EvictionRequest at the same time for one Eviction.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: There is no validation, and have I added a more detailed explanation and the consequences of deletion.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: The implementation details are still not fully known at this time, but we will resolve/implement this at the later stage.

For pod targets, this will be guided by the PodGroup disruptionMode, which we will have to compare with the workload aware preemption. For PodGroups, we will disrupt all pods.

If any coordination is needed between Evictions, I presume it could be done by placing and prioritizing responders correctly. This would allow one responder to wait for another.

We could experiment with Eviction for a PodGroup target to allow for further coordination, but I am not sure if we have a use case for that.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Right, we don't have a good answer for the mitigation. The metric just helps with that. The explanation has been updated.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Nope, thanks for pointing that out.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: It is managed by the eviction responder controller (please see `Responder Selection`). I added more explanation under this section as well.

> It seems difficult for each responder to know where it should be in the list without understanding the semantics of the other responders.

This is indeed difficult, but we may be able to mitigate this with other features later on (e.g. a global priority).

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -469,75 +549,72 @@ we can see many administrators override these default settings and many componen
...
-between the eviction requesters and the interceptors. This type is a bit similar to
-`leases.coordination.k8s.io` in that it requires multiple actors to synchronize the state. Which in
-our case is the progress of the eviction.
+We will introduce `evictionrequest.lifecycle.k8s.io` and `eviction.lifecycle.k8s.io` types to
+enforce the contract between the eviction requesters and the responders. This type is a bit similar
+to `leases.coordination.k8s.io` in that it requires multiple actors to synchronize the state. Which
+in our case is the progress of the eviction. The coordination group is not being reused for these
+new types, because it should only contain coordination primitives for generic apiservers. And even
+though PDBs are in the policy group, evictions are ephemeral and do not adhere properly to being a
+policy. Therefore, we will introduce a new lifecycle group.
```
atiratree, line 562: Following @jpbetz [feedback](https://github.com/kubernetes/kubernetes/pull/137050#discussion_r3376480059), I changed the group name to `lifecycle`. 

There is also a [suggestion](https://github.com/kubernetes/kubernetes/pull/137050#discussion_r3376488051) to rename the EvictionRequest to EvictionClaim. I would like to hear additional opinions before doing the rename :)  cc @thockin

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -265,68 +276,124 @@ For completeness here is a complete list of open PDB issues. Most are relevant t
...
+### EvictionRequest and Eviction API
+
+`EvictionRequest` is used to request and store an eviction intent and to report the final state of
+the eviction (`Evicted`, `Failed`). 
+
+`Eviction` is used to coordinate the eviction process. Each eviction responder is given the
+opportunity to progress or complete the eviction. The entire eviction process and Eviction object
+lifecycle is managed by the eviction request controller. 
+
+Alternatively, the EvictionRequest and Eviction could be combined into a single object. However,
```
mortent: Maybe move this into the "Alternatives Considered" section. It disrupts the flow of the explanation of the API.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -265,68 +276,124 @@ For completeness here is a complete list of open PDB issues. Most are relevant t
...
+same time. Each requester should communicate their intent using their own instance of
+EvictionRequest. They can observe the conditions to check whether the eviction has been successful
+or not. They can also look up Eviction objects to gather more detailed status.
+
+For more details see [Eviction Requester](#eviction-requester-1) section in the Design Details.
+
+### Pod and Responder
+
+Any pod may be the subject to an eviction request type of eviction. There can be multiple responders
+for a single pod, and they should all advertise which pods they are responsible for. Normally, the
```
mortent, line 379: "they should all advertise which pods they are responsible for": The mechanism for this is just that responders add themselves to the pod `.spec.evictionResponders` list right?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,364 +845,612 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
-- Other interceptors should insert themselves into the `.spec.evictionInterceptors` according to
-  their own needs. Lower index interceptors are selected first by the eviction request controller. 
-- The number of the interceptors is limited to 15 for the Pod and for the EvictionRequest. If there
-  is a need for a larger number of interceptors, the current use case should be re-evaluated.
-  Limiting the number of interceptors ensures that the EvictionRequest cannot be blocked
-  indefinitely by setting an abnormally large number of these interceptors on a pod.
-- To prevent misuse, we will maintain a list of allowed `*.k8s.io` interceptor names. And reject
+- Other responders should insert themselves into the `.spec.evictionResponders` according to
+  their own needs. Lower index responders are selected first by the eviction request controller. 
+- The number of the responders is limited to 15 for the Pod and for the Eviction. If there
```
mortent: This comments is inconsistent with the proposed API. The number of evictionResponders in the Pod is limited to 16 and the number of targetResponders in the EvictionStatus is 17. Are these discrepancies intentional?

The max length here also limits the number of default responders that can be added. Is this the reason why the max length in the EvictionStatus is one higher than on the Pod? Should we expect there will ever be more than one default responder?

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1095,65 +1509,73 @@ metadata:
...
-   `.spec.requesters`.
-3. The descheduling controller notices that the pod p-1 is running in the wrong zone. It wants to
-   create an EvictionRequest (named after the pod's UID) for this pod, but the EvictionRequest
-   already exists. It sets the `descheduling.avalanche.io` value to the `.spec.requesters`.
-4. The eviction request controller designates Actor B as the next interceptor by updating
-   `.status.activeInterceptors[0]`.
-5. Actor B begins notifying users of application P that the application will experience
+   `.spec.requesterName` and `Eviction` to `.spec.intent`.
+3. The descheduling controller notices that the pod p-1 is running in the wrong zone. It creates a
+   new EvictionRequest and sets the`descheduling.avalanche.io` value to the
```
mortent: Nit: This will flow better by phrasing it as setting a field to a value. So this sentence would be "It creates a new EvictionRequest, setting `.spec.requesterName` to `descheduling.avalanche.io` and `.spec.intent` to `Eviction`". This applies to some other places too.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1053,29 +1459,37 @@ The following diagrams describe what the EvictionRequest process will look like
 ![eviction-request-process](eviction-request-process.svg)
 
 
-#### EvictionRequest Completion and Deletion
+#### EvictionRequest and Eviction Completion and Deletion
 
-The eviction request is considered evicted if:
+The EvictionRequest and Eviction is considered evicted if:
```
mortent: So I assume the EvictionRequest and the Eviction are not considered evicted, but rather completed? It is the target (i.e. the pod) that is evicted when the EvictionRequest and Eviction have completed successfully.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -1095,65 +1509,73 @@ metadata:
...
-7. The eviction request controller notices the change in `.spec.requesters`, and sets `Canceled`
-   condition to `True`  as there is no requester present. It also clears out
-   `.status.activeInterceptors` field.
-8. Actor B can detect the cancellation of the EvictionRequest object and notify users of application
+7. The node drain controller sets the intent to `Withdrawn` in `.spec.intent` of its own
+   EvictionRequest.
+8. The eviction request controller notices the change in `.spec.intent`, and sets `Canceled`
+   condition to `True` in both EvictionRequest and Eviction as there is no requester present. It 
+   also updates `.status.targetResponders[0].state` to `Canceled`.
+9. Actor B can detect the cancellation of the EvictionRequest object and notify users of application
```
mortent, line 1554: What happens if another EvictionRequest is created at this point? Does the Eviction transition back to active and then picks up where it left off with regard to the responders? Or will this result in a new Eviction being created and the eviction process will start from the beginning again?

mortent (commented):
I agree with the concern in https://github.com/kubernetes/kubernetes/pull/137050#pullrequestreview-4453577356 about the unbounded worst case time to eviction. I had a few comments, but otherwise this looks pretty good to me. Haven't had a chance to properly look at the PRR section yet.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Sure, updated. Btw, this alternative is already mentioned in https://github.com/atiratree/kube-enhancements/blob/eviction-request-api/keps/sig-node/4563-eviction-request-api/README.md#track-evictionrequests-and-eviction-in-a-single-object

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Yes. The `Proposal` section is more high level. The API is discussed in more detail in the `Design Details` section.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: I forgot to update this one. And yes, there can be up to 16 of user defined responders. And the status allow for +1 default responder. If the number of default responders grows over time, we can increase the validation.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Thanks for the suggestion. Updated.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Good point. I renamed the `Evicted` condition to `TargetEvicted` to make the subject clear.

For now, we are reserving the Completed condition, as there is a proposal to extend the Eviction lifecycle beyond the Pod lifecycle. That is, there could be a responder performing post-migration/validation tasks after the pod has been terminated. We would report Completed once all responders have completed.

keps/sig-node/4563-eviction-request-api/README.md
atiratree: If the Canceled condition and responder state are updated by the eviction controller, then yes, a new Eviction will object will be created and the process will start again.

If the controller is slower, then it will only result in updating .status.requesters.

atiratree:
> I agree with the concern in https://github.com/kubernetes/kubernetes/pull/137050#pullrequestreview-4453577356 about the unbounded worst case time to eviction.

@mortent @jpbetz 

One important issue we see in /eviction+PDB is the lack of pusback/coordination mechanism that would allow the workload to interact with the disruption (node drain) process. This results in applications defining blocking PDBs. The admin or disruptor doesn't understand the state or purpose/importance of each application in the cluster and therefore finds it difficult to proceed with the disruption during the cluster maintenance.

The EvicitonRequest and Eviction API solves this issue by providing the `.status.responders[].expectedCompletionTime` field, which allows us to asses how much time the workload requires. Together with the `.status.responders[].heartbeatTime` and `.status.responders[].message` fields, it offers us a better visibility into the workload.
 
The disruptor can observe the status of Eviction and increase the level of disruption if desired (/eviction or DELETE on the Pod). We plan to enhance our API to communicate the level of disruption, for example by adding a `.spec.type` field which could contain a deadline ([New EvictionRequest Types and Synchronization of Pod Termination Mechanisms](https://github.com/atiratree/kube-enhancements/blob/eviction-request-api/keps/sig-node/4563-eviction-request-api/README.md#new-evictionrequest-types-and-synchronization-of-pod-termination-mechanisms)). Adding the least disruptive type first helps us to define the safest default for critical workloads. It also helps us to improve the observability, as it is useful for the application to implement a responder instead of relying on a blocking PDB.


For critical workloads (live VM migration, checkpoint/restore, non-distributed DB), it is very hard to predict how much time should be allocated for viction by the admin or node drain/maintenance controller. This is because it can depend on workload size (process memory, local storage), the network bandwidth and the number of active sessions. Also, [Declarative Node Maintenance Motivation](https://github.com/atiratree/kube-enhancements/blob/improve-node-maintenance/keps/sig-apps/4212-declarative-node-maintenance/README.md#motivation) offers additional insight into disruption flows.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -616,68 +700,137 @@ be decided solely by the user deploying the application and resolved by creating
...
+
+If a single pod target is shared across multiple EvictionRequests, this should result in a single
+Eviction object.
 
-The controller should synchronize target's `.metatada.labels` to the EvictionRequest's
-`metadata.labels`. This will overwrite any conflicting labels in the EvictionRequest. This is done
-to allow for custom label selectors that the interceptors can use when watching the eviction
-requests.
+There are multiple options for naming the Evictions, but we must be careful as new targets could be
+added later.
```
haircommander, line 731: I'm left a little confused on what this proposal has chosen among these options

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,364 +840,612 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
 
-// Requester allows you to identify the entity, that requested the eviction of the target.
-// +structType=atomic
-type Requester struct {
-	// Name must be a fully qualified domain name of at most 253 characters in length, consisting
-	// only of lowercase alphanumeric characters, periods and hyphens (e.g. foo.example.com).
-	// This field must be unique for each requester.
-    // This field is required.
+    // metadata is the standard object metadata; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata.
+    // .metadaata.name set by the evictionrequest-controller is purely informative and subject to change.
```
haircommander: nit: metadata*

keps/sig-node/4563-eviction-request-api/README.md
mortent: Sounds good. The language just made me wonder if there were some other mechanism for this that I had missed in the rest of the proposal.

keps/sig-node/4563-eviction-request-api/README.md
mortent: I think increasing the limit is doable, but it will probably require ratcheting, so it will take multiple versions.

haircommander (commented):
some small notes, but generally LGTM. I think the new API direction is much more elegant

keps/sig-node/4563-eviction-request-api/README.md
atiratree: That should be okay. I don't expect new default responders to be added very often. Thanks!

keps/sig-node/4563-eviction-request-api/README.md
atiratree: The last one, I added a mention to the end of the section. Thanks!

intUnderflow:
Leaving my thoughts here, although I think you've all heard them in different forms over the last few months.

Overall I think the API is elegant, while not perfect for every use case that's not what we're aiming for, we're aiming for the minimal viable API to get user signal in alpha1, and I think that's where we are.

I have seen commentary around how some of the components of this version of the API are not fully user configurable (ex: deadlines) or are not mutable, I think that's actually a good thing, the biggest issue with this KEP throughout the process has been complexity, so there should be heightened scrutiny towards increasing the API footprint in this revision. I am hoping user feedback will allow us to be selective in where we would want to increase that footprint. Perfect is the enemy of good, especially at this early stage.

I think at this point the best signal is user feedback, and we can get that in alpha1 and refine from there. I know at least two companies who have actively reimplemented this KEP into their own CRDs to use it months ago, which demonstrates users are agitating for this feature and it will get signal.

wojtek-t:
I didn't have time to review carefully, but there is no objections about those changes from PRR perspective.

mortent:
With more configurability around the deadlines planned for future releases, I think this is good enough for alpha. LGTM

dchen1107:
@atiratree you need API reviewers to approve your API changes. But I like the current API approach though.

dchen1107:
Just want to log a review status: I am still in the process of digesting the new APIs, but really like the direction: decoupling of EvictionRequest and Eviction simplifies the overall architecture a lot and also make the design easier to understand, at least to me.

keps/sig-node/4563-eviction-request-api/README.md
```diff
@@ -687,364 +843,612 @@ Example message: `Could not evict a pod due to failing eviction requests, number
...
-	// +k8s:maxLength=15
-	EvictionInterceptors []Interceptor `json:"evictionInterceptors,omitempty"  patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,42,rep,name=evictionInterceptors"`
+    // Responders should periodically report on an eviction progress by updating the
+    // .status.responders[].heartbeatTime field of the Eviction object. If this field is
+    // not updated within 20 minutes, the eviction request is passed over to the next responder at
+    // a higher index. If there is no other responder, the last default
+    // imperative-eviction.k8s.io/evictor responder will evict the pod using the imperative
+    // Eviction API (/evict endpoint).
+    //
+    // The maximum length of the responders list is 16.
```
dchen1107, line 859: I raised this before, and we don't need to address this one for alpha, but want to put it here as a record: 

I still think the maximum length of the responders is too high:

1) Thinking about the worst scenario of a pod with 16 broken responders, 20mins timeout for each, it takes 16*20 minutes to finally hit the fall back one. 
2) Increasing PodSpec object side for 16 responders' names. 
3) Potentially we have 16 responders watching and interacting with a single Pod status, which creates unnecessary watch event churn for the API server.

dchen1107: My comment is NOT alpha blocker, but we should aggressively seek an alternative solution or pushing down the number of max responders for beta.

dchen1107:
/lgtm
/approve

Really like the new APIs which simplify the architecture. Also like we formally introduced topology-aware validation for AI workloads. Of course simply rejection eviction because pod is belonging a schedulingGroup isn't what we desire for, but that is a good start point. Now I feel I didn't lie / exaggerated when I called out EvictionRequest API is the one addressing AI workload aware maintainence. :-) Thanks!

dchen1107 (approved):

keps/sig-node/4563-eviction-request-api/README.md
atiratree: Linking my last response: https://github.com/kubernetes/enhancements/pull/4565#discussion_r2798340948.

I think we could accept lowering this to 10 and then potentially increase it back based on community feedback.

Alternatively, we could implement some unresponsive responder flood prevention. For example, with each failed responder, we could decrease the time by one minute? Or if a single responder fails, we decrease the time by a constant for the rest?

I will bring this topic to the WG for a discussion.

atiratree:
Thank you everyone!

keps/sig-node/4563-eviction-request-api/README.md
atiratree: We have decreased the number of responders to 10 for now and we will revisit this in alphaX/beta.

PR #6075, opened 2026-05-14
KEP-6075: Allow FQDN with trailing dot in HostAliases

kairosci:
- One-line PR description: Add feature gate HostAliasesAllowFQDN to allow trailing dot in hostAliases.hostnames for FQDN entries

- Issue link: https://github.com/kubernetes/kubernetes/issues/135273

- Other comments: This KEP modifies ValidateHostAliases to strip trailing dot before DNS-1123 subdomain validation, gated by HostAliasesAllowFQDN feature flag.

### DIFF
--- a/keps/sig-network/6075-hostaliases-allow-fqdn/README.md
+++ b/keps/sig-network/6075-hostaliases-allow-fqdn/README.md
@@ -1,0 +1,419 @@
+# KEP-6075: Allow FQDN with trailing dot in HostAliases
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Validation](#validation)
+  - [Storage: trailing dot is preserved](#storage-trailing-dot-is-preserved)
+    - [/etc/hosts Generation with Graceful Degradation](#etchosts-generation-with-graceful-degradation)
+    - [Discrepancy with <code>man 5 hosts</code>](#discrepancy-with-man-5-hosts)
+  - [Interaction with existing HostAliases semantics](#interaction-with-existing-hostaliases-semantics)
+  - [Feature gate](#feature-gate)
+  - [Test Plan](#test-plan)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+Allow a trailing dot in hostnames under `spec.hostAliases[].hostnames`. The trailing dot denotes a Fully Qualified Domain Name (FQDN) as defined in RFC 1034, preventing DNS resolvers from appending the search path.
+
+## Motivation
+
+DNS resolution in modern systems supports two styles of domain name queries: relative and absolute. A relative query (e.g., `example.com`) is subject to search path expansion, where the resolver appends configured search domains and retries if the initial lookup fails. An absolute query (denoted by a trailing dot, e.g., `example.com.`) bypasses this expansion entirely. This distinction is standardized in RFC 1034 and RFC 2181, and is universally implemented across operating systems, system resolvers, and language runtime libraries.
+
+Kubernetes `HostAliases` currently validates hostnames using `ValidateDNS1123Subdomain`, which rejects trailing dots. This restriction prevents users from pinning absolute FQDN names to specific IP addresses within their Pods. Applications that perform absolute FQDN lookups (common in distributed systems where exact name resolution is critical) cannot use `HostAliases` to override those names, even though the system resolver implementations inside containers fully support this use case.
+
+The practical consequence is that users who need FQDN overrides must either accept incorrect DNS behavior (by using relative names and hoping search domains resolve correctly), or use privileged init containers to manually rewrite `/etc/hosts` to include the trailing dots. Neither approach is satisfactory: the former breaks applications designed to use absolute names, and the latter requires elevated privileges and circumvents Kubernetes security controls.
+
+By allowing trailing dots in `HostAliases` hostnames, Kubernetes can support the full spectrum of DNS resolution patterns that applications require, making the feature more complete and usable for real-world scenarios.
+
+### Goals
+
+- Allow trailing dots in `HostAliases` hostnames to denote fully-qualified domain names (FQDNs), following RFC 1034 semantics.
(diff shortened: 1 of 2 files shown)


aojea:
We have some internet ancient history here, and I think we need to ensure we are providing correctness based on the actual expectations of the `/etc/hosts` file.

If we read the linux manual man 5 hosts (https://man7.org/linux/man-pages/man5/hosts.5.html):

> IP_address canonical_hostname [aliases...]
> 
> The IP address can conform to either IPv4 or IPv6. Fields of the entry are separated by any number of blanks and/or tab characters. Text from a "#" character until the end of the line is a comment, and is ignored. Host names may contain only alphanumeric characters, minus signs ("-"), and periods ("."). They must begin with an alphabetic character and end with an alphanumeric character.

[RFC 1123](https://www.rfc-editor.org/rfc/rfc1123.html) relaxed the "begin with an alphabetic character" rule from RFC 952 to allow digits:  

> The syntax of a legal Internet host name was specified in RFC-952 [DNS:4]. One aspect of host name syntax is hereby changed: the restriction on the first character is relaxed to allow either a letter or a digit. Host software MUST support this more liberal syntax.  

The trailing dot comes from DNS ([RFC 1034)](https://www.rfc-editor.org/rfc/rfc1034.html):

> The most common interpretation uses the root "." as either the single origin or as one of the members of the search list, so a multi-label relative name is often one where the trailing dot has been omitted to save typing.  

It is unclear to me if a trailing dot is actually well-specified for the /etc/hosts file itself. The man page explicitly says it must end with an alphanumeric character. A dot is not alphanumeric.  

It seems that putting a trailing dot in /etc/hosts is an undocumented behavior that may to work because glibc/musl parsers are mapping the RFC 1034 DNS query strictly against the file entry ... and we should not do validation based on the undocumented behaviors

kairosci:
I have updated the KEP with a new section to explicitly document this discrepancy and clarify why we need this behavior.

In short, while man 5 hosts expects alphanumeric endings, system resolvers inside containers (like glibc, musl, and CGO-enabled Go binaries) perform exact literal string matches against /etc/hosts. 

If a workload queries an absolute FQDN (e.g., example.com.), the local lookup will completely bypass the /etc/hosts entry unless the entry contains the trailing dot verbatim.

Since Kubernetes runs a wide variety of workloads beyond pure Go applications, preserving the trailing dot in /etc/hosts is a pragmatic necessity to make FQDN overrides function correctly across all runtimes. The KEP preserves RFC 1123 compliance by stripping the dot before API validation, but keeps it in /etc/hosts to satisfy the system resolvers.

aojea:
It is an undocumented behavior and the use case it solves is very very very rare , if we can agree whoever maintains /etc/hosts to document it is different ... but relying on anecdotal evidence of three parsers risk it may not work in language resolver foo ... so we break someone else

kairosci:
RFC 1034 (https://www.rfc-editor.org/rfc/rfc1034.html) and RFC 2181 (https://www.rfc-editor.org/rfc/rfc2181.html) explicitly document trailing dots as the standard notation for absolute FQDNs. This is not undocumented behavior. it's DNS standard.

Every major system resolver implements this consistently: glibc, musl, BSD, Windows. All major language runtimes support it: Go, Python, Java, Node.js, Rust, Ruby. The behavior is uniform across the ecosystem, not anecdotal.

Any workload making absolute FQDN queries requires this functionality. The use case is not rare. it's fundamental to correct DNS resolution semantics.

The KEP has been updated with RFC citations and comprehensive analysis of all major platform implementations. The trailing dot convention is normative DNS semantics that Kubernetes should support.

In addition, the KEP explicitly states that it intends to use "double entry", meaning that it will include both the domain name with a dot and without a dot, so that traditional (very rare) parsers can use the dot-less name.

aojea:
So the feature is you provide hostAliases and we build an /etc/hosts with those aliases

```
piVersion: v1
kind: Pod
metadata:
  name: hostaliases-pod
spec:
  restartPolicy: Never
  hostAliases:
  - ip: "127.0.0.1"
    hostnames:
    - "foo.local"
    - "bar.local"
  - ip: "10.1.2.3"
    hostnames:
    - "foo.remote"
    - "bar.remote"
 ```

IMHO the validation should match what `/etc/hosts` defines, not what the parsers of that file interpret

kairosci:
Your principle is correct but the standard you're applying is wrong.

You state validation should match what /etc/hosts defines, not what parsers interpret. Correct. The question is what actually defines /etc/hosts.

Man 5 hosts describes Unix implementation conventions. It is not the specification that defines what /etc/hosts must be for DNS resolution. RFC 1034 is. /etc/hosts is not a standalone format it is the system resolver's DNS datasource. Its entries must be valid for RFC 1034 DNS queries.

RFC 1034 normatively defines an absolute FQDN as requiring a trailing dot. When a workload queries "example.com." (absolute notation), the system resolver performs exact literal string matching on /etc/hosts. If the entry reads "example.com" (relative), there is no match. The query fails. This is not parser interpretation. This is DNS semantics defined in 1987.

The concrete case: any workload making absolute FQDN queries Go net.LookupHost, Python socket, Java InetAddress, any standard resolver will fail if hostAliases strips the trailing dot. This is not edge case behavior. This is how DNS works.

Should /etc/hosts validation follow man 5 hosts (Unix implementation convention) or RFC 1034 (the DNS standard that defines what /etc/hosts actually is for system resolution)?

Kubernetes orchestrates workloads that use RFC 1034. The validation must match RFC 1034, not man 5 hosts.

The double-entry pattern solves your concern: write both "example.com." and "example.com". Satisfies RFC 1034. No violations of man 5 hosts.

aojea:
Sorry, I do not think that argument justify the risk ... I'm also skeptical people use urls with trailing dots that is the only way of benefitting of this change ...

kairosci:
Specifically, what are you skeptical about? And more importantly, given the "double entry" definition, how do you think KEP could break Kubernetes?

aojea:
> Specifically, what are you skeptical about? 

First, what problem are we solving? Do we have more evidence that people writes urls using fdqn with trailing dots?

> how do you think KEP could break Kubernetes?

I know now we don't have any problem, I can't say exactly what can break but I've seen a lot of similar changes that in theory were harmless later break in one resolver (muslc, glibc) or in some specific language (golang, java, ...)  ...

kairosci:
Sorry for the late reply.

The real problem we’re solving isn’t about users typing URLs into browsers, but rather optimizing network performance in Kubernetes. By default, the cluster is configured with an `ndots` value of 5. This means that every time an application tries to contact an external endpoint like `api.example.com` without the trailing dot, the container resolver performs up to five internal DNS queries (appending search domains like `.default.svc.cluster.local`, `.svc.cluster.local`, etc.). Because both A and AAAA records are queried, a single external request generates up to 10 useless queries before resolving. For a microservice making just 1,000 API calls per second, this translates to an [amplification of 10,000 queries per second hitting CoreDNS](https://github.com/kubernetes/kubernetes/issues/14051). This well-documented amplification often leads to high CoreDNS CPU utilization, latency spikes, and [UDP packet loss due to Linux conntrack race conditions](https://www.weave.works/blog/racy-conntrack-and-dns-lookup-timeouts). To avoid this waste, the established best practice in distributed systems (publicly documented by engineering teams at [Datadog](https://www.datadoghq.com/blog/coredns-metrics/), Zendesk, and others) is to configure applications to query the absolute FQDN directly, including the trailing dot. If a team needs to override that endpoint in a test environment using `hostAliases`, they currently cannot do so: the container’s resolver will look for a literal match with the dot in `/etc/hosts`, and the resolution will fail.

The concern that a strict or legacy parser might fail is mitigated by the very structure of the double entry on the IP line. Looking at the actual source code of the most ubiquitous C libraries, [`glibc` (specifically `nss_files`)](https://sourceware.org/git/?p=glibc.git;a=blob;f=nss/nss_files/files-hosts.c) and [`musl libc`](https://git.musl-libc.org/cgit/musl/tree/src/network/) tokenize `/etc/hosts` lines using simple whitespace delimiters (`isspace()`). They do not enforce strict POSIX alphanumeric regex validation during the parsing phase that would trigger a panic, buffer overflow, or line rejection. By writing the IP followed first by the domain with a period and then by the domain without a period, modern parsers associate both records and correctly resolve absolute queries. In contrast, older or strictly manual-compliant parsers, scanning the line from left to right, will simply invalidate the token with the period and move on to the next valid token (the dot-less alias). The legacy system ignores the absolute fallback but does not break, maintaining exactly the same behavior as today.

To eliminate any further concerns regarding unexpected impacts on specific runtimes, this feature will not be rolled out globally and immediately. The feature will initially be released protected by an [Alpha-stage Feature Gate](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/), which is disabled by default. This will allow the community to test the behavior of containers in real-world scenarios with a wide variety of operating systems and programming languages, ensuring empirical validation (and verifying metrics like CoreDNS query drop rates) before moving the feature to a stable release.

kairosci:
@aojea as soon as you can, could you take another look at it? Thanks!

PR #6076, opened 2026-05-14
KEP-4671: Update Workload API and Gang Scheduling KEP for beta in v1.37

macsko:
- One-line PR description: Update KEP-4671 for beta in v1.37

- Issue link: https://github.com/kubernetes/enhancements/issues/4671

- Other comments:
  - Apart from beta graduation itself, this KEP proposes:
    - minCount mutability in PodGroup and PodGroupTemplate (based on the [document](https://docs.google.com/document/d/1tflhZdsgxCXfVgb2OrNFuodBStVXd3fdLz-SlC2zxHU)).
    - Adding new PlacementFeasible extension point to kube-scheduler to check PodGroup's feasibility (related to https://github.com/kubernetes/kubernetes/pull/138643).
    - Replacing delayed preemption with workload-aware preemption in beta criteria, eliminating the delayed preemption effort completely.
    - Merging GenericWorkload, GangScheduling and WorkloadAwarePreemption feature gates into one - GenericWorkload.
  - PR reuses some of the PRR answers written by @mm4tt in https://github.com/kubernetes/enhancements/pull/5854

### DIFF
--- a/keps/sig-scheduling/4671-gang-scheduling/README.md
+++ b/keps/sig-scheduling/4671-gang-scheduling/README.md
@@ -34,6 +34,7 @@
   - [PodGroup Creation Ordering](#podgroup-creation-ordering)
   - [Ownership and Object Relationship](#ownership-and-object-relationship)
   - [Pod Group minCount Mutability](#pod-group-mincount-mutability)
+  - [Workload Controllers Integration](#workload-controllers-integration)
   - [Scheduler Changes](#scheduler-changes)
     - [North Star Vision](#north-star-vision)
     - [GangScheduling Plugin](#gangscheduling-plugin)
@@ -126,11 +127,10 @@ their resources.  This allows other workloads to try to allocate those resources
 New core types called `Workload` and `PodGroup` are introduced to tell the kube-scheduler that a 
 group of pods should be scheduled together and to define policy options related to gang scheduling. 
 Pods may have an object reference in their spec to the `PodGroup` they belong to. The `Workload` 
-and `PodGroup` objects are intended to evolve[^2] via  future KEPs to support additional 
+and `PodGroup` objects are intended to evolve[^2] via future KEPs to support additional 
 kube-scheduler improvements, such as topology-aware scheduling.
 
-In 1.36, we redesigned the API in order to clearly decouple the Workload API from the runtime PodGroup API [^4].
-In the updated design:
+The API is structured to decouple template from runtime grouping:
 
 - `Workload` represents a static template defining the scheduling hierarchy and scheduling policy 
 definition that specifies what workload behavior should be applied.
@@ -226,34 +226,15 @@ See [Future plans](#future-plans) for more details.
 
 ## Proposal
 
-This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io/v1beta1`. The 1.36 revision, 
-as detailed in the original design [^4], addresses feedback regarding the ambiguity of the original "monolithic" 
-Workload. By decoupling policy (Workload) from runtime grouping (PodGroup), we improve clarity, scalability 
-and lifecycle management. This approach also provides a more readable and intuitive structure for complex workloads 
-like JobSet and LeaderWorkerSet.
-
+This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io`.
 The `Workload` API defines the scheduling policy and references one or more `podGroupTemplates`. Each `PodGroup` 
 is a standalone runtime object created from those templates, representing a self-contained scheduling unit that 
 encapsulates the runtime state.
 
-To maintain long-term API quality and ensure a clean path to GA, the Workload API remains in Alpha for the 1.36 
-cycle. We have decided to completely abandon the original v1alpha1 version of the API in favor of v1alpha2, where 
-only the new decoupled model is supported. This allows for the finalization of the architectural changes and ensures 
-that the graduated Beta API will be clean and free of transitional technical debt or legacy semantics.
-
 In v1.37, the API will be promoted to `v1beta1`. At the same time, a new `v1alpha3` version will be created
 to replace `v1alpha2`, enabling backward-incompatible changes around `DisruptionMode` for planned alpha features.
 
-The `spec.workloadRef` field was introduced in v1.35 to identify the scheduling context. In the v1.36 revision, we are
-replacing this with a new field, `spec.schedulingGroup`, to align with the decoupled architecture:
-
-- Immediate Execution Context: The Pod now points strictly to the immediate execution context (the runtime PodGroup). 
-  We are removing the direct Workload reference because it is not strictly necessary for the scheduler's operation. 
-  We may re-introduce a direct workload reference in the future (most probably in status, not spec) if concrete use 
-  cases (e.g., enhanced debuggability, UX) emerge. 
-- Future Extensibility: We anticipate the API will evolve to include concepts like PodSubGroup[^5]. Therefore, we 
-  structure this reference as a PodSchedulingGroup object. This allows us to easily extend the API in the future
-  (e.g., via a oneOf pattern) to support hierarchical scheduling.
+The `spec.schedulingGroup` on the Pod object is used to identify the scheduling context, which is the runtime `PodGroup`.
 
 A sample pod with these new fields looks like this:
 
@@ -752,11 +733,13 @@ type PodGroupStatus struct {
     // Conditions represent the latest observations of the PodGroup's state.
     //
     // Known condition types:
-    // - "PodGroupScheduled": Indicates whether the scheduling requirement has been satisfied.
+    // - "PodGroupInitiallyScheduled": Indicates whether the scheduling requirement has been satisfied.
+    //   Once this condition transitions to True, it serves as a terminal state and will never revert to False,
+    //   even if pods are subsequently evicted and group constraints are no longer met.
     // - "DisruptionTarget": Indicates whether the PodGroup is about to be terminated
     //   due to disruption such as preemption.
     //
-    // Known reasons for the PodGroupScheduled condition:
+    // Known reasons for the PodGroupInitiallyScheduled condition:
     // - "Unschedulable": The PodGroup cannot be scheduled due to resource constraints,
     //   affinity/anti-affinity rules, or insufficient capacity for the gang.
     // - "SchedulerError": The PodGroup cannot be scheduled due to some internal error
@@ -789,23 +772,6 @@ type WorkloadPodGroupTemplateReference struct {
     // +required
     PodGroupTemplateName string
 }
-
-// PodGroupStatus represents information about the status of a pod group.
-type PodGroupStatus struct {
-	// Conditions represent the latest observations of the PodGroup's state.
-	//
-	// Known condition types:
-	// - "PodGroupScheduled": Indicates whether the scheduling requirement has been satisfied.
-	//
-	// Known reasons for the PodGroupScheduled condition:
-	// - "Unschedulable": The PodGroup cannot be scheduled due to resource constraints,
-	//   affinity/anti-affinity rules, or insufficient capacity for the gang.
-	// - "SchedulerError": The PodGroup cannot be scheduled due to some internal error
-	//   that happened during scheduling, for example due to nodeAffinity parsing errors.
-	//
-	// +optional
-	Conditions []metav1.Condition
-}
 ```
 
 Individual `PodGroup` objects are treated as independent scheduling units. If a `Workload` defines 
@@ -821,32 +787,38 @@ in the code to allow for extending this structure if needed in the future.
 
 ### PodGroup Status Lifecycle
 
-The `PodGroup.Status` is managed by kube-scheduler to reflect the scheduling status. For alpha, we introduce 
-a `Conditions` field with the `PodGroupScheduled` condition type, but more fields may be added for beta and GA.
+The `PodGroup.Status` is managed by kube-scheduler to reflect the scheduling status. We introduce
+a `Conditions` field with the `PodGroupInitiallyScheduled` condition type.
 
 `PodGroup` status mirrors `Pod` status semantics rather than defining PodGroup-specific reasons:
 - If pods are unschedulable (i.e., timeout, resources, affinity, etc.), the scheduler updates the 
-`PodGroupScheduled` condition to `False` and sets the reason fields accordingly.
-- If pods are scheduled, the scheduler updates the `PodGroupScheduled` condition to `True` after the group 
+`PodGroupInitiallyScheduled` condition to `False` and sets the reason fields accordingly.
+- If pods are scheduled, the scheduler updates the `PodGroupInitiallyScheduled` condition to `True` after the group 
 got accepted by the Permit phase.
 
 For basic scheduling policy, when the pod related to the `PodGroup` gets scheduled, the scheduler updates 
-the `PodGroupScheduled` condition to `True`.
+the `PodGroupInitiallyScheduled` condition to `True`.
 
 #### Status Transition Rules
 
-Once a `PodGroup` transitions to `PodGroupScheduled=True`, it is treated as a terminal scheduling 
+Once a `PodGroup` transitions to `PodGroupInitiallyScheduled=True`, it is treated as a terminal scheduling 
 state and does not revert to `False`. Specifically, once the group's scheduling constraint (`minCount`) 
 has been satisfied, subsequent failed scheduling cycles for additional pods beyond `minCount` do not 
 regress the condition. On same-status transitions (e.g., `True` → `True`), the condition message may 
 be updated, but `LastTransitionTime` remains unchanged.
 
-This is a deliberate simplification. In practice, scheduled pods may later be evicted
-or impacted by node failures, but the `PodGroup` status will not track
-these post-scheduling disruptions. While a more robust status lifecycle capable of reflecting
-post-scheduling state changes is a desirable future enhancement, we do not consider it
-a blocker for beta graduation. Handling such behavior is tied to rescheduling 
-mechanisms, which remain out of scope for this KEP.
+Note that in practice, scheduled pods may later be evicted or impacted by node failures,
+but the `PodGroup` status will not track these post-scheduling disruptions.
+The current condition provides enough  visibility into the initial scheduling result
+to support beta and GA of this feature. However, future extensions to the status could:
+
+- Add a new condition tracking the outcome of the most recent scheduling cycle
+  (except in cases where the PodGroup is still feasible but additional pods weren't scheduled).
+  This would be a useful extension of the current condition and straightforward to implement.
+
+- Add a more robust status lifecycle mechanism capable of reflecting live post-scheduling state changes,
+  including current pod counts. It's likely that a new, separate component would be responsible
+  for tracking such changes and updating the status.
 
 #### Implementation Notes (Alpha)
 
@@ -971,6 +943,16 @@ and `PodGroupTemplate` is proposed to be mutable in v1.37. Specifically:
 While broader PodGroup spec mutability, such as modifying the number of `PodGroupTemplates` in a `Workload`,
 may be desirable, we are strictly scoping API mutability in v1.37 to the `minCount` field.
 Further relaxation of validation rules will be considered in the future if driven by strong use cases.
+
+### Workload Controllers Integration
+
+The Job controller integrated with this API in v1.36 (Alpha) via [KEP-5547], proving its initial capability. Currently,
+there are ongoing integrations with broader ecosystem workload controllers, such as JobSet, LeaderWorkerSet, KubeRay, and TrainJob.
+
+We are working closely with the communities behind these workloads, and the feedback is positive and supportive.
+There are no identified blockers for Beta graduation. Any feedback we have received can be delivered
+in a backward-compatible manner through subsequent KEPs, meaning the current API remains
+a solid baseline for these use cases.
 
 ### Scheduler Changes
 
@@ -1124,8 +1106,8 @@ The cycle proceeds as follows:
 #### Queuing and Ordering
 
 Workload-aware preemption ([KEP-5710](https://github.com/kubernetes/enhancements/pull/5711))
-will introduce a specific scheduling priority for a workload.
-Having that in mind, the queueing mechanism should support the workload's scheduling priority.
+will introduce a specific scheduling priority for a `PodGroup`.
+Having that in mind, the queueing mechanism should support the `PodGroup`'s scheduling priority.
 
 To ensure that we process the `PodGroup` instance at an appropriate time and
 don't starve other pods from being scheduled, we need to have a good queueing mechanism
@@ -1206,6 +1188,18 @@ The list and configuration of plugins used by this algorithm will be the same as
      the number of pods that have passed the Workload Scheduling Cycle to ensure
      that Pods do not wait unnecessarily if some have been rejected while new pods
      have been added to the cluster.
+
+     In this successful case, preemption will not be attempted even if some pods remain unschedulable
+     This means that as long as the scheduling constraints (`minCount`) are met, the schedulable pods
+     will move to binding. Any subsequent preemption for unschedulable pods will be handled in the next
+     [Workload Scheduling Cycle](#workload-scheduling-cycle) if still needed. Triggering both binding
+     and preemption in the same cycle would be ambiguous, and such precedence wouldn't be clear
+     from the scheduler's perspective. Alternatively, always attempting preemption to free up space,
+     even for schedulable groups, would be unnecessarily disruptive and delay the startup of the group. 
+     
+     Unschedulable pods will be requeued using their old timestamp, meaning the subsequent scheduling
+     and preemption attempt should start immediately after the current cycle,
+     unless a higher-priority pod or pod group comes in between.
 
    * If `schedulableCount < minCount`, the cycle fails. The scheduler attempts
      [Workload-aware Preemption](#workload-aware-preemption) to free sufficient space for the `PodGroup` through disruption.
@@ -1384,6 +1378,9 @@ became a beta graduation criterion for this KEP, the `WorkloadAwarePreemption` f
 will also be merged into `GenericWorkload`. This means the lifecycle of workload-aware preemption will be directly tied
 to the Workload API and gang scheduling features.
 
+However, this KEP and [KEP-5710] remain separate because they introduce different functionalities
+and operate on different abstraction layers.
+
 ### Test Plan
 
 [X] I/we understand the owners of the involved components may require updates to
@@ -1422,7 +1419,6 @@ With Workload Scheduling Cycle features, we will significantly expand test cover
 - Deadlocks and livelocks do not occur when multiple gangs compete for resources or interleave with standard pods.
 - Failed pod groups are requeued correctly and retry successfully when resources become available.
 - Scheduler correctly captures updated `minCount` value for pending pod groups, potentially unblocking them from PreEnqueue.
-- Changes to `minCount` value during Workload Scheduling Cycle doesn't affect the current cycle, but the subsequent ones.
 
 We will also benchmark the performance impact of these changes to measure:
 
@@ -1474,6 +1470,8 @@ In 1.36:
   and deliver a proper algorithm of handling PodGroup preemptions.
 - Implement PodGroup queueing algorithm.
 - Both `Workload` and `PodGroup` APIs are integrated (alpha) with at least one true workload[^6] controller.
+- There are no blockers or concerns from the true workload controller[^6] communities,
+  proving that the Workload and PodGroup APIs serve as a solid baseline for their use cases.
 - A deletion protection mechanism is implemented for `PodGroup` objects and finalizer is added to the API.
 - All e2e tests for `PodGroup` are added.
 - Performance tests are created and are being run in CI to protect against regressions.
@@ -1481,6 +1479,7 @@ In 1.36:
 #### GA
 
 - All issues and gaps identified as feedback during beta are resolved
+- Promote the e2e API tests to conformance together with a test for gang scheduling behavior.
 
 ### Upgrade / Downgrade Strategy
 
@@ -1535,12 +1534,9 @@ This section must be completed when targeting alpha to a release.
     - kube-apiserver
     - kube-scheduler
     - kube-controller-manager
-- [ ] Other
-  - Describe the mechanism:
-  - Will enabling / disabling the feature require downtime of the control
-    plane?
-  - Will enabling / disabling the feature require downtime or reprovisioning
-    of a node?
+
+This KEP and workload-aware preemption ([KEP-5710]) are tightly coupled and controlled by the same feature gate.
+They graduate in lockstep, meaning their graduation paths (Alpha, Beta, GA) and timelines are identical.
 
 ###### Does enabling the feature change any default behavior?
 
@@ -1607,19 +1603,19 @@ behavior).
 2. Attempt to create a Pod with `spec.schedulingGroup` set.
 3. The `spec.schedulingGroup` field is dropped by the API server. The pod is created successfully
    but without the `schedulingGroup` reference, resulting in immediate standard scheduling (one-by-one).
-4. Restart/Upgrade API Server and Scheduler to v1.37 (with feature gates enabled).
+4. Restart/Upgrade API Server and Scheduler to v1.37 with feature gates enabled.
 5. Create two PodGroup objects: `gang-test-A` and `gang-test-B` (both with `minCount=2`).
 6. Create a Pod `test-pod-1` with `spec.schedulingGroup` pointing to `gang-test-A`.
 7. The Pod stays in `Pending` state (waiting for the gang). Verify that
    `scheduler_pending_entities{type="podgroup", queue="gated"}` metric is incremented.
 8. Create a Pod `test-pod-2` pointing to the same pod group.
 9. Both pods are scheduled successfully in the same cycle (Gang Scheduling works). 
-10. Downgrade API Server and Scheduler to v1.36 (with feature gates disabled).
+10. Downgrade API Server and Scheduler to v1.36 with feature gates disabled.
 11. Create `test-pod-3` pointing to `gang-test-B`. Note: We use a pod group created in step 5 because creating new
     PodGroup objects is disabled.
 12. The pod is scheduled immediately (PodGroup logic is ignored because the schedulingGroup field is dropped by
     the v1.36 API server). If Gang Scheduling were active, this pod would hang pending waiting for a second member.
-13. Upgrade API Server and Scheduler back to v1.37 (feature gates enabled).
+13. Upgrade API Server and Scheduler back to v1.37 with feature gates enabled.
 14. Create `test-pod-4` and `test-pod-5` pointing to `gang-test-B`; verifying that Gang Scheduling functionality is
     restored (these pods wait for `minCount=2` before scheduling).
 
@@ -1656,7 +1652,7 @@ Recall that end users cannot usually observe component logs or access metrics.
 -->
 
 - [x] API .status
-  - Condition name: `PodGroupScheduled`
+  - Condition name: `PodGroupInitiallyScheduled`
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
@@ -2013,8 +2009,8 @@ Kubernetes uses for most resource names.
 
 [#5501]: https://github.com/kubernetes/enhancements/pull/5501
 [extended proposal]: https://docs.google.com/document/d/1ulO5eUnAsBWzqJdk_o5L-qdq5DIVwGcE7gWzCQ80SCM/edit?
-[KEP-5547]: https://github.com/kubernetes/enhancements/pull/5871
-[KEP-5710]: https://github.com/kubernetes/enhancements/pull/5710
+[KEP-5547]: https://github.com/kubernetes/enhancements/issues/5547
+[KEP-5710]: https://github.com/kubernetes/enhancements/issues/5710
 [kubernetes.io]: https://kubernetes.io/
 [kubernetes/enhancements]: https://git.k8s.io/enhancements
 [kubernetes/website]: https://git.k8s.io/website
--- a/keps/sig-scheduling/4671-gang-scheduling/README.md
+++ b/keps/sig-scheduling/4671-gang-scheduling/README.md
@@ -793,11 +793,11 @@ a `Conditions` field with the `PodGroupInitiallyScheduled` condition type.
 `PodGroup` status mirrors `Pod` status semantics rather than defining PodGroup-specific reasons:
 - If pods are unschedulable (i.e., timeout, resources, affinity, etc.), the scheduler updates the 
 `PodGroupInitiallyScheduled` condition to `False` and sets the reason fields accordingly.
-- If pods are scheduled, the scheduler updates the `PodGroupInitiallyScheduled` condition to `True` after the group 
-got accepted by the Permit phase.
+- If pods are scheduled, the scheduler updates the `PodGroupInitiallyScheduled` condition to `True` after
+the member pods moved to the binding phase.
 
-For basic scheduling policy, when the pod related to the `PodGroup` gets scheduled, the scheduler updates 
-the `PodGroupInitiallyScheduled` condition to `True`.
+For basic scheduling policy, when the pod related to the `PodGroup` gets scheduled (moves to binding),
+the scheduler updates the `PodGroupInitiallyScheduled` condition to `True`.
 
 #### Status Transition Rules
 
@@ -1480,6 +1480,9 @@ In 1.36:
 
 - All issues and gaps identified as feedback during beta are resolved
 - Promote the e2e API tests to conformance together with a test for gang scheduling behavior.
+
+This KEP and workload-aware preemption ([KEP-5710]) are tightly coupled and controlled by the same feature gate.
+They graduate in lockstep, meaning the graduation criteria of [KEP-5710] should be also met when targeting to a release.
 
 ### Upgrade / Downgrade Strategy
 
--- a/keps/sig-scheduling/4671-gang-scheduling/README.md
+++ b/keps/sig-scheduling/4671-gang-scheduling/README.md
@@ -27,12 +27,13 @@
   - [Associating Pod into PodGroups](#associating-pod-into-podgroups)
   - [API](#api)
   - [PodGroup Status Lifecycle](#podgroup-status-lifecycle)
-    - [Alpha Status Transition Rules](#alpha-status-transition-rules)
+    - [Status Transition Rules](#status-transition-rules)
     - [Implementation Notes (Alpha)](#implementation-notes-alpha)
   - [PodGroup Deletion Protection](#podgroup-deletion-protection)
   - [SchedulingPolicy Reference vs. Copy/Inline in PodGroup](#schedulingpolicy-reference-vs-copyinline-in-podgroup)
   - [PodGroup Creation Ordering](#podgroup-creation-ordering)
   - [Ownership and Object Relationship](#ownership-and-object-relationship)
+  - [Pod Group minCount Mutability](#pod-group-mincount-mutability)
   - [Scheduler Changes](#scheduler-changes)
     - [North Star Vision](#north-star-vision)
     - [GangScheduling Plugin](#gangscheduling-plugin)
@@ -43,9 +44,9 @@
     - [Scheduling Algorithm](#scheduling-algorithm)
     - [Algorithm Limitations](#algorithm-limitations)
     - [Interaction with Basic Policy](#interaction-with-basic-policy)
-    - [Delayed Preemption](#delayed-preemption)
     - [Workload-aware Preemption](#workload-aware-preemption)
     - [Failure Handling](#failure-handling)
+  - [Feature gates merge in v1.37](#feature-gates-merge-in-v137)
   - [Test Plan](#test-plan)
       - [Prerequisite testing updates](#prerequisite-testing-updates)
       - [Unit tests](#unit-tests)
@@ -225,7 +226,7 @@ See [Future plans](#future-plans) for more details.
 
 ## Proposal
 
-This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io/v1alpha2`. The 1.36 revision, 
+This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io/v1beta1`. The 1.36 revision, 
 as detailed in the original design [^4], addresses feedback regarding the ambiguity of the original "monolithic" 
 Workload. By decoupling policy (Workload) from runtime grouping (PodGroup), we improve clarity, scalability 
 and lifecycle management. This approach also provides a more readable and intuitive structure for complex workloads 
@@ -239,6 +240,9 @@ To maintain long-term API quality and ensure a clean path to GA, the Workload AP
 cycle. We have decided to completely abandon the original v1alpha1 version of the API in favor of v1alpha2, where 
 only the new decoupled model is supported. This allows for the finalization of the architectural changes and ensures 
 that the graduated Beta API will be clean and free of transitional technical debt or legacy semantics.
+
+In v1.37, the API will be promoted to `v1beta1`. At the same time, a new `v1alpha3` version will be created
+to replace `v1alpha2`, enabling backward-incompatible changes around `DisruptionMode` for planned alpha features.
 
 The `spec.workloadRef` field was introduced in v1.35 to identify the scheduling context. In the v1.36 revision, we are
 replacing this with a new field, `spec.schedulingGroup`, to align with the decoupled architecture:
@@ -258,7 +262,7 @@ apiVersion: v1
 kind: Pod
 spec:
   ...
-  # In 1.36 schedulngGroup replaces workloadRef.
+  # In 1.36 schedulingGroup replaces workloadRef.
   schedulingGroup:
     podGroupName: pg1  # Points to the standalone PodGroup object
   ...
@@ -301,7 +305,7 @@ serves as a policy template, containing the PodGroupTemplates with their corresp
 
 The `Workload` object defines these templates:
 ```yaml
-apiVersion: scheduling.k8s.io/v1alpha2
+apiVersion: scheduling.k8s.io/v1beta1
 kind: Workload
 metadata:
   namespace: ns-1
@@ -318,7 +322,7 @@ spec:
 
 A sample `PodGroup` instantiated from the above template would look like this:
 ```yaml
-apiVersion: scheduling.k8s.io/v1alpha2
+apiVersion: scheduling.k8s.io/v1beta1
 kind: PodGroup
 metadata:
   name: training-worker-0
@@ -512,7 +516,7 @@ The example below shows how this could look with the decoupled architecture for 
 A Workload object defines the static PodGroup template:
 
 ```yaml
-apiVersion: scheduling.k8s.io/v1alpha2
+apiVersion: scheduling.k8s.io/v1beta1
 kind: Workload
 metadata:
   name: jobset
@@ -527,7 +531,7 @@ spec:
 A standalone PodGroup object is created to define the scheduling policy and track a specific runtime instance:
 
 ```yaml
-apiVersion: scheduling.k8s.io/v1alpha2
+apiVersion: scheduling.k8s.io/v1beta1
 kind: PodGroup
 metadata:
   name: job-instance-worker-0
@@ -685,17 +689,25 @@ type BasicSchedulingPolicy struct {
 type GangSchedulingPolicy struct {
 	// MinCount is the minimum number of pods that must be schedulable or scheduled
 	// at the same time for the scheduler to admit the entire group.
-	// It must be a positive integer.
+	// It must be a positive integer. This field is mutable to support workload scaling.
+	//
+	// Note that the scheduler operates on an eventually consistent model. Updates
+	// to minCount may not be immediately reflected in scheduling decisions due to
+	// propagation delays. If minCount is updated while a scheduling cycle is in
+	// progress for that group, the new value may not take effect until the next
+	// cycle. Moreover, minCount is only enforced during scheduling, meaning that
+	// modifications to this field do not affect already-scheduled pods, applying
+	// only to those evaluated in future cycles.
 	//
 	// +required
 	MinCount int32
 }
 ```
 
-The `PodGroup` resource is a separate API object in `scheduling.k8s.io/v1alpha2`:
+The `PodGroup` resource is a separate API object in `scheduling.k8s.io/v1beta1`:
 
 ```go
-// API Group: scheduling.k8s.io/v1alpha2
+// API Group: scheduling.k8s.io/v1beta1
 
 // PodGroup represents a runtime instance of pods grouped together.
 // PodGroups are created by workload controllers (Job, LWS, JobSet, etc...) from
@@ -821,7 +833,7 @@ got accepted by the Permit phase.
 For basic scheduling policy, when the pod related to the `PodGroup` gets scheduled, the scheduler updates 
 the `PodGroupScheduled` condition to `True`.
 
-#### Alpha Status Transition Rules
+#### Status Transition Rules
 
 Once a `PodGroup` transitions to `PodGroupScheduled=True`, it is treated as a terminal scheduling 
 state and does not revert to `False`. Specifically, once the group's scheduling constraint (`minCount`) 
@@ -829,10 +841,12 @@ has been satisfied, subsequent failed scheduling cycles for additional pods beyo
 regress the condition. On same-status transitions (e.g., `True` → `True`), the condition message may 
 be updated, but `LastTransitionTime` remains unchanged.
 
-This is a deliberate simplification for alpha. In practice, pods may later become unschedulable 
-(for example, due to node failures or evictions), but the `PodGroup` status will not reflect that. 
-A more robust status lifecycle that can capture post-scheduling state changes must be designed for 
-beta graduation.
+This is a deliberate simplification. In practice, scheduled pods may later be evicted
+or impacted by node failures, but the `PodGroup` status will not track
+these post-scheduling disruptions. While a more robust status lifecycle capable of reflecting
+post-scheduling state changes is a desirable future enhancement, we do not consider it
+a blocker for beta graduation. Handling such behavior is tied to rescheduling 
+mechanisms, which remain out of scope for this KEP.
 
 #### Implementation Notes (Alpha)
 
@@ -938,6 +952,26 @@ Pods reference their `PodGroup` via `schedulingGroup.podGroupName`, which allows
 look up the `PodGroup` object. The scheduler requires `PodGroup` object to exist before scheduling 
 pods that reference them.
 
+### Pod Group minCount Mutability
+
+To support workload scaling (e.g. for elastic Jobs), the `minCount` field in both `PodGroup`
+and `PodGroupTemplate` is proposed to be mutable in v1.37. Specifically:
+
+* Modifying a `PodGroupTemplate` will not affect existing `PodGroup` instances.
+  Changes should apply exclusively to new instances created from the updated template by the controllers.
+
+* Because the scheduler operates on an eventually consistent view of the cluster state,
+  updates to `minCount` may not be immediately visible. If a modification occurs while the scheduler
+  is actively processing that specific `PodGroup` in a scheduling cycle, the previous value
+  may still be evaluated for that cycle. The new value will take effect in the subsequent cycle.
+
+* `minCount` is enforced strictly during scheduling, not at runtime. This means modifications to the field
+  will not affect already-scheduled pods, applying only to those evaluated in future scheduling cycles.
+
+While broader PodGroup spec mutability, such as modifying the number of `PodGroupTemplates` in a `Workload`,
+may be desirable, we are strictly scoping API mutability in v1.37 to the `minCount` field.
+Further relaxation of validation rules will be considered in the future if driven by strong use cases.
+
 ### Scheduler Changes
 
 The kube-scheduler will add a new informer to watch `PodGroup` objects. If the `PodGroup` is missing,
@@ -983,6 +1017,8 @@ We will implement a new plugin implementing the following hooks:
   scheduling. The extension will check if the `PodGroup` object exists. If not, it will return 
   `UnschedulableAndUnresolvable` status. Then it verifies that at least `minCount` pods have 
   been observed, ensuring there are enough pods to consider before enqueuing them.
+  If the `minCount` was updated, the kube-scheduler should re-evaluate the PreEnqueue with the new value.
+
 **- WaitOnPermit**: used as a barrier to wait for the pods to be assigned to the nodes before 
 initiating potential preemptions and their bindings. The extension waits for all pods in the 
 `PodGroup` to reach permit stage by using each pod's `schedulingGroup.podGroupName` to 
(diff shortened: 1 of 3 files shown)


macsko:
/hold
For approvers' review

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -685,17 +689,23 @@ type BasicSchedulingPolicy struct {
 type GangSchedulingPolicy struct {
 	// MinCount is the minimum number of pods that must be schedulable or scheduled
 	// at the same time for the scheduler to admit the entire group.
-	// It must be a positive integer.
+	// It must be a positive integer. This field is mutable to support workload scaling.
+  //
```
mm4tt: NIT: Indent

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -685,17 +689,23 @@ type BasicSchedulingPolicy struct {
...
+  // Note that the scheduler operates on an eventually consistent model. Updates
```
tosi3k: Nitpick - indentation.
```suggestion
	//
	// Note that the scheduler operates on an eventually consistent model. Updates
```

keps/sig-scheduling/4671-gang-scheduling/README.md
mm4tt: Shouldn't we get rid of all these 1.36 references and instead focus on current shape?

tosi3k:
```suggestion
#### Status Transition Rules
```

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1564,16 +1525,10 @@ This section must be completed when targeting alpha to a release.
 ###### How can this feature be enabled / disabled in a live cluster?
 
 - [X] Feature gate (also fill in values in `kep.yaml`)
-  - Feature gate name: GenericWorkload (alternatives: NativeWorkload/Workload)
+  - Feature gate name: GenericWorkload
   - Components depending on the feature gate:
```
tosi3k, line 1579: We should add `kube-controller-manager` here - PodGroup protection controller (added as a part of KEP-5832 that got merged with this KEP) runs if and only if `GenericWorkload` flag is enabled.

Ditto in `kep.yaml`.

I guess it would be also good to mention this in the `Scalability` section for the sake of completeness.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -938,6 +950,23 @@ Pods reference their `PodGroup` via `schedulingGroup.podGroupName`, which allows
 look up the `PodGroup` object. The scheduler requires `PodGroup` object to exist before scheduling 
 pods that reference them.
 
+### Pod Group minCount Mutability
```
tosi3k, line 927: Let's say we have a gang PodGroup which was scheduled and its Pods are already running for some time.

Now, let's assume the minCount value for that group got bumped up, new Pods appear but the scheduler is not able to fit those even with preemption.

What is going to happen with the already running Pods? Are they going to be evicted by the scheduler or will they continue to run as they are?

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added here, to `kep.yaml` and mentioned in `Scalability`

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Good point, added a sentence about this to the field documentation and to this KEP section. The already-scheduled pods should remain intact in such case

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1411,16 +1367,28 @@ calculating gang-level schedulability inside the event handler can be difficult
...
 we can start by skipping the Queueing Hints mechanism and relying solely on the backoff time.
+As a later step, we can decide whether to re-enable Queueing Hints for PodGroups with homogeneous pods only.
+
+### Feature gates merge in v1.37
 
+Initially, the KEP was introduced with two feature gates: `GenericWorkload` and `GangScheduling`.
+We identified that such a split is artificial and enabling only `GenericWorkload`
+might provide unclear results (as it enables the entire API but only a subset of kube-scheduler functionalities).
+
+In v1.37, both feature gates will be merged into one: `GenericWorkload`. Moreover, since workload-aware preemption
```
kannon92, line 1394: This is going to be sorta difficult to track. Is the goal that WAP will graduate at same time as gang scheduling?

Then I would maybe suggest we merge the KEPs also. The point of having separate feature gates for KEPs is to enable the features in isolation.

macsko: > Is the goal that WAP will graduate at same time as gang scheduling?

Yes. The preemption mechanism is necessary and we don't see a point in enabling the feature that doesn't support it. That's why having separate feature gates doesn't make sense, as they would need to graduate and be enabled in exactly the same way.

> Then I would maybe suggest we merge the KEPs also.

This KEP is already kinda large and operates on a different abstraction layer than the WAP KEP. Personally, I would keep them separate, pointing to each other where needed, unless there is a strong reason to merge them.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -225,7 +226,7 @@ See [Future plans](#future-plans) for more details.
 
 ## Proposal
 
-This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io/v1alpha2`. The 1.36 revision, 
+This KEP introduces both the `Workload` and `PodGroup` APIs in `scheduling.k8s.io/v1beta1`. The 1.36 revision,
```
wojtek-t: nit: I would remove the "/v1beta1" here - let's focus on the API group, as versions will change over time as it's visible here.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -821,18 +833,20 @@ got accepted by the Permit phase.
...
-This is a deliberate simplification for alpha. In practice, pods may later become unschedulable 
-(for example, due to node failures or evictions), but the `PodGroup` status will not reflect that. 
-A more robust status lifecycle that can capture post-scheduling state changes must be designed for 
-beta graduation.
+This is a deliberate simplification. In practice, scheduled pods may later be evicted
+or impacted by node failures, but the `PodGroup` status will not track
+these post-scheduling disruptions. While a more robust status lifecycle capable of reflecting
+post-scheduling state changes is a desirable future enhancement, we do not consider it
+a blocker for beta graduation. Handling such behavior is tied to rescheduling 
+mechanisms, which remain out of scope for this KEP.
```
wojtek-t: I would remove the last sentence - this is not necessarily tied to rescheduling. It may be used to trigger other actions by related, but different components.

Also, currently kube-scheduler, why it tracks pod terminations/removals to ensure its cache reflects the available resources, it doesn't take any external actions based on those.
So arguably, we should consider having a separate (at least logically) component for that.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1086,20 +1118,14 @@ The cycle proceeds as follows:
...
+Workload-aware preemption ([KEP-5710](https://github.com/kubernetes/enhancements/pull/5711))
 will introduce a specific scheduling priority for a workload.
-Having that in mind, it is beneficial to design a queueing mechanism open
-for taking a workload's scheduling priority into account.
-However, as we need to support ordering before that feature can be enabled,
-we also need to derive the priority from the pod group's pods.
-One such formula can be to set it to the lowest priority found within the pod group,
-what will be effectively the weakest link to determine if the whole pod group is schedulable
-and reduce unnecessary preemption attempts.
+Having that in mind, the queueing mechanism should support the workload's scheduling priority.
```
wojtek-t: PodGroup's scheduling priority

[this is what scheduler cares about]

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1146,84 +1171,88 @@ by sorting the Pods appropriately by their signatures. In case Opportunistic Bat
...
-       and retried even before the victims are fully removed, but they must pass through
-       the Workload Scheduling Cycle again. Crucially, initiating *new* preemptions
-       will be forbidden during this retry. This ensures that the pod group
-       can be scheduled in a different location if resources become available earlier,
-       but cannot cause additional disruption to do so.
-
-     * If preemptions are not needed: Pods proceed directly to their binding cycles
-       using the nodes selected during the Workload Scheduling Cycle.
+   * If `schedulableCount >= minCount`, the cycle succeeds. Pods proceed directly to their binding cycles
+     using the nodes selected during the Workload Scheduling Cycle.
```
wojtek-t: What we've been talking offline couple weeks ago was if we don't have to run the preemption here.

If we managed to schedule at least minCount pods, the group is technically schedulable.
But if there remained some unscheduled pods and we can actually make place for them via preempting other lower-priority workloads, we should actually still do that.

Now the remaining question is whether:
1) we immediately send the schedulable pods to binding
or
2) we actually trigger preemption to find space for the missing ones, because as part of that we may actually find a more optimal space for the whole group

The answer to that is not a clear tradeoff to me.
On one hand I would like to have better placement, but on the other hand preemption may take significant time and I don't want to unnecessary delay startup of this workload.
But I would personally lean towards not unnecessary delaying startup of those pods.

So I believe what should happen is:
1) if schedulable >= minCount the cycle succeeds and pods move to binding
2) if some pods remained unschedulable - trigger the preemption for them
  [now the question is whether it's the same cycle - which would be a bit weird, because the cycle technically already succeeded; or we somehow immediately trigger the next cycle to handle the remaining pods]

@macsko @sanposhiho @dom4ha @mm4tt

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: +1 to Maciek
They are both pretty complex and merging them would result in a giant beast hard to read.

That being said, we should add a bit note at the beginning of the PRR section that these 2 KEPs will progress in lockstep.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1446,20 +1414,19 @@ Initially, we created integration tests to ensure the basic functionalities of g
...
 - Pods referencing a `PodGroup` (both gang and basic policies) are correctly processed via the Workload Scheduling 
   Cycle.
 - `PodGroup` queuing ensures that all available members are retrieved and processed correctly.
 - Deadlocks and livelocks do not occur when multiple gangs compete for resources or interleave with standard pods.
-- Delayed Preemption feature doesn't break pod-by-pod (non-workload) scheduling.
-- Delayed Preemption ensures atomicity, i.e., victims are deleted only if the scheduler determines the 
-entire gang can fit, otherwise, the cycle aborts with zero disruption.
 - Failed pod groups are requeued correctly and retry successfully when resources become available.
+- Scheduler correctly captures updated `minCount` value for pending pod groups, potentially unblocking them from PreEnqueue.
+- Changes to `minCount` value during Workload Scheduling Cycle doesn't affect the current cycle, but the subsequent ones.
```
wojtek-t: This sounds like something fairly tricky to test and inherently racy (I mean if this requirement wouldn't be satisfied, it will be very hard to make this test consistently fail).

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -821,18 +833,20 @@ got accepted by the Permit phase.
...
 state and does not revert to `False`. Specifically, once the group's scheduling constraint (`minCount`) 
 has been satisfied, subsequent failed scheduling cycles for additional pods beyond `minCount` do not 
 regress the condition. On same-status transitions (e.g., `True` → `True`), the condition message may 
 be updated, but `LastTransitionTime` remains unchanged.
 
-This is a deliberate simplification for alpha. In practice, pods may later become unschedulable 
-(for example, due to node failures or evictions), but the `PodGroup` status will not reflect that. 
-A more robust status lifecycle that can capture post-scheduling state changes must be designed for 
-beta graduation.
+This is a deliberate simplification. In practice, scheduled pods may later be evicted
```
mm4tt: I agree we shouldn't track runtime status here, but calling it a "deliberate simplification" leaves the actual semantics ambiguous. Here is how I see it:

1. First, a status condition can reflect either scheduling decisions or runtime state. For a single Pod, these effectively mean more or less the same thing because a Pod is scheduled only once. For a PodGroup, this is not true. A PodGroup can be evaluated multiple times as new pods appear, triggering new scheduling cycles.

2. Second, the scheduler currently lacks the mechanisms to track runtime state (e.g., evictions, node failures). If we ever want the scheduler to handle rescheduling and broader runtime failure handling, it will be a paradigm shift requiring a separate KEP. For now, we must operate within the existing framework and restrict this status strictly to scheduling decisions.

3. However, even with these boundaries established, we cannot just leave it as is. The exact scheduling semantics remain unclear. We need to explicitly choose and document one of two approaches:

   1. Latched Status: The condition turns True once minCount is satisfied and never reverts. Crucially, if some pods are later deleted and new pods are added, the scheduler runs a new cycle. Even if this new cycle fails to satisfy minCount, the status still remains True.

   2. Latest Cycle Status: The condition always reflects the outcome of the most recent scheduling cycle. If new pods are added, triggering a new cycle, the status updates to reflect whether the PodGroup is currently schedulable (e.g., it reverts to False if the current state no longer satisfies minCount).

Regardless of which option we choose, we must explicitly document these exact semantics instead of waving them off as a beta simplification. Depending on the choice, we also need to ensure the implementation actually matches it (e.g. if we go with 2). 

Personally, I think I like 2 better given it provides up2date scheduling decision / status.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1504,28 +1470,28 @@ In 1.36:
...
 - A deletion protection mechanism is implemented for `PodGroup` objects and finalizer is added to the API.
-- All e2e tests for `PodGroup` are added and graduate to conformance tests.
+- All e2e tests for `PodGroup` are added.
 - Performance tests are created and are being run in CI to protect against regressions.
 
 #### GA
 
-- TBD in for Beta release
-
+- All issues and gaps identified as feedback during beta are resolved
```
wojtek-t, line 1515: Promotion tests to conformance:
1) API tests
2) I would also push for having some gang-scheduling test to be part of conformance.

wojtek-t (commented):
Great update - added few comments, the only bigger one is about triggering preemption when some pods remain unschedulable

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: I was going back and forth with writing something similar and finally decided that we can live with (1). But reading your comment now, I think that it's still probably OK *if* we would adjust the name of the condition.
`PodGroupScheduled` suggests the current state and that is not really what it is reflecting.
So maybe we should rename it to something different (PodGroupScheduledOnce? - I don't like it but I don't have a better name).

Latest Cycle Status - I don't like this semantic, because it reflects last "scheduling attempt" not really what is happening with the group. The group can technically be fully scheduled (satisfying all constraints including minCount) but e.g. the additional pod (above minCount) couldn't scheduled.
If we want something like that - the name should clearly reflect that it's not about current state, but rather the status of the last scheduling cycle.

keps/sig-scheduling/4671-gang-scheduling/README.md
Argh4k: > because as part of that we may actually find a more optimal space for the whole group 

Is that relevant for the non TAS pod scheduling cycle? What would more optimal placement in a non TAS case mean? I think that actually this problem is more of a TAS problem, whether we should take placement that meets minCount or we should try to look for a placement that with preemption can give us placement that will fit all pods.

keps/sig-scheduling/4671-gang-scheduling/README.md
kannon92: > That being said, we should add a bit note at the beginning of the PRR section that these 2 KEPs will progress in lockstep.

+1.

Will we keep the WAP KEP up to date? Remove the feature gate and update design in tandem with this one.

Argh4k: Yes, I am working on updating WAP KEP for 1.37.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: TAS is the clearest example, but the exact same thing you can achieve with pod affinity/anfi-affinity.

But even ignoring hard constraints, we can phrase this same problem in the context of scores - if we approximate score(PodGroup) as sum of scores of individual pods for non-TAS case, with preemption we may be able to find visibly better place.
Now - we don't do that if we can find a place without preemption, but given we didn't find it for all pods and we need to preempt something, we may actually find a better one (e.g. if we will preempt some large low-priority job).

macsko: I think it depends on how we treat the minCount. If we say that the minCount is the main schedulability condition and later pods can be considered best-effort, then we can send schedulable pods to binding, move unschedulable pods back to the scheduling queue, and then trigger the preemption in a subsequent scheduling cycle, if it's still needed. I would prefer this option, as it would be the easiest to implement and, I think, pretty easy to understand by the users.

The problem would be when we say that we **always** want to find the best placement out of the current cluster state for **all** pods that we have. Obviously, it still might not be the best placement because we (will) optimize the number of nodes/placements to score, but the preemption could help us achieve something better than we have without it. In this case, I wouldn't like to send subset of pods to binding and trigger the preemption for the rest. It would be unintuitive in the code. However, if we run the preemption with all the pods, we may unnecessarily delay the startup of schedulable pods, because they would have to come back to the scheduling queue to wait for the victims to terminate.

Given that, I'm leaning towards executing the preemption in a subsequent cycle for unschedulable pods.

wojtek-t: > Given that, I'm leaning towards executing the preemption in a subsequent cycle for unschedulable pods.

This seems to be aligned with what I wrote in my first comment in this thread.
Now my primary question is - can we ensure that the remaining pods will actually not be send to the back of the queue, and instead land on its front (obviously assuming there aren't higher priority ones added in the meantime)?

macsko: > Now my primary question is - can we ensure that the remaining pods will actually not be send to the back of the queue, and instead land on its front (obviously assuming there aren't higher priority ones added in the meantime)?

It's a bit tricky. We could send the pod group directly to the activeQ after such attempt, but it would have a newer timestamp (we calculate the order based on the timestamp), so it will be considered later than the equal priority pod(group)s. We could use the previous timestamp, although this might be unintuitive in the code, it shouldn't cause any problems.

wojtek-t: > although this might be unintuitive in the code, it shouldn't cause any problems.

That was primarily my question.
If this "should work", I think it's effectively what we should do to strengthen the story of "workload is scheduled together".

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: > The group can technically be fully scheduled (satisfying all constraints including minCount) but e.g. the additional pod (above minCount) couldn't scheduled.
> If we want something like that - the name should clearly reflect that it's not about current state, but rather the status of the last scheduling cycle.

+1, Also, even if the minCount is no longer met, the PodGroup has scheduled pods that may run, and in case of TAS, the selected placement which new pods will follow. So, I think it's beneficial to keep the current condition.

If we need the "latest cycle status", we could create a new condition to capture it pretty easily.

> So maybe we should rename it to something different (PodGroupScheduledOnce? - I don't like it but I don't have a better name).

`PodGroupInitiallyScheduled`?

mm4tt: > Latest Cycle Status - I don't like this semantic, because it reflects last "scheduling attempt" not really what is happening with the group. The group can technically be fully scheduled (satisfying all constraints including minCount) but e.g. the additional pod (above minCount) couldn't scheduled.

I think my previous point might not have been entirely clear. In this exact scenario, in option 2., the group should have the Scheduled status because the user intent (minCount) is satisfied. Individual pods above the minCount might remain Unscheduled, but the group as a whole meets the user's specified requirements. 

To me, this is the "relatively cleanest" approach and I would argue that Option 2 actually reflects "what is happening with the group" better than Option 1. As noted above, if minCount is met, the user intent is satisfied and the group is considered scheduled, even if additional pods are not. With Option 1, the user loses visibility after the first successful scheduling attempt. If a disruption occurs, pods terminate, and the scheduler subsequently fails to place the recreated pods (dropping the feasible count below minCount), the status remains Scheduled. This is misleading. Option 2 updates the status based on whether minCount is satisfied in the most recent cycle. This provides an up-to-date answer to whether the user intent is currently met by the scheduler. And obviously this also requires thorough documentation regarding its main limitation. Because it is strictly a scheduling status, if pods are disrupted and not recreated (meaning no new scheduling cycle is triggered), the status will not update. This can also be confusing.


If we decide to go with latching the first successful state, I agree we must rename the condition and document its semantics thoroughly. @macsko's proposal (PodGroupInitiallyScheduled) seems fine. It accurately reflects that we are recording the first successful scheduling attempt (even if preceded by multiple failed attempts before) and ignoring subsequent state changes.

wojtek-t: OK - so it's not really about "Last Cycle State", but whether the PodGroup satisfies the conditions.
I agree this will be useful, but the primary question is whether it's really kube-scheduler who should be setting that. The "healthiness" of PodGroup can break as a consequence of e.g. disruption of a pod not even triggered by scheduler and if either there will not be recreated pod it will never be updated.

To be clear - I think we actually need both (Maciek above pointed more good arguments why the current condition is beneficial - e.g. locked in TAS placement).
The primary question is - what is a beta blocker and what can be added later.

Personally, I'm leaning towards say that:
1) we keep the existing condition and only that is the beta blocker - we just rename it - I'm fine with `PodGroupInitiallyScheduled`
2) we postpone the "PodGroup status" for now and will address it in a follow-up

macsko: Initially, I thought it wouldn't be simple to handle the "latest cycle status" in the code, but I realized that with the new PlacementFeasible extension point it will be fairly easy. So, apart from API changes (description?), it should be quick to implement the support for both statuses.

I see the benefits of both options, so I think the primary question is even whether we have the capacity to implement the "latest cycle status" this release or not.

> I agree this will be useful, but the primary question is whether it's really kube-scheduler who should be setting that. The "healthiness" of PodGroup can break as a consequence of e.g. disruption of a pod not even triggered by scheduler and if either there will not be recreated pod it will never be updated.

I think we are mixing now different concepts. In my opinion, live tracking the state (maybe including the current pod counts, I remember such idea being raised some time ago) of the pod group shouldn't be done by kube-scheduler, but rather a different component. The scheduler, however, can display its current decision in the status, which is also important.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added a few more sentences here and there in the KEP regarding both KEPs graduations

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Removed unnecessary sentences, PTAL

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: I rewrote that paragraph completely based on a discussion from different thread

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added to the algorithm description that the preemption will be attempted in the subsequent cycle and the pod group will be requeued with the old timestamp in case it was mostly scheduled.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: That's right. Removed it from this list

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added a sentence about both

macsko:
I pushed a commit that applies the pending comments, PTAL

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: Thank you - this looks great now!

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1617,30 +1571,53 @@ This section must be completed when targeting beta to a release.
...
-are missing a bunch of machinery and tooling and can't do that now.
--->
+We'll perform manual testing of the upgrade -> downgrade -> upgrade path using the following sequence:
+
+1. Start a local Kubernetes v1.36 cluster with GenericWorkload feature gate disabled (default 
+behavior).
+2. Attempt to create a Pod with `spec.schedulingGroup` set.
+3. The `spec.schedulingGroup` field is dropped by the API server. The pod is created successfully
+   but without the `schedulingGroup` reference, resulting in immediate standard scheduling (one-by-one).
+4. Restart/Upgrade API Server and Scheduler to v1.37 (with feature gates enabled).
```
wojtek-t: I would remove brackets - the FG will be disabled by default in 1.37 because the API will be disabled by default.
So let's clarify this.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1617,30 +1571,53 @@ This section must be completed when targeting beta to a release.
...
+7. The Pod stays in `Pending` state (waiting for the gang). Verify that
+   `scheduler_pending_entities{type="podgroup", queue="gated"}` metric is incremented.
+8. Create a Pod `test-pod-2` pointing to the same pod group.
+9. Both pods are scheduled successfully in the same cycle (Gang Scheduling works). 
+10. Downgrade API Server and Scheduler to v1.36 (with feature gates disabled).
+11. Create `test-pod-3` pointing to `gang-test-B`. Note: We use a pod group created in step 5 because creating new
+    PodGroup objects is disabled.
+12. The pod is scheduled immediately (PodGroup logic is ignored because the schedulingGroup field is dropped by
+    the v1.36 API server). If Gang Scheduling were active, this pod would hang pending waiting for a second member.
+13. Upgrade API Server and Scheduler back to v1.37 (feature gates enabled).
```
wojtek-t: Again - remove brackets, because you need to explicitly enable FG.

wojtek-t (commented):
Just some super minor nits from then - but overall it looks great.

/lgtm

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Done

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Done

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -125,11 +127,10 @@ their resources.  This allows other workloads to try to allocate those resources
 New core types called `Workload` and `PodGroup` are introduced to tell the kube-scheduler that a 
 group of pods should be scheduled together and to define policy options related to gang scheduling. 
 Pods may have an object reference in their spec to the `PodGroup` they belong to. The `Workload` 
-and `PodGroup` objects are intended to evolve[^2] via  future KEPs to support additional 
+and `PodGroup` objects are intended to evolve[^2] via future KEPs to support additional 
 kube-scheduler improvements, such as topology-aware scheduling.
 
-In 1.36, we redesigned the API in order to clearly decouple the Workload API from the runtime PodGroup API [^4].
-In the updated design:
+In the API:
```
helayoty: Can we update it to something like:
```
The API is structured to decouple policy from runtime grouping:
```
rather than the bare "In the API"? The reader who is unfamiliar with the history might benefit from it.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -938,6 +924,36 @@ Pods reference their `PodGroup` via `schedulingGroup.podGroupName`, which allows
...
+  may still be evaluated for that cycle. The new value will take effect in the subsequent cycle.
+
+* `minCount` is enforced strictly during scheduling, not at runtime. This means modifications to the field
+  will not affect already-scheduled pods, applying only to those evaluated in future scheduling cycles.
+
+While broader PodGroup spec mutability, such as modifying the number of `PodGroupTemplates` in a `Workload`,
+may be desirable, we are strictly scoping API mutability in v1.37 to the `minCount` field.
+Further relaxation of validation rules will be considered in the future if driven by strong use cases.
+
+### Workload Controllers Integration
```
helayoty, line 947: We should add link to the KEP (#5547). And it would be good to add feedback from these projects about the integration or list issues.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Good idea, updated

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added link to the KEP. Regarding adding the feedback, I don't think we need that level of details, listing the exact feedback. The confirmation that we go in a good direction and don't miss anything should be enough, and further discussion and design should be done on the controllers side. We don't want to keep such section being constantly updated based on the controllers integration stage.

wojtek-t:
For PRR review

/assign @soltysh

keps/sig-scheduling/4671-gang-scheduling/README.md
sanposhiho, line 1498: nit: better to mention that now each graduation needs to meet the criteria mentioned on KEP-5710 side too.

sanposhiho (commented):
overall looking great. One tiny nit, 

/lgtm

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -809,30 +787,38 @@ in the code to allow for extending this structure if needed in the future.
...
+The `PodGroup.Status` is managed by kube-scheduler to reflect the scheduling status. We introduce
+a `Conditions` field with the `PodGroupInitiallyScheduled` condition type.
 
 `PodGroup` status mirrors `Pod` status semantics rather than defining PodGroup-specific reasons:
 - If pods are unschedulable (i.e., timeout, resources, affinity, etc.), the scheduler updates the 
-`PodGroupScheduled` condition to `False` and sets the reason fields accordingly.
-- If pods are scheduled, the scheduler updates the `PodGroupScheduled` condition to `True` after the group 
+`PodGroupInitiallyScheduled` condition to `False` and sets the reason fields accordingly.
+- If pods are scheduled, the scheduler updates the `PodGroupInitiallyScheduled` condition to `True` after the group 
 got accepted by the Permit phase.
```
mm4tt: We should get rid of / replace those Permit references if we're planning to retire its usage with PlacementFeasible.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -938,6 +924,36 @@ Pods reference their `PodGroup` via `schedulingGroup.podGroupName`, which allows
 look up the `PodGroup` object. The scheduler requires `PodGroup` object to exist before scheduling 
 pods that reference them.
 
+### Pod Group minCount Mutability
+
```
mm4tt, line 928: I would clarify that minCount is a "scheduling constraint" and not a "runtime constraint". This means if there is a scheduled PodGroup with minCount=10 and 10 pods, and the minCount is increased to 20 without creating new pods, those 10 pods will remain running even though the condition is no longer met. Consequently, given how we report PodGroup status, it will also remain "Scheduled".

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: In this specific case, the mention of Permit was incorrect. Fixed the wording.

However, currently we are not planning to get rid of Permit. It still works as a barrier before the pods are send to binding, to reverify the constraints.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: We have "`minCount` is enforced strictly during scheduling, not at runtime. This means modifications to the field will not affect already-scheduled pods, applying only to those evaluated in future scheduling cycles."

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Good point, added

macsko:
@soltysh PTAL at PRR section

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -740,11 +733,13 @@ type PodGroupStatus struct {
     // Conditions represent the latest observations of the PodGroup's state.
     //
     // Known condition types:
-    // - "PodGroupScheduled": Indicates whether the scheduling requirement has been satisfied.
+    // - "PodGroupInitiallyScheduled": Indicates whether the scheduling requirement has been satisfied.
```
dom4ha, line 736: Frankly I'm not sure if adding "initially" make it any cleaner. I feel like the original name is good enough to reflect that it holds the initial status.

I'd suggest to keep it consistent with Pod status as I don't see a clear justification for this change.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -938,6 +924,36 @@ Pods reference their `PodGroup` via `schedulingGroup.podGroupName`, which allows
...
+### Pod Group minCount Mutability
+
+To support workload scaling (e.g. for elastic Jobs), the `minCount` field in both `PodGroup`
+and `PodGroupTemplate` is proposed to be mutable in v1.37. Specifically:
+
+* Modifying a `PodGroupTemplate` will not affect existing `PodGroup` instances.
+  Changes should apply exclusively to new instances created from the updated template by the controllers.
+
+* Because the scheduler operates on an eventually consistent view of the cluster state,
+  updates to `minCount` may not be immediately visible. If a modification occurs while the scheduler
```
dom4ha:
```suggestion
* In case of `PodGroup` instances, updates to `minCount` may not be immediately visible,
as the scheduler operates on an eventually consistent view of the cluster state. If a modification occurs while the scheduler
```

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1040,10 +1058,6 @@ to process the entire gang together.
 The single scheduling cycle, together with blocking resources using nomination,
 will address requirement (3).
 
-We will also introduce [Delayed Preemption](#delayed-preemption).
```
dom4ha, line 1043: Maybe we could call the section "### Scheduler Changes release history" and describe briefly 1.37 scope:
- drop delayed preemption in favor of Workload Aware Preemption
- introduction of minCount mutability
- most of all, promotion to beta
- new extension point
- feature gates merge
- etc.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1126,8 +1134,7 @@ The queue will support queuing `PodGroup` instances alongside individual Pods.
    Based on the cycle's outcome:
```
dom4ha, line 1134: I think we have just added a case in which PodGroup goes directly into ActiveQ after scheduling attempt at least in the following cases:
- preemption above minCount
- new pods that appeared during scheduling

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1617,30 +1571,53 @@ This section must be completed when targeting beta to a release.
...
+3. The `spec.schedulingGroup` field is dropped by the API server. The pod is created successfully
+   but without the `schedulingGroup` reference, resulting in immediate standard scheduling (one-by-one).
+4. Restart/Upgrade API Server and Scheduler to v1.37 with feature gates enabled.
+5. Create two PodGroup objects: `gang-test-A` and `gang-test-B` (both with `minCount=2`).
+6. Create a Pod `test-pod-1` with `spec.schedulingGroup` pointing to `gang-test-A`.
+7. The Pod stays in `Pending` state (waiting for the gang). Verify that
+   `scheduler_pending_entities{type="podgroup", queue="gated"}` metric is incremented.
+8. Create a Pod `test-pod-2` pointing to the same pod group.
+9. Both pods are scheduled successfully in the same cycle (Gang Scheduling works). 
+10. Downgrade API Server and Scheduler to v1.36 with feature gates disabled.
```
dom4ha:
```suggestion
10. Downgrade API Server and Scheduler to v1.36 with feature gates (`GenericWorkload` and `GangScheduling` defined in that release) disabled.
```

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1289,81 +1311,9 @@ optional. In the v1.36 timeframe, this cycle will be applied to
...
-
-The rationale behind the above design is to maintain the current scheduling property where preemption
-doesn't result in a commitment for a particular placement. If a different possible placement appears
-in the meantime (e.g. due to other pods terminating or new nodes appearing), subsequent scheduling
-attempts may pick it up, improving the end-to-end scheduling latency. Returning pods to scheduling
-queue if these need to wait for preemption to become schedulable maintains that property.
-
 #### Workload-aware Preemption
 
+Proper preemption support is a critical requirement for moving Gang Scheduling to Beta.
```
dom4ha, line 1335: I think it's worth to add a paragraph about per-pod PostFilter status. Currently we skip it only for preemption plugin, but execute it for other plugins.

They need to eventually migrate to the PodGroup level PostFilter (introduced in WAP KEP) as future WAS features will require repeating the scheduling process several times, so per-pod PostFilter will be eventually disabled for PodGroups.

It's possible that the migration will come with the PostFilter refactoring effort which may simplify that process.

dom4ha (commented):
Overall LGTM.

I left a comment to bring a paragraph about per-pod PostFilter migration.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: Pod - once scheduled remains scheduled.
This is not true for PodGroup, where individual pods may come and go.

PodGroupScheduled suggests that the group remains scheduled, and it may not really be the case.

PodGroupInitiallyScheduled may not be the best name and I'm open to hear other suggestions, but I'm pretty sure it better reflects what it is that PodGroupScheduled.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: This comment should go to WAS KEP, not to this KEP.

/cc @Argh4k

dom4ha: It's definitely a comment for WAP, but this KEP also mentions disabling per-pod default preemption, so it may be worth to mention that other PostFilter executions stays (agree it does not have to be a separate section).

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: This KEP is already quite complicated and contains multiple sections written based on the specific releases' states. I would like to avoid having further sections like this that will be outdated soon anyway.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: > It's definitely a comment for WAP, but this KEP also mentions disabling per-pod default preemption, so it may be worth to mention that other PostFilter executions stays (agree it does not have to be a separate section).

Will other PostFilter executions stay? Didn't want we to move all PostFilter use cases to PodGroupPostFilter?

Also, I don't think this KEP should focus that much on the WAP KEP outcomes. We can always adjust the wording here and there later during the release cycle.

keps/sig-scheduling/4671-gang-scheduling/README.md
dom4ha: Ack.

keps/sig-scheduling/4671-gang-scheduling/README.md
dom4ha: This section already contains a history of changes for 1.35 and 1.36, so maybe we should either drop it as well?

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: > Also, I don't think this KEP should focus that much on the WAP KEP outcomes. We can always adjust the wording here and there later during the release cycle.

+1 - let's move all those things to WAP KEP

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Changed

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: I adjusted this section now to minimize references to specific versions. I know that we could get rid of more paragraphs, but it could be done in a cleanup later.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Good point. Added sentences about both

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1086,20 +1100,14 @@ The cycle proceeds as follows:
...
-Workload-aware preemption (an `Alpha` effort in [KEP-5710](https://github.com/kubernetes/enhancements/pull/5711))
-will introduce a specific scheduling priority for a workload.
-Having that in mind, it is beneficial to design a queueing mechanism open
-for taking a workload's scheduling priority into account.
-However, as we need to support ordering before that feature can be enabled,
-we also need to derive the priority from the pod group's pods.
-One such formula can be to set it to the lowest priority found within the pod group,
-what will be effectively the weakest link to determine if the whole pod group is schedulable
-and reduce unnecessary preemption attempts.
+Workload-aware preemption ([KEP-5710](https://github.com/kubernetes/enhancements/pull/5711))
```
soltysh: It's best to link to the tracking KEP issue https://github.com/kubernetes/enhancements/issues/5710 rather than actual PRs 😉

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1446,20 +1413,18 @@ Initially, we created integration tests to ensure the basic functionalities of g
...
-entire gang can fit, otherwise, the cycle aborts with zero disruption.
 - Failed pod groups are requeued correctly and retry successfully when resources become available.
+- Scheduler correctly captures updated `minCount` value for pending pod groups, potentially unblocking them from PreEnqueue.
 
 We will also benchmark the performance impact of these changes to measure:
 
-- The scheduling throughput of the workload scheduling, including gang and basic policies, and preemptions.
+- The scheduling throughput of the workload scheduling, including gang and basic policies.
 
 ##### e2e tests
```
soltysh, line 1455: Can you please make sure to update both integration and e2e sections with the tests you've already written? Those are required for beta graduation.

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1504,28 +1468,34 @@ In 1.36:
...
-- TBD in for Beta release
+- All issues and gaps identified as feedback during beta are resolved
+- Promote the e2e API tests to conformance together with a test for gang scheduling behavior.
 
+This KEP and workload-aware preemption ([KEP-5710]) are tightly coupled and controlled by the same feature gate.
+They graduate in lockstep, meaning the graduation criteria of [KEP-5710] should be also met when targeting to a release.
 
 ### Upgrade / Downgrade Strategy
 
 This KEP is completely additive and can safely fallback to the original behavior on downgrade.
```
soltysh, line 1523: Can you explain how the merge of the two FGs (GangScheduling and GenericWorkload) will affect upgrade/downgrade paths? We error out if you specify unknown FG, which might be a problem during upgrade, and similarly the lack of the other FG during downgrade might cause unexpected behaviors, right?

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1588,26 +1550,23 @@ those are not yet created automatically behind the scenes.
...
-cleared, as well as the existing Workload objects will not be deleted.
+Yes. The `GenericWorkload` feature gate needs to be switched off to disable the API changes, admission and gang scheduling
+functionality. However, the content of `spec.schedulingGroup` fields in Pod objects will not be
+cleared, as well as the existing Workload and PodGroup objects will not be deleted.
 
 ###### What happens if we reenable the feature if it was previously rolled back?
 
 The feature should start working again.
-However, the user need to remember that some Workload objects could already be stored
+However, the user need to remember that some Workload and PodGroup objects could already be stored
```
soltysh:
```suggestion
However, the user needs to remember that some Workload and PodGroup objects could already be stored
```

keps/sig-scheduling/4671-gang-scheduling/README.md
soltysh: Given that you're entirely removing the v1alpha2 version, should we also call out here explicitly that users should remove all the resources stored with that version, b/c they won't be able to reach to them after the upgrade?

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1588,26 +1550,23 @@ those are not yet created automatically behind the scenes.
...
 ###### Are there any tests for feature enablement/disablement?
 
-No.
-The enablement/disablement for the new field in Pod API will be added similarly to this PR:
-https://github.com/kubernetes/kubernetes/pull/97058/files#diff-7826f7adbc1996a05ab52e3f5f02429e94b68ce6bce0dc534d1be636154fded3R246-R282
+The feature has unit tests that verifies enablement and disablement of the `schedulingGroup` field in Pod API:
+https://github.com/kubernetes/kubernetes/blob/cc9fd614c429f43e67fd8ded5e160fd784d860a8/pkg/registry/core/pod/strategy_test.go#L3904
 
-Note that gang-scheduling itself is purely in-memory feature, so feature themselves are enough.
+Note that gang-scheduling itself is purely in-memory feature, so features themselves are enough.
```
soltysh: Maybe something like: 
```suggestion
Note that gang-scheduling itself is purely in-memory feature, so these feature tests are enough.
```
will be better sounding?

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1617,30 +1576,54 @@ This section must be completed when targeting beta to a release.
...
+4. Restart/Upgrade API Server and Scheduler to v1.37 with feature gates enabled.
+5. Create two PodGroup objects: `gang-test-A` and `gang-test-B` (both with `minCount=2`).
+6. Create a Pod `test-pod-1` with `spec.schedulingGroup` pointing to `gang-test-A`.
+7. The Pod stays in `Pending` state (waiting for the gang). Verify that
+   `scheduler_pending_entities{type="podgroup", queue="gated"}` metric is incremented.
+8. Create a Pod `test-pod-2` pointing to the same pod group.
+9. Both pods are scheduled successfully in the same cycle (Gang Scheduling works). 
+10. Downgrade API Server and Scheduler to v1.36 with feature gates
+    (`GenericWorkload` and `GangScheduling` defined in that release) disabled.
+11. Create `test-pod-3` pointing to `gang-test-B`. Note: We use a pod group created in step 5 because creating new
```
soltysh, line 1660: I don't think this will work for you. Remember that v1.37 will have either v1alpah3 or v1beta1 versions, both of which are not available in 1.36, so this might not work for you, or am I missing something?

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1617,30 +1576,54 @@ This section must be completed when targeting beta to a release.
...
+    PodGroup objects is disabled.
+12. The pod is scheduled immediately (PodGroup logic is ignored because the schedulingGroup field is dropped by
+    the v1.36 API server). If Gang Scheduling were active, this pod would hang pending waiting for a second member.
```
soltysh: No, it would not, b/c the podgroup would not be visible due to version mismatch. See my previous comment.

keps/sig-scheduling/4671-gang-scheduling/README.md
soltysh: I believe you'll need to update this, b/c you're dropping v1alpha2, right?

soltysh: Can you fix this typo here, should be `dependencies`

soltysh, line 1746: Can you update the `XX` numbers in this and followup section with actual values?

soltysh (commented):
Left a few comments from PRR review

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1146,84 +1155,100 @@ by sorting the Pods appropriately by their signatures. In case Opportunistic Bat
...
-     can be attempted on that place. See *Failure Handling*.
+     In this successful case, preemption will not be attempted even if some pods remain unschedulable
+     This means that as long as the scheduling constraints (`minCount`) are met, the schedulable pods
+     will move to binding. Any subsequent preemption for unschedulable pods will be handled in the next
+     [Workload Scheduling Cycle](#workload-scheduling-cycle) if still needed. Triggering both binding
+     and preemption in the same cycle would be ambiguous, and such precedence wouldn't be clear
+     from the scheduler's perspective. Alternatively, always attempting preemption to free up space,
+     even for schedulable groups, would be unnecessarily disruptive and delay the startup of the group. 
+     
+     Unschedulable pods will be requeued using their old timestamp, meaning the subsequent scheduling
```
mm4tt: Let's make it clear what condition will actuallu trigger preemption for them. 

According what we have in the text, the preemption will not be triggered in the next cycle. 

Let's assume we have a PG with 15 pods and `MinCount=10`. If 12 pods succeed in the first pass and 3 fail, the flow looks like this:

1. `schedulableCount >= 10`: We bind the 12 pods, do not trigger preemption, and finish the cycle.
2. The 3 remaining pods are requeued, so the PG enters the scheduling cycle again.
3. We still have `schedulableCount >= 10`, so according to the condition preemption is not triggered

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1146,84 +1155,100 @@ by sorting the Pods appropriately by their signatures. In case Opportunistic Bat
...
+     and preemption attempt should start immediately after the current cycle,
```
mm4tt: Running preemption in subsequent cycles effectively means running preemption for an already scheduled `PodGroup`. I think this becomes problematic for TAS for three reasons:

1. This is similar to what you pointed out in the CPG KEP discussion ([#6017](https://github.com/kubernetes/enhancements/pull/6017#discussion_r3304252339)), we choose a placement based on the current number of schedulable pods. In subsequent cycles, those placements may not fit extra pods. Conversely, running preemption from the beginning allows us to choose alternative placements that can accommodate the entire `PodGroup`.
2. The current design for the TAS preemption algorithm relies on re-generating placements. However, in this scenario, we cannot re-generate placements because they are already locked in for the currently running pods and cannot be changed.
3. To be clear, this issue exists today and would exist anyway even if we went with a single greedy cycle with preemption - this will happen anyway if pods are added to a `PodGroup` later. We likely need to accept this tradeoff and acknowledge that it won't work ideally in all cases, which is simply a property of TAS+WAP. This stems from the fact that pod-level preemption is much simpler because the preemptor itself is never already scheduled, which is no longer true for `PodGroups`.

@wojtek-t @Argh4k

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: Good callout.
The condition should actually be:
1) not every pod has been scheduled
and
2) it is NOT an initial scheduling of the PG

The alternative for (2) might be "we didn't find a place for any pods", but if the PG is already scheduled, I think we should prioritize finding the place for as many pods as possible over quickly adding a subset of pods (and soon after adding more).
The reasoning is that adding more pods to a gang generally has some overhead anyway (e.g. requires pod restart) so let's avoid triggering this twice.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: > Given that you're entirely removing the v1alpha2 version, should we also call out here explicitly that users should remove all the resources stored with that version, b/c they won't be able to reach to them after the upgrade?

That's actually a good question. v1alpha2 was added as part of Alpha.
We had multiple cases in the past when doing breaking changes between Alpha and Beta and removing Alpha APis together. I think we were generally not describing that in the PRR (or at least I wasn't bringing that up) according to the rule: "if you're using Alpha features, you do that on your own risk".

I'm not against mentioning things like this, but is that a requirement? Might worth discussing during PRR meeting.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: Great catch.

The fix for that is changing point (10) above - instead of running 1.36 with FG disabled, let's simply run 1.37 with FG disabled.
That would address this problem.

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: But it was introduced as part of this KEP.
So technically we added and removed - if you don't enable alpha APIs - it doesn't matter.
And if you do - you do that on your own risk anyway...

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: Yes - I agree it will be suboptimal and may lead to unschedulability of those additional pods. But it doesn't break things.

I commented on that in https://github.com/kubernetes/enhancements/pull/6017#discussion_r3309809542 already

macsko: Yes, TAS is an example of a behavior where trying the preemption, even with schedulable pods, would be beneficial. However, the motivation for not doing the preemption still holds - we don't want to delay the startup of schedulable pods. So, I agree that we have to accept this tradeoff and document it on the TAS + WAP level.

> 1. This is similar to what you pointed out in the CPG KEP discussion ([#6017](https://github.com/kubernetes/enhancements/pull/6017#discussion_r3304252339)), we choose a placement based on the current number of schedulable pods. In subsequent cycles, those placements may not fit extra pods. Conversely, running preemption from the beginning allows us to choose alternative placements that can accommodate the entire `PodGroup`.

It's different, because compared to the CPG case, preemption is much heavier and may cause significant delays, waiting for victims to terminate.

> 2\. The current design for the TAS preemption algorithm relies on re-generating placements. However, in this scenario, we cannot re-generate placements because they are already locked in for the currently running pods and cannot be changed.

In such case, TAS will generate only one placement, matching already-scheduled pods, so the preemption will consider victims only in this specific placement.

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: I'll adjust this section to capture this scenario correctly.

> The alternative for (2) might be "we didn't find a place for any pods", but if the PG is already scheduled, I think we should prioritize finding the place for as many pods as possible over quickly adding a subset of pods (and soon after adding more).
> The reasoning is that adding more pods to a gang generally has some overhead anyway (e.g. requires pod restart) so let's avoid triggering this twice.

I don't understand that you meant here. Could you clarify how  "it is NOT an initial scheduling of the PG" is better than "we didn't find a place for any pods"? Is is that because we already met the contract (scheduled minCount), so we can try the preemption, even when some new pods are schedulable?

How would that play with the case where the minCount is no longer met (some pods disappeared) and the scheduler has a decision to either schedule up to minCount (and try preemption in subsequent cycle for other pods) or preempt (without scheduling schedulable pods)? Also, what would we want to do with the basic policy?

wojtek-t: > Is is that because we already met the contract (scheduled minCount), so we can try the preemption, even when some new pods are schedulable?

I tried to explain in my second paragraph, but let me try to clarify.
We need to take a step back and think what effectively happens at the application level when we change the size of the gang. In those cases we're generally talking about cases, where individual pods within a gang talk to other pods. So changing the size of a gang effectively requires "reconfiguration of the workload at runtime".
Different frameworks solve that differently, but this isn't instant (it may require in-place restarts, it may be solved by distributing some config and reinitializing etc.). But it's not instant.
From the perspective, having multiple cases where we add the next (set of) pod(s) soon after the other is simply not efficient (due to hit on the "goodput" of those workloads).

So once the workload is running, we should rather try to find place for as many remaining pods as possible and them shoot them into the workload in one batch.


> How would that play with the case where the minCount is no longer met (some pods disappeared) and the scheduler has a decision to either schedule up to minCount (and try preemption in subsequent cycle for other pods) or preempt (without scheduling schedulable pods)?

That's a great question. And clearly we can do better in this case.
But that's more of a corner case to me (which will happen of course, but shouldn't be super frequent).
So I would suggest ignoring that for now, getting some production experience with the first simple version of it and react based on users feedback.

> Also, what would we want to do with the basic policy?

I would bet on simplicity at the first step - let's do the same. And in the future we can iterate on that based on the feedback.

macsko: > So once the workload is running, we should rather try to find place for as many remaining pods as possible and them shoot them into the workload in one batch.

Agree, trying preemption is such cases sounds like a reasonable approach. However, this might have a tricky outcome when pods from a single group are constantly coming and triggering additional preemptions. For example, we could have:
- Scheduler already scheduled 5 pods of pod group with minCount=5
- New pods from the pod group are coming
- 1 pod is observed, triggers the unsuccessful scheduling cycle and runs preemption
- 1 more pod is observed, while the previous pod is waiting for victims to terminate
- Scheduling cycle is triggered, 1 pod can be scheduled (as it freed up space via preemption), but another pod cannot. The scheduler tries yet another preemption and moves both pods to scheduling queue.
- 1 more pod is observed ...

Such scenario might significantly delay the binding of these pods.

wojtek-t: Agree - but that sounds a bit like a maliciously constructed problematic scenario :)

I would ignore it for now - let's try to graduate what solves huge majority of problems and then iteratively improve it in follow-ups.

macsko: Alright, if we say that we accept such "malicious" scenarios, I'm fine. Let me update the KEP, so the scheduler will follow a rule that if:

- not every pod has been scheduled AND
- at least one pod was already scheduled at the beginning of the scheduling cycle (i.e., it's not an initial attempt, but any subsequent one)

then we trigger the preemption, even if some pods may be schedulable without it.

macsko: Added, PTAL

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Updated with the links and wrote what tests are remaining, but needed for beta

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Adjusted the algorithm to use v1.37 only

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added a few sentences here describing what changed

macsko: IIUC these `XX/s` numbers mean two-digit values for the number of API calls per second. It's hard to provide some numbers here, as they depend on the exact configurations. Whatever concrete number we put here, it would be more or less random

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Added explanation of both cases

macsko:
@soltysh All your comments should be addressed, PTAL again

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1181,27 +1181,43 @@ The list and configuration of plugins used by this algorithm will be the same as
...
+       [Workload-aware Preemption](#workload-aware-preemption).
+
+       * If the preemption can accommodate more pods, it will be actuated (as described below)
+         and all considered pods, including the schedulable ones, will be moved back to the scheduling queue.
+
+       * If preemption cannot help, schedulable pods will be moved to binding,
+         just as in the initial scheduling attempt case.
+       
+       This aligns with real applications, where attempting preemption to secure capacity for as many additional,
+       remaining pods as possible is preferable to binding only a schedulable subset of pods.
```
wojtek-t: nit:

... if soon after that we will schedule another set unblocked by preemption.

Or something along those lines.

wojtek-t (commented):
One minor nit from me. But overall looks great.

/lgtm

keps/sig-scheduling/4671-gang-scheduling/README.md
macsko: Done

keps/sig-scheduling/4671-gang-scheduling/README.md
soltysh: > I'm not against mentioning things like this, but is that a requirement? Might worth discussing during PRR meeting.

Normally, I'd agree with that "if you're using Alpha features, you do that on your own risk", but somehow I feel a lot of users is actually testing this functionality and that wording in this section hopefully serve as a good reminder for them.

soltysh: On top of that, is the FG merge, which might be a surprise to some, and again I know the alpha risk comment applies 😉

keps/sig-scheduling/4671-gang-scheduling/README.md
soltysh: Not quite upgrade, but it serves its purpose. We should be able to update this flow in the next iteration.

keps/sig-scheduling/4671-gang-scheduling/README.md
soltysh: Gotcha, I wasn't quite sure the meaning behind the `XX/s`, so thx for the explanation, that makes sense.

soltysh (approved):
/approve
the PRR section

keps/sig-scheduling/4671-gang-scheduling/README.md
wojtek-t: OK - the "exposure to community" comment makes sense to me.
I think that many of those people don't read KEPs, but it won't hurt.

And Maciek already added it anyway :)

wojtek-t:
Cancelling hold given LGTMs from both KEP approvers (Kensei and Dominik) above.

/hold cancel

PR #6078, opened 2026-05-14
KEP-6078: MinReadyReplicas for Workload Controllers

kairosci:
- One-line PR description: KEP-6076: MinReadyReplicas for Workload Controllers

- Issue link: https://github.com/kubernetes/kubernetes/issues/134336

- Other comments: Initial KEP draft for adding minReadyReplicas field to Deployment, StatefulSet, and DaemonSet specs. Allows operators to define how many controller replicas must be ready before the controller is considered Available.

### DIFF
--- a/keps/sig-apps/6078-min-ready-replicas/README.md
+++ b/keps/sig-apps/6078-min-ready-replicas/README.md
@@ -1,0 +1,401 @@
+# KEP-6078: MinReadyReplicas for Workload Controllers
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Controller Behavior](#controller-behavior)
+  - [Interaction with maxUnavailable](#interaction-with-maxunavailable)
+  - [Deployment Controller](#deployment-controller)
+  - [StatefulSet Controller](#statefulset-controller)
+  - [DaemonSet Controller](#daemonset-controller)
+  - [HPA Integration](#hpa-integration)
+  - [Test Plan](#test-plan)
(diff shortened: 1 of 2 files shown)


keps/sig-apps/6078-min-ready-replicas/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+## Summary
+
+Add a `minReadyReplicas` field to the spec of Deployment, StatefulSet, and DaemonSet. When set to a positive integer, the controller is considered Available once at least that many replicas are Ready, rather than waiting for all replicas. This gives operators control over the availability signal used by higher-level systems (HPA, network controllers, rollouts).
+
+## Motivation
+
+When pods require long-lived connections or expensive initialization, the first pods to become Ready during a rollout receive a disproportionate share of traffic. This causes uneven load distribution, resource contention, and degraded user experience.
+
+Currently, controllers expose a binary Available signal only when all replicas are Ready. There is no way to signal "enough replicas are ready to serve traffic safely." This forces operators to either accept uneven load or implement complex workarounds (readiness gates with external controllers, custom operators, etc.).
```
aojea: but network traffic is orthogonal to the `Deployment, StatefulSet, DaemonSet specs` , it only uses Pod readiness and EndpintSlices ... I don't fully understand this KEP, @kairosci how much of this is generated and how much is something you put some work to understand it?

kairosci: The KEP is not about changing how Services or EndpointSlices route traffic. The actual scope is about the controllers' Available condition: today it is binary (all replicas Ready or not), and this adds a configurable threshold so consumers like rollback tools or progressive delivery controllers have an intermediate signal to work with. I can explain the reasoning behind any section if it is unclear.

On the implementation side: the changes are in the deployment controller's condition computation, adding conditions to StatefulSet and DaemonSet status, and admission validation and feature gates across the board. Happy to walk through the details if needed.

kairosci: I've just noticed there was some overlap between this version and an earlier draft I had written, which is why there was a bit of confusion. I apologize for that. It should now be complete and contain only the information relevant to this version.

PR #6079, opened 2026-05-14
KEP-5823: Pod-Level Checkpoint/Restore

haircommander:
- One-line PR description: Enable support for Pod-level checkpoint and restore.

- Issue link: https://github.com/kubernetes/enhancements/issues/5823

- Other comments:  Related to https://github.com/kubernetes/enhancements/issues/2008 and https://github.com/kubernetes/enhancements/issues/5091

### DIFF
--- a/keps/sig-node/5823-pod-level-checkpoint-restore/README.md
+++ b/keps/sig-node/5823-pod-level-checkpoint-restore/README.md
@@ -6,11 +6,12 @@
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
-- [Proposal](#proposal)
-  - [Implementation](#implementation)
+  - [Targeted use cases](#targeted-use-cases)
     - [Accelerating startup of applications with long initialization times](#accelerating-startup-of-applications-with-long-initialization-times)
     - [Enabling fault-tolerance for long-running workloads](#enabling-fault-tolerance-for-long-running-workloads)
     - [Pod migration across nodes for load balancing and maintenance](#pod-migration-across-nodes-for-load-balancing-and-maintenance)
+- [Proposal](#proposal)
+  - [Implementation](#implementation)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [CRI API Extensions](#cri-api-extensions)
@@ -193,36 +194,7 @@ follow-on enhancement.
   of being cloned. For example, gVisor provides a special file (`/proc/gvisor/checkpoint`) that
   blocks until a restore is complete, allowing applications to refresh state on resume.
 
-## Proposal
-
-### Implementation
-
-In this proposal, we aim to provide CRI functionality to checkpoint and restore a running Pod, which
-includes all containers running in the Pod, along with Pod-level metadata and configurations. This
-functionality is inspired by [kubelet checkpoint], but extends it to the Pod level, allowing to capture
-and restore the execution state of a Pod, rather than individual containers. The exact implementation
-details of this checkpoint/restore mechanism are left to the container runtime, but we expect the Pod
-checkpoint to capture the complete execution context of all processes running in containers, including
-in-memory state, process hierarchies, open file descriptors, and Pod-level configuration and metadata.
-
-The implementation consists of three layers:
-
-1. **CRI APIs** (`CheckpointPod`, `RestorePod`): Container runtime interface for the actual
-   checkpoint/restore operations, implemented by container runtimes such as containerd and CRI-O.
-
-2. **Kubelet endpoints** (`POST /checkpoint/{ns}/{pod}`, `POST /restore/{ns}/{checkpointName}`):
-   Kubelet HTTP endpoints that translate API requests into CRI calls. The kubelet suspends health
-   checks during checkpointing, resolves the CRI sandbox ID, and manages checkpoint storage at
-   `/var/lib/kubelet/pod-checkpoints/`. For restore, the kubelet reads pod sandbox configuration
-   from the checkpoint, assigns a new Pod UID, updates cgroup parent paths, and delegates to the
-   container runtime.
-
-3. **API objects** in the `checkpoint.k8s.io` API group that provide declarative management of
-   checkpoint operations. `PodCheckpoint` is a namespace-scoped standalone object. Restore is
-   triggered by a new optional `restoreFrom` field on Pod spec rather than a separate object;
-   see [Restore Mechanism](#restore-mechanism). A [pod-snapshot-controller] watches
-   `PodCheckpoint` objects and calls the kubelet checkpoint endpoint through the API server
-   node proxy.
+### Targeted use cases
 
 #### Accelerating startup of applications with long initialization times
 
@@ -255,6 +227,38 @@ state across the move, significantly reducing recovery time compared to full Pod
 Partially served by the alpha scope: checkpoint and create a new Pod on the same node is
 covered; cross-node migration requires a follow-on cross-node transport enhancement, and live
 migration semantics require a follow-on live migration enhancement.
+
+
+## Proposal
+
+### Implementation
+
+In this proposal, we aim to provide CRI functionality to checkpoint and restore a running Pod, which
+includes all containers running in the Pod, along with Pod-level metadata and configurations. This
+functionality is inspired by [kubelet checkpoint], but extends it to the Pod level, allowing to capture
+and restore the execution state of a Pod, rather than individual containers. The exact implementation
+details of this checkpoint/restore mechanism are left to the container runtime, but we expect the Pod
+checkpoint to capture the complete execution context of all processes running in containers, including
+in-memory state, process hierarchies, open file descriptors, and Pod-level configuration and metadata.
+
+The implementation consists of three layers:
+
+1. **CRI APIs** (`CheckpointPod`, `RestorePod`): Container runtime interface for the actual
+   checkpoint/restore operations, implemented by container runtimes such as containerd and CRI-O.
+
+2. **Kubelet endpoints** (`POST /checkpoint/{ns}/{pod}`, `POST /restore/{ns}/{checkpointName}`):
+   Kubelet HTTP endpoints that translate API requests into CRI calls. The kubelet suspends health
+   checks during checkpointing, resolves the CRI sandbox ID, and manages checkpoint storage at
+   `/var/lib/kubelet/pod-checkpoints/`. For restore, the kubelet reads pod sandbox configuration
+   from the checkpoint, assigns a new Pod UID, updates cgroup parent paths, and delegates to the
+   container runtime.
+
+3. **API objects** in the `checkpoint.k8s.io` API group that provide declarative management of
+   checkpoint operations. `PodCheckpoint` is a namespace-scoped standalone object. Restore is
+   triggered by a new optional `restoreFrom` field on Pod spec rather than a separate object;
+   see [Restore Mechanism](#restore-mechanism). A [pod-snapshot-controller] watches
+   `PodCheckpoint` objects and calls the kubelet checkpoint endpoint through the API server
+   node proxy.
 
 ### Risks and Mitigations
 
--- a/keps/sig-node/5823-pod-level-checkpoint-restore/README.md
+++ b/keps/sig-node/5823-pod-level-checkpoint-restore/README.md
@@ -1,0 +1,1250 @@
+# KEP-5823: Pod-Level Checkpoint/Restore
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Implementation](#implementation)
+    - [Accelerating startup of applications with long initialization times](#accelerating-startup-of-applications-with-long-initialization-times)
+    - [Enabling fault-tolerance for long-running workloads](#enabling-fault-tolerance-for-long-running-workloads)
+    - [Pod migration across nodes for load balancing and maintenance](#pod-migration-across-nodes-for-load-balancing-and-maintenance)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [CRI API Extensions](#cri-api-extensions)
+    - [CheckpointPod](#checkpointpod)
+    - [RestorePod](#restorepod)
+  - [Kubelet API Endpoints](#kubelet-api-endpoints)
+    - [Checkpoint Endpoint](#checkpoint-endpoint)
+    - [Restore Endpoint](#restore-endpoint)
+  - [PodCheckpoint Objects](#podcheckpoint-objects)
+    - [PodCheckpoint](#podcheckpoint)
+    - [Pod-Snapshot-Controller](#pod-snapshot-controller)
+  - [Restore Mechanism](#restore-mechanism)
+  - [Checkpoint Content](#checkpoint-content)
+    - [Pod Specification and Metadata](#pod-specification-and-metadata)
+    - [Container Runtime State](#container-runtime-state)
+    - [Shared Pod Resources](#shared-pod-resources)
+  - [Pod Lifecycle](#pod-lifecycle)
+  - [TCP Connection Handling](#tcp-connection-handling)
+  - [Security Implications](#security-implications)
+    - [Privilege model](#privilege-model)
+    - [Sensitive memory contents](#sensitive-memory-contents)
+    - [Denial of service via excessive checkpointing](#denial-of-service-via-excessive-checkpointing)
+    - [automountServiceAccountToken on restore](#automountserviceaccounttoken-on-restore)
+    - [Path traversal protection](#path-traversal-protection)
+    - [Status and spec separation](#status-and-spec-separation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing](#prerequisite-testing)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Rejected Approaches](#rejected-approaches)
+- [Prior Art](#prior-art)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/website]: https://git.k8s.io/website
+[kubelet checkpoint]: https://kubernetes.io/docs/reference/node/kubelet-checkpoint-api
+[OCI image annotations]: https://kubernetes.io/blog/2022/12/05/forensic-container-checkpointing-alpha/#restore-checkpointed-container-k8s
+[criu-coordinator]: https://github.com/checkpoint-restore/criu-coordinator
+[criu-image-streamer]: https://github.com/checkpoint-restore/criu-image-streamer
+[TCP connection repair]: https://lwn.net/Articles/495304/
+[KEP-2008]: https://git.k8s.io/enhancements/keps/sig-node/2008-forensic-container-checkpointing
+[volume snapshots]: https://kubernetes.io/docs/concepts/storage/volume-snapshots/
+[pod-snapshot-controller]: https://github.com/checkpoint-restore/pod-snapshot-controller
+
+## Summary
+
+This proposal defines CRI APIs, kubelet endpoints, and controllers together with  Kubernetes objects
+for managing the lifecycle and artifacts of these operations to enable native support for Pod-level
+checkpoint and restore. The scope of the proposal is limited to warm start and fault-tolerance use cases,
+with outline of the API design to accommodate other use cases such as suspend/resume (with IP preservation)
+and live migration (streaming checkpoint data between nodes). These use cases will be addressed in future KEPs.
+
+The core idea is to outline the minimal set of Container Runtime Interface (CRI) and kubelet
+extensions required for Pod-level checkpoint and restore, and to provide a clear path for
+iteratively building on top of these APIs to address the broader set of use cases and requirements.
+
+In this KEP, checkpoints represent the runtime state of a Pod, where the checkpoint format and low-level
+implementation details are left to the container runtime (e.g., containerd, CRI-O), the OCI runtime (runc, crun),
+and the underlying checkpoint/restore mechanism (e.g., CRIU, gVisor).
+
+While Pod-level checkpointing is inspired by the existing [kubelet checkpoint] API and extends that
+container checkpointing mechanism to Pods, the restore functionality is a larger addition as Kubernetes
+currently supports container restore only via [OCI image annotations].
+
+This proposal defines Pod-level checkpoint and restore as a single, cohesive feature as checkpointing
+without restore would be incomplete and impractical for the use cases motivating this work.
+
+## Motivation
+
+The existing [kubelet checkpoint] API was originally inspired by the checkpoint/restore functionality
+of container engines such as Podman. However, unlike these container engines, Kubernetes is responsible
+for managing, scaling, and coordinating workloads across an entire cluster of machines. As a result,
+container-level checkpointing alone does not adequately support many Kubernetes-native workflows and
+higher-level operations that require preserving and restoring the full Pod state. This KEP aims to
+remove this barrier by enabling a Pod-level checkpoint and restore mechanisms that is aligned with
+the core Kubernetes abstractions.
+
+### Goals
+
+- Introduce Pod-level checkpoint and restore support to the CRI API (`CheckpointPod`, `RestorePod`).
+- Extend the kubelet API with checkpoint and restore endpoints for Pod runtime state.
+- Define `PodCheckpoint` object and the Pod-level restore operation.
+
+### Non-Goals
+
+The following items are out of scope for this KEP. Each is expected to be addressed in a
+follow-on enhancement.
+
+- Pod live migration with low latency or SLO guarantees. This requires streaming checkpoint data
+  directly between nodes (without intermediate storage) and IP-address preservation for
+  established TCP connections across nodes. This is partially addressed today by [criu-image-streamer]
+  and [TCP connection repair], but once Pods are scheduled they are bound to a specific node
+  (`pod.Spec.NodeName`), and Kubernetes does not currently guarantee network identity preservation
+  across restores.
+
+- In-place restore (same Pod UID, same Pod object). The initial implementation creates a new Pod
+  from a checkpoint. Restoring into the same Pod object requires modifying Pod lifecycle
+  semantics, which has deep ecosystem implications for controllers, schedulers, and monitoring
+  tools.
+
+- Cross-node restore. The initial implementation focuses on same-node restore. Cross-node restore
+  requires a checkpoint transport mechanism.
+
+- Checkpoint and restore of shared Pod resources such as shared memory, volumes, and Dynamic
+  Resource Allocation (DRA) devices.
+
+- Scheduling integration (workload-aware preemption with checkpoint awareness, eviction request
+  interceptors).
+
+- Distributed or multi-Pod coordinated checkpointing (e.g., synchronized checkpoint of a
+  distributed training job). Requires external coordination tools such as [criu-coordinator].
+
+- Handling of exec sessions and port-forward. Support for preserving and restoring exec sessions
+  and port-forward can be explored in a future enhancement proposal.
+
+- Checkpoint portability across heterogeneous environments such as different CPU and GPU
+  architectures, kernel versions, container runtimes, or device drivers.
+
+- Checkpoint lifecycle management including resource quotas, limits, retention policies, and
+  garbage collection of checkpoint data.
+
+- Application-triggered checkpointing. When creating multiple clones from the same checkpoint,
+  the workload may need to refresh state such as session keys, random number generator states,
+  and certificates. A future KEP will explore a common mechanism for applications to be notified
+  of being cloned. For example, gVisor provides a special file (`/proc/gvisor/checkpoint`) that
+  blocks until a restore is complete, allowing applications to refresh state on resume.
+
+## Proposal
+
+### Implementation
+
+In this proposal, we aim to provide CRI functionality to checkpoint and restore a running Pod, which
+includes all containers running in the Pod, along with Pod-level metadata and configurations. This
+functionality is inspired by [kubelet checkpoint], but extends it to the Pod level, allowing to capture
+and restore the execution state of a Pod, rather than individual containers. The exact implementation
+details of this checkpoint/restore mechanism are left to the container runtime, but we expect the Pod
+checkpoint to capture the complete execution context of all processes running in containers, including
+in-memory state, process hierarchies, open file descriptors, and Pod-level configuration and metadata.
+
+The implementation consists of three layers:
+
+1. **CRI APIs** (`CheckpointPod`, `RestorePod`): Container runtime interface for the actual
+   checkpoint/restore operations, implemented by container runtimes such as containerd and CRI-O.
+
+2. **Kubelet endpoints** (`POST /checkpoint/{ns}/{pod}`, `POST /restore/{ns}/{checkpointName}`):
+   Kubelet HTTP endpoints that translate API requests into CRI calls. The kubelet suspends health
+   checks during checkpointing, resolves the CRI sandbox ID, and manages checkpoint storage at
+   `/var/lib/kubelet/pod-checkpoints/`. For restore, the kubelet reads pod sandbox configuration
+   from the checkpoint, assigns a new Pod UID, updates cgroup parent paths, and delegates to the
+   container runtime.
+
+3. **API objects** in the `checkpoint.k8s.io` API group that provide declarative management of
+   checkpoint operations. `PodCheckpoint` is a namespace-scoped standalone object. Restore is
+   triggered by a new optional `restoreFrom` field on Pod spec rather than a separate object;
+   see [Restore Mechanism](#restore-mechanism). A [pod-snapshot-controller] watches
+   `PodCheckpoint` objects and calls the kubelet checkpoint endpoint through the API server
+   node proxy.
+
+#### Accelerating startup of applications with long initialization times
+
+This is the primary driver of the alpha scope. The cold start time of many applications, such
+as LLM inference services and Java applications, can reach several minutes due to complex
+initialization steps that must complete before the service can accept requests or process data.
+Pod checkpointing allows the initialized state of a running application to be saved to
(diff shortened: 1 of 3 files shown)


keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1250 @@
...
+   checks during checkpointing, resolves the CRI sandbox ID, and manages checkpoint storage at
+   `/var/lib/kubelet/pod-checkpoints/`. For restore, the kubelet reads pod sandbox configuration
+   from the checkpoint, assigns a new Pod UID, updates cgroup parent paths, and delegates to the
+   container runtime.
+
+3. **API objects** in the `checkpoint.k8s.io` API group that provide declarative management of
+   checkpoint operations. `PodCheckpoint` is a namespace-scoped standalone object. Restore is
+   triggered by a new optional `restoreFrom` field on Pod spec rather than a separate object;
+   see [Restore Mechanism](#restore-mechanism). A [pod-snapshot-controller] watches
+   `PodCheckpoint` objects and calls the kubelet checkpoint endpoint through the API server
```
haircommander: TODO: discuss mechanism

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1250 @@
...
+  artifacts as sensitive data with the same handling as Secrets; (d) provide pre-defined
+  viewer/editor/admin ClusterRoles for per-namespace binding. See
+  [Security Implications](#security-implications).
+
+- Application awareness is required. Checkpoint and restore are not transparent to applications:
+  in-memory secrets, tokens, environment variables, and cached hostnames persist through restore,
+  and selective memory scrubbing is not feasible. Applications must cooperate for correctness.
+  See [Application Awareness and Restore Contract](#application-awareness-and-restore-contract).
+
+- GPU memory overhead. Checkpointing workloads that run on NVIDIA GPUs typically requires the GPU
```
haircommander, line 285: true but out of scope for this KEP, probably should break these into two sections: risks for this KEP, and future risks

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  of scope for this KEP and must be handled by external tools such as [criu-coordinator] or by
+  application-level synchronization.
+
+- Temporary unavailability during checkpoint. During the checkpointing window, the containers in
+  the checkpointed Pod are frozen to create a consistent checkpoint. The duration of this window
+  varies with the workload (for example, amount of memory at the time of checkpointing) and the
+  underlying checkpoint mechanism, leading to temporary unavailability. The checkpointing state
+  must be exposed by the container runtime via the Pod or Container Status API so clients can
+  detect it. During this window, Ephemeral Containers may not be started for the checkpointed
+  Pod, and any existing Ephemeral Containers are frozen along with the regular containers.
```
haircommander, line 313: hm did we decide to allow ephemeral containers in the initial checkpoint impl?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+    // ID of the restored pod sandbox
+    string pod_sandbox_id = 1;
+}
+```
+
+For security reasons, certain fields in the Pod spec (e.g., namespace) are blocked from changing during
+the restore. In particular, the Pod UID must remain automatically generated by Kubernetes and must not
+be user-specifiable. While this KEP does not include the complete set of restricted fields, we expect
+the exact fields and validation rules to be fully specified as part of the implementation enforcing
+strict validation to Pod spec. A future KEP might also explore utilizing RBAC permissions to verify
```
haircommander, line 408: we potentially need to do this in this kep, depending on how we do restore

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+that the requesting user has access to both the source and destination namespaces.
+
+In the event of a timeout, the container runtime should return an error indicating that the restore
+operation did not complete within the specified time limit and ensuring that any partially restored
+artifacts are cleaned up. The kubelet should handle this error appropriately, for example by
+returning a timeout error to the caller of the `RestorePod` API.
```
haircommander, line 414: I don't love the duplication here. maybe we reference the section above but say the same applies for restore

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+```
+POST /checkpoint/{podNamespace}/{podName}[?timeout={seconds}]
+```
+
+The checkpoint endpoint:
+1. Validates the target Pod exists and is in the `Running` phase.
+2. Resolves the CRI sandbox ID from the Pod's runtime status.
+3. Generates a checkpoint path: `/var/lib/kubelet/pod-checkpoints/checkpoint-{podName}_{namespace}-{timestamp}`.
+4. Calls the `CheckpointPod` CRI API with the sandbox ID and checkpoint path.
+5. Returns the checkpoint path in the response:
```
haircommander, line 432: I think this is a little different than what I was imagining. I guess I was thinking because we have the restoreFrom field, the checkpoint controller could watch pods coming in and if there's restoreFrom with a reference to an existing  checkpoint, the controller could edit the pod spec to make the link, and pass that back to the apiserver. apiserver would then do a normal pod creation/update, but the kubelet would read the restoreFrom field and do a restore rather than normal create. i.e: do we want a restore api in the kubelet?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+status to reflect the result.
+
+```yaml
+apiVersion: checkpoint.k8s.io/v1alpha1
+kind: PodCheckpoint
+metadata:
+  name: my-checkpoint
+spec:
+  # Name of the running Pod to checkpoint.
+  sourcePodName: my-app
```
haircommander, line 481: I think we'll need to think through this. can we be fancier? like if there's a pod template or a replicaset, can we find a way to reference the latest pod or the node local pod without finding the exact name? not sure

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+#### Pod-Snapshot-Controller
+
+The [pod-snapshot-controller] manages the lifecycle of `PodCheckpoint` objects. It
+communicates with the kubelet through the API server node proxy
+(`/api/v1/nodes/{node}/proxy/checkpoint/...`).
+
+The checkpoint reconciliation flow:
+
+1. Watch for new `PodCheckpoint` objects.
+2. Validate that the target Pod exists, is `Running`, and has all init containers completed.
```
haircommander, line 510: should the controller do this or the kubelet? kubelet would close TOCTOU race a bit

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+3. Set phase to `InProgress` and record the node name.
+4. Call `POST /api/v1/nodes/{node}/proxy/checkpoint/{namespace}/{pod}` with optional timeout.
+5. On success, set phase to `Ready` and store the checkpoint location and container info.
+6. On failure, set phase to `Failed` with error details in conditions.
+
+Restore does not require controller involvement: the kubelet drives restore directly from the
+Pod spec (see [Restore Mechanism](#restore-mechanism)).
```
haircommander, line 517: I'm not positive about this. for instance: who is responsible for saying when a user has access to a snapshot? is that apiserver, or is it controller?

haircommander: also: if the restore is targetting a different node than the checkpoint is on, now that will fail. in the future: don't we imagine the checkpoint controller makign the transfer?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+A user creates a Pod with `spec.restoreFrom` set to a checkpoint reference; the kubelet
+observes this during `SyncPod` and calls `restorePodSandbox()` instead of `createPodSandbox()`.
+Pod creation is the restore, in a single step.
+
+This shape collapses the restore flow into the normal Pod admission and scheduling path: the
+scheduler places the Pod (subject to the node-affinity constraints needed to land on a node
+that has the checkpoint data), the CNI plugin sets up networking against the Pod object
+exactly as it would for a fresh Pod, and the kubelet swaps the sandbox creation step for
+sandbox restore. No placeholder Pod, no separate object lifecycle, and no `nodes/proxy`
+permission for restore.
```
haircommander, line 531: this contradicts above where we say we want a restore api in the kubelet (which I disagree with)

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+### Checkpoint Content
+
+Different container and OCI runtimes may implement the Pod checkpointing mechanism differently, and the
+exact content of a checkpoint may vary. However, we expect that a Pod checkpoint will capture the
```
haircommander, line 536: if this is an expectation we should find a implementation agnostic way to test it. if we can't, then maybe it's more of a suggestion. I can imagine a world where we want a lossful checkpoint implementation that maybe doesn't capture everything but does many things. for instance: pytorch checkpointing really checkpoints very little of the process, but it does create a recreatable state. maybe the heuristic is we can do a roundtrip check (checkpoint, restore) and then have a suite of tests to make sure the data we expect to be retained was

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+Note that the `RestorePod` API allows users to optionally override fields such as resource requests and
+limits. During restore, the process tree inside containers is recreated from the application state
+captured during checkpointing. This operation requires, for example, open file descriptors and memory
+allocations to be recreated with the same offsets and contents as at the time of checkpointing, in order
+to ensure correct application behavior.
+
+For Pods using Dynamic Resource Allocation, resource claims are expected to be resolved during restore,
+and the checkpoint should capture the necessary information to allow the runtime to reattach to the same
+resource claims.
```
haircommander, line 575: out of scope here

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+- Pod UID, namespace, labels, annotations, and owner references
+- Resource requests and limits
+- Scheduling constraints and security contexts
+- Status of all containers, including containers that have completed execution
+
+This proposal considers checkpointing while init containers are running out for
+the initial implementation, and the handling of init containers may be further
+explored in future iterations.
+
+Note that the `RestorePod` API allows users to optionally override fields such as resource requests and
```
haircommander, line 567: maybe this should not be allowed to start. a user can always in-place resize after the restore if they want, or before. but i think it'll simplify things if we disallow it

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+follow-on enhancement.
+
+- Pod live migration with low latency or SLO guarantees. This requires streaming checkpoint data
+  directly between nodes (without intermediate storage) and IP-address preservation for
+  established TCP connections across nodes. This is partially addressed today by [criu-image-streamer]
+  and [TCP connection repair], but once Pods are scheduled they are bound to a specific node
+  (`pod.Spec.NodeName`), and Kubernetes does not currently guarantee network identity preservation
+  across restores.
+
+- In-place restore (same Pod UID, same Pod object). The initial implementation creates a new Pod
```
greenblade29, line 165: I understand this is in the `Non Goals` section, but I am struggling to understand if this needs to be brought up as a follow up enhancement as well. What would be the use cases where we need to restore with the same Pod ID? Kubernetes by design is supposed to be ephemeral: and I believe as long as we are able to restore a process from a state, pod IDs should not matter.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+#### Privilege model
+
+The existing container-level checkpoint API requires node administrator or SSH access to reach
+the kubelet endpoint. Exposing Pod-level checkpoint and restore through namespaced API objects
+is a different security model. Mitigations:
+
+- `PodCheckpoint` is namespace-scoped and may only target Pods in the same namespace; the
+  controller enforces same-namespace lookups. Restore is triggered by `spec.restoreFrom` on a
+  Pod and so is naturally scoped to the namespace in which the Pod is created. Cross-namespace
+  restore is not possible by construction.
```
haircommander, line 630: in the warm start case this is not immediately true but we can make it so

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+- The `nodes/proxy` permission permits arbitrary kubelet API calls and is granted only to the
+  pod-snapshot-controller service account. It is never granted to end-user roles. Restore
+  does not require `nodes/proxy` because it flows through the normal Pod admission path.
+- The controller exposes no user-facing API beyond the `PodCheckpoint` resource; all kubelet
+  interaction for checkpointing is mediated by the controller.
+- Pre-defined namespaced ClusterRoles (viewer, editor, admin) are provided so administrators
+  can bind checkpoint and restore access per namespace with `RoleBinding`.
+- `sourcePodName` on `PodCheckpoint` and `spec.restoreFrom` on Pod are immutable after
+  creation, preventing post-creation namespace-escape attempts.
```
haircommander, line 639: this does mean on the live migration case you only get one migration. that may be okay to start but something to think about

haircommander: the other option being in the restore case we can block updates to the spec where the status says restore is in progress. but after the restore is complete, we could allow updates to the restoreFrom field again

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+Service account tokens mounted into the original Pod may be invalid or expired when a
+checkpoint is restored. Checkpointable workloads should disable token automounting and refresh
+tokens explicitly after restore; a formal opt-out or automatic token refresh mechanism will be
+specified before Beta.
+
+#### Path traversal protection
+
+The kubelet restore endpoint validates checkpoint names, rejecting paths containing `/` or
+`..`, and verifies that the resolved checkpoint path remains within the checkpoint storage
```
haircommander, line 664: the other option being users can only reference a snapshot by name, and then the controller uses pod status to say where the checkpoint is

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  exists and the archive is non-empty.
+- `RestorePod` happy path: restore that Pod; verify a new sandbox ID is returned.
+- Probe suspension: a Pod with a 1 second liveness probe is not killed during a multi-second
+  checkpoint window.
+- Runtime does not implement the new RPC: the kubelet returns a typed error rather than a
+  panic or generic 500.
+- Feature gate disabled: the checkpoint endpoint returns 405 and `spec.restoreFrom` is
+  rejected at Pod admission.
+- `spec.restoreFrom` happy path: the kubelet sees the field during `SyncPod`, calls
+  `restorePodSandbox()`, and the Pod transitions to `Running`.
```
haircommander, line 720: we also could write cri proxy tests to spoof checkpointing/restoring

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1250 @@
...
+
+#### Enabling fault-tolerance for long-running workloads
+
+Training jobs for large AI models run on hundreds or thousands of GPUs and often execute for
+weeks or months. Hardware and system failures are inevitable and can force jobs to restart from
+scratch, resulting in significant loss of time and computational resources. Pod-level
+checkpointing allows the runtime state of these workloads to be captured and restored on
+failure. For example, when a training job is preempted by a batch scheduler, Pod-level
+checkpoint/restore can capture and later resume the runtime state to avoid restarting the
+training job. Partially served by the alpha scope: single-Pod checkpoint/restore is covered;
```
greenblade29, line 245: For single pod checkpoint/restore in these cases, do we plan to make use of something like [CRIUgpu](https://arxiv.org/html/2502.16631v1)?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+- **Container image name override for restore.** Replacing the container image name with a checkpoint
+  archive path to trigger restore (as used in the existing forensic checkpointing feature) was rejected
+  because it does not work for Pod-level restore (what image name to use for a multi-container Pod?)
+  and creates confusing Pod generation semantics.
+
+- **Parent cgroup freezer for atomic Pod freeze.** Using the parent cgroup freezer to freeze an entire
+  Pod atomically was rejected because CRIU is not aware of the parent cgroup freezer. CRIU needs to
+  unfreeze individual containers for parasite code injection, and processes are frozen one-by-one
+  internally. Per-container cgroup freezing is simpler and works correctly with CRIU.
```
haircommander, line 1201: this is an implementation detail not discussed above, I'd suggest removing

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  checkpoint and restore through namespaced API objects is a different security model: it lets
+  regular users trigger an operation that captures full process memory, including secrets.
+  Mitigations: (a) scope checkpoint resources as namespace-scoped; (b) keep the node-proxy
+  privilege on the controller service account and never grant it to users; (c) treat checkpoint
+  artifacts as sensitive data with the same handling as Secrets; (d) provide pre-defined
+  viewer/editor/admin ClusterRoles for per-namespace binding. See
+  [Security Implications](#security-implications).
+
+- Application awareness is required. Checkpoint and restore are not transparent to applications:
+  in-memory secrets, tokens, environment variables, and cached hostnames persist through restore,
```
greenblade29, line 285: Will the pods that have these values referenced as environment variables read stale value if the Secret objects change between checkpoint and restore?

haircommander:
oh darn, I meant for this to be opened against https://github.com/rst0git/enhancements/pulls. oh well

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  surrogate spec fields to satisfy CNI plugins, requires `nodes/proxy` on the restore path,
+  and introduces a second status state machine that must be kept in sync with the restored
+  Pod's own status. The `spec.restoreFrom` field on Pod spec collapses restore into the
+  normal Pod create flow: the scheduler, CNI plugins, controllers, and observability tooling
+  all see a single Pod object with its standard lifecycle, and the only change is that the
+  kubelet swaps `createPodSandbox()` for `restorePodSandbox()` when `spec.restoreFrom` is
+  set. The trade-off is a small Pod spec addition, which is justified by the simplification
+  on every other axis.
+
+- **`PodCheckpoint` as a CRD.** Shipping `PodCheckpoint` as a CRD (in an out-of-tree
```
kannon92, line 1221: I still am unclear on why this should be a core API. 

out-of-tree have built-in validation, conversion and defaulting machinery also. Most projects implement this. 

The downsides of putting this in-tree is the velocity of development for this API will be a lot slower. To me, the big question is how comfortable are we have the API of PodCheckpoint. Is this API feature complete and can proceed to GA?

Or will we expect a lot of iteration on this API. If so I think we should still consider putting PodCheckpoint into a kubernets-sig project. We also could implement things like GC of checkpoints.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+without restore would be incomplete and impractical for the use cases motivating this work.
+
+## Motivation
+
+The existing [kubelet checkpoint] API was originally inspired by the checkpoint/restore functionality
+of container engines such as Podman. However, unlike these container engines, Kubernetes is responsible
+for managing, scaling, and coordinating workloads across an entire cluster of machines. As a result,
+container-level checkpointing alone does not adequately support many Kubernetes-native workflows and
+higher-level operations that require preserving and restoring the full Pod state. This KEP aims to
+remove this barrier by enabling a Pod-level checkpoint and restore mechanisms that is aligned with
```
kannon92, line 144:
```suggestion
remove this barrier by enabling a Pod-level checkpoint and restore mechanism that is aligned with
```

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  Mitigations: (a) scope checkpoint resources as namespace-scoped; (b) keep the node-proxy
+  privilege on the controller service account and never grant it to users; (c) treat checkpoint
+  artifacts as sensitive data with the same handling as Secrets; (d) provide pre-defined
+  viewer/editor/admin ClusterRoles for per-namespace binding. See
+  [Security Implications](#security-implications).
+
+- Application awareness is required. Checkpoint and restore are not transparent to applications:
+  in-memory secrets, tokens, environment variables, and cached hostnames persist through restore,
+  and selective memory scrubbing is not feasible. Applications must cooperate for correctness.
+  See [Application Awareness and Restore Contract](#application-awareness-and-restore-contract).
```
kannon92, line 287: This doesn't point to anything.

keps/sig-node/5823-pod-level-checkpoint-restore/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: KubeletLocalPodCheckpointRestore
+    components:
+      - kubelet
+disable-supported: true
+
+# The following PRR answers are required at beta release
+metrics:
+  - "N/A"
```
kannon92, line 40: I don't think this is correct. I think you list new metrics you are adding in the KEP.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
greenblade29: > the checkpoint controller could watch pods coming in and if there's restoreFrom with a reference to an existing checkpoint

As discussed in the call: we need to think of tradeoffs between-

1. Adding a dedicated label (for performant listing of pods that are in the category of being restored) **and** `restoreFrom` field (for having the data on restoration).
2. Only keeping `restoreFrom` field and then listing down all pods, followed by filtering on the ones that have it.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+    ...
+    // RestorePod restores a pod sandbox from a checkpoint
+    rpc RestorePod(RestorePodRequest) returns (RestorePodResponse) {}
+    ...
+}
+
+message RestorePodRequest {
+    // Location path where the checkpoint will be restored from.
+    string path = 1;
+    // Optional pod sandbox configuration to override checkpoint metadata.
```
greenblade29, line 381: As we discussed yesterday: for the scope of this KEP, we can think of config being the same. Anything that is allowed to change on the Pod (such as resource requests) can be later done in a separate operation. This KEP will only focus on restoring the pod as is, and modification of pod sandbox can be separate.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+For Pods using Dynamic Resource Allocation, resource claims are expected to be resolved during restore,
+and the checkpoint should capture the necessary information to allow the runtime to reattach to the same
+resource claims.
+
+#### Container Runtime State
+
+The container runtime state includes OCI container configurations, security contexts, the writable layer
+of the container filesystem, and checkpoint images representing the state of the Pod that capture the
+complete execution context of all processes and threads running in containers necessary to recreate them,
```
greenblade29, line 581: We should also clarify whether we will be giving users the option to keep their pods running or kill it (in a completed state). Also: since this KEP talks about restore on the same node, it is possible that the original pod moves out of the node, so expectations should be clarified that when we are saying restoration happens on the same node: the node being referred to is the one that will have the checkpoint taken on rather than the pod's current node.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
haircommander: it follows from a logical conclusion beginning at the fact the kubelet needs to be aware of checkpointing and restoring. If we checkpoint a pod, that means the processes will pause, which will interrupt probes. When we restore, we need a way to express that the pod is being created from the snapshot and not the majority of the spec fields.  These things can't be done with a CRD. I think this initial implementation is useful for warm start cases or pause/unpause (if the IP address doesn't matter and the process is stateless)

keps/sig-node/5823-pod-level-checkpoint-restore/kep.yaml
dfeigin-nv: The README has several metrics (kubelet_pod_checkpoint_operations_total, kubelet_pod_checkpoint_duration_seconds, kubelet_pod_restore_operations_total, kubelet_pod_restore_duration_seconds, kubelet_pod_checkpoint_size_bytes, podcheckpoint_phase_total and podcheckpoint_reconcile_duration_seconds). @haircommander maybe we should paste them here too?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
dfeigin-nv: Yes - env vars are read into process memory at process start, captured in the checkpoint, and persist through restore. We can point that out in the doc:
1. Short-lived credentials (SA tokens, registry tokens) can expire between checkpoint and restore.
2. If a Secret's value changes between checkpoint and restore, the env-injected copy is invisibly stale.
It's part of why applications must cooperate for correctness.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+### Kubelet API Endpoints
+
+The kubelet exposes HTTP endpoints for Pod-level checkpoint and restore, gated behind the
+`KubeletLocalPodCheckpointRestore` feature gate.
+
+#### Checkpoint Endpoint
+
+```
+POST /checkpoint/{podNamespace}/{podName}[?timeout={seconds}]
```
dfeigin-nv, line 424: I couldn't understand if `restoreFrom` is the canonical path for restoring or the Kubelet HTTP endpoint. Do they meet somewhere or use the same building block?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
dfeigin-nv: Good question, but I think that GPU state is complex enough that we probably shouldn't tie ourselves to a specific tool like `CRIUgpu ` right out of the gate. Since device checkpointing is a non-goal for Alpha anyway, we can figure out the GPU tooling later and discuss it separately

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  # Path to checkpoint data on the node.
+  checkpointLocation: /var/lib/kubelet/pod-checkpoints/checkpoint-my-app_default-2026-03-10T20:38:11Z
+  # List of checkpointed containers.
+  containers:
+  - name: main
+    image: my-app:latest
+  conditions:
+  - type: Ready
+    status: "True"
+    reason: CheckpointCompleted
```
atiratree, line 498: The API needs polishing:
- Should there be rather conditions instead of a phase? - from API design PoV
- should we also use UIDs for nodes? What if the node goes missing, but a new one is created with the same name?
- checkpointLocation should be a struct  to allow for additional storage types.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: If you find more than one pod it will be quite painful to handle.
- pod selection - rules are needed to select only a single pod
- if more than one pod can be selected, the status and lifecycle of the PodCheckpoint would be quite difficult to reconcile.

Would it make sense to include the UID here as well? To exactly match a pod?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+```
+
+#### Pod-Snapshot-Controller
+
+The [pod-snapshot-controller] manages the lifecycle of `PodCheckpoint` objects. It
+communicates with the kubelet through the API server node proxy
+(`/api/v1/nodes/{node}/proxy/checkpoint/...`).
```
atiratree, line 505: Why is the subresource on the node? Shouldn't we either make this a subresource of a PodCheckpoint or a Pod like `pods/exec`.

This has implications for RBAC. Node is not namespaced and pods are. It might make sense to bundle it with pod as the management of privileges is collocated. 

Not sure how it is with versioning subresources though...

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: Having early validation in the controller doesn't have any disadvantages IMO. 

I think having a UID of the pod should help here. Especially if you consider that the controller can stop running for some time - for example during a leader change.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+In the event of a timeout, the container runtime should return an error indicating that the restore
+operation did not complete within the specified time limit and ensuring that any partially restored
+artifacts are cleaned up. The kubelet should handle this error appropriately, for example by
+returning a timeout error to the caller of the `RestorePod` API.
+
+### Kubelet API Endpoints
+
+The kubelet exposes HTTP endpoints for Pod-level checkpoint and restore, gated behind the
+`KubeletLocalPodCheckpointRestore` feature gate.
```
atiratree, line 419: It would be good to mention the concurrency model for the endpoints here. I.e. it seems only a single controller can issue or restore checkpoints for a pod at the same time. I wonder if it makes sense to invest in some kind of a lock or async lock?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+The [pod-snapshot-controller] manages the lifecycle of `PodCheckpoint` objects. It
+communicates with the kubelet through the API server node proxy
+(`/api/v1/nodes/{node}/proxy/checkpoint/...`).
+
+The checkpoint reconciliation flow:
+
+1. Watch for new `PodCheckpoint` objects.
+2. Validate that the target Pod exists, is `Running`, and has all init containers completed.
+3. Set phase to `InProgress` and record the node name.
```
atiratree, line 511: It is better to use ObservedGeneration to indicate a progress and a conditions for additional info/state.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+4. Call `POST /api/v1/nodes/{node}/proxy/checkpoint/{namespace}/{pod}` with optional timeout.
+5. On success, set phase to `Ready` and store the checkpoint location and container info.
+6. On failure, set phase to `Failed` with error details in conditions.
```
atiratree, line 514: Is the controller going to hold the connection open until it completes? I expect that the timeouts can be on the order of minutes or even more. This seems like a major bottle neck for the controller.

Also what happens when you need to restart the controller in the middle of the call?

I think it would be preferable to also explore an async model.

Also, what happens when the kubelet restarts during the operation?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: I agree. The pod spec field will limits the usage and evolution of the restore.
- It is not possible to apply RBAC to pod fields.
- There might be different storage locations to choose from - even in the different ns or outside of the cluster. This might need to be resolved by the controller.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+distinct from the full `v1.PodSpec` defined in the API server. This KEP aims to describe an
+initial high-level approach to Pod-level checkpointing and leave the implementation details
+to each container runtime.
+
+The checkpoint includes:
+- Serialized Pod specification
+- Pod UID, namespace, labels, annotations, and owner references
+- Resource requests and limits
+- Scheduling constraints and security contexts
+- Status of all containers, including containers that have completed execution
```
atiratree, line 561: Finalizers should be at least discussed here IMO.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+Restore does not require controller involvement: the kubelet drives restore directly from the
+Pod spec (see [Restore Mechanism](#restore-mechanism)).
+
+### Restore Mechanism
+
+Restore is triggered by a new optional field on Pod spec rather than by a separate API object.
+A user creates a Pod with `spec.restoreFrom` set to a checkpoint reference; the kubelet
+observes this during `SyncPod` and calls `restorePodSandbox()` instead of `createPodSandbox()`.
+Pod creation is the restore, in a single step.
```
atiratree, line 524: So what happens to pods with init containers here? Do we run them first? What if they differ from the pod restore snapshot? Is this allowed/validated?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+runtime. Init container semantics and partial-ready states are out of scope for this KEP.
+
+During checkpointing, all containers in the Pod are frozen (using the Pod-level cgroup freezer)
+as a prerequisite for creating a consistent checkpoint. Each container is then checkpointed
+individually, and the cgroup is unfrozen at the end of this operation.
+
+The kubelet must suspend liveness and readiness probes while a Pod is being checkpointed. Frozen
+cgroups may cause probes to time out, and without suspension the kubelet would kill the Pod
+mid-checkpoint. A Pod status condition (`Checkpointing=True`) is set so that higher-level
+controllers can observe this state.
```
atiratree, line 604: This is very problematic. Pod which is not ready is prone to be disrupted. For example StatefulSets and PDBs consider the pod to be unhealthy and allow the pod to be killed or even actively delete it in some scenarios. I expect other controllers might behave in a similar manner.

On the other hand, masking this and marking the pod Ready during checkpoints will result in the network traffic being sent there.

The condition could solve this but an adoption will be slow.

We also have similar issue https://github.com/kubernetes/kubernetes/issues/116965 and I think we need to find a common approach with right granularity.

EvictionRequest API might need more signal in the future as well during termination.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+  `restoreFrom` field on Pod spec.
+- Pod-snapshot-controller implemented.
+- End-to-end warm start workflow: checkpoint a running Pod, create a new Pod from that
+  checkpoint on the same node. Demonstrated against at least one CRI implementation.
+- e2e tests described in the [Test Plan](#test-plan) pass in CI on supported runtimes and skip
+  cleanly on unsupported runtimes.
+- Pre-defined viewer, editor, and admin ClusterRoles published for the namespaced resources.
+- Alpha-level PRR answered.
+
+#### Beta
```
atiratree, line 768: I feel like the checkpoint retention is important here as well.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+
+- **Kubelet-only scope (no API server changes).** Keeping the KEP scope to kubelet-level changes only
+  was rejected because a restored Pod with no API server representation is not useful. Even if alpha
+  does not fully implement API server awareness, the KEP must describe the path to a useful end-to-end
+  feature.
+
+- **Separate `PodRestore` API object.** A standalone `PodRestore` resource that references a
+  `PodCheckpoint` and is reconciled by a controller (which then creates a placeholder Pod and
+  calls the kubelet restore endpoint through `nodes/proxy`) was considered and rejected. The
+  separate-object shape duplicates the Pod lifecycle, requires a placeholder Pod with
```
atiratree, line 1211: Does it have to be complicated? A simple PodRestore API object with a checkpoint/snapshot location/reference should be enough. The Pod itself could have a reference to this PodRestore. The goal is to have the flexibility for the feature evolution, not to make it handle everything from the start.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
```diff
@@ -0,0 +1,1254 @@
...
+5. On success, set phase to `Ready` and store the checkpoint location and container info.
+6. On failure, set phase to `Failed` with error details in conditions.
+
+Restore does not require controller involvement: the kubelet drives restore directly from the
+Pod spec (see [Restore Mechanism](#restore-mechanism)).
+
+### Restore Mechanism
+
+Restore is triggered by a new optional field on Pod spec rather than by a separate API object.
+A user creates a Pod with `spec.restoreFrom` set to a checkpoint reference; the kubelet
```
atiratree, line 522: This expects that the user manages the Pod YAMLs for each checkpoint. I guess this can work for simple examples, but I suppose this can drift considerably.

Having a controller to validate that or fully construct the pod object would help here. Btw for example we track revisions with ControllerRevision objects for a StatefulSet. We don't let users to manage the revisions manually.

Also what guarantees what application can be in what checkpoint location. This goes against the workload and revisions management model. How are the workloads expected to handle and navigate that?

Does the image in the spec mean anything?

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: Yes. I think the pod field might become limiting when implementing more granular RBAC permissions

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: > , the controller could edit the pod spec to make the link, and pass that back to the apiserver. apiserver would then do a normal pod creation/update

This would require an admission plugin. This is something we would like to avoid if we can. 

other options:

1. Make the restoreFrom mutable and change it later, or update the pod status.
3. Can we make the kubelet list the checkpoints itself? And find the correct path.
4. I think, it would also be good to consider a separate Restore API and the controller would translate it  into a pod.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: It would be useful to have full apidoc for `spec.restoreFrom` in the KEP.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
haircommander: for 3, a separate restore api would still lead to pod spec changes. at some point that translation has to happen, so we could either hide it behind the sugar of a restore api, or make it explicit in the pod spec. I think the best way forward is having the kubelet listing the checkpoints itself in the restore path. in the future, we may need to create a api call to move checkpoints between nodes, but I think that's fine for now.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
haircommander: hm how would this resolution work in the case of https://github.com/kubernetes/enhancements/pull/6079#discussion_r3284735594

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: Yes, but a separate API would give you more options when and how this translation happens:

> we may need to create a api call to move checkpoints between nodes

This is a good example; it might require a coordination to determine where to store/move/pull the checkpoint before restoring it.

Nevertheless, if the restoreFrom is generic enough and well documented, these features could be added later.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
atiratree: You mean the access rights if we do not use the Restore API?

A very rough measure could be PSA? Or develop a new security feature for this.

haircommander:
I chatted with @rst0git and we're going to continue this in https://github.com/kubernetes/enhancements/pull/5851

PR #6081, opened 2026-05-15
DRA: Derived Attributes

gauravkghildiyal:
- One-line PR description: Add KEP-6080 introducing derivedAttributes for flexible DRA device co-allocation.

- Issue link: https://github.com/kubernetes/enhancements/issues/6080

- Other comments: This KEP introduces scoped CEL expressions (`derivedAttributes`) to `DeviceRequest` to synthesize virtual grouping keys on the fly. 

/sig scheduling
/wg device-management

### DIFF
--- a/keps/sig-scheduling/6080-dra-derived-attributes/README.md
+++ b/keps/sig-scheduling/6080-dra-derived-attributes/README.md
@@ -318,6 +318,7 @@ type DeviceRequest struct {
 	// +optional
 	// +listType=map
 	// +listMapKey=name
+	// +k8s:optional
 	// +k8s:maxItems=8
 	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
 }
--- a/keps/sig-scheduling/6080-dra-derived-attributes/README.md
+++ b/keps/sig-scheduling/6080-dra-derived-attributes/README.md
@@ -1,0 +1,614 @@
+# KEP-6080: DRA Derived Attributes
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [End-User Rigidity](#end-user-rigidity)
+  - [Attribute Standardization Bottleneck](#attribute-standardization-bottleneck)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Core Manifest Example: GPU &amp; NIC NUMA Alignment](#core-manifest-example-gpu--nic-numa-alignment)
+  - [User Stories](#user-stories)
+    - [Custom Multi-Device Grouping (GPU + NIC + CPU)](#custom-multi-device-grouping-gpu--nic--cpu)
+    - [Substring / Regex Topology Extraction](#substring--regex-topology-extraction)
+    - [Dynamic Capacity Tiering](#dynamic-capacity-tiering)
+    - [Implicit Hardware Alignment via Naming Conventions](#implicit-hardware-alignment-via-naming-conventions)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+    - [KEP-5254 Comparison (DRA: Constraints with CEL)](#kep-5254-comparison-dra-constraints-with-cel)
+    - [Interaction with KEP-5491 (List Types for Attributes)](#interaction-with-kep-5491-list-types-for-attributes)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [CEL Environment &amp; Validation](#cel-environment--validation)
+  - [Scheduler Plugin Implementation](#scheduler-plugin-implementation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+DRA currently relies on exact literal matching of device attributes across
+requests (`matchAttributes`). While performant, literal matching is rigid
+for end users. Workloads needing to pair devices across complex boundaries—
+such as extracting a PCI or NUMA ID from a topology string—cannot express
+these requirements inline.
+
+This KEP introduces `derivedAttributes` to `.devices.requests`. It allows
+users to synthesize virtual grouping keys on the fly using scoped, per-device
+CEL expressions. The scheduler's constraint engine then evaluates these
+derived keys exactly like static attributes.
+
+## Motivation
+
+### End-User Rigidity
+Currently, DRA co-allocation requires exact literal matching of attributes via
+`matchAttributes`. If a user wants to co-locate a GPU and a NIC based on a
+shared PCIe locality, both device drivers must publish the exact same attribute
+name and value. If one driver publishes `pcie_locality: 0` and the other
+publishes `pcie_root: 0` (or embeds it in a topology string like `numa0-pcie1`),
+the scheduler cannot match them. Users have no mechanism to bridge these schemas
+inline.
+
+### Attribute Standardization Bottleneck
+To achieve cross-vendor or multi-driver device pairing today, hardware vendors
+must agree on standardized attribute naming conventions (e.g.,
+`resource.kubernetes.io/pcieRoot`).
+
+However, core Kubernetes components like the scheduler treat these attribute
+names as opaque strings during constraint matching. Forced standardization is
+essentially a human-coordinated workaround for a missing API capability: **the
+inability of the constraint engine to match different attribute names across
+requests**. Relying on this workaround forces vendors into a slow pipeline for
+API approval whenever a new pairing boundary is needed ([real example](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/issues/1123)).
+While standardized labels provide clear benefits for downstream observability
+tools, **forcing their use as the sole mechanism for device alignment creates
+rigid, human-coordinated dependencies rather than enabling flexible, API-driven
+logic.**
+
+While ongoing community efforts to standardize schemas are highly valuable for
+long-term hardware representation, the rapidly evolving AI/ML ecosystem often
+outstrips the turnaround time of formal standardization cycles. Frequently, the
+physical topology information needed for co-allocation is already present on
+device objects (embedded in vendor-specific strings, naming conventions, or
+capacity metrics). The only bottleneck preventing end users from consuming this
+data immediately is the strict requirement that attribute names match exactly
+across requests.
+
+### Goals
+- Allow `.devices.requests` to define virtual grouping keys via scoped,
+  per-device CEL expressions (`derivedAttributes`).
+- Enable co-allocation of heterogeneous hardware across differing vendor
+  attribute schemas without prior human coordination or schema standardization.
+- Preserve the early-pruning performance of the scheduler's constraint engine
+  by scoping CEL evaluation to individual candidate devices.
+
+### Non-Goals
+- Replacing existing static attribute matching (`matchAttributes`).
+- Supporting arbitrary cross-device constraints in CEL (explored in KEP-5254
+  but did not proceed due to scheduler scaling bottlenecks).
+
+## Proposal
+
+We propose extending `.devices.requests` with `derivedAttributes` and
+`.devices.constraints` with `matchDerivedAttribute`. A derived attribute defines
+a CEL expression evaluated against each candidate device object. The resulting
+value acts as a virtual attribute that can be referenced in constraints.
+
+### Core Manifest Example: GPU & NIC NUMA Alignment
+
+*(Note: The community is actively standardizing `numaNode` in [PR #6073](https://github.com/kubernetes/enhancements/pull/6073).
+This example illustrates the API mechanics for bridging disparate schemas).*
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaim
+metadata:
+  name: gpu-numa-alignment-claim
+spec:
+  devices:
+    requests:
+    - name: gpu
+      exactly:
+        deviceClassName: gpu.nvidia.com
+        count: 8
+        # [NEW]: Compute virtual grouping key from GPU driver
+        derivedAttributes:
+        - name: shared-numa-node
+          expression: "device.attributes['gpu.nvidia.com'].numa"
+    - name: nic
+      exactly:
+        deviceClassName: dranet
+        count: 1
+        # [NEW]: Compute virtual grouping key from dranet driver
+        derivedAttributes:
+        - name: shared-numa-node
+          expression: "device.attributes['dra.net'].numaNode"
+    constraints:
+    # [NEW]: Match the derived attribute across both requests
+    - matchDerivedAttribute: shared-numa-node
+      requests: [gpu, nic]
+```
+
+### User Stories
+
+#### Custom Multi-Device Grouping (GPU + NIC + CPU)
+Users need to co-locate heterogeneous hardware (GPUs, high-speed NICs, host
+CPUs) managed by independent drivers. Each driver publishes topology metadata
+under different attribute names. Users can synthesize an identical virtual
+grouping key across all three requests:
+- **GPU**: `expression: "device.attributes['gpu.nvidia.com'].foo_key"`
+- **NIC**: `expression: "device.attributes['dra.net'].bar_id"`
+- **CPU**: `expression: "device.attributes['cpu.intel.com'].baz_domain"`
+
+#### Substring / Regex Topology Extraction
+Device drivers often publish monolithic topology strings (e.g.,
+`topology: "numa0-pcieDomain1-nic0"`). Users can use CEL string manipulation
+to extract the specific boundary needed inline:
+- `expression: "device.attributes['vendor.com'].topology.split('-')[0]"`
+
+#### Dynamic Capacity Tiering
+Users want to group heterogeneous devices into matching performance tiers based
+on capacity or numeric attributes. CEL comparisons can quantize continuous
+ranges into discrete matching bins:
+- `expression: "device.attributes['vendor.com'].memory_gb >= 80 ? 'tier-1' : 'tier-2'"`
+
+#### Implicit Hardware Alignment via Naming Conventions
+When physical topology alignment is known implicitly via device naming
+conventions (e.g., `gpu0` aligns with `eth0`) rather than explicit topology
+attributes, CEL can extract and match the underlying hardware index:
+- **GPU**: `expression: "device.name.replace('gpu', '')"`
+- **NIC**: `expression: "device.name.replace('eth', '')"`
+
+### Notes/Constraints/Caveats
+
+#### KEP-5254 Comparison (DRA: Constraints with CEL)
+[KEP-5254 (PR #5391)](https://github.com/kubernetes/enhancements/pull/5391)
+explored using CEL to express arbitrary constraints across entire device groups
+(`cel.expression`). For example:
+
+```yaml
+    constraints:
+    # KEP-5254 proposed evaluating CEL across an entire device group
+    - cel:
+        expression: "devices[0].attributes['gpu.nvidia.com'].numa == devices[1].attributes['dra.net'].numaNode"
+      requests: [gpu, nic]
+```
+
+While highly flexible, evaluating expressions across entire device groups
+prevented the scheduler from pruning invalid permutations early. Because the CEL
+environment required candidate devices for both `gpu` and `nic` simultaneously
+within the `devices` list, the scheduler had to generate combinatorial device
+permutations before evaluating the constraint. This led to combinatorial
+explosion during device filtering.
+
+`derivedAttributes` resolves this by scoping CEL evaluation strictly to
+individual candidate devices. The scheduler evaluates derived keys as individual
+devices are filtered, maintaining the exact same early-pruning efficiency as
+static attribute matching.
+
+#### Interaction with KEP-5491 (List Types for Attributes)
+[KEP-5491 (PR #5492)](https://github.com/kubernetes/enhancements/pull/5492)
+introduces list-typed attributes to `ResourceSlice` (e.g., lists of strings or
+integers) and redefines `matchAttribute` to evaluate as a non-empty set
+intersection ($\cap v_k \neq \emptyset$), treating scalar values as single-
+element lists.
+
+`derivedAttributes` is fully forward-compatible with KEP-5491 and will inherit its
+capabilities in three key areas:
+1. **CEL Runtime Environment**: The CEL environment for `expression` will expose
+   list-typed attributes exactly as defined in `ResourceSlice` by KEP-5491.
+2. **List Return Types & Synthesis**: The allowed return types for `expression`
+   will be expanded to include list types (`[]string`, `[]int64`, `[]bool`).
+   Crucially, manifest authors can use CEL list literal syntax to dynamically
+   synthesize a list from multiple individual scalar attributes inline (e.g.,
+   `expression: "[device.attributes['v'].r1, device.attributes['v'].r2]"`).
+   This allows bridging older scalar-only drivers with KEP-5491 list matching.
+3. **Constraint Matching Semantics**: When `matchDerivedAttribute` evaluates a
+   derived attribute that returns a list, it will adopt the exact same non-empty
+   intersection matching semantics defined by KEP-5491. Scalar return values
+   will be treated as single-element lists.
+
+### Risks and Mitigations
+- **Risk**: A new CEL expression needs to be evaluated for each candidate device
+  (in addition to any CEL expressions evaluated for device selectors).
+- **Mitigation**: This evaluation adds a constant time increment to each device
+  evaluation. The remainder of the scheduling cycle—including constraint
+  evaluation—remains exactly the same because the derived attribute acts as a
+  standard, non-derived attribute during constraint matching.
+
+## Design Details
+
+### API Changes
+
+We extend `DeviceRequest` in `resource.k8s.io/v1` with `DerivedAttributes`:
+
+```go
+// package resource
+
+type DeviceRequest struct {
+	// Existing fields...
+	Name string `json:"name"`
+	DeviceClassName string `json:"deviceClassName"`
+	// ...
+
+	// DerivedAttributes defines a set of virtual attributes computed via CEL expressions
+	// for each candidate device.
+	// +featureGate=DRADerivedAttributes
+	// +optional
+	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
+}
+
+type DerivedAttribute struct {
+	// Name is the identifier for this derived attribute, used in constraints.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// Expression is a CEL expression evaluated against each candidate device.
+	// The expression must evaluate to a primitive scalar (string, integer,
+	// boolean) or a list of scalars ([]string, []int64, []bool) to act as a
+	// virtual grouping key. Any other return type is an error and causes
+	// device validation or allocation to fail.
+	//
+	// The expression's input is an object named "device", which carries the
+	// following properties:
+	//  - driver (string): the name of the driver which defines this device.
+	//  - attributes (map[string]object): the device's attributes, grouped by prefix
+	//    (e.g. device.attributes["dra.example.com"] evaluates to an object with all
+	//    of the attributes which were prefixed by "dra.example.com").
+	//  - capacity (map[string]object): the device's capacities, grouped by prefix.
+	//
+	// +required
+	Expression string `json:"expression"`
+}
+```
+
+We extend `DeviceConstraint` with `MatchDerivedAttribute`:
+
+```go
+type DeviceConstraint struct {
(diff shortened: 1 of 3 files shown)


keps/sig-scheduling/6080-dra-derived-attributes/kep.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: DRADerivedAttributes
+    components:
+      - kube-apiserver
+      - kube-controller-manager
+      - kube-scheduler
+      - kubelet
```
johnbelamaric: I doubt kubelet needs to know about these

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,614 @@
...
+
+### Non-Goals
+- Replacing existing static attribute matching (`matchAttributes`).
+- Supporting arbitrary cross-device constraints in CEL (explored in KEP-5254
+  but did not proceed due to scheduler scaling bottlenecks).
+
+## Proposal
+
+We propose extending `.devices.requests` with `derivedAttributes` and
+`.devices.constraints` with `matchDerivedAttribute`. A derived attribute defines
```
johnbelamaric: We should not change the constraints. These should just look like any other attribute to the entire constraints section. The only API addition needed should be the `derivedAttributes` field.

johnbelamaric:
@thameem-abbas I think this might be useful for your use cases.

johnbelamaric:
cc @pohly

keps/sig-scheduling/6080-dra-derived-attributes/kep.yaml
gauravkghildiyal: Makes sense removed.

(I'm not too sure about kube-controller-manager either, but maybe its needed there because it clones the ResourceClaimTemplate to ResourceClaims)

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Great idea! Even better.

Updated.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+Recommended: **Option 1 (Allowing FQDNs)**. Most manifest authors will naturally
+choose simple bare names. Conversely, authoring an FQDN override requires
+deliberate effort to duplicate a driver's schema, indicating clear user intent.
+Enabling this pattern provides great flexibility.
+
+### Risks and Mitigations
+- **Risk**: A new CEL expression needs to be evaluated for each candidate device
+  (in addition to any CEL expressions evaluated for device selectors).
+- **Mitigation**: This evaluation adds a constant time increment to each device
+  evaluation. The remainder of the scheduling cycle—including constraint
```
pohly: It's constant per device, but each device might be considered more than once. Caching derived attributes mitigates that.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+	// for each candidate device.
+	// +featureGate=DRADerivedAttributes
+	// +optional
+	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
+}
+
+type DerivedAttribute struct {
+	// Name is the identifier for this derived attribute, used in constraints.
+	//
+	// +required
```
pohly: Document the valid string format here.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+	// Existing fields...
+	Name string `json:"name"`
+	DeviceClassName string `json:"deviceClassName"`
+	// ...
+
+	// DerivedAttributes defines a set of virtual attributes computed via CEL expressions
+	// for each candidate device.
+	// +featureGate=DRADerivedAttributes
+	// +optional
+	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
```
pohly, line 325: Size limit?

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+
+type DerivedAttribute struct {
+	// Name is the identifier for this derived attribute, used in constraints.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// Expression is a CEL expression evaluated against each candidate device.
+	// The expression must evaluate to a primitive scalar (string, integer,
+	// boolean) or a list of scalars ([]string, []int64, []bool) to act as a
```
pohly: We also have semver. Are those valid results?

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+
+`DeviceConstraint` in `resource.k8s.io/v1` requires zero Go struct modifications.
+Existing `MatchAttribute *string` fields will be reused to reference derived
+attributes.
+
+### CEL Environment & Validation
+- **Environment**: The CEL environment for `Expression` is exactly the same
+  as that for `CELDeviceSelector`, containing a single variable `device`.
+- **Return Type**: The CEL expression must evaluate to a scalar (string,
+  integer, boolean) or a list of scalars (`[]string`, `[]int64`, `[]bool`).
```
pohly: This "must evaluate to" cannot be validated during admission time (depends on unknown input types).

How are evaluation errors at runtime handled?

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+	//
+	// The expression's input is an object named "device", which carries the
+	// following properties:
+	//  - driver (string): the name of the driver which defines this device.
+	//  - attributes (map[string]object): the device's attributes, grouped by prefix
+	//    (e.g. device.attributes["dra.example.com"] evaluates to an object with all
+	//    of the attributes which were prefixed by "dra.example.com").
+	//  - capacity (map[string]object): the device's capacities, grouped by prefix.
+	//
+	// +required
```
pohly: Size limit?

pohly (commented):
Looks okay to me.

Shadow API review: we need to sort out some details, see comments.

keps/sig-scheduling/6080-dra-derived-attributes/kep.yaml
```diff
@@ -0,0 +1,47 @@
+title: DRA Derived Attributes
+kep-number: 6080
+authors:
+  - "@gauravkghildiyal"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-network
```
aojea, line 7: I can review from sig network

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+  but did not proceed due to scheduler scaling bottlenecks).
+
+## Proposal
+
+We propose extending `.devices.requests` with `derivedAttributes`. A derived
+attribute defines a CEL expression evaluated against each candidate device
+object. The resulting value acts as a virtual attribute that can be referenced
+directly by existing `.devices.constraints[].matchAttribute` fields.
+
+### Core Manifest Example: GPU & NIC NUMA Alignment
```
aojea, line 163: This is a great example , cc: @ffromani @pravk03

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+We debate two design paths for validating `derivedAttributes` names:
+
+1. **Option 1 (Allow FQDNs e.g., `vendo.com/attr`)**: Enables derived attributes
+   to cleanly **override** static driver attributes of the same name, providing
+   powerful inline flexibility at the slight risk of accidental shadowing.
+2. **Option 2 (Strictly Bare Names e.g., `shared-numa`)**: Enforces an absolute
+   syntactic boundary between static (FQDN) and derived (bare) attributes,
+   eliminating shadowing risks but preventing direct attribute overrides.
+
+Recommended: **Option 1 (Allowing FQDNs)**. Most manifest authors will naturally
```
aojea, line 287: I also think that Option 2 will break existing users of the well defined pcieRoot standard attribute, if a new driver that does not implement the standard attribute want to use it I think it should have to override it with the FQDN

keps/sig-scheduling/6080-dra-derived-attributes/README.md
yongruilin, line 325:
```suggestion
	// +optional
	// +k8s:optional
	// +k8s:maxItems=<int>
	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
```

yongruilin (commented):
Declarative Validation plans to support validation for new API fields without handwritten validation counterpart directly.
But for complicated validation(e.g. CEL compilable), might still need to reply on handwritten.
/cc @jpbetz

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,621 @@
...
+- API validation and scheduler plugin implementation complete.
+- Unit and integration tests passing.
+
+#### Beta
+- Gather feedback from DRA driver maintainers (SIG Node / SIG Network).
+- Any additional e2e tests implemented and running in Testgrid canaries.
+- Verify scheduler performance and latency overhead with large device counts.
+
+#### GA
+- Proven adoption in real-world DRA drivers (e.g., dra-driver-cpu,
```
ffromani: question: from a driver author perspective, what is expectation in order to validate the KEP value proposition? IOW, what a driver should do?

ffromani (commented):
+1 (non binding) for general idea. Offers a clear short term viability/improvement while the naming standardization is ongoing.

General thought: I wonder which incentives do we have (or should we try to set up?) to keep the standardization process ongoing and avoid just to short circuit to the derived attributes? Actually, is this even a concern at all?

keps/sig-scheduling/6080-dra-derived-attributes/README.md
aojea: I think this is just a condition for us , the community, to proof that idea really works in production, so from my perspective seeing tutorials or companies documentation as evidence of that just checks the box  ...

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Thanks. Reworded the 'Mitigation' to include this too.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: I started thinking we could keep it 32, similar to the current attributes+capacities combined on the ResourceSlice.

But I can't think of people needing that many. CEL expressions should allow them to derive a single attribute from multiple original attributes. I'm proposing starting out with 8 derived attributes.

gauravkghildiyal: I've added a maxItems. But I'm not sure what `k8s:optional` is needed in addition to the already existing `optional`

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Updated type to QualifiedName since that seems like what we want.

(I've included a link QualifiedName's definition just for the sake of people reading the KEP. The actual comment in the api spec can exclude it as appropriate)

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Wow, I had no idea that was a thing. I've updated comments to include `semver` as a valid return type to keep things consistent with existing ResourceSlice attribute types.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Copy pasted our `CELDeviceSelector` size limit. For the foreseeable future, I can't see this needing anything different.

```
	// The length of the expression must be smaller or equal to 10 Ki. The
	// cost of evaluating it is also limited based on the estimated number
	// of logical steps.
```

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: +1 to what Antonio said.

I updated the KEP with this bit of context to clarify that there's no "implementation" work which drivers have to do here.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Ah yes I should have clarified this. Missed including it here but my thoughts were to have the same behaviour as CELDeviceSelector. Precisely [this](https://github.com/kubernetes/kubernetes/blob/ebae4193371a0bcdd97f05f2c660a791c1780fdc/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources_test.go#L783-L787)

I've update the KEP to that effect.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Makes sense.

gauravkghildiyal:
> [ffromani](https://github.com/ffromani):  +1 (non binding) for general idea. Offers a clear short term viability/improvement while the naming standardization is ongoing.
> 
> General thought: I wonder which incentives do we have (or should we try to set up?) to keep the standardization process ongoing and avoid just to short circuit to the derived attributes? Actually, is this even a concern at all?

Thanks for the feedback @ffromani and asking about this. I have a couple of thoughts in different form on this topic and I've tried to summarize some of them in the Motivation section of the KEP. Adding to that:

- I share the sentiment behind your question: "Actually, is this even a concern at all?" Currently, resource alignment is the only active use case that I'm aware of for these standard attributes. If other use cases emerge in the future, it is those use cases that should be the ones which incentivize attribute standardization. Forcing upfront standardization for hypothetical future use cases at the expense of the immediate user experience is counterproductive.

- Forcing users to only be able to constraint match on 'standard attributes' is very limiting. The pace of defining 'standard attributes' can never keep up with the pace of newer user cases needing constraint matching based on some new criteria. For a use case that we expect to show up again and again the future, the API should present a generic solution to the problem (which is possible here without many cons) rather than requiring modification (in the form of a new `standard attribute`) each and every time.

- I see DRA API as being built on the premise of a generic interface that allows vendors to expose their hardware as open-ended attributes and the end users to consume and appropriately configure this hardware. The very fact that we chose an `attributes` field instead of 'explicit fields' for hardware properties in the API shows that standardizing hardware properties is a very difficult problem. (Case in point, we've been discussing a numa related attributes close to an year now with the latest discussion at https://github.com/kubernetes/enhancements/pull/6073). DRA is designed to be highly generic, and `derivedAttributes` is the logical next step to solve for this other common user requirement.

ffromani:
>     * I see DRA API as being built on the premise of a generic interface that allows vendors to expose their hardware as open-ended attributes and the end users to consume and appropriately configure this hardware. The very fact that we chose an `attributes` field instead of 'explicit fields' for hardware properties in the API shows that standardizing hardware properties is a very difficult problem. (Case in point, we've been discussing a numa related attributes close to an year now with the latest discussion at [KEP-6072: DRA: Standard numaNode Device Attribute #6073](https://github.com/kubernetes/enhancements/pull/6073)). DRA is designed to be highly generic, and `derivedAttributes` is the logical next step to solve for this other common user requirement.

Thanks @gauravkghildiyal , very good point and nice food for thought.

Just thinking aloud now: we may want to reposition, over time, the standardization of attributes as acknowledgement of well-known widespread attributes after-the-fact rather than be proactive.
IOW, we may want to swap priorities: derived attributes as primary vehicle to actually enable users and observe the behavior, and consolidate later through standardization of attributes once real pattern emerge and solidify.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
yongruilin: `+k8s:optional` is the tag required for DeclarativeValidation while `+optional` is the pre-existing tag for openAPI.

gauravkghildiyal: Thanks. Updated.

gauravkghildiyal:
> [ffromani](https://github.com/ffromani): derived attributes as primary vehicle to actually enable users and observe the behavior, and consolidate later through standardization of attributes once real pattern emerge and solidify.

Fully agreed. Having actual customers already using a pattern should definitely assist with justifying standardization of some of those attributes.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,644 @@
...
+spec:
+  devices:
+    requests:
+    - name: gpu
+      exactly:
+        deviceClassName: gpu.nvidia.com
+        count: 8
+        # [NEW]: Compute virtual grouping key from GPU driver
+        derivedAttributes:
+        - name: shared-numa-node
```
pravk03, line 182: Since we call out that allowing FQDNs (Option 1) as the recommended approach below, we can use that in the example here ?

gauravkghildiyal: Ah, so the intent is that FQDNs are allowed, in addition to 'bare names' (without any prefix domain). I actually expect non-prefixed names, like `shared-numa-node` to be the more commonly used names here, hence the example include just that.

But that's a good point and as requested by Patrick earlier, I've updated the comment above the `Name` API field with the acceptable format = https://github.com/kubernetes/kubernetes/blob/0f935974ee0e38fb77a09e3be1e8754b5720605e/pkg/apis/resource/types.go#L621-L637

pravk03: SG. Thanks.

johnbelamaric (commented):
you need a PRR file too

gauravkghildiyal:
> you need a PRR file too

Update, thanks for the reminder

kad:
LGTM, thanks for the KEP: I believe derived attributes will be good help to express complex topological concepts and give HW vendors a bit more flexibility to provide customers the most optimal combinations of the  devices in custom system configurations.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -314,20 +316,26 @@ type DeviceRequest struct {
...
+	// +listMapKey=name
+	// +k8s:optional
+	// +k8s:maxItems=8
 	DerivedAttributes []DerivedAttribute `json:"derivedAttributes,omitempty"`
 }
 
 type DerivedAttribute struct {
 	// Name is the identifier for this derived attribute, used in constraints.
+  //
+	// Format: https://github.com/kubernetes/kubernetes/blob/0f935974ee0e38fb77a09e3be1e8754b5720605e/pkg/apis/resource/types.go#L621-L637
```
pohly: Such a link is not going to be very useful in generated docs. Better just say "It has the same format as the name of attributes in a ResourceSlice."

This is also the right place to add some guidance for users whether a domain prefix is needed or when it should (or shouldn't) be used.

We don't seem to have a declarative validation format for this yet.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -314,20 +316,26 @@ type DeviceRequest struct {
...
 	//
 	// +required
```
pohly: Instead of `+required` and `+optional`, use `+k8s:required` and `+k8s:optional` (the newer declarative validation tags). You then don't need to write that validation manually.

pohly (changes requested):

thameem-abbas:
We're hoping to adopt derivedAttributes and KEP-5491 concurrently for our cross-driver GPU-NIC pairing use case (`gpu.nvidia.com` + `dra.net`, [webhook source](https://github.com/openshift-psap/dra-rail-admission-webhook)).

From our [KEP-5491 feedback](https://github.com/kubernetes/enhancements/issues/5491#issuecomment-4337499390), @everpeace confirmed that scalar-as-singleton-set normalization is purely scheduler-side — the raw attribute type in ResourceSlice changes when a driver migrates from scalar to list.

How should derivedAttributes expressions handle this? If a driver migrates `pcieRoot` from `string` to `list(string)`, does `device.attributes['gpu.nvidia.com'].pcieRoot` break in the CEL expression? Should we be writing defensive expressions, or does the CEL environment normalize this?

Curious how others are thinking about this overlap.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Sure thing, I've updated the description and removed the link. Thanks!

(The link was anyways a temporary thing for the KEP review, so people can look up what that format is, and not something I intended to be part of the final spec.)

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Done. I've completely removed `+required` and `+optional` and replaced with their declarative validation counterparts.

gauravkghildiyal (commented):
Thanks for the review @pohly. Incorporated your suggestions.

gauravkghildiyal:
> We're hoping to adopt derivedAttributes and KEP-5491 concurrently for our cross-driver GPU-NIC pairing use case (`gpu.nvidia.com` + `dra.net`, [webhook source](https://github.com/openshift-psap/dra-rail-admission-webhook)).
> 
> From our [KEP-5491 feedback](https://github.com/kubernetes/enhancements/issues/5491#issuecomment-4337499390), @everpeace confirmed that scalar-as-singleton-set normalization is purely scheduler-side — the raw attribute type in ResourceSlice changes when a driver migrates from scalar to list.
> 
> How should derivedAttributes expressions handle this? If a driver migrates `pcieRoot` from `string` to `list(string)`, does `device.attributes['gpu.nvidia.com'].pcieRoot` break in the CEL expression? Should we be writing defensive expressions, or does the CEL environment normalize this?
> 
> Curious how others are thinking about this overlap.

It's gonna be extremely difficult for DRA Driver to directly change pcieRoot from a single element to a list type without breaking users. In the best case scenario, newer users of the pcieRoot attribute are aware of this nuance (single element v/s list) and they write CEL expressions in `derivedAttributes` to be resilient of this change. But that still doesn't safeguard any existing users of this attribute who don't have such safeguards within their CEL expressions.

How I see things is that DRA Drivers should instead introduce a `pcieRoots` (plural) attribute if/when they want a list type. Then, `derivedAttributes` gives end users the flexibility to start consuming `pcieRoots` list in whatever way they want.

johnbelamaric:
We have already progressed on doing it transparently, with a change in the matchAttribute semantics and a special CEL function to simplify that. I wonder if there is some best practice + CEL function to help manage derived attributes that manipulate such an attribute. For example, use of a type-guarded "map" function or something?

gauravkghildiyal:
> We have already progressed on doing it transparently, with a change in the matchAttribute semantics and a special CEL function to simplify that.

Good point. Yes for matchAttribute its simpler because there is no CEL expression involved and the kubernetes scheduler can be made intelligent to be able to 'match' an element with a single-element-list. 

I'm referring more to the existing use of pcieRoot attribute within CELExpressions for selection purposes. In particular, this limitation highlighted in the [List Attributes KEP](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5491-dra-list-types-for-attributes#introduce-include-function-in-cel)
> image

In light of this, I'm thinking it may serve our end users better if DRA Drivers don't try to migrate their single-element to list-type and instead introduce a different attribute for lists when needed.

thameem-abbas:
I like the idea of different APIs if that's just made clear now before more DRA drivers standardize and it becomes harder

johnbelamaric:
Two different attributes force users to know which one each driver publishes. The same attribute avoids that.

gauravkghildiyal:
> Two different attributes force users to know which one each driver publishes. The same attribute avoids that.

But as things exist today in [KEP-5491-dra-list-types-for-attributes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5491-dra-list-types-for-attributes#introduce-include-function-in-cel), we don't yet have a solution which covers all backward compatibility issues and the migration of single scalar to a list type isn't transparent. So the cost of migrating an attribute from scalar to a list is:

- Potential of breaking existing CEL expressions that expect scalar value
- While it simplifies the naming (by keeping the same name), it forces users to use specialized CEL functions to handle the overloaded type, which may negate the UX gains.

These are all great discussions to have, but is it fair to say a solution to this problem is better discussed in the scope of KEP-5491 and is independent of this KEP? CC: @everpeace

ffromani:
> > Two different attributes force users to know which one each driver publishes. The same attribute avoids that.
> 
> But as things exist today in [KEP-5491-dra-list-types-for-attributes](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5491-dra-list-types-for-attributes#introduce-include-function-in-cel), we don't yet have a solution which covers all backward compatibility issues and the migration of single scalar to a list type isn't transparent. So the cost of migrating an attribute from scalar to a list is:
> 
>     * Potential of breaking existing CEL expressions that expect scalar value
> 
>     * While it simplifies the naming (by keeping the same name), it forces users to use specialized CEL functions to handle the overloaded type, which may negate the UX gains.
> 
> 
> These are all great discussions to have, but is it fair to say a solution to this problem is better discussed in the scope of KEP-5491 and is independent of this KEP? CC: @everpeace

+1. There's very active [work about adding the pcieRoot attribute support to the dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu/pull/156) which we aim to ship in 0.2.0 and we could use some guidance, but I don't want to derail the KEP review. Perhaps a topic for the WG-device-management meeting?

dom4ha:
There are no concerns regarding this feature from the scheduler perspective. The time overhead is linear, so fully acceptable. Leaving the usability aspects on wg-device-management to decide.
/approve

dom4ha:
There are no concerns regarding this feature from the scheduler perspective. The time overhead is linear, so fully acceptable. Leaving the usability aspects on wg-device-management to decide.
/approve

aojea:
/lgtm (sig network)

networking community is using DRA as extension mechanism, but is unrealistic for networking to agree in all the possible standard attributes in such fragmented space. This allow us to provide a simple mechanism for iteration

gauravkghildiyal:
Thanks for the reviews @dom4ha and @aojea 

@johnbelamaric and @pohly -- Can one of you give the final approvals (including PRR)? Thanks

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+	Name string `json:"name"`
+	DeviceClassName string `json:"deviceClassName"`
+	// ...
+
+	// DerivedAttributes defines a set of virtual attributes computed via CEL expressions
+	// for each candidate device.
+	// +featureGate=DRADerivedAttributes
+	// +listType=map
+	// +listMapKey=name
+	// +k8s:optional
```
pohly, line 323: Nit: I know I suggested `+k8s:optional` as replacement for `+optional` earlier, but since then I learned that *both* are needed (the former for declarative validation, the latter for OpenAPI).

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+	// []bool, []semver) to act as a virtual grouping key. Any other return type
+	// is an error and causes CEL evaluation for the device to fail.
+	//
+	// The expression's input is an object named "device", which carries the
+	// following properties:
+	//  - driver (string): the name of the driver which defines this device.
+	//  - attributes (map[string]object): the device's attributes, grouped by prefix
+	//    (e.g. device.attributes["dra.example.com"] evaluates to an object with all
+	//    of the attributes which were prefixed by "dra.example.com").
+	//  - capacity (map[string]object): the device's capacities, grouped by prefix.
```
pohly: Nit: let's just say "which carries the same properties as in a CELDeviceSelector`?

For example, `allowMultipleAllocations` probably will be available (why should it not be?), but isn't described here.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+  ([Coverage](https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit&include-filter-by-regex=k8s.io%2Fkubernetes%2Fpkg%2Fapis%2Fresource%2Fvalidation%7Ck8s.io%2Fkubernetes%2Fpkg%2Fscheduler%2Fframework%2Fplugins%2Fdynamicresources&show-stale-tests=))
+  - Verify validation of `derivedAttributes` (valid/invalid CEL syntax).
+- `k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources`:
+  `2026-05-01` - `>80%`
+  ([Coverage](https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit&include-filter-by-regex=k8s.io%2Fkubernetes%2Fpkg%2Fapis%2Fresource%2Fvalidation%7Ck8s.io%2Fkubernetes%2Fpkg%2Fscheduler%2Fframework%2Fplugins%2Fdynamicresources&show-stale-tests=))
+  - Verify CEL compilation caching.
+  - Verify correct CEL evaluation and constraint matching.
+
+##### Integration tests
+- `test/integration/scheduler/dra`:
```
pohly: Instead of doing this in `test/integration/scheduler/dra`, do this in `test/integration/scheduler_perf/dra/derived-attributes` (new folder). Make sure to include a realistic scenario (number of attributes, complexity of the CEL expressions), then define a simple case for correctness checking and larger cases for performance measurement.

See consumable capacity (currently being updated in https://github.com/kubernetes/kubernetes/pull/139511 because we didn't do that well initially).

With that done I don't see a need to add more testcases for this to test/integration/scheduler/dra.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+
+#### Alpha
+- Feature gate `DRADerivedAttributes` implemented.
+- API validation and scheduler plugin implementation complete.
+- Unit and integration tests passing.
+
+#### Beta
+- Gather feedback from DRA driver maintainers (SIG Node / SIG Network).
+- Any additional e2e tests implemented and running in Testgrid canaries.
+- Verify scheduler performance and latency overhead with large device counts.
```
pohly: This is what you'll need the scheduler_perf cases for.

pohly (commented):
Some nits, a suggestion for integration/performance testing, otherwise almost there as far as I am concerned.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: DRADerivedAttributes
+  - Components depending on the feature gate: kube-apiserver,
+    kube-controller-manager, kube-scheduler, kubelet
```
johnbelamaric: why kubelet?

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+Existing `ResourceClaims` with `derivedAttributes` will resume being evaluated
+correctly by the scheduler.
+
+###### Are there any tests for feature enablement/disablement?
+
+Yes. Unit tests in `pkg/apis/resource/validation` verify that
+`derivedAttributes` and non-FQDN `matchAttribute` strings are rejected when the
```
johnbelamaric: Also cover `distinctAttribute`

keps/sig-scheduling/6080-dra-derived-attributes/README.md
```diff
@@ -0,0 +1,651 @@
...
+  detected and resolved.
+
+### Scheduler Plugin Implementation
+In `pkg/scheduler/framework/plugins/dynamicresources`:
+1. The plugin compiles the CEL expressions defined in `derivedAttributes` for
+   all requests in the Pod's `ResourceClaims`. Compiled ASTs are cached.
+2. When evaluating candidate devices for a request, the plugin executes the
+   cached CEL expressions against each `Device` object. The computed values are
+   stored in a temporary map associated with the candidate device.
+3. When evaluating `matchAttribute` constraints during the plugin
```
johnbelamaric: it should apply to anything in `constraints`, not just `matchAttribute`. Today that means also `distinctAttribute`.

Also: can derived attributes be used in device request CEL selectors? ie, are they evaluated BEFORE or AFTER the selectors?

johnbelamaric (commented):
A few minor questions/comments but PRR generally looks good. Please respond to these and then I can give approval

keps/sig-scheduling/6080-dra-derived-attributes/README.md
pohly: That's a good point. In CEL expressions, the derived attributes aren't really needed (the expression itself can embed the necessary mapping). But it would be more consistent and perhaps even expected if they were present. I'm +1 for that.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: No worries. Updated with `+k8s:optional`

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Sure, updated the statement to:

```
	// The expression's input is an object named "device", which carries the
	// same properties as in a CELDeviceSelector.
```

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Thanks. Updated the directory and things to keep in mind during implementation.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Thanks, updated a reference to the scheduler_perf test discussed in the other comment.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: My bad. You did ask me to clean this up from the PRR but I missed cleaning it up here. Removed now.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Explicitly worded `distinctAttributes` too.

keps/sig-scheduling/6080-dra-derived-attributes/README.md
gauravkghildiyal: Thanks @johnbelamaric and @pohly! The use of derivedAttributes within CELDeviceSelector is something I've thought about, but intentionally did not include in this KEP.

My main reason for keeping it out of scope—as you both noted—is that it isn't strictly necessary. If a user needs that logic in the selector, the CELDeviceSelector can just embed the necessary mapping directly. Attempting to include derivedAttributes into selectors brings additional complexities, such as how they impact performance by forcing early evaluation, and how they should be referenced in the CEL environment (i.e. whether we add them to `device.attributes` or have to introduce a new `device.derivedAttributes` map)

Perhaps earlier in the KEP cycle we could have had these discussions in detail, but seeing as we're close to the KEP deadline, I'd rather avoid introducing this change without more careful consideration. If this usability experience is needed in the future, a different KEP would be a perfect opportunity to explore that avenue. Hope you both share the same thoughts :)

With that in mind, I've added this bit of clarification:

```
- **Evaluation Order**: Derived attributes are evaluated **after** the device
  request's `CELDeviceSelector` has filtered candidate devices. They are not
  injected into the selector's CEL environment and are exclusively used for
  evaluating constraints like `matchAttribute` and `distinctAttribute`.
```

and also added a separate section `## Future Considerations > ### Derived Attributes in Device Selectors` on the current rationale and future improvement through a separate enhancement (if needed)

PR #6085, opened 2026-05-16
Mark DRA Prioritized List as implemented

mortent:
- One-line PR description: The DRA Prioritized List is GA and implemented

- Issue link: https://github.com/kubernetes/enhancements/issues/4816

- Other comments:

PR to lock the feature to default-on for 1.37: https://github.com/kubernetes/kubernetes/pull/139110

### DIFF
--- a/keps/sig-scheduling/4816-dra-prioritized-list/kep.yaml
+++ b/keps/sig-scheduling/4816-dra-prioritized-list/kep.yaml
@@ -7,7 +7,7 @@ owning-sig: sig-scheduling
 participating-sigs:
   - sig-node
   - sig-autoscaling
-status: implementable
+status: implemented
 creation-date: 2024-09-24
 reviewers:
   - "@pohly"
@@ -44,7 +44,7 @@ feature-gates:
       - kube-apiserver
       - kube-controller-manager
       - kube-scheduler
-disable-supported: true
+disable-supported: false
 
 # The following PRR answers are required at beta release
 metrics:


kannon92 (commented):
/lgtm

/assign @johnbelamaric

johnbelamaric:
/lgtm

Needs sig-scheduling approval.
/assign @macsko

macsko:
Shouldn't we mark the KEP as implemented **after** it's GA and locked in the code? @mortent could you work on resolving the comments and merging the https://github.com/kubernetes/kubernetes/pull/139110 first?

pohly:
> Shouldn't we mark the KEP as implemented after it's GA and locked in the code?

Agreed. IMHO cleanup after GA graduation doesn't need tracking by SIG release, so let's just flip the feature gate and the merge this PR, without tracking the feature for another release cycle in 1.37.

kannon92:
> Agreed. IMHO cleanup after GA graduation doesn't need tracking by SIG release, so let's just flip the feature gate and the merge this PR, without tracking the feature for another release cycle in 1.37.

Technically this stable graduation did not go through in 1.36/1.37. So this KEP should track the move to stable graduation.

Once the feature gate merges, we shouldn't need to track this anymore imo.

pohly:
> Technically this stable graduation did not go through in 1.36/1.37. 

It did?

https://github.com/kubernetes/kubernetes/blob/2f5e22e8bb66836f767715d99a83662250968bf5/pkg/features/kube_features.go#L1335

It was the ResourceClaim status which was left in beta, in contrast to what we had planned.

kannon92:
I realize that the odd thing is that we promoted the feature to GA but did not lock to default. I wasn't aware that this was possible honestly.

So I agree with you that it sounds like this doesn't really need to be tracked. I am looking at the issue template and it wasn't kept up to date so I missed that this merged.

PR #6086, opened 2026-05-18
KEP-5328: Promote Node Declared Features to GA.

pravk03:
- One-line PR description: Promote Node Declared Features to GA in 1.37

- Issue link: https://github.com/kubernetes/enhancements/issues/5328

- Other comments:

### DIFF
--- a/keps/sig-node/5328-node-declared-features/README.md
+++ b/keps/sig-node/5328-node-declared-features/README.md
@@ -70,16 +70,16 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 - [x] (R) KEP approvers have approved the KEP status as `implementable`
 - [x] (R) Design details are appropriately documented
 - [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
-  - [] e2e Tests for all Beta API Operations (endpoints)
-  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
-  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+  - [x] e2e Tests for all Beta API Operations (endpoints)
+  - [x] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [x] (R) Minimum Two Week Window for GA e2e tests to prove flake free
 - [x] (R) Graduation criteria is in place
-  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
-- [] (R) Production readiness review completed
-- [] (R) Production readiness review approved
+  - [x] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
 - [x] "Implementation History" section is up-to-date for milestone
 - [x] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [x] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 
 [kubernetes.io]: https://kubernetes.io/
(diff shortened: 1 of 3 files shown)


swatisehgal (commented):
Graduation to GA criteria mentions the following:  

> - N examples of real-world usage
> - N installs
> - Allowing time for feedback
> - All issues and gaps identified as feedback during beta are resolved

Of these, the only criterion I'm concerned about is "allowing time for feedback."

We've been fairly aggressive with the graduation timeline, moving from Alpha in v1.35 to Beta in v1.36 and now proposing GA in v1.37. I would typically expect a feature to spend at least a couple of release cycles in Beta to allow for broader adoption, feedback, and identification of any issues that may not have surfaced during earlier testing.

Do we have any SIG Node guidance or precedent on how this criterion is evaluated? If not, is this something we should discuss in a SIG Node meeting?

In the meantime, it would be helpful to understand:
1. Have we received feedback from actual Beta users? If so, what feedback have we received and how has it been addressed?
2. Have downstream consumers and distributions had sufficient opportunity to test the feature in Beta?
3. Are there any known edge cases, limitations, or operational concerns that may only emerge with broader adoption?

pravk03:
Thanks @swatisehgal for reviewing and raising this valid concern.

The main rationale for pushing for GA in v1.37 is that we currently have [5 features](https://github.com/kubernetes/kubernetes/tree/7a1385a332e503ac051edc39a57e0d84fc5c4e98/staging/src/k8s.io/component-helpers/nodedeclaredfeatures/features) already using Node Declared Features for version skew management. As of right now, 4 of these dependent features are in Beta and 1 is in Alpha. We believe that this framework needs to graduate to GA before the features depending on it can safely graduate to GA themselves.

> 1. Have we received feedback from actual Beta users? If so, what feedback have we received and how has it been addressed?

Since this feature is not user facing, our primary "Beta users" are other feature owners using this for skew management. There hasn't been any issues reported in the features currently using the framework. We did remove one feature declaration during beta (`GuaranteedQoSPodCPUResize`) as it would cause a regression on older nodes ([issue](https://github.com/kubernetes/kubernetes/issues/135329)). KEP was updated to address this and tests were added in the beta implementation to check for strict feature gate dependency in any registered features.

> 2. Have downstream consumers and distributions had sufficient opportunity to test the feature in Beta?

Downstream consumers like the Cluster Autoscaler implicitly use the feature during scheduling simulations to evaluate existing node groups (scale-from-1) for pending pods. Testing was added to validate this (https://github.com/kubernetes/autoscaler/pull/8952). "Scale-from-0" was discussed  during the 1.36 Beta cycle, and it was [agreed](https://github.com/kubernetes/enhancements/pull/5797#issuecomment-3886560750) that it can remain out of scope for the KEP.  We added the necessary APIs in the framework to enable autoscalers to integrate this in the future.

> 3. Are there any known edge cases, limitations, or operational concerns that may only emerge with broader adoption?

We have not identified edge cases that block GA. There was a recent discussion in SIG Node meeting about dynamically updating declared features ([issue](https://github.com/kubernetes/kubernetes/issues/138099)), but the consensus was to target this functionality with a separate KEP.  During Alpha, a concern was raised regarding scheduler performance. This was addressed by introducing a [bitmap based approach](https://github.com/kubernetes/kubernetes/pull/136185) for declared feature based node filtering. We haven't seen any significant impact on scheduler throughput since this optimization was introduced in beta (benchmark results added in the KEP).

Happy to bring this up for discussion in the next SIG Node meeting. Please let me know what you think!

swatisehgal:
Thanks for the detailed explanation about the current state of this KEP and its rationale behind its graduation to GA. I think we have a strong case here to proceed but it would really help to capture some of this rationale in the GA graduation section of the KEP. In particular, the fact that the framework is already being consumed by multiple features, the feedback received during Beta, and the issues that were identified and addressed. This would help make the case for GA more explicit for future reviewers.

swatisehgal (commented):
/lgtm

from sig-node perspective

pravk03:
Thanks @swatisehgal. I have updated the GA graduation criteria with details from https://github.com/kubernetes/enhancements/pull/6086#issuecomment-4615255802.

swatisehgal:
> Thanks @swatisehgal. I have updated the GA graduation criteria with details from [#6086 (comment)](https://github.com/kubernetes/enhancements/pull/6086#issuecomment-4615255802).

Thanks!

/lgtm

swatisehgal:
/assign @SergeyKanzhelev @dchen1107 
for node approval
/assign @johnbelamaric
for PRR

keps/sig-node/5328-node-declared-features/README.md
```diff
@@ -520,9 +535,28 @@ implementing this enhancement to ensure the enhancements have also solid foundat
...
+| **NodeDeclaredFeatures Enabled** (20 Declared Features) | [perf-dash](https://perf-dash.k8s.io/#/?jobname=scheduler-perf-benchmark&metriccategoryname=Scheduler&metricname=BenchmarkPerfScheduling&Metric=SchedulingThroughput&Name=BenchmarkPerfScheduling%2FNodeDeclaredFeaturesEnabled%2F5000Nodes20DeclaredFeatures%2Ftest&event=not%20applicable&extension_point=not%20applicable&plugin=not%20applicable&result=not%20applicable) | P50:1205 P90:1623 P99:1648 |
+| **NodeDeclaredFeatures Enabled** (100 Declared Features) | [perf-dash](https://perf-dash.k8s.io/#/?jobname=scheduler-perf-benchmark&metriccategoryname=Scheduler&metricname=BenchmarkPerfScheduling&Metric=SchedulingThroughput&Name=BenchmarkPerfScheduling%2FNodeDeclaredFeaturesEnabled%2F5000Nodes100DeclaredFeatures%2Ftest&event=not%20applicable&extension_point=not%20applicable&plugin=not%20applicable&result=not%20applicable) | P50:1185 P90:1616 P99:1693|
```
macsko: It would be good to write how many benchmark runs were performed. Rule of thumb is to have at least 6 runs for each scenario. You can put then the averages of percentiles (ang average of averages) here.

macsko: They run in CI anyway, so you can just explain how you computed the percentiles numbers.

SergeyKanzhelev (approved):
/approve

keps/sig-node/5328-node-declared-features/README.md
pravk03: Thanks. I have updated the text and table with the average of the percentiles across the 6 most recent CI runs.

macsko:
/lgtm
From scheduling POV

keps/sig-node/5328-node-declared-features/README.md
```diff
@@ -70,13 +70,13 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
-  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+  - [x] e2e Tests for all Beta API Operations (endpoints)
+  - [x] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [x] (R) Minimum Two Week Window for GA e2e tests to prove flake free
 - [x] (R) Graduation criteria is in place
-  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
-- [] (R) Production readiness review completed
-- [] (R) Production readiness review approved
+  - [x] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
```
johnbelamaric: check these off, so they are marked done along with the approval

johnbelamaric (commented):
PRR LGTM, awaiting SIG approval

keps/sig-node/5328-node-declared-features/README.md
pravk03: Updated.

PR #6088, opened 2026-05-18
Update KEP 3157 (watch-list) for milestone 1.37

p0lyn0mial:
- One-line PR description: Update KEP 3157 (watch-list) for milestone 1.37

- Issue link: https://github.com/kubernetes/enhancements/issues/3157

- Other comments: N/A

### DIFF
--- a/keps/sig-api-machinery/3157-watch-list/README.md
+++ b/keps/sig-api-machinery/3157-watch-list/README.md
@@ -104,6 +104,9 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Beta2](#beta2)
     - [Beta3](#beta3)
     - [Beta4](#beta4)
+    - [Beta5](#beta5)
+      - [Backward compatibility](#backward-compatibility)
+      - [Scale test results (5000 nodes, ~165K pods)](#scale-test-results-5000-nodes-165k-pods)
     - [GA](#ga)
     - [Post-GA](#post-ga)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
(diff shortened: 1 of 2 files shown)


keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -795,6 +798,39 @@ The discussion happened in [this document] and resulted in the following update
 
 [this document]: https://docs.google.com/document/d/1x30DjXSSF5krpyoTCwManJg6vphpnGI37xfCRaiHAbs
 
+#### Beta5
+- Enable gzip compression for the WatchList request. Gated behind a new server-side
+  feature gate `WatchListCompression` (Beta, enabled by default).
+  Regular watch requests are not affected.
```
wojtek-t, line 804: I'm ok with that to reduce blast radius.
And then treat expanding it to all watch requests as a follow-up enhancement.

OTOH, with watch-list enabled by-default and used by reflector, we already have pretty large exposure, so maybe that won't by us much anyway?

What splitting protects us from is when it has some negative impact on watch under some scenarios we didn't predict, and then people has to disable it and revert the gains for watch-list.
So maybe (given the implementation is common), we effectively need two separate FGs:
- one that enables it for watch-list
- second that enables it for regular watch requests?

@jpbetz @liggitt - for your thoughts too

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -795,6 +798,39 @@ The discussion happened in [this document] and resulted in the following update
...
+
+##### Backward compatibility
+
+No client-side changes are required. Go's `http.Transport` automatically sends
+`Accept-Encoding: gzip` when `DisableCompression` is `false` (the default in
+`rest.Config`), the same mechanism already used for LIST compression. Older
+clients will transparently receive compressed WatchList responses. 
+The `pull-kubernetes-gce-master-scale-performance-5000` test validates this: all
```
wojtek-t: Link the POC that you were running the test against.

keps/sig-api-machinery/3157-watch-list/README.md
```diff
@@ -795,6 +798,39 @@ The discussion happened in [this document] and resulted in the following update
...
+
+We ran the `pull-kubernetes-gce-master-scale-performance-5000` test with WatchList
+compression enabled and compared against a [baseline][baseline-json] run without it.
+The WatchList P99 latency for pods dropped from 60s (capped at the histogram bucket
+ceiling, actual latency was much worse) to ~30s:
+
+|  | [Baseline][baseline-json] | [With compression][compression-json] |
+|--|----------|-----------------|
+| P50 | 989ms | 969ms |
+| P90 | 5,384ms | 17,150ms |
```
wojtek-t: Why is p90 3x higher with compression?

p0lyn0mial: good catch, now compared against `ci-kubernetes-e2e-gce-scale-performance-5000` not `ci-kubernetes-e2e-gce-scale-resource-size`


now, he ~2s difference at P90 is probably just run-to-run variance. also the run with compression made more requests to the server (higher variance)

keps/sig-api-machinery/3157-watch-list/README.md
p0lyn0mial: done.

keps/sig-api-machinery/3157-watch-list/README.md
p0lyn0mial: > OTOH, with watch-list enabled by-default and used by reflector, we already have pretty large exposure, so maybe that won't by us much anyway?


I would personally consider enabling compression for WATCH requests only when needed. It seems that enabling compression in this case could reduce the amount of data sent from the server, at the cost of higher CPU usage.

jpbetz: I favor the reduced blast radius of `WatchListCompression`. I'm not convinced we should be transitioning to on-by-default compression for watch, and I don't see a strong reason to bundle work into this KEP.

jpbetz (approved):
/approve
/lgtm

Issue #6090, opened 2026-05-18
Kubelet-CRI Exec Environment Variable Injection Contract


nispriha:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add `repeated KeyValue envs` field to CRI `ExecRequest` and `ExecSyncRequest` to enable environment variable injection into exec sessions without relying on in-container binaries or shell wrappers.
- Kubernetes Enhancement Proposal:  https://github.com/kubernetes/enhancements/pull/6091
- Discussion Link:  https://kubernetes.slack.com/archives/C019LFTGNQ3/p1774874170937829
- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6091
    - [ ] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/139023
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._
cc: @haircommander @ngopalak-redhat @sigtrapd

karimzakzouk:
Hello @nispriha 👋, v1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for v1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [x] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

For this KEP, we would just need to update the following:
- PR with the PRR approval file created for the `alpha` stage

[...]

PR #6091, opened 2026-05-18
KEP-6090: Add KEP for kubelet-CRI exec environment variable injection

nispriha:
- One-line PR description: Add KEP for kubelet-CRI exec environment variable injection via `repeated KeyValue envs` in `ExecRequest` and `ExecSyncRequest`.

- Issue link: https://github.com/kubernetes/enhancements/issues/6090

- Implementation PR: https://github.com/kubernetes/kubernetes/pull/139023

- Other comments:

### DIFF
--- a/keps/sig-node/6090-exec-env-injection/README.md
+++ b/keps/sig-node/6090-exec-env-injection/README.md
@@ -1,0 +1,928 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+Follow the guidelines of the [documentation style guide].
+In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-6090: Kubelet-CRI Exec Environment Variable Injection Contract
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Environment Variable Precedence](#environment-variable-precedence)
+  - [Scope](#scope)
+  - [Notes on Silent Degradation](#notes-on-silent-degradation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+Add a `repeated KeyValue envs` field to `ExecRequest` and `ExecSyncRequest` in the CRI protobuf. Establish a contract that the container runtime must inject whatever environment variables the kubelet provides into the exec'd process via `execve()` `envp[]`, without any filtering, interpretation, or policy enforcement by the runtime.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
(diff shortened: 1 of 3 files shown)


keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+Add a `repeated KeyValue envs` field to `ExecRequest` and `ExecSyncRequest` in the CRI protobuf. Establish a contract that the container runtime must inject whatever environment variables the kubelet provides into the exec'd process via `execve()` `envp[]`, without any filtering, interpretation, or policy enforcement by the runtime.
```
SergeyKanzhelev, line 182: make sure to read https://github.com/kubernetes/kubernetes/issues/138798 and select the right data types.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+- Align `ExecRequest` with the existing `ContainerConfig.envs` pattern used for container creation
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+- Exposing exec-time env var injection as a user-facing feature (deferred to a separate KEP)
```
SergeyKanzhelev, line 222: why is it deferred? We typically ask to develop a feature alongside the API to ensure that API can be used: https://www.kubernetes.dev/docs/code/cri-api-version-skew-policy/#feature-development

In this particular case, API change is very straightforward. So the biggest question for me is how exactly it will be used.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- Add a `repeated KeyValue envs` field to `ExecRequest` and `ExecSyncRequest` in the CRI protobuf
+- Establish a clear contract: the container runtime must mechanically forward the provided env vars into the exec'd process via `envp[]`
+- Define environment variable precedence rules upfront so that dependent KEPs (user-facing env injection, audit ID injection) do not need to relitigate them
+- Align `ExecRequest` with the existing `ContainerConfig.envs` pattern used for container creation
+
+### Non-Goals
```
SergeyKanzhelev, line 215: what about env vars expansion? If expansion supported, is it expanded against the existing environment? If not - how will it be protected? Should we allow to inject some dynamic data similar to the downward API?

SergeyKanzhelev, line 215: what is relations of these env vars with other env vars like what was specified on container creation?

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+    int64 timeout = 3;
+    repeated KeyValue envs = 4;  // new field
+}
+```
+The contract established by this KEP:
+
+- The runtime must inject all provided env vars into the exec'd process via `envp[]`
+- The runtime must not filter, modify, or interpret the env vars
+- The runtime must not use the env vars for any policy enforcement
+- The runtime treats this as a mechanical forwarding obligation, not a semantic one
```
SergeyKanzhelev, line 264: this KEP repeating those statements. What is the risk this statement is trying to protect from? Container Runtime injecting something? Filtering out some variables for security purposes?

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+}
+
+message ExecSyncRequest {
+    string container_id = 1;
+    repeated string cmd = 2;
+    int64 timeout = 3;
+    repeated KeyValue envs = 4;  // new field
+}
+```
+The contract established by this KEP:
```
SergeyKanzhelev, line 259: is there any contract on logging? Today exec session output is shown in container logs and having these env var support may be a good way to work around exposing some secrets in logs

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: The reason why I thought we can implement the API side changes as a later step was because I was wondering if it would increase the review surface of this KEP. But I would be happy to include those changes here as well. 

On the use cases for that, here are some things that came to my mind:
- exec'ing a specific binary in a distroless container with no shell or no env binary, while also having to inject some additional environment variables like `LOG_LEVEL=DEBUG`
- Injecting dynamic credentials for the exec'd process session based on the incoming user's identity. 
- Can enforce custom runtime security policies based on the identity. For example, if a user is not allowed to read a certain file inside an exec session, you can inject the policy context into the session, from where eBPF tools can contextually and selectively block those calls for that user alone. (I feel that this can be very powerful)
- And of course, can be used to inject TraceID/ Audit ID for  attribution of commands run inside.

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: We haven't thought about the logging part yet. Can you please help with a little bit more context here?
- Are there separate logs for container exec sessions?
- Are you suggesting something like redacting known environment variables from the logs? Similar to how Github Actions does?

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: The intention behind having these statements was simply to highlight the semantic boundaries. The container runtime should pass on the variables as is, without any interpretation or modification.

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: What typically happens in the container runtime implementation is it clones the main process, modifies the `cmd` field alone to the specified command and passes it on to `runc/crun` as a process.json file. So the environment variables present in the main process are also present in the process to be exec'd. 
We will handle collisions by giving more prirority to the environment variables coming from kubelet via the `ExecRequest`. Meaning, if there is a same environment variable key in the main process, it will be overwritten by this new contract.

sigtrapd: I haven't really thought about this direction yet. But this leads to many interesting questions. 
- Where would you resolve the values for expansions? Is it the Kubelet or API server? Maybe it would depend on what kind of information is needed to resolve the variable.
- The existing environment vars are cloned already. So do we need to allow expansion at that level?
- And as for protection and precedence, I have answered that in the comment just below. Can you please clarify if you were specifically looking for something else?

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: There are two proto definitions. One is `staging/src/k8s.io/api/core/v1/generated.proto` which captures all API object structures. Here, the value field for an environment variable is of type `string` as highlighted in the KEP you have shared. The other proto definition is `staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto` which is the definition for the CRI calls. Here, The `CreateContainerRequest` uses `KeyValue` in which the `value` is of type `bytes`. We will be adopting the same data types that are used while creating a container. If I'm not wrong, the Kubelet would type cast from string to bytes before wiring it down to the CR.

SergeyKanzhelev: as long as it it well aligned, I am fine. Just make sure to learn what is the latest in addressing it or even help with fixing it.

keps/sig-node/6090-exec-env-injection/README.md
SergeyKanzhelev: covering scenarios as part of the API change will help answer some question on whether some reuse of pod-spec defined env vars is needed, whether expansion of values is needed, etc. If the scenarios is just a constant key and value injection, than it is fine. I would suggest to start with some "easiest" use case so the KEP will have logic end-to-end.

keps/sig-node/6090-exec-env-injection/README.md
SergeyKanzhelev: it's easiest to not expand. But it really depends on whether you have use cases for expansion or some dynamic substitution.

SergeyKanzhelev: is the fact that "main" env vars are inherited documented somewhere? Should we have a critest to make it an official conformance requirement? Or it is implementation specific? If it is implementation specific, would the override order also be an implementation specific? This is not a complicated questions on a surface - logically it is clear what should happen. The questions are mostly: can it introduce some security issue (PATH override) and what is the guarantees we should place in conformance critests and what is an implementation details

keps/sig-node/6090-exec-env-injection/README.md
SergeyKanzhelev: how can we test (in terms of conformance) that the container runtime doesn't behave differently when specific env var is present? Also if container runtime injects it's own env var for some reason, and does not want CRI API overriding it, what's the issue with it?

keps/sig-node/6090-exec-env-injection/README.md
SergeyKanzhelev: I remember, but cannot find an issue when all stdout logs from remote executions were forwarded to container logs. So if secret was needed in that session, it was exposed to logs. If we support env vars setting like this, I do not think they will show up in logs (unless echo'd explicitly). So this is a great improvement for such scenarios that is worth mentioning as a use case.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,927 @@
...
+### Environment Variable Precedence
+
+The final `envp[]` passed to `execve()` is constructed in the following order, where earlier entries take precedence on collision (first-match wins via `getenv()`):
+
+```
+final_envp[] = kubelet_envs + user_envs + container_envs
+```
+
+Highest to lowest precedence:
+1. Kubelet-injected envs
```
mikebrow, line 324:
```suggestion
A. Container Runtime adjustments to the ENVs (i.e. from NRI when exec is extended to the NRI, and here is an example in the runtime:

https://github.com/containerd/containerd/blob/main/internal/cri/server/container_execsync.go#L140-L145
 
1. Kubelet-injected envs
```

keps/sig-node/6090-exec-env-injection/README.md
mikebrow: nod... a couple use cases, e.g. cases for exec probe.., an agent use case case, kubectl exec case

mikebrow (commented):
couple early comments on this draft :-)

keps/sig-node/6090-exec-env-injection/README.md
sigtrapd: Agreed. I am thinking of leaving the expansion for a future enhancement if there are usecases.

sigtrapd: The main env vars being inherited is a convergent behavior across various implementations, but as far as I have seen, there's no documentation of this. So yes, the override order would also be implementation specific.

@mikebrow , Can you please help us confirm if this is a documented behavior, or it just happens to be implemented similarly?

And on the security front, anyone with `pods/exec` permissions can still inject env vars into the process using the `env` binary except for the fact that it is not reliable. So I feel that this is something that needs to be gated at the RBAC level. On top of it, since this is coming through the API server, such injections can be prevented with policy engines like Gatekeeper or Kyverno.

keps/sig-node/6090-exec-env-injection/kep.yaml
```diff
@@ -0,0 +1,51 @@
...
+participating-sigs:
+  - sig-api-machinery
+
+status: provisional
+creation-date: 2026-05-18
+reviewers:
+  - "@haircommander"
+
+approvers:
+  - TBD
```
soltysh, line 18: This needs to be filled in. I see @SergeyKanzhelev already looked at it, so I believe he'll one of approvers.

keps/sig-node/6090-exec-env-injection/kep.yaml
```diff
@@ -0,0 +1,51 @@
...
+authors:
+  - "@nispriha"
+  - "@sigtrapd"
+  - "@ngopalak-redhat"
+
+owning-sig: sig-node
+participating-sigs:
+  - sig-api-machinery
+
+status: provisional
```
soltysh, line 12: implementable?

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
```
soltysh, line 157: Nit: make sure to check appropriate boxes above.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+### Risks and Mitigations
+
+<!--
```
soltysh, line 294: This section is required.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Environment Variable Precedence
+
+The final `envp[]` passed to `execve()` is constructed in the following order, where earlier entries take precedence on collision (first-match wins via `getenv()`):
```
soltysh, line 317: That seems exactly opposite of what I would normally expect. In all other cases, the closest envvar to the process is taken into account. Why reverting this order?

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+The goal is to ensure that we don't accept enhancements with inadequate testing.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[ ] I/we understand the owners of the involved components may require updates to
```
soltysh, line 350: Check the box.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+This can be done with:
+- permalinks to the GitHub source code
+- links to the periodic job (typically a job owned by the SIG responsible for the feature), filtered by the test name
+- a search in the Kubernetes bug triage tool (https://storage.googleapis.com/k8s-triage/index.html)
+
+We expect no non-infra related flakes in the last month as a GA graduation criteria.
+If e2e tests are not necessary or useful, explain why.
+-->
+
+- [test name](https://github.com/kubernetes/kubernetes/blob/2334b8469e1983c525c0c6382125710093a25883/test/e2e/...): [SIG ...](https://testgrid.k8s.io/sig-...?include-filter-by-regex=MyCoolFeature), [triage search](https://storage.googleapis.com/k8s-triage/index.html?test=MyCoolFeature)
```
soltysh, line 427: Tests are required for alpha, please make sure to list the planned ones. For units, list the affected packages and their current code coverage.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+- Gather feedback from developers and surveys
+
+#### GA
+
+- Stable for at least two releases
+- No major issues reported
+
+### Upgrade / Downgrade Strategy
+
+<!--
```
soltysh, line 521: This is required.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+components? What are the guarantees? Make sure this is in the test plan.
+
+Consider the following in developing a version skew strategy for this
+enhancement:
+- Does this enhancement involve coordinating behavior in the control plane and nodes?
+- How does an n-3 kubelet or kube-proxy without this feature available behave when this feature is used?
+- How does an n-1 kube-controller-manager or kube-scheduler without this feature available behave when this feature is used?
+- Will any other components on the node change? For example, changes to CSI,
+  CRI or CNI may require updating that component before the kubelet.
+-->
```
soltysh, line 546: This is required.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+-->
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+## Implementation History
+
+<!--
```
soltysh, line 897: This section should be filled in.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
```
soltysh, line 911: Same here.

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+
+#### Alpha
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- No major bugs reported during alpha
+- Gather feedback from developers and surveys
```
soltysh, line 512: Since you're changing CRI, I'd expect that one of the beta promotion criteria explicitly call out 2 major CRI implementing the new functionality (CRIO and containerd).

keps/sig-node/6090-exec-env-injection/README.md
```diff
@@ -0,0 +1,928 @@
...
+implementation difficulties, etc.).
+-->
+
+### Dependencies
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### Does this feature depend on any specific services running in the cluster?
```
soltysh, line 758: Since you're explicitly modifying CRI, I'd expect this section be filled in already for alpha.

keps/sig-node/6090-exec-env-injection/README.md
mikebrow: docs on this behavior: https://docs.docker.com/reference/cli/docker/container/exec/#env

In my opinion, my memory of the initial docker integrations, docker shim, cri-o, and cri-containerd were attempting to integrate the kubelet with the docker api, then later CRI api to run docker containers (then OCI containers, which begat from docker containers).  The CRI plugins should follow the behavior as specified therein:  `exec command inherits the environment variables that are set at the time the container is created.` and overrides/additions are allowed by user request. 

> Should we have a critest to make it an official conformance requirement?

yes. 

> can it introduce some security issue (PATH override) 

yes.

> and what is the guarantees we should place in conformance critests and what is an implementation details

nod needs to discussed, documented, and tested .. one way (if allowed to override/amend env for exec) or the other way (no overriding)

SergeyKanzhelev: make sure to add a test case that there is NO expansion indeed

sigtrapd:
/hold

This KEP is temporarily being placed on hold as we are working on merging the two KEPS https://github.com/kubernetes/enhancements/pull/6091 and https://github.com/kubernetes/enhancements/pull/6036 into one (merging into https://github.com/kubernetes/enhancements/pull/6036). Feel free to take a look at the KEPs as the high level problem and approach are going to remain the same. We are just working on refactoring the explanation to better answer some questions that were raised as part of the reviews.

Issue #5945, opened 2026-03-05
DRA: Optional Node Operations


jiefeng-xu:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add an explicit DRA opt-out for node-local prepare/unprepare so kubelet can skip plugin calls for allocations that do not require node preparation.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/5945-dra-optional-node-preparation/README.md
- Discussion Link: WG Device Management meeting
- PRs by stage and milestone:
  - [x] Alpha - v1.37
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6104
    - [x] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/139933
    - [x] Docs (`k/website`) update PR(s): https://github.com/kubernetes/website/pull/56258

pohly:
@jiefeng-xu: can you change the PR description such that it uses the normal template for an enhancement issue?

You have too many details there. Those belong into the KEP README.md, not the issue or PR description.

mortent:
@jiefeng-xu Are you planning to work on this feature for 1.37? We want this to land in 1.37, so we might look at finding someone else to work on it if you don't plan to.

jiefeng-xu:
@pohly , @mortent , thank you for checking. It is a bit challenging for me with the 1.37 schedule. I would not let it be blocked by me. Feel free to find someone else to work on it.

mortent:
Thanks for the quick response @jiefeng-xu. @troychiu will take a look at this, so he will probably create a new KEP proposal that supersedes your proposal in https://github.com/kubernetes/enhancements/pull/5946. If you have time to provide input on the design, that will be very helpful.

/assign @troychiu

johnbelamaric:
"Node Lifecycle" means something different to most SIG Node readers. This is actually "Node-Local Resource Lifecycle" but that seems too long. What about "DRA: Controller-Only Drivers"?

troychiu:
Would something straightforward like "DRA Optional Node Preparation" be okay?

johnbelamaric:
@troychiu please let me know when you have a KEP PR ready to review

troychiu:
@johnbelamaric I have created the KEP PR. Please take a look. Thank you!

lasomethingsomething:
[...]

PR #6104, opened 2026-05-22
KEP-5945: DRA Optional Node Preparation

troychiu:
- One-line PR description: This KEP introduces Optional Node Preparation to Dynamic Resource Allocation (DRA), allowing a DeviceClass to declare that node preparation is not required.

- Issue link: https://github.com/kubernetes/enhancements/issues/5945

- Other comments: Gemini helped create this PR

### DIFF
--- a/keps/sig-node/5945-dra-optional-node-preparation/README.md
+++ b/keps/sig-node/5945-dra-optional-node-preparation/README.md
@@ -423,17 +423,81 @@ driver based on the individual skip flags, allowing mixed-mode execution.
     - Delete the Pod.
     - Assert that the driver's `NodeUnprepareResources` was **not** called.
 
+###### Scenario 3: Upgrade / Downgrade and Feature Gate Rollback
+This scenario validates that the system behaves correctly during a rolling
+upgrade or downgrade/rollback of the feature gate.
+
+- **Test Case 3.1: Rolling Upgrade (N-1 to N)**:
+  - **Setup**: Start with a cluster running version N-1 (feature gate disabled).
+    Deploy a DRA driver.
+  - **Action 1 (Control Plane Upgrade)**: Upgrade the control plane to version N
+    (feature gate enabled).
+    - **Assertions**:
+      - If we deploy a new workload using a control-plane-only driver (no
+        node-local components):
+        - The Pod gets stuck in `ContainerCreating` on the N-1 kubelet, as the
+          N-1 kubelet attempts to contact the non-existent node-local driver.
+  - **Action 2 (Kubelet Upgrade)**: Upgrade the kubelets to version N.
+    - **Assertions**:
+      - Verify that the control-plane-only workload (which was stuck in
+        `ContainerCreating`) now successfully bypasses node preparation,
+        transitions to `Running`, and runs successfully.
+      - Verify that deleting the control-plane-only workload completes
+        immediately without trying to contact a node-local driver.
+- **Test Case 3.2: Feature Gate Rollback / Downgrade (N to N-1)**:
+  - **Setup**: Start with a cluster running version N (feature gate enabled).
+    Deploy a standard DRA driver and a workload using `SkipNodePrepare: true`
+    and `SkipNodeUnprepare: true`.
+  - **Action 1 (Control Plane Downgrade)**: Downgrade the control plane to N-1
+    (feature gate disabled).
+    - **Assertions**:
+      - The API server's ratcheting validation allows the existing
+        `ResourceSlice` objects to remain valid and not be rejected on unrelated
+        updates.
+      - The running workload on the N kubelet continues to run without
+        interruption.
+  - **Action 2 (Kubelet Downgrade)**: Downgrade the kubelets to N-1 (feature
+    gate disabled).
+    - **Assertions**:
+      - Verify that the kubelet behaves according to the
+        [Disabled Feature Gate Behavior](#disabled-feature-gate-behavior) section
+        upon workload deletion.
+- **Test Case 3.3: Upgrade -> Downgrade -> Upgrade (N-1 -> N -> N-1 -> N)**:
+  - **Setup**: Start with a cluster running version N-1 (feature gate disabled).
+    Deploy a standard DRA driver.
+  - **Action 1 (Upgrade)**: Upgrade the cluster to version N (feature gate
+    enabled).
+    - Deploy a workload using a driver configured with `SkipNodePrepare: true`
+      and `SkipNodeUnprepare: true`.
+    - Assert that the workload runs successfully.
+  - **Action 2 (Downgrade)**: Downgrade the cluster to version N-1 (feature gate
+    disabled).
+    - **Assertions**:
+      - Assert that the API server's ratcheting validation allows the existing
+        `ResourceSlice` (which has `SkipNodePrepare: true` and `SkipNodeUnprepare:
+        true`) to remain valid and unmodified.
+      - Delete the workload and verify that the kubelet behaves according to the
+        [Disabled Feature Gate Behavior](#disabled-feature-gate-behavior) section.
+  - **Action 3 (Upgrade Again)**: Upgrade the cluster back to version N (feature
+    gate enabled).
+    - Deploy a new workload using the same driver.
+    - Assert that the new fields are respected, and the workload runs
+      successfully.
+    - Assert that any pre-existing resource slices that survived the downgrade
+      cycle continue to function correctly with the re-enabled feature gate.
+
 ### Graduation Criteria
 
 #### Alpha
 
 - Feature implemented behind the `DRAOptionalNodePreparation` feature flag (off
   by default).
-- Full unit, integration, and E2E test suites implemented and green.
+- Full unit and basic E2E test suites (Scenario 1 & 2) implemented and green.
 
 #### Beta
 
 - Enable the feature gate by default.
+- E2E upgrade/downgrade and rollback test suites (Scenario 3) implemented and green.
 - Gather real-world feedback from developers and vendors deploying
   controller-managed DRA drivers.
 - Ensure no regressions or performance issues are observed in large clusters.
@@ -589,7 +653,10 @@ is attempting node preparation and blocking/failing due to missing node drivers.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-To be completed at Beta stage.
+This will be tested as part of the Beta graduation criteria using the
+upgrade/downgrade E2E test plan. See [Scenario 3: Upgrade / Downgrade and
+Feature Gate Rollback](#scenario-3-upgrade--downgrade-and-feature-gate-rollback)
+in the Test Plan for the detailed test cases.
 
 ###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
 
--- a/keps/sig-node/5945-dra-optional-node-preparation/README.md
+++ b/keps/sig-node/5945-dra-optional-node-preparation/README.md
@@ -246,6 +246,28 @@ this architectural difference.
    }
    ```
 
+#### API Server Handling & Ratcheting Validation
+
+To ensure fail-fast feedback and prevent workloads from entering a broken state,
+the `kube-apiserver` validates the new fields against the state of the
+`DRAOptionalNodePreparation` feature gate.
+
+To support safe cluster rollbacks and downgrades, the API server implements
+**Ratcheting Validation** (in accordance with the [Kubernetes API Changes
+Guide](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md#ratcheting-validation)):
+
+* **When the feature gate is disabled**:
+  * **New Resources (POST)**: Any attempt to create a `ResourceSlice` or
+    allocate a `ResourceClaim` (via its status) with `SkipNodePrepare` or
+    `SkipNodeUnprepare` set to `true` is **rejected** with a validation error.
+  * **Existing Resources (PUT)**: The API server allows updates to existing
+    resources that already have these fields set to `true` (e.g., persisted
+    while the feature gate was enabled before a downgrade), provided the update
+    does not attempt to newly enable or modify these fields. Any transition of
+    these fields from `nil`/`false` to `true` is **rejected**.
+* **When the feature gate is enabled**:
+  * The fields are validated and persisted normally.
+
 ### Allocator Changes
 
 During scheduling, the structured parameters allocator resolves `ResourceSlices`
@@ -309,6 +331,12 @@ None.
 
 ##### Unit tests
 
+- **API Server Validation Unit Tests**: In `pkg/apis/resource/validation/validation_test.go`:
+  - Verify the four ratcheting validation cases when the feature gate is disabled:
+    - New valid (fields `nil`/`false`) -> Succeeds.
+    - New invalid (fields `true`) -> Fails.
+    - Old valid (fields `nil`/`false`) -> Succeeds.
+    - Old invalid (fields `true` in old object, unchanged in new) -> Succeeds.
 - **Allocator Unit Tests**: In
   `staging/src/k8s.io/dynamic-resource-allocation/structured/allocator_test.go`:
   - Verify that `SkipNodePrepare` and `SkipNodeUnprepare` in `ResourceSliceSpec`
--- a/keps/sig-node/5945-dra-optional-node-preparation/README.md
+++ b/keps/sig-node/5945-dra-optional-node-preparation/README.md
@@ -153,8 +153,10 @@ We propose adding a boolean field `SkipNodeOperations` to
 
 #### Deploying controller-managed resources without node-local drivers
 As a cluster administrator or vendor using a central driver controller, I want
-to offer resources (e.g., cluster-wide shared resource pools or logically
-partitioned network services) where availability is discovered and published as
+to offer resources (e.g., cluster-wide shared resource pools, logically
+partitioned network services, or pure control-plane gating drivers like
+[dra-driver-image-configurator](https://github.com/gke-labs/dra-drivers/tree/main/dra-driver-image-configurator))
+where availability is discovered and published as
 `ResourceSlice` resources centrally by the controller. Because the devices
 require no node-local plumbing or mount operations on worker nodes, there is no
 node driver deployed. The controller publishes these resources with
@@ -418,6 +420,7 @@ deploy any node-local components.
   - **Workload**: Deploy a Pod referencing this resource.
   - **Assertions**:
     - The Pod reaches the `Running` phase successfully.
+    - The allocated device can be accessed.
     - No `FailedPrepareDynamicResources` warnings are posted to the Pod events.
     - Pod deletion completes cleanly and immediately (does not hang in
       `Terminating` waiting for unprepare).
--- a/keps/sig-node/5945-dra-optional-node-preparation/README.md
+++ b/keps/sig-node/5945-dra-optional-node-preparation/README.md
@@ -13,6 +13,7 @@
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [API Changes](#api-changes)
+    - [API Server Handling and Ratcheting Validation](#api-server-handling-and-ratcheting-validation)
   - [Allocator Changes](#allocator-changes)
   - [Kubelet Changes](#kubelet-changes)
   - [Test Plan](#test-plan)
@@ -39,6 +40,7 @@
   - [Alternative 1: DeviceClass-level configuration](#alternative-1-deviceclass-level-configuration)
   - [Alternative 2: Claim-level declaration](#alternative-2-claim-level-declaration)
   - [Alternative 3: Kubelet Auto-Discovery / gRPC probe with timeout](#alternative-3-kubelet-auto-discovery--grpc-probe-with-timeout)
+  - [Alternative 4: Centralized catch-all no-op plugin](#alternative-4-centralized-catch-all-no-op-plugin)
 <!-- /toc -->
 
 ## Release Signoff Checklist
@@ -138,11 +140,18 @@ We propose adding boolean fields `SkipNodePrepare` and `SkipNodeUnprepare` to
 1. **API Definition**: The driver/controller publisher sets `SkipNodePrepare:
    true` and/or `SkipNodeUnprepare: true` in `ResourceSlice` resources if the
    published devices do not require node-local setup or cleanup.
-2. **Control Plane Resolution**: The allocator/scheduler resolves the referenced
+2. **API Validation**: To prevent pods from getting stuck during termination, we enforce a
+   validation rule: **If `SkipNodeUnprepare` is `false/nil`, `SkipNodePrepare` must
+   not be `true`**. The combination of skipping preparation but requiring
+   unpreparation (`SkipNodePrepare: true` and `SkipNodeUnprepare: false/nil`) is
+   invalid and will be rejected. This configuration is fragile because the kubelet
+   wouldn't verify the driver's presence at startup, leading to pods getting stuck
+   in `Terminating` if the driver is missing during deletion.
+3. **Control Plane Resolution**: The allocator/scheduler resolves the referenced
    `ResourceSlice` during allocation, and copies this configuration into
    `ResourceClaim.Status.Allocation.Devices.Results[i].SkipNodePrepare` and
    `SkipNodeUnprepare`.
-3. **Node Execution**: The kubelet reads these fields from the `ResourceClaim`'s
+4. **Node Execution**: The kubelet reads these fields from the `ResourceClaim`'s
    allocation results. If all allocated devices for a given driver within a
    claim set `SkipNodePrepare: true` (or `SkipNodeUnprepare: true`), the kubelet
    bypasses the corresponding gRPC call to the node-local resource driver.
@@ -246,7 +255,18 @@ this architectural difference.
    }
    ```
 
-#### API Server Handling & Ratcheting Validation
+#### API Server Handling and Ratcheting Validation
+
+The API server enforces the following co-dependency validation rule on both
+`ResourceSlice` and `ResourceClaim` (via its status):
+*   **Co-dependency Validation**: If `SkipNodePrepare` is `true`,
+    `SkipNodeUnprepare` **must** be `true`.
+    *   If `SkipNodePrepare` is `true` and `SkipNodeUnprepare` is `false` (or
+        `nil`/omitted), the resource is **rejected** with a validation error
+        (e.g., `forbidden: skipNodePrepare cannot be true when skipNodeUnprepare
+        is false`).
+    *   This ensures we fail early and prevent workloads from entering a state
+        where they cannot terminate.
 
 To ensure fail-fast feedback and prevent workloads from entering a broken state,
 the `kube-apiserver` validates the new fields against the state of the
@@ -337,6 +357,9 @@ None.
     - New invalid (fields `true`) -> Fails.
     - Old valid (fields `nil`/`false`) -> Succeeds.
     - Old invalid (fields `true` in old object, unchanged in new) -> Succeeds.
+  - Verify co-dependency validation:
+    - `SkipNodePrepare: true` and `SkipNodeUnprepare: false/nil` -> Fails.
+    - `SkipNodePrepare: true` and `SkipNodeUnprepare: true` -> Succeeds.
 - **Allocator Unit Tests**: In
   `staging/src/k8s.io/dynamic-resource-allocation/structured/allocator_test.go`:
   - Verify that `SkipNodePrepare` and `SkipNodeUnprepare` in `ResourceSliceSpec`
@@ -381,19 +404,17 @@ deploy any node-local components.
     - No `FailedPrepareDynamicResources` warnings are posted to the Pod events.
     - Pod deletion completes cleanly and immediately (does not hang in
       `Terminating` waiting for unprepare).
-- Test Case 1.2: Missing node component failure (`SkipNodePrepare: false` or
-  `SkipNodeUnprepare: false`)
-  - **API Configuration**: Publish `ResourceSlices` with at least one skip field
-    set to `false` (or omitted).
+- Test Case 1.2: Missing node component failure (`SkipNodePrepare: false`)
+  - **API Configuration**: Publish `ResourceSlices` with `SkipNodePrepare` set
+    to `false` (or omitted).
   - **Workload**: Deploy a Pod referencing this resource.
   - **Assertions**:
-    - If `SkipNodePrepare` is `false`: The Pod gets stuck in `ContainerCreating`
+    - The Pod gets stuck in `ContainerCreating`
       with `FailedPrepareDynamicResources` errors because the kubelet tries to
       contact the non-existent node driver.
-    - If `SkipNodePrepare` is `true` but `SkipNodeUnprepare` is `false`: The Pod
-      runs successfully, but upon deletion, it gets stuck in `Terminating` phase
-      as the kubelet fails to call the non-existent node driver for
-      unpreparation.
+    - Note: The configuration `SkipNodePrepare: true` and `SkipNodeUnprepare:
+      false` is rejected at the API level and cannot be tested with a running
+      Pod.
 
 ###### Scenario 2: Driver with node-local components (Standard Driver)
 This scenario validates that the kubelet selectively invokes the node-local
@@ -401,18 +422,7 @@ driver based on the individual skip flags, allowing mixed-mode execution.
 
 - **Setup**: Deploy a standard DRA test driver that includes node-local gRPC
   components.
-- Test Case 2.1: Skip Prepare Only (`SkipNodePrepare: true`,
-  `SkipNodeUnprepare: false`)
-  - **API Configuration**: Publish `ResourceSlices` with `SkipNodePrepare: true`
-    and `SkipNodeUnprepare: false`.
-  - **Workload**: Deploy a Pod.
-  - **Assertions**:
-    - The Pod reaches the `Running` phase.
-    - Assert that the driver's
-      `NodePrepareResources` was **not** called.
-    - Delete the Pod.
-    - Assert that the driver's `NodeUnprepareResources` **was** called.
-- Test Case 2.2: Skip Unprepare Only (`SkipNodePrepare: false`,
+- Test Case 2.1: Skip Unprepare Only (`SkipNodePrepare: false`,
   `SkipNodeUnprepare: true`)
   - **API Configuration**: Publish `ResourceSlices` with `SkipNodePrepare:
     false` and `SkipNodeUnprepare: true`.
@@ -843,7 +853,7 @@ is not needed and starts the pod.
   containers launch before their local devices are fully prepared. Explicit
   declaration via the API is highly deterministic and secure.
 
-### Alternative 4: Centralized "catch-all" no-op plugin
+### Alternative 4: Centralized catch-all no-op plugin
 Deploy a generic, "no-op" DRA driver (such as
 [dra-driver-noop](https://github.com/gke-labs/dra-drivers/tree/main/dra-driver-noop))
 configured centrally to register under specific DRA driver names and handle the
--- a/keps/sig-node/5945-dra-optional-node-preparation/README.md
+++ b/keps/sig-node/5945-dra-optional-node-preparation/README.md
@@ -1,0 +1,587 @@
+# KEP-5945: DRA Optional Node Preparation
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Deploying controller-managed resources without node-local drivers](#deploying-controller-managed-resources-without-node-local-drivers)
+    - [Flexibility for drivers with mixed preparation requirements](#flexibility-for-drivers-with-mixed-preparation-requirements)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Allocator Changes](#allocator-changes)
+  - [Kubelet Changes](#kubelet-changes)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: DeviceClass-level configuration](#alternative-1-deviceclass-level-configuration)
+  - [Alternative 2: Claim-level declaration](#alternative-2-claim-level-declaration)
+  - [Alternative 3: Kubelet Auto-Discovery / gRPC probe with timeout](#alternative-3-kubelet-auto-discovery--grpc-probe-with-timeout)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+[Conformance Tests]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md
+[all GA Endpoints]: https://github.com/kubernetes/community/pull/1806
+
+## Summary
+
+This KEP introduces **Optional Node Preparation** to Dynamic Resource Allocation
+(DRA), allowing resource drivers to declare that node preparation and/or node
+unpreparation is not required for their devices. Currently, the kubelet assumes
+it must always coordinate with a node-local DRA driver via gRPC to prepare
+allocated devices before container start (`NodePrepareResources`), and to
+unprepare them during pod termination (`NodeUnprepareResources`).
+
+In some cases, node preparation or cleanup is a pure no-op. Requiring it forces
+administrators and vendors to deploy and maintain empty node-local drivers on
+every node, which introduces unnecessary operational complexity and risk.
+
+By introducing `SkipNodePrepare` and `SkipNodeUnprepare` fields to the
+`ResourceSliceSpec` and propagating them to the device allocation results at
+scheduling time, the kubelet can safely skip driver lookup and gRPC calls for
+devices that do not require these node-local actions.
+
+## Motivation
+
+In Dynamic Resource Allocation (DRA), the kubelet coordinates with a node-local
+driver via gRPC to prepare allocated devices before container start
+(`NodePrepareResources`) and to unprepare them upon pod termination
+(`NodeUnprepareResources`). For node-local accelerators (such as PCIe GPUs or
+local FPGAs), this node-level setup is critical to check device health,
+partition memory, and configure mount paths.
+
+However, there is an emerging class of resources whose lifecycles are managed
+entirely in the control plane and published centrally by a controller as
+`ResourceSlice` objects. These resources require absolutely zero node-local
+setup. Under the current architecture, the kubelet still assumes a node-local
+driver exists, forcing administrators to deploy and maintain wasteful "no-op"
+node DaemonSets just to answer gRPC calls with empty success responses. If one
+of these dummy helper plugins crashes or is missing, the kubelet's unprepare
+hook fails and retries indefinitely, leaving terminating pods permanently "stuck
+in Terminating" and blocking cluster upgrades and node drains.
+
+To resolve this architectural mismatch and accommodate modern deployments, we
+need a way for resource drivers to declare that node preparation and/or cleanup
+can be skipped. Bypassing these gRPC hooks directly at the `ResourceSlice` level
+allows vendors to deploy central-only controllers with zero worker node
+footprints. It also provides the flexibility to support mixed hardware
+topologies—where a single driver manages some devices requiring node-level
+preparation and others that do not—without splitting the driver or forcing
+unnecessary footprints onto worker nodes.
+
+### Goals
+
+- Allow resource drivers to declare that node-local preparation and/or
+  unpreparation is optional for devices.
+- Propagate this configuration from the `ResourceSlice` to the final allocated
+  `ResourceClaim.Status.Allocation` result.
+- Update the kubelet to skip driver lookup and gRPC preparation/unpreparation
+  steps when node preparation or clean-up is explicitly configured as skipped.
+- Maintain backward compatibility: by default, all existing DRA drivers must
+  continue to require node-local preparation and unpreparation.
+
+### Non-Goals
+
+- Eliminate node-local preparation entirely.
+- Enable users to override this infrastructure requirement at the individual
+  `ResourceClaimSpec` level.
+
+## Proposal
+
+We propose adding boolean fields `SkipNodePrepare` and `SkipNodeUnprepare` to
+`ResourceSliceSpec` and `DeviceRequestAllocationResult`.
+
+1. **API Definition**: The driver/controller publisher sets `SkipNodePrepare:
+   true` and/or `SkipNodeUnprepare: true` in `ResourceSlice` resources if the
+   published devices do not require node-local setup or cleanup.
+2. **Control Plane Resolution**: The allocator/scheduler resolves the referenced
+   `ResourceSlice` during allocation, and copies this configuration into
+   `ResourceClaim.Status.Allocation.Devices.Results[i].SkipNodePrepare` and
+   `SkipNodeUnprepare`.
+3. **Node Execution**: The kubelet reads these fields from the `ResourceClaim`'s
+   allocation results. If all allocated devices for a given driver within a
+   claim set `SkipNodePrepare: true` (or `SkipNodeUnprepare: true`), the kubelet
+   bypasses the corresponding gRPC call to the node-local resource driver.
+
+### User Stories
+
+#### Deploying controller-managed resources without node-local drivers
+As a cluster administrator or vendor using a central driver controller, I want
+to offer resources (e.g., cluster-wide shared resource pools or logically
+partitioned network services) where availability is discovered and published as
+`ResourceSlice` resources centrally by the controller. Because the devices
+require no node-local plumbing or mount operations on worker nodes, there is no
+node driver deployed. The controller publishes these resources with
+`SkipNodePrepare: true` and `SkipNodeUnprepare: true`. When users request these
+devices, the kubelet launches the pods immediately and cleanly, without
+complaining about missing node-local drivers, and without requiring any node
+driver DaemonSet to be present in the cluster.
+
+#### Flexibility for drivers with mixed preparation requirements
+As a hardware vendor, I develop a DRA driver that supports multiple classes of
+devices. Some devices of our hardware pool require node-local interface setups
+(e.g., local host bindings), while others are fully managed and provisioned
+out-of-band in the control plane. By setting `SkipNodePrepare: false` (default)
+on node-local `ResourceSlices` and `SkipNodePrepare: true` on
+control-plane-managed `ResourceSlices`, the kubelet can dynamically determine
+whether to run node prep based on which device was selected by the scheduler. I
+do not need to split my driver into multiple separate drivers just to handle
+this architectural difference.
+
+### Risks and Mitigations
+
+- **Dynamic ResourceSlice Changes**: An administrator or controller could update
+  `SkipNodePrepare` or `SkipNodeUnprepare` in a `ResourceSlice` while claims are
+  already allocated.
+  - *Mitigation*: The allocation result is written and frozen into the
+    `ResourceClaim` status at scheduling time. The kubelet reads the values from
+    the `AllocationResult` of the claim itself, not from the `ResourceSlice`
+    directly. This ensures that the behavior of currently running pods is
+    consistent and matches the configuration at the time they were scheduled.
+- **Backward Compatibility & Out-of-Tree / Custom Allocators**: Old scheduler
+  clients or out-of-tree custom driver controllers/allocators might write
+  allocation results without setting the new `SkipNodePrepare` and
+  `SkipNodeUnprepare` fields.
+  - *Mitigation*: The pointer fields default to `nil` when omitted or sent by
+    older versions, which is treated as `false` (not skipped). A value of
+    `false` (explicit or default nil) indicates that the kubelet **should not
+    skip** the action, meaning it defaults to executing node preparation and
+    clean-up as normal. This guarantees 100% backward compatibility with all
+    existing schedulers, custom controllers, and running workloads.
+
+## Design Details
+### API Changes
+
+1. **`ResourceSliceSpec`** in `pkg/apis/resource`:
+   ```go
+   type ResourceSliceSpec struct {
+       ...
+       // SkipNodePrepare indicates that node-local resource preparation (NodePrepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodePrepare *bool `json:"skipNodePrepare,omitempty" protobuf:"varint,5,opt,name=skipNodePrepare"`
+
+       // SkipNodeUnprepare indicates that node-local resource cleanup (NodeUnprepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodeUnprepare *bool `json:"skipNodeUnprepare,omitempty" protobuf:"varint,6,opt,name=skipNodeUnprepare"`
+   }
+   ```
+
+2. **`DeviceRequestAllocationResult`** in `pkg/apis/resource`:
+   ```go
+   type DeviceRequestAllocationResult struct {
+       ...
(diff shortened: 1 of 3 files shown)


keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,412 @@
...
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+[Conformance Tests]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md
+[all GA Endpoints]: https://github.com/kubernetes/community/pull/1806
+
+## Summary
+
+This KEP introduces **Optional Node Preparation** to Dynamic Resource Allocation (DRA), allowing a `DeviceClass` to declare that node preparation is not required. Currently, the kubelet assumes it must always coordinate with a node-local DRA driver via gRPC to prepare allocated devices before container start (`NodePrepareResources`), and to unprepare them during pod termination (`NodeUnprepareResources`).
```
johnbelamaric: Why `DeviceClass`? I think it probably should be the `Device` (ie, the `ResourceSlice`). The driver author should know if it needs preparation. `DeviceClass` is for the cluster administrator, who could easily make a mistake with this being at that level.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,412 @@
...
+
+In some cases, node preparation is a pure no-op. Requiring it forces administrators and vendors to deploy and maintain empty node-local drivers on every node, which introduces unnecessary operational complexity and risk.
+
+By introducing a `RequiresNodePreparation` field to the cluster-scoped `DeviceClass` and propagating it to the device allocation results at scheduling time, the kubelet can safely skip driver lookup and gRPC calls for drivers and device classes that do not require node-local actions.
```
johnbelamaric: I think the Device (or maybe the ResourceSlice with possible override at the Device level?) should have two fields: `NoNodePrepare` and `NoNodeUnprepare`. This allows us to skip unprepare even if prepare is needed. It can eliminate pods stuck in terminating if the driver has gone away, and the devices do not need any unpreparation (which I suspect is common; the containers with the mounts going away is sufficient in many cases).

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,412 @@
...
+
+## Proposal
+
+We propose adding a boolean `RequiresNodePreparation` field to both `DeviceClassSpec` and `DeviceRequestAllocationResult`.
+
+1. **API Definition**: The cluster administrator sets `RequiresNodePreparation: false` in a `DeviceClass` if the devices in that class do not require node-local preparation.
+2. **Control Plane Resolution**: The allocator/scheduler resolves the referenced `DeviceClass` during allocation, and copies/translates this preference into `ResourceClaim.Status.Allocation.Devices.Results[i].RequiresNodePreparation`.
+3. **Node Execution**: The kubelet reads the `RequiresNodePreparation` field from the `ResourceClaim`'s allocation results. If all allocated devices for a given driver within a claim explicitly set this field to `false`, the kubelet bypasses resource preparation and cleanup for that claim (completely bypassing driver registration lookup and gRPC calls) during pod admission and deletion.
+
+### User Stories
```
johnbelamaric, line 152: also the "stuck in terminating" issue - maybe in motivation rather than use cases, though

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,412 @@
...
+  - *Mitigation*: The allocation result is written and frozen into the `ResourceClaim` status at scheduling time. The kubelet reads the value from the `AllocationResult` of the claim itself, not from the cluster-scoped `DeviceClass` directly (which the kubelet does not have permission to read, maintaining least privilege). This ensures that the behavior of currently running pods is consistent and matches the configuration at the time they were scheduled.
+- **Backward Compatibility & Out-of-Tree / Custom Allocators**: Old scheduler clients or out-of-tree custom driver controllers/allocators might write allocation results without setting the new `RequiresNodePreparation` field (resulting in `nil`).
+  - *Mitigation*: The `RequiresNodePreparation` field defaults to `true` if it is `nil`. The kubelet treats `nil` as requiring node-local preparation. This guarantees 100% backward compatibility with all existing scheduler versions and custom vendor controllers. An out-of-tree allocator must be recompiled/upgraded with the updated API to explicitly set this field to `false` and bypass node-local preparation for their claims.
```
johnbelamaric: I think default false for optional fields is simpler and more often what we do; thus I flipped the orientation in the fields i suggested

troychiu: Yeah I agree! Updated.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yeah what you mentioned makes sense to me. I updated the KEP to introduce fields in `ResourceSlice`. Placing them on `Device` seems repetitive and might inflate the size of `ResourceSlice` instance. We can always add them to `Device` if there are use cases.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yeah that makes sense! Updated.

johnbelamaric (commented):
LGTM from the DRA perspective, have not reviewed the PRR section yet

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,392 @@
...
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+[Conformance Tests]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md
+[all GA Endpoints]: https://github.com/kubernetes/community/pull/1806
+
+## Summary
+
+This KEP introduces **Optional Node Preparation** to Dynamic Resource Allocation (DRA), allowing resource drivers to declare that node preparation and/or node unpreparation is not required for their devices. Currently, the kubelet assumes it must always coordinate with a node-local DRA driver via gRPC to prepare allocated devices before container start (`NodePrepareResources`), and to unprepare them during pod termination (`NodeUnprepareResources`).
```
pohly: Please line-wrap, I'll do a pass once that is done.

https://github.com/kubernetes/enhancements/blob/f515e2d79a601a1a85b29ca074e8534c7f23cfbf/keps/NNNN-kep-template/README.md?plain=1#L4-L7

Out of curiosity: which editor or tool did you use which prefers not wrapping?

troychiu: Updated. I use VSCode, and it requires an extension to perform line wrapping.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,586 @@
...
+- Eliminate node-local preparation entirely.
+- Enable users to override this infrastructure requirement at the individual
+  `ResourceClaimSpec` level.
+
+## Proposal
+
+We propose adding boolean fields `SkipNodePrepare` and `SkipNodeUnprepare` to
+`ResourceSliceSpec` and `DeviceRequestAllocationResult`.
+
+1. **API Definition**: The driver/controller publisher sets `SkipNodePrepare:
```
yujuhong: Curious: are there use cases where only node prepare or unprepare is skipped? If one of them is not skipped, there still needs to be a local driver, no?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,586 @@
...
+### API Changes
+
+1. **`ResourceSliceSpec`** in `pkg/apis/resource`:
+   ```go
+   type ResourceSliceSpec struct {
+       ...
+       // SkipNodePrepare indicates that node-local resource preparation (NodePrepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to false.
+       // +optional
+       SkipNodePrepare bool `json:"skipNodePrepare,omitempty" protobuf:"varint,5,opt,name=skipNodePrepare"`
```
yujuhong: Should we make the field `*bool` for an optional field? I'm a bit rusty on the API changes.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: Yes, I think it will be pretty common to skip the unprepare but require the prepare. You still need the local driver in that case. But if the driver is down or deleted, you can still terminate the pods if you can skip the unprepare. Right now, they get stuck.

yujuhong: I see. 

> if a driver is down or deleted

That sounds like a common issue for any driver though, if a local driver is anyway required. :)

troychiu: I would see
> if the driver is down or deleted, you can still terminate the pods if you can skip the unprepare. 

as a side benefit but not the main goal of the enhancement.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: That is a good question. I am unsure if there is a standard practice, as I see both types used in the codebase. I initially chose to use bool because its default value is false, which aligns with the current behavior.

However, I feel that `*bool` is more future-proof should the default behavior ever change. I will update it.

pohly: New alpha fields have to be a pointer to avoid storing the value when not set.

keps/sig-node/5945-dra-optional-node-preparation/README.md
pohly: I'm surprised that something so fundamental isn't built in. But I guess that explains why documents don't follow that guidance.

Note that there's a corollary: when changing some text, change as few lines as possible, even if that leads to wider or shorter lines than usual. Do not reflow the entire paragraph, that defeats the purpose.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,587 @@
...
+### Risks and Mitigations
+
+- **Dynamic ResourceSlice Changes**: An administrator or controller could update
+  `SkipNodePrepare` or `SkipNodeUnprepare` in a `ResourceSlice` while claims are
+  already allocated.
+  - *Mitigation*: The allocation result is written and frozen into the
+    `ResourceClaim` status at scheduling time. The kubelet reads the values from
+    the `AllocationResult` of the claim itself, not from the `ResourceSlice`
+    directly. This ensures that the behavior of currently running pods is
+    consistent and matches the configuration at the time they were scheduled.
```
pohly: This doesn't work in all cases:
- SkipNodePrepare is set in a slice.
- Allocation copies it.
- Driver gets updated in the cluster such that it now needs NodePrepare.
- kubelet skips NodePrepare.

The driver cannot prevent this. The admin must do such a migration very carefully. This is not a blocker for the feature because such a transition seems unlikely, but the KEP should be clear about how risky this is.

Copying the value is still the right thing to do because of the availability of the information for kubelet (shouldn't depend on ResourceSlices).

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,587 @@
...
+- **Backward Compatibility & Out-of-Tree / Custom Allocators**: Old scheduler
+  clients or out-of-tree custom driver controllers/allocators might write
+  allocation results without setting the new `SkipNodePrepare` and
+  `SkipNodeUnprepare` fields.
+  - *Mitigation*: The pointer fields default to `nil` when omitted or sent by
+    older versions, which is treated as `false` (not skipped). A value of
+    `false` (explicit or default nil) indicates that the kubelet **should not
+    skip** the action, meaning it defaults to executing node preparation and
+    clean-up as normal. This guarantees 100% backward compatibility with all
+    existing schedulers, custom controllers, and running workloads.
```
pohly: ... but only of the feature is not used. As soon as a driver wants to use it, allocation results must store the value or the kubelet will try to call the driver and fail.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,587 @@
...
+  - Verify that if a claim with skip fields set to `false` is passed, the
+    kubelet behaves normally and attempts to call the local driver.
+
+##### Integration tests
+
+We will add integration tests in `test/integration/dra` to verify:
+- A pod using a claim with `SkipNodePrepare: true` is admitted and scheduling
+  succeeds.
+- The kubelet processes the allocation without raising any errors or registering
+  warnings even when no physical node-local driver is registered.
```
pohly: There's no kubelet in integration testing. You don't need an integration test for this feature, it wouldn't add much value.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,587 @@
...
+   ```
+
+### Allocator Changes
+
+During scheduling, the structured parameters allocator resolves `ResourceSlices`
+that contain the allocated devices. The allocator extracts the `SkipNodePrepare`
+and `SkipNodeUnprepare` boolean values from the corresponding
+`ResourceSliceSpec` and copies them directly into each
+`DeviceRequestAllocationResult` under
+`ResourceClaim.Status.Allocation.Devices.Results`.
```
pohly: What is the behavior if the feature is turned off in the scheduler and there are ResourceSlices with the fields set?

This can happen during a cluster upgrade from alpha (off) to beta (on) if the admin installs a DRA driver using the feature before the upgrade is complete, because of a misconfiguration (inconsistent feature gates) or because the admin turned off the feature and there are lingering slices from the time when it was on.

This also influences the PRR answers.

My preference is to fail allocation when the scheduler detects this, because this enables retries after fixing the cluster. Scheduling a pod incorrectly (= fields not copied which were in the slice) is worse (no retry).

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I see. That's a helpful tip. Thank you!

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,587 @@
...
+devices. Some devices of our hardware pool require node-local interface setups
+(e.g., local host bindings), while others are fully managed and provisioned
+out-of-band in the control plane. By setting `SkipNodePrepare: false` (default)
+on node-local `ResourceSlices` and `SkipNodePrepare: true` on
+control-plane-managed `ResourceSlices`, the kubelet can dynamically determine
+whether to run node prep based on which device was selected by the scheduler. I
+do not need to split my driver into multiple separate drivers just to handle
+this architectural difference.
+
+### Risks and Mitigations
```
x0rw, line 168: Hi 👋 
My main concern here is that this assumes the driver implementations are setting `SkipNodePrepare` correctly. Otherwise, we would unlock a whole class of silent runtime errors for devices that require preparation but are marked as skippable incorrectly.

I know this is a bit outside of our **control**(to guarantee that the driver implementations are correct) but this risk should be made more explicit in the documentation in the future.

Otherwise, I think this is okay for alpha from a PRR review perspective.

x0rw (commented):
PRR shadow.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Your point makes sense. I have updated the "Risks and Mitigations" section and the PRR to mention migration risks and mitigations.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yes basically any inconsistency between driver/scheduler/kubelet would result in unexpected behavior. Updated the KEP to state the risks and mitigations.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: That's a good point; it's a possible scenario, and I agree with your preference. 

Then I wonder if the kubelet should also check for this scenario, but I'm unsure about the intended behavior. What are your thoughts on the following?

- For admission (`SkipNodePrepare`), if a claim's allocation result specifies `SkipNodePrepare: true`, the kubelet fails pod admission immediately with a clear error.
- For teardown (`SkipNodeUnprepare`), if a claim's allocation result specifies `SkipNodeUnprepare: true` but the node-local driver is not registered, the kubelet logs a warning, records an event, and bypasses the gRPC cleanup call, allowing the pod to be deleted cleanly.

Another approach is for the kubelet not to perform this check.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Thank you for your review, and yes, I agree. I will make sure we document that well.

keps/sig-node/5945-dra-optional-node-preparation/README.md
mortent: I agree with putting on the `ResourceSlice`. If someone wants to publish both devices that require preparation and devices that don't in the same resource pool, they can always split it across multiple `ResourceSlices`. After we scoped counters to the resource pool rather than the slice, I don't think there are any reasons why publishing the devices across multiple slices won't work.

inirah02:
Hi! I’m wrangling KEP #5945 for SIG Node v1.37. I noticed the release tracking status is currently “At risk for enhancements freeze,” and the KEP PR seems to still need prod-readiness and SIG Node approval. Is there anything blocking approval/merge right now, or is it mainly waiting on reviewer/approver attention? I’ll keep the SIG Node board updated accordingly.

keps/sig-node/5945-dra-optional-node-preparation/README.md
pohly: Whatever we pick, it must:
- be observable
- be recoverable, ideally without too much manual intervention
- not make the situation worse.

Failing during admission satisfies this. An admin can choose to re-enable the feature and the pod will run eventually or delete the pod without causing more chaos on node. One addition: admission should also check for SkipNodeUnprepare.

For teardown, I prefer to implement the skip. If the feature was disabled after admitting a pod with the field, then allowing the the pod to terminate is consistent with the original intent and must have been valid at the time when NodePrepareResources was called (assuming that it was called), so this causes no harm and cleans up the situation.

troychiu: This is a really clever idea. I think it covers most of the cases well, and I have updated the KEP accordingly. Thank you.

troychiu:
Hi @inirah02, there are a few open comments to resolve, but I don't expect this to take too long. Thank you!

keps/sig-node/5945-dra-optional-node-preparation/kep.yaml
```diff
@@ -0,0 +1,43 @@
+title: DRA Optional Node Preparation
+kep-number: 5945
+authors:
+  - "@troychiu"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scheduling
```
johnbelamaric: I don't think sig scheduling needs to be involved in this one

keps/sig-node/5945-dra-optional-node-preparation/kep.yaml
```diff
@@ -0,0 +1,42 @@
...
+kep-number: 5945
+authors:
+  - "@troychiu"
+owning-sig: sig-node
+# participating-sigs:
+status: implementable
+creation-date: 2026-05-21
+reviewers:
+  - "@mortent"
+approvers:
```
SergeyKanzhelev, line 11: You need sig node approver here. @mrunalp ?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+partitioned network services) where availability is discovered and published as
+`ResourceSlice` resources centrally by the controller. Because the devices
+require no node-local plumbing or mount operations on worker nodes, there is no
+node driver deployed. The controller publishes these resources with
+`SkipNodePrepare: true` and `SkipNodeUnprepare: true`. When users request these
+devices, the kubelet launches the pods immediately and cleanly, without
+complaining about missing node-local drivers, and without requiring any node
+driver DaemonSet to be present in the cluster.
+
+#### Flexibility for drivers with mixed preparation requirements
```
SergeyKanzhelev: Do we envision the GPU DRA driver to skip unprepare for non-partitioned devices the same way as Device Plugin doesn't bother with it?

I guess one thing will be needed is for that driver to ignore leftover CDI files

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+       ...
+       // SkipNodePrepare indicates that node-local resource preparation (NodePrepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodePrepare *bool `json:"skipNodePrepare,omitempty" protobuf:"varint,5,opt,name=skipNodePrepare"`
+
+       // SkipNodeUnprepare indicates that node-local resource cleanup (NodeUnprepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodeUnprepare *bool `json:"skipNodeUnprepare,omitempty" protobuf:"varint,6,opt,name=skipNodeUnprepare"`
```
SergeyKanzhelev: Can only unprepare be required and prepare skipped? is it a valid configuration? If not - it may be best to not allow it for now

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: Since prepare/unprepare is published in the slice, I am not sure if it's dynamic enough to do this. At least, I don't think generally it would respond to reconfiguration.

Ignoring leftover CDI files I think happens accidentally because at least the NVIDIA driver I believe names them with a claim UID. I suppose they could accumulate, maybe we need some cleanup process.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+   `ResourceClaim.Status.Allocation.Devices.Results[i].SkipNodePrepare` and
+   `SkipNodeUnprepare`.
+3. **Node Execution**: The kubelet reads these fields from the `ResourceClaim`'s
+   allocation results. If all allocated devices for a given driver within a
+   claim set `SkipNodePrepare: true` (or `SkipNodeUnprepare: true`), the kubelet
+   bypasses the corresponding gRPC call to the node-local resource driver.
+
+### User Stories
+
+#### Deploying controller-managed resources without node-local drivers
```
SergeyKanzhelev, line 154: do we need some sort of downward API for Pod to know which devices are associated with it? Today's design put all the downward API responsibility on a local driver. Will we need to move parts of it to the kubelet?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+
+- **2026-05-21**: KEP drafted and proposed as Provisional for Alpha stage.
+
+## Drawbacks
+
+- Adds a new boolean configuration field to the API, which increases API surface
+  area. However, this is necessary to support controller-managed or logical
+  resources natively without node-local drivers in a clean way.
+
+## Alternatives
```
SergeyKanzhelev, line 848: there is an alternative to have "catch all no-op plugin" that can be configured centrally with the knowledge of which device class to register for. But it will be hard to coordinate it for the mixed mode resources. Worth listing

johnbelamaric: https://github.com/gke-labs/dra-drivers/tree/main/dra-driver-noop

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: but it will be desireable as consuming the whole GPU typically doesn't require unprepare, assuming that every time it is partitioned we can unpartition in unprepare

keps/sig-node/5945-dra-optional-node-preparation/kep.yaml
troychiu: @mrunalp let me know if you are okay as an approver!

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I currently don't have such a use case in mind. @johnbelamaric do you foresee any?

Also, @SergeyKanzhelev what is the reason for adding this kind of limitations?

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: For Alpha, we can keep the scope tight: if a driver requires exposing metadata to the container via the Downward API, it must deploy a node-local driver and cannot use SkipNodePrepare. If they want a hybrid model, they can deploy a minimal, "no-op" node-local driver that only handles the metadata injection. This keeps the initial implementation of this KEP focused purely on the mechanics of skipping the gRPC hooks.

For Beta, we can consider supporting Kubelet-managed Downward API for these "no-prep" resources.

What do you think?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+driver via gRPC to prepare allocated devices before container start
+(`NodePrepareResources`) and to unprepare them upon pod termination
+(`NodeUnprepareResources`). For node-local accelerators (such as PCIe GPUs or
+local FPGAs), this node-level setup is critical to check device health,
+partition memory, and configure mount paths.
+
+However, there is an emerging class of resources whose lifecycles are managed
+entirely in the control plane and published centrally by a controller as
+`ResourceSlice` objects. These resources require absolutely zero node-local
+setup. Under the current architecture, the kubelet still assumes a node-local
```
bart0sh, line 102: Since you've mentioned GPUs and FPGAs above, would it make sense to also give examples of devices that don't need a node-level setup or cleanup? Especially those that don't require setup.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+When Kubelet prepares resources for an allocated claim, it evaluates the
+allocated devices' status:
+1. **Aggregation**: Because Kubelet invokes preparation and clean-up per-claim,
+   Kubelet can only bypass a gRPC step if **all** devices for a given driver
+   allocated in a claim have the respective skip field (`SkipNodePrepare` or
+   `SkipNodeUnprepare`) set to `true`.
+2. **Checkpointing**: Kubelet caches these aggregated properties inside its
+   checkpointed, claim-specific state (`ClaimInfo`) so they are safely preserved
+   across Kubelet restarts.
+3. **Bypassing**: During Pod admission and teardown, Kubelet's DRA manager
```
bart0sh: DRA manager is not called on pod admission by design, at least to my knowledge.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+       // SkipNodePrepare indicates that node-local resource preparation (NodePrepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodePrepare *bool `json:"skipNodePrepare,omitempty" protobuf:"varint,5,opt,name=skipNodePrepare"`
+
+       // SkipNodeUnprepare indicates that node-local resource cleanup (NodeUnprepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // +optional
+       SkipNodeUnprepare *bool `json:"skipNodeUnprepare,omitempty" protobuf:"varint,6,opt,name=skipNodeUnprepare"`
+   }
```
bart0sh, line 217: Should API server actions (e.g. skipping these fields if feature gate is disabled) be explained here?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+
+When Kubelet prepares resources for an allocated claim, it evaluates the
+allocated devices' status:
+1. **Aggregation**: Because Kubelet invokes preparation and clean-up per-claim,
+   Kubelet can only bypass a gRPC step if **all** devices for a given driver
+   allocated in a claim have the respective skip field (`SkipNodePrepare` or
+   `SkipNodeUnprepare`) set to `true`.
+2. **Checkpointing**: Kubelet caches these aggregated properties inside its
+   checkpointed, claim-specific state (`ClaimInfo`) so they are safely preserved
+   across Kubelet restarts.
```
bart0sh: Is ClaimInfo structure going to be updated, i.e. SkipNode* fields going to be added to it?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+3. **Bypassing**: During Pod admission and teardown, Kubelet's DRA manager
+   checks the claim's cached properties. If skipping is enabled for the driver
+   under that claim (meaning all allocated devices have the respective skip field
+   explicitly set to `true` in the allocation result), it bypasses driver registry
```
bart0sh, line 283: It should be explained what happens if some devices (should it be slices?) have skip field set to `false` and some - to `true`. I suspect the gRPC API would be called and it's up to driver to decide what to do in this case, correct?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+  and `SkipNodeUnprepare: true`.
+- **Workload**: Deploy a Pod referencing this template.
+- **Assertions**:
+  - The Pod reaches the `Running` phase successfully.
+  - No `FailedPrepareDynamicResources` warnings are posted to the Pod events.
+  - Bypassing the driver lookup does not crash the kubelet or produce resource
+    leaks.
+  - Pod deletion completes cleanly and immediately (no hanging in `Terminating`
+    status waiting for unprepare calls).
+
```
bart0sh, line 451: Should we also test upgrade / downgrade scenarios?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+- No need to add integration tests since the feature can be validated with unit and e2e tests.
+
+##### e2e tests
+
+We will add a new End-to-End test case inside `test/e2e/dra/dra.go` using a
+driver without node-local components:
+- **Setup**: Deploy a DRA test driver without node gRPC components running on
+  worker nodes (`WithKubelet = false`).
+- **API Configuration**: Publish `ResourceSlices` with `SkipNodePrepare: true`
+  and `SkipNodeUnprepare: true`.
```
bart0sh: This is just one use case. It's not enough to properly test this feature in my opinion. All combinations should be tested.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+  - Mock a claim requiring allocation with `SkipNodePrepare: true` and
+    `SkipNodeUnprepare: true`.
+  - Assert that `prepareResources` and `unprepareResources` complete
+    successfully without querying the plugin manager or invoking gRPC calls.
+  - Verify that if a claim with skip fields set to `false` is passed, the
+    kubelet behaves normally and attempts to call the local driver.
+
+##### Integration tests
+
+- No need to add integration tests since the feature can be validated with unit and e2e tests.
```
bart0sh: Proposed tests don't involve Kubelet, so either e2e tests should be extended to cover all pod lifecycle or integration tests should be implemented.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,697 @@
...
+
+We will add a new End-to-End test case inside `test/e2e/dra/dra.go` using a
+driver without node-local components:
+- **Setup**: Deploy a DRA test driver without node gRPC components running on
+  worker nodes (`WithKubelet = false`).
+- **API Configuration**: Publish `ResourceSlices` with `SkipNodePrepare: true`
+  and `SkipNodeUnprepare: true`.
+- **Workload**: Deploy a Pod referencing this template.
+- **Assertions**:
+  - The Pod reaches the `Running` phase successfully.
```
bart0sh: Is it even possible for the pod to reach `Running` phase if `WithKubelet = false` ?

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: I don't think we can predict that, and I don't see why we would add arbitrary restrictions. That would seem to add extra code, and it is simpler to just treat each field as independent.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: I would just never unprepare, and make sure we have a clean state by un-/re-partitioning in the prepare call.

But I do wonder if more generally we need CDI cleanup (outside of unprepare)? So we don't accumulate files if there is a misbehaving controller. I think that's orthogonal to this KEP - maybe @troychiu you can check and see if this is a problem and add an issue in k/k?

keps/sig-node/5945-dra-optional-node-preparation/kep.yaml
johnbelamaric: I would list him, he is generally our SIG Node DRA approver. If he can't do it, he'll let us know :)

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: For now, I would say if Pods need that, then the there needs to be a node-local driver. We can add that as a separate feature if this turns out to be an issue.

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: I am thinking from the usability perspective. I am not clear whether kubelet will try to check if the local driver present on pod scheduling. So it will try to instantiate a bunch of things on pod termination, making termination more fragile. Frankly, it is best if we fail to start the Pod than keeping it spinning in termination.

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: So DRA resources are just an accounting of some external resources? Or some external controller will associate things with the Pod based on claims? My question is coming from usability - if each time there will be a need to route some metadata inside the Pod separately from DRA mechanism, it may be easier just to support some basic downward API

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: yep.  May be good to mention it for completeness. I understand it is not scalable, I like this proposal better

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: indeed, it is all inside the sandbox creation stage

SergeyKanzhelev:
WIll kubelet try to connect to Health endpoint of DRA driver is allocation is skipped? cc @harche

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: There are three things:
- Driver registration with kubelet
- SkipPrepare
- SkipUnprepare

The scheduler does not know if the driver has registered with the kubelet, so a Pod can get scheduled whether or not there is a registered driver.

Right now, we fail on Prepare if there is no registered driver. I think what you are saying is if we don't fail then, but we have to unprepare and their is no registered driver, then they will be stuck in terminating.

Maybe we should verify that a driver is registered for any devices that have prepare or unprepare, so we can fail up front? Could we do that on pod admission?

johnbelamaric:
> WIll kubelet try to connect to Health endpoint of DRA driver is allocation is skipped? cc @harche

It should call the health endpoint of all registered drivers, IMO.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: The controller-only use cases are accounting and provisioning of external resources. The application may utilize those, so I can see your point. I just don't have a specific use case for that right now.

SergeyKanzhelev: what is the specific use case for just accounting? Or external controller and how that external controller react quickly on Pod being started. Maybe KEP may give an example. I am fine if the Downward API is postponed as long as there is a clear use case end-to-end explained.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yeah that's a good point. A configuration where `SkipNodePrepare: true` and `SkipNodeUnprepare: false` is fragile because the Kubelet wouldn't verify the driver's presence at startup, leading to pods getting stuck in `Terminating` if the driver is missing during deletion.

If a resource requires node-local cleanup (Unprepare), it must have a node-local driver, which can easily serve Prepare even if it's a no-op.

Disallowing this configuration makes sense to me.

SergeyKanzhelev: > Maybe we should verify that a driver is registered for any devices that have prepare or unprepare, so we can fail up front? Could we do that on pod admission?

we do not do pod admission with the DRA. But yes, failing early is best and this is why I am thining it is easier to enforce that both must be set if unprepare is needed - less complexity in kubelet, the while complexity is in API type checking

johnbelamaric: yeah, makes sense. and as Troy said, they already need a driver so they need to implement the method. the primary reason to split this into two fields instead of one was to allow prepare without unprepare and reduce the chance of stuck in termination. so I am good with only allowing skipprepare/skipunprepare combinations of  false/false, true/true, false/true and disallowing true/false.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: IMO both patterns you mentioned are valid, but for Control-plane associated resources, the Pod does need the metadata.

If we want to support this natively without a node driver, we should indeed allow Kubelet to manage the basic downward API for these claims. Potentially, this could work by having the scheduler/allocator copy the attributes from the `ResourceSlice` into the `ResourceClaim` status during allocation, allowing Kubelet to read them and write the CDI/JSON file directly.

Should we treat this as a separate feature, or is it better to include this design as part of the Beta transition plan for this KEP?

johnbelamaric: We discussed this quite a bit with @alaypatel07 for #5304 and landed on the driver focused implementation in his KEP. I'd rather leave it there. Copying all the attributes to status would be a huge size increase on the object.

harche:
> WIll kubelet try to connect to Health endpoint of DRA driver is allocation is skipped? cc @harche

No, the health stream is tied to plugin registration, not to allocation or prepare.

Kubelet only starts the `WatchResources` health stream when a driver's kubelet plugin registers via the plugin registration socket. This happens in `DRAPluginManager.add()`, and the only conditions are the `ResourceHealthStatus` feature gate being enabled and a stream handler being present. Nothing about claims or allocation is involved:
https://github.com/kubernetes/kubernetes/blob/d57b170cda65d284030d6773f7b967f224a6158b/pkg/kubelet/cm/dra/plugin/dra_plugin_manager.go#L387-L409

So for the main use case in this KEP, a controller only driver with no node component, there is no registered plugin and hence no gRPC connection at all. Kubelet won't try to reach a health endpoint for it, just like it won't call `NodePrepareResources`.

For the mixed case where a driver does run a node plugin but some devices have `SkipNodePrepare: true`, the health stream runs as usual for that registered plugin, independent of whether prepare was skipped for any particular claim. Today kubelet attempts the stream against every registered DRA plugin and simply retries every 5s if the driver doesn't implement it. kubernetes/kubernetes#139477 changes that so kubelet only connects when the plugin advertises the `DRAResourceHealth` service in its supported services at registration time.

keps/sig-node/5945-dra-optional-node-preparation/README.md
SergeyKanzhelev: Yep. I am not advocating for the whole suite of downward API. The issue is that there is no extensibility point between the Pod bound to specific resources and it is running to add any metadata to it or catch the exact moment of a pod starting (short of init container that will query API). And it may impose limitations on this design, even though I see it as a great thing to avoid the whole failure mode of maintaining the external DRA driver on the node. So maybe some CEL-based downward api querying from resources might do it.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Added it as an alternative. Thank you.

keps/sig-node/5945-dra-optional-node-preparation/kep.yaml
troychiu: Added. Thank you!

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: The kubelet can only see devices and cannot see slices.

gRPC call is skipped only when all allocated devices have the respective skip field explicitly set to `true`. If any device has a `nil` or `false` value, it defaults to `false` (do not skip).

https://github.com/kubernetes/enhancements/pull/6104/changes#diff-88cfe74750f75b39d67987baab792ebdbc286928ff4d02d8474f96b967262c02R282-R283

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Good point. Updated the test section to include all combinations.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yes this plan also sounds good to me. Let me update the KEP.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I agree. That indeed seems a gap, and a CEL-based downward API for resources is potentially a solution.

I am trying to figure out where this design work best belongs:

1. Should we extend KEP-5304 (since it defines the DRA downward API contract)?
2. Should we extend this KEP, potentially as a Beta graduation?
3. Or, since modifying the core Pod spec to support CEL is a significant API change, does it deserve a separate KEP entirely?

How should we make this decision?

I'd love to get your advice on the best way to steer this. Thank you.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Yeah that's wrong wording. Updated.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: added a new section. Do you mind helping review?

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I plan to update `DriverState` (https://github.com/kubernetes/kubernetes/blob/c2e3492837be213a2ffc143bb81588d20df91247/pkg/kubelet/cm/dra/state/state.go#L49-L51) because this information is specific to each driver. I didn't mention this since I thought it was too detailed. Should I?

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Added upgrade / downgrade e2e section

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: We decided not to have integration tests (see https://github.com/kubernetes/enhancements/pull/6104#discussion_r3347228710). I have updated e2e tests sections to cover more scenarios.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Hmm that's what I understood. Let me double check.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Updated

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: I see this as a major scope increase. I'd rather treat that as a separate feature.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,876 @@
...
+preparation and cleanup unless explicitly set to `true` in the published
+`ResourceSlice` by the driver.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Setting the feature gate to `false` and restarting components will disable
+it. If disabled, any new allocations for standard drivers will proceed normally
+(propagating `false` or `nil`). However, any allocation requests targeting drivers
+that set `SkipNodePrepare: true` or `SkipNodeUnprepare: true` in their `ResourceSlices`
+will fail during allocation, preventing workloads from scheduling into a state
```
johnbelamaric, line 648: If the feature gate is disabled in the apiserver, how will the scheduler know this to fail allocation?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,876 @@
...
+    to the legacy behavior of expecting node preparation.
+  - If any pods are running using a driver without node-local drivers, those
+    pods will fail to restart or delete cleanly if the kubelet tries to invoke
+    node-local gRPC calls that don't exist. Operators must ensure all pods using
+    no-prep claims are terminated before downgrading, or ensure temporary no-op
+    drivers are running during downgrade transitions.
+
+### Version Skew Strategy
+
+- **Older kubelet (N-1) / Upgraded Control Plane (N)**:
```
johnbelamaric: For kubelet skew we go back further. I think the main point here is that any node running an old kubelet must continue to run a noop driver.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,876 @@
...
+    - For already running pods (in case the feature gate was disabled after
+      successful `PrepareResources`), the kubelet will honor `SkipNodeUnprepare: true` and skip the
+      unprepare call during `UnprepareResources`, allowing the pod to terminate cleanly.
+
+- **Scheduler Feature Gate Disabled / SkipNodePrepare or SkipNodeUnprepare set to true**:
+  - If the `DRAOptionalNodePreparation` feature gate is disabled in the
+    scheduler/allocator, but a driver publishes `ResourceSlices` with
+    `SkipNodePrepare: true` or `SkipNodeUnprepare: true` (e.g., due to
+    inconsistent feature gates in a rolling upgrade, or lingering slices after
+    downgrade), the scheduler/allocator will fail the allocation of those
```
johnbelamaric, line 614: I don't think that's true? I could be wrong.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: Or maybe NodeDeclaredFeatures will prevent scheduling to those nodes?

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: If the feature gate is disabled in the apiserver, then the `resourceslice` cannot set `SkipNodePrepare`/`SkipNodeUnprepare` to `true`

So here I try to discuss the case where the scheduler (feature disabled) should fail the resource allocation for a `resourceslice` that had fields set to true before the feature was disabled.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I was unaware of `NodeDeclaredFeatures`, but it seems like a better approach for dealing with version skew. Should I include it in this KEP?

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Perhaps my writing was confusing, but this is an intended action to avoid pods getting stuck, which I mentioned in Allocator Changes section. Also see https://github.com/kubernetes/enhancements/pull/6104#discussion_r3347291544.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: yes, we should integrate with that

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: My understanding is that having `WithKubelet = false`, the Kubelet on that node won't have the driver registered.

https://github.com/kubernetes/kubernetes/blob/9f664f538b8acd07897a2607252baf71bf6a8c76/test/e2e/dra/utils/deploy.go#L715-L730

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Updated. Thank you.

mrunalp (approved):

keps/sig-node/5945-dra-optional-node-preparation/README.md
bart0sh: > The kubelet can only see devices and cannot see slices.

Kubelet (DRA plugin manager to be exact) can see and already works with slices. It wipes them when DRA driver can't do that. Other parts of DRA manager can do that too if needed as it's just a matter of extra API calls. Why do you think Kubelet can't do that? Not that I'm/was insisting on this, asking just out of curiosity.

> gRPC call is skipped only when all allocated devices have the respective skip field explicitly set to true. If any device has a nil or false value, it defaults to false (do not skip).

Yes, that's an implicit way to say that if not all allocated devices have this field set to true, [Un]Prepare GRPC API will be called and it's up to the DRA driver to decide how to deal with this. Would it make sense to explain it explicitly?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,936 @@
...
+from being scheduled to incompatible nodes.
+
+We register a new declared feature:
+*   **Feature Name**: `DraOptionalNodePreparation`
+*   **Associated Feature Gate**: `DRAOptionalNodePreparation`
+*   **Discovery Logic (Kubelet)**: A node declares support for
+    `DraOptionalNodePreparation` in its `node.status.declaredFeatures` if and
+    only if the `DRAOptionalNodePreparation` feature gate is enabled on the
+    Kubelet.
+*   **Inference Logic (Scheduler & Admission)**: The control plane infers that a
```
bart0sh, line 329: What does "Admission" mean in this context? Is it related to Kubelet? Currently Kubelet doesn't call prepare/unprepare gRPC APIs on admission.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,936 @@
...
+
+- **Setup**: Deploy a DRA test driver without node gRPC components running on
+  worker nodes (`WithKubelet = false`).
+- Test Case 1.1: Fully skipped preparation (`SkipNodePrepare: true`,
+  `SkipNodeUnprepare: true`)
+  - **API Configuration**: Publish `ResourceSlices` with both fields set to
+    `true`.
+  - **Workload**: Deploy a Pod referencing this resource.
+  - **Assertions**:
+    - The Pod reaches the `Running` phase successfully.
```
bart0sh, line 422: We should also verify that pod can access allocated devices.

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,936 @@
...
+2. **Checkpointing**: Kubelet caches these aggregated properties inside its
+   checkpointed, claim-specific state (`ClaimInfo`) so they are safely preserved
+   across Kubelet restarts.
+3. **Bypassing**: During `PrepareResources` and `UnprepareResources`, the DRA manager
+   checks the claim's cached properties. If skipping is enabled for the driver
+   under that claim (meaning all allocated devices have the respective skip field
+   explicitly set to `true` in the allocation result), it bypasses driver registry
+   lookup and the respective gRPC calls (`NodePrepareResources` or `NodeUnprepareResources`),
+   allowing container startup/pod termination to proceed immediately. If any device
+   has a `nil` or `false` value, it defaults to `false` (do not skip).
```
bart0sh, line 286: Should we explain here how allocated device info is going to be passed to the container runtime if `NodePrepareResources` is skipped? Currently it's done using CDI and CDI files are created by the DRA Driver during `NodePrepareResources` call.

keps/sig-node/5945-dra-optional-node-preparation/README.md
bart0sh: No need to mention, but we need to test how old Kubelet handles updated checkpoints. Should this be explained somewhere in this document?

keps/sig-node/5945-dra-optional-node-preparation/README.md
```diff
@@ -0,0 +1,936 @@
...
+
+1. **`ResourceSliceSpec`** in `pkg/apis/resource`:
+   ```go
+   type ResourceSliceSpec struct {
+       ...
+       // SkipNodePrepare indicates that node-local resource preparation (NodePrepareResources gRPC)
+       // is not required for the devices in this slice. Defaults to nil (false).
+       // (SkipNodePrepare: true and SkipNodeUnprepare: false/nil) is invalid and will be rejected.
+       // +optional
+       SkipNodePrepare *bool `json:"skipNodePrepare,omitempty" protobuf:"varint,5,opt,name=skipNodePrepare"`
```
bart0sh:
```suggestion
       SkipNodePrepare *bool `json:"skipNodePrepare,omitempty" protobuf:"varint,9,opt,name=skipNodePrepare"`
```

<= 8 protobuf numbers are already taken: https://github.com/kubernetes/kubernetes/blob/v1.36.2/staging/src/k8s.io/api/resource/v1/types.go#L212

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: But if it's disabled in the api server, will the scheduler see the fields at all?

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: ok, i think I see

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Sure! Added. Thank you for the suggestion.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: I might be wrong, but it seems like API machinery typically does not prune fields on the read path.

keps/sig-node/5945-dra-optional-node-preparation/README.md
bart0sh: You're right. I didn't look at the code and thought that WithKubelet = false means that the Kubelet is not running at all.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: Hmm. Ok.

johnbelamaric (commented):
/approve

but leaving the LGTM to @bart0sh

keps/sig-node/5945-dra-optional-node-preparation/README.md
bart0sh: As far as I understand we have more or less 2 real use cases for skipping and both seem to create additional complications:
- Skipping unprepare: we need to somehow remove CDI leftovers.
- Skipping both: no real use cases + it's unclear how devices are allocated and become available in containers + the concerns expressed above (metadata, downward API etc).

Considering the above, does this really worth it comparing to implementing 1/2 no-op gRPC calls?

bart0sh:
> /approve
> 
> but leaving the LGTM to @bart0sh

Before lgtm it, I need these review comments to be addressed:
- https://github.com/kubernetes/enhancements/pull/6104#discussion_r3415668192
- https://github.com/kubernetes/enhancements/pull/6104#discussion_r3394473948
- https://github.com/kubernetes/enhancements/pull/6104#discussion_r3412987038

The rest is minor and can be done during implementation.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: > Skipping unprepare: we need to somehow remove CDI leftovers.

Actually I think this is unrelated to skipping unprepare. I am not sure unprepare always removes this, because claims may be shared. I may be wrong on that though. @troychiu are prepare/unprepare called for every pod using a claim or just on the first and last pods (respectively)?

 > Skipping both: no real use cases + it's unclear how devices are allocated and become available in containers + the concerns expressed above (metadata, downward API etc).

So, for example we have an internal use case that is using binding conditions to update annotations on the Pod after scheduling. We don't need any new downward API, it is a controller-only flow. Something like this:
- Controller published resource slices, including a binding condition
- User requests a device from those
- Scheduler allocates a device
- Controller sees that and does some stuff; when it is done, it sets an annotation on the Pod and sets the binding condition
- Pod uses existing downward api to read the annotation

The alternative is for the Pod to have an init container that reads the resourceclaims associated with itself from the API server.

We *could* add in some way to publish data from the RCs via downward API. I am OK with that, I just don't think it's necessary. The question is what data should be made available this way? In #5304 we wanted device attributes, which come from the *ResourceSlice*. I think in that case we can require an on-node driver. If we only need data from the *ResourceClaim*, maybe we can add that into this scope? Seems late in the KEP cycle to add that though.

I would prefer to proceed with this KEP as-is, and if we need that add it next cycle in here or a separate KEP.

keps/sig-node/5945-dra-optional-node-preparation/README.md
johnbelamaric: https://github.com/gke-labs/dra-drivers/tree/main/dra-driver-image-configurator

is one example. we also have an internal use case (described in the other comment) but I can't go into detail on that

keps/sig-node/5945-dra-optional-node-preparation/README.md
pohly: > are prepare/unprepare called for every pod using a claim or just on the first and last pods (respectively)?

First and last. Claims are prepared once, not for individual pods.

johnbelamaric: Ok. So the CDI file cleanup is expected to happen in unprepare? Because I believe with shared claims those files are shared as well, we can't clean them up on Pod teardown. AFAIK we don't have any other way to clean them up from kubelet, because the driver can make up whatever names they want for the files. We wouldn't know which ones we can do that with.

So, does it still make sense to have "prepare without unprepare"? Maybe we just need one flag after all?

pohly: > So the CDI file cleanup is expected to happen in unprepare? 

Yes. That's when it is certain that nothing is depending on them.

> So, does it still make sense to have "prepare without unprepare"?

It only makes sense if "prepare" is a NOP (but then why does it not get skipped?) or whatever it does gets cleaned up some other way, without involving the kubelet and the DRA driver. This sounds a bit far-fetched to me.

I prefer to not add functionality and complexity, no matter how small, when there is no real need for it.

johnbelamaric: Ok. @troychiu how about we go back to just a single boolean then? Maybe `SkipNodeOperations`? We don't need the final name now, it can be figured out in implementation PRs.

@bart0sh if we do that, and add a reference to the `image-configurator` use case, are we good to go?

troychiu: Yes that sounds good to me. Thank you

troychiu:
Thanks @bart0sh for the review. I have updated the enhancement according to https://github.com/kubernetes/enhancements/pull/6104#discussion_r3416217910. Let me know what you think. Thank you.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: Fixed!

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: This refers to the kubelet pod admission, which leverage node declared features.

keps/sig-node/5945-dra-optional-node-preparation/README.md
troychiu: https://github.com/kubernetes/enhancements/pull/6104#discussion_r3415794807 is a way to do that. Should I mention that in the KEP?

keps/sig-node/5945-dra-optional-node-preparation/README.md
bart0sh: > @bart0sh if we do that, and add a reference to the `image-configurator` use case, are we good to go?

Absolutely.

PR #6106, opened 2026-05-22
KEP-5710: Update Workload-aware preemption KEP for beta in v1.37

Argh4k:
- One-line PR description: Update KEP-5710 for beta in v1.37

- Issue link: https://github.com/kubernetes/enhancements/issues/5710

- Other comments: 
  - Apart from Beta graduation itself this update:
    - Eliminates delayed preemption completely
    - Removes Workload Aware preemption specific feature gate and ties the effort with GangScheduling behind GenericWorkload feature gate
    - Describes in details the proposed "in place reprieval" 
    - Describes in details PodGroupPostFilter extension point

PR #6113, opened 2026-05-25
KEP-5690: Add proposal (DRA preemption)

mortent:
- One-line PR description: Preemption for DRA

- Issue link: https://github.com/kubernetes/enhancements/issues/5690

- Other comments: Google Gemini and Claude was used while writing this KEP

### DIFF
--- a/keps/prod-readiness/sig-scheduling/5690.yaml
+++ b/keps/prod-readiness/sig-scheduling/5690.yaml
@@ -1,0 +1,3 @@
+kep-number: 5690
+alpha:
+  approver: "@johnbelamaric"
--- a/keps/prod-readiness/sig-scheduling/5690.yaml
+++ b/keps/prod-readiness/sig-scheduling/5690.yaml
@@ -1,3 +1,3 @@
 kep-number: 5690
-alpha:
-  approver: "@johnbelamaric"
+beta:
+  approver: "@wojtek-t"
--- a/keps/sig-scheduling/5690-dra-preemption/README.md
+++ b/keps/sig-scheduling/5690-dra-preemption/README.md
@@ -203,8 +203,8 @@ Features/scenarios that we will not support:
   pods on a single node, meaning that it will not be able to free up devices that are allocated to claims
   referenced by pods spanning multiple nodes.
 * **ResourceClaims for PodGroups**: These claims are allocated to a PodGroup, meaning that all pods in the
-  group must be scheduled together. The default preemption algorithm does not consider the PodGroup at all,
-  so we will not allow preemption of pods that reference such claims.
+  group must be scheduled together. The dynamicresources plugin will not support preemption for
+  ResourceClaims allocated at the PodGroup level.
 
 
 ### Test Plan
@@ -313,7 +313,6 @@ scenarios will be handled by integration tests.
 
 #### Beta
 
-- Gather feedback
 - Additional tests are in Testgrid and linked in KEP
 
 #### GA
@@ -561,7 +560,7 @@ There are no SLOs for this feature.
 
 ## Implementation History
 
-1.33: first KEP revision and implementation directly to Beta
+1.37: first KEP revision and implementation directly to Beta
 
 ## Drawbacks
 
--- a/keps/sig-scheduling/5690-dra-preemption/README.md
+++ b/keps/sig-scheduling/5690-dra-preemption/README.md
@@ -1,0 +1,674 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+Follow the guidelines of the [documentation style guide].
+In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-NNNN: Your short, descriptive title
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
+**Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
+-->
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+Dynamic Resource Allocation (DRA) was promoted to stable in v1.35, but one major capability—preemption of pods
+utilizing DRA resources—is currently not supported. Under the current state, lower-priority workloads that
+happen to be scheduled early can occupy scarce devices indefinitely, leaving higher-priority workloads
+pending. By introducing native preemption support in DRA, we equip the scheduler with the necessary
+capabilities to ensure that the most critical workloads gain access to scarce hardware resources.
+
+## Motivation
+Acquisition and allocation of specialized hardware (e.g., GPUs and TPUs) is a primary operational concern for
+AI/ML users. Because accelerator resources are frequently constrained, it is essential that Kubernetes
+provides robust tools to run high-priority workloads on the available hardware. Introducing preemption
+support for DRA ensures that the scheduler can reclaim resources from lower-priority workloads to satisfy
+scheduling requirements of higher-priority tasks.
+
+### Goals
+* Enable the scheduler to preempt pods consuming node-local DRA devices.
+
+### Non-Goals
+* Support preemption of workloads utilizing PodGroup-level ResourceClaims.
+* Support preemption of workloads using multi-node or network-attached devices.
+
+## Proposal
+We will implement the `fwk.PreFilterExtensions` interface in the `dynamicresources` scheduler plugin, which
+requires implementing the `AddPod` and `RemovePod` methods. These functions are invoked by the core
+`DefaultPreemption` plugin to incrementally simulate the removal or recovery (reprieval) of candidate victim
+pods during the preemption planning loop.
+Implementing these functions requires updating the internal `stateData` structure in the `dynamicresources`
+plugin to track state transitions and maintain local capacity maps transactionally throughout the preemption
+simulation.
+The scheduler must correctly handle several advanced DRA features during this simulation, or explicitly
+bypass preemption if those features are in use.
+
+Features that we will support but require careful implementation:
+* **Partitionable Devices**: Freeing up the exact device partition requested by a higher-priority workload
+  may require preempting multiple lower-priority pods that collectively occupy fractional partitions
+  (represented as counters) on the same physical device, so that the raw capacity can be consolidated
+  and re-partitioned.
+* **Consumable Capacity**: We might need to free up just a subset of the capacity on a device to satisfy
+  the request of a higher-priority pod. We need to make sure the remaining capacity on a device is correctly
+  tracked during preemption simulations, and that we don't overcommit.
+* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
+  sure this is handled correctly during preemption simulations.
+
+Features/scenarios that we will not support:
+* **ResourceClaims that span multiple nodes**: The default preemption algorithm only considers preempting
+  pods on a single node, meaning that it will not be able to free up devices that are allocated to claims
+  referenced by pods spanning multiple nodes.
+* **ResourceClaims for PodGroups**: These claims are allocated to a PodGroup, meaning that all pods in the
+  group must be scheduled together. The default preemption algorithm does not consider the PodGroup at all,
+  so we will not allow preemption of pods that reference such claims.
+
+
+### Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+The goal is to ensure that we don't accept enhancements with inadequate testing.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+<!--
+Based on reviewers feedback describe what additional tests need to be added prior
+implementing this enhancement to ensure the enhancements have also solid foundations.
+-->
+
+##### Unit tests
+
+<!--
+In principle every added code should have complete unit test coverage, so providing
+the exact set of tests will not bring additional value.
+However, if complete unit test coverage is not possible, explain the reason of it
+together with explanation why this is acceptable.
+-->
+
+<!--
+Additionally, for Alpha try to enumerate the core package you will be touching
+to implement this enhancement and provide the current unit coverage for those
+in the form of:
+- <package>: <date> - <current test coverage>
+The data can be easily read from:
+https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit
+
+This can inform certain test coverage improvements that we want to do before
+extending the production code to implement this enhancement.
+-->
+
+<!--
+Generated with:
+go test -cover ./pkg/scheduler/framework/plugins/dynamicresources/... | sed -e 's/.*\(k8s.io[a-z/-]*\).*coverage: \(.*\) of statements/- `\1`: \2/' | sort
+-->
+
+- `k8s.io/kubernetes/pkg/scheduler/framework/plugins/dynamicresources`: 82.4%
+
+##### Integration tests
+
+<!--
+Integration tests are contained in https://git.k8s.io/kubernetes/test/integration.
+Integration tests allow control of the configuration parameters used to start the binaries under test.
+This is different from e2e tests which do not allow configuration of parameters.
+Doing this allows testing non-default options and multiple different and potentially conflicting command line options.
+For more details, see https://github.com/kubernetes/community/blob/master/contributors/devel/sig-testing/testing-strategy.md
+
+If integration tests are not necessary or useful, explain why.
+-->
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, document that tests have been written,
+have been executed regularly, and have been stable.
+This can be done with:
+- permalinks to the GitHub source code
+- links to the periodic job (typically https://testgrid.k8s.io/sig-release-master-blocking#integration-master), filtered by the test name
+- a search in the Kubernetes bug triage tool (https://storage.googleapis.com/k8s-triage/index.html)
+-->
+
+Integration tests will be added that covers the normal functionality of preemption with DRA, with
+specific tests covering all the special scenarios called out above.
+
+##### e2e tests
+
+<!--
+This question should be filled when targeting a release.
+For Alpha, describe what tests will be added to ensure proper quality of the enhancement.
+
+For Beta and GA, document that tests have been written,
+have been executed regularly, and have been stable.
+This can be done with:
+- permalinks to the GitHub source code
+- links to the periodic job (typically a job owned by the SIG responsible for the feature), filtered by the test name
+- a search in the Kubernetes bug triage tool (https://storage.googleapis.com/k8s-triage/index.html)
+
+We expect no non-infra related flakes in the last month as a GA graduation criteria.
+If e2e tests are not necessary or useful, explain why.
+-->
+
+E2e tests will be added to cover the basic scenarios for preemption with DRA. The more complicated
+scenarios will be handled by integration tests.
+
+### Graduation Criteria
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- Gather feedback
+- Additional tests are in Testgrid and linked in KEP
+
+#### GA
+
+- 2 examples of real-world usage
+- Allowing time for feedback
+
+
+### Upgrade / Downgrade Strategy
+
+Standard upgrade/downgrade strategies may be used, no special configuration
+changes are needed. The changes are local to the scheduler.
+
+### Version Skew Strategy
+
+This feature only impacts the scheduler, so there will be no version skew issues.
+
+## Production Readiness Review Questionnaire
+
+<!--
+
+Production readiness reviews are intended to ensure that features merging into
+Kubernetes are observable, scalable and supportable; can be safely operated in
+production environments, and can be disabled or rolled back in the event they
+cause increased failures in production. See more in the PRR KEP at
+https://git.k8s.io/enhancements/keps/sig-architecture/1194-prod-readiness.
+
+The production readiness review questionnaire must be completed and approved
+for the KEP to move to `implementable` status and be included in the release.
+
+In some cases, the questions below should also have answers in `kep.yaml`. This
+is to enable automation to verify the presence of the review, and to reduce review
+burden and latency.
+
+The KEP must have a approver from the
+[`prod-readiness-approvers`](http://git.k8s.io/enhancements/OWNERS_ALIASES)
+team. Please reach out on the
+[#prod-readiness](https://kubernetes.slack.com/archives/CPNHUMN74) channel if
+you need any help or guidance.
+-->
+
+### Feature Enablement and Rollback
+
+<!--
+This section must be completed when targeting alpha to a release.
+-->
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+<!--
+Pick one of these and delete the rest.
+
+Documentation is available on [feature gate lifecycle] and expectations, as
+well as the [existing list] of feature gates.
+
+[feature gate lifecycle]: https://git.k8s.io/community/contributors/devel/sig-architecture/feature-gates.md
+[existing list]: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/
+-->
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: DRAPreemption
+  - Components depending on the feature gate:
+    - kube-scheduler
+
+
+###### Does enabling the feature change any default behavior?
+
+<!--
+Any change of default behavior may be surprising to users or break existing
+automations, so be extremely careful here.
+-->
+
+Yes, it will cause the scheduler to preempt pods referencing ResourceClaims to
+make room for higher priority pods. There is no API change for this feature, so it
+will potentially impact any Pod in the cluster when the feature is enabled.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes, disabling the feature will prevent the scheduler from preempting pods that
+reference ResourceClaims.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+That it was previously rolled back have no impact. Reenabling it just means that
+preemption of pods referencing ResourceClaims will again be considered.
+
+###### Are there any tests for feature enablement/disablement?
+
+We will cover this scenario in both unit tests and integration tests.
+
+### Rollout, Upgrade and Rollback Planning
+
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+A rollout will immediately make pods referencing ResourceClaims eligible for
+preemption. So if there are higher priority pods that can't be scheduled due
+to pending ResourceClaims, the result will be that lower priority pods will
+get preempted.
+
+Similarly, a rollback means that pods referencing ResourceClaims will no longer
+be eligible for preemption. But already preempted workload will remain in the
+pending state until sufficient resources are freed up.
+
+###### What specific metrics should inform a rollback?
+
+The specific signal that would suggest this feature should be rolled back, would
+be if pods are being preempted when they shouldn't be. This means there is a bug
+somewhere in the implementation.
+
+If the `scheduler_preemption_victims` increases significantly when the feature is enabled, but we don't 
+see a corresponding increase in pods being scheduled, we should investigate. It would suggest
+that pods are being preempted incorrectly and the higher-priority pods are not actually
+being scheduled.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+It will be tested by bringing up a KinD cluster and enabling the feature gate.
+We will then run a simple workload with higher priority pods and then enable
+the feature gate to verify that the lower priority pods are preempted. We will then 
+disable the feature gate and verify that anotherpod with an even higher priority
+does not preempt the existing pod.
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+It is enabled on a cluster-level, so if it is enabled, it is in use. It will only impact
+pods that are referencing ResourceClaims.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [ ] Events
+  - Event Reason: 
+- [ ] API .status
+  - Condition name: 
+  - Other field: 
+- [ ] Other (treat as last resort)
+  - Details:
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+<!--
+This is your opportunity to define what "normal" quality of service looks like
+for a feature.
+
+It's impossible to provide comprehensive guidance, but at the very
+high level (needs more precise definitions) those may be things like:
+  - per-day percentage of API calls finishing with 5XX errors <= 1%
+  - 99% percentile over day of absolute value from (job creation time minus expected
+    job creation time) for cron job <= 10%
+  - 99.9% of /health requests per day finish with 200 code
+
+These goals will help you determine what you need to measure (SLIs) in the next
+question.
+-->
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+<!--
+Pick one more of these and delete the rest.
+-->
+
+- [ ] Metrics
+  - Metric name:
+  - [Optional] Aggregation method:
+  - Components exposing the metric:
+- [ ] Other (treat as last resort)
+  - Details:
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+<!--
+Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
+implementation difficulties, etc.).
+-->
+
+### Dependencies
+
+<!--
+This section must be completed when targeting beta to a release.
+-->
+
+###### Does this feature depend on any specific services running in the cluster?
+
+<!--
+Think about both cluster-level services (e.g. metrics-server) as well
+as node-level agents (e.g. specific version of CRI). Focus on external or
+optional services that are needed. For example, if this feature depends on
+a cloud provider API, or upon an external software-defined storage or network
+control plane.
+
+For each of these, fill in the following—thinking about running existing user workloads
+and creating new ones, as well as about cluster-level services (e.g. DNS):
+  - [Dependency name]
+    - Usage description:
+      - Impact of its outage on the feature:
+      - Impact of its degraded performance or high-error rates on the feature:
+-->
+
+### Scalability
+
+<!--
+For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them.
+
+For beta, this section is required: reviewers must answer these questions.
+
+For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field.
+-->
+
+###### Will enabling / using this feature result in any new API calls?
+
+<!--
+Describe them, providing:
+  - API call type (e.g. PATCH pods)
+  - estimated throughput
+  - originating component(s) (e.g. Kubelet, Feature-X-controller)
+Focusing mostly on:
+  - components listing and/or watching resources they didn't before
+  - API calls that may be triggered by changes of some Kubernetes resources
+    (e.g. update of object X triggers new updates of object Y)
+  - periodic API calls to reconcile state (e.g. periodic fetching state,
+    heartbeats, leader election, etc.)
+-->
+
+###### Will enabling / using this feature result in introducing new API types?
+
+<!--
+Describe them, providing:
+  - API type
+  - Supported number of objects per cluster
+  - Supported number of objects per namespace (for namespace-scoped objects)
+-->
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+<!--
+Describe them, providing:
+  - Which API(s):
+  - Estimated increase:
+-->
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+<!--
+Describe them, providing:
+  - API type(s):
+  - Estimated increase in size: (e.g., new annotation of size 32B)
+  - Estimated amount of new objects: (e.g., new Object X for every existing Pod)
+-->
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+<!--
+Look at the [existing SLIs/SLOs].
+
+Think about adding additional work or introducing new steps in between
+(e.g. need to do X to start a container), etc. Please describe the details.
+
+[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
+-->
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any components?
+
+<!--
+Things to keep in mind include: additional in-memory state, additional
+non-trivial computations, excessive access to disks (including increased log
+volume), significant amount of data sent and/or received over network, etc.
+This through this both in small and large cases, again with respect to the
+[supported limits].
+
+[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
+-->
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+<!--
+Focus not just on happy cases, but primarily on more pathological cases
+(e.g. probes taking a minute instead of milliseconds, failed pods consuming resources, etc.).
+If any of the resources can be exhausted, how this is mitigated with the existing limits
+(e.g. pods per node) or new limits added by this KEP?
+
+Are there any tests that were run/should be run to understand performance characteristics better
+and validate the declared limits?
+-->
+
+### Troubleshooting
+
+<!--
+This section must be completed when targeting beta to a release.
+
+For GA, this section is required: approvers should be able to confirm the
+previous answers based on experience in the field.
+
+The Troubleshooting section currently serves the `Playbook` role. We may consider
+splitting it into a dedicated `Playbook` document (potentially with some monitoring
+details). For now, we leave it here.
+-->
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+###### What are other known failure modes?
+
+<!--
+For each of them, fill in the following information by copying the below template:
+  - [Failure mode brief description]
+    - Detection: How can it be detected via metrics? Stated another way:
+      how can an operator troubleshoot without logging into a master or worker node?
+    - Mitigations: What can be done to stop the bleeding, especially for already
+      running user workloads?
+    - Diagnostics: What are the useful log messages and their required logging
+      levels that could help debug the issue?
+      Not required until feature graduated to beta.
+    - Testing: Are there any tests for failure mode? If not, describe why.
+-->
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+## Implementation History
+
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+## Alternatives
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
+
+## Infrastructure Needed (Optional)
+
+<!--
+Use this section if you need things from the project/SIG. Examples include a
+new subproject, repos requested, or GitHub details. Listing these here allows a
+SIG to get the process for these resources started right away.
+-->
--- a/keps/sig-scheduling/5690-dra-preemption/README.md
+++ b/keps/sig-scheduling/5690-dra-preemption/README.md
@@ -65,7 +65,7 @@ If none of those approvers are still appropriate, then changes to that list
 should be approved by the remaining approvers and/or the owning SIG (or
 SIG Architecture for cross-cutting KEPs).
 -->
-# KEP-NNNN: Your short, descriptive title
+# KEP-5690: DRA Preemption
 
 <!--
 This is the title of your KEP. Keep it short, simple, and descriptive. A good
@@ -434,7 +434,7 @@ being scheduled.
 It will be tested by bringing up a KinD cluster and enabling the feature gate.
 We will then run a simple workload with higher priority pods and then enable
 the feature gate to verify that the lower priority pods are preempted. We will then 
-disable the feature gate and verify that anotherpod with an even higher priority
+disable the feature gate and verify that another pod with an even higher priority
 does not preempt the existing pod.
 
 ###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
@@ -450,30 +450,18 @@ pods that are referencing ResourceClaims.
 
 ###### How can someone using this feature know that it is working for their instance?
 
-- [ ] Events
-  - Event Reason: 
-- [ ] API .status
-  - Condition name: 
-  - Other field: 
-- [ ] Other (treat as last resort)
-  - Details:
+When a Pod is preempted, it can be observed in the following ways:
+
+- [x] Events
+  - Event Reason: Preempted
+- [x] API .status
+  - Condition name: DisruptionTarget
+
+Seeing this on a Pod that has references ResourceClaims whows that the feature is working.
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
-<!--
-This is your opportunity to define what "normal" quality of service looks like
-for a feature.
-
-It's impossible to provide comprehensive guidance, but at the very
-high level (needs more precise definitions) those may be things like:
-  - per-day percentage of API calls finishing with 5XX errors <= 1%
-  - 99% percentile over day of absolute value from (job creation time minus expected
-    job creation time) for cron job <= 10%
-  - 99.9% of /health requests per day finish with 200 code
-
-These goals will help you determine what you need to measure (SLIs) in the next
-question.
--->
+Kubernetes does not have SLOs for preemption, so neither will this enhancement.
 
 ###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
 
@@ -481,19 +469,13 @@ question.
 Pick one more of these and delete the rest.
 -->
 
-- [ ] Metrics
-  - Metric name:
-  - [Optional] Aggregation method:
-  - Components exposing the metric:
-- [ ] Other (treat as last resort)
-  - Details:
+- [x] Metrics
+  - Metric name: `scheduler_preemption_victims` for pod that references ResourceClaims
+  - Components exposing the metric: kube-scheduler
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
-<!--
-Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
-implementation difficulties, etc.).
--->
+No
 
 ### Dependencies
 
@@ -503,20 +485,7 @@ This section must be completed when targeting beta to a release.
 
 ###### Does this feature depend on any specific services running in the cluster?
 
-<!--
-Think about both cluster-level services (e.g. metrics-server) as well
-as node-level agents (e.g. specific version of CRI). Focus on external or
-optional services that are needed. For example, if this feature depends on
-a cloud provider API, or upon an external software-defined storage or network
-control plane.
-
-For each of these, fill in the following—thinking about running existing user workloads
-and creating new ones, as well as about cluster-level services (e.g. DNS):
-  - [Dependency name]
-    - Usage description:
-      - Impact of its outage on the feature:
-      - Impact of its degraded performance or high-error rates on the feature:
--->
+No
 
 ### Scalability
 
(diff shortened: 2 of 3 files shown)


keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+see a corresponding increase in pods being scheduled, we should investigate. It would suggest
+that pods are being preempted incorrectly and the higher-priority pods are not actually
+being scheduled.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+It will be tested by bringing up a KinD cluster and enabling the feature gate.
+We will then run a simple workload with higher priority pods and then enable
+the feature gate to verify that the lower priority pods are preempted. We will then 
+disable the feature gate and verify that anotherpod with an even higher priority
```
pohly:
```suggestion
disable the feature gate and verify that another pod with an even higher priority
```

pohly (approved):
KEP looks reasonable to me. The purpose is clear and it describes well what can be done.

The devil will be in implementing this correctly - lot's of details!

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-NNNN: Your short, descriptive title
```
johnbelamaric:
```suggestion
# KEP-5690: DRA Preemption
```

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+  the request of a higher-priority pod. We need to make sure the remaining capacity on a device is correctly
+  tracked during preemption simulations, and that we don't overcommit.
+* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
+  sure this is handled correctly during preemption simulations.
+
+Features/scenarios that we will not support:
+* **ResourceClaims that span multiple nodes**: The default preemption algorithm only considers preempting
+  pods on a single node, meaning that it will not be able to free up devices that are allocated to claims
+  referenced by pods spanning multiple nodes.
+* **ResourceClaims for PodGroups**: These claims are allocated to a PodGroup, meaning that all pods in the
```
johnbelamaric, line 205: @nojnhuh have you looked at how we can integrate PodGroup RCs with https://github.com/kubernetes/enhancements/issues/5710 ?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+It is enabled on a cluster-level, so if it is enabled, it is in use. It will only impact
+pods that are referencing ResourceClaims.
```
johnbelamaric, line 448: Is there a metric to show it is happening? Is it worth decorating any existing metric?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [ ] Events
```
johnbelamaric: Is there some event that fires when preemption happens? If so, then the answer is "this event happens on a pod, even when that pod is using RCs".

keps/sig-scheduling/5690-dra-preemption/kep.yaml
```diff
@@ -0,0 +1,40 @@
...
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.37"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.37"
```
johnbelamaric: Since there is no API, in theory we can go straight to default-on beta. I think this could be possible if:

1) We can show that this code ONLY kicks in when a pod has RCs.
2) We have really good tests.
3) Of course the FG can be turned off

keps/prod-readiness/sig-scheduling/5690.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 5690
+alpha:
+  approver: "@johnbelamaric"
```
wojtek-t: I'm happy to take that too.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,674 @@
...
+  may require preempting multiple lower-priority pods that collectively occupy fractional partitions
+  (represented as counters) on the same physical device, so that the raw capacity can be consolidated
+  and re-partitioned.
+* **Consumable Capacity**: We might need to free up just a subset of the capacity on a device to satisfy
+  the request of a higher-priority pod. We need to make sure the remaining capacity on a device is correctly
+  tracked during preemption simulations, and that we don't overcommit.
+* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
+  sure this is handled correctly during preemption simulations.
+
+Features/scenarios that we will not support:
```
wojtek-t, line 201: FWIW - we already do have PodGroup-oriented preemption implemented here:
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L455
invoked from:
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/schedule_one_podgroup.go#L219
We want to graduate workload-aware preemption to Beta in 1.37:
https://github.com/kubernetes/enhancements/pull/6106
so there are places which DRA-preemption should integrate with.

That being said, I think it makes sense to focus on single-node devices first (given that we will be making non-trivial changes in WAP in 1.37 to avoid conflicts). And expand to multinode devices as a potential followup.

wojtek-t: /cc @Argh4k

keps/prod-readiness/sig-scheduling/5690.yaml
johnbelamaric: that'd be great, thanks @wojtek-t

keps/sig-scheduling/5690-dra-preemption/README.md
nojnhuh: @wojtek-t mentioned in https://github.com/kubernetes/enhancements/pull/6113#discussion_r3316000320 that #5710 is keeping this in mind so I'm not too worried about being backed into a corner by the time we're ready to integrate these features. I'll keep tabs on the WAS pieces.

keps/sig-scheduling/5690-dra-preemption/README.md
Argh4k: I think that support for WAP and single/podgroup devices are actually two different things.  

In the end in WAP we will do the following: 
- remove all potential victims via NodeInfo.RemovePod
- run PodGroupSchedulingCycle (creates fresh cycle state based on nodes with victims removed)
- for each victim try to reprieve it via Reprieve method
- if Reprieve returns true, add back the node via NodeInfo.AddPod
- at the end preempt all victims that failed Reprieve

From DRA perspective, IIUC, if we remove pods from the "scheduler" view of the cluster via NodeInfo.RemovePod, it won't be actually reflected by the DRAManager as manager gathers resource claims allocated state based on its informer cache. 

That means that in WAP model from beta proposal, even when we remove pods with resource claims that could unblock the preemptor, preemptor will not be able to use them.

We would need a way to create an allocatedState that also includes pods removals. Maybe we could use for that podGroupState (store in it information about all removals and put that on top of allocatedState?).

Another option would be to actually keep the cycle state of pods from pod group and pass it to WAP. Then during victim removal we would run PreFilterExtensions.RemovePod for each pod from pod group and later reuse the cycle state in PodGroupSchedulingCycle. This would require a major (?) refactoring of the schedule_one_podgroup as everything there is tightly coupled. 

Nevertheless, I think that for this proposal we can aim at support single pod cycle only as I do not think that we will have the capacity to implement: this one, WAP beta update and  marry them together.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: There are two metrics that are particularly relevant for this, `scheduler_preemption_attempts_total` and `scheduler_preemption_victims`. But they don't currently don't have any labels and it is not clear to me if adding information about the pods as labels on metrics is a good solution. At the very least, it doesn't seem scalable.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Yeah, I've filled out this section.

keps/sig-scheduling/5690-dra-preemption/kep.yaml
mortent: I think all of those are doable. So I have updated the KEP to target direct to beta. I assume that in the end it will come down to whether we feel confident enough in the implementation to go through with it.

keps/prod-readiness/sig-scheduling/5690.yaml
mortent: I have updated it.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,577 @@
...
+scheduling requirements of higher-priority tasks.
+
+### Goals
+* Enable the scheduler to preempt pods consuming node-local DRA devices.
+
+### Non-Goals
+* Support preemption of workloads utilizing PodGroup-level ResourceClaims.
+* Support preemption of workloads using multi-node or network-attached devices.
+
+## Proposal
```
troychiu, line 179: I am curious about how deallocation would work. If I understand correctly, the current `ResourceClaim` controller asynchronously checks if the resource claim is free for deallocation. This means a pod triggering eviction may not be able to be scheduled immediately, as it has to wait for the `ResourceClaim` controller to complete.

johnbelamaric: That's an interesting point. While deallocation is generally handled by the RC controller, I don't know of any real reason the scheduler couldn't do it in this case. What about shared claims that are node-local? We can't deallocate those unless all the pods are being preempted. @pohly

pohly: There's no reason why the scheduler shouldn't update the ResourceClaim status to remove an unused allocation. It has to be prepared for that not succeeding when something adds to ReservedFor in the meantime (shouldn't happen, but the API allows it), but that's business as usual (any API call can fail).

Why would the scheduler do this? Will the pod then still go through the same flow of unschedulable + informer events? Beware that directly after an Update call, the informer cache used for the next scheduling attempt may still have the status from before the Update, unless the scheduler waits for the informer event or we do more tricks with the assume cache.

johnbelamaric: It may not be necessary. I am not really sure how pre-emption works internally in the scheduler. Once it finds viable preemption candidates, what does it do? Does it actually schedule the new Pod and evict the other Pods? Or does it set nominated node? I think the point is - will waiting for the RC controller to do the deallocation result in a race condition?

troychiu: My understanding is that the scheduler evicts victim pods and then reschedules the high-priority pod once the victim pods terminate. This introduces a potential race condition where the high-priority pod might be scheduled either before or after the RC controller performs the deallocation.

mortent: Good question. So I think there is a potential problem here, but not for the basic scenario where maybe the resourceclaim controller is a bit slow.

So what happens is that once a high priority preemptor pod fails to get schedule, the scheduler will trigger the preemption logic. That can find a lower priority victim pod that, if removed, would allow the preemptor pod to be scheduled. Once the scheduler has found a victim pod, it will terminate it, set the `nominatedNode` field on the preemptor pod and then put the preemptor pod in the unschedulable queue. It will only retry scheduling of the preemptor pod when the scheduler gets one of the queuing hints that are considered relevant for the dynamicresources plugin, since that is the plugin that blocked the scheduling of the preemptor pod in the first place. The hint we are waiting for here, is an update on the resourceclaim referenced by the pod. So essentially rescheduling will not be attempted until there is a signal that rescheduling is likely to succeed and deletion/termination of a pod is not one of those hints.

However, there are several hints that can trigger rescheduling, including updates the DeviceClass, ResourceSlice and [more](https://github.com/kubernetes/kubernetes/blob/f92c74d803de1109f0d9a55b966b0d0061ed0dc7/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L250-L262) and also after a timeout (I think it is 5 minutes). If this happens, and the resourceclaim controller has not yet deallocated the claim, the pod will still not be schedulable, and will trigger another preemption attempt. I think the consequence here is that if the resourceclaim controller fails to deallocate claims, it can trigger a cascade of preemptions of lower priority pods.

I think maybe doing the deallocation through the scheduler might address this, but I'm not yet sure of the consequences.

mortent:
@wojtek-t Do you have any concerns about the PRR section of this KEP?

keps/sig-scheduling/5690-dra-preemption/README.md
johnbelamaric: I think we need input from @kubernetes/sig-scheduling-approvers as to the right approach here.

keps/sig-scheduling/5690-dra-preemption/README.md
Argh4k: So, we reached an agreement in WAP KEP () that we want to rerun the whole pod group scheduling algorithm within WAP. So the option of keeping the cycle state is out. This means that we will do the following:
- remove all potential victims via NodeInfo.RemovePod
- run PodGroupSchedulingCycle (creates fresh cycle state based on nodes with victims removed)
- for each victim try to reprieve it. Reprieving will be done by running Filter plugins on all preemptor pods on nodes where they are supposed to run, with the victim added back onto its original node 
-  if Reprieve returns true, add back the node via NodeInfo.AddPod

So from this KEP, my main question is whether the PreFilter method of DRA is implemented in such a way that it detects that some pods holding resource claims are actually not in the NodeInfos? And if yes, whether PreFilterExtension for DRA plugin will work on top of that?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,577 @@
...
+
+None known at this point.
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+There are no SLOs for this feature.
+
+## Implementation History
+
+1.33: first KEP revision and implementation directly to Beta
```
dom4ha:
```suggestion
1.37: first KEP revision and implementation directly to Beta
```

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,577 @@
...
+scenarios will be handled by integration tests.
+
+### Graduation Criteria
+
+- Feature implemented behind a feature flag
+- Initial e2e tests completed and enabled
+
+#### Beta
+
+- Gather feedback
```
dom4ha: The feature goes straight to beta, so it won't be possible to gather feedback :)

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: As @mortent said, the preemptor pod will wait in the unschedulable queue until the cluster events wakes it up. So it may be victim pod deletion or ResourceClaim deallocation. The mechanism that prevents scheduler from making further preemptions is base on NNN and it's a heuristic, so unfortunately the concern is valid.

The mechanism is implemented under `PodEligibleToPreemptOthers` https://github.com/kubernetes/kubernetes/blob/d88c48520ecd5a3bd9d0c9e90a48bcfca6f2ae11/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L364-L365

1. If the preemptor is `UnschedulableAndUnresolvable` on the node, it means there is no point waiting for preemption process to finish and new preemption will be triggered
2. If the preemptor is just `Unschedulable` and there are still pods terminating with "Preemption" in reason, the preemption waits.

So despite all pods terminated, the preemptor may remain `Unschedulable` (due to ResourceClaim not being deallocated) and may trigger yet another preemption as there are no pods on the node in terminated state. 

Deallocation as a part of the preemption process most likely won't help either as some pods needs to honor the grace period, so we cannot deallocate claims earlier.

I think we may need to extend the NNN mechanism to not only nominate a candidate Node, but also "nominate" DRA devices. It would prevent other ResourceClaim from stealing the devices for which the preemptor pod (ResourceClaim) is waiting for. The DRA plugin would have to internally maintain a map from devices to nominator ResourceClaims, similar to the nominator mechanism in the scheduler cache. Based on that, the check in point 2 would be extended to wait until all nominated devices are released.

dom4ha (commented):
Simulating the pod add and removal in DRA plugin should be fairly easy to achieve, but preventing secondary preemptions seems to be a bigger problem which may require much more in depth design.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,577 @@
...
+* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
+  sure this is handled correctly during preemption simulations.
+
+Features/scenarios that we will not support:
+* **ResourceClaims that span multiple nodes**: The default preemption algorithm only considers preempting
+  pods on a single node, meaning that it will not be able to free up devices that are allocated to claims
+  referenced by pods spanning multiple nodes.
+* **ResourceClaims for PodGroups**: These claims are allocated to a PodGroup, meaning that all pods in the
+  group must be scheduled together. The default preemption algorithm does not consider the PodGroup at all,
+  so we will not allow preemption of pods that reference such claims.
```
macsko: For workload-aware preemption, AddPod/RemovePod will also be used, so this KEP cannot say that the "default preemption algorithm does not consider the PodGroup at all" or "default preemption algorithm only considers preempting pods on a single node", because for workload-aware preemption the default preemption algorithm may consider such cases. It would be good to explicitly emphasize that preemption support for such RCs will be restricted on the DRA plugin level.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: @Argh4k The dynamicresources plugin does not check the NodeInfo slice in the `PreFilter` implementation, so it will not discover if any pods were removed. It fetches the state from the DRAManager, which is a snapshot of the live state of the cluster. Is this a problem? It might be possible to fix this by reconciling the state from the DRAManager with the nodes available from NodeInfo, but I'm not yet sure if we have the necessary information to do this without larger changes.

In my WIP implementation of this KEP, I store some of the information from the DRAManager in the stateData struct so it can be updated as `AddPod` and `RemovePod` are called during the preemption simulations: https://github.com/kubernetes/kubernetes/pull/139266. I think that aligns with what you are suggesting here, where the dynamicresources plugin needs to handle that some pods are removed during simulation of the pod group scheduling. Am I understanding this correctly?

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Right, so the basic solution would work for simple use-cases for DRA, but I don't think it will work for more complicated features like Partitionable Devices and Consumable Capacity. In both cases, it is not as simple as a device being deallocated from one claim and allocated to another, but rather that some capacity are released from one claim which will allow another claim to be allocated. So we want to prevent other claims from taking the capacity (rather than just the device) and we are not necessarily waiting for the device we need to be deallocated, but rather for one or more other devices to be deallocated so the capacity is freed up.

Could we solve this by nominating devices, but rather than checking from them to be released, we set the criteria as the claims referenced by the preempted pod has been deallocated. Although, we have to be careful with this, since a pod might be using a shared claim that won't be deallocated, but also doesn't need to be in order for the preemptor pod to schedule.

And for the implementation, are you suggesting that implementing this will require changes to the scheduling framework outside of the dynamicresources plugin (i.e. changes to the `PodEligibleToPreemptOthers` function) which would make the framework aware of DRA? Or are you thinking that this is something that we can implement in the dynamicresources plugin, possible as part of the queuingHint functions?

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Good point. Updated.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Yeah, I've updated the language here.

wojtek-t:
> @wojtek-t Do you have any concerns about the PRR section of this KEP?

We need to figure out the design first, and I believe that both:
1) Add/Remove Pod: https://github.com/kubernetes/enhancements/pull/6113#discussion_r3316000320
2) avoiding unnnecessary preemptions: https://github.com/kubernetes/enhancements/pull/6113#discussion_r3336545353

should first be addressed in the KEP before we even get to PRR review

keps/sig-scheduling/5690-dra-preemption/README.md
Argh4k: Well, yes and no. I took a look at your WIP implementation and this will work with the current pod by pod scheduling cycle. 

The problem is that in preemption for Workload Aware scheduling we want to: remove all potential victims (NodeInfo.RemovePod) and rerun the scheduling algorithm (including rerunning prefilter). So, we will not call `RemovePod` but instead we expect that plugins should be able to create a CycleState reflecting the state of the cluster based on NodeInfos passed in PreFilter. So I think it boils down to understanding whether

> It might be possible to fix this by reconciling the state from the DRAManager with the nodes available from NodeInfo, but I'm not yet sure if we have the necessary information to do this without larger changes.

is actually feasible. 

I think that for Alpha it's not a big problem that we will support only pod by pod cycle, but for beta we will need to have support for WAS. @wojtek-t ?

wojtek-t: Given that the primary bottlenecks is review bandwidth, I would like to avoid doing things twice and start already with something that will work for WAP (unless you can show that making it work for WAP will be pure extension - but I don't see that now).

dom4ha: @wojtek-t To support WAP, DRA plugin would have to also support PodGroup-level snapshotting, so it may be a bigger change.

Currently the plugin snapshots DRA state (DRAManager) only for a single pod scheduling cycle (not PodGroup scheduling cycle), so the state is constantly updating after each pod sub-cycle.

The problem results from the fact that DRA has its own state even thought the plugins should be stateless, so there is no (yet) an extension point that would allow the plugin to follow PodGroup scheduling cycle boundaries (when it starts and ends) and assure scheduling consistency, required to assure WAP and PodGroups algorithm correctness.

@mortent What @Argh4k is saying is that the framework will not call `PreFilterExtensions.RemovePod`, but `NodeInfo.RemovePod` on PodGroup level snapshot, but it will call `PreFilterExtensions.AddPod` on reprieve, so in fact there DRA plugin does not need to implement `PreFilterExtensions.RemovePod` at all, but it should be able to rebuild its state from the PodGroup level snapshot.

mortent: So I think this means that the dynamicresources plugin must be able to look at the information in the DRAManager and NodeInfos and determine the removed Pods and the impact that has on the available devices and their capacity. I will take a look at the code today and try to come up with a proposal for how we can implement this.

mortent: So I looked at the code and I definitely think it is doable for the dynamicresources plugin to rebuild the state in PreFilter based on information in `NodeInfos`:
* Capture the UIDs of all remaining pods (i.e. not removed by the `NodeInfo.RemovePod` calls) by iterating over the provided `nodeInfos` and the referenced pods. 
* Iterate over all the claims from `DRAManager.ResourceClaims` and determine which should be simulated as released by since none of the pods referenced in their `.status.reservedFor` list remain in the simulation.

This should let us track the state in a similar way to what I have in my POC for the pod-by-pod scheduling cycle.

@Argh4k Does the new WAP implementation replace the current one, meaning we no longer need to implement `PreFilterExtensions.RemovePod`? Does this impact other parts of the preemption lifecycle, meaning does it change when/if the cyclestate is cloned?

@dom4ha What is the consequence of PodGroup-level shapshotting, in particular what changes do we need to make to the DRAManager to handle it?

dom4ha: > @dom4ha What is the consequence of PodGroup-level shapshotting, in particular what changes do we need to make to the DRAManager to handle it?

It's mainly important to support WAS itself, as the PodGroup scheduling algorithm assumes the cluster state does not change during (what we call) the workload scheduling cycle, so we do snapshot the cluster state at the beginning of the cycle and perform scheduling simulation of all PodGroups' pods. After each pod scheduling sub-cycle, the pod is assumed before the next sub-cycle starts. Later the pods may be reverted and scheduled again, but everything needs to happen within a consistent cluster state. 

Preemption (WAP) is a part of this cycle and (as @Argh4k said) reruns the scheduling algorithm after removal all eligible victim pods. Performing it under a consistent cluster state is essential for correctness. We also need to store cycle states of sub-cycles to perform simulations like `PreFilterExtensions.AddPod/RemovePod`.

So we should think how the DRA state could be snapshotted. It could either be migrated to the framework or snapshotted inside the plugin. But to achieve that, the framework would have to define new extension points defining the workload scheduling cycles boundaries.

Initially I was thinking that DRA plugin could just implement `PreFilterExtensions.AddPod/RemovePod` and support preemption for a single pod, but the issue with resource nomination seems to be a bigger blocker.

mortent:
We have made the decision to postpone this until 1.38.

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: > So we should think how the DRA state could be snapshotted. It could either be migrated to the framework or snapshotted inside the plugin. But to achieve that, the framework would have to define new extension points defining the workload scheduling cycles boundaries.

I believe that we should invest more time to think how we can migrate it to the framework. I acknowledge the fact that there are differences wrt DRA resources, but if we don't unify it, it will be hitting us again and again and we will need more and more ducktaping over time and reasoning about the whole scheduling will just be harder.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+When the preemption logic selects a winning candidate for a preemptor pod, the `dynamicresources`
+plugin records a *claim nomination* for that preemptor, consisting of:
+
+* the nominated node;
+* the set of ResourceClaims that the preemption simulation released in order to make the placement
+  feasible;
+* the allocation result of each of those claims;
+* an expiry time.
+
+The set of claims is taken directly from the simulation, which already computes it in order to
```
dom4ha, line 337: The DRA plugin does not have any visibility to the set of victims that were identified (it doesn't know when the simulation ends), so it probably cannot construct the list. I don't think this is even necessary, but there is a separate comment.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+grace period.
+
+Erring on the long side is intentional. A nomination that outlives its usefulness holds capacity for
+a pod that is still pending and still wants it, and that capacity remains available to pods of
+higher priority. A nomination that expires while the preemptor is still waiting removes both the
+capacity hold and the deferral at once, so the preemptor preempts a second set of victims and
+creates more work for the controller whose slowness caused the expiry. The `expired` outcome on
+`scheduler_dra_claim_nominations_total` will show whether five minutes is the right value.
+
+When a nomination expires the plugin requeues the preemptor, so that a pod whose wake-ups were
```
dom4ha, line 368: Except from requeueing the preemptor, we need to requeue other workload waiting for devices that were freed up. This is achieved by running the same wake-up mechanism as if the preemptor pod was removed, but this is triggered when nominatedNodeName is removed.

The important detail is that rescheduled preemptor may actually land on the same node, but pick different devices, so this mechanism may not work without remembering what exactly allocation was nominated (there's a separate comment about it)

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+withheld while the nomination was unsettled does not then wait for the periodic flush of the
+unschedulable queue.
+
+Nominations are held in memory in the `dynamicresources` plugin and are keyed by the preemptor's pod
```
dom4ha, line 372: This is fine for the first implementation, but eventually they should be persisted in api-server. Note that autoscalers (and potentially other external systems) run simulations that should consider resources being blocked to not take conflicting decisions. Since they cannot guess what exact resources were identify, they can only simulate it without guarantee the results will match.

I wouldn't call it beta blocker, as resource nomination seems to be a feature in itself (useful in other scenario), although preemption depend on it.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+controller rather than by the removal of the pod itself; DRA is simply the first case where the
+interval is long enough to matter in practice.
+
+### Claim nomination
+
+When the preemption logic selects a winning candidate for a preemptor pod, the `dynamicresources`
+plugin records a *claim nomination* for that preemptor, consisting of:
+
+* the nominated node;
+* the set of ResourceClaims that the preemption simulation released in order to make the placement
```
dom4ha, line 332: The nomination should mainly remember the the preemptor's ResourceClaim allocation of a pod that has nominatedNodeName set, which is a aligned with how node nominations work.

It's because we need to stick to a ResourceClaim allocation that was determined during preemption scheduling and treat these resources as if they were in use to guard them from being allocated by other pods.

Storing the list of victim ResourceClaims does not give such guarantee, unless we want to reserve resources of all victim claims, but that would be a way too much. Consider a case when a pod needs to allocate a portion of a device that is going to be deallocated, but it should not prevent other pods from allocating the remaining capacity once it's freed up.

The mechanism that the scheduler uses for that is AddPod/RemovePod, which not only simulates ResourceClaim victims deallocation (RemovePod), but also simulates allocation of nominated resources to treat them as potentially in-use (AddPod).

We indeed may also need a list of victim ResourceClaims that are supposed to be deallocated as an opposite mechanism to nomination (victims nomination). This information should be used to determine devices for which there is an ongoing preemption. It's a similar to the mechanism of annotating pods with `deletionTimestam` (grace period) to avoid starting new preemption as long as any required device is supposed to be deallocated. But for that, I'd rather annotate victim claims without coupling it with information which preemptor triggered it (it's not important)

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+
+## Motivation
+Acquisition and allocation of specialized hardware (e.g., GPUs and TPUs) is a primary operational concern for
+AI/ML users. Because accelerator resources are frequently constrained, it is essential that Kubernetes
+provides robust tools to run high-priority workloads on the available hardware. Introducing preemption
+support for DRA ensures that the scheduler can reclaim resources from lower-priority workloads to satisfy
+scheduling requirements of higher-priority tasks.
+
+### Goals
+* Enable the scheduler to preempt pods consuming node-local DRA devices, in the pod-by-pod
```
dom4ha, line 193: How we are going to enforce it? Does the plugin know whether a device is node local or not?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+  a nominated device has been freed, so there is no reliable point at which to release the hold. A
+  claim's `Status.Allocation` being cleared is directly observable.
+
+Claims do not have these problems, because the claim is the unit at which capacity is actually
+released. "These claims will be released" remains true and checkable regardless of how the
+underlying capacity is subsequently recombined.
+
+### Deferred: workload-aware preemption
+
+Workload-aware preemption ([KEP-5710]) removes candidate victims by mutating the scheduler snapshot
```
dom4ha, line 494: WAP also changes the preemption scope from node-local to cluster-global. On one hand it is desired to be able to preempt pods on a different node if only they block scheduling of a higher priority pod.

But the second part of the story is nomination mechanism, which is supposed to prevent other pods/claims from using resources reserved for preemptor. Currently scheduler applies nominations in filtering phase on each node separately, but in case of multi-node DRA claims, those nomination may need be applied globally, before filtering starts.

It's a big change to the scheduler. It's not blocking alpha, but blocks WAP global preemption adoption in DRA.

/cc @wojtek-t @mm4tt @macsko @Argh4k

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Every device has an associated node selector, which can be one of `nodeName`, `nodeSelector` or `allNodes`. This can be either set directly on a device, or it is "inherited" from the `ResourceSlice` on which the device is defined. So in this case, we can say that we only support devices use the `nodeName` node selector. That can in theory leave out devices that have `nodeSelector` defined, but only end up selecting a single node, but I think we can ignore that case.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: So if I understand you correctly here, you are suggesting that rather than keeping track of the resourceclaims that needs to be deallocated in order for the preemptor to find the needed capacity, we keep track of which devices/capacity was allocated to the preemptor during the simulation, and then "pretend" that this is taken during future simulations, thereby preventing other pods from grabbing the devices? It seems like that could be a better solution, and I agree that reserving whole devices as in the current proposal can be wasteful.

Does the dynamicresources plugin have a way to know which ResourceClaim allocation were chosen during simulation? Or will this require that the core scheduler logic is aware of ResourceClaims?

Keeping track of victim ResourceClaims does make sense. But to clarify, I think the same applies here as earlier, that we need to keep track of the actual capacity from devices that are supposed to be deallocated, since deallocation of a claim doesn't mean that the full device will be available.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Right, I think https://github.com/kubernetes/enhancements/pull/6113/changes#r4032294371 explains the way to think about it. But it seems like the dynamicresources plugin does need to know the allocation result for the preemptor pod from the simulation.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: So the workloads that might be queuing waiting for the devices/capacity that will be freed up, where are these coming from? Are these just other workloads that failed to schedule while preemption is in progress, and that would potentially succeed once the victim's ResourceClaims have been deallocated?

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Yeah, but it seems like the interesting question is where would we persist this information? It requires more information than we do for NNN.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,999 @@
...
+and then reruns the pod group scheduling algorithm with fresh CycleStates, instead of calling
+`PreFilterExtensions.RemovePod` for each victim. A plugin that learns about removals only through
```
sohankunkerkar, line 496: Kueue hits exactly this today. We check per-node device feasibility with `structured.Allocator` and simulate preemption by removing the victims' Pods, so the allocator keeps counting their devices as allocated and the preemption frees nothing and the Workload stays pending, as described here.

One question on the deferred design: you note devices become allocatable when the resourceclaim controller deallocates, not when the victim Pods are deleted. For the workload-aware path, is the intent that the nomination record covers that interval too, or would a workload-aware scheduler be expected to track it itself?

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: Yes. The main question is whether we make the nomination mechanism plugin agnostic or plugin specific. 

- If we make it plugin specific, we need to have a new extension point to communicate nomination back to the plugin, since such plugins would be eventually responsible for persisting nomination in the api-server. Currently it's done in FailureHandler in case of preemption or in PreBind in case of resource reservation (communicating scheduler decision to external components in case of binding is delayed, like in DRA BindingConditions feature)
- in the plugin agnostic approach, we need to have a way to serialize that information, which eventually would need to be persisted in the api-server along with nominatedNodeName. In that case the internal in-memory information would be stored only in scheduler cache (nominator) and whenever nominated pods needs to be considered, the plugin would be requested to apply it to its internal tracking structured based on this serialized information (AddPod(serialized_nomination)).

Since we'd rather need to store nominations in ResourceClaim objects which are managed by DRA plugin, I'd rather add necessary extension points. In that case, the DRA plugin would have to keep information about nominations in its internal structures

It's however another good topic for WAS+DRA alignment discussion.
@wojtek-t @macsko

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: Yes. When ResourceClaim is deallocated, the standard mechanism will wake them up, but when in-memory nomination disappears, a similar mechanism (mimic pod and resources deallocation) need to wake them up.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: I think it could be in ResourceClaim status.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: > So if I understand you correctly here, you are suggesting that rather than keeping track of the resourceclaims that needs to be deallocated in order for the preemptor to find the needed capacity, we keep track of which devices/capacity was allocated to the preemptor during the simulation, and then "pretend" that this is taken during future simulations, thereby preventing other pods from grabbing the devices?

Yes

> Does the dynamicresources plugin have a way to know which ResourceClaim allocation were chosen during simulation? Or will this require that the core scheduler logic is aware of ResourceClaims?

There is separate comment about it

> Keeping track of victim ResourceClaims does make sense. But to clarify, I think the same applies here as earlier, that we need to keep track of the actual capacity from devices that are supposed to be deallocated, since deallocation of a claim doesn't mean that the full device will be available.

Note that for pod preemption we don't store victim nominations in-memory at all. Scheduler does not persist it either, just go straight to actuation (remove pod). In effect of actuation, the api-server sets deletionTimestamp on pods which require waiting for their grace period or inform about immediate pod deletion. So what we currently have is prone to races in scheduler.

We have to fix it in case of ResourceClaim preemption implementation (set in-memory with timeout), for the reasons we identified in 1.37 (after pods removal, some other pods will be identified as victims causing a chain reaction).

I believe that in-memory tracking should be enough, but it's not clear whether it should be within plugin or plugin agnostic (see other comment)

keps/sig-scheduling/5690-dra-preemption/README.md
sohankunkerkar: cc @mortent @dom4ha, could you shed some light here? This is really important for TAS + DRA in Kueue.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: I have updated the proposal to reflect the discussions, including adding a new `PreemptionExtensions` interface that gives the scheduler a way to inform the dynamicresources plugin the result of a simulation.

wojtek-t: I still need to re-read the whole proposal now, but quickly commenting here.

I agree with the two potential options described by Dominik above.
But I actually don't really agree with the recommendations - I think the second option is more cohesive.

The way I think about nominator is that is it should handle all the information related to nominations. I acknowledge the fact that it may need to be persisted in ResourceClaims, but in the end it's about nominations.

So we either agree to make DRA a first-class citizen in scheduler, in which case nomination is a single concept (whether we're nominating node, device or whatever else) - whether it's DRA-related or not it doesn't matter.

Or we build two parallel solution for different things in scheduler - which I'm fairly against it.


I will provide more comprehensive comments overall tomorrow, sorry for delay...

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: This is a known issue and it's one of the reasons why WAS is out of scope of the DRA preemption KEP in 1.38. We need to align first whether we allow plugins to maintain their state or whether DRA state should be moved to the core.

When WAP simulates Pod removals, it does so in it's core scheduler cache, but does not impact DRA state. The AddPod/RemovePod extension points does not cover this operation, so they are not called and this enhancement won't fix it in 1.38.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: > So we either agree to make DRA a first-class citizen in scheduler, in which case nomination is a single concept (whether we're nominating node, device or whatever else) - whether it's DRA-related or not it doesn't matter.

I think it boils down to the question where DRA plugin state lives (on which we need to get the generic alignment). Keeping nominations in the scheduler Nominator does not make much sense if scheduler is not capable of interpreting it (tracking resource usage when we simulate Add or Remove resource).

 See also the comment https://github.com/kubernetes/enhancements/pull/6113#discussion_r4043335090

mortent: Thanks for the feedback here. The design choice here seems to be more on the scheduler side than DRA, but I'm happy to contribute.

keps/sig-scheduling/5690-dra-preemption/kep.yaml
```diff
@@ -1,40 +1,43 @@
...
+
 see-also:
-  - "/keps/sig-scheduling/4815-dra-partitionable-devices"
   - "/keps/sig-node/4381-dra-structured-parameters"
-  - "/keps/sig-scheduling/5075-dra-consumable-capacity"
+  - "/keps/sig-scheduling/5710-workload-aware-preemption"
 
 # The target maturity stage in the current dev cycle for this KEP.
-stage: alpha
+stage: beta
```
wojtek-t: Didn't we want to start with Alpha?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+capabilities to ensure that the most critical workloads gain access to scarce hardware resources.
+
+Simulating the removal of victims is not sufficient on its own. The devices held by preempted pods
+are reclaimed asynchronously by the resourceclaim controller, so there is a window during which a
+preemptor has been promised capacity that it does not yet hold. This KEP therefore also specifies
+how the scheduler records and holds that capacity, so that a preemptor is not made to preempt
+repeatedly and its promised resources are not taken by other pods while being reclaimed.
+
+Most of this is contained in the `dynamicresources` plugin. The one exception is a new optional
+scheduler framework extension point, `PreemptionExtensions`, which notifies plugins when a
```
wojtek-t, line 182: This may need to be adjusted based on on the final outcome of this comment thread:
https://github.com/kubernetes/enhancements/pull/6113#discussion_r4062328963

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+  preemption path implemented by `DefaultPreemption`.
+* Ensure that a preemptor is not made to preempt repeatedly because the devices it freed are
+  reclaimed asynchronously, or are taken by another pod before it can be scheduled.
+
+### Non-Goals
+* Support preemption in the workload-aware preemption path. Pods belonging to a PodGroup are out of
+  scope, whether their ResourceClaims are reserved for the PodGroup or for individual pods. What
+  this would require is outlined in
+  [Deferred: workload-aware preemption](#deferred-workload-aware-preemption).
+* Support preemption of workloads using multi-node or network-attached devices.
```
wojtek-t: nit: or even individual pods using those devices, right?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+Features that we will support but require careful implementation:
+* **Partitionable Devices**: Freeing up the exact device partition requested by a higher-priority workload
+  may require preempting multiple lower-priority pods that collectively occupy fractional partitions
+  (represented as counters) on the same physical device, so that the raw capacity can be consolidated
+  and re-partitioned.
+* **Consumable Capacity**: We might need to free up just a subset of the capacity on a device to satisfy
+  the request of a higher-priority pod. We need to make sure the remaining capacity on a device is correctly
+  tracked during preemption simulations, and that we don't overcommit.
+* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
+  sure this is handled correctly during preemption simulations.
```
wojtek-t: It might be answered below, but - what does "handled correctly" mean in this context?

Scheduler can't predict what will happen with binding conditions, so what we can do with them?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+* **Pods belonging to a PodGroup**: Preemption for these goes through the workload-aware preemption
+  path, which removes victims from the simulation differently. The plugin will not free devices for
+  them, whether their ResourceClaims are reserved for the PodGroup or for individual pods. See
+  [Deferred: workload-aware preemption](#deferred-workload-aware-preemption).
+
+Simulating the removal of victims is not the whole problem. The devices freed by a preemption do not
+become allocatable when the victim pods are deleted, but later, when the resourceclaim controller
+deallocates their ResourceClaims. During that interval the scheduler can preempt further pods
+unnecessarily, and the freed devices can be taken by an unrelated pod. When a preemption candidate is
+selected, `dynamicresources` records the simulated `AllocationResult`s computed for the preemptor
```
wojtek-t: Where it record that AllocationResult?

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+together with the victim `ResourceClaim`s being released. It uses the simulated `AllocationResult`s
+in `AddPod` to hold the exact capacity the preemptor needs on the nominated node, and uses the victim
+`ResourceClaim`s to defer any further preemption by the same pod until those claims have been
+deallocated. This is specified in [Design Details](#design-details).
+
+Supporting this requires a small, generic addition to the scheduling framework: an optional
```
wojtek-t, line 258: I acknowledge lack of alignment between us - but I don't personally agree with this new extension point - see this comment thread for discussion:
https://github.com/kubernetes/enhancements/pull/6113#discussion_r4062328963

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: > I think it boils down to the question where DRA plugin state lives (on which we need to get the generic alignment). Keeping nominations in the scheduler Nominator does not make much sense if scheduler is not capable of interpreting it (tracking resource usage when we simulate Add or Remove resource).

Agree this is tightly coupled to decision where the state lives.

But I don't understand the second part - the nominator itself is not interpreting the resource usage or anything like that - it only stores the nominations (assignments). The interpretations are done by plugins themselves and running `PreFilterExtensions.AddPod()` function.

So we don't need nominator to interpret the DRA resources - we only need to store the nominations of individual RCs there and the intepretation will be done by DRA plugin.

The other comment also doesn't change my perception and so I still believe that instead of new extension point, we should rather extend the nominator to keep DRA-related information and keep nominations in a single place (rather than splitting the tightly coupled state across few different places).

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+  feasible.
+
+During `Filter` in the preemption simulation, `dynamicresources` caches the computed
+`AllocationResult`s per candidate node in its own `CycleState` entry. When `AddNominatedPod` is
+called with the scheduling cycle's `CycleState`, the winning `nodeName`, and `victims`,
+`dynamicresources` looks up the cached `AllocationResult`s for `nodeName` and derives the released
+victim claims from `victims` (including only claims whose every reserving pod is in `victims`, not
+shared claims still held by a non-preempted pod).
+
+A nomination's lifetime is bound 1-to-1 to the pod's `nominatedNodeName` in the scheduler's
```
wojtek-t, line 362: FWIW - this is another strong signal to me that this information should be stored in the Nominator - otherwise you're required to synchronize two parts of the same state across two different places.

@dom4ha

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+higher-priority pod took the freed device, or the node became unschedulable), `PodEligibleToPreempt`
+returns `true` and `DefaultPreemption` immediately re-evaluates the pod—either replacing the
+nomination with a new candidate node or clearing `nominatedNodeName`, which removes the nomination.
+Before t3, while the victim claims are still allocated, waiting without a timeout matches how
+`DefaultPreemption` behaves when a victim pod is stuck terminating: if the resourceclaim controller
+is unhealthy, expiring the nomination would only cause the preemptor to evict additional victims on
+other nodes while the controller is down.
+
+When a nomination is removed without a corresponding `ResourceClaim` deallocation event in the API
+server (for example when the preemptor is deleted or its `nominatedNodeName` is cleared or changed),
```
wojtek-t: Maybe I'm missing something and it's a dumb question, but how do you distinguish the "nomination remove without corresponding RC deallocation" from "nomination remove with RC deallocation"

Given that those two pieces of information are stored  in different objects (pod & RC), they are by definition asynchronous to each other, so by observing one you don't know if the other will happen soon after or will not happen.

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: Certainly not Alpha blocker, but even if we say that having that implemented is not beta blocker, I think that having a clear fleshed out design for that should actually be a beta blocker (to ensure that what we're doing with nominations is compatible with future plans).

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+preemptor lands on the same node but selects different devices—would remain in the unschedulable
+queue until the periodic flush.
+
+Nominations are held in memory in the `dynamicresources` plugin and are keyed by the preemptor's pod
+UID. They are not persisted; see [Risks and Mitigations](#risks-and-mitigations).
+
+A claim nomination is distinct from, but paired with, the pod's `nominatedNodeName`. The latter is an
+API field recording which node the scheduler intends to place the pod on; the former is scheduler-local
+state recording which DRA capacity on that node is being held for it and which victim claims it is
+waiting on. A claim nomination never exists without a corresponding `nominatedNodeName`.
```
wojtek-t, line 396: Yet another signal that it should be stored together (so it's part of nominator).

@dom4ha

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+
+Because the gate is disabled by default, upgrading to a release that contains this feature does not
+by itself change any behavior. An operator has to enable the gate.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes, disabling the feature will prevent the scheduler from preempting pods that
+reference ResourceClaims.
+
+Disabling it also discards any claim nominations that are currently held, which releases the
```
wojtek-t: held by by the scheduler in its memory

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+there is no state to clean up and no reconciliation is required.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+That it was previously rolled back have no impact. Reenabling it just means that
+preemption of pods referencing ResourceClaims will again be considered.
+
+###### Are there any tests for feature enablement/disablement?
+
+We will cover this scenario in both unit tests and integration tests.
```
wojtek-t: For purely in-memory features, enablement/disablement (via FG that requires component restart) is exactly the same thing as feature test.

So no additional tests are really neede here :)

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1065 @@
...
+somewhere in the implementation.
+
+If the `scheduler_preemption_victims` increases significantly when the feature is enabled, but we don't
+see a corresponding increase in pods being scheduled, we should investigate. It would suggest
+that pods are being preempted incorrectly and the higher-priority pods are not actually
+being scheduled.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+It will be tested by bringing up a KinD cluster and enabling the feature gate.
```
wojtek-t, line 852: Not strictly required for Alpha if we agree to target alpha in 1.38, but for Beta please provide a bit more details. See https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4671-gang-scheduling#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested as an example

wojtek-t (commented):
Sorry for a huge delay from my side - this is a very solid KEP and my main concern is about storing DRA nominations in DRA plugin instead of in a single place.

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: I discussed that offline with @dom4ha and @macsko 

The part where our view was disjoint is I'm living in a mental model where we will move DRA state to the framework soon and in that mode having a "joint" nominator make sense. However, we're not there yet (and we don't believe it will happen in 1.38) - with that in mind keeping it closer to where the state is seems like a better decision short-term.

To unblock this proposal, we decided to go with what's currently proposed with the big disclaimer that the proposed extension point will be alpha (or even just hard-coded internally in the framework and not exposed to plugins) for Alpha in 1.37. And in the meantime have a discussion how to move the state of DRA from the plugin to framework and revisit that decision based on the outcome of this discussion.

keps/sig-scheduling/5690-dra-preemption/kep.yaml
mortent: Yeah, we had a hope we could take this directly to beta, but it has turned out to be a more complicated feature than originally expected, so I think doing alpha first makes sense. I have updated the KEP.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: As discussed in that comment, we can introduce this extension point under the FG (tentatively), so that we can revision that decision in 1.39 without risking it will gain some other unexpected adoption.

You  can add that disclaimer somewhere (maybe at the bottom of this section)

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Yeah, this is just imprecise language. I have updated this to say `pods` rather than `workloads`.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -242,43 +245,52 @@ Features/scenarios that we will not support:
...
-controller reclaims afterwards. The extension point lets a plugin answer the question directly.
+unnecessarily, and the freed devices can be taken by an unrelated pod. When a preemption candidate is
+selected, `dynamicresources` records the simulated `AllocationResult`s computed for the preemptor
+together with the victim `ResourceClaim`s being released. It uses the simulated `AllocationResult`s
+in `AddPod` to hold the exact capacity the preemptor needs on the nominated node, and uses the victim
+`ResourceClaim`s to defer any further preemption by the same pod until those claims have been
+deallocated. This is specified in [Design Details](#design-details).
+
+Supporting this requires a small, generic addition to the scheduling framework: an optional
+`PreemptionExtensions` interface. Because `DefaultPreemption` evaluates candidate nodes concurrently
```
dom4ha, line 259: I wouldn't tight the nomination mechanism with preemption, as we already use it in other use cases which requires communication of scheduler decisions to external components. So consider naming it something like `NominationExtensions`

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: This section was meant to point out that DRA has several "advanced features" that are important to consider and not necessarily because they will require special handling. I've updated the language here.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -242,43 +245,52 @@ Features/scenarios that we will not support:
...
+  change; we defer that decision to GA, informed by Beta feedback.
+* **Claims that are never deallocated.** If the resourceclaim controller is unhealthy and fails to
+  deallocate a victim claim, its capacity remains occupied in the API server (which is standard DRA
+  behavior whenever a pod is deleted while the controller is down). The preemptor remains waiting
+  for its nominated node—matching how `DefaultPreemption` behaves when a victim pod is stuck
+  terminating—rather than timing out and evicting additional victims on other nodes while the
+  controller is unhealthy. Once the controller recovers and deallocates the claim (or if the
+  preemptor is deleted), the nomination resolves. Persistently stuck nominations are visible through
+  the `scheduler_dra_claim_nominations` gauge metric below.
 * **Multiple schedulers.** A nomination is only known to the scheduler that created it. Another
```
dom4ha, line 298: This is not much different from the first point here "Nominations are in-memory", so you can probably split that point into two. The Cluster Autoscaler is actually an example of a scheduler, so persisting the nominations will address this point as well.

> This matches the existing limitation of nominated nodes.

Nominations persisting is actually a way to address this limitation to some extent. There were even attempts to extend this mechanism to reduce possible races between different actors (like scheduler and kubelet), so DRA nominations may evolve in the future as well.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: It is described in more detailed further down, but clarified here that we are recording this in memory for the alpha version.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha, line 339: It is a slightly different case than pod preemption, since native resources are bound to the pod lifetime. The DRA resources need to be anchored to a pod, which is a subject for preemption analysis (it has a priority). So it brings a new problem that the pod may disappear earlier than the anchored resources get released.

So even though scheduler is not able to preempt DRA resources that are not anchored to any pod, it will still know that they are about to disappear. It will refrain from running another preemption, in which it would have to pick a different pod victim.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: I've updated the KEP with the disclaimer that `PreemptionExtensions` is provisional for alpha (and may be kept internal to the framework rather than exposed as a public plugin interface), and added a beta graduation criterion to revisit whether DRA state and claim nominations should move into the core framework.

Just to make sure we are on the same page: my understanding from https://github.com/kubernetes/enhancements/pull/6113#discussion_r4071987565 is that the concern here is the broader architectural question of where DRA state and nominations should live (plugin vs. core framework), rather than the mechanics of `PreemptionExtensions` itself while DRA state remains in the dynamicresources plugin. Please let me know if you also had any concerns with the interface methods themselves.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Yeah, the wording here is a bit misleading. We don't really need to know which one happens before the other, we just want to only wake up waiting pods when they can actually be scheduled (the victims claims have been deallocated). I've updated the KEP.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Fixed.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Ack. Updated the KEP to reflect this.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Ack. The KEP has been updated to target alpha, so leaving this as is for now.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Added having an agreed-upon design for persisting claim nominations in the API as a Beta graduation criterion.

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: Yes - correct. Sorry for not being precise with my comment.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -225,8 +226,10 @@ Features that we will support but require careful implementation:
 * **Consumable Capacity**: We might need to free up just a subset of the capacity on a device to satisfy
   the request of a higher-priority pod. We need to make sure the remaining capacity on a device is correctly
   tracked during preemption simulations, and that we don't overcommit.
-* **Device Binding Conditions**: Pods might be waiting for a resource binding condition, so we want to make
-  sure this is handled correctly during preemption simulations.
+* **Device Binding Conditions**: A lower-priority pod may be blocked in `PreBind` waiting for device
+  binding conditions to be satisfied while already holding an in-flight or allocated ResourceClaim.
+  We need to make sure its reserved capacity is accounted for during preemption simulations and that
+  preempting the pod immediately aborts its `PreBind` wait so the claim can be deallocated.
```
wojtek-t, line 232: @dom4ha - can this be easily done in the code?

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: The preemptor nomination and victim nomination are two different things. What this paragraph should refer to is preemptor nomination (which is not supposed to be deleted or dallocated as a victim).

So when a preemptor pod nomination disappears, all other workloads blocked on that nomination resources, should get a chance to scheduled.

The mention of "without a corresponding `ResourceClaim` deallocation event in the API server" is bit misleading here.

keps/prod-readiness/sig-scheduling/5690.yaml
```diff
@@ -1,3 +1,3 @@
-kep-number: 6089
-alpha:
+kep-number: 5690
+beta:
```
wojtek-t, line 2: Please change this to Alpha given we decided to only target Alpha now.

wojtek-t (commented):
Given we're now targeting Alpha I'm fine with that both from PRR perspective aand wearing my scheduling-related hat.

Please fix the PRR file and I will approve the PRR part.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1078 @@
...
+implementation, `dynamicresources` records an internal *claim nomination* for that preemptor,
+consisting of:
+
+* the nominated node;
+* the simulated `AllocationResult` for each of the preemptor's ResourceClaims on that node;
+* the UIDs of the victim ResourceClaims that the simulation released in order to make the placement
+  feasible.
+
+During `Filter` in the preemption simulation, `dynamicresources` caches the computed
+`AllocationResult`s per candidate node in its own `CycleState` entry. When `AddNominatedPod` is
```
dom4ha, line 356: The CycleState exists only during scheduling cycle, but we need to remember nominated resources (AllocationResult) in some cross-cycle structure. In core we have Nominator, which is a part of scheduler cache. Here we may that a part of the dra mananger.

When `AddNominatedPod` is called (specifying a pod to add), the plugin should simulate scheduling, so look up ResourceClaims and `AllocationResult` for that pod and simulate allocating them

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1078 @@
...
+   `ResourceClaim.Status.Allocation` remains present in the state built by `PreFilter` until t3.
+2. **Add the preemptor's simulated `AllocationResult`s (t0 to t4):** `dynamicresources` adds
+   `P`'s recorded `AllocationResult`s into the allocated device state, marking the exact devices,
+   consumable capacity shares, or partition counters selected for `P` as in use.
+
+When both Pass 1 and Pass 2 run, `dynamicresources.Filter` preserves the `AllocationResult` computed
+for `Q` in Pass 1 (which respected `AddPod(P)` and avoided `P`'s nominated devices) rather than
+letting Pass 2 overwrite `nodeAllocations[nodeName]`.
+
+The combination of Pass 1 (with nominated pods) and Pass 2 (without nominated pods) ensures accurate
```
dom4ha, line 432: The purpose of Pass 2 is to ensure that pods won't get schedule under assumption some other nominated pod is already running. The main use case was to avoid scheduling a pod which has pod affinity to a nominated pod (which is not running yet).

In DRA, this mechanism is interesting, as we also cannot schedule a pod and allocate ResourceClaims under assumption that other claims are already allocated (by nominated pods). I can't imagine a case where we had similar dependency to pod affinity in DRA, but it may matter at the point we want to support shared claims.

There is one interesting detail though, that the Pass 1 scheduling determines an allocation based on other nominated allocations, while the Pass 2 could construct a different allocation (not constrained to the nominated one).

So it seems to me that we need to use Pass 1 allocation as the resulting one and probably check that exact allocation in Pass 2 to verify whether it still holds even if other nominated pods won't get eventually scheduled.

keps/sig-scheduling/5690-dra-preemption/README.md
```diff
@@ -0,0 +1,1078 @@
...
+### Claim nomination
+
+When `DefaultPreemption` selects a winning candidate for a preemptor pod, it invokes
+`PreemptionExtensions.AddNominatedPod` on registered plugins. In its `AddNominatedPod`
+implementation, `dynamicresources` records an internal *claim nomination* for that preemptor,
+consisting of:
+
+* the nominated node;
+* the simulated `AllocationResult` for each of the preemptor's ResourceClaims on that node;
+* the UIDs of the victim ResourceClaims that the simulation released in order to make the placement
```
dom4ha, line 352: The nominated resources part sounds good to me, but I have much more doubts regarding the nominated victims:
1. I wouldn't bundle them with the preemptor pod that determined them, as other preemptors should also benefit from victims determined by others during simulation. Eventually they will one resources get deallocated.
2. We need to decide whether we want to follow the existing (not ideal, but well established) path, in which we track victims only after api-server "confirmed" them or track them in-memory right after they are determined (as currently described in the KEP). The well established path for pods depend on setting `deletionTimestamp` in pods, which probably could translate to empty `ReservedFor` list. Doing so would not require to have any sophisticated tracking, but we just risk running into a problematic scenario that pod will get removed before scheduler learns about dropping to zero the `ReservedFor` list and preempting a new pod.
3. If we want to track victims in memory, I think we may want to use timeout, since we don't have guarantee the pod removal request was  successful at all, so at some point we should issue pod removal again (currently it happens in another preemption which hopefully picks the same set of victim). In general, this design part needs some more work.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: Regarding the reserved capacity (in the earlier sentence), we don't need to worry, as the pod is already assumed, so resources are already accounted in the scheduler cache and in the dra manager.

Regarding aborting `PreBind`, I wouldn't worry about it in Alpha. For Beta we probably should have a dedicated section about it. Potentialy we could handle it somehow in the process of pod removal (actuation) or in the resource claim controller as a part of deallocation process.

So I wouldn't say we can do it easily.

keps/sig-scheduling/5690-dra-preemption/README.md
wojtek-t: >  as other preemptors should also benefit from victims determined by others during simulation.

Nothing prevents us from storing the same victim for multiple preemptors.

> Doing so would not require to have any sophisticated tracking, but we just risk running into a problematic scenario that pod will get removed before scheduler learns about dropping to zero the ReservedFor list and preempting a new pod.

I think it's not a potential risk - I would expect it to happen pretty frequently actually.

keps/sig-scheduling/5690-dra-preemption/README.md
macsko: > Regarding aborting `PreBind`, I wouldn't worry about it in Alpha. For Beta we probably should have a dedicated section about it. Potentialy we could handle it somehow in the process of pod removal (actuation) or in the resource claim controller as a part of deallocation process.

We already cancel the PreBind in case the pod gets preempted:
https://github.com/kubernetes/kubernetes/blob/877e7690935292fccee581dc08357d0721512768/pkg/scheduler/framework/preemption/executor.go#L116-L118

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: > Nothing prevents us from storing the same victim for multiple preemptors.

Not sure if we know which victims of other preemptor are useful (needed) for a consecutive preemptor, as we don't reprieve victims. In general I don't see a use case in which this information would be useful in anything, so I suggest storing both nomination types independently. What we just need to achieve is that selecting a victim for one preemptor, can be consumed by the other one.

I think the confusion comes from the fact I incorrectly suggested to call them victim nominations, as the it's rather equivalent to assume mechanism instead of nomination mechanism. The difference is that we just cover the actuation part as the decision itself cannot be changed.

If we store them together with the preemptor nomination, I infer from it that these victim nominations along with preemptor nomination change, but in fact the victims will eventually disappear anyway, as the preemption is already actuated.

> I think it's not a potential risk - I would expect it to happen pretty frequently actually.

Yes, but we still could think about some workarounds, like skipping UIDs of removed pod while checking whether the ReservedFor list empty.

I don't mind going forward with the proper victim assume mechanism, but I worry that it would be better to design it separately both for pods and Resource Claims in the next cycle, rather than doing it just for claims now.

wojtek-t: >  In general I don't see a use case in which this information would be useful in anything, so I suggest storing both nomination types independently. What we just need to achieve is that selecting a victim for one preemptor, can be consumed by the other one.

This information is necessary to know if all the preemptions (including reclamation of RCs) were already done. And thus if are allowed to trigger new preemption (if the original placement no longer works for some reason).
Without knowing this information, we have no way to infer this information.

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: Ah, that's good

keps/sig-scheduling/5690-dra-preemption/README.md
dom4ha: > This information is necessary to know if all the preemptions (including reclamation of RCs) were already done. And thus if are allowed to trigger new preemption (if the original placement no longer works for some reason). Without knowing this information, we have no way to infer this information.

Eventually, as long as a workload is schedulable after all assumed victims are removed and higher/equal priority nominations considered, IIMO it doesn't matter which preemptor determined them.

This approach should be aligned with what you suggested in the Workload-Aware Preemption KEP https://github.com/kubernetes/enhancements/pull/6367#discussion_r4035083426, but before we have that, we need to fallback to checking whether there is no victim waiting for deallocation in the scope of a single node, which should be good enough for alpha. We still need to consider this as a part of this KEP design.

keps/prod-readiness/sig-scheduling/5690.yaml
mortent: Fixed.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Updated the KEP to reflect this discussion.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Updated.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: Updated the KEP to be more specific here that we need to keep the information about the preemptor and victim ResourceClaims in the DRA manager.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: I have updated the KEP to clarify this.

keps/sig-scheduling/5690-dra-preemption/README.md
mortent: I made an attempt to update the KEP to reflect this discussion, but I'm not 100% sure if I got it right. Please take a look.

PR #6115, opened 2026-05-26
KEP-5237: add route corrections metric and adjust graduation criteria

lukasmetzner:
- One-line PR description: Adds a new gated metric `route_controller_route_corrections_total` to measure how often periodic reconciles correct routes (signal for filter accuracy), and adjusts graduation criteria for milestone beta.

- Issue link: https://github.com/kubernetes/enhancements/issues/5237

### DIFF
--- a/keps/sig-cloud-provider/5237-watch-based-route-controller-reconciliation/README.md
+++ b/keps/sig-cloud-provider/5237-watch-based-route-controller-reconciliation/README.md
@@ -145,8 +145,6 @@ This is mitigated by:
 
 We use a similar concept as already implemented in the node and service controller. Through node informers we register event handlers for the add, delete and update node events, where updates are filtered by certain criteria. To introduce the feature we establish a new feature gate called `CloudControllerManagerWatchBasedRoutesReconciliation`.
 
-Additionally, we introduce a new non-gated metric `route_controller_route_sync_total`, which tracks the number of route reconciliations. This metric can be used for A/B testing to evaluate the impact of watch based route reconciliation.
-
 #### Full reconcile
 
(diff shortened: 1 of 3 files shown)


lukasmetzner:
@elmiko Should we also adjust the graduation criteria in this PR, or do you think the scope would be too large then?

elmiko (commented):
@lukasmetzner good question, up to you in terms of doing the work. since this is already looking good you might want to create another PR, but if you feel like expanding this one that would be fine also.

/lgtm

lukasmetzner:
@elmiko I have put all changes now in this PR. Should be fine, PR is still quite small.

lukasmetzner:
@elmiko Should we set this directly to `stage: beta`, or maybe first do the implementation and then open a second PR for actually graduating this to `beta`?

elmiko:
content looks good to me, seems the update toc command needs to be re-run:

``` Table of content not up to date. Did you run 'make update-toc' ? ```

> @elmiko Should we set this directly to stage: beta, or maybe first do the implementation and then open a second PR for actually graduating this to beta?

good question, i don't want to be overly pedantic, i think you could set to beta now. the only issue i can see is that if we don't get the metrics PRs merged, we should come back here and set back to alpha.

lukasmetzner:
@elmiko Yeah TOC was an issue, also we need another production readiness review for moving to beta.

Updated the PRR file and assigning @deads2k who was our previous PRR reviewer. Hope that's fine :pray: 

/assign @deads2k

keps/sig-cloud-provider/5237-watch-based-route-controller-reconciliation/README.md
deads2k, line 307: Reminder to implement this metric when going beta.

deads2k:
Reminder to add the metric indicated in PRR.  PRR lgtm

/approve

PR #6120, opened 2026-05-27
KEP-5495: update IPVS deprecation plan

danwinship:
I'm changing #5343 to be just about promoting nftables to default and re-opening #5495; it will be simpler to have a separate KEP to track the deprecation of IPVS.

As discussed in the SIG meeting last week, this updates the plan to say that we're just totally killing it off, not moving it out of tree.

/assign @adrianmoisey @aojea

### DIFF
--- a/keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
+++ b/keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
@@ -1,73 +1,11 @@
-<!--
-**Note:** When your KEP is complete, all of these comment blocks should be removed.
-
-To get started with this template:
-
-- [X] **Pick a hosting SIG.**
-  Make sure that the problem space is something the SIG is interested in taking
-  up. KEPs should not be checked in without a sponsoring SIG.
-- [X] **Create an issue in kubernetes/enhancements**
-  When filing an enhancement tracking issue, please make sure to complete all
-  fields in that template. One of the fields asks for a link to the KEP. You
-  can leave that blank until this KEP is filed, and then go back to the
-  enhancement and add the link.
-- [ ] **Make a copy of this template directory.**
-  Copy this template into the owning SIG's directory and name it
-  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
-  leading-zero padding) assigned to your enhancement above.
-- [ ] **Fill out as much of the kep.yaml file as you can.**
-  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
-  "Status", and date-related fields.
-- [ ] **Fill out this file as best you can.**
-  At minimum, you should fill in the "Summary" and "Motivation" sections.
-  These should be easy if you've preflighted the idea of the KEP with the
-  appropriate SIG(s).
-- [ ] **Create a PR for this KEP.**
-  Assign it to people in the SIG who are sponsoring this process.
-- [ ] **Merge early and iterate.**
-  Avoid getting hung up on specific details and instead aim to get the goals of
-  the KEP clarified and merged quickly. The best way to do this is to just
-  start with the high-level sections and fill out details incrementally in
-  subsequent PRs.
-
-Just because a KEP is merged does not mean it is complete or approved. Any KEP
-marked as `provisional` is a working document and subject to change. You can
-denote sections that are under active debate as follows:
-
-```
-<<[UNRESOLVED optional short context or usernames ]>>
-Stuff that is being argued.
-<<[/UNRESOLVED]>>
-```
-
-When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
-focused. If you disagree with what is already in a document, open a new PR
-with suggested changes.
-
-One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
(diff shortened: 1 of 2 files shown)


keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
```diff
@@ -175,31 +102,56 @@ to implement this enhancement.
...
+
+- Docs and warning messages are updated to indicate that the plan is
+  for the feature gate to go `Default: false` in 1.40 and
+  `LockToDefault: true` in 1.43 (after which `ipvs` mode will no
+  longer be available).
+
+#### Stage 3 (1.40)
+
+- The feature gate is flipped to `Default: false`, and the docs
+  updated.
```
adrianmoisey: What should happen here if someone sets `--proxy-mode=ipvs` in kube-proxy (while the gate is set to false)?
My thinking is possibly one of these two things:
1. Fail to start kube-proxy
2. Start kube-proxy in nftables|iptables mode

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
```diff
@@ -175,31 +102,56 @@ to implement this enhancement.
...
+
+#### Stage 4 (1.43)
+
+- The feature gate is flipped to `LockToDefault: false` and
+  `pkg/proxy/ipvs` is removed. Remaining mentions of IPVS mode in the
+  docs are removed.
```
adrianmoisey, line 151: Same question as above for 1.40 release: What happens if someone sets `--proxy-mode=ipvs` at this point?

I assume the answer for the 1.40 release will be the same as the answer for this one, but thought I'd ask anyway

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/kep.yaml
adrianmoisey, line 1: Should you add yourself to the authors list?

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
```diff
@@ -175,31 +102,56 @@ to implement this enhancement.
...
 ### Graduation Criteria
 
-#### Deprecation
+#### Stage 1 (1.35)
 
 - The Kubernetes web site has been updated with deprecation notices for the `ipvs` mode of kube-proxy
 - Kube-proxy prints a warning when a user starts kube-proxy in `ipvs` mode
 - All nftables-mode bugfixes have been backported to 1.34 and 1.33, to ensure that `ipvs` users on older releases can still migrate to `nftables`.
 
+#### Stage 2 (1.37)
```
adrianmoisey, line 125: I plan to start working on the blog post for the deprecation soon.
And with the modification of plans, I have 2 questions:
1. Should we mention it in the KEP?
2. Should we do the blog post whenever (ie: not linked to a specific release) or should we delay it for the 1.37?

My thinking is we possibly start working on it as this PR is merged and the KEP is updated, and it doesn't need to be tied to a release.

adrianmoisey (commented):
Solid plan, thanks for handling it

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
danwinship: kube-proxy should fail to start, just like if you said `--proxy-mode=userspace` or `--proxy-mode=ebpf`.

We used to try to have fallback between modes, but it was a bad idea. (https://github.com/kubernetes/kubernetes/pull/111806)

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
danwinship: Oh! I definitely mentioned the IPVS deprecation blog post when I was initially writing this stuff in #5343, and then I guess I missed that part when I was moving the text over to here. So, yes.

I think we should do the blog post whenever, because we've already been getting questions. The blog post is more about "how to migrate" and "why you don't actually care about IPVS schedulers" than it is about the exact timeline anyway.

keps/sig-network/5495-deprecate-ipvs-mode-in-kube-proxy/README.md
danwinship: from that PR:

```
In particular, if the user passed configuration options for ipvs, then
they presumably *didn't* pass configuration options for iptables, and
so even if the iptables proxy is able to run, it is likely to be
misconfigured.
```

adrianmoisey: Is it worth mentioning that in the KEP? I'm thinking if a provider reads the KEP to figure out how to handle the change, they may care about those details.

danwinship: Added. (I was going to quote the exact error message until I realize it probably will change as part of the other KEP.)

adrianmoisey:
Hmmm, I forgot that it was already approved by Dan, I didn't mean for it to merge then

aojea:
LGTM

Issue #6128, opened 2026-05-29
LoadBalancer resource for explicitly managing and monitoring load balancers


rikatz:
### Enhancement Description

- One-line enhancement description (can be used as a release note): LoadBalancer resource for explicitly managing and monitoring load balancers
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6129
- Discussion Link: [TBD](https://github.com/kubernetes/enhancements/pull/6129)
- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6129
    - [ ] Code (`k/k`) update PR(s): TBD
    - [ ] Docs (`k/website`) update PR(s): TBD

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

**Note**: This KEP will have follow-ups of https://github.com/kubernetes/enhancements/issues/4631

/sig network
/sig cloud-provider

rikatz:
this is WIP

danwinship:
/label lead-opted-in
/milestone v1.38

targeting Alpha for 1.38

lasomethingsomething:
Hello @rikatz :wave:, 1.38 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 22nd September 2026 (AoE) / Wednesday 23th September 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.38 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [ ] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [ ] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

For this KEP, we would just need to update the following:
- kep.yaml updated for 1.38
[...]

PR #6129, opened 2026-05-29
KEP-6128: (alpha) LoadBalancer resource for explicitly managing and monitoring load balancers

rikatz:
- One-line PR description: LoadBalancer resource for explicitly managing and monitoring load balancers

- Issue link: https://github.com/kubernetes/enhancements/issues/6128

- Other comments: This is a WIP from a discussion me and Dan had during Kubecon, lots of open questions yet.

This proposal was written with the assistance of Claude/Opus 4.6

### DIFF
--- a/keps/sig-network/6128-loadbalancer-resource/README.md
+++ b/keps/sig-network/6128-loadbalancer-resource/README.md
@@ -112,7 +112,7 @@ The new `LoadBalancer` resource provides:
 
 The resource is optional and additive. Existing Services of type
 `LoadBalancer` continue to work unchanged. When a `LoadBalancer`
-resource exists with the same name as a Service in the same namespace,
+resource exists with the same name as a `Service` in the same namespace,
 the cloud controller manager uses it as the source of truth, falling
 back to `Service.Spec` for unset fields.
 
@@ -137,7 +137,7 @@ several problems:
    balancer is being provisioned, has failed, is serving traffic, or
    is operating in a degraded mode. This was the primary motivation
    behind [KEP-4631], which attempted to add conditions directly to
-   the Service resource but was never implemented.
+   the `Service` resource but was never implemented.
 
 3. **Service is overloaded.** The `Service` resource serves multiple
    purposes: service discovery, internal cluster routing, and
@@ -149,7 +149,7 @@ several problems:
    infrastructure.** Gateway API implementations that run their
    proxy inside the cluster need load balancers to route and forward
    traffic from external clients to the proxy pods. Today, this is
-   done through `Service` of type `LoadBalancer` with cloud-specific
+   done through `Service` of type [`LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) with cloud-specific
    annotations. A standalone `LoadBalancer` resource provides a
    structured way to configure and observe this routing/forwarding
    layer. (Gateway API implementations that run externally and manage
@@ -174,9 +174,9 @@ several problems:
 - Provide structured status reporting with conditions (`Accepted`,
   `Programmed`, `Degraded`) aligned with Gateway API conventions.
 
-- Mirror load balancer addresses (IPs/hostnames) in
-  `LoadBalancer.Status` so users can observe load balancer state from
-  a single resource.
+- Mirror load balancer addresses (IPs/hostnames) from `Services` in
+  `LoadBalancer.Status` so users can observe load balancer state 
+  (conditions, IPs) from `LoadBalancer` resource only.
 
 - Provide a feature reporting mechanism so cloud providers can
   advertise supported capabilities.
@@ -473,7 +473,7 @@ type LoadBalancerRoutability string
 
 const (
     // LoadBalancerRoutabilityExternal indicates the load balancer
-    // is reachable from the public internet.
+    // is reachable from outside of the cluster network.
     LoadBalancerRoutabilityExternal LoadBalancerRoutability = "External"
 
     // LoadBalancerRoutabilityInternal indicates the load balancer
@@ -630,7 +630,7 @@ support.
 
 <<[UNRESOLVED service-status-interaction]>>
 
-When a `LoadBalancer` resource exists alongside a Service, the
+When a `LoadBalancer` resource exists alongside a `Service`, the
 relationship between `LoadBalancer.Status.Addresses` and
 `Service.Status.LoadBalancer.Ingress` must be clearly defined:
 
@@ -903,7 +903,7 @@ None.
     `LoadBalancer` resource.
   - Populating `.Status.Addresses` when the load balancer is
     provisioned.
-  - Falling back to Service.Spec when `LoadBalancer` fields are unset.
+  - Falling back to `Service.Spec` when `LoadBalancer` fields are unset.
   - Using `LoadBalancer.Spec.Routability` to provision internal vs
     external load balancers.
   - Reporting `SupportedFeatures` in status.
--- a/keps/sig-network/6128-loadbalancer-resource/README.md
+++ b/keps/sig-network/6128-loadbalancer-resource/README.md
@@ -301,14 +301,6 @@ of cloud-specific annotations per environment.
 
 ### Risks and Mitigations
 
-**Risk: Cloud provider adoption.** If cloud providers do not adopt
-the new resource, users gain no benefit.
-
-*Mitigation:* The resource is designed to be optional and additive.
-Cloud providers can adopt it incrementally. The structured status
-reporting alone provides enough value to motivate adoption, as it
-solves the e2e testing problem identified in KEP-4631.
-
 **Risk: Configuration drift between Service and LoadBalancer.** Users
 may set conflicting values on the Service and LoadBalancer resource.
 
@@ -316,15 +308,6 @@ may set conflicting values on the Service and LoadBalancer resource.
 fields take precedence when set, with fallback to `Service.Spec`
 fields when unset. Cloud controller managers should set a condition or
 event when they detect conflicting configuration.
-
-**Risk: RBAC misconfiguration allows users to overwrite status.**
-The `LoadBalancer.Status` should only be written by the cloud
-controller manager, not by end users.
-
-*Mitigation:* RBAC policies must ensure that only the cloud
-controller manager's service account has `update` and `patch`
-permissions on the `loadbalancers/status` subresource. Default
-ClusterRoles should be configured accordingly.
 
 **Risk: Lateral configuration override via spec create permissions.**
 A user or controller with `create` permissions on `loadbalancers` in
--- a/keps/sig-network/6128-loadbalancer-resource/README.md
+++ b/keps/sig-network/6128-loadbalancer-resource/README.md
@@ -1,0 +1,1275 @@
+# KEP-6128: LoadBalancer Resource
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Explicit internal load balancer](#story-1-explicit-internal-load-balancer)
+    - [Story 2: Observing load balancer provisioning state](#story-2-observing-load-balancer-provisioning-state)
+    - [Story 3: In-cluster Gateway API integration](#story-3-in-cluster-gateway-api-integration)
+    - [Story 4: Multi-cloud portability](#story-4-multi-cloud-portability)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Definition](#api-definition)
+  - [Status](#status)
+  - [Conditions](#conditions)
+    - [Accepted](#accepted)
+    - [Programmed](#programmed)
+    - [Degraded](#degraded)
+  - [Binding Model](#binding-model)
+  - [Validation](#validation)
+  - [Interaction with Gateway API](#interaction-with-gateway-api)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Add fields directly to Service (KEP-4631 approach)](#add-fields-directly-to-service-kep-4631-approach)
+  - [Use Gateway API for all load balancer configuration](#use-gateway-api-for-all-load-balancer-configuration)
+  - [Use CRDs instead of a built-in type](#use-crds-instead-of-a-built-in-type)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes a new `LoadBalancer` resource in the
+`networking.k8s.io/v1alpha1` API group that decouples load balancer
+configuration and status from the `Service` resource.
+
+Today, load balancer behavior in Kubernetes is configured through a
+combination of `Service.Spec` fields (e.g., `sessionAffinity`,
+`loadBalancerSourceRanges`) and cloud-provider-specific annotations
+(e.g., internal vs external routing, IP allocation, health check
+intervals). Load balancer status is limited to the IP/hostname list in
+`Service.Status.LoadBalancer`, with no structured way for cloud
+providers to communicate provisioning state, feature support, or
+degraded behavior.
+
+The new `LoadBalancer` resource provides:
+
+- **Standardized configuration** for common load balancer features
+  currently handled through provider-specific annotations
+  (routability, IP allocation, health checks).
+
+- **Structured status** with conditions aligned to Gateway API
+  semantics (`Accepted`, `Programmed`, `Degraded`).
+
+- **Feature reporting** so cloud providers can advertise supported
+  capabilities, enabling e2e tests and users to determine provider
+  support without hard-coded skip rules.
+
+- **Gateway API integration point**. In-cluster Gateway
+  implementations need load balancers to route and forward traffic
+  from external clients to the Gateway proxy. The `LoadBalancer`
+  resource provides a structured way to configure and observe this
+  routing/forwarding layer without Gateway API having to redefine
+  load balancer semantics.
+
+The resource is optional and additive. Existing Services of type
+`LoadBalancer` continue to work unchanged. When a `LoadBalancer`
+resource exists with the same name as a Service in the same namespace,
+the cloud controller manager uses it as the source of truth, falling
+back to `Service.Spec` for unset fields.
+
+## Motivation
+
+The current model for configuring load balancers in Kubernetes has
+several problems:
+
+1. **Configuration is scattered across annotations.** Core load
+   balancer behaviors like internal vs external routing, IP allocation
+   strategy, and health check configuration are handled through
+   cloud-provider-specific annotations. Every cloud provider has
+   invented its own annotation names for the same concepts (e.g.,
+   `service.beta.kubernetes.io/aws-load-balancer-internal`,
+   `networking.gke.io/internal-load-balancer`,
+   `service.beta.kubernetes.io/azure-load-balancer-internal`). This
+   makes multi-cloud deployments and tooling unnecessarily difficult.
+
+2. **Status reporting is inadequate.** `Service.Status.LoadBalancer`
+   only contains a list of ingress IPs/hostnames. There is no
+   structured way for a cloud provider to communicate whether a load
+   balancer is being provisioned, has failed, is serving traffic, or
+   is operating in a degraded mode. This was the primary motivation
+   behind [KEP-4631], which attempted to add conditions directly to
+   the Service resource but was never implemented.
+
+3. **Service is overloaded.** The `Service` resource serves multiple
+   purposes: service discovery, internal cluster routing, and
+   external load balancer management. Mixing load balancer
+   infrastructure concerns with service abstraction concerns makes
+   the API harder to reason about and evolve independently.
+
+4. **In-cluster Gateway API implementations need load balancer
+   infrastructure.** Gateway API implementations that run their
+   proxy inside the cluster need load balancers to route and forward
+   traffic from external clients to the proxy pods. Today, this is
+   done through `Service` of type `LoadBalancer` with cloud-specific
+   annotations. A standalone `LoadBalancer` resource provides a
+   structured way to configure and observe this routing/forwarding
+   layer. (Gateway API implementations that run externally and manage
+   their own load balancer infrastructure are not affected.)
+
+5. **E2e tests cannot determine provider capabilities.** As
+   identified in [KEP-4631] and [kubernetes/kubernetes#123714], load
+   balancer e2e tests have historically hard-coded provider-specific
+   skip rules. With in-tree cloud providers removed, there is no
+   scalable way for tests to determine which features a cloud provider
+   supports.
+
+[KEP-4631]: https://github.com/kubernetes/enhancements/pull/4632
+[kubernetes/kubernetes#123714]: https://github.com/kubernetes/kubernetes/issues/123714
+
+### Goals
+
+- Define a new namespaced `LoadBalancer` resource in the
+  `networking.k8s.io/v1alpha1` API group with explicit fields for
+  common load balancer configuration.
+
+- Provide structured status reporting with conditions (`Accepted`,
+  `Programmed`, `Degraded`) aligned with Gateway API conventions.
+
+- Mirror load balancer addresses (IPs/hostnames) in
+  `LoadBalancer.Status` so users can observe load balancer state from
+  a single resource.
+
+- Provide a feature reporting mechanism so cloud providers can
+  advertise supported capabilities.
+
+- Maintain full backward compatibility: existing `Service` of type
+  `LoadBalancer` without a corresponding `LoadBalancer` resource must
+  continue to work exactly as today.
+
+- Provide a clear integration point for in-cluster Gateway API
+  implementations to configure and observe the load balancer
+  routing/forwarding layer without redefining these semantics.
+
+### Non-Goals
+
+- Replacing or deprecating `Service` of type `LoadBalancer`. The
+  `Service` resource remains the primary way to expose workloads.
+
+- Requiring all cloud providers to support the new resource. Cloud
+  controller managers opt in to `LoadBalancer` resource support.
+
+- Moving service-routing concerns (`externalTrafficPolicy`,
+  `internalTrafficPolicy`, `ipFamilies`, `ipFamilyPolicy`,
+  `allocateLoadBalancerNodePorts`) to the `LoadBalancer` resource.
+  These are service behavior, not load balancer infrastructure
+  concerns.
+
+- Defining how cloud providers should migrate from annotation-based
+  configuration to the new resource. Migration strategy is left to
+  individual cloud providers.
+
+- Defining a mechanism for cloud-provider-specific extension fields
+  beyond the common set. Cloud providers may continue to use
+  annotations for provider-specific features.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Explicit internal load balancer
+
+A platform team wants to create an internal load balancer without
+relying on provider-specific annotations. They create a `LoadBalancer`
+resource with `routability: Internal` alongside their Service:
+
+```yaml
+apiVersion: networking.k8s.io/v1alpha1
+kind: LoadBalancer
+metadata:
+  name: my-service
+  namespace: default
+spec:
+  routability: Internal
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: my-service
+  namespace: default
+spec:
+  type: LoadBalancer
+  ports:
+    - port: 80
+      targetPort: 8080
+  selector:
+    app: my-app
+```
+
+#### Story 2: Observing load balancer provisioning state
+
+A user creates a `Service` of type `LoadBalancer` and a corresponding
+`LoadBalancer` resource. Instead of repeatedly checking whether
+`Service.Status.LoadBalancer.Ingress` is populated, they can watch
+the `LoadBalancer` resource conditions:
+
+```
+$ kubectl get loadbalancer my-service -o jsonpath='{.status.conditions}'
+[
+  {"type": "Accepted", "status": "True", ...},
+  {"type": "Programmed", "status": "True", ...}
+]
+```
+
+#### Story 3: In-cluster Gateway API integration
+
+An in-cluster Gateway API implementation creates a `LoadBalancer`
+resource to configure the routing/forwarding infrastructure that
+delivers external traffic to the Gateway proxy pods, instead of
+relying on cloud-specific annotations on a Service. The Gateway
+controller watches `LoadBalancer` status to determine when the
+infrastructure is ready to receive traffic.
+
+#### Story 4: Multi-cloud portability
+
+An organization deploying the same application across AWS, GCP, and
+Azure uses a `LoadBalancer` resource with `routability: Internal` and
+`loadBalancerSourceRanges` instead of maintaining three different sets
+of cloud-specific annotations per environment.
+
+### Notes/Constraints/Caveats
+
+- The `LoadBalancer` resource is bound to a `Service` by name within
+  the same namespace. This follows the same pattern as `Service` and
+  `Endpoints`.
+
+- Cloud controller managers are not required to support the
+  `LoadBalancer` resource. If a cloud controller manager does not
+  recognize the resource, the `Service` continues to work as it does
+  today.
+
+- When a `LoadBalancer` resource exists, it serves as the source of
+  truth for the fields it defines. Fallback semantics depend on the
+  field type:
+  - **Pointer fields** (`Routability`, `SessionAffinity`,
+    `HealthCheck`, `ServiceRef`): `nil` means "no opinion, fall back
+    to Service." A non-nil value is the source of truth.
+  - **Slice fields** (`SourceRanges`): `nil` means "no opinion, fall
+    back to Service." An explicit empty slice (`[]`) means "the
+    LoadBalancer explicitly sets no value" and does **not** fall back
+    to the Service. For `SourceRanges`, `nil` falls back to
+    `Service.Spec.LoadBalancerSourceRanges`, while `[]` means "no
+    source range restrictions."
+
+### Risks and Mitigations
+
+**Risk: Cloud provider adoption.** If cloud providers do not adopt
+the new resource, users gain no benefit.
+
+*Mitigation:* The resource is designed to be optional and additive.
+Cloud providers can adopt it incrementally. The structured status
+reporting alone provides enough value to motivate adoption, as it
+solves the e2e testing problem identified in KEP-4631.
+
+**Risk: Configuration drift between Service and LoadBalancer.** Users
+may set conflicting values on the Service and LoadBalancer resource.
+
+*Mitigation:* The precedence rule is clear: `LoadBalancer` resource
+fields take precedence when set, with fallback to `Service.Spec`
+fields when unset. Cloud controller managers should set a condition or
+event when they detect conflicting configuration.
+
+**Risk: RBAC misconfiguration allows users to overwrite status.**
+The `LoadBalancer.Status` should only be written by the cloud
+controller manager, not by end users.
+
+*Mitigation:* RBAC policies must ensure that only the cloud
+controller manager's service account has `update` and `patch`
+permissions on the `loadbalancers/status` subresource. Default
+ClusterRoles should be configured accordingly.
+
+**Risk: Lateral configuration override via spec create permissions.**
+A user or controller with `create` permissions on `loadbalancers` in
+a namespace can create a `LoadBalancer` resource that silently alters
+the behavior of any `Service` of type `LoadBalancer` in that namespace
+(e.g., changing routability or source ranges) without the Service
+owner's knowledge.
+
+<<[UNRESOLVED rbac-binding-security]>>
+
+The name-match binding model means that any principal with `create`
+permissions on `loadbalancers` can affect any Service in the same
+namespace. Options to mitigate:
+
+1. **Restrict `loadbalancers` creation** to the same set of
+   principals that can create Services. This is the simplest approach
+   but may be too broad.
+
+2. **Require an explicit opt-in annotation on the Service** (e.g.,
+   `networking.k8s.io/loadbalancer-resource: "true"`) before the
+   cloud controller manager considers a matching `LoadBalancer`
+   resource. This prevents unsolicited overrides.
+
+3. **Use an explicit `serviceRef` field** on the `LoadBalancer`
+   resource (see Binding Model section) which could enable
+   admission-time validation of cross-resource permissions.
+
+4. **Rely on namespace-level trust boundaries.** Kubernetes
+   namespaces are already trust boundaries — principals with write
+   access to a namespace can already modify Services, Endpoints, and
+   other resources that affect networking behavior.
+
+<<[/UNRESOLVED]>>
+
+## Design Details
+
+### API Definition
+
+The `LoadBalancer` resource is defined in the
+`networking.k8s.io/v1alpha1` API group:
+
+```golang
+// LoadBalancer defines the desired configuration and reports the
+// status of a load balancer associated with a Service of type
+// LoadBalancer. The LoadBalancer must have the same name and
+// namespace as the Service it is associated with.
+type LoadBalancer struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec   LoadBalancerSpec   `json:"spec,omitempty"`
+    Status LoadBalancerStatus `json:"status,omitempty"`
+}
+
+type LoadBalancerSpec struct {
+    // Routability indicates whether the load balancer should be
+    // reachable from the public internet (External) or only from
+    // within the network (Internal). If unset, the cloud controller
+    // manager falls back to the Service annotations or its own
+    // default behavior.
+    // +optional
+    Routability *LoadBalancerRoutability `json:"routability,omitempty"`
+
+    // SessionAffinity defines the session affinity type and
+    // configuration for the load balancer. If nil, the cloud
+    // controller manager falls back to Service.Spec.SessionAffinity
+    // and Service.Spec.SessionAffinityConfig.
+    // +optional
+    SessionAffinity *LoadBalancerSessionAffinity `json:"sessionAffinity,omitempty"`
+
+    // SourceRanges restricts traffic to the load balancer to the
+    // specified client IP ranges. If nil, the cloud controller
+    // manager falls back to Service.Spec.LoadBalancerSourceRanges.
+    // An explicit empty list means "no source range restrictions"
+    // and does not fall back to the Service.
+    // +optional
+    SourceRanges []CIDR `json:"sourceRanges,omitempty"`
+
+    // HealthCheck configures how the cloud load balancer performs
+    // health checks against backends. These are cloud infrastructure
+    // health checks, independent of Pod readiness probes.
+    // If nil, the cloud controller manager uses its default health
+    // check configuration.
+    // +optional
+    HealthCheck *HealthCheckConfig `json:"healthCheck,omitempty"`
+
+    // ServiceRef is an optional explicit reference to the Service
+    // this LoadBalancer is associated with. If unset, the
+    // LoadBalancer binds to a Service with the same name in the
+    // same namespace. When set, the name-match binding is not used
+    // and the referenced Service is used instead.
+    // +optional
+    ServiceRef *ServiceReference `json:"serviceRef,omitempty"`
+}
+
+// CIDR is a string type that represents a network in CIDR notation
+// (e.g., "203.0.113.0/24", "2001:db8::/32"). Values must be valid
+// CIDR notation and are validated at admission time.
+// +kubebuilder:validation:Format=cidr
+type CIDR string
+
+// ServiceReference identifies a Service in the same namespace.
+type ServiceReference struct {
+    // Name is the name of the Service.
+    Name string `json:"name"`
+}
+
+// LoadBalancerSessionAffinity defines session affinity mode and
+// configuration for a load balancer.
+type LoadBalancerSessionAffinity struct {
+    // Type selects the session affinity mode.
+    // Valid values are "None" and "ClientIP".
+    // +enum
+    Type corev1.ServiceAffinity `json:"type"`
+
+    // Config holds additional session affinity configuration.
+    // Applicable only when Type is "ClientIP".
+    // +optional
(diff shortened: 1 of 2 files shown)


keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+
+```
+$ kubectl get loadbalancer my-service -o jsonpath='{.status.conditions}'
+[
+  {"type": "Accepted", "status": "True", ...},
+  {"type": "Programmed", "status": "True", ...}
+]
+```
+
+#### Story 3: In-cluster Gateway API integration
```
rikatz, line 262: a side note: I have explicitly added this story because I needed it on Gateway API, but reading it out loud doesn't solve many Gateway API issues tbh, as Gateway API won't have the same semantics on it and "how a LoadBalancer" is provisioned is up to the creator of the service (which means the problem still exists on the implementation side of Gateway API).

One thing I have considered and wondered was about having a LoadBalancerClass and this class behaving like GatewayClass, where you could actually define "what a LoadBalancer of class 'internal' means for the enterprise", which would make something like:

- Gateway - uses loadBalancerClass internal
- Implementation - Creates a loadbalancer and service resource, point the loadbalancer to that class
- waits for the loadbalancer status to be reported
- now you have a class-level configuration for loadbalancer, not a per-lb specific

This is open to discussion, I just wanted to be sure that the original story was captured

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+#### Story 4: Multi-cloud portability
+
+An organization deploying the same application across AWS, GCP, and
+Azure uses a `LoadBalancer` resource with `routability: Internal` and
+`loadBalancerSourceRanges` instead of maintaining three different sets
+of cloud-specific annotations per environment.
+
+### Notes/Constraints/Caveats
+
+- The `LoadBalancer` resource is bound to a `Service` by name within
```
rikatz, line 280: and this is fine, right? right? Otherwise we must discuss what is a better approach here (serviceRef? some relationship like services and endpointSlices?)

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+
+- Cloud controller managers are not required to support the
+  `LoadBalancer` resource. If a cloud controller manager does not
+  recognize the resource, the `Service` continues to work as it does
+  today.
+
+- When a `LoadBalancer` resource exists, it serves as the source of
+  truth for the fields it defines. Fallback semantics depend on the
+  field type:
+  - **Pointer fields** (`Routability`, `SessionAffinity`,
```
rikatz, line 292: this is more like an attempt to keep compatibility, tho I know that maybe the usage of a pointer is not always welcome on spec fields.

The semantic here is still undefined tbh, but purely from "a user perspective" having a null field means "I have no opinion, fallback to the default" which is different of an empty field that imo means "I have an opinion, and I don't want anything to be accepted".

rikatz: probably this does not make much sense for non-slices fields as it makes for slices fields (see below)

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+    `HealthCheck`, `ServiceRef`): `nil` means "no opinion, fall back
+    to Service." A non-nil value is the source of truth.
+  - **Slice fields** (`SourceRanges`): `nil` means "no opinion, fall
+    back to Service." An explicit empty slice (`[]`) means "the
+    LoadBalancer explicitly sets no value" and does **not** fall back
+    to the Service. For `SourceRanges`, `nil` falls back to
+    `Service.Spec.LoadBalancerSourceRanges`, while `[]` means "no
```
rikatz, line 299: hum, I guess I wanted to say "no source range allowed" 🤔  

I think no-source range restrictions is 0.0.0.0/0 (I prefer to fail-close than fail-open here)

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+    // same namespace. When set, the name-match binding is not used
+    // and the referenced Service is used instead.
+    // +optional
+    ServiceRef *ServiceReference `json:"serviceRef,omitempty"`
+}
+
+// CIDR is a string type that represents a network in CIDR notation
+// (e.g., "203.0.113.0/24", "2001:db8::/32"). Values must be valid
+// CIDR notation and are validated at admission time.
+// +kubebuilder:validation:Format=cidr
```
rikatz, line 409: IIRC there is an internal validation tag for IPs? Need to check

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+
+The `Accepted` condition indicates that the cloud controller manager
+has seen the `LoadBalancer` resource and acknowledges responsibility
+for it.
+
+- `Status=True, Reason=Accepted`: The controller has accepted the
+  resource and will attempt to provision a load balancer.
+- `Status=False, Reason=Unsupported`: The controller cannot provision
+  a load balancer for this configuration (explained in `Message`).
+- `Status=False, Reason=Conflict`: The `LoadBalancer` resource
```
rikatz, line 674: open to discussion and should be defined based on precedence rules definition

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+- `Status=False, Reason=ServiceNotFound`: The `serviceRef` references
+  a Service that does not exist.
+
+If no `Accepted` condition appears within a reasonable time after
+creation, the user should assume no cloud controller manager is
+handling this resource.
+
+#### Programmed
+
+The `Programmed` condition indicates that the load balancer
```
rikatz, line 689: this is very similar to the "Serving" condition from the original GEP of LoadBalancer status IMO, but maybe we want additional conditions here?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require
+updates to existing tests to make this code solid enough prior to
+committing the changes necessary to implement this enhancement.
+
+##### Prerequisite testing updates
+
+None.
```
rikatz, line 877: well....we will need some implementation that supports the fields like cloud-provider-kind to be changed here so we can test conformance

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+
+The time to reach `Programmed=True` depends on the cloud provider's
+infrastructure provisioning time and is not bounded by this KEP.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- [ ] Metrics
+  - Metric name: `loadbalancer_resource_reconcile_duration_seconds`
+  - Aggregation method: histogram
+  - Components exposing the metric: cloud-controller-manager
```
rikatz, line 1081: nope, we probably need this on kind provider, this is wrong

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1275 @@
...
+A metric tracking the count of `LoadBalancer` resources by condition
+status (e.g., how many are in `Programmed=True` vs
+`Programmed=False`) would be useful for cluster-level observability
+but is not required for alpha.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+- Cloud controller manager
```
rikatz, line 1097: fix to "any cloud controller" that can deal with LoadBalancer actually

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+- **Gateway API integration point**. In-cluster Gateway
+  implementations need load balancers to route and forward traffic
+  from external clients to the Gateway proxy. The `LoadBalancer`
+  resource provides a structured way to configure and observe this
+  routing/forwarding layer without Gateway API having to redefine
+  load balancer semantics.
+
+The resource is optional and additive. Existing Services of type
+`LoadBalancer` continue to work unchanged. When a `LoadBalancer`
+resource exists with the same name as a Service in the same namespace,
```
mtulio:
```suggestion
resource exists with the same name as a `Service` in the same namespace,
```

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   `service.beta.kubernetes.io/azure-load-balancer-internal`). This
+   makes multi-cloud deployments and tooling unnecessarily difficult.
+
+2. **Status reporting is inadequate.** `Service.Status.LoadBalancer`
+   only contains a list of ingress IPs/hostnames. There is no
+   structured way for a cloud provider to communicate whether a load
+   balancer is being provisioned, has failed, is serving traffic, or
+   is operating in a degraded mode. This was the primary motivation
+   behind [KEP-4631], which attempted to add conditions directly to
+   the Service resource but was never implemented.
```
mtulio:
```suggestion
   the `Service` resource but was never implemented.
```

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   purposes: service discovery, internal cluster routing, and
+   external load balancer management. Mixing load balancer
+   infrastructure concerns with service abstraction concerns makes
+   the API harder to reason about and evolve independently.
+
+4. **In-cluster Gateway API implementations need load balancer
+   infrastructure.** Gateway API implementations that run their
+   proxy inside the cluster need load balancers to route and forward
+   traffic from external clients to the proxy pods. Today, this is
+   done through `Service` of type `LoadBalancer` with cloud-specific
```
mtulio:
```suggestion
   done through `Service` of type [`LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer) with cloud-specific
```

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   structured way to configure and observe this routing/forwarding
+   layer. (Gateway API implementations that run externally and manage
+   their own load balancer infrastructure are not affected.)
+
+5. **E2e tests cannot determine provider capabilities.** As
+   identified in [KEP-4631] and [kubernetes/kubernetes#123714], load
+   balancer e2e tests have historically hard-coded provider-specific
+   skip rules. With in-tree cloud providers removed, there is no
+   scalable way for tests to determine which features a cloud provider
+   supports.
```
mtulio, line 163: cc/fwiw @elmiko @miyadav may be interesred by test topic to enhance e2e across cloud provider controllers. :)

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+- Define a new namespaced `LoadBalancer` resource in the
+  `networking.k8s.io/v1alpha1` API group with explicit fields for
+  common load balancer configuration.
+
+- Provide structured status reporting with conditions (`Accepted`,
+  `Programmed`, `Degraded`) aligned with Gateway API conventions.
+
+- Mirror load balancer addresses (IPs/hostnames) in
+  `LoadBalancer.Status` so users can observe load balancer state from
+  a single resource.
```
mtulio: I am not sure if I followed that goal! Mirror from `Service` to "observe load balancer state from
  a single resource"?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+
+- Provide a feature reporting mechanism so cloud providers can
+  advertise supported capabilities.
+
+- Maintain full backward compatibility: existing `Service` of type
+  `LoadBalancer` without a corresponding `LoadBalancer` resource must
```
mtulio, line 185: supposing it will be managed by same controller (CCM) or a new controller/service is planned to be introduced aside CCM?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+relying on cloud-specific annotations on a Service. The Gateway
+controller watches `LoadBalancer` status to determine when the
+infrastructure is ready to receive traffic.
+
+#### Story 4: Multi-cloud portability
+
+An organization deploying the same application across AWS, GCP, and
+Azure uses a `LoadBalancer` resource with `routability: Internal` and
+`loadBalancerSourceRanges` instead of maintaining three different sets
+of cloud-specific annotations per environment.
```
mtulio, line 276: if cloud provider standardize those "common" annotations letting provider implementations using those, creating some mapping to keep compatibility, will it resolve the annotation discrepancy pain point?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+
+A hybrid approach is also possible: per-class features for general
+capabilities, with per-instance overrides for configuration-dependent
+support.
+
+<<[/UNRESOLVED]>>
+
+<<[UNRESOLVED service-status-interaction]>>
+
+When a `LoadBalancer` resource exists alongside a Service, the
```
mtulio:
```suggestion
When a `LoadBalancer` resource exists alongside a `Service`, the
```

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+)
+```
+
+<<[UNRESOLVED feature-reporting]>>
+
+The `SupportedFeatures` field on `LoadBalancer.Status` reports
+per-instance feature support. An alternative is a cluster-scoped
+mechanism (similar to `StorageClass` or `IngressClass`, or Gateway
+API's `SupportedFeatures` on `GatewayClass`) where a cloud provider
+advertises its general capabilities once, rather than per-instance.
```
mtulio, line 611: interesting suggestion. 👍🏽 
I think capabilities in per-instance would be a good idea. There may be examples that the cluster(CCM) has capability for certain features but the instance (LB) may not (and never will). e.g. old LBs created when CCM did not supported certain capability by actual CCM. for provider AWS there is one example SG for Service type Loadbalancer NLB, which needs 1) CCM with certain version and configuration, 2) LB instance with Service created with that capability enabled. So even if the Service NLB has been created prior the CCM with "NLBManagedSG" capability, the instance(Service) does not support it, and can't support without recreation.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+
+##### e2e tests
+
+- Cloud provider implementations (cloud-provider-kind at minimum for
+  alpha) should demonstrate:
+  - Setting `Accepted` and `Programmed` conditions on the
+    `LoadBalancer` resource.
+  - Populating `.Status.Addresses` when the load balancer is
+    provisioned.
+  - Falling back to Service.Spec when `LoadBalancer` fields are unset.
```
mtulio:
```suggestion
  - Falling back to `Service.Spec` when `LoadBalancer` fields are unset.
```

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+`LoadBalancer` resource may be created before the corresponding
+Service, and admission validation cannot enforce cross-resource
+constraints.
+
+### Interaction with Gateway API
+
+In-cluster Gateway API implementations that need load balancers to
+route and forward external traffic to their proxy pods can create a
+`LoadBalancer` resource instead of configuring a Service with
+cloud-specific annotations. This allows the Gateway controller to:
```
mtulio, line 854: I may be missed, but how the `Service`'s annotations will looks like when they have bound to a `LoadBalancer` resource? Will it keep with same cloud provider's specific annotations, set by `LoadBalancer` spec, or will it be empty delegated to `LoadBalancer`resource?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   owner reference to the Service. Deleting the Service deletes the
+   LoadBalancer. Deleting the LoadBalancer reverts to Service-only
+   behavior.
+
+2. **LoadBalancer owns Service**: The `LoadBalancer` resource is the
+   primary object. This is unusual in Kubernetes and likely not the
+   right model.
+
+3. **No ownership**: Both resources are independent. The cloud
+   controller manager observes both and reconciles. Deleting either
```
mtulio, line 772: afaict (@elmiko please LMK if I am wrong =] ), at least for cloud-provider AWS CCM, it uses even drivel callback to manage `Service` resource by design. can you see challenges in this implementations to support `LoadBalancer` to be "observed" by CCM to reconcile both resources?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  (routability, IP allocation, health checks).
+
+- **Structured status** with conditions aligned to Gateway API
+  semantics (`Accepted`, `Programmed`, `Degraded`).
+
+- **Feature reporting** so cloud providers can advertise supported
+  capabilities, enabling e2e tests and users to determine provider
+  support without hard-coded skip rules.
+
+- **Gateway API integration point**. In-cluster Gateway
```
aojea, line 106: This is a kind of chicken an egg problem since Gateway API represent a LoadBalancer object decomposed in indepedent objects to avoid the exact problem of having to use annotations for each combination ... IIUIC the problem here is only for "In-Cluster" gateways, isn't it?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  These are service behavior, not load balancer infrastructure
+  concerns.
+
+- Defining how cloud providers should migrate from annotation-based
+  configuration to the new resource. Migration strategy is left to
+  individual cloud providers.
+
+- Defining a mechanism for cloud-provider-specific extension fields
+  beyond the common set. Cloud providers may continue to use
+  annotations for provider-specific features.
```
mfbonfigli, line 212: I think the proposal is interesting, but I am a bit concerned about this. 

If custom annotation proliferation will continue (as I expect it will given the many provider-specific configurations available, e.g. check out the just how many the [AWS LBC](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.3/guide/service/annotations/) has just for NLBs), I expect cloud providers to start attaching custom annotations to the new resource, effectively shifting/duplicating this problem onto the LoadBalancer resource too.

If we are standardizing the core fields/configurations without an extension mechanism do we risk of just shifting the current annotation sprawl onto a new resource?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+
+##### Integration tests
+
+- Creating a `LoadBalancer` resource with and without a corresponding
+  Service verifies the binding behavior.
+- Creating a `LoadBalancer` with conflicting fields from the Service
+  verifies the precedence behavior.
+- Feature gate enable/disable tests for the new API type.
+
+##### e2e tests
```
aojea, line 898: e2e test need to provide end to end, as datapath traffic ... and here is when you enter in contradictions with the functionalities  because a lot of them depend on the internal infrastructure of the cloud providers and each cloud provider most probably has different meanings for similar functionality or completely incompatible functionalities ... Let's start with Routability, what will be the definition for Internal and External that works for all environments? Is a proxy or an L4 loadbalancer?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+
+The resource is optional and additive. Existing Services of type
+`LoadBalancer` continue to work unchanged. When a `LoadBalancer`
+resource exists with the same name as a Service in the same namespace,
+the cloud controller manager uses it as the source of truth, falling
+back to `Service.Spec` for unset fields.
+
+## Motivation
+
+The current model for configuring load balancers in Kubernetes has
```
aojea, line 121: I think that this is referring to a particular subset of loadbalancers only, people configure loadbalancers with Gateway API too

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   infrastructure.** Gateway API implementations that run their
+   proxy inside the cluster need load balancers to route and forward
+   traffic from external clients to the proxy pods. Today, this is
+   done through `Service` of type `LoadBalancer` with cloud-specific
+   annotations. A standalone `LoadBalancer` resource provides a
+   structured way to configure and observe this routing/forwarding
+   layer. (Gateway API implementations that run externally and manage
+   their own load balancer infrastructure are not affected.)
+
+5. **E2e tests cannot determine provider capabilities.** As
```
aojea, line 158: Theoretically these test should not be in tree too https://github.com/kubernetes/enhancements/tree/master/keps/sig-cloud-provider/2395-removing-in-tree-cloud-providers

keps/sig-network/6128-loadbalancer-resource/README.md
aojea: I'm skeptical cloud providers will also move to a new object, the amount of work needed and the risk of breaking workloads during migrations make it unfortunately very difficult.

Just one example https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/1435-mixed-protocol-lb/README.md , until that KEP loadbalancers were restricting the use of TCP and UDP at the same time and a KEP was done for relaxing validation, only one cloud provider implemented it in only one subset of all the loadbalancers flavors in 1.26 ... Google implemented it in 1.34 only for the external flavor https://docs.cloud.google.com/kubernetes-engine/docs/how-to/mixed-protocol-lb ... as a person that worked on the internals of the load balancers cloud providers implementations for Kubernetes ... trust, is more complicated than it seems

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+// status of a load balancer associated with a Service of type
+// LoadBalancer. The LoadBalancer must have the same name and
+// namespace as the Service it is associated with.
+type LoadBalancer struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata,omitempty"`
+
+    Spec   LoadBalancerSpec   `json:"spec,omitempty"`
+    Status LoadBalancerStatus `json:"status,omitempty"`
+}
```
aojea, line 363: From existing annotations how many will you aggregate with this object?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  - Setting `Accepted` and `Programmed` conditions on the
+    `LoadBalancer` resource.
+  - Populating `.Status.Addresses` when the load balancer is
+    provisioned.
+  - Falling back to Service.Spec when `LoadBalancer` fields are unset.
+  - Using `LoadBalancer.Spec.Routability` to provision internal vs
+    external load balancers.
+  - Reporting `SupportedFeatures` in status.
+
+### Graduation Criteria
```
aojea, line 911: If the goal is to replace the annotations we have a measurable goal, we need to define the existing annotations we want to migrate and we use the coverage of annotations as criteria.

If there are 100 annotations today, I expect beta to replace at least 75 % of them per example , otherwise we enter in the xkcd thing

Image

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+needs.
+
+### Use Gateway API for all load balancer configuration
+
+Gateway API's `Gateway` resource already models infrastructure with
+conditions and status. However:
+
+- Not all users are on Gateway API. `Service` of type `LoadBalancer`
+  remains the primary mechanism for external load balancing.
+- Gateway API focuses on proxying semantics. Load balancer
```
aojea, line 1238: TCPRoute and UDPRoute are not proxying semantics, the fact that most of use case are via proxy does not mean that is not in scope of that subproject

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  routing/forwarding infrastructure is a separate concern that
+  in-cluster Gateway implementations need but shouldn't have to
```
aojea, line 1240: I think that this is the key problem, we are talking about "in-cluster Gateway implementations" and "out-cluster Gateway implementations" ... is this distinction clearly defined? do we differentiate between in and out of cluster gateway today?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  define.
+- The `LoadBalancer` resource complements Gateway API by providing
```
aojea, line 1242: It look like you are talking only about L4 loadbalancing here?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  about it.
+
+- **Split configuration.** Load balancer configuration split across
+  two resources (Service and LoadBalancer) could confuse users.
+  Documentation and tooling (e.g., `kubectl describe service` showing
+  the associated `LoadBalancer` resource) can mitigate this.
+
+- **Cloud provider adoption burden.** Cloud providers need to update
+  their controllers. The resource is optional, and structured status
+  reporting provides a strong incentive for adoption.
```
aojea, line 1208: is this based on someone feedback?

keps/sig-network/6128-loadbalancer-resource/README.md
aojea: cloud controller manager traditionally only works for L4 loadbalancers

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+    IPAddressAllocations []string `json:"ipAddressAllocations,omitempty"`
+```
+
+<<[/UNRESOLVED]>>
+
+<<[UNRESOLVED subnet-selection]>>
+
+Subnet selection is common but highly implementation-specific. Some
+providers use subnet IDs, others use tags or names. This may be
+better left to annotations or a future KEP.
```
aojea, line 515: 😄

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+}
+
+// LoadBalancerRoutability defines whether a load balancer is
+// externally or internally routable.
+// +enum
+type LoadBalancerRoutability string
+
+const (
+    // LoadBalancerRoutabilityExternal indicates the load balancer
+    // is reachable from the public internet.
```
aojea: is public Internet the right definition for External?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+    // Applicable only when Type is "ClientIP".
+    // +optional
+    Config *corev1.SessionAffinityConfig `json:"config,omitempty"`
+}
+
+// HealthCheckConfig configures how the cloud load balancer
+// performs health checks against backends. These are
+// infrastructure-level health checks performed by the load
+// balancer itself, independent of Kubernetes readiness probes.
+type HealthCheckConfig struct {
```
aojea, line 436: are healtcheck always HTTP?
do they target the backends directly?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+    to the Service. For `SourceRanges`, `nil` falls back to
+    `Service.Spec.LoadBalancerSourceRanges`, while `[]` means "no
+    source range restrictions."
+
+### Risks and Mitigations
+
+**Risk: Configuration drift between Service and LoadBalancer.** Users
+may set conflicting values on the Service and LoadBalancer resource.
+
+*Mitigation:* The precedence rule is clear: `LoadBalancer` resource
```
aojea, line 307: This can be confusing during upgrades/downgrades , specially since loadbalancer controllers do not necessarely follow kubernetes releases that will make this even more challenging ...  so there is the in cluster components skew and the cloud-controller-manager loadbalancer controller vs kubernetes skew

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+fields take precedence when set, with fallback to `Service.Spec`
+fields when unset. Cloud controller managers should set a condition or
```
aojea, line 309: This will break existing workloads ... it sounds more like the LoadBalancer object should always be reconciled to match the Service.Spec if exist to avoid this discrepancy

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+1. **Restrict `loadbalancers` creation** to the same set of
+   principals that can create Services. This is the simplest approach
+   but may be too broad.
+
+2. **Require an explicit opt-in annotation on the Service** (e.g.,
+   `networking.k8s.io/loadbalancer-resource: "true"`) before the
+   cloud controller manager considers a matching `LoadBalancer`
+   resource. This prevents unsolicited overrides.
+
+3. **Use an explicit `serviceRef` field** on the `LoadBalancer`
```
aojea, line 334: This is how IPAddress works

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+**Risk: Lateral configuration override via spec create permissions.**
+A user or controller with `create` permissions on `loadbalancers` in
+a namespace can create a `LoadBalancer` resource that silently alters
+the behavior of any `Service` of type `LoadBalancer` in that namespace
+(e.g., changing routability or source ranges) without the Service
+owner's knowledge.
+
+<<[UNRESOLVED rbac-binding-security]>>
+
+The name-match binding model means that any principal with `create`
```
aojea, line 321: Endpoints are also autogenerated from Services, is not the same problem?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   `networking.gke.io/internal-load-balancer`,
+   `service.beta.kubernetes.io/azure-load-balancer-internal`). This
+   makes multi-cloud deployments and tooling unnecessarily difficult.
+
+2. **Status reporting is inadequate.** `Service.Status.LoadBalancer`
+   only contains a list of ingress IPs/hostnames. There is no
+   structured way for a cloud provider to communicate whether a load
+   balancer is being provisioned, has failed, is serving traffic, or
+   is operating in a degraded mode. This was the primary motivation
+   behind [KEP-4631], which attempted to add conditions directly to
```
aojea, line 139: We should implement this conditions , there was agreement on doing that https://github.com/kubernetes/enhancements/pull/4632#issuecomment-2531726692 , is also a good next step to validate assumptions with zero risk

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+`loadBalancerSourceRanges` instead of maintaining three different sets
+of cloud-specific annotations per environment.
+
+### Notes/Constraints/Caveats
+
+- The `LoadBalancer` resource is bound to a `Service` by name within
+  the same namespace. This follows the same pattern as `Service` and
+  `Endpoints`.
+
+- Cloud controller managers are not required to support the
```
elmiko, line 284: i think it would be cool if we could incorporate some support for LoadBalancer resources in the cloud-provider framework to improve interactions where Services are backed by loadbalancers.

elmiko (commented):
i think this is an interesting idea, we have certainly had some issues in the past few years (mostly around loadbalancer health checks) where i feel like an improved loadbalancer integration would help.

i mostly reviewed for changes around the cloud controller managers, and don't have any request for changes at this time.

rikatz:
sorry folks, I've lost the time for this KEP mostly (actually totally) because of my own lack of organization. I will work on it soon, add some time to answer all of the questions.

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: no, it is not. GKE today solves this with different GatewayClasses https://docs.cloud.google.com/kubernetes-engine/docs/how-to/gatewayclass-capabilities which IMO is far from optimal, when the admin could be defining their own Gateway classes and allowing users to define their LoadBalancer behavior as it is already today with services

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: fixed on my local branch, thanks!

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: depends on what you consider a LoadBalancer in this case. Gateway API mostly does a proxy, not a LoadBalancer. For TLSRoute and above, it will be always a proxy unless you are able to decompose the packets (tls, http) and take the proper routing decision without having to proxy them.

For TCPRoute and UDPRoute I believe this may be a case soon, but I didn't really saw it happening. In fact I have tried to start implementing a TCPRoute load balancer but then, if you are doing it on a in-cluster LB (kube-vip, metallb) you start to bump on things like "where this IP should be announced to be closer to the service? Is a hop acceptable?" and then we are back to services and loadbalancer semantics. So IMO using Gateway API for a LoadBalancer today would just work if you are an external box for loadbalancing.

Also AFAIK no provider can do it as of today, given TCPRoute and UDPRoute are very recently promoted services.

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: your point is that we should have a separate KEP to implement them on services, or that we can use LoadBalancer to implement the missing conditions?

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: applied locally, thanks!

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: applied locally, thanks

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: honest question, did we ended up removing them? if so this can be removed from the motivations IMO and cloud providers should present more of a conformance test than a e2e test.

I tried to find on our k/k codebase but at least on services e2e I couldn't find it, so maybe it was mine (or claude) hallucination

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: "observe load balancer state from a single resource" is more about being able to debug a loadbalancer from a loadbalancer resource. Let me put on another way: I create a LoadBalancer:
* Was it created? if not why - this should be reflected on loadbalancer conditions
* If it was created, what are the IPs associated to it? - same status.

That said, I agree this wording is lame and I will improve a bit

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: IMO this should be part of CCM

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: I see both points, but I think we need to start somewhere. Gateway API could proof that if providers want to be (and keep being) part of validated implementation list, they should follow the direction that the community wants, and not make users hostages of custom annotations. Maybe this is a bit harsh from my side, but internal/external loadbalancer, ip assignment are basic features that any cloud provider follow on their own way, and no discussion even happened to allow a standardization.

That said, I agree that at least an extension point that doesn't make LoadBalancer a new Service (or a new Ingress) is required, something that signals the users there is a "level of compatibility" between a LoadBalancer configuration.

I need to think a bit better in this specific case, but I am happy to hear your thoughts about a bare minimum required bar, and making a case for explicit "non portable" extension points for users

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: For this case I think so. But I also agree with the comment on https://github.com/kubernetes/enhancements/pull/6129/changes#r3335577769 that we may want to add some "extension point" and some compatibility warning if a cloud provider decides to use these extension points

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: see my comment above about proxy semantics and TCPRoute and UDPRoute :) those are new resources, and assuming they can be used for LoadBalancers is also not straightforward, while something like Services already exists.

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: we do: https://gateway-api.sigs.k8s.io/geps/gep-1762/

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: I am, the LoadBalancer is about sending the traffic to the right proxy (Gateway) and Gateway will do its proxy stuff

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: yours :P when you told me you are skeptical about cloud providers adopting it

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: done locally

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: so here is one thing. I am wondering if we want to make this a Kubernetes thing or a Cloud provider thing/API (ping @elmiko ) 

I keep thinking that for me it makes zero sense Kubernetes supporting annotations as "official" for things that are cloud specifics, without no standard (https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer). It actually bothers me a lot and brings me some ingress mood.

I don't want honestly to see new cloud providers coming to Kubernetes docs and adding more annotations that are not portable. And I for sure think that this is a way cloud providers do their lock-in stuff for users.

I totally agree with the e2e case, and that it becomes confusing when you start to think "how will we test this", so yes, probably this must live outside of k/k thinking louder. But also, I want it to have proper outcome. I don't want us to say "yeah, you know what, this is a problem no one will solve, nothing to see here".

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: There will not be annotations. The idea is that the load balancer controller has priority over service when it exists.

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: fixed locally

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: yeah I hate it as well

keps/sig-network/6128-loadbalancer-resource/README.md
rikatz: changed to "outside of cluster network" given using the word "VPC" is very AWS specific

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: We still have loadbalancer tests. They are all behind `[Feature:LoadBalancer]`, but not tagged beyond that, so you have to just know which ones are and aren't expected to pass on particular platforms.

danwinship: Note that in addition to LB tests that pass on some providers but not others (eg, UDP, session affinity), there are also Service features that have no LB tests and we never even defined how/if they interact with load balancers (eg, `trafficDistribution: PreferSameZone`).

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: This probably doesn't work, since you can have `LoadBalancer`s that don't correspond to `Service`s (and thus, someone might create a `LoadBalancer` object with a given name first, and then later create an unrelated LoadBalancer `Service` with the same name).

I'm not sure why we used labels to correlate EndpointSlices to Services rather than having an explicit field... Oh, no, actually, I do know, it's so you can do `kubectl get endpointslices -l kubernetes.io/service-name=myservice`. So the same consideration may apply here.

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: I think I dislike having the semantics fall back like this.

I would say one of
- There's no duplication between Service and LoadBalancer fields
- Duplicate fields are required to be identical between the two objects (... or what?)
- When there are duplicate fields and the Service and LoadBalancer have different values, the fields on the Service apply to non-LB traffic and the fields on the LoadBalancer apply to LB traffic.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+fields take precedence when set, with fallback to `Service.Spec`
+fields when unset. Cloud controller managers should set a condition or
+event when they detect conflicting configuration.
+
+**Risk: Lateral configuration override via spec create permissions.**
+A user or controller with `create` permissions on `loadbalancers` in
+a namespace can create a `LoadBalancer` resource that silently alters
+the behavior of any `Service` of type `LoadBalancer` in that namespace
+(e.g., changing routability or source ranges) without the Service
+owner's knowledge.
```
danwinship, line 317: Yeah, if we're not matching Service and LoadBalancer objects by name, it may be best to say that Services have to point to their LoadBalancers

keps/sig-network/6128-loadbalancer-resource/README.md
elmiko: imo, this would be cool

keps/sig-network/6128-loadbalancer-resource/README.md
elmiko: imo, we do want some sort of common tests around load balancers but i think we need them to be coupled to the cloud-provder staging repo, and then have an interface for providers to implement.

i have been working with a colleague (@miyadav) to build a PoC of these style of tests. ideally we should be able to have a test that says something like "when you make an LB service, the LB actually gets created" and then provider implementers can utilize those.

happy to share more details, we don't have these done yet but we've been working on them.

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: I guess part of the question here is, does GKE's current approach make sense? It's not _obviously_ wrong that "externally-reachable load balancer" and "cluster-internal load balancer" should be different GatewayClasses.

What do non-`Service`/`type: LoadBalancer`-based Gateways do? Since Envoy/Istio/etc don't have to worry about exposing their options in a way that is compatible with legacy k8s APIs/components underneath, presumably they implement this in the preferred/logical/"obvious" way, and we should want GKE/AWS/etc to implement it the same way?

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: The KEP probably needs to be more precise with its terminology. e.g., it's not clear here if "load balancer" means specifically "Gateways backed by Services of `type: LoadBalancer`" or not.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+[kubernetes/kubernetes#123714]: https://github.com/kubernetes/kubernetes/issues/123714
+
+### Goals
+
+- Define a new namespaced `LoadBalancer` resource in the
+  `networking.k8s.io/v1alpha1` API group with explicit fields for
+  common load balancer configuration.
+
+- Provide structured status reporting with conditions (`Accepted`,
+  `Programmed`, `Degraded`) aligned with Gateway API conventions.
```
danwinship, line 175: There was some discussion about LoadBalancers and Gateway's conditions in https://github.com/kubernetes/enhancements/pull/4632#discussion_r1796998336. Even after resolving the question there, I don't think we should assume that Gateway conditions will necessarily map perfectly to LB conditions.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  `Service` resource remains the primary way to expose workloads.
+
+- Requiring all cloud providers to support the new resource. Cloud
+  controller managers opt in to `LoadBalancer` resource support.
+
+- Moving service-routing concerns (`externalTrafficPolicy`,
+  `internalTrafficPolicy`, `ipFamilies`, `ipFamilyPolicy`,
+  `allocateLoadBalancerNodePorts`) to the `LoadBalancer` resource.
+  These are service behavior, not load balancer infrastructure
+  concerns.
```
danwinship, line 204: You need `ipFamilies` if you want to be able to select IPv4-only LB vs IPv6-only LB vs dual-stack LB.

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: This user story is confusing because it appears to describe the user using the API to do something that the API does not support doing? (Like, maybe it originally said `LoadBalancerClass` and you just search+replaced it?)

In general, user stories should not pre-suppose the solution you are proposing. They should talk about what the user wants to do, not how they will use the new API to do it.

If the API you're proposing doesn't actually solve this problem (or solves it badly), then this may fit better under "Alternatives".

> * Gateway - uses loadBalancerClass internal
> * Implementation - Creates a loadbalancer and service resource, point the loadbalancer to that class
> * waits for the loadbalancer status to be reported
> * now you have a class-level configuration for loadbalancer, not a per-lb specific

Still confusing. Does the first step mean "The user creates a `Gateway` resource whose `gatewayClassName` points to a `GatewayClass` corresponding to an internal load balancer"? Or maybe "Some implementation of the Gateway API creates a `LoadBalancerClass` object in response to the user creating a `Gateway`"?

Likewise, what is the "Implementation" in the second step? The Gateway API implementation? The cloud-provider implementation?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   resource. This prevents unsolicited overrides.
+
+3. **Use an explicit `serviceRef` field** on the `LoadBalancer`
+   resource (see Binding Model section) which could enable
+   admission-time validation of cross-resource permissions.
+
+4. **Rely on namespace-level trust boundaries.** Kubernetes
+   namespaces are already trust boundaries — principals with write
+   access to a namespace can already modify Services, Endpoints, and
+   other resources that affect networking behavior.
```
danwinship, line 341: I think I vote for this. Namespaces are our trust boundaries. We don't generally worry about ways in which one user in a namespace might "attack" another user in the same namespace.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+    // controller manager falls back to Service.Spec.SessionAffinity
+    // and Service.Spec.SessionAffinityConfig.
+    // +optional
+    SessionAffinity *LoadBalancerSessionAffinity `json:"sessionAffinity,omitempty"`
+
+    // SourceRanges restricts traffic to the load balancer to the
+    // specified client IP ranges. If nil, the cloud controller
+    // manager falls back to Service.Spec.LoadBalancerSourceRanges.
+    // An explicit empty list means "no source range restrictions"
+    // and does not fall back to the Service.
```
danwinship, line 385: (You can't distinguish `nil`, `[]`, and `&[]`. You _can_ distinguish `nil` from `&{[]}` (pointer to struct containing an empty array)

keps/sig-network/6128-loadbalancer-resource/README.md
danwinship: needs to validate as `isCIDR(val) && cidr(val).isMask()`, to allow only subnet/mask CIDRs

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+    // healthy.
+    // +optional
+    HealthyThreshold *int32 `json:"healthyThreshold,omitempty"`
+
+    // UnhealthyThreshold is the number of consecutive failed
+    // health checks required before a backend is considered
+    // unhealthy.
+    // +optional
+    UnhealthyThreshold *int32 `json:"unhealthyThreshold,omitempty"`
+}
```
danwinship, line 467: How portable is this? Do at least GCP, AWS, and Azure support every one of the fields there?

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+different providers model this differently:
+
+- AWS: `service.beta.kubernetes.io/aws-load-balancer-eip-allocations`
+  (references to Elastic IP allocation IDs)
+- GKE: `networking.gke.io/load-balancer-ip-addresses` (references
+  to GCP address resources)
+- Azure: Public IP resource IDs
+
+A potential approach is a string slice of provider-interpreted
+references:
```
danwinship, line 499: Not clear how that's better than provider-specific annotations

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+a class, and it conflicts with the Service's class, the behavior is
+undefined. Options:
+
+1. The `LoadBalancer` resource's class takes precedence.
+2. The class must match if both are set, otherwise the cloud
+   controller sets an error condition.
+3. The class field only lives on the `LoadBalancer` resource and is
+   removed from Service in a future version.
+
+<<[/UNRESOLVED]>>
```
danwinship, line 533: You don't actually include "class" on the LB currently. If you do, it should work the same way as all of the other duplicated fields.

Note that `loadBalancerClass` exists to allow people to have `LoadBalancer` Services that are backed by in-cluster implementations, so this isn't really needed in your model anyway. (An Envoy-based Gateway implementation is not going to provision an nginx-based `LoadBalancer` Service as the backend of a `Gateway`.)

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+  information to determine whether the load balancer is degraded.
+
+<<[UNRESOLVED degraded-condition]>>
+
+The `Degraded` condition serves the same purpose as
+`LoadBalancerDegraded` from KEP-4631. An alternative is to remove
+this condition and rely solely on the feature reporting mechanism
+(`SupportedFeatures`) combined with `Accepted: False` for hard
+failures. The `Degraded` condition provides a more immediate signal
+that something is "working but not as requested."
```
danwinship, line 729: `Accepted: False` would be a backward-incompatible change for existing cloud providers. Currently if a Service requests a feature that the cloud doesn't support, the cloud provider just ignores the unsupported feature.

keps/sig-network/6128-loadbalancer-resource/README.md
```diff
@@ -0,0 +1,1258 @@
...
+   primary object. This is unusual in Kubernetes and likely not the
+   right model.
+
+3. **No ownership**: Both resources are independent. The cloud
+   controller manager observes both and reconciles. Deleting either
+   requires explicit cleanup by the user or controller.
+
+4. **Cloud provider decides**: The ownership model is left to the
+   cloud controller manager implementation. This provides flexibility
+   but may lead to inconsistent behavior across providers.
```
danwinship, line 777: No

keps/sig-network/6128-loadbalancer-resource/kep.yaml
```diff
@@ -0,0 +1,33 @@
+title: LoadBalancer Resource
```
jpbetz, line 1: Please add a keps/prod-readiness/sig-storage/6128.yaml and list me as the PRR reviewer for alpha.

jpbetz:
Reminder that the above comments need to be addressed by EOD to meet the KEP readiness deadline. xref: https://github.com/kubernetes/enhancements/issues/6128#issuecomment-5705078311

PR #6130, opened 2026-05-29
KEP-5526: Pod Level Resource Managers 1.37

KevinTMtz:
- One-line PR description:

- Issue link: https://github.com/kubernetes/enhancements/issues/5526

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-node/5526.yaml
+++ b/keps/prod-readiness/sig-node/5526.yaml
@@ -1,3 +1,5 @@
 kep-number: 5526
 alpha:
   approver: "@jpbetz"
+beta:
+  approver: "@jpbetz"
--- a/keps/sig-node/5526-pod-level-resource-managers/README.md
+++ b/keps/sig-node/5526-pod-level-resource-managers/README.md
@@ -31,7 +31,7 @@ To get started with this template:
   appropriate SIG(s).
 - [X] **Create a PR for this KEP.**
   Assign it to people in the SIG who are sponsoring this process.
-- [ ] **Merge early and iterate.**
+- [X] **Merge early and iterate.**
   Avoid getting hung up on specific details and instead aim to get the goals of
   the KEP clarified and merged quickly. The best way to do this is to just
   start with the high-level sections and fill out details incrementally in
@@ -84,7 +84,9 @@ Ensure the TOC is wrapped with
 tags, and then generate with `hack/update-toc.sh`.
 -->
 
+<!-- mdformat off() -->
 <!-- toc -->
+
 - [Release Signoff Checklist](#release-signoff-checklist)
 - [Glossary](#glossary)
 - [Summary](#summary)
@@ -110,11 +112,16 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Policy Options](#policy-options-1)
     - [Pod Scope Allocation and Partitioning Algorithm](#pod-scope-allocation-and-partitioning-algorithm)
     - [State Management and Container Removal](#state-management-and-container-removal)
+      - [Compatibility with Generalized V4 State](#compatibility-with-generalized-v4-state)
     - [Interaction with CPU Quota Management](#interaction-with-cpu-quota-management)
   - [Memory Manager](#memory-manager)
     - [Policies](#policies-2)
     - [Pod Scope Allocation and Partitioning Algorithm](#pod-scope-allocation-and-partitioning-algorithm-1)
     - [State Management and Container Removal](#state-management-and-container-removal-1)
+      - [Compatibility with Generalized V4 State](#compatibility-with-generalized-v4-state-1)
+  - [PodResources API](#podresources-api)
+      - [Protobuf API Updates](#protobuf-api-updates)
+      - [Illustrative output](#illustrative-output)
   - [Future Enhancements and Long-Term Vision](#future-enhancements-and-long-term-vision)
   - [Feature Gate](#feature-gate)
   - [Test Plan](#test-plan)
@@ -124,6 +131,7 @@ tags, and then generate with `hack/update-toc.sh`.
       - [e2e tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
+    - [Beta](#beta)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
@@ -138,6 +146,7 @@ tags, and then generate with `hack/update-toc.sh`.
 - [Alternatives](#alternatives)
 - [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
 <!-- /toc -->
+<!-- mdformat on -->
 
 ## Release Signoff Checklist
 
@@ -176,8 +185,8 @@ release*.
         within one minor version of promotion to GA
 -   [x] `R` Production readiness review completed
 -   [x] `R` Production readiness review approved
--   [ ] "Implementation History" section is up-to-date for milestone
--   [ ] User-facing documentation has been created in [kubernetes/website], for
+-   [x] "Implementation History" section is up-to-date for milestone
+-   [x] User-facing documentation has been created in [kubernetes/website], for
     publication to [kubernetes.io]
 -   [ ] Supporting documentation—e.g., additional design documents, links to
     mailing list discussions/SIG meetings, relevant PRs/issues, release notes
@@ -536,6 +545,34 @@ New behavior <br> Pod Level Resources <br> Some Guaranteed containers | Pod: 5 C
 New behavior <br> Pod Level Resources <br> No Guaranteed containers   | Pod: 5 CPU <br> Container 1 <br> Container 2 <br> Container 3                      | {NUMA node 0}   | Container 1: {1,2,3,4,5} <br> Container 2: {1,2,3,4,5} <br> Container 3: {1,2,3,4,5}
 New behavior <br> Pod Level Resources <br> Admission Failure          | Pod: 5 CPU <br> Container 1: 3 CPU <br> Container 2: 2 CPU <br> Container 3        | Admission Error | Pod Rejected
 
+###### Ephemeral Containers
+
+For pods under the `pod` scope:
+
+-   **No Resource Reservation:** Ephemeral containers do not participate in the
+    partition of the pod's pool and are not allocated exclusive resources from
+    the pod-level budget.
+-   **Node-level Resource Inheritance:** Ephemeral containers inherit the node's
+    default CPU set (the node-wide shared pool) rather than being confined to
+    the pod's assigned CPU set or the pod's shared pool.
+-   **Pod-level Limit Enforcement:** Although ephemeral containers run with the
+    node's default CPU set, they are executed inside the pod's cgroups. They
+    remain restricted by the overall pod-level resource limits (e.g., CPU quota
+    and memory limits) defined on the pod.
+-   **Troubleshooting Utility:** Pinned CPU sets allocated to Guaranteed pod
+    containers can become fully saturated by active workloads. By escaping the
+    pod's restricted CPU set, ephemeral containers can run on any available CPU
+    in the node-wide shared pool. This ensures they can be scheduled and execute
+    for debugging, even if the pod's exclusively allocated CPUs are fully
+    saturated. Furthermore, latency-sensitive workloads expect zero performance
+    interference on their exclusively assigned CPUs. Running ephemeral
+    containers on the node-wide shared pool (and specifically not on the pod's
+    exclusively allocated CPUs) ensures that troubleshooting/debugging
+    operations do not introduce CPU contention or cache disruption for the
+    primary workload. However, because they are inside the pod's cgroup, they
+    still share the overall pod-level cgroup quota limits with the regular
+    containers.
+
 ###### Rejected Configurations
 
 **Empty podSharedPool with non-Guaranteed containers**
@@ -597,6 +634,7 @@ shared resource pool.
 <!-- mdformat off() -->
 
 **Pod Spec:**
+
 ```yaml
 apiVersion: v1
 kind: Pod
@@ -639,6 +677,7 @@ use the remainder.
 <!-- mdformat off() -->
 
 **Pod Spec:**
+
 ```yaml
 apiVersion: v1
 kind: Pod
@@ -687,6 +726,7 @@ admission.
 <!-- mdformat off() -->
 
 **Pod Spec:**
+
 ```yaml
 apiVersion: v1
 kind: Pod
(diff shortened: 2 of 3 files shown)


jpbetz:
Please update `keps/prod-readiness/sig-node/5526.yaml` as part of this KEP update. You can assign me as PRR approver.

keps/prod-readiness/sig-node/5526.yaml
```diff
@@ -1,3 +1,3 @@
 kep-number: 5526
-alpha:
+beta:
   approver: "@jpbetz"
```
jpbetz, line 5:
```suggestion
alpha:
  approver: "@jpbetz"
beta:
  approver: "@jpbetz"
```

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1791,12 +1915,23 @@ would also warrant investigation and potential rollback.
...
+-   **Memory Manager Checkpoint:** Unit tests verify forward compatibility of
+    V1/V2 checkpoints, migration flow from V1 to V2, and verify that the Kubelet
+    correctly rejects corrupted checkpoints under rollback scenarios.
+
+E2E upgrade/downgrade/rollback scenarios are covered by the following DRA test
+workflows:
+
+-   `pull-kubernetes-kind-dra-n-1`
+-   `pull-kubernetes-kind-dra-n-2`
+-   `pull-kubernetes-kind-dra-n-3`
```
jpbetz: Do we need to update [run_if_changed](https://github.com/kubernetes/test-infra/blob/5fcd8c67b8870d58b9878e00a5225d1f6d87c09c/config/jobs/kubernetes/sig-node/dra-presubmit.yaml#L11) so that these test run for pod level resource code changes? Or is there another test strategy we need to put in place?

keps/sig-node/5526-pod-level-resource-managers/README.md
jpbetz, line 1945: If I'm understanding correctly, after disabling the feature gate, checksum validation can fail and the next step is to drain the node and delete the memory manager checkpoint file?

Let me know if I'm misunderstanding. Goal is for a cluster admin to be able to follow the instructions here to get back to safety.

jpbetz: @tallclair should we do more to make it easier for cluster admins to deal with checkpoint file during downgrade or is this typically how we handle it?

jpbetz (commented):
A few comments then LGTM for PRR.

keps/sig-node/5526-pod-level-resource-managers/README.md
jpbetz: cc @pohly

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -536,6 +543,28 @@ New behavior <br> Pod Level Resources <br> Some Guaranteed containers | Pod: 5 C
...
+    default CPU set (the node-wide shared pool) rather than being confined to
+    the pod's assigned CPU set or the pod's shared pool.
+-   **Pod-level Limit Enforcement:** Although ephemeral containers run with the
+    node's default CPU set, they are executed inside the pod's cgroups. They
+    remain restricted by the overall pod-level resource limits (e.g., CPU quota
+    and memory limits) defined on the pod.
+-   **Troubleshooting Utility:** Pinned CPU sets allocated to Guaranteed pod
+    containers can become fully saturated by active workloads. By escaping the
+    pod's restricted CPU set, ephemeral containers can run on any available CPU
+    in the node-wide shared pool. This ensures they can be scheduled and execute
```
ffromani, line 565: Furthermore, latency-sensitive workload want to have no interference on the exclusively assigned CPUs, so running ephemeral containers on the pod-shared pool (or anyway NOT on the exclusively assigned CPUs) is definitely the right move

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1292,6 +1350,43 @@ back to the node's free pool and delete the entry from `PodMemoryAssignments`.
...
+##### Compatibility with Generalized V4 State
+
+Following the CPU Manager's adoption of the generalized V4 state checkpoint
+format, the Memory Manager checkpoint format is planned to be updated similarly
+in a future release.
+
+For the GA milestone of this feature, the Memory Manager's pod-level resource
+manager support (specifically `PodMemoryAssignments`) will be formally tied to
+and dependent on the Memory Manager adopting the same generalized V4 checkpoint
+format structure.
```
ffromani, line 1370: cc @lukaszwojciechowski let's make sure the memory manager counterpart of the V4 checkpoint state lands in 1.37

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1292,6 +1350,43 @@ back to the node's free pool and delete the entry from `PodMemoryAssignments`.
...
+    selectively:
+    -   If a container is allocated exclusive resources individually (e.g., a
+        Guaranteed container with exclusive CPU/memory), its allocated CPU and
+        memory are reported in the container-level `cpu_ids` and `memory`
+        fields.
+    -   If a container shares resources within the pod's budget (running in the
+        pod's shared pool or the node's shared pool), the container-level
+        `cpu_ids` and `memory` fields are returned empty. Instead, these
+        allocations are surfaced at the pod level in the new `PodResources`
+        fields.
```
ffromani, line 1396: need to think if it is better to report empty or omit these entirely. I tend to believe report empty (plus another optional flag?) is better here, but I wanted to doublecheck

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1520,7 +1621,17 @@ Below are some examples to consider, in addition to the aforementioned [maturity
 - All known pre-release issues and gaps resolved
 
 **Note:** Beta criteria must include all functional, security, monitoring, and testing requirements along with resolving all issues and gaps identified
+-->
+
+-   The `PodLevelResourceManagers` feature gate is enabled by default.
+-   The PodResources API is updated to reflect the pod-level exclusive resource
+    allocation for both CPU and memory resources.
```
ffromani: I'd add `podresources API changes include e2e tests` which is probably implied but should be mentioned explicitly

keps/sig-node/5526-pod-level-resource-managers/README.md
ffromani: This is likely to be solved by the changes we're making in parallel to the cpumanager and memorymanager checkpoint format. The only remaining gap is if an user
1. opted-in in the alpha feature in 1.36
2. is upgrading to 1.37

this is the one and only known case on which the checkpoint data is incompatible and we need to remove the checkpoints.

cc @KevinTMtz let's make sure this is indeed the case

ffromani (commented):
provisional LGTM from sig-node. The only thing left to fully iron out is to shape of podresources API changes. I'm mostly OK with the proposed approach but I wanted to take a bit of time to fully explore

keps/prod-readiness/sig-node/5526.yaml
KevinTMtz: Done, thank you.

keps/sig-node/5526-pod-level-resource-managers/README.md
KevinTMtz: > Do we need to update [run_if_changed](https://github.com/kubernetes/test-infra/blob/5fcd8c67b8870d58b9878e00a5225d1f6d87c09c/config/jobs/kubernetes/sig-node/dra-presubmit.yaml#L11) so that these test run for pod level resource code changes?

The dra workflows are helpful, since problems in recent state changes have been caught by failures in dra: https://github.com/kubernetes/kubernetes/issues/137750, https://github.com/kubernetes/kubernetes/issues/139069. However, it might be better to have a dedicated workflow for the resource managers in general. Wdyt @ffromani?

keps/sig-node/5526-pod-level-resource-managers/README.md
KevinTMtz: Integrated this into the description.

keps/sig-node/5526-pod-level-resource-managers/README.md
KevinTMtz: Added explicit mention.

keps/sig-node/5526-pod-level-resource-managers/README.md
ffromani: We are (ab)using the DRA tests because these are the ones readily available which implement version skew, which in turn enable us to test the upgrade flow.
I'd argue the proper fix is to setup similar lanes for the builtin resource managers: we need version skew tests, and the resource manager testing doesn't belong to DRA lanes.

KevinTMtz: I updated the KEP to mention the addition of a dedicated version skew test workflow for the resource managers.

keps/sig-node/5526-pod-level-resource-managers/README.md
KevinTMtz: @jpbetz

> should we do more to make it easier for cluster admins to deal with checkpoint file during downgrade or is this typically how we handle it?

Historically, draining the node and clearing/deleting the checkpoint files has been the standard operational procedure when downgrading the Kubelet or disabling resource manager feature gates, as forward compatibility was not guaranteed. Until the checkpoint formats updates are completed, we should expect to continue with this procedure for downgrades.

> This is likely to be solved by the changes we're making in parallel to the cpumanager and memorymanager checkpoint format. The only remaining gap is if an user
> 
>   1. opted-in in the alpha feature in 1.36
>   2. is upgrading to 1.37

I confirm that is indeed the only remaining gap. This happens because the generalized V4 checkpoint format embeds the V2 checkpoint data inline for backward compatibility, but does not embed the Alpha V3 structure. As a result, a Kubelet running the 1.36 Alpha version will fail checksum verification on downgrade/rollback.

keps/sig-node/5526-pod-level-resource-managers/README.md
ffromani: we're on track: https://github.com/kubernetes/kubernetes/pull/139626

keps/sig-node/5526-pod-level-resource-managers/README.md
ffromani: crosslinking https://github.com/kubernetes-sigs/node-feature-discovery/pull/2509#issuecomment-4488489247 , the answer is:
- podresources API wants to report _exclusive_ assignment. If a resource is not exclusive for whatever reason, is not reported
- report empty slices for non-exclusive resources
- in the case of pod-level excusive resources subsequently allocated to a container, the exclusive resource should be represented _both_ at pod-level and at container-level (I think we already agreed about this, I'm mentioning for completeness in this context)

@KevinTMtz please clarify (another bullet point should be sufficient) how pod-level exclusive resources interact with container-level resources *in the context of the podresources API reporting*. Worth adding mock-ups of how the podresources API output would look like. The examples are meant to be illustrative but realistic, so the implementation is not bound to give byte-identical output.

ffromani (commented):
provisional LGTM once we clarify further the podresources API portion. We need, I believe, a simple extra bullet point and some examples. Once that is done, pending a final due diligence pass, LGTM from sig-node perspective.

keps/sig-node/5526-pod-level-resource-managers/README.md
lukaszwojciechowski: I didn't get notification about previous comment from github, but the PR is ready and verified

keps/sig-node/5526-pod-level-resource-managers/README.md
lukaszwojciechowski: The only not working transitions are:
| scenario | from | to | result |
| --- | --- | --- | --- |
| downgrade | 1.36 with opted-in alpha | 1.35 | checksum fail |
| upgrade | 1.36 with opted-in alpha | 1.37 | checksum fail |
| downgrade | 1.37 with beta | 1.36 with opted-in alpha | loading CPU checkpoint succeeds but in V2 format - pod level data is lost, but operator might expect feature to work |

In fact every downgrade from 1.37 using V4 checkpoint in CPU manager to 1.36 or lower will cause lose of pod level resources data (as it is loaded as embedded V2 version).
So a round-trip scenario 1.37 -> 1.36 -> 1.37 will also lose the data if there are any changes in 1.36 and checkpoint is overwritten.

All other transitions work fine.

keps/sig-node/5526-pod-level-resource-managers/README.md
KevinTMtz: I added a new paragraph to be explicit about the reporting behavior at both levels when dealing with exclusive resources, and I also added an illustrative output section for three different scenarios.

ffromani (commented):
/lgtm

thanks for addressing all my comments

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1677,35 +1986,40 @@ This section must be completed when targeting beta to a release.
...
+    *   **CPU Manager (Safe Rollback):** The CPU Manager utilizes a generalized
+        V4 checkpoint format that embeds the legacy V2 checkpoint data inline
+        and implements a dual-checksum format. When the feature gate is disabled
+        or the Kubelet is downgraded to a version only supporting V2, the
+        Kubelet can parse the embedded V2 payload and verify its checksum,
+        safely ignoring the newer V4 fields.
+    *   **Memory Manager (Higher Risk):** The Memory Manager does not support
+        forward compatibility. Disabling the feature gate (or downgrading to a
+        V1-only Kubelet) will cause the Kubelet to fail checksum validation of
+        the V2 state file, resulting in a startup failure.
```
jpbetz, line 2018: Should an issue be opened to track the need to add support for for future compatibility (so if we ever do a V3 the migration story is better?) or is that infeasible?

jpbetz (approved):
/approve
For PRR
(one non-blocking suggestion about future compatibility of the mem manager)

keps/sig-node/5526-pod-level-resource-managers/README.md
lukaszwojciechowski: @jpbetz 
There is a PR fixing that issue, that generalizes Memory Manager checkpoint file similar to solution merged for CPUManager: https://github.com/kubernetes/kubernetes/pull/139626
It introduces V3 version embedding V1 (similar to V4 embedding V2 for CPUs)

With that being merged forward compatibility will be supported and rollback scenario will be safe.

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -785,6 +825,19 @@ New behavior <br> Pod Level Resources <br> All Guaranteed containers  | Pod: 5 C
...
+For pods under the `container` scope:
+
+-   **Current Behavior Preserved:** Since there is no pod-level resource
+    allocation, ephemeral containers continue to follow their current behavior.
+    They do not receive exclusive allocations and run in the node-wide shared
+    pool.
+-   **Pod-level Limit Enforcement:** If the pod defines pod-level resource
+    requests/limits, ephemeral containers (along with all other containers in
+    the pod) remain restricted by the overall pod-level cgroup limits at
+    runtime.
```
ndixita, line 839: Consider adding that if ephemeral containers can't be accomodated on the pod-level resources boundaries, it will be required to resize the pod first before injecting ephemeral containers.

keps/sig-node/5526-pod-level-resource-managers/README.md
```diff
@@ -1292,6 +1358,224 @@ back to the node's free pool and delete the entry from `PodMemoryAssignments`.
...
+
+Following the CPU Manager's adoption of the generalized V4 state checkpoint
+format, the Memory Manager checkpoint format is planned to be updated similarly
+in a future release.
+
+For the GA milestone of this feature, the Memory Manager's pod-level resource
+manager support (specifically `PodMemoryAssignments`) will be formally tied to
+and dependent on the Memory Manager adopting the same generalized V4 checkpoint
+format structure.
+
```
ndixita, line 1371: Do we know when memory manager is adopting the v4 checkpoint? Will there be any issues because of the gap? Please considering adding the tracking issue link for memory manager checkpoint upgrade to v4

ffromani:
/approve

as sig-node delegate

Issue #6132, opened 2026-05-31
PreQueueingHint Extension Point for Scheduler Event Processing


geetasg:
### Enhancement Description

   - One-line enhancement description (can be used as a release note): Add PreQueueingHintFn extension point to
   the scheduling framework, allowing plugins to narrow the set of pods evaluated on cluster events from O(N)
   to O(1).
   - Kubernetes Enhancement Proposal:
   https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/NNNN-prequeueing-hints
   - Discussion Link: https://github.com/kubernetes/kubernetes/pull/138531#discussion_r3181733527
   - PRs by stage and milestone:
     - [ ] Alpha - v1.37 (straight to beta, but change to alpha later)
       - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6133
       - [x] Code (`k/k`) update PR(s): 
         - [x] https://github.com/kubernetes/kubernetes/pull/138916
         - [x] https://github.com/kubernetes/kubernetes/pull/140959
       - [ ] Docs (`k/website`) update PR(s):
   - [ ] Beta - v1.38
     - [ ] KEP (`k/enhancements`) update PR(s):
     - [ ] Code (`k/k`) update PR(s):
       - [ ] https://github.com/kubernetes/kubernetes/pull/141732
     - [ ] Docs (`k/website`) update(s):

pohly:
Let's use a shorter title: "PreQueueingHint" seems sufficient to me. If I cut-and-paste the current one into e.g. https://docs.google.com/presentation/d/1COH8dG8qZ9jEMXzQVdco74XmroZQTMYcUgwAES8FYWU/edit?usp=sharing&resourcekey=0-RCm3FFt8pxiK-y-_YHUkOg it's too wide...

johnbelamaric:
/stage beta

Hoping for straight-to-beta since there is no API.

whtssub:
Hello @geetasg  👋, v1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **PRR Freeze on Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `beta` for v1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

[...]

PR #6133, opened 2026-05-31
KEP-6132: PreQueueingHint extension point for scheduler event processing

geetasg:
- One-line PR description: Adding new KEP for PreQueueingHintFn extension point in the scheduling framework

 - Issue link: https://github.com/kubernetes/enhancements/issues/6132

- Other comments: Implementation PR: https://github.com/kubernetes/kubernetes/pull/138916

KEP geenration is assisted by kiro-cli

### DIFF
--- a/keps/sig-scheduling/6132-prequeueing-hints/README.md
+++ b/keps/sig-scheduling/6132-prequeueing-hints/README.md
@@ -1,0 +1,452 @@
+# KEP-6132: PreQueueingHint Extension Point for Scheduler Event Processing
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [PreQueueingHintFn Interface](#prequeueinghintfn-interface)
+  - [DRA Plugin Implementation](#dra-plugin-implementation)
+  - [Feature Gate](#feature-gate)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+    - [E2E Tests](#e2e-tests)
+    - [Performance Tests](#performance-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Alternatives](#alternatives)
+  - [PodGroup (GenericWorkload) Interaction](#podgroup-genericworkload-interaction)
+  - [Other Alternatives](#other-alternatives)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone
+/ release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP
+  dir in [kubernetes/enhancements]
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG
+  Architecture and SIG Testing input
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in
+  [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation
+
+## Summary
+
+This KEP introduces a `PreQueueingHintFn` extension point to the
+Kubernetes scheduling framework. When a cluster event occurs, the
+scheduler currently iterates all unschedulable pods to evaluate
+per-pod `QueueingHintFn`. The new `PreQueueingHintFn` is called
+once per event *before* this iteration, allowing plugins to return
+a targeted list of pods to check — or signal that no pods need
+checking. This reduces the requeue path from O(N) to O(1) for
+events that affect only specific pods.
+
+The DRA plugin implements `PreQueueingHintFn` for ResourceClaim
+events using a pod informer index that maps ResourceClaims to the
+pods referencing them.
+## Motivation
+
+When a burst of pods using DRA ResourceClaimTemplates are created
+simultaneously (e.g., a large Deployment scaling up), each pod
+generates a ResourceClaim. Each ResourceClaim creation or allocation
+event triggers `MoveAllToActiveOrBackoffQueue`, which iterates all
+unschedulable pods to evaluate per-pod `QueueingHintFn`. With N pods
+in the unschedulable queue and N ResourceClaim events, the total
+work is O(N²).
+
+This quadratic behavior becomes the dominant CPU cost in the
+scheduler, limiting scheduling throughput during burst workload
+creation. The existing `QueueingHintFn` is per-pod (called once per
+unschedulable pod per event). There is no mechanism to narrow the
+candidate set *before* iterating all pods.
+
+### Goals
+
+- Provide a `PreQueueingHintFn` extension point that plugins can
+  implement to narrow the set of pods evaluated on cluster events.
+- Reduce ResourceClaim event processing from O(N) to O(1) for
+  claims referenced by a known set of pods.
+- Maintain correctness: provide an explicit "evaluate all" signal
+  for events where the affected pod set cannot be determined.
+
+### Non-Goals
+
+- Replacing or modifying the existing `QueueingHintFn` mechanism.
+- Optimizing events other than ResourceClaim (though the framework
+  supports it).
+- Changing the scheduling algorithm or pod priority ordering.
+
+## Proposal
+
+### User Stories
+
+**Story 1: Large-scale DRA workloads**
+
+As a cluster operator rapidly creating thousands of pods (e.g., 10K)
+with DRA ResourceClaimTemplates, I want the scheduler to maintain
+high throughput without being bottlenecked by ResourceClaim event
+processing.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|-----------|
+| PreQueueingHintFn incorrectly signals "no pods affected" | Deallocation events and index errors always signal "evaluate all". The periodic flush of unschedulable pods provides an additional safety net. |
+| A generated claim referenced by a second pod | Pod informer index looks up all pods referencing the claim by spec, not just the owner. Verified by E2E test. |
+| Pod informer index returns stale empty result | This is correct: if no pod references the claim yet, no pod can benefit from the event. The pod status update (linking pod to claim) triggers a separate scheduling event. |
+
+## Design Details
+
+### PreQueueingHintFn Interface
+
+A new optional function is added to `ClusterEventWithHint` in the
+scheduling framework. When registered, it is invoked once per
+matching cluster event inside `moveAllToActiveOrBackoffQueue`,
+*before* the loop that iterates unschedulable pods.
+
+The function returns a `PreQueueingHintResult` struct:
+- `AllPods bool`: if true, all unschedulable pods should be
+  evaluated (existing behavior). The pod list is ignored.
+- `Pods []types.NamespacedName`: when `AllPods` is false, only
+  the listed pods are evaluated by the per-pod `QueueingHintFn`.
+  An empty or nil slice means no pods need evaluation for this
+  event.
+
+If no `PreQueueingHintFn` is registered for an event, the framework
+falls back to evaluating all pods (existing behavior).
+
+Multiple plugins may register a `PreQueueingHintFn` for the same
+event. The framework tracks PreQueueingHint results per-plugin:
+each plugin's QueueingHint is only invoked for the pods its own
+PreQueueingHintFn identified. Plugins without PreQueueingHintFn
+default to "all pods" only for their own QueueingHint evaluation.
+If a plugin signals "all pods", the framework evaluates all pods
+for that plugin's QueueingHint only.
+
+### DRA Plugin Implementation
+
+The DRA plugin maintains a pod informer index that maps each
+ResourceClaim (by namespace/name) to the pods that reference it
+in their `spec.resourceClaims`. This index is registered at plugin
+initialization via the shared pod informer.
+
+When a ResourceClaim event occurs:
+
+1. **Deallocation**: If the claim's allocation transitions from
+   non-nil to nil, signal "evaluate all pods" — freed capacity
+   may make any waiting pod schedulable.
+2. **Other events**: Look up all pods referencing the claim via
+   the informer index. Return their NamespacedNames with
+   `allPods=false`. If the index lookup fails, signal
+   "evaluate all pods" as a safe fallback.
+
+This avoids the OwnerReference heuristic entirely. The index
+correctly handles the case where multiple pods reference the
+same generated claim.
+
+### Feature Gate
+
+`SchedulerPreQueueingHints` (beta, default=true):
+- When disabled: `PreQueueingHintFn` is ignored, all pods are
+  evaluated on every event (existing behavior).
+- When enabled: `PreQueueingHintFn` narrows the pod set before
+  per-pod QueueingHint evaluation.
+
+### Test Plan
+
+#### Prerequisite testing updates
+
+The existing QueueingHint tests provide baseline coverage for the
+requeue path.
+
+#### Unit Tests
+
+- `TestPreQueueingHint`: Tests pod lookup via index, deallocation
+  fallback, empty index (no pods affected), error handling.
+- `TestPriorityQueue_PreQueueingHint`: Tests gate enabled/disabled,
+  hint narrows to specific pods, hint signals all, multiple hints
+  union.
+
+#### Integration Tests
+
+- Existing scheduler integration tests pass with feature gate
+  enabled and disabled.
+- DRA integration tests (`pull-kubernetes-dra-integration`) pass.
+- Integration test verifying PreQueueingHintFn correctly narrows
+  the pod set and that missed pods are rescued by periodic flush.
+
+#### E2E Tests
+
+- `pull-kubernetes-e2e-kind-alpha-beta-features`: Full e2e with
+  all alpha/beta gates enabled.
+- Dedicated test: multiple pods sharing one generated ResourceClaim
+  all get scheduled after driver starts (verifies the index-based
+  approach handles shared claims correctly).
+
+#### Performance Tests
+
+Measured using `test/integration/scheduler_perf` with the
+`SchedulingWithResourceClaimTemplate` benchmark:
+
+- Before: ~297 pods/sec
+- After: ~711 pods/sec (~2.4x improvement)
+
+This benchmark is part of CI and will detect regressions. Once
+this feature is merged, the threshold for expected throughput
+needs to be revised to match the improvement.
+
+### Graduation Criteria
+
+#### Beta
+
+- [x] Feature gate `SchedulerPreQueueingHints` (default=true)
+- [x] `PreQueueingHintFn` interface added to scheduling framework
+- [x] DRA plugin implements `PreQueueingHintFn` using pod informer
+  index
+- [x] Deallocation handling (signal "evaluate all pods")
+- [x] Unit tests for all code paths
+- [x] E2E test for shared claim correctness
+- [x] `scheduler_perf` benchmark demonstrating improvement
+- [x] All e2e tests pass with feature enabled
+
+#### GA
+
+- [ ] Feature gate locked to true, removal planned
+- [ ] No reported correctness issues for 2 releases
+- [ ] Documentation updated
+
+### Upgrade / Downgrade Strategy
+
+- **Upgrade**: Feature defaults to enabled. No action required.
+  Scheduling behavior is unchanged; only performance improves.
+- **Downgrade**: Disabling the feature gate reverts to existing
+  O(N) behavior. No state is persisted.
+
+### Version Skew Strategy
+
+The feature is entirely within `kube-scheduler`. No version skew
+concerns with other components.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- Feature gate name: `SchedulerPreQueueingHints`
+- Components depending on the feature gate: `kube-scheduler`
+- Can be toggled by restarting the scheduler with/without the flag.
+
+###### Does enabling the feature change any default behavior?
+
+No. The feature only changes the *performance* of event processing,
+not scheduling decisions. All pods that would have been re-evaluated
+are still re-evaluated when `allPods=true` is signaled.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Restart the scheduler without the feature gate. The scheduler
+reverts to O(N) evaluation. No persistent state.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+The scheduler resumes using PreQueueingHintFn. No persistent state.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollback be performed?
+
+Set `--feature-gates=SchedulerPreQueueingHints=false` on the
+scheduler and restart.
+
+###### What specific metrics should inform a rollback?
+
+- `scheduler_pod_scheduled_after_flush_total` increasing: indicates
+  pods are being rescued by the periodic flush timeout rather than
+  by event-driven requeue. This suggests PreQueueingHintFn is
+  incorrectly signaling "no pods affected" for events that should
+  trigger re-evaluation.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Yes. Toggling the feature gate between scheduler restarts was tested
+in scale environments with no issues.
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use?
+
+- Scheduler logs at V(5): "PreQueueingHint narrowed pod set"
+- Feature gate status visible in
+  `kubernetes_feature_enabled{name="SchedulerPreQueueingHints"}`
+
+###### How can someone using this feature know that it is working for their instance?
+
+- For DRA workloads using ResourceClaimTemplates: scheduling
+  throughput during burst pod creation improves compared to the
+  feature being disabled.
+- `scheduler_pod_scheduled_after_flush_total` remains at zero or
+  near-zero (pods are being requeued by events, not by flush
+  timeout).
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+- Scheduling latency for DRA pods should not increase.
+- No pods should remain stuck in Pending due to missed requeue
+  events.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- Type: Metric
+  - Metric name: `scheduler_schedule_attempts_total`
+  - Components exposing the metric: `kube-scheduler`
+  - Interpretation: an unexpected increase in failed attempts
+    may indicate pods being requeued prematurely.
+- Type: Metric
+  - Metric name: `scheduler_pod_scheduled_after_flush_total`
+  - Components exposing the metric: `kube-scheduler`
+  - Interpretation: should remain at zero. Non-zero indicates
+    pods were missed by event-driven requeue and rescued by
+    the periodic flush.
+- Type: Metric
+  - Metric name: `scheduler_pre_queueing_hint_evaluations_total`
+  - Components exposing the metric: `kube-scheduler`
+  - Interpretation: a high ratio of `narrowed` vs `all_pods`
+    indicates the optimization is effective. If `all_pods`
+    dominates, most events are falling back to the full scan.
+
+###### Are there any missing metrics that would be useful to have in this context?
+
+A metric `scheduler_pre_queueing_hint_evaluations_total` with
+`plugin` and `result` labels (where result is `all_pods` or
+`narrowed`) will be added to track PreQueueingHint invocations
+and their outcomes.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+No. It is internal to kube-scheduler.
+
+###### Does it have a known list of any hard or soft dependencies on other Kubernetes features?
+
+The DRA plugin's PreQueueingHintFn requires DRA to be active.
+DRA is GA.
+
+### Scalability
+
+###### Will enabling / using this feature result in any new API calls?
+
+No.
+
+###### Will enabling / using this feature result in introducing new API types?
+
+No.
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+No.
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+No. It reduces scheduling latency.
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any component?
+
+The pod informer index adds a small memory overhead proportional to
+the number of pods with ResourceClaims. This is negligible compared
+to the existing pod informer memory usage.
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No.
+
+### Troubleshooting
+
+###### How does the feature react if the API server and/or etcd is unavailable?
+
+No impact. The feature is internal to the scheduler's in-memory
+queue processing.
+
+###### What are other known failure modes?
+
+| Failure Mode | Detection | Mitigation |
+|---|---|---|
+| PreQueueingHintFn incorrectly signals no pods | `scheduler_pod_scheduled_after_flush_total` increases | Disable feature gate |
+| Pod informer index stale | No impact: empty index means no pod references the claim yet; correct behavior | Self-healing via pod status update event |
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+1. Check `scheduler_pod_scheduled_after_flush_total` — if elevated,
+   the PreQueueingHint is missing pods that should be re-evaluated.
+2. Check scheduler logs at V(5) for "PreQueueingHint narrowed pod
+   set" messages to verify the feature is active.
+3. Disable the feature gate and restart the scheduler.
+
+## Alternatives
+
+### PodGroup (GenericWorkload) Interaction
+
+When the `GenericWorkload` feature is enabled, the scheduling queue
+stores `QueuedPodGroupInfo` entities rather than individual pods.
+To resolve which entity a pod belongs to, the framework will use
+the pod's `schedulingGroup` field via the pod informer to look up
+the owning PodGroup entity in O(1).
+
+This will be supported in 1.37. The framework will use the pod's
+`schedulingGroup` field to look up the owning PodGroup entity
+and evaluate it accordingly.
+
+### Other Alternatives
+
+1. **OwnerReference heuristic**: Use the ResourceClaim's
+   OwnerReference to identify the owning pod. Simpler but incorrect
+   when multiple pods reference the same generated claim. Rejected
+   in favor of the pod informer index.
+
+2. **Optimize QueueingHintFn directly**: Make per-pod evaluation
+   faster. Rejected because the O(N) iteration is the bottleneck,
+   not per-pod evaluation cost.
+
+3. **No feature gate**: Ship as always-on. Accepted for beta since
+   the pod informer index approach is correct by construction and
+   verified by E2E tests.
+
+## Implementation History
+
+- 2026-05: Initial KEP for beta proposed
+  ([#138916](https://github.com/kubernetes/kubernetes/pull/138916))
--- a/keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
+++ b/keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
@@ -1,0 +1,37 @@
+title: PreQueueingHint Extension Point for Scheduler Event Processing
+kep-number: 6132
+authors:
+  - "@geetasg"
+owning-sig: sig-scheduling
+participating-sigs:
+  - sig-node
+status: implementable
+creation-date: 2026-05-31
+reviewers:
+  - "@sanposhiho"
+  - "@macsko"
+  - "@pohly"
+approvers:
+  - "@sanposhiho"
+
+see-also:
+  - "/keps/sig-scheduling/624-scheduling-framework"
+
+stage: beta
+
+latest-milestone: "v1.37"
+
+milestone:
+  beta: "v1.37"
+  stable: "v1.39"
+
+feature-gates:
+  - name: SchedulerPreQueueingHints
+    components:
+      - kube-scheduler
+disable-supported: true
+
+metrics:
+  - scheduler_schedule_attempts_total
+  - scheduler_pod_scheduled_after_flush_total
+  - scheduler_pre_queueing_hint_evaluations_total
--- a/keps/prod-readiness/sig-scheduling/6132.yaml
+++ b/keps/prod-readiness/sig-scheduling/6132.yaml
@@ -1,0 +1,3 @@
+kep-number: 6132
+beta:
+  approver: "@johnbelamaric"


keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG Meetings, relevant PRs/issues, release notes
+
+## Summary
+
+This KEP introduces a `PreQueueingHintFn` extension point to the Kubernetes scheduling framework. This function is called once per cluster event before iterating unschedulable pods, allowing plugins to return a targeted set of pod keys to evaluate instead of scanning all unschedulable pods. This reduces the requeue path from O(N) to O(1) for events that affect only a single pod.
```
pohly: Before we review: please line-wrap as mentioned in 
https://github.com/kubernetes/enhancements/blob/3cf8faaf57c6bef31d5878d07afe93f8554e71e3/keps/NNNN-kep-template/README.md?plain=1#L5-L7

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+The initial implementation provides a `PreQueueingHintFn` in the DRA (Dynamic Resource Allocation) plugin for ResourceClaim events, along with an async event queue that decouples ResourceClaim event handling from the informer thread.
+
+## Motivation
```
pohly, line 73: You are saying "At scale" here and the dive into a discussion of scheduler implementation details. For someone unfamiliar with this space it's not clear whether they are affected. Can you describe the scenario that you need to improve? Along which dimension do you need to scale up?

I think "at scale" here refers to the number of pods which accumulate in the unschedulable queue, not number of nodes, right? For a workload with a steady rate of new pods per second over a long period of time, the scheduler should be able to keep up. I believe you want to improve the handling of a "thundering herd" of pods which all get created more or less at the same time.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+As a scheduler plugin author, I want to implement a `PreQueueingHintFn` that tells the framework which specific pods are affected by a cluster event, so the framework doesn't need to iterate all unschedulable pods.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|-----------|
+| PreQueueingHintFn returns too narrow a set, missing pods that should be re-evaluated | Return nil to fall back to evaluating all pods. Deallocation events always return nil. |
+| Per-pod claim heuristic (OwnerReference) is wrong for shared claims | Only apply optimization when OwnerReference Kind=Pod; return nil otherwise |
+| Async event queue introduces ordering issues | Use workqueue with deduplication; ordering within a single event type is preserved |
+| Feature gate interaction with other scheduler features | Comprehensive e2e testing with all alpha/beta features enabled |
```
pohly: There's one risk missing: what if a generated claim gets referenced by a second pod? It is not a common scenario, but the API allows it. We can't break such a usage and cause the second pod to not get scheduled.

Here's how an E2E test for this could be written:
- Don't deploy the DRA test driver yet.
- Create a Pod which references a ResourceClaimTemplate.
- Wait for ResourceClaim to get created.
- Create a second Pod which references the ResourceClaim.
- Wait for the scheduler to attempt scheduling of both pods once.
- Deploy the DRA driver.

One mitigation would be the periodic sweep of the unscheduleable queue. But that causes an unpredictable, potentially long delay for the second pod. Also, isn't that sweep meant to go away, if it hasn't already? cc @sanposhiho 

Another, more expensive mitigation would be to maintain an index in the pod informer to look up pods which reference a ResourceClaim. This completely avoids using the Owner reference. It may also have another advantage: the flow in the controller is "create ResourceClaim with Owner" + "update Pod status to link to the ResourceClaim". Scheduling can happen when the second part is complete. With the proposed PreQueueingHintFn based on Owner, the pod might get checked before the scheduler has the Pod status in its informer cache, causing some unnecessary work. When using the index, the pod won't show up yet and the unnecessary check is avoided. The pod status update then triggers the scheduling.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+```go
+type ClusterEventWithHint struct {
+    Event             ClusterEvent
+    QueueingHintFn    QueueingHintFn
+    PreQueueingHintFn PreQueueingHintFn
+}
+
+// PreQueueingHintFn is called once per event before iterating unschedulable pods.
+// It returns a set of pod keys (in "name_namespace" format) that should be evaluated.
```
pohly: Underscore is valid in both name and namespace, isn't it? Then pod "a" in namespace "b_c" and pod "a_b" in namespace "c" both have "a_b_c" as concatenated string.

But isn't the implementation using a slash as separator? Then the docs here would need to be updated.

Let's use NamespacedName to avoid this ambiguity? Then no parsing of the string is needed to extract the parts and plugin and framework don't need to agree on how to encode both strings. I think that's a cleaner API.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+// Returning nil means all pods should be evaluated (existing behavior).
```
pohly: That's a very subtle difference and means that for the not uncommon case of "evaluate no pods" memory must be allocated to return an empty set.

Maybe add an explicit boolean for "all"? `true, nil` means "all", `false` means "check the set" (which can be nil, empty or non-empty).

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+- Existing scheduler integration tests pass with feature gate enabled
+- DRA integration tests (`pull-kubernetes-dra-integration`) pass
+
+#### E2E Tests
+
+- `pull-kubernetes-e2e-kind-alpha-beta-features`: Full e2e with all alpha/beta gates enabled
+- `pull-kubernetes-e2e-gce-cos-alpha-features`: GCE e2e with alpha features
+
+#### Scale Tests
```
pohly: Replace "Scale Tests" with "Performance Tests"?

A one-time check with KWOK is useful, but not sufficient. We also have to ensure that there are not future regressions. Scale tests are expensive, in this case the right tool is test/integration/scheduler_perf.

Do you see any improvements there? See the README.md there for instructions on how to do a before/after comparison.

If not, then we should add a scenario that demonstrates the poor performance and then benefits from this enhancement.

pohly: I was able to reproduce with scheduler_perf:
```
make test-integration WHAT=./test/integration/scheduler_perf/dra KUBE_CACHE_MUTATION_DETECTOR=false KUBE_TIMEOUT=-timeout=1h ETCD_LOGLEVEL=warn KUBE_TEST_VMODULE="''" FULL_LOG=y ARTIFACTS=/tmp SHORT=-short=false KUBE_TEST_ARGS='-run=^$ -benchtime=1x -bench=^BenchmarkPerfScheduling$/^SchedulingWithResourceClaimTemplate$/5000pods_500nodes -perf-scheduling-label-filter= -cpuprofile /tmp/cpu-profile.out' >before.log 2>&1
```

Repeat and write into `after.out` after flipping the feature gate:
```patch
diff --git a/pkg/features/kube_features.go b/pkg/features/kube_features.go
index cf280242e3a..56dca771499 100644
--- a/pkg/features/kube_features.go
+++ b/pkg/features/kube_features.go
@@ -1871,7 +1871,7 @@ var defaultVersionedKubernetesFeatureGates = map[featuregate.Feature]featuregate
        },
 
        SchedulerPreQueueingHints: {
-               {Version: version.MustParse("1.37"), Default: false, PreRelease: featuregate.Alpha},
+               {Version: version.MustParse("1.37"), Default: true, PreRelease: featuregate.Beta},
        },
 
        SeparateTaintEvictionController: {
```

```console
$ benchstat before.log after.log

SchedulingThroughput/Average

PerfScheduling/SchedulingWithResourceClaimTemplate/5000pods_500nodes-24
                    296.6 ± ∞ ¹                    710.7 ± ∞ ¹  ~ (p=1.000 n=1) ²
```

I also see the change in CPU consumption in the cpu profile.

```
go tool pprof -http=pohly-desktop:8080 -no_browser  /tmp/cpu-profile.out
```

So definitely a worthwhile enhancement.

keps/sig-scheduling/6132-prequeueing-hints/README.md
pohly: > Here's how an E2E test for this could be written: [...]

Below is a patch adding such a test. It passes with https://github.com/kubernetes/kubernetes/pull/138916 when the feature is off and fails with some pods not getting scheduled when on.

```patch
diff --git a/test/e2e/dra/dra.go b/test/e2e/dra/dra.go
index 8ec1458d5ae..9a15c269a30 100644
--- a/test/e2e/dra/dra.go
+++ b/test/e2e/dra/dra.go
@@ -46,6 +46,7 @@ import (
 	applyv1 "k8s.io/client-go/applyconfigurations/core/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/dynamic-resource-allocation/resourceclaim"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/kubelet/events"
@@ -2209,6 +2210,58 @@ var _ = framework.SIGDescribe("node")(framework.WithLabel("DRA"), func() {
 	// as long as the test names the still remain unique overall.
 
 	framework.Context("control plane", func() { singleNodeTests(false) })
+	framework.Context("control plane", func() {
+		ginkgo.It("schedules pods sharing one generated ResourceClaim after driver starts", func(ctx context.Context) {
+			tCtx := f.TContext(ctx)
+			nodes := drautils.NewNodesNow(tCtx, 1, 1)
+			driver := drautils.NewDriverInstance(tCtx)
+			driver.WithKubelet = false
+			b := drautils.NewBuilderNow(tCtx, driver)
+
+			pod, template := b.PodInline()
+			b.Create(tCtx, template, pod)
+			allPods := []*v1.Pod{pod}
+
+			// Cannot run pod, no devices.
+			framework.ExpectNoError(e2epod.WaitForPodNameUnschedulableInNamespace(ctx, f.ClientSet, pod.Name, pod.Namespace))
+
+			// Determine which claim was generated for the pod.
+			pod, err := tCtx.Client().CoreV1().Pods(pod.Namespace).Get(tCtx, pod.Name, metav1.GetOptions{})
+			tCtx.ExpectNoError(err, "get pod")
+			claimName, _, err := resourceclaim.Name(pod, &pod.Spec.ResourceClaims[0])
+			tCtx.ExpectNoError(err, "get ResourceClaimName from pod")
+			tCtx.Expect(claimName).NotTo(gomega.BeNil())
+
+			numExtraPods := 10
+			var extraPods []*v1.Pod
+			for range numExtraPods {
+				pod := b.PodExternal(*claimName)
+				extraPods = append(extraPods, pod)
+				allPods = append(allPods, pod)
+				b.Create(tCtx, pod)
+			}
+			for i, pod := range extraPods {
+				framework.ExpectNoError(e2epod.WaitForPodNameUnschedulableInNamespace(ctx, f.ClientSet, pod.Name, pod.Namespace), "extra pod #%d", i)
+			}
+
+			// Set up driver, which makes devices available.
+			driver.Run(tCtx, framework.TestContext.KubeletRootDir, nodes, drautils.DriverResourcesNow(nodes, 1))
+
+			// Now all should run.
+			for _, pod := range allPods {
+				b.TestPod(tCtx, pod)
+			}
+
+			// We need to clean up explicitly because the normal
+			// cleanup doesn't work (driver shuts down first).
+			for _, pod := range allPods {
+				framework.ExpectNoError(f.ClientSet.CoreV1().Pods(pod.Namespace).Delete(ctx, pod.Name, metav1.DeleteOptions{}))
+			}
+			for _, pod := range allPods {
+				framework.ExpectNoError(e2epod.WaitForPodNotFoundInNamespace(ctx, f.ClientSet, pod.Name, pod.Namespace, f.Timeouts.PodDelete))
+			}
+		})
+	})
 	framework.Context("kubelet", feature.DynamicResourceAllocation, "on single node", func() { singleNodeTests(true) })
 	framework.Context("kubelet", feature.DynamicResourceAllocation, "on single node with multiple claims allocation", singleNodeMultipleClaimsTests)
 ```

File: keps/sig-scheduling/6132-prequeueing-hints/README.md (hunk 5 above)
Comment (pohly) [reply keps/sig-scheduling/6132-prequeueing-hints/README.md]: A set is unnecessarily complex. A slice is simpler and sufficient: the contract is that the PreQueueingHintFn doesn't return duplicates. We don't need to enforce that via the function prototype.

File: keps/sig-scheduling/6132-prequeueing-hints/README.md (hunk 3 above)
Comment (pohly) [reply keps/sig-scheduling/6132-prequeueing-hints/README.md]: > Another, more expensive mitigation would be to maintain an index in the pod informer to look up pods which reference a ResourceClaim. 

Here's a patch against your code PR which implements this idea:
```patch
diff --git a/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go b/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go
index fb909dbdb06..735ecca46c9 100644
--- a/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go
+++ b/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go
@@ -39,6 +39,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/component-helpers/scheduling/corev1/nodeaffinity"
 	"k8s.io/dynamic-resource-allocation/cel"
@@ -170,8 +171,15 @@ type DynamicResources struct {
 	clientset      kubernetes.Interface
 	celCache       *cel.Cache
 	draManager     fwk.SharedDRAManager
+	podIndexer     cache.Indexer
 }
 
+const (
+	// podResourceClaimIndex is the lookup name for the index function which indexes pod by the ResourceClaims that they reference.
+	// The key for looking up pods is "<namespace>/<name>" of a ResourceClaim.
+	podResourceClaimIndex = "podResourceClaim"
+)
+
 var (
 	ErrDeviceBindingTimeout = errors.New("device binding timeout")
 	ErrDeviceBindingFailed  = errors.New("device binding failed")
@@ -192,6 +200,11 @@ func New(ctx context.Context, plArgs runtime.Object, fh fwk.Handle, fts feature.
 		return nil, err
 	}
 
+	podInformer := fh.SharedInformerFactory().Core().V1().Pods()
+	if err := podInformer.Informer().AddIndexers(cache.Indexers{podResourceClaimIndex: podResourceClaimIndexFunc}); err != nil {
+		return nil, fmt.Errorf("add pod indexer: %w", err)
+	}
+
 	pl := &DynamicResources{
 		enabled:       true,
 		fts:           fts,
@@ -210,6 +223,7 @@ func New(ctx context.Context, plArgs runtime.Object, fh fwk.Handle, fts feature.
 			EnableListTypeAttributes: fts.EnableDRAListTypeAttributes,
 		}),
 		draManager: fh.SharedDRAManager(),
+		podIndexer: podInformer.Informer().GetIndexer(),
 	}
 
 	return pl, nil
@@ -252,7 +266,7 @@ func (pl *DynamicResources) EventsToRegister(_ context.Context) ([]fwk.ClusterEv
 		// A new or updated node may make pods schedulable.
 		{Event: fwk.ClusterEvent{Resource: fwk.Node, ActionType: fwk.Add | fwk.UpdateNodeLabel | fwk.UpdateNodeAllocatable}},
 		// Allocation is tracked in ResourceClaims, so any changes may make the pods schedulable.
-		{Event: fwk.ClusterEvent{Resource: fwk.ResourceClaim, ActionType: fwk.Add | fwk.Update}, QueueingHintFn: pl.isSchedulableAfterClaimChange, PreQueueingHintFn: claimPreQueueingHint},
+		{Event: fwk.ClusterEvent{Resource: fwk.ResourceClaim, ActionType: fwk.Add | fwk.Update}, QueueingHintFn: pl.isSchedulableAfterClaimChange, PreQueueingHintFn: pl.preQueueingHint},
 		// Adding the ResourceClaim name to the pod status makes pods waiting for their ResourceClaim schedulable.
 		{Event: fwk.ClusterEvent{Resource: fwk.TargetPod, ActionType: fwk.UpdatePodGeneratedResourceClaim}, QueueingHintFn: pl.isSchedulableAfterTargetPodUpdate},
 		// A pod might be waiting for a class to get created or modified.
@@ -278,17 +292,35 @@ func (pl *DynamicResources) PreEnqueue(ctx context.Context, pod *v1.Pod) (status
 	return nil
 }
 
-// claimPreQueueingHint returns the set of pod keys affected by a ResourceClaim event.
-// For per-pod claims (ResourceClaimTemplate-based, with a Pod OwnerReference),
-// only that pod needs evaluation. Uses util.GetPodFullName to construct the key
-// matching the unschedulablePods podInfoMap format.
-func claimPreQueueingHint(logger klog.Logger, oldObj, newObj interface{}) sets.Set[string] {
-	// Deallocation frees resources that may make other pods schedulable.
+// podResourceClaimIndexFunc is an index function that returns ResourceClaim keys (=
+// namespace/name) for any ResourceClaim referenced by a pod.
+func podResourceClaimIndexFunc(obj interface{}) ([]string, error) {
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		return nil, nil
+	}
+	keySet := sets.New[string]()
+	for _, podClaim := range pod.Spec.ResourceClaims {
+		claimName, _, err := resourceclaim.Name(pod, &podClaim)
+		if err != nil || claimName == nil {
+			// Index functions are not supposed to fail, the caller will panic.
+			// For both error reasons (claim not created yet, unknown API)
+			// we simply don't index.
+			continue
+		}
+		keySet.Insert(fmt.Sprintf("%s/%s", pod.Namespace, *claimName))
+	}
+	return sets.List(keySet), nil
+}
+
+// preQueueingHint returns the set of pod keys affected by a ResourceClaim event.
+func (pl *DynamicResources) preQueueingHint(logger klog.Logger, oldObj, newObj interface{}) sets.Set[string] {
 	if oldObj != nil && newObj != nil {
 		if oldClaim, ok := oldObj.(*resourceapi.ResourceClaim); ok {
 			if newClaim, ok := newObj.(*resourceapi.ResourceClaim); ok {
+				// Deallocation frees resources that may make other pods schedulable.
 				if oldClaim.Status.Allocation != nil && newClaim.Status.Allocation == nil {
-					return nil
+					return nil // all pods
 				}
 			}
 		}
@@ -300,14 +332,20 @@ func claimPreQueueingHint(logger klog.Logger, oldObj, newObj interface{}) sets.S
 	}
 	claim, ok := obj.(*resourceapi.ResourceClaim)
 	if !ok {
+		return nil // all pods
+	}
+	objs, err := pl.podIndexer.ByIndex(podResourceClaimIndex, fmt.Sprintf("%s/%s", claim.Namespace, claim.Name))
+	if err != nil {
+		// Shouldn't happen. Check all pods.
 		return nil
 	}
-	for _, ref := range claim.OwnerReferences {
-		if ref.Kind == "Pod" && ref.APIVersion == v1.SchemeGroupVersion.String() {
-			return sets.New(schedutil.GetPodFullNameFromNamespacedName(types.NamespacedName{Name: ref.Name, Namespace: claim.Namespace}))
+	podKeys := sets.New[string]()
+	for _, obj := range objs {
+		if pod, ok := obj.(*v1.Pod); ok {
+			podKeys.Insert(schedutil.GetPodFullNameFromNamespacedName(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}))
 		}
 	}
-	return nil // shared claim — evaluate all pods
+	return podKeys // Non-nil, so only explicitly listed pods are checked.
 }
 
 // isSchedulableAfterClaimChange is invoked for add and update claim events reported by
```

The E2E test now passes with the feature enabled. Performance is about the same. I also don't see the indexer in the CPU profile.

If we do it like this, then I have no concerns about directly going to beta with on-by-default.

johnbelamaric (commented):
@geetasg please add a PRR approver file. That needs to be done today.

geetasg:
Added PRR and updated the design to use informer. cc @johnbelamaric @pohly

keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
```diff
@@ -0,0 +1,39 @@
...
+    components:
+      - kube-scheduler
+disable-supported: true
+
+metrics:
+  - scheduler_scheduling_attempt_duration_seconds
+  - scheduler_pod_scheduled_after_flush_total
+
+prr-approvers:
+  - "@johnbelamaric"
```
macsko: I think the `kep.yaml` shouldn't have the `prr-approvers` list? They are defined in the `prod-readiness` yaml

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+        if pod, ok := obj.(*v1.Pod); ok {
+            pods = append(pods, types.NamespacedName{
+                Name:      pod.Name,
+                Namespace: pod.Namespace,
+            })
+        }
+    }
+    return pods
+}
+```
```
macsko: We rather don't expect code blocks in the KEPs. It's likely that they will be outdated soon. It's better to describe the core behavior in text

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+| PreQueueingHintFn returns too narrow a set | Return nil to fall back to evaluating all pods. Deallocation events always return nil. |
+| Pod informer index is stale (pod created but not yet indexed) | Pod won't appear in index; claim event falls through to nil (evaluate all). Pod status update triggers separate scheduling. |
+| A generated claim referenced by a second pod | Pod informer index looks up all pods referencing the claim, not just the owner. Correctness verified by E2E test. |
+| Async event queue introduces ordering issues | Use workqueue with deduplication; ordering within a single event type is preserved. |
+
+## Design Details
+
+### PreQueueingHintFn Interface
+
+Added to `ClusterEventWithHint` in the scheduling framework:
```
macsko: You should describe more in the KEP how this function will be called and where.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+### Async Event Queue
+
+ResourceClaim Add/Update events are decoupled from the informer
+thread using a workqueue:
+
+- The informer handler enqueues a lightweight item (claim key).
+- A dedicated worker goroutine drains the queue and calls
+  `MoveAllToActiveOrBackoffQueue`.
+- This prevents the informer from blocking on scheduling queue
+  lock acquisition.
```
macsko: I don't follow what you mean by this section. Why do we need dedicated calls to the `MoveAllToActiveOrBackoffQueue`? Could you elaborate more?

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+
+###### How can an operator determine if the feature is in use?
+
+- Scheduler logs at V(5): "PreQueueingHint narrowed pod set"
+- Feature gate status visible in
+  `kubernetes_feature_enabled{name="SchedulerPreQueueingHints"}`
+
+###### How can someone using this feature know that it is working for their instance?
+
+- Scheduling throughput improves for DRA workloads.
```
macsko: All DRA workloads or specific workloads?

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+- `scheduler_scheduling_attempt_duration_seconds` decreases.
+- `scheduler_pod_scheduled_after_flush_total` remains low.
```
macsko: It's too generically described to be a useful indicator

macsko (commented):
Left some initial comments

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+iterating all pods.
+
+### Goals
+
+- Provide a `PreQueueingHintFn` extension point that plugins can
+  implement to narrow the set of pods evaluated on cluster events.
+- Reduce ResourceClaim event processing from O(N) to O(1) for
+  claims referenced by a known set of pods.
+- Decouple ResourceClaim event handling from the informer thread
+  to prevent blocking.
```
pohly: Note that I haven't looked at this at all - I'm leaning on SIG Scheduling reviewers to tell us whether this can lead to correctness problems.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+- Changing the scheduling algorithm or pod priority ordering.
+
+## Proposal
+
+### User Stories
+
+**Story 1: Large-scale DRA workloads**
+
+As a cluster operator running thousands of pods (e.g., 10K) with
+DRA ResourceClaimTemplates, I want the scheduler to maintain high
```
pohly:
```suggestion
As a cluster operator rapidly creating thousands of pods (e.g., 10K) with
DRA ResourceClaimTemplates, I want the scheduler to maintain high
```

The "thundering herd" aspect is important, let's call it out.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+`PreQueueingHintFn` that tells the framework which specific pods
+are affected by a cluster event, so the framework doesn't need to
+iterate all unschedulable pods.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|-----------|
+| PreQueueingHintFn returns too narrow a set | Return nil to fall back to evaluating all pods. Deallocation events always return nil. |
+| Pod informer index is stale (pod created but not yet indexed) | Pod won't appear in index; claim event falls through to nil (evaluate all). Pod status update triggers separate scheduling. |
```
pohly: This is not how I implemented it, I believe. When no pod is found, `PreQueueingHintFn` should result in "don't check any pods".

Also note my earlier comment about "nil" vs. "empty" - we should use a separate boolean.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+    Event             ClusterEvent
+    QueueingHintFn    QueueingHintFn
+    PreQueueingHintFn PreQueueingHintFn
+}
+
+// PreQueueingHintFn is called once per event before iterating
+// unschedulable pods. It returns a list of pods that should be
+// evaluated. Returning nil means all pods should be evaluated
+// (existing behavior). Returning an empty slice means no pods
+// need evaluation.
```
pohly: Let's add the explicit boolean for "all" that I mentioned in https://github.com/kubernetes/enhancements/pull/6133#discussion_r3339284030.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+### DRA Plugin Implementation
+
+The DRA plugin maintains a pod informer index that maps
+`namespace/name` of ResourceClaims to the pods referencing them.
+This is registered at plugin initialization:
+
+```go
+podInformer.Informer().AddIndexers(cache.Indexers{
+    podResourceClaimIndex: podResourceClaimIndexFunc,
+})
```
pohly: You can remove this code snippet. It doesn't tell the reader of the KEP anything. Let's focus on concepts in the KEP, not specific Go code.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+The `PreQueueingHintFn` implementation:
+
+1. **Deallocation detection**: If allocation transitions from
+   non-nil to nil, return nil (evaluate all pods, since freed
+   capacity may help any waiting pod).
+2. **Index lookup**: Look up all pods referencing the claim via
+   the informer index. Return their NamespacedNames.
+3. **Index error or empty result**: Return nil (evaluate all pods).
```
pohly: "empty result" is fine, it means that no pods can become schedulable because none reference the claim.

keps/sig-scheduling/6132-prequeueing-hints/README.md
pohly: Fully agreed.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,482 @@
...
+
+#### Performance Tests
+
+Measured using `test/integration/scheduler_perf` with the
+`SchedulingWithResourceClaimTemplate` benchmark:
+
+- Before: ~297 pods/sec
+- After: ~711 pods/sec (~2.4x improvement)
+
+This benchmark is part of CI and will detect regressions.
```
pohly:
```suggestion
This benchmark is part of CI and will detect regressions. Once this feature
is merged, the threshold for expected
throughput needs to be revised to
match the improvement.
```

pohly (changes requested):

keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
geetasg: updated. PRR approver is in 6132.yaml now.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated.. removed code blocks.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated - Added - the PreQueueingHintFn section now explains it is invoked once per matching cluster event inside moveAllToActiveOrBackoffQueue, before the loop that iterates unschedulable pods.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Updated. Currently, ResourceClaim informer handlers call MoveAllToActiveOrBackoffQueue synchronously,
   acquiring the scheduling queue lock on the informer goroutine. Under high claim event volume this blocks the informer from delivering other events. The async queue decouples the two with a workqueue in between.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Narrowed to "DRA workloads using ResourceClaimTemplates"

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Updated to be bit more specific: scheduler_pod_scheduled_after_flush_total should remain at zero. Non-zero indicates pods were missed by event-driven requeue and rescued by the periodic flush.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Updated the user story description

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Updated. Return type is now (allPods bool, pods []types.NamespacedName). Empty list with allPods=false means no pods need evaluation. Stale/empty index is correct behavior, not a fallback to all.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: removed code snippets

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,460 @@
...
+- Type: Metric
+  - Metric name: `scheduler_pod_scheduled_after_flush_total`
+  - Components exposing the metric: `kube-scheduler`
+  - Interpretation: should remain at zero. Non-zero indicates
+    pods were missed by event-driven requeue and rescued by
+    the periodic flush.
+
+###### Are there any missing metrics that would be useful to have in this context?
+
+A metric for PreQueueingHint invocations and their result
```
aws-ruhip: We can call out the metric name to be added here, something like
`scheduler_pre_queueing_hint_evaluations_total` with `plugin` and `result` labels (result could be either `all_pods` or `narrowed`)

geetasg: updated

keps/sig-scheduling/6132-prequeueing-hints/README.md
pohly: And what's the advantage of decoupling? What's the benefit of delivering other events first? You describe the "what" here, but not the "why".

XREF: code is currently in https://github.com/kubernetes/kubernetes/pull/138916/changes#diff-5886f7d1bbfda26de989732aba1cb0f658f2835af1062bb05be1ac1d3531c836

As you have this implemented already, can you benchmark whether that change is worth it? How much speedup do you see compared to only doing the PreQueuingHint?

We have two different and unrelated optimizations:
- PreQueuingHint
- workqueue for ResourceClaim events

I think both should have separate feature gates. Then one can be turned off without affecting the other. A single KEP seems sufficient.

macsko: +1 to @pohly. It would be good to see how much improvement the workqueue gives. While PreQueueingHint looks like a solid idea, we have to be careful with managing the events.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+**Story 1: Large-scale DRA workloads**
+
+As a cluster operator rapidly creating thousands of pods (e.g., 10K)
+with DRA ResourceClaimTemplates, I want the scheduler to maintain
+high throughput without being bottlenecked by ResourceClaim event
+processing.
+
+**Story 2: Plugin authors**
+
+As a scheduler plugin author, I want to implement a
```
ania-borowiec: Nit:
I'm not sure whether this is a valid user story. I mean - it's valid, but not a user story. The user story would probably say something about using the DRA plugin without impacting the scheduling throughput severely (which boils down to story 1).

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+scheduling framework. When registered, it is invoked once per
+matching cluster event inside `moveAllToActiveOrBackoffQueue`,
+*before* the loop that iterates unschedulable pods.
+
+The function returns two values:
+- `allPods bool`: if true, all unschedulable pods should be
+  evaluated (existing behavior). The pod list is ignored.
+- `pods []types.NamespacedName`: when `allPods` is false, only
+  the listed pods are evaluated by the per-pod `QueueingHintFn`.
+  An empty list means no pods need evaluation for this event.
```
macsko: When queueing PodGroups, we put the entire QueuedPodGroupInfo object into the `unschedulableEntities` structure - we don't store the member pods directly in the queue. When receiving a list of pods from `PreQueueingHintFn`, how could we know which pod groups these pods belong to? Without such knowledge, we would need to iterate over all unschedulable entities and check whether the pod fits to any.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+As a scheduler plugin author, I want to implement a
+`PreQueueingHintFn` that tells the framework which specific pods
+are affected by a cluster event, so the framework doesn't need to
+iterate all unschedulable pods.
+
+### Risks and Mitigations
+
+| Risk | Mitigation |
+|------|-----------|
+| PreQueueingHintFn incorrectly signals "no pods affected" | Deallocation events and index errors always signal "evaluate all". |
```
macsko: In case the `PreQueueingHintFn` incorrectly signals that the no pods (or invalid pods due to a bug) are affected by the event, we also have a periodical flush of unschedulable pods, which would partially mitigate the issue.

keps/sig-scheduling/6132-prequeueing-hints/README.md
macsko: What is the meaning of nil list?

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+  fallback, empty index (no pods affected), error handling.
+- `TestPriorityQueue_PreQueueingHint`: Tests gate enabled/disabled,
+  hint narrows to specific pods, hint signals all, multiple hints
+  union.
+
+#### Integration Tests
+
+- Existing scheduler integration tests pass with feature gate
+  enabled.
+- DRA integration tests (`pull-kubernetes-dra-integration`) pass.
```
macsko, line 200: It would be good to also add an integration test to verify the PreQueueingHint behavior

johnbelamaric (commented):
LGTM from PRR perspective. Awaiting SIG approval

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+
+Multiple plugins may register a `PreQueueingHintFn` for the same
+event. The framework unions their pod lists. If any plugin signals
+"all pods", the framework evaluates all pods.
+
+### DRA Plugin Implementation
+
+The DRA plugin maintains a pod informer index that maps each
+ResourceClaim (by namespace/name) to the pods that reference it
+in their `spec.resourceClaims`. This index is registered at plugin
```
ania-borowiec, line 156: IIUC it's possible that the index will have to be filled on scheduler's startup, to reflect the current state of the cluster. What is the cost of initializing the index?
Would the initialization be handled asynchronously? Would the plugin know when the index is not yet up to date (so that it can return "evaluate all", as stated in line 130)?

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+- Scheduling latency for DRA pods should not increase.
+- No pods should remain stuck in Pending due to missed requeue
+  events.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- Type: Metric
+  - Metric name: `scheduler_scheduling_attempt_duration_seconds`
```
ania-borowiec: I don't think this metric is very useful in this case - this is the length of a scheduling cycle, and processing QueueingHints happens outside of the scheduling cycle. Plus this metric doesn't allow you to filter by plugins.

How about looking at the number of failed scheduling attempts that were rejected by the DRA plugin: `scheduler_unschedulable_pods`? If that number increases significantly compared to feature disabled, that could indicate too many pods being moved to activeQ?
Alternatively, maybe the related metric `scheduler_schedule_attempts_total`? This one doesn't allow filtering by plugin, though

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,461 @@
...
+  - Interpretation: should remain at zero. Non-zero indicates
+    pods were missed by event-driven requeue and rescued by
+    the periodic flush.
+
+###### Are there any missing metrics that would be useful to have in this context?
+
+A metric `scheduler_pre_queueing_hint_evaluations_total` with
+`plugin` and `result` labels (where result is `all_pods` or
+`narrowed`) will be added to track PreQueueingHint invocations
+and their outcomes.
```
ania-borowiec, line 353: How would you recommend interpreting the outcomes?

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Benchmarked independently. No throughput benefit from the async queue. Profiling shows ~18% reduction in CPU available for scheduling due to worker goroutine overhead.
   
    Pros:
    - Cleaner separation of concerns between event delivery and event processing
   
    Cons:
    - ~18% reduction in effective scheduler CPU utilization
   
Open to dropping the async queue from this KEP based on how this thread converges.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: ack. This is open discussion - tracked by above thread - https://github.com/kubernetes/enhancements/pull/6133#discussion_r3408664449

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: ack. evaluating updates required for PodGroups - will post back on this thread once KEP is updated.

geetasg: When allPods is false, a nil slice is equivalent to an empty slice — no pods need evaluation.
Clarified in the KEP.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated the risks table

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Added to the test plan.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: Fair point — removed Story 2 and consolidated into Story 1

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: The index is built by the shared pod informer during its initial list+watch sync — same as any
  other informer index. The scheduler already waits for informer caches to sync before scheduling
  begins, so the index is populated before PreQueueingHintFn is ever called. No separate
  initialization needed.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: replaced with scheduler_schedule_attempts_total and scheduler_pre_queueing_hint_evaluations_total. Updated in the KEP.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: A high ratio of narrowed vs all_pods indicates the optimization is effective. If all_pods
  dominates, most events are falling back to the full scan (e.g., deallocation events or index
  errors). Added interpretation to the SLI section.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated interface to return  []types.NamespacedName

keps/sig-scheduling/6132-prequeueing-hints/README.md
pohly: Yes, let's remove it.

keps/sig-scheduling/6132-prequeueing-hints/README.md
macsko: Probably, we could just use an informer to check the `schedulingGroup` of the pod. Worth evaluate the performance impact

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,465 @@
...
+- `TestPreQueueingHint`: Tests pod lookup via index, deallocation
+  fallback, empty index (no pods affected), error handling.
+- `TestPriorityQueue_PreQueueingHint`: Tests gate enabled/disabled,
+  hint narrows to specific pods, hint signals all, multiple hints
+  union.
+
+#### Integration Tests
+
+- Existing scheduler integration tests pass with feature gate
+  enabled.
```
ania-borowiec:
```suggestion
- Existing scheduler integration tests pass with feature gate
  enabled and disabled.
```

As we need to make sure that the newly added logic does not break the existing logic.

ania-borowiec (commented):
One last comment, otherwise LGTM

keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+feature-gates:
+  - name: SchedulerPreQueueingHints
+    components:
+      - kube-scheduler
+disable-supported: true
+
+metrics:
+  - scheduler_scheduling_attempt_duration_seconds
+  - scheduler_pod_scheduled_after_flush_total
```
macsko: Update the metrics with the current list from "What are the SLIs..."

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,465 @@
...
+- [x] Async event queue for ResourceClaim processing
+- [x] Unit tests for all code paths
+- [x] E2E test for shared claim correctness
+- [x] `scheduler_perf` benchmark demonstrating improvement
+- [x] All e2e tests pass with feature enabled
+
+#### GA
+
+- [ ] Feature gate locked to true, removal planned
+- [ ] No reported correctness issues for 2 releases
```
macsko, line 241: You are mentioning 2 releases here, however in the kep.yaml you're proposing GA in v1.38. Please align the criteria (we should probably GA this in v1.39)

macsko (commented):
KEP still needs an update with "Async Event Queue" removal and PodGroup processing support

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,465 @@
...
+matching cluster event inside `moveAllToActiveOrBackoffQueue`,
+*before* the loop that iterates unschedulable pods.
+
+The function returns two values:
+- `allPods bool`: if true, all unschedulable pods should be
+  evaluated (existing behavior). The pod list is ignored.
+- `pods []types.NamespacedName`: when `allPods` is false, only
+  the listed pods are evaluated by the per-pod `QueueingHintFn`.
+  An empty or nil slice means no pods need evaluation for this
+  event.
```
sanposhiho: nit: can we define a single struct that contains both, instead of returning them separately? They together show the single result (which pods to evaluate in QHint), and each value is related to the other one (if you have allPods:true, pods should be nil and will be ignored).

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,465 @@
...
+
+If no `PreQueueingHintFn` is registered for an event, the framework
+falls back to evaluating all pods (existing behavior).
+
+Multiple plugins may register a `PreQueueingHintFn` for the same
+event. The framework unions their pod lists. If any plugin signals
+"all pods", the framework evaluates all pods.
+
```
sanposhiho, line 151: We need one step further - I believe we need to map which plugin returned which PreQueueingHint result. Like, if pluginA PreQHint returned [podA], while pluginB PreQHint returned allPods:true, we should only run pluginA QHint for podA, should not run pluginA QHint for all pods.

This is not only about the optimization - if your plugin is using pod/add event, and want to reduce the num of executions with PreQHint, you would need to get ALL pod/add QHints to implement PreQHint with the current proposal. Otherwise other QHints w/o PreQHint would be regarded as allPods:true, and your PreQHint on your plugin doesn't change anything.

geetasg: updated

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated.

keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
geetasg: updated.

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,460 @@
...
+- 2026-05-08: Initial PR with OwnerReference-based approach
+  ([#138916](https://github.com/kubernetes/kubernetes/pull/138916))
+- 2026-05-13: Added deallocation handling
+- 2026-05-27: Rebased, CI green
+- 2026-06-09: Revised to pod informer index approach per reviewer
+  feedback
+- 2026-06-10: KEP updated: removed code blocks, added explicit
+  boolean return, added PRR file
+- 2026-06-13: Removed async event queue (no benefit), added
+  PodGroup interaction design note
```
macsko: nit: one item per release is enough. You can just say that the initial KEP for beta was proposed in 2026-05

keps/sig-scheduling/6132-prequeueing-hints/kep.yaml
geetasg: Also removed async queue and added design for PodGroup handling

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,460 @@
...
+   the PreQueueingHint is missing pods that should be re-evaluated.
+2. Check scheduler logs at V(5) for "PreQueueingHint narrowed pod
+   set" messages to verify the feature is active.
+3. Disable the feature gate and restart the scheduler.
+
+## Alternatives
+
+### PodGroup (GangScheduling) Interaction
+
+When the `GangScheduling` feature is enabled, the scheduling queue
```
macsko:
```suggestion
When the `GenericWorkload` feature is enabled, the scheduling queue
```

keps/sig-scheduling/6132-prequeueing-hints/README.md
```diff
@@ -0,0 +1,460 @@
...
+stores `QueuedPodGroupInfo` entities rather than individual pods.
+To resolve which entity a pod belongs to, the framework will use
+the pod's `schedulingGroup` field via the pod informer to look up
+the owning PodGroup entity in O(1).
+
+This will be addressed in a follow-up release. For 1.37,
+PreQueueingHintFn operates on individual pod entities (the common
+case without GangScheduling enabled).
```
macsko: I think we would need to support this from day 1 since it's a beta feature and it's even hard to determine whether the pod is a pod group member without inspecting the pod group object. It should be fairly easy to implement this

macsko:
The current shape of the KEP looks good
/lgtm
/approve
/hold in case @sanposhiho has anything to add (I'll lift it before the freeze)

macsko:
@johnbelamaric This should be ready for PRR approval

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated.

keps/sig-scheduling/6132-prequeueing-hints/README.md
geetasg: updated

geetasg:
Thanks everyone for the review. All feedback has been addressed. Waiting on PRR approval and hold removal per https://github.com/kubernetes/enhancements/pull/6133#issuecomment-4722640920

PR #6134, opened 2026-06-01
KEP-5732: Update Topology-aware workload scheduling for beta in v1.37

44past4:
- One-line PR description: Update KEP-5732: Topology-aware workload scheduling for beta in v1.37

- Issue link: https://github.com/kubernetes/enhancements/issues/5732

- Other comments:
  - Apart from beta graduation itself, this KEP proposes:
    - Adding support for defining  multiple PlacementGeneratePlugins
    - Implementation of two performance improvements for TAS
    - Moving DRA support implementation to [KEP-5729: DRA: ResourceClaim Support for Workloads](https://kep.k8s.io/5729)

### DIFF
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -556,6 +556,9 @@ Topology Aware Scheduling With Basic Policy:
 
 Those tests are located at [tas_test.go‎](https://github.com/kubernetes/kubernetes/blob/eb01d62d2676cfe009382cadd5d65c2ad654998d/test/integration/scheduler/podgroup/topology_aware_scheduling/tas_test.go‎).
 
+Apart from those integration tests with promoting to beta performance tests
+for basic scenario of Topology Aware Scheduling With Gang Policy will be added.
+
 #### e2e tests
 
 - End-to-End Workload Scheduling: Submit a Workload with TopologyConstraint
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -557,7 +557,9 @@ Topology Aware Scheduling With Basic Policy:
 Those tests are located at [tas_test.go‎](https://github.com/kubernetes/kubernetes/blob/eb01d62d2676cfe009382cadd5d65c2ad654998d/test/integration/scheduler/podgroup/topology_aware_scheduling/tas_test.go‎).
 
 Apart from those integration tests with promoting to beta performance tests
-for basic scenario of Topology Aware Scheduling With Gang Policy will be added.
+for basic scenario of Topology Aware Scheduling With Gang Policy have been
+added. Those tests are located at [tas_test.go]
+(https://github.com/kubernetes/kubernetes/tree/42d9ed9c47706cb04457410e387e07889203b888/test/integration/scheduler_perf/tas).
 
 #### e2e tests
 
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -256,7 +256,7 @@ type TopologyConstraint struct {
     // All pods within the PodGroup must be colocated within the same domain instance.
     // Different PodGroups can land on different domain instances even if they derive from the same PodGroupTemplate.
     // Examples: "topology.kubernetes.io/rack"
-    Level string
+    Key string
 }
 ```
 
@@ -413,7 +413,7 @@ The algorithm proceeds in three main phases for a given PodGroup.
 **TopologyPlacementGenerator (New)** Implements `PlacementGeneratePlugin`. Generates
 Placements based on distinct values of the designated node label (TAS).
 
-**NodeResourcesFit (Existsing)** Implements `PlacementScorePlugin`. Scores
+**NodeResourcesFit (Existing)** Implements `PlacementScorePlugin`. Scores
 Placements to maximize utilization (tightest fit) and minimize fragmentation.
 
 **PodGroupPodsCount (New)** Implements `PlacementScorePlugin`. Scores
@@ -565,7 +565,7 @@ Those tests are located at [tas_test.go‎](https://github.com/kubernetes/kubern
 
 - Scalability tests on large clusters with high placement counts.
 - Comprehensive e2e testing.
-- Cluster autoscaling compomnents are aware of workload topology constraints.
+- Cluster autoscaling components are aware of workload topology constraints.
 
 #### GA
 
@@ -723,7 +723,7 @@ are actively using the feature.
 
 - [X] API .status
   - Object: PodGroup
-  - Condition Name: `PodGroupScheduled`
+  - Condition Name: `PodGroupInitiallyScheduled`
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
--- a/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
+++ b/keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
@@ -9,7 +9,6 @@
 - [Proposal](#proposal)
   - [User Stories (Optional)](#user-stories-optional)
     - [Story 1: AI Training in a Single Rack](#story-1-ai-training-in-a-single-rack)
-    - [Story 2: Workload using Interconnected DRA Devices](#story-2-workload-using-interconnected-dra-devices)
   - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
@@ -23,7 +22,7 @@
     - [Phase 3: Placement Scoring and Selection](#phase-3-placement-scoring-and-selection)
   - [Scheduler Plugins](#scheduler-plugins)
   - [Beta Extensions](#beta-extensions)
-  - [Potential Future Extensions](#potential-future-extensions)
+  - [Future Extensions](#future-extensions)
   - [Test Plan](#test-plan)
     - [Prerequisite testing updates](#prerequisite-testing-updates)
     - [Unit tests](#unit-tests)
@@ -32,6 +31,7 @@
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
     - [Beta](#beta)
+    - [GA](#ga)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
@@ -64,9 +64,9 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
   - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
 - [X] (R) Production readiness review completed
 - [X] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
-- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [X] "Implementation History" section is up-to-date for milestone
+- [X] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [X] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 <!--
 **Note:** This checklist is iterative and should be reviewed and updated every time this enhancement is being considered for a milestone.
@@ -89,15 +89,15 @@ lifecycle introduced in KEP-4671, rather than fundamentally altering the schedul
 loop itself. It extends this foundation by enabling the evaluation of scheduling
 options within specific "Placements" (subsets of the cluster). These Placements
 represent candidate domains (sets of
-nodes or DRA resources) where the entire workload is theoretically feasible. The
+nodes) where the entire workload is theoretically feasible. The
 scheduler then simulates the placement of the full group of pods within these
 domains, utilizing existing filtering and scoring logic to ensure high-fidelity
 decisions before committing resources.
 
 This design introduces specific extensions to the Kubernetes Workload API to
-support `TopologyConstraints` and `DRAConstraints`, defines new interfaces
-within the Scheduling Framework (`PlacementGeneratorPlugin`, `PlacementStatePlugin`,
-`PlacementScorerPlugin`), and details the algorithmic flow required to schedule Pod
+support `TopologyConstraints`, defines new interfaces
+within the Scheduling Framework (`PlacementGeneratePlugin`,
+`PlacementScorePlugin`), and details the algorithmic flow required to schedule Pod
 Groups while maintaining compatibility with the scheduler's existing ecosystem.
 
 ## Motivation
@@ -110,11 +110,10 @@ primarily establishes the API structure. For workloads sensitive to inter-pod
 communication, simply grouping pods is insufficient; their physical placement
 within the cluster's network topology is a decisive factor in their performance.
 
-In this KEP, we propose an algorithm for topology-aware and DRA-aware scheduling
+In this KEP, we propose an algorithm for topology-aware scheduling
 that operates directly within the Kubernetes kube-scheduler. The core objective
 is to ensure that pods belonging to a Workload are co-located within optimal
-topological domains - such as specific racks or blocks - or are bound to shared
-Dynamic Resource Allocation (DRA) devices that require cohesive management.
+topological domains - such as specific racks or blocks that require cohesive management.
 Without this level of precision, workloads may be fragmented across disparate
 network domains, drastically degrading performance and wasting the potential of
 expensive hardware.
@@ -130,7 +129,7 @@ Topology-aware scheduling is not a new concept and is currently addressed by
 external admission control systems like Kueue or alternative schedulers like
 Volcano. However, relying on external admission controllers decouples the
 topology decision from the scheduler's core logic, while alternative schedulers
-introduce operational complexity. We believe that embedding topology and DRA
+introduce operational complexity. We believe that embedding topology
 awareness deeply into the kube-scheduler is critical enough to warrant
 standardization. This integration allows the algorithm to leverage the full
 fidelity of the scheduler's existing pod-level filtering and scoring plugins,
@@ -139,14 +138,13 @@ need for external dependencies.
 
 ### Goals
 
-- To enhance kube-scheduler to be able to perform topology-aware and DRA-aware
+- To enhance kube-scheduler to be able to perform topology-aware
   scheduling for multi-pod workloads, as defined by the Workload API
   ([KEP-4671](https://kep.k8s.io/4671)).
 - To optimize the placement of distributed workloads by co-locating pods based
-  on network topology and DRA resource availability.
+  on network topology.
 - To introduce new extension points and phases within the Kubernetes scheduler
-  framework to support the concept of "Placements" (candidate sets of nodes
-  and DRA resources).
+  framework to support the concept of "Placements" (candidate sets of nodes).
 - To define the required changes to the Workload API (KEP-4671) to support
   Topology scheduling constraints.
 - To leverage the scheduler's existing pod-level filtering and scoring logic
@@ -157,8 +155,8 @@ need for external dependencies.
 ### Non-Goals
 
 - To define the required changes to the Workload API (KEP-4671) to support
-  ResourceClaims for DRA-aware workload scheduling. These changes will be
-  proposed in a separate KEP:
+  ResourceClaims for DRA-aware workload scheduling and their scheduling.
+  These changes have been proposed in a separate KEP:
   [KEP-5729: DRA: ResourceClaim Support for Workloads](https://kep.k8s.io/5729)
 - To replace the functionality of external workload queueing and admission
   control systems like Kueue. This proposal focuses on the in-scheduler
@@ -168,11 +166,10 @@ need for external dependencies.
 - To handle all aspects of the workload lifecycle management beyond
   scheduling.
 - To implement Workload-level preemption logic.
-- To integrate with cluster autoscaling mechanisms in this phase.
-- To support complex multi-PodSet dependency resolution with backtracking or
+- To support complex multi pod dependency resolution with backtracking or
   parallel processing in the initial version.
 - To automatically discover network topology; the mechanisms rely on topology
-  information being present (e.g., via node labels or DRA ResourceSlices).
+  information being present (e.g., via node labels).
 
 ## Proposal
 
@@ -180,18 +177,16 @@ This proposal introduces an API to define constraints on a PodGroup (a
 collection of pods within a Workload) requiring it to be scheduled onto a
 specific subset of nodes or resources.
 
-We support two fundamental types of constraints:
+We support one type of constraints:
 
 1. **Topology Constraint (Node Label Co-location)**: Ensures all pods in a
    PodGroup are placed onto nodes sharing a common topological characteristic
    (e.g., same rack), defined by a specific node label.
 
-2. **DRA Constraint (Shared Dynamic Resource Allocation)**: Ensures all pods in a
-   PodGroup bind to a single DRA claim fulfilled from a single, shared,
-   co-located resource (e.g., interconnected network interfaces or
-   accelerators).
-
-The scheduler is extended to interpret these new PodGroup level scheduling constraints and similarly to scheduling pods on nodes (available scheduling options), find a "Placement" for this PodGroup among the feasible options (subsets of nodes and DRA resources) that satisfies them.
+The scheduler is extended to interpret these new PodGroup level scheduling constraints
+and similarly to scheduling pods on nodes (available scheduling options), find
+a "Placement" for this PodGroup among the feasible options (subsets of nodes)
+that satisfies them.
 
 ### User Stories (Optional)
 
@@ -203,14 +198,6 @@ need to be located in the same server rack to minimize latency. I define a
 label. The scheduler identifies a rack with sufficient capacity and schedules
 all pods there at once.
 
-#### Story 2: Workload using Interconnected DRA Devices
-
-As a cluster administrator, I want to schedule a workload that requires a set of
-specialized accelerators that are physically interconnected. I use a
-`DRAConstraint` targeting a specific `ResourceClaimTemplate`. The scheduler
-finds a set of DRA resources (ResourceSlice) that are co-located and binds the
-workload's pods to them.
-
 ### Notes/Constraints/Caveats (Optional)
 
 ### Risks and Mitigations
@@ -218,15 +205,14 @@ workload's pods to them.
 - **Scheduling Latency:** Evaluating multiple placements involves running
   filter/score plugins multiple times (multiple attempts to schedule a PodGroup considering all topology options).
 
-  - **Mitigation:** Implement pre-filtering optimizations to reject infeasible
-    placements early based on aggregate resource availability.
+  - **Mitigation:** Implement early placement rejection optimization to stop
+    feasibility checks for placements which cannot fit gang scheduled PodGroups.
 
 - **Complexity of Pod Group Scheduling:** Scheduling heterogeneous Pod Groups
   can be complex.
 
-  - **Mitigation:** The initial version supports sequential processing of pods
-    within a PodGroup, avoiding complex backtracking or parallel processing
-    in the alpha release.
+  - **Mitigation:** We support only sequential processing of pods
+    within a PodGroup, avoiding complex backtracking or parallel processing.
 
 ## Design Details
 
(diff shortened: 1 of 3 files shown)


pmady:
i did something similar in Volcano - added GPU NUMA topology awareness to the scheduler plugin ([volcano#5095](https://github.com/volcano-sh/volcano/pull/5095), [apis#229](https://github.com/volcano-sh/apis/pull/229), [resource-exporter#12](https://github.com/volcano-sh/resource-exporter/pull/12)). couple things that came up:

the TopologyConstraint covers rack/zone level stuff well but does it handle intra-node topology? we had to deal with GPU-to-NUMA-node affinity - if a pod gets a GPU on NUMA node 0 but the CPU/memory is on NUMA node 1 you get hit with cross-NUMA penalties. would that fall under DRAConstraint or does TopologyConstraint need to go deeper than node labels?

also re the Volcano alternative section - biggest pain point with the external scheduler path was keeping topology data current. we ended up building a DaemonSet that scrapes sysfs for GPU NUMA mappings, pushes to a CRD, then the scheduler plugin reads it. lots of moving parts. having the scheduler talk directly to DRA state is way cleaner.

on the latency mitigation - we saw the same thing, pre-filtering by aggregate GPU count per NUMA node before doing full placement made a big difference. sounds like that's what you're going for here too?

lmk if any of the Volcano implementation details would be useful context.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/kep.yaml
```diff
@@ -45,3 +46,6 @@ disable-supported: true
 
 # The following PRR answers are required at beta release
 metrics:
+  - scheduler_placement_count
+  - scheduler_placement_schedule_attempts_total
+  - scheduler_placement_scheduling_attempt_duration_seconds
```
pacoxu: Can you share some details about the metrics here? 

- Is `scheduler_placement_count` a counter(if yes, `_total` is a general name) or a histogram.
- `scheduler_placement_schedule_attempts_total` is similar to `scheduler_schedule_attempts_total` and `scheduler_podgroup_schedule_attempts_total`, but placement may be different. Is `xxx_placement_evaluations_total` better? https://github.com/kubernetes/enhancements/pull/6134/changes#diff-7ae886ce193e1c72e6a0673ded84f5552891a2b813918acd8cd5e5a3b6748c06R827
- ditto here: Is `xxx_placement_evaluation_duration_seconds` better?


BTW, in the alpha implementation PR, the metrics update is about `scheduler_framework_extension_point_duration_seconds{extension_point=PlacementGenerate}`. in https://github.com/kubernetes/kubernetes/pull/137083/changes#diff-0aa9024787a576d2bfcd5766f717a75b88978dbd1c927bf4139faa91b508ddddR1996.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
pacoxu:
```suggestion
integrating a Topology-Aware workload scheduling algorithm into
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -288,125 +276,78 @@ Placement represents a candidate domain (nodes and resources) for a PodGroup.
...
+// PlacementScorePlugin is an interface for plugins that score feasible Placements.
+type PlacementScorePlugin interface {
+    Plugin
+
+    // ScorePlacement calculates a score for a given Placement.
+    // This function is called only for Placements that have been deemed feasible for the sufficient number of pods in the PodGroup scheduling cycle.
+    // The PodGroupAssignments indicates the node assigned to each pod within this Placement.
+    // The returned score is a int64 with higher scores generally indicating more preferable Placements.
+    // Plugins can implement various scoring strategies, such as bin packing to minimize resource fragmentation.
+    ScorePlacement(ctx context.Context, state PodGroupCycleState, podGroup PodGroupInfo, placement *PodGroupAssignments) (int64, *Status)
```
pacoxu:
```suggestion
    ScorePlacement(ctx context.Context, state PlacementCycleState, podGroup PodGroupInfo, placement *PodGroupAssignments) (int64, *Status)
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -288,125 +276,78 @@ Placement represents a candidate domain (nodes and resources) for a PodGroup.
...
-    // of the given Placement for the PodGroup.
-    AssumePlacement(ctx context.Context, state *framework.CycleState, podGroup *PodGroupInfo, placement *Placement) *framework.Status
-
-    // RevertPlacement reverts the temporary scheduling context changes made by AssumePlacement.
-    // This should be called after the evaluation of a Placement is complete to restore
-    // the scheduler's state and allow other Placements to be considered.
-    RevertPlacement(ctx context.Context, state *framework.CycleState, podGroup *PodGroupInfo, placement *Placement) *framework.Status
+    // GeneratePlacements generates a list of potential Placements for the given PodGroup within the parent placement.
+    // Each Placement represents a candidate set of resources, e.g., nodes matching a selector.
+    GeneratePlacements(ctx context.Context, state PodGroupCycleState, podGroup PodGroupInfo, parentPlacement *Placement) (*GeneratePlacementsResult, *Status)
```
sohankunkerkar, line 314: qq: The `parentPlacement` parameter supports nested placement generation, but Non-Goals lists multi-level constraints as out of scope. Is this parameter exercised at beta, or is it forward-compatible scaffolding?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -583,11 +554,14 @@ necessary to implement this enhancement.
 
 #### Beta
 
-- Support for "Potential Future Extensions" (Prioritized placement, etc.)
-  evaluated.
 - Scalability tests on large clusters with high placement counts.
 - Comprehensive e2e testing.
-- Cluster autoscaling compomnents are aware of workload topology constraints. 
+- Cluster autoscaling compomnents are aware of workload topology constraints.
```
sohankunkerkar: What does _aware_ mean here? Does the autoscaler need to scale up nodes within a specific topology domain, or avoid scaling down nodes in a TAS-placed PodGroup? If this work isn't ready for v1.37, does it block beta?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +643,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
...
+  (`scheduler_pod_scheduling_duration_seconds`) for both TAS and non-TAS pods compared to the baseline
+  (behavior with the feature disabled).
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
-<!--
-Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
-implementation difficulties, etc.).
--->
+No.
```
sohankunkerkar: Beta introduces `PercentageOfPlacementsToScore` which samples placements. Operators tuning this parameter need to know how many placements were generated, how many were evaluated, and how many were feasible. Without that, there's no way to tell if the sampling is too aggressive

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
sohankunkerkar, line 832: Note: 
The implementation in PR: [139064](https://github.com/kubernetes/kubernetes/pull/139064) iterates over sampled placements, and for each evaluates all pods across that placement's nodes. The actual cost is O(#sampled_placements * #pods * #nodes_per_placement). `PercentageOfPlacementsToScore` bounds the first term but the per-placement cost is still multiplicative.

sohankunkerkar (commented):
First pass from the PRR reviewer side.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -241,43 +227,45 @@ scheduling constraints. An optional `ScheduleConstraints` field is added to the
...
 }
 
-// PodGroupSchedulingConstraints holds the scheduling constraints for the PodGroup.
+// PodGroupSchedulingConstraints defines scheduling constraints (e.g. topology) for a PodGroup.
 type PodGroupSchedulingConstraints struct {
-    // TopologyConstraints specifies desired topological placements for all pods
-    // within this PodGroup.
+    // Topology defines the topology constraints for the pod group.
+    // Currently only a single topology constraint can be specified. This may change in the future.
     TopologyConstraints []TopologyConstraint
```
mm4tt: We ended up with just `Topology` to not repeat constraints - https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/scheduling/v1alpha3/types.go#L600

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -433,17 +375,9 @@ The algorithm proceeds in three main phases for a given PodGroup.
...
-  2. Run default workload scheduling algorithm with the given context.
-
-  3. If all pods fit, the Placement is marked Feasible.
-
-  4. Call `RevertPlacement`.
+  1. Run default workload scheduling algorithm with the given set of nodes.
 
-- **Potential Optimization:** Pre-filtering can check aggregate resources
-  requested by PodGroup Pods before running the full simulation.
+  2. If all required pods fit, the Placement is marked Feasible.
```
mm4tt: Can you clarify "all required"? What does it mean for PodGroups with Gang and Basic policies?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -469,44 +403,65 @@ The algorithm proceeds in three main phases for a given PodGroup.
...
+
+1. Multiple PlacementGeneratePlugins
+   
+   The beta version supports defining multiple PlacementGeneratePlugins. When multiple
+   such plugins are configured, the scheduler framework runs them independently. It then
+   merges their results by calculating non-empty intersections of the placements returned
+   by the different plugins, allowing the system to easily handle complex placement
+   requirements.
+
+2. Early Rejection of Placements
```
mm4tt, line 436: This is already described in the recent revision of https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/4671-gang-scheduling/README.md. 

We do that optimization for gang-scheduling not TAS per-se and it's implemented in the gang-scheduling plugin and not the "label-based TAS" plugin. The only TAS related thing is that calling it is placement aware (we call per placement). I'd consider either removing this or linking to the gang-scheduling KEP.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Supporting multi-level constraints is a non-goal for this KEP however in parallel we are working on introduction of the CompositePodGroup API (https://github.com/kubernetes/enhancements/issues/6012) which will allow defining multi-level topology constraints for the workloads. `parentPlacement` will be used there for nested placement generation.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
mm4tt, line 635: As discussed in https://github.com/kubernetes/enhancements/pull/6017, what is currently written is oversimplification and it's not true in general. There might be leader running older version and api-server might have API enabled. It will cause scheduler to ignore PG topology constraints and schedule pods without it, which is an irreversible action.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Triggering scale up for TAS-placed PodGroup without taking into account the TAS constraints will most likely lead to workload being unschedulable on newly provisioned capacity. Because of this as a bare minimum we want to block scale ups for PodGroups which have TAS constraints. For selected well-known node labels autoscalers may reenable scale ups but this should be done only if the autoscaler is able to predict the node label value for a newly provisioned node and make sure that the PodGroup will be schedulable on it.

The actual implementation of this change on the autoscalers side should not block beta of this KEP (in parallel to this KEP we are also graduating PodGroup API to beta and this will most likely be a blocker for the actual implementation) but autoscalers should have this work planned.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Yes, that is true. In the kep.yaml I have included introduction of a new `scheduler_placement_count` metric which could be used for tuning `PercentageOfPlacementsToScore`.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: If we assume that the placements are exclusive (they do not share common nodes) which is true for the current node label based constraints implementation O(#sampled_placements * #nodes_per_placement) should be equal to O(#nodes) as each node can appear in only one placement at most. For larger clusters because of the way how a default value for `PercentageOfPlacementsToScore` is calculated the average number of nodes processed during TAS algorithm execution can be as low as 5% of the total number of nodes in the cluster but this is still O(#nodes) asymptotically.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Indeed I have missed updating this one. Thanks for pointing this out.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: It is at least `minCount` pods for Gang scheduling policy and at least one pod for Basic scheduling policy. I will update the description.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have added a link to this KEP below.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
mm4tt, line 358: Can we describe what happens for PGs with already scheduled pods - e.g. some pods were added later. What part of this algorithm will ensure we honor the placement selected in the previous scheduling cycle?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/kep.yaml
44past4: All three suggestions make sense for me. I have updated the names of the proposed metrics to:
  - scheduler_placement_total
  - scheduler_placement_evaluations_total
  - scheduler_placement_evaluation_duration_seconds

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: If some of PodGroups's pods are already scheduled TopologyPlacementGenerator plugin when generating placements for a PodGroups will only return domains which include all nodes on which those already scheduled pods are running - https://github.com/kubernetes/kubernetes/blob/643e407efef84a7bd37ef23d9ed4ec0ddc123571/pkg/scheduler/framework/plugins/topologyaware/topology_placement.go#L73

44past4: I have added a note to the Phase 1: Candidate Placement Generation section about already scheduled pods handling.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -241,43 +227,45 @@ scheduling constraints. An optional `ScheduleConstraints` field is added to the
...
-// TopologyConstraint describes a desired topological colocation for all pods in the PodGroup.
+// TopologyConstraint defines a topology constraint for a PodGroup.
 type TopologyConstraint struct {
-    // Level specifies the key of the node label representing the topology domain.
+    // Key specifies the key of the node label representing the topology domain.
     // All pods within the PodGroup must be colocated within the same domain instance.
-    // Different replicas of the PodGroup can land on different domain instances.
+    // Different PodGroups can land on different domain instances even if they derive from the same PodGroupTemplate.
     // Examples: "topology.kubernetes.io/rack"
     Level string
```
macsko:
```suggestion
    Key string
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -469,44 +410,67 @@ The algorithm proceeds in three main phases for a given PodGroup.
 
 ### Scheduler Plugins
 
-**TopologyPlacementPlugin (New)** Implements `PlacementGeneratorPlugin`. Generates
+**TopologyPlacementGenerator (New)** Implements `PlacementGeneratePlugin`. Generates
 Placements based on distinct values of the designated node label (TAS).
 
-**PlacementBinPackingPlugin (New)** Implements `PlacementScorerPlugin`. Scores
+**NodeResourcesFit (Existsing)** Implements `PlacementScorePlugin`. Scores
```
macsko:
```suggestion
**NodeResourcesFit (Existing)** Implements `PlacementScorePlugin`. Scores
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +652,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
...
-- [ ] Events
-  - Event Reason:
-- [ ] API .status
-  - Condition name:
-  - Other field:
-- [ ] Other (treat as last resort)
-  - Details:
+- [X] API .status
+  - Object: PodGroup
+  - Condition Name: `PodGroupScheduled`
```
macsko: We updated the condition name in #4671:
```suggestion
  - Condition Name: `PodGroupInitiallyScheduled`
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +652,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
...
 
-<!--
-Describe manual testing that was done and the outcomes.
-Longer term, we may want to require automated upgrade/rollback tests, but we
-are missing a bunch of machinery and tooling and can't do that now.
--->
+We'll perform manual testing of the upgrade -> downgrade -> upgrade path using the following sequence:
+
+1. Start a local Kubernetes v1.36 cluster with `TopologyAwareWorkloadScheduling` feature
+   gate disabled and `GenericWorkload` feature gate enabled (default behavior).
```
macsko: `GenericWorkload` was disabled in v1.36, so it's not a default behavior

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -583,11 +563,14 @@ necessary to implement this enhancement.
 
 #### Beta
 
-- Support for "Potential Future Extensions" (Prioritized placement, etc.)
-  evaluated.
 - Scalability tests on large clusters with high placement counts.
 - Comprehensive e2e testing.
-- Cluster autoscaling compomnents are aware of workload topology constraints. 
+- Cluster autoscaling compomnents are aware of workload topology constraints.
```
macsko:
```suggestion
- Cluster autoscaling components are aware of workload topology constraints.
```

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -524,54 +488,70 @@ future separate KEPs improving and extending the proposed functionality:
...
+  labels.
+  
+- PlacementScorePlugin: Test scoring of placements.
 
 - Algorithm Logic: Test the sequential processing of Placements and the
   selection logic based on scores.
 
-- DRA Integration: specific tests for DRATestPlugin plugin.
-
 #### Integration tests
```
macsko, line 533: What about performance tests?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -469,44 +410,67 @@ The algorithm proceeds in three main phases for a given PodGroup.
...
-**DRATestPlugin (New)** Implements `PlacementGeneratorPlugin` and `PlacementStatePlugin`
-and is used only for testing the algorithm's support for DRA-aware scheduling.
-
-- **Generator:** Returns Placements derived from available Devices satisfying
-  claims shared by all Pods within a PodGroup.
-
-- **State:** Temporarily assigns AllocationResults to Devices during the
-  Assume phase.
-
 ### Beta Extensions
```
macsko, line 422: You could also mention that TAS algorithm will support NNNs, evaluating the placement with nominated nodes first

44past4: Added.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have removed the remark that this is a default behaviour.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Added at the end of the Integration tests section.

johnbelamaric:
> the TopologyConstraint covers rack/zone level stuff well but does it handle intra-node topology? we had to deal with GPU-to-NUMA-node affinity - if a pod gets a GPU on NUMA node 0 but the CPU/memory is on NUMA node 1 you get hit with cross-NUMA penalties. would that fall under DRAConstraint or does TopologyConstraint need to go deeper than node labels?

@pmady this KEP is focused on inter-node topology not intra-node topology. However, since WAS simulates scheduling of the individual Pods, any intra-node topology constraints expressed in DRA must be satisfied to successfully schedule the workload. These two types of topology are orthogonal.

Originally we had the concept of a DRA constraint in here; it looks like @44past4 pulled it out in this revision. I think that's OK - it's not clear to me what that added that is not already applied if you use #5729 to generate resource claims that contain node selectors in their allocation results. But again - that is about inter-node constraints, and in that case both the label-based TAS and the resourceclaim node selector constraints would be applied to any successfully scheduled workload.

Eventually, we may want to be able to model inter-node constraints that cannot be done with label selectors. For example, rails-based alignment of GPUs that are not all in the same node. DRA will likely be the way we do that, but the use cases are not as common. It's not clear to me that the original "DRA constraints" concept in this KEP is the approach we would use there.

@44past4 @wojtek-t please confirm this summary.

44past4:
> @44past4 @wojtek-t please confirm this summary.

As written by John this KEP is focused on inter-node multi-pod topology not intra-node single pod topology and therefore the things like node GPU/CPU/memory NUMA alignment is not what we want to solve here. This in my mind is something that DRA could be used for with GPU and CPU resource requests being represented as separate ResourceClaims with a constraint that both should belong to the same NUMA node.

sanposhiho:
/unassign

I will leave approvals to @dom4ha @macsko about this. If I read through the diff correctly, the major part is the clean up and there's nothing newly controversial around the implementation changes. Of course, still feel free to explicitly tag me in the discussion which I should be involved.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -524,54 +506,75 @@ future separate KEPs improving and extending the proposed functionality:
...
   selection logic based on scores.
 
-- DRA Integration: specific tests for DRATestPlugin plugin.
-
 #### Integration tests
 
-- Topology Awareness: Verify that pods with TopologyConstraint are correctly
-  co-located on nodes sharing the label.
+For Alpha we implemented integration tests to ensure basic functionalities of
+to
```
wojtek-t: nit: remove

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
wojtek-t: The previous paragraph addresses that:

"Thus, users should not set those fields before confirming all control-plane instances were upgraded to the version supporting those."


However, it would be good to add a callout what happens if a PodGroup having SchedulingConstraints was scheduled when the FG was disabled and then FG is reenabled.

IIUC, if this was scheduled across different domains, we will not be able to add any more pods to it later, but what runs will continue to run, right?

@44past4

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +675,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
 enablement/disablement tests - the logic will be covered by regular feature tests.
 
 For the newly introduced API fields, dedicated enablement/disablement tests at the
-kube-apiserver registry layer will be added in Alpha.
+kube-apiserver registry layer will be added.
```
wojtek-t: Were those added already?

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +675,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
...
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-<!--
-Describe manual testing that was done and the outcomes.
-Longer term, we may want to require automated upgrade/rollback tests, but we
-are missing a bunch of machinery and tooling and can't do that now.
--->
+We'll perform manual testing of the upgrade -> downgrade -> upgrade path using the following sequence:
+
+1. Start a local Kubernetes v1.36 cluster with `TopologyAwareWorkloadScheduling` feature
```
wojtek-t: This comment is relevant here too:
https://github.com/kubernetes/enhancements/pull/6076#discussion_r3306183704

Please adjust

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -669,119 +675,99 @@ The scheduler algorithm changes are purely in-memory and doesn't require any ded
...
-- [ ] Events
-  - Event Reason:
-- [ ] API .status
-  - Condition name:
-  - Other field:
-- [ ] Other (treat as last resort)
-  - Details:
+- [X] API .status
+  - Object: PodGroup
+  - Condition Name: `PodGroupInitiallyScheduled`
```
wojtek-t, line 759: That doesn't necessarily mean that TAS is enabled...

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
wojtek-t: Can you please describe this metric in the PRR itself too?
It actually seems relevant to "How can someone using this feature know that it is working for their instance?" question...

wojtek-t (commented):
Just few minor comments from me - overall this looks great already.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Removed.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Yes, they were. I have updated the section with links to the existing tests.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: As suggested I have updated the test scenario to rely on API server and scheduler using 1.37 version only.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have updated the description to clarify that `schedulingConstrains`need to be set on PodGroup and I have added information about `scheduler_placement_total` metric which can be used as an alternative.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: I have added description for the 3 new metrics.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: As suggested I have added a section describing the scheduler behaviour after FG enablement for already scheduled PodGroup.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
```diff
@@ -668,120 +678,115 @@ The feature starts working again.
...
+active scheduler instance has the feature disabled, it will schedule pods using the
+standard non-TAS method, falling back to a default workload scheduling algorithm.
 
-Be sure to consider highly-available clusters, where, for example,
-feature flags will be enabled on some API servers and not others during the
-rollout. Similarly, consider large clusters and how enablement/disablement
-will rollout across nodes.
--->
+This results in a fallback to the status quo behavior, meaning that pods will be
+still scheduled, but PodGroup-level toplogy scheduling constraints won't be applied.
```
wojtek-t, line 696: Can you please briefly mention the addition to Version Skew here?

wojtek-t (commented):
Just one more very minor comment from me - other than that both PRR and update itself LGTM.

keps/sig-scheduling/5732-topology-aware-workload-scheduling/README.md
44past4: Added.

wojtek-t:
/approve PRR

LGTM overall - thanks!
/lgtm

macsko:
/approve
From scheduling POV

pmady:
that makes sense — intra-node via DRA + inter-node via TopologyConstraint as separate layers. our volcano work is all intra-node (GPU-to-NUMA affinity scoring) so that maps to the DRA side. will follow how DRA ends up modeling device-to-NUMA relationships

PR #6138, opened 2026-06-01
Update volume health enhancement to support comprehensive health 

gnufied:
xref https://github.com/kubernetes/enhancements/issues/1432

### DIFF
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -324,6 +324,11 @@ a CSI driver reports same adverse condition for a volume
 repeatedly no updates are written to API server because 
 we only store `lastTransitionTime`.
 
+Another way we can mitigate this risk is by using in-process
+queuing system so that unhealthy update to same PVC and pod can
+be queued and only latest values are written to the underlying
+object if multiple health updates are queued.
+
 A second risk is compromised or malicious nodes lying about a
 volume's health to influence operator behavior. This was the
 original objection to per-PVC node-keyed health, and it's why the
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -149,7 +149,12 @@ fields that automation can consume.
 - Define a small, machine-parseable error vocabulary
   (`Inaccessible`, `DataLoss`, `Degraded`, `StorageUnreachable`,
   `StorageDegraded`) that admits driver-specific elaboration via
-  `reason` and `message`.
+  `reason` and `message`. The vocabulary is intentionally
+  extensible: future CSI spec revisions MAY add values, drivers
+  that adopt a newer spec MAY report them alongside existing
+  ones, and COs MUST tolerate values they do not recognize. A
+  CO that does not recognize a value SHOULD surface it for
+  observability and MUST NOT misclassify it as healthy.
 - Make health reporting opt-in per driver (via CSI capabilities) and
   per cluster (via the `CSIVolumeHealth` feature gate). A driver
   that implements no health capability is not probed and incurs no
@@ -349,11 +354,12 @@ under this KEP.
 
 The shared types are `VolumeHealth` (per-volume) and
 `StorageBackendHealth` (per-backend). Each carries a status drawn
-from a small enum, an optional CamelCase `reason` for grouping,
-and an optional human-readable `message`. `VolumeHealth` carries a
-list of entries rather than a single status, because a volume can
-exhibit multiple concurrent conditions and forcing the driver to
-choose one would lose information.
+from a small enum, a required CamelCase `reason` that distinguishes
+distinct conditions sharing the same status, and an optional
+human-readable `message`. `VolumeHealth` carries a list of entries
+rather than a single status, because a volume can exhibit multiple
+concurrent conditions and forcing the driver to choose one would
+lose information.
 
 ```protobuf
 message VolumeHealth {
@@ -363,9 +369,11 @@ message VolumeHealth {
     // The health status category. REQUIRED.
     VolumeHealthErrorType status = 1;
 
-    // A brief CamelCase machine-parseable reason. OPTIONAL.
-    // Used for grouping volumes with similar conditions in
-    // dashboards and CLI displays.
+    // A brief CamelCase machine-parseable reason. REQUIRED.
+    // Together with status, reason forms the unique identity of
+    // an entry: the Plugin MUST NOT return multiple
+    // VolumeHealthEntry messages for the same volume with the
+    // same (status, reason) combination.
     string reason = 2;
 
     // A user-friendly description. OPTIONAL.
@@ -416,7 +424,10 @@ message StorageBackendHealth {
   // Health status. REQUIRED.
   StorageHealthErrorType status = 1;
 
-  // A brief CamelCase machine-parseable reason. OPTIONAL.
+  // A brief CamelCase machine-parseable reason. REQUIRED.
+  // The Plugin MUST NOT return multiple StorageBackendHealth
+  // messages with the same (status, reason, volume_capability)
+  // combination.
   string reason = 2;
 
   // A user-friendly description. OPTIONAL.
@@ -478,7 +489,11 @@ message ControllerListVolumeHealthResponse {
 ```
 
 For drivers whose backends cannot enumerate efficiently, the sidecar
-falls back to per-volume `ControllerGetVolumeHealth` calls.
+falls back to per-volume `ControllerGetVolumeHealth` calls. A driver
+that advertises neither `LIST_VOLUME_HEALTH` nor `GET_VOLUME_HEALTH`
+is not probed by the sidecar at all, and no controller-side health
+is recorded for its volumes; this is the default for drivers that
+have not opted in.
 
 ```protobuf
 message ControllerGetVolumeHealthRequest {
@@ -580,9 +595,13 @@ type VolumeHealthStatus struct {
     // conditions is the set of adverse conditions reported by
     // the CSI controller plugin. An empty list (or absence)
     // means the controller plugin reports no adverse condition.
+    // Conditions are uniquely identified by the (status, reason)
+    // tuple, matching the CSI spec's uniqueness rule for
+    // VolumeHealthEntry.
     // +optional
     // +listType=map
     // +listMapKey=status
+    // +listMapKey=reason
     Conditions []VolumeHealthCondition `json:"conditions,omitempty"`
 
     // lastProbeTime is the most recent time the CO obtained a
@@ -595,9 +614,9 @@ type VolumeHealthCondition struct {
     // "Inaccessible", "DataLoss", "Degraded".
     Status VolumeHealthStatusType `json:"status"`
     // reason is a brief CamelCase machine-parseable reason
-    // (e.g. "VolumeNotFound").
-    // +optional
-    Reason string `json:"reason,omitempty"`
+    // (e.g. "VolumeNotFound"). Required; together with status
+    // it forms the unique identity of a condition entry.
+    Reason string `json:"reason"`
     // message is a human-readable description.
     // +optional
     Message string `json:"message,omitempty"`
@@ -614,11 +633,15 @@ const (
 )
 ```
 
-`Conditions` is a `+listType=map` keyed by `status`, so each adverse
-category appears at most once per PVC. A driver reporting two
-distinct categories produces two entries; a driver re-reporting an
-existing category updates its `reason`, `message`, and
-`lastTransitionTime` (the last only when `reason` changes).
+`Conditions` is a `+listType=map` keyed by `(status, reason)`,
+matching the CSI spec's uniqueness rule for `VolumeHealthEntry`. A
+driver reporting two distinct categories produces two entries, and a
+driver reporting the same category with two different reasons (for
+example `Degraded`/`HighLatency` and `Degraded`/`MultipathLoss`) also
+produces two entries; both forms are first-class conditions, not
+collisions. A driver re-reporting an entry whose `(status, reason)`
+tuple is unchanged updates `message` and `lastProbeTime` only and
+preserves `lastTransitionTime`.
 
 The Pod field is owned by the kubelet on the pod's node. It mirrors
 the two-level shape of KEP-4680: an outer entry per `pod.spec.volumes`
@@ -644,10 +667,13 @@ type PodVolumeHealth struct {
     Name string `json:"name"`
 
     // conditions is the set of adverse conditions reported by
-    // the CSI node plugin for this volume on this node.
+    // the CSI node plugin for this volume on this node. Keyed by
+    // (status, reason) to match the CSI spec's uniqueness rule for
+    // VolumeHealthEntry.
     // +optional
     // +listType=map
     // +listMapKey=status
+    // +listMapKey=reason
     Conditions []VolumeHealthCondition `json:"conditions,omitempty"`
 
     LastProbeTime metav1.Time `json:"lastProbeTime,omitempty"`
@@ -655,18 +681,27 @@ type PodVolumeHealth struct {
 ```
 
 The CSINode field is owned by the kubelet on the node, keyed by
-driver name. The Node Authorizer is extended to allow the kubelet
-to PATCH `csinodes/status.storageHealth` for its own node only,
-matching how it already PATCHes `csinodes/spec.drivers`.
+driver name. CSINode does not have a `/status` subresource today;
+this KEP adds one, and extends the Node Authorizer and the
+NodeRestriction admission plugin to allow the kubelet to PATCH it
+for its own node only. This matches the scoping pattern the kubelet
+already uses to PATCH `csinodes` (no subresource) for
+`spec.drivers` registration.
 
 ```go
 type CSINodeStatus struct {
     // storageHealth is the set of backend health reports for
     // each CSI driver registered on the node, as observed by
-    // the kubelet via NodeGetStorageHealth.
+    // the kubelet via NodeGetStorageHealth. A single driver may
+    // report multiple conditions; entries are uniquely identified
+    // by the (name, status, reason) tuple, matching the CSI
+    // spec's uniqueness rule for StorageBackendHealth scoped to
+    // the registered driver.
     // +optional
     // +listType=map
     // +listMapKey=name
+    // +listMapKey=status
+    // +listMapKey=reason
     // +featureGate=CSIVolumeHealth
     StorageHealth []StorageHealthCondition `json:"storageHealth,omitempty"`
 }
@@ -679,8 +714,9 @@ type StorageHealthCondition struct {
     Status StorageHealthStatusType `json:"status"`
 
     // reason is a brief CamelCase machine-parseable reason.
-    // +optional
-    Reason string `json:"reason,omitempty"`
+    // Required; together with name and status it forms the unique
+    // identity of a condition entry.
+    Reason string `json:"reason"`
 
     // message is a human-readable description.
     // +optional
@@ -721,22 +757,36 @@ follow the same rules. Each reads the current state from its
 configured CSI plugin, computes the desired status, compares it to
 what's stored on the API object, and PATCHes only on difference.
 
-The detection rule is direct. A probe that returns one or more
-adverse conditions writes those conditions to the corresponding
-status field. If the same conditions are already present, the
-write is a no-op and is suppressed.
+The desired-state rule is replacement, not union. The driver's
+current report is the authoritative set of conditions for the
+volume (or backend) it is reporting on; the writer replaces the
+stored `Conditions` list with the driver's list rather than merging
+the two. A condition entry that the driver no longer returns is
+dropped from the API object on the same PATCH. Without this rule,
+stale `(status, reason)` entries would accumulate on the object and
+never be cleared, because no single later report would name them
+again. Per-condition `LastTransitionTime` is preserved across
+replacement when the same `(status, reason)` tuple is still present.
 
-The recovery rule is more careful, because the spec does not let
-the CO trust an absence in a single list pass. A writer treats a
-volume as recovered, and clears its conditions, only on a positive
-healthy signal:
+The detection rule follows from replacement. A probe that returns
+one or more adverse conditions writes that exact set to the
+corresponding status field. If the set already matches what is
+stored (same `(status, reason)` tuples, same `message`, same
+ancillary fields), the write is a no-op and is suppressed.
 
-- an empty `health_statuses` from `ControllerGetVolumeHealth`,
-  which is per-volume and therefore consistent;
-- an empty `health_statuses` from `NodeGetVolumeHealth`, also
-  per-volume; or
+The recovery rule needs special handling only for
+`ControllerListVolumeHealth`, because the spec explicitly does not
+let the CO trust an absence in a single paged list pass. For the
+other RPCs, recovery is just a special case of replacement:
+
+- an empty `health_statuses` from `ControllerGetVolumeHealth`
+  replaces the stored conditions with the empty set, which is the
+  recovered state;
+- an empty `health_statuses` from `NodeGetVolumeHealth` does the
+  same on the node side; and
 - a `ControllerListVolumeHealth` result in which the volume has
-  been absent across two consecutive complete list cycles.
+  been absent across two consecutive complete list cycles is
+  treated as a recovery, and stored conditions are cleared.
 
 The two-cycle rule for List absorbs the paging inconsistency the
 spec warns about. Without it, a volume that
@@ -749,7 +799,7 @@ A driver that supports both `LIST_VOLUME_HEALTH` and
 into one: when a previously-unhealthy volume is absent from the
 current list, the sidecar issues a single `ControllerGetVolumeHealth`
 to confirm. An empty `health_statuses` from Get clears the
-condition; a non-empty result simply updates it.
+condition; a non-empty result replaces it.
 
 A failed RPC is never a recovery signal. The writer leaves existing
 conditions untouched and increments
@@ -758,13 +808,17 @@ conditions untouched and increments
 from the explicit-healthy case, and conflating them would have the
 cluster report transient driver failures as volume recoveries.
 
-`LastTransitionTime` is per-condition. When a condition is added,
-the writer sets it to the probe time. When an existing condition's
-`(status, reason)` tuple is unchanged, the writer preserves it. When
-`reason` changes for the same `status`, the writer updates it.
-Consumers reading `LastTransitionTime` get a stable answer to
-"how long has this condition been in this shape" rather than a
-moving timestamp that updates on every probe.
+`LastTransitionTime` is per-condition. When a new
+`(status, reason)` entry first appears, the writer sets it to the
+probe time. As long as that entry is present in subsequent reports,
+the writer preserves it on every replacement, even when `message`
+changes. When the entry is dropped from the report (the condition
+recovered, or the driver replaced it with a different
+`(status, reason)` tuple), the entry disappears from the API object,
+and a future re-report of the same tuple gets a fresh
+`LastTransitionTime`. Consumers reading `LastTransitionTime` get a
+stable answer to "how long has this exact condition been present"
+rather than a moving timestamp that updates on every probe.
 
 The sidecar's polling and the kubelet's polling are independent.
 Each has its own configurable interval, its own timeout, and its
@@ -781,22 +835,32 @@ health for volumes whose mount failed.
 
 ### Authorization
 
-Three new permissions are granted, each scoped as tightly as the
-existing model allows:
+The three writers each need permission scoped as tightly as the
+existing model allows. Two of the three already exist; one is new.
 
 - The external monitor sidecar's service account is granted PATCH
-  on `persistentvolumeclaims/status`. It already has the GET and
-  LIST permissions it needs on PVCs and PVs.
-- Kubelets are granted PATCH on `pods/status.volumeHealth` for pods
-  bound to their own node.
-- Kubelets are granted PATCH on `csinodes/status.storageHealth` for
-  their own node only.
+  on `persistentvolumeclaims/status`. The sidecar already has the
+  GET and LIST permissions it needs on PVCs and PVs in its
+  upstream RBAC manifest, so this is the only permission added.
+- Kubelet PATCH on `pods/status` is already authorized today by
+  the Node Authorizer, with the NodeRestriction admission plugin
+  scoping it to pods bound to the kubelet's own node. No new
+  authorization surface is needed for `Pod.Status.VolumeHealth`;
+  it is just another field reachable through the existing
+  `pods/status` subresource. NodeRestriction's existing
+  pod-binding check enforces own-pod scoping.
+- Kubelet PATCH on `csinodes/status` is new. CSINode has no
+  `/status` subresource today, and the Node Authorizer explicitly
+  rejects all `csinodes` subresources. This KEP adds the `/status`
+  subresource on the CSINode resource and extends both the Node
+  Authorizer and NodeRestriction admission so a kubelet can PATCH
+  the CSINode whose name matches its own node and no other. The
+  scoping is identical to the kubelet's existing PATCH of the main
+  `csinodes` resource for `spec.drivers`.
 
-Both kubelet extensions follow the same scoping pattern the Node
-Authorizer already applies to existing kubelet writes on those
-objects. Nodes are explicitly not granted any access to PVC status,
-which is the central security improvement over the
-originally-rejected "node writes per-PVC health" design.
+Nodes are explicitly not granted any access to PVC status, which
+is the central security improvement over the originally-rejected
+"node writes per-PVC health" design.
 
 ### Feature gate
 
@@ -832,9 +896,11 @@ implementation:
 ##### Unit tests
 
 - `k8s.io/kubernetes/pkg/apis/core/validation` and
-  `k8s.io/kubernetes/pkg/apis/storage/validation`: validation of the
-  new fields, including the `+listType=map +listMapKey=status`
-  constraint that rejects duplicate-status entries.
+  `k8s.io/kubernetes/pkg/apis/storage/validation`: validation of
+  the new fields, including the composite-key uniqueness rules
+  (`(status, reason)` for `VolumeHealthCondition` and
+  `(name, status, reason)` for `StorageHealthCondition`) and
+  rejection of empty `reason` values.
 - `pkg/registry/core/{persistentvolumeclaim,pod}` and
   `pkg/registry/storage/csinode`: feature-gate drop-on-save,
   including the case where the field is already set on the old
@@ -851,11 +917,16 @@ implementation:
 
 - The apiserver round-trips the new fields with the gate on and
   drops them with it off.
-- The Node Authorizer accepts kubelet writes only for own-node
-  entries.
+- The new `csinodes/status` subresource is registered, accepts
+  kubelet PATCHes, and is gated by `CSIVolumeHealth`.
+- The Node Authorizer and NodeRestriction admission accept kubelet
+  PATCH on `pods/status` and `csinodes/status` only for the
+  kubelet's own pods and own node, respectively, and reject all
+  cross-node attempts.
 - The external monitor sidecar runs end-to-end against the mock
   driver and exercises the unhealthy → updated → healthy → cleared
-  cycle.
+  cycle, including the replace-not-union semantics when a follow-up
+  report drops a previously-reported `(status, reason)` entry.
 
 ##### e2e tests
 
@@ -878,8 +949,9 @@ implementation:
   [container-storage-interface/spec][spec-pr].
 - The mock CSI driver implements the new RPCs.
 - The `CSIVolumeHealth` feature gate is plumbed through the apiserver
-  and kubelet, including field validation, drop-on-save, and the
-  Node Authorizer extensions.
+  and kubelet, including field validation, drop-on-save, the new
+  `csinodes/status` subresource, and the Node Authorizer and
+  NodeRestriction admission extensions.
 - The external monitor sidecar uses the new RPCs and writes to
   `pvc.status.healthStatus`.
 - Initial unit and integration tests are in place.
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -783,7 +783,7 @@ const (
 ```
 
 `AccessModes` and `VolumeMode` exist because backends can degrade
-asymmetrically: a multipath fault may affect RWX multi-attach while
+asymmetrically: a NFS connectivity problem may affect NFS RWX volumes while
 RWO works fine, and a block-mode regression on a backend that serves
 both block and filesystem volumes may not affect filesystem volumes
 at all. These fields let drivers express that asymmetry; consumers
@@ -1158,7 +1158,7 @@ The feature has two dependencies, both of which are opt-in:
 In steady state with no unhealthy volumes, no-op suppression brings
 sustained PATCH rate to zero. Under sustained all-unhealthy load,
 the rate is bounded by the writer's polling interval, which is
-operator-configurable.
+configured via kubelet's `VolumeStatsAggPeriod` command line parameter.
 
 ###### Will enabling / using this feature result in introducing new API types?
 
@@ -1335,7 +1335,10 @@ is opaque to k8s. But We could in future add `backend_id` as a first class field
 specify it in health responses. This will allow drivers to report volume health at more granular 
 level from different backends.
 
-3. Report entire storage backend health from control-plane as well.
+3. Report entire storage backend health from control-plane as well.We will have to find out the use case
+of such reporting more thoroughly - such as, what would be recovery if entire backend is down etc. We will also
+have to figure out which k8s object will reflect this cluster health from control-plane.
+
 
 ## Infrastructure Needed
 
--- a/keps/sig-storage/1432-volume-health-monitor/README.md
+++ b/keps/sig-storage/1432-volume-health-monitor/README.md
@@ -1,1003 +1,1271 @@
 # KEP-1432: Volume Health Monitor
 
-## Table of Contents
-
 <!-- toc -->
 - [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
 - [Motivation](#motivation)
-- [User Experience](#user-experience)
-  - [Use Cases](#use-cases)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
 - [Proposal](#proposal)
-- [Implementation](#implementation)
-  - [Kubelet Metrics changes](#kubelet-metrics-changes)
-  - [CSI changes](#csi-changes)
-    - [Add ControllerGetVolume RPC](#add-controllergetvolume-rpc)
-    - [Add Node Volume Health Function](#add-node-volume-health-function)
-  - [External controller](#external-controller)
-    - [CSI interface](#csi-interface)
-    - [Node down event](#node-down-event)
-  - [Kubelet](#kubelet)
-    - [CSI interface](#csi-interface-1)
-  - [Alternatives](#alternatives)
-    - [Alternative option 1](#alternative-option-1)
-    - [Alternative option 2](#alternative-option-2)
-    - [Alternative option 3](#alternative-option-3)
-    - [Optional HTTP(RPC) service](#optional-httprpc-service)
-    - [External node agent](#external-node-agent)
-      - [CSI interface](#csi-interface-2)
-- [Graduation Criteria](#graduation-criteria)
-  - [Alpha](#alpha)
-  - [Alpha -&gt; Beta](#alpha---beta)
-  - [Beta -&gt; GA](#beta---ga)
-- [Test Plan](#test-plan)
-  - [Unit tests](#unit-tests)
-  - [E2E tests](#e2e-tests)
+  - [User Stories](#user-stories)
+    - [Story 1: Volume deleted out-of-band](#story-1-volume-deleted-out-of-band)
+    - [Story 2: Backend unreachable from a subset of nodes](#story-2-backend-unreachable-from-a-subset-of-nodes)
+    - [Story 3: Cross-driver dashboards](#story-3-cross-driver-dashboards)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Where reports live, and why](#where-reports-live-and-why)
+  - [The CSI surface](#the-csi-surface)
+  - [The Kubernetes surface](#the-kubernetes-surface)
+  - [Reconciliation contract](#reconciliation-contract)
+  - [Authorization](#authorization)
+  - [Feature gate](#feature-gate)
+  - [Test Plan](#test-plan)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
-  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
   - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
   - [Monitoring Requirements](#monitoring-requirements)
   - [Dependencies](#dependencies)
   - [Scalability](#scalability)
   - [Troubleshooting](#troubleshooting)
 - [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Embed VolumeCondition in existing RPCs (the original alpha)](#embed-volumecondition-in-existing-rpcs-the-original-alpha)
+  - [A standalone per-PVC VolumeHealth CRD](#a-standalone-per-pvc-volumehealth-crd)
+  - [A per-PVC map[node]Health directly on pvc.status](#a-per-pvc-mapnodehealth-directly-on-pvcstatus)
+  - [PV taints with NoEffect](#pv-taints-with-noeffect)
+  - [DRA-style device taints for storage (KEP-5055)](#dra-style-device-taints-for-storage-kep-5055)
+  - [Push-based health ingest](#push-based-health-ingest)
+  - [A richer error enum](#a-richer-error-enum)
+- [Infrastructure Needed](#infrastructure-needed)
 <!-- /toc -->
 
 ## Release Signoff Checklist
 
-<!--
-**ACTION REQUIRED:** In order to merge code into a release, there must be an
-issue in [kubernetes/enhancements] referencing this KEP and targeting a release
-milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
-of the targeted release**.
-
-For enhancements that make changes to code or processes/procedures in core
-Kubernetes i.e., [kubernetes/kubernetes], we require the following Release
-Signoff checklist to be completed.
-
-Check these off as they are completed for the Release Team to track. These
-checklist items _must_ be updated for the enhancement to be released.
--->
-
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
-- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [x] (R) KEP approvers have approved the KEP status as `implementable`
-- [x] (R) Design details are appropriately documented
-- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
-- [x] (R) Graduation criteria is in place
-- [x] (R) Production readiness review completed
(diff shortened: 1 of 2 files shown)


keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
-The external node agent will go through all the pods on that node and find out all volumes used by those pods. It will watch all those volumes and call NodeGetVolumeStatus() to find out their health status.
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
 
-The external node agent reports events on the pod object. If multiple pods are using the same volume, report events on all pods.
+- [X] Metrics
+  - `csi_volume_health_probe_duration_seconds`: histogram of
+    health-probe RPC latency.
+  - `csi_volume_health_probe_total`: counter of probes by
+    outcome.
+  - `csi_volume_health_status`: gauge per posted condition.
```
jsafrane: Per what condition? Controller or node health? I think we may need a separate metric for them, esp. when they are reported by different components.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+original alpha could see only volumes that were already healthy enough
+to mount, which inverts the use case.
+
+The third constraint was that findings were surfaced only as
+Kubernetes events. Events are an operator-affordance, not a state
+machine. They have a TTL, they aren't joined to the PVC by API
+consumers in any structured way, and they can't be driven by a
+remediation controller without screen-scraping. Anyone serious about
+acting on volume health needs durable status.
+
```
jsafrane: This belongs to "alternatives".

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+  ones, and COs MUST tolerate values they do not recognize. A
+  CO that does not recognize a value SHOULD surface it for
+  observability and MUST NOT misclassify it as healthy.
+- Make health reporting opt-in per driver (via CSI capabilities) and
+  per cluster (via the `CSIVolumeHealth` feature gate). A driver
+  that implements no health capability is not probed and incurs no
+  cost.
+- Keep ownership consistent with the existing Node Authorizer model:
+  nodes never write PVC status.
+
```
jsafrane, line 183: I think metrics + alerts should still be a goal. Cluster admins needs to know if something fails and an alert is the usual method. Then they can figure out the details using the new API fields.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+This KEP separates health from stats, separates the controller and
+node planes, and stores the result on durable, well-typed status
+fields that automation can consume.
+
+### Goals
+
+- Provide a CSI-spec-level mechanism for drivers to report
+  per-volume health from the controller, per-volume health from the
+  node (including for volumes that have not been successfully
+  published), and overall storage-backend health from a node.
```
jsafrane, line 162: Nit: I am not sure about wording "storage-backend health from a node". What does it mean? The storage backend health should have the same  single value, regardless which node reports it.

We want to report health of the driver on the node, its configuration and network between the storage backend and the node. Is there a better term for this? I'd call it driver health / node health, not backend health.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+- Surface those reports on stable Kubernetes API status fields:
+  controller-side volume health on
+  `PersistentVolumeClaim.Status.HealthStatus`, node-side per-volume
+  health on `Pod.Status.VolumeHealth`, and node-side backend health
+  on `CSINode.Status.StorageHealth`.
```
jsafrane, line 167: Nit: `Storage` is already implied by presence in CSINode object. Or do we plan to introduce other healths there?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+reconciliation against the backend and starts reporting
+`{status: Inaccessible, reason: VolumeNotFound}` from
+`ControllerListVolumeHealth`. The external monitor sidecar writes
+that onto `pvc.status.healthStatus`. The application owner sees the
+condition on `kubectl describe pvc` and recreates the PVC from a
+snapshot. None of the nodes hosting the workload need to be involved.
+In fact, if the workload still has a mount cached in the kernel page
+cache, the nodes might not notice for some time.
+
+#### Story 2: Backend unreachable from a subset of nodes
```
jsafrane, line 253: I'd like to see a story of a CSI driver that has multiple storage backends, such as multiple Ceph clusters. What `NodeGetStorageHealth` reports? Does the driver need to find all possible backends, typically defined in StorageClasses or in PVs?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+## Design Details
+
+### Where reports live, and why
+
+Three Kubernetes objects gain one optional status field each. The
+choice of object is the most consequential decision in this KEP, so
+it's worth saying plainly why each is where it is.
+
+Controller-reported volume health goes on
```
jsafrane, line 358: It would be nice to mention also metric for each health component (controller / node / driver)

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
-}
-```
+A driver may implement any combination of these or none at all. A
+driver that advertises only `GET_VOLUME_HEALTH` on the controller is
+probed per-volume by the sidecar; a driver that advertises
+`LIST_VOLUME_HEALTH` is probed via list, even if it also advertises
+`GET_VOLUME_HEALTH`.
 
-### External controller
+### The Kubernetes surface
```
jsafrane, line 632: Would it make sense to merge it with "Where reports live, and why" above? The above chapter says a lot about "why" (which is great!), but both describe "where".

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
-        ConditionUnknown ConditionStatus = "Unknown"
+    StorageHealthUnreachable StorageHealthStatusType = "StorageUnreachable"
+    StorageHealthDegraded    StorageHealthStatusType = "StorageDegraded"
 )
 ```
 
-To avoid stale information being stored on a PVC, each periodic update will mark the PVC with the latest health information, replacing previous health information added by the external controller. If the PVC becomes healthy after being marked as unhealthy previously, the controller should remove the previous information.
+`AccessModes` and `VolumeMode` exist because backends can degrade
+asymmetrically: a multipath fault may affect RWX multi-attach while
+RWO works fine, and a block-mode regression on a backend that serves
```
jsafrane: This is not a good example: "a multipath fault may affect RWX multi-attach while RWO works fine"
Multipath failure will affect both in the same way.

RWX filesystem volumes with NFS and RWX block volumes with iSCSI, whose multipath degraded, would be a better example.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+[ ] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+Two pieces of test infrastructure are extended in lockstep with the
+implementation:
+
+- [`kubernetes-csi/csi-test`](https://github.com/kubernetes-csi/csi-test):
+  the mock CSI driver implements the four new RPCs and capabilities,
+  with configurable health responses.
```
jsafrane: csi-test is not a mock CSI driver, it's a test suite for real CSI drivers. We need to add tests for the new CSI calls.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+- [`kubernetes-csi/csi-driver-host-path`](https://github.com/kubernetes-csi/csi-driver-host-path):
+  the hostpath driver gains a flag to inject specific health states,
+  for upstream e2e jobs.
```
jsafrane, line 909: Test-specific return values are injected into csi-driver-host-path via  `-proxy-endpoint` and k/k e2e tests can add  hooks to return them (errors, volume health conditions, ...). [Example](https://github.com/kubernetes/kubernetes/blob/5ad79eae2dcbf33df3b35c48ec993d30fbda46dd/test/e2e/storage/csi_mock_volume.go#L803)

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+  `pkg/registry/storage/csinode`: feature-gate drop-on-save,
+  including the case where the field is already set on the old
+  object and the gate is now disabled.
+- `pkg/kubelet/volumemanager`: probing, `Pod.Status` patching, and
+  no-op suppression.
+- `plugin/pkg/auth/authorizer/node`: own-pod and own-node scoping
+  for the new PATCH permissions.
+- `kubernetes-csi/external-health-monitor`: capability detection,
+  list-vs-get fallback, leader election, no-op suppression, and the
+  two-cycle (or Get-confirmed) recovery rule.
```
jsafrane, line 929: are you really unit testing the leader election? 😅

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+##### Integration tests
+
+- The apiserver round-trips the new fields with the gate on and
+  drops them with it off.
```
jsafrane: This is usually an unit test

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+- Volume-side enums (`Inaccessible`, `DataLoss`, `Degraded`)
+  injected via the hostpath driver: verify each surfaces on
+  `pvc.status.healthStatus` and `pod.status.volumeHealth`, then
+  recovery clears the condition.
+- Backend-side enums (`StorageUnreachable`, `StorageDegraded`)
+  injected via the hostpath driver: verify each surfaces on
+  `csinode.status.storageHealth`, then recovery clears.
+- Negative authorization: a kubelet attempting to PATCH another
+  node's CSINode is rejected; a kubelet attempting to PATCH a PVC
+  status is rejected.
```
jsafrane: How are you going to convince kubelet to try patch another CSINode in e2e?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+### Upgrade / Downgrade Strategy
+
+The upgrade order is:
+
+1. Roll out the apiserver with `CSIVolumeHealth` enabled. The new
+   status fields begin to be accepted.
+2. Roll out kubelets with the gate enabled.
+3. Update CSI drivers that implement the new capabilities. Drivers
+   that don't are unaffected.
```
jsafrane, line 990: I think the driver can be upgraded anywhere in this sequence and kubelet won't just call the new RPCs.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+(preserving them if already set on the old object, per the standard
+field-level feature-gate handling), the kubelet does not call the
+new RPCs, and the Node Authorizer extensions are inactive.
+
+The external monitor sidecar does not depend on a feature gate. Its
+deployment is the controller-side opt-in. If the apiserver does not
+have `CSIVolumeHealth` enabled, the sidecar's writes are silently
+dropped server-side, which is the right behavior: the sidecar
+doesn't need to know whether the cluster has the feature on, only
+that its own writes succeed or fail.
```
jsafrane, line 889: While the approach looks solid, devil can be in details: the sidecar will see empty health conditions and will try to update them on every health sync.

We need to ensure it detects the API server clearing the fields and stop the health sync (or at least stop updating remaining PVCs). Possibly also resync at a lower rate.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+  object and the gate is now disabled.
+- `pkg/kubelet/volumemanager`: probing, `Pod.Status` patching, and
+  no-op suppression.
+- `plugin/pkg/auth/authorizer/node`: own-pod and own-node scoping
+  for the new PATCH permissions.
+- `kubernetes-csi/external-health-monitor`: capability detection,
+  list-vs-get fallback, leader election, no-op suppression, and the
+  two-cycle (or Get-confirmed) recovery rule.
+
+##### Integration tests
```
jsafrane, line 931: In general, please double check that the tests below actually can be done as integration tests. Integration tests run a single component (e.g. KCM controller) with an API server. Nobody has ever added CSIDriver to the mix or tested presence of subresources.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+new condition and that condition appearing on the corresponding
+Kubernetes status field is bounded by one probe interval.
 
-The external node agent will go through all the pods on that node and find out all volumes used by those pods. It will watch all those volumes and call NodeGetVolumeStatus() to find out their health status.
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
 
-The external node agent reports events on the pod object. If multiple pods are using the same volume, report events on all pods.
+- [X] Metrics
+  - `csi_volume_health_probe_duration_seconds`: histogram of
+    health-probe RPC latency.
```
jsafrane, line 1138: I assume it's a kubelet metric. Is it both `NodeGetVolumeHealth` and `NodeGetStorageHealth`?
Plus we will need a similar one in the sidecar.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
-of this feature?**
-  <!--
-  Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
-  implementation difficulties, etc.).
-  -->
-  No.
+In steady state with no unhealthy volumes, no-op suppression brings
+sustained PATCH rate to zero. Under sustained all-unhealthy load,
+the rate is bounded by the writer's polling interval, which is
+operator-configurable.
```
jsafrane: kubelet currently shares the health probing period with `VolumeStatsAggPeriod` (How frequently to calculate and cache volume disk usage for all pods). Do you still want to reuse it? Should it be configurable per driver?

jsafrane: Since the nr. of PVCs is not bound and all PVCs can get suddenly unhealthy on a network failure and the health state has `LastProbeTime` that needs to be updated after each probe (both node and control plane), the load on the API server can be high, even with a modest polling interval.

I don't have any smart idea how to prevent it, but it should be at least clearly written that the API server load can get really high and stay like that until volumes get healthy again.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+The cost depends on the driver and backend; this KEP does not
+prescribe an implementation.
 
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
 
-### Scalability
+Three new optional status fields. Each entry holds a small list
+of conditions (status / reason / message / timestamps). Fields
+are populated only when a driver reports adverse conditions, so
+typical steady-state cost is zero.
```
jsafrane, line 1207: Will we set an upper bound of nr. of health states for each PVC to keep the size manageable?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
 
-* **What steps should be taken if SLOs are not being met to determine the problem?**
-  If SLOs are not being met, analysis should be made to understand what have caused the problem. Debug level logging should be enabled to collect verbose logs. Look at logs to find out what might have caused the events to be missed. If it indicates an underlying problem on the storage system, then storage admin can be pulled in to help find the root cause.
+### Troubleshooting
 
-[supported limits]: https://git.k8s.io/community//sig-scalability/configs-and-limits/thresholds.md
-[existing SLIs/SLOs]: https://git.k8s.io/community/sig-scalability/slos/slos.md#kubernetes-slisslos
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+The kubelet and the sidecar log errors and back off. No status
```
jsafrane, line 1230: Not sure if it should be mentioned anywhere: there is no exp. backoff on write error, right? It could multiply the API server load.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+Considered as a way to encode health in an existing mechanism. PV
+taints were never approved upstream for general use, and `Taint` was
+designed to express scheduling effects, not status; the `value` field
+is a single string, so encoding multiple concurrent conditions
+degenerates into an ad-hoc protocol.
+
+### DRA-style device taints for storage (KEP-5055)
+
+Considered. We could not identify a use case where a workload author
+would write a `Toleration` for, say, `Degraded` storage. Storage
```
jsafrane, line 1323: We could not find any use case, because this KEP does not propose any remediations.
Tolerating / not tolerating `Degraded` is indeed app specific - some apps absolutely need high throughput and won't tolerate it, some other apps may be less sensitive.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+remediation is overwhelmingly an operator concern (rebuild from
+snapshot, fail over to a replica), not a tolerations concern.
+`DataLoss` does not translate cleanly into a taint effect, because
+the right reaction is application-specific. Backend health on
```
jsafrane, line 1327: `DataLoss` can translate easily to scheduler effect - NoSchedule, NoExecute. A pod that tries to run `fsck` can tolerate it.

jsafrane: TBH, I don't see much difference between device taints and volume taints except for complexity of driver health on a node. That was IMO the main reason not to use them.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+A storage admin accidentally deletes a backend volume that is bound
+to a PVC. The controller plugin learns about it on its next
+reconciliation against the backend and starts reporting
+`{status: Inaccessible, reason: VolumeNotFound}` from
+`ControllerListVolumeHealth`. The external monitor sidecar writes
+that onto `pvc.status.healthStatus`. The application owner sees the
+condition on `kubectl describe pvc` and recreates the PVC from a
+snapshot. None of the nodes hosting the workload need to be involved.
+In fact, if the workload still has a mount cached in the kernel page
+cache, the nodes might not notice for some time.
```
jsafrane, line 251: I don't think that a workload can survive re-creation of a PV. Maybe of NFS / CIFS.
But anything based on a block volume will need a Pod restart and this KEP does not allow any automation on Kubernetes side. The cluster operator (human or automation) needs to do it on its own.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+A network change makes the storage backend's data path unreachable
+from one rack of nodes while the control plane keeps working fine.
+CSI controller calls all succeed and report the volume healthy. On
+the affected nodes the kubelet calls `NodeGetStorageHealth`, gets
+back `{status: StorageUnreachable, reason: NetworkPartition}`, and
+writes that onto `csinode.status.storageHealth` for the driver. A
+future scheduler opt-in (out of scope) can use that to avoid placing
+new pods needing the driver onto those nodes. Today, an operator can
+see the condition with a single `kubectl get csinode` and route
+around it manually.
```
jsafrane, line 264: The KEP misses any automatic remediation. An example could fit here well: a database operator sees that a database replica runs on a node where the storage is not healthy and automatically moves the replica to a healthier node.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+N/A.
+
+## Alternatives
+
+### Embed VolumeCondition in existing RPCs (the original alpha)
+
+The original alpha embedded `VolumeCondition` in `ListVolumes`,
+`ControllerGetVolume`, and `NodeGetVolumeStats` and surfaced findings
+as Kubernetes events. The Motivation section explains why this shape
```
jsafrane: "as Kubernetes events" and metrics

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
-This option is not in the main proposal because it is a push-based method while CSI uses pull-based method.
+The CSI spec removes the legacy alpha `VOLUME_CONDITION` capability
+and `volume_condition` fields. No deprecation period applies because
+the prior shape was alpha and never declared stable. Drivers that
+implemented the alpha shape continue to compile against the old
+spec but must migrate to the new RPCs to surface health under this
+KEP.
 
-#### External node agent
+### Monitoring Requirements
```
x0rw, line 1107: Do you think scalability tests need to be performed against health flapping(oscillating between Healthy - Inaccessible) rather than only steady-state conditions?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+    condition.
 
-## Graduation Criteria
-### Alpha
-* Initial feature implementation, including:
-  * External controller volume health monitoring.
-  * Kubelet volume health monitoring.
-* Implementation in the csi-mock driver.
-* Add basic unit tests.
+  Exact label sets are defined alongside the implementation, not
```
x0rw, line 1145: Could we at least document expectations around label cardinality (for example, avoiding volume-specific, node-specific, or unbounded driver-provided labels)?

x0rw (commented):
PRR shadow.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: >I'd call it driver health / node health

I think that driver health is already taken. It generally means the health of CSI driver and we have separate health checks for that. 

I am still open to ideas about what to call it.  But IMO calling it storage backend health is not that bad of a idea.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: That is a good idea. We can set `nmaxitems` to 4-5 for each PVC's health status. But still I don't have a solid idea of how many entries will be sufficient.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: It is possible that - when you delete a underlying volume of a PV, something in the backend prevents deletion of actual volume until volume is unmounted/detached from the node. So, in this case - PVC can keep working for atime, but as soon as you try to migrate the workload to a different node it will break. That is why, I have left this as a valid use case.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I thought we discussed this in the call , but `DataLoss` recovery varies from workload to workload - say an application such as etcd/cassandra has replicas from which it can build new replica (like sync from other nodes that make up the quorum), but for some other apps it may mean restoring a new PV from snapshot. 

So it isn't ideal to just say - if app just ran `fsck` then it can tolerate this health effect, because in many cases that may not be enough. 

Again - I am not sure taints and tolerations fit neatly into this. Generally speaking, I don't know why will a application want to tolerate `DataLoss` health effect, rather than just rebuild the replica (using whatever strategy).

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Currently we do not have ability to mark just one backend as unavailable, since ceph backends are typically separated by `clusterID` parameter in storageclass (or embedded in pv path). 

I am leaning towards reporting `NodeGetStorageHealth` as unhealthy with `message` string expressing the granularity more clearly. But it isn't designed to be machine parseable. 

This was discussed in design doc - https://docs.google.com/document/d/1QsziFslmA5yD8eX_BYSKFxebjdQLguzBIb74roN4cDQ/edit?tab=t.0 but I am afraid we punted this for now.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I was thinking of making two things configurable via `CSIDriver` object:

1. How often the checks run. 
2. If there is any intertia before we flip the condition from failing to healthy or back to failing. 

But - I am not sure if we have to do this for current alpha.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
```diff
@@ -1,33 +1,45 @@
...
+
+latest-milestone: "v1.37"
 
-latest-milestone: "v1.24"
-stage: "alpha"
 milestone:
-  alpha: "v1.21"
-  beta: "v1.25"
-  stable: "v1.26"
+  alpha: "v1.37"
```
xing-yang: "v1.21" was when the feature was initially introduced. I think it should be kept. The latest milestone will show that it is a new alpha in 1.37.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
```diff
@@ -1,33 +1,45 @@
-title: Volume Health Monitor 
+title: Volume Health Monitor
 kep-number: 1432
 authors:
-  - "@NickrenREN"
-  - "@xing-yang"
```
xing-yang, line 5: I think the original authors should be kept. Just move them to the end of the author list.
That's what I usually do.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+The second constraint was that `NodeGetVolumeStats` is contractually
+defined to be valid only for staged or published volumes. The volumes
+most worth reporting on are precisely the ones the kubelet can never
+call `NodeGetVolumeStats` against: the ones whose filesystems are
+corrupt, the ones the driver could not stage in the first place. The
+original alpha could see only volumes that were already healthy enough
+to mount, which inverts the use case.
+
+The third constraint was that findings were surfaced only as
```
xing-yang: The reason that the original feature only provides volume health condition as events and metrics is that we didn't want to add first class fields to PVCs and Pods that Kubernetes built-in controllers do not know how to react to. 
Can you add this to the KEP?

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: "health from a node" sounds like it refers to local storage only. Do you mean whether this node has access to storage?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+
+The two reports are independent. A volume can be reported as
+`Inaccessible` from a node, because that node lost its data-plane
+route to the backend, while the controller plugin reports the same
+volume as healthy. A volume can also be reported as `Degraded` from
+the controller plane while a node sees it just fine, because the
+backend's monitoring caught a slow trend that the data path hasn't
+yet expressed. Forcing these two perspectives into one truth would
+hide useful information; this design surfaces them separately and
+lets consumers decide what to do.
```
xing-yang, line 236: I think the original CSI spec has language on how to reconcile if the conflict happens.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
+The node plugin learns to answer two questions: "what's the condition
+of this volume on this node" (via `NodeGetVolumeHealth`) and "what's
+the condition of the storage backend from this node" (via
+`NodeGetStorageHealth`). The kubelet calls them and writes the
+results to `pod.status.volumeHealth` for the pods running on that
+node and to `csinode.status.storageHealth` for the driver registered
+on that node. Both targets are objects the kubelet can already write
+to under its existing Node Authorizer permissions, so no new
+authorization surface is needed for the node-side path.
+
```
xing-yang, line 227: Can you add a paragraph and describe what CSI RPCs will be removed?

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Can you add a use case for local storage?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1298 @@
...
 
-### Kubelet Metrics changes
+    // A brief CamelCase machine-parseable reason. REQUIRED.
+    // Together with status, reason forms the unique identity of
+    // an entry: the Plugin MUST NOT return multiple
+    // VolumeHealthEntry messages for the same volume with the
+    // same (status, reason) combination.
+    string reason = 2;
 
-Add a new field in the [VolumeStats metrics API](https://github.com/kubernetes/kubernetes/blob/v1.22.1/staging/src/k8s.io/kubelet/pkg/apis/stats/v1alpha1/types.go#L263).
```
xing-yang, line 148: We should support VolumeStats metrics?

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Agreed. I will add back metric emission part. Since, we don't have ability to emit alerts in plain-k8s, we will have to leave that as exercise to the k8s vendor.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I think we still want to emit metrics for volume health but they will be different.

keps/sig-storage/1432-volume-health-monitor/README.md
jsafrane: You can add new fields later, but they will need to start as alpha and will slow down graduation to beta / GA.

keps/sig-storage/1432-volume-health-monitor/README.md
jsafrane: The app (etcd / cassandra) won't tolerate `DataLoss` and I don't expect their pods to run fsck. But a cluster admin / automation may try to run some recovery pod with `fsck` and they need to tolerate the data loss.

> Future KEPs may layer behavior on top (a CSIDriver opt-in to
prevent pod scheduling on `Inaccessible` volumes is a natural next
step), 

If you're postponing scheduler changes to future, you will still need to add some sort of toleration to run these "storage recovery pods".

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: In many cases IMO we have done both, we have injected the specific return values but also added new capabilities in the host-path driver. One doesn't exclude the other, we will leave further granular implementation details for later I guess.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Added to alternative section but briefly keeping old design here as well.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Can you elaborate why you think it means local-storage?  We are reporting two distinct health  from the node:

1. Health of individual volumes from the node from the perspective of CSI driver.
2. Health of overall storage backend from the node from the perspective of CSI driver. 

Currently IMO, we do not have a better word/language to describe 2nd use case but "Storage backend health" makes sense and is actually incredibly useful.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: My intention is to express health of storage backend as visible from CSI driver. I am not yet sure, what is the best word to describe this in API but `StorageHealth` seemed close enough. Just calling it `CSINode.Status.Health` seems bit ambigious because we don't know if it expresses driver health or the health of the storage subsystem. 

But some other options might be `CSINode.Status.StorageBackendHealth`  or shorter `BackendHealth` or `StorageSystemsHealth` or `BackendStatus`

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Updated this to reflect csi sanity tests.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: removed.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: removed.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: removed.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Updated wording to reflect this.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Added wording that says - if API server is rejecting our health updates, we will retry at a slower rate.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I have removed integration tests entirely. I do not think we use it in storage much. cc @torredil

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Added a use case for local-storage with topolvm.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: added a new alternative section for this.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
gnufied: fixed.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
gnufied: fixed.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Updated the proposal to include newer stats metrics.

keps/sig-storage/1432-volume-health-monitor/README.md
torredil: Monitoring Requirements section is required when targeting a release to beta, so we have time to finalize metrics - should not be a blocker for alpha and we will definitely revisit this as we hammer it out in implementation.

FFR +1 on separate metric per source.  We'll also want to drop `csi_volume_health_probe_total` (the histogram already emits a `_count` series, so probe totals come for free),  and drop the `_status` suffix on the gauges since they carry status as a label and the name would just repeat it (see https://prometheus.io/docs/practices/naming/#labels).

So following proper naming conventions, we'd have something like:

- `csi_volume_health_probe_duration_seconds`
- `csi_controller_volume_health`
- `csi_node_volume_health`
- `csi_node_storage_backend_health`

keps/sig-storage/1432-volume-health-monitor/README.md
torredil: Sure, scalability testing is a beta requirement so we can add it to the Beta graduation criteria section. The current scale test we have defined with 5k PVCs is a little aggressive honestly, so I suggest updating it to something like:

>  A scalability test on a cluster of at least 1,000 PVCs across at least 50 nodes confirms that (a) steady-state apiserver write rate stays at or near zero, and (b) under sustained health flapping (volumes oscillating between healthy and unhealthy), PATCH rate stays bounded by the configured probe interval and does not amplify.

cc: @gnufied

keps/sig-storage/1432-volume-health-monitor/README.md
torredil: I agree with dropping the integration tests, we can achieve comprehensive code coverage via unit and e2e testing

keps/sig-storage/1432-volume-health-monitor/README.md
torredil: bounded, low cardinality labels are standard instrumentation rules we'll strive to follow and they're heavily scrutinized by sig-instrumentation during implementation, so I don't think we need to spell it out in this KEP. This is a beta requirement so I'm happy to revisit this when targeting that release.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
```diff
@@ -1,33 +1,45 @@
...
+    components:
+      - kube-apiserver
+      - kubelet
 disable-supported: true
 
 metrics:
+  - csi_volume_health_probe_duration_seconds
+  - csi_volume_health_probe_total
+  - csi_volume_health_status
+  - csi_storage_health_status
```
xing-yang: These don't match the metrics defined in README.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
xing-yang: Looks like this is not fixed?

keps/sig-storage/1432-volume-health-monitor/kep.yaml
xing-yang: Not fixed?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1354 @@
...
+The `csi-external-health-monitor-controller` sidecar consumes the
+controller-side RPCs and writes to
+`PersistentVolumeClaim.Status.HealthStatus`. The kubelet consumes the
+node-side RPCs and writes to `Pod.Status.VolumeHealth` (per pod, per
+volume) and `CSINode.Status.StorageHealth` (per driver, per node). A
+single feature gate, `CSIVolumeHealth`, gates the kubelet and the
+apiserver.
+
+This is not the first attempt. An alpha for KEP-1432 shipped in
+Kubernetes v1.21 that overloaded `ListVolumes`, `ControllerGetVolume`,
```
xing-yang: overloaded -> embedded

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1354 @@
...
+in existing `ListVolumes`, `ControllerGetVolume`, and
+`NodeGetVolumeStats` RPCs and surfacing findings as Kubernetes
+events. That approach had three structural limitations that
+prevented graduation: health was coupled to stat and list RPCs
+that serve different purposes on different timelines;
+`NodeGetVolumeStats` can only be called for staged or published
+volumes, which excludes the most interesting unhealthy cases
+like corrupt filesystems or failed mounts; and events are
+ephemeral, unstructured, and cannot drive a remediation
+controller.
```
xing-yang, line 149: We should add this to the KEP:

"The original feature used only events and metrics because we didn't want to add first-class fields to PVCs and Pods that Kubernetes built-in controllers couldn't react to. However, experience showed this conservative approach was insufficient - events are ephemeral and metrics lack the structured, durable state that operators and automation need. This enhanced design represents our learning that well-designed status fields, following established Kubernetes patterns, provide essential value even when not directly consumed by core controllers."

Now we are making a step forward with the enhanced design, we should acknowledge what we have learned from the initial design.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Thanks. I have another local PV use case for DP operators.

```
Use case: Database operator with local PV resilience

A database operator manages stateful workloads using local persistent volumes for performance. It periodically checks pvc.status.healthStatus for storage accessibility. When a local disk fails, the CSI driver reports {status: Inaccessible, reason: DiskFailure}. If the volume remains inaccessible beyond a configured threshold (e.g., 5 minutes), the operator automatically:

1. Shuts down the database pod
2. Deletes the failed local PV to trigger provisioning of a new volume
3. Restores from backup and brings the database online

This structured health reporting enables precise, automated recovery that reduces downtime compared to relying on pod restart patterns or external monitoring alone.
```

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1354 @@
...
+which already runs alongside many CSI controllers, polls whichever is
+available and writes the result to
+`pvc.status.healthStatus`. Because the sidecar lives in the control
+plane, the cluster's existing trust boundary handles the question of
+who can write what to a PVC.
+
+The node plugin learns to answer two questions: "what's the condition
+of this volume on this node" (via `NodeGetVolumeHealth`) and "what's
+the condition of the storage backend from this node" (via
+`NodeGetStorageHealth`). The kubelet calls them and writes the
```
xing-yang, line 221: I think usually the controller side knows the condition of the entire storage system. Why is this information on the node side only?

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Can you add this? 
```
In cases where controller-side and node-side health reports conflict, the more severe health condition should take precedence, whether it originates from the controller or the node.
```

torredil: That rule existed because the old enhancement had the controller and the node both populating a single field, but with our current design we actually have a cleaner split between node side and controller side volume health reporting such that the conflict is naturally addressed and there is nothing to reconcile in the writers. Both reporting mechanisms write to different places (PVC for controller and pod for node) so its better if they are treated as independent sources of truth.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
gnufied: Sorry I forgot to push. This should be fixed now.

keps/sig-storage/1432-volume-health-monitor/kep.yaml
gnufied: same.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1354 @@
...
+    // VolumeHealthEntry.
+    // +optional
+    // +listType=map
+    // +listMapKey=status
+    // +listMapKey=reason
+    Conditions []VolumeHealthCondition `json:"conditions,omitempty"`
+
+    // lastProbeTime is the most recent time the CO obtained a
+    // response from the controller plugin.
+    LastProbeTime metav1.Time `json:"lastProbeTime,omitempty"`
```
deads2k: this did not age well when we did it with Node conditions because it produced a standing write load that scaled linearly with the size of the cluster.  Why would this be different?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1354 @@
...
+    // the CSI controller plugin. An empty list (or absence)
+    // means the controller plugin reports no adverse condition.
+    // Conditions are uniquely identified by the (status, reason)
+    // tuple, matching the CSI spec's uniqueness rule for
+    // VolumeHealthEntry.
+    // +optional
+    // +listType=map
+    // +listMapKey=status
+    // +listMapKey=reason
+    Conditions []VolumeHealthCondition `json:"conditions,omitempty"`
```
deads2k, line 667: Looks like we'd get similar results with a well-known condition.

deads2k:
PRR lgtm. Some comments on the API.

deads2k:
PRR

/approve

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I have updated the wording.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: So, which specific RPC call is being probed can be added as a label. I am not sure if we have to document the labels in alpha version of KEP. We will come to know exact metrics, when we get to implement it.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: ack. I am thinking for now, we will reuse the `VolumeStatsAggPeriod` for polling interval and have a hardcoded inertia for changing volume's health status.

gnufied: I am thinking this can be solved by not updating pvc or pod's health if `status` and `reason` does not changes  including not even updating `lastProbeTime`, so only timestamp in health is `lastTransitionTime`. 

I think we can do this because `Conditions` in health is not a merge, but actually straight replacement with whatever CSI driver reports.  So if what existss in current `Conditions` is different from incoming health status from CSI driver - then and only then `Conditions` is updated and then only then, `VolumeHealthStatus` changes.

This means that - if a lot of PVCs or Pods have failed for same reason, they don't have to be updated again and again. 

This does have a downside that, user only knows when something last changed, so if a PVC is reporting bad health for last 5 days (with same `status` and `reason`) then `lastTransitionTime` could be 5 days old. 

But this would reduce load on API server I think.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Yes, it does suffer from similar problems. I am currently leaning towards removing `lastProbeTime` entirely from health statuses, which would avoid this problem. See - https://github.com/kubernetes/enhancements/pull/6138#discussion_r3382743688

jsafrane: @deads2k do you have any link to scaling issues that node conditions introduced? Maybe not even related to lastProbeTime.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: In the old design, controller and node also report to different fields.
- If a PVC has an abnormal health condition, an event will be reported on the pod object using the PVC. 
- The controller watches volume health of the PVCs and reports events on the PVCs.
https://kubernetes.io/blog/2021/04/16/volume-health-monitoring-alpha-update/#what-is-volume-health-monitoring

So we have the same problem in the new design.

torredil: Thank you for clarifying Xing, +1 to adding more guidance for consumers when this conflict arises. The wording you suggested looks pretty good to me.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I think exact naming can be figured out during API reviews as well. We don't have to debate them here.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I think I have specified this clearly now.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: ack. will be fixed latest push.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: ack, will be fixed by next push.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: This has been added as a topic for future enhancement, but currently out of scope for this KEP.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I have removed `lastProbeTime` from API for now and added details about how this will be solved.

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1463 @@
...
+either or both. The `csi-external-health-monitor-controller` sidecar,
+which already runs alongside many CSI controllers, polls whichever is
+available and writes the result to
+`pvc.status.healthStatus`. Because the sidecar lives in the control
+plane, the cluster's existing trust boundary handles the question of
+who can write what to a PVC.
+
+The node plugin learns to answer two questions: "what's the condition
+of this volume on this node" (via `NodeGetVolumeHealth`) and "what's
+the condition of the storage backend from this node" (via
```
msau42, line 220: Is it useful for controller to have a storage backend health status?

keps/sig-storage/1432-volume-health-monitor/README.md
```diff
@@ -1,1003 +1,1463 @@
...
+runs whenever a volume's reported health changes, every driver
+reports on every volume it manages, and at scale this could amplify
+into a substantial PATCH rate against the apiserver. The design
+addresses this in two ways: both writers compute the desired
+`Conditions` list, compare it element-by-element to the on-disk
+value, and PATCH only on difference; in steady state, with no
+unhealthy volumes, the sustained PATCH rate is zero. If 
+a CSI driver reports same adverse condition for a volume 
+repeatedly no updates are written to API server because 
+we only store `lastTransitionTime`.
```
msau42, line 325: We could also do something similar to what kubelet does with node/pod status updates and have them queued.

msau42 (commented):
API changes lgtm

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I have added an example of automatic remediation for future enhancement but it is currently out of scope for this KEP.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: I have incorporated this one.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Yes, so this was brought up by @xing-yang as well as in CSI spec PR. I have added this as a future possible enhancement to this - https://github.com/gnufied/enhancements/tree/8b9982dda3688dcbe10fc27b4debd722107e5517/keps/sig-storage/1432-volume-health-monitor#future-enhancement

keps/sig-storage/1432-volume-health-monitor/kep.yaml
gnufied: This should match now.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: So none of the RPCs are removed right? We are only removing fields from certain response types. The deletion in CSI spec is specifically documented.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: We can add the wording but  we **do not define** in CSI spec which error is more severe. Is `DATA_LOSS` more sever than `INACCESSIBLE`?  The thing is - I do not think the conflict we are thinking exists when it comes to remediation (either manual or automatic). 

If controller reports a volume as `INACCESSIBLE` in general that means volume is inaccessible from everywhere. If node is reporting `DATA_LOSS` at the same time for same volume (for some reason), the scheduler and operator which consumes the storage have different remediation.  Scheduler will see volume as inaccessible and prevent pod scheduling.  Scheduler in general has no logic for preventing `DATA_LOSS` errors.  The operator which sees both errors, realizes that volume is both inaccessible and data is gone, so it needs to build new replica or restore from backup. 

If add wording that says:

> In cases where controller-side and node-side health reports conflict, the more severe health condition should take precedence, whether it originates from the controller or the node.

then we need to define what severity is for the errors, but at least in the current proposal we aren't defining severity of one error over the other. It is upto operator/component to decide which one they think is more severe.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: This was added.

keps/sig-storage/1432-volume-health-monitor/README.md
gnufied: Added wording about update queuing.

keps/sig-storage/1432-volume-health-monitor/README.md
xing-yang: Inaccessible is more severe than DataLoss which is more severe than Degraded. We can take a closer look at this during the CSI spec review.

PR #6140, opened 2026-06-01
Add note to notify SRC when KEP-5707 becomes default

tabbysable:
- One-line PR description: Add note to notify SRC when KEP-5707 becomes default

- Issue link: https://github.com/kubernetes/enhancements/issues/5707

- Other comments: When this KEP defaults to on, users' clusters will no longer be vulnerable to CVE-2020-8554 unless they specifically choose to re-enable support. At that point I consider the CVE to be fixed, and we should update the published CVE record to indicate that fact. Putting a little to-do note in here seems the best way to make sure we all remember to do so at the appropriate time.

### DIFF
--- a/keps/sig-network/5707-deprecate-service-externalips/README.md
+++ b/keps/sig-network/5707-deprecate-service-externalips/README.md
@@ -119,6 +119,7 @@ Tentative version: 1.36
 
 - Switch the `AllowServiceExternalIPs` feature gate to false
 - Create e2e test ensuring that externalIPs is no longer used by kube-proxy
+- Notify the Security Response Committee to update CVE-2020-8554 indicating that it now has a fix available.
 
 Clusters administrators can enable the feature gate if needed to restore the externalIPs functionality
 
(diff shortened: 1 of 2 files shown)


tabbysable:
cc @adrianmoisey

keps/sig-network/5707-deprecate-service-externalips/README.md
```diff
@@ -119,6 +119,7 @@ Tentative version: 1.36
 
 - Switch the `AllowServiceExternalIPs` feature gate to false
 - Create e2e test ensuring that externalIPs is no longer used by kube-proxy
+- Notify the Security Response Committee to update CVE-2020-8554 indicating that it now has a fix available.
```
adrianmoisey: This seems good to me, once merged I'll also add a an item to the description of https://github.com/kubernetes/enhancements/issues/5707

Another idea is to reword this to something like "update the published CVE record to indicate that CVE-2020-8554 is fixed" and add sig-security as a participating-sigs

danwinship:
/approve
i'll let Adrian lgtm

keps/sig-network/5707-deprecate-service-externalips/README.md
tabbysable: I like that idea and have pushed a commit implementing it.

Issue #5986, opened 2026-03-30
Expose cgroup v2 memory.events counters in CRI stats API


sohankunkerkar:
### Enhancement Description                                                                                                                                                
                                                                                                                                                                             
  - One-line enhancement description (can be used as a release note): Expose cgroup v2 `memory.events` counters via a new `MemoryEvents` message in CRI `MemoryUsage` stats for container memory debugging and Memory QoS observability                                                                             
  - Kubernetes Enhancement Proposal: <!-- link to KEP PR once created -->                                                                                                    
  - Discussion Link: Discussed in KEP-2570 (Memory QoS) review — Peter Hunt and Sascha Grunert requested this be split into a separate KEP (see kubernetes/kubernetes#137760)
  - PRs by stage and milestone:                                                                                                                                              
    - [ ] Alpha -                                                                                                                                                       
      - [ ] KEP (`k/enhancements`) update PR(s):                                                                                                                             
      - [ ] Code (`k/k`) update PR(s):                                                                                                                                       
      - [ ] Docs (`k/website`) update PR(s):                                                                                                                                 
                                                                                                                                                                             
  <!-- Uncomment these as you prepare the enhancement for the next stage                                                                                                     
  - [ ] Beta - v1.xx                                                                                                                                                         
    - [ ] KEP (`k/enhancements`) update PR(s):                                                                                                                               
    - [ ] Code (`k/k`) update PR(s):                                                                                                                                         
    - [ ] Docs (`k/website`) update(s):                                                                                                                                      
  - [ ] Stable - v1.xx                                                                                                                                                       
    - [ ] KEP (`k/enhancements`) update PR(s):                                                                                                                               
    - [ ] Code (`k/k`) update PR(s):                                                                                                                                         
    - [ ] Docs (`k/website`) update(s):                                                                                                                                      
  -->                                                                                                                                                                        
                                                                                                                                                                             
  _Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

sohankunkerkar:
cc @QiWang19

lasomethingsomething:
Hello @QiWang19 @sohankunkerkar 👋, 1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9th June 2026 (AoE) / Wednesday 10th June 2026, 12:00 UTC**.

[...]

PR #6141, opened 2026-06-02
KEP-5986: Per-container memory pressure eviction

sohankunkerkar:
- One-line PR description: Add KEP for per-container memory pressure detection using memory.events counters and PSI with two-stage remediation (relax memory.high, then evict)

- Issue link: https://github.com/kubernetes/enhancements/issues/5986

- Other comments: This KEP depends on KEP-2570 (MemoryQoS) which sets memory.high on container cgroups

### DIFF
--- a/keps/prod-readiness/sig-node/5986.yaml
+++ b/keps/prod-readiness/sig-node/5986.yaml
@@ -1,0 +1,3 @@
+kep-number: 5986
+alpha:
+  approver: "@kannon92"
--- a/keps/sig-node/5986-memory-high-eviction/README.md
+++ b/keps/sig-node/5986-memory-high-eviction/README.md
@@ -1,0 +1,947 @@
+# KEP-5986: Per-container memory pressure eviction
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Detection: Three Signals](#detection-three-signals)
+  - [Why Not a New Eviction Signal](#why-not-a-new-eviction-signal)
+  - [Response: Two-Stage Remediation](#response-two-stage-remediation)
+  - [Rate Calculation](#rate-calculation)
+  - [KubeletConfiguration](#kubeletconfiguration)
+  - [User Stories](#user-stories)
+    - [Perpetually Throttled Java Workload](#perpetually-throttled-java-workload)
+    - [Bursty ML Training Job](#bursty-ml-training-job)
+    - [memory.min Contention Across Pods](#memorymin-contention-across-pods)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+    - [Feature Interactions](#feature-interactions)
+    - [Kernel Version Considerations](#kernel-version-considerations)
+    - [CRI Runtime Requirements](#cri-runtime-requirements)
+    - [Platform Support](#platform-support)
+    - [No-Swap Scenario](#no-swap-scenario)
+    - [Eviction-Recreation Loop](#eviction-recreation-loop)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Feature Gate](#feature-gate)
+  - [Stats API Extension](#stats-api-extension)
+  - [Eviction Manager Integration](#eviction-manager-integration)
+    - [Stage 1 Dependencies](#stage-1-dependencies)
+    - [Detection](#detection)
+    - [Candidate Selection](#candidate-selection)
+    - [Eviction Observability](#eviction-observability)
+  - [Metrics](#metrics)
+  - [Prior Art](#prior-art)
+  - [Future: EvictionRequest API Integration](#future-evictionrequest-api-integration)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha (v1.37)](#alpha-v137)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+  - [Prior Work](#prior-work)
+- [FAQ](#faq)
+  - [Why are memory.events counters needed when PSI already detects pressure?](#why-are-memoryevents-counters-needed-when-psi-already-detects-pressure)
+  - [Why does stage 1 modify cgroup state from the eviction manager?](#why-does-stage-1-modify-cgroup-state-from-the-eviction-manager)
+  - [Why do all thresholds default to 0?](#why-do-all-thresholds-default-to-0)
+  - [How does the feature behave when MemoryQoS is disabled?](#how-does-the-feature-behave-when-memoryqos-is-disabled)
+  - [How is the eviction-recreation loop mitigated?](#how-is-the-eviction-recreation-loop-mitigated)
+  - [How does this feature co-exist with EvictionRequest API (KEP-4563)?](#how-does-this-feature-co-exist-with-evictionrequest-api-kep-4563)
+  - [How will VPA learn about memory pressure evictions?](#how-will-vpa-learn-about-memory-pressure-evictions)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [PSI-Only Detection (No memory.events)](#psi-only-detection-no-memoryevents)
+  - [Eviction-Only (No Stage 1 Remediation)](#eviction-only-no-stage-1-remediation)
+  - [New Eviction Signal in the Signal Framework](#new-eviction-signal-in-the-signal-framework)
+  - [systemd-oomd Integration](#systemd-oomd-integration)
+  - [Absolute Threshold Instead of Rate-Based](#absolute-threshold-instead-of-rate-based)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [x] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation, e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+When MemoryQoS ([KEP-2570](https://github.com/kubernetes/enhancements/issues/2570))
+sets `memory.high` on a container cgroup, the kernel throttles allocations
+that exceed the boundary. The kubelet currently has no way to detect this
+throttling or act on it. This KEP adds per-container memory pressure
+detection using `memory.events` counters and PSI, with a two-stage response:
+relax the `memory.high` boundary first, evict only if pressure persists.
+
+## Motivation
+
+cgroup v2 `memory.high` is designed to work with a userspace management
+agent that monitors throttling and takes action, either by granting more
+memory or terminating the workload
+([kernel docs](https://docs.kernel.org/admin-guide/cgroup-v2.html)).
+KEP-2570 (MemoryQoS) introduced `memory.high` on container cgroups to
+throttle memory overcommit, but the kubelet currently has no mechanism to
+detect this throttling or act on it.
+
+When a container exceeds `memory.high`, the kernel throttles its allocations
+and tries to reclaim memory. If the container stays between `memory.high`
+and `memory.max`, it gets perpetually throttled with no automatic recovery.
+Containers that repeatedly approach `memory.max` see severe latency spikes
+but never quite trigger OOM kill.
+When multiple pods have `memory.low` set and total reserved memory exceeds
+what the node has available, the kernel reclaims from the weakest pods,
+causing sustained pressure that `memory.events:high` alone cannot detect.
+
+The kernel tracks all of these via `memory.events` counters and PSI
+(Pressure Stall Information). cadvisor already reads both. This KEP uses
+them to detect sustained pressure and respond.
+
+The existing kubelet eviction manager monitors node-level resource pressure
+(memory.available, disk, PIDs) and evicts pods when the node is under stress.
+All nine existing eviction signals are quantity-based (capacity/available),
+measuring absolute resource levels. Per-container throttle rate and pressure
+percentage are different signal types that do not fit this model.
+This design gap was discussed in
+[KEP-2570](https://github.com/kubernetes/enhancements/issues/2570#issuecomment-3964464265)
+and confirmed as requiring a follow-up KEP.
+
+### Goals
+
+- Detect per-container sustained memory pressure using three kernel signals:
+  `memory.events:high` rate, `memory.events:max` rate, and PSI
+  `memory.pressure` avg10
+- Respond with two-stage remediation: relax the `memory.high` throttle
+  boundary first, evict only if pressure persists
+- Provide configurable per-signal thresholds and grace period via
+  `KubeletConfiguration`
+- Emit metrics and pod conditions for operator observability
+
+### Non-Goals
+
+- PSI-based node-level protection (system-wide eviction trigger, planned
+  for beta)
+- Multi-signal weighted scoring (simple OR at alpha, weighted combination
+  at beta)
+- Per-pod opt-out of memory pressure eviction (beta consideration)
+- Direct systemd-oomd integration (at beta, node-level PSI eviction
+  provides equivalent functionality within the kubelet)
+- New node condition for memory.high pressure (this is per-container, not
+  node-level; `NodeMemoryPressure` already covers node-level)
+
+## Proposal
+
+Add a per-container memory pressure check to the kubelet eviction manager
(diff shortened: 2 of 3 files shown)


sohankunkerkar:
cc @haircommander @QiWang19

keps/prod-readiness/sig-node/5986.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 5986
+alpha:
+  approver:
```
kannon92:
```suggestion
  approver: @kannon92
```

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+  - [New Eviction Signal in the Signal Framework](#new-eviction-signal-in-the-signal-framework)
+  - [systemd-oomd Integration](#systemd-oomd-integration)
+  - [Absolute Threshold Instead of Rate-Based](#absolute-threshold-instead-of-rate-based)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
```
kannon92: Please make sure to fill this out as you complete it.

keps/sig-node/5986-memory-high-eviction/kep.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - "@SergeyKanzhelev"
+stage: alpha
+latest-milestone: "v1.37"
+milestone:
+  alpha: "v1.37"
+feature-gates:
+  - name: MemoryHighEviction
+    components:
+      - kubelet
+disable-supported: true
```
kannon92, line 22: You are missing the metrics field for this.

kannon92: I think you are suggesting to add new metrics if I understand your KEP correctly.

kannon92:
> **Disclaimer: This is an AI-assisted review of the PRR (Production Readiness Review) for alpha requirements. It should not be treated as a substitute for a human PRR approver's review.**

## PRR Alpha Review Summary — KEP-5986: Per-container memory pressure eviction

**Feature Gate:** `MemoryHighEviction` (kubelet, default OFF)
**Stage:** Alpha, targeting v1.37

### Alpha Requirements: All Pass

| Area | Status |
|---|---|
| `kep.yaml` fields (`feature-gates`, `components`, `disable-supported`) | :white_check_mark: |
| Feature Enablement and Rollback (all 5 questions answered) | :white_check_mark: |
| Graduation Criteria (alpha, beta, GA defined) | :white_check_mark: |
| Test Plan (unit + e2e_node) | :white_check_mark: |
| Scalability (encouraged at alpha, all questions answered) | :white_check_mark: |

### Strengths

- **Double-gating design**: Even with the feature gate ON, all thresholds default to 0 (no-op). Operators must explicitly configure thresholds to activate detection/eviction. This is an excellent safety pattern.
- **Stage 1 remediation is non-destructive**: Relaxing `memory.high` to `memory.max` returns to pre-MemoryQoS behavior. The hard limit is unchanged. This avoids unnecessary evictions.
- **MemoryQoS dependency is well-documented**: `memory.events` signals require MemoryQoS; PSI works independently. A validation warning is emitted when thresholds are set without MemoryQoS enabled.
- **Beta/GA sections answered early**: Rollout/Rollback, Monitoring, Dependencies, and Troubleshooting sections are filled in with substantive answers even though they are only required at beta.
- **Comprehensive feature interaction table**: Covers MemoryQoS, InPlacePodVerticalScaling, VPA, QoS classes, pod priority, sidecars, container restart, EvictionRequest API, PDB, DRA, swap, and more.

### Minor Items to Address

1. **Release Signoff Checklist**: The "KEP approvers have approved the KEP status as `implementable`" checkbox is unchecked (`README.md` line 80). Should be checked once approvers confirm.
2. **PRR checkboxes**: "Production readiness review completed" and "Production readiness review approved" are unchecked (`README.md` lines 88-89). Should be checked once PRR is approved.
3. **`metrics` field in `kep.yaml`**: Not required until beta, but the KEP defines two metrics (`kubelet_evictions{eviction_signal="memory.high.pressure"}` and `kubelet_memory_high_relaxed_total`) in the README. Consider adding them to `kep.yaml` proactively.

### Known Limitations (Acceptable for Alpha)

- **Eviction-recreation loop**: Acknowledged with reasonable mitigation (grace period bounds loop frequency, stage 1 reduces eviction rate, VPA integration planned for beta).
- **No PDB support**: Kubelet eviction does not check PDBs. EvictionRequest API integration at beta will address this.
- **No production-validated default thresholds**: All thresholds default to 0 (disabled). Data-driven defaults planned for beta based on alpha feedback.

### Verdict

The KEP meets all alpha PRR requirements. The design is well-reasoned with strong safety guarantees for rollback and incremental adoption.

kannon92:
Really nice PRR work for alpha.

https://github.com/kubernetes/enhancements/pull/6141#discussion_r3357365125

Very minor nit.

/approve

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+  two-stage remediation.
+- Three detection signals and four configuration fields increase operator
+  cognitive load.
+- Rate-based and percentage-based detection is new to the eviction
+  framework since all existing signals are quantity-based.
+- Requires MemoryQoS as a prerequisite for `memory.events` signals.
+- Stage 1 (relaxing `memory.high`) modifies container cgroup state, unlike
+  existing eviction which is read-only until the kill.
+
+## Alternatives
```
haircommander, line 899: an alternative @kannon92 hinted at offline could be an external eviction controller that triggers an api level eviction. that may be too slow, so another alternative could be an eviction endpoint in the kubelet that the controller would hook into. I think structurally eviction code is a bit brittle and I do think it's worth considering whether want to continue to extend in-tree. Can you put an agenda item in sig node next week to discuss @sohankunkerkar ?

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+On each cycle, the eviction manager reads per-container
+`memory.events` counters and PSI data from the existing stats pipeline,
+computes rates for `memory.events` counters, and compares all three signals
+against their configured thresholds. If any signal exceeds its threshold
+and the grace period has elapsed, the two-stage response triggers.
+
+#### Stage 1 Dependencies
+
+Stage 1 needs to call CRI `UpdateContainerResources` to relax
+`memory.high`. The eviction manager does not have direct CRI access. A
```
haircommander, line 433: I kind of think this should be tiered similar to memory protection. a guaranteed pod or high priority pod should have its memory.high increased, but a best-effort probably shouldn't. further, how do we make a signal to VPA that the memory limit of the pod should increase as well?

haircommander: https://github.com/kubernetes/enhancements/pull/6141/changes#diff-afe656367ee3f1986c64f7208b9410337adee92ae1cfbedeac7372c046efab1eR459 hm jk

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+  a 10-second window. It is compared directly against the threshold.
+
+#### Candidate Selection
+
+When multiple pods exceed the grace period, candidates are sorted by
+QoS class (BestEffort before Burstable), then by pod priority (lowest
+first), then by the highest ratio of any active signal value to its
+threshold across the pod's containers. Guaranteed pods are naturally
+exempt since MemoryQoS does not set `memory.high` on them. Mirror pods
+and static pods are skipped.
```
haircommander: i think this is more accurate to say static pods are skipped. mirror pods are just an api representation of static pods

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+
+#### Eviction Observability
+
+Evicted pods receive:
+- A `DisruptionTarget` pod condition with reason `TerminationByKubelet`
+- A warning event with reason `Evicted` and a message indicating which
+  container triggered the eviction and which signal was exceeded (e.g.,
+  "Container foo exceeded memory.high throttle threshold")
+- Annotations: `OffendingContainersKey` (throttled container name) and
```
haircommander, line 469: annotations on what? the pod? that seems odd imo. probably something in status would be better

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+
+### Future: EvictionRequest API Integration
+
+[KEP-4563 (EvictionRequest API)](https://github.com/kubernetes/enhancements/issues/4563)
+is also targeting v1.37 alpha. At alpha, both features ship independently:
+this feature uses the kubelet's pod eviction path directly, same as all
+evictions. Coupling two alpha features would create fragility since
+EvictionRequest's API may change before beta.
+
+At beta, the stage 2 eviction step routes through EvictionRequest, giving
```
haircommander, line 507: who routes? the kubelet? seems at risk of timing issues where the kernel may act too fast

haircommander: also this is another reason to me to have this managed by an out of tree controller. tbh

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,935 @@
...
+
+On each cycle, the eviction manager reads per-container
+`memory.events` counters and PSI data from the existing stats pipeline,
+computes rates for `memory.events` counters, and compares all three signals
+against their configured thresholds. If any signal exceeds its threshold
+and the grace period has elapsed, the two-stage response triggers.
+
+#### Stage 1 Dependencies
+
+Stage 1 needs to call CRI `UpdateContainerResources` to relax
```
haircommander, line 432: btw a similar thing could be achieved by just increasing memory limit. then the memory.high would be increased (right?) and the overall goal of making the container less close to OOM would be acheived

sohankunkerkar: I think stage 1 only removes `memory.high`, which is a throttle boundary the kubelet set via MemoryQoS. `memory.max` stays unchanged, pod spec stays unchanged. Increasing the memory limit is an API-level operation that changes `memory.max`, scheduling, and quota. The eviction manager has no API client to do that, and shouldn't. Stage 1 is scoped to undo what the kubelet did. If the workload stabilizes below `memory.max`, no eviction needed.

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: Guaranteed pods are exempt (no memory.high set). QoS-tiered stage 1 policy and VPA signal are not in alpha today, but good candidates for beta refinement.

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: good catch, will update that!

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: These are event annotations (`AnnotatedEventf`), not pod annotations. They're the same schema the node-pressure eviction path uses (`OffendingContainersKey`, `StarvedResourceKey`). The primary status signal is the `DisruptionTarget` pod condition.

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: At alpha, no `EvictionRequest` routing. Stage 2 uses the same direct kubelet eviction path as all existing evictions. On timing, `memory.high` throttles but doesn't kill. The kernel only kills at memory.max/OOM, which exists with or without this feature.

sohankunkerkar: >also this is another reason to me to have this managed by an out of tree controller. tbh

Even with external detection, stage 1 still needs a privileged node-local actor with CRI access for `UpdateContainerResources`. Stage 2 via API adds a control-plane hop. So the controller becomes another      privileged node-local resource manager with its own stats pipeline and cgroup writes.

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: Agree, worth discussing at sig-node. No kubelet eviction endpoint exists today  as building one needs its own design. An external controller still needs a node-local actor for stage 1 CRI writes, so it's a multi-KEP architecture split. This KEP follows the existing `localStorageEviction` parallel check pattern. It can proceed at alpha while the broader architecture discussion happens, but the two are orthogonal.

kannon92: If we draw parallels to localStorageEviction, should there be a pod limit (like ephemeral-storage)?

sohankunkerkar: Actually, memory.high already is the per-container limit. MemoryQoS computes it from the pod spec's memory request and limit and sets it on the cgroup. `localStorageEviction` needs ephemeral-storage limits because there's no kernel enforcement. Here the kernel enforcement (memory.high throttling) already exists. This KEP adds detection and remediation for when that throttling becomes sustained.

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,937 @@
...
+
+Evicted pods receive:
+- A `DisruptionTarget` pod condition with reason `TerminationByKubelet`
+- A warning event with reason `Evicted` and a message indicating which
+  container triggered the eviction and which signal was exceeded (e.g.,
+  "Container foo exceeded memory.high throttle threshold")
+- Annotations: `OffendingContainersKey` (throttled container name) and
+  `StarvedResourceKey` ("memory")
+
+Stage 1 emits a warning event indicating that `memory.high` relaxation was
```
HirazawaUi: Should we consider adding a condition or a status field to indicate that a Pod is under high memory pressure after the first stage is triggered but before eviction? This could allow automated operators to trigger a Pod resize, especially since events are not always reliable.

keps/sig-node/5986-memory-high-eviction/README.md
```diff
@@ -0,0 +1,937 @@
...
+kubelet config validation rejects them at startup. If MemoryQoS is OFF,
+`memory.events` signals are no-ops (PSI still works independently). If the
+container runtime does not support stage 1 updates, detection and stage 2
+still function normally.
+
+Impact on running workloads: only containers under sustained pressure above
+the configured thresholds are affected. Containers operating normally are
+never touched. Stage 1 (relaxing `memory.high`) is non-destructive.
+
+Rollback: disable the gate, restart kubelet. Immediate, minimal cleanup
```
HirazawaUi: Could you please clarify what is meant by "minimal cleanup"? A more detailed explanation would be very helpful. Thanks!

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: Good point! I added a `ContainerMemoryPressure` pod condition at stage 1. It's queryable and watchable, unlike events which can be missed.

keps/sig-node/5986-memory-high-eviction/README.md
sohankunkerkar: Sure, added a few lines to clarify this part.

keps/sig-node/5986-memory-high-eviction/README.md
haircommander: BTW an advantage to an external controller is better handling of https://github.com/kubernetes/enhancements/pull/6141/changes#r3358885558. In fact, something like KEDA could be close to being able to do this today. it may not have the knobs to trigger an api eviction, but that may be an easier extension in the grand scheme of changes, and it could better handle scaling up important pods, vs triggering an eviciton for less important pods

haircommander:
Some pieces of this feel a bit funky to me (changing eviction mgr semantics) but an overarching worry I have is this is actually possible with an external controller after we get eviction request API (or even today with triggering a graceful termination of a pod). An external controller could use the existing metrics exposed (like KEDA does today) and:
- for high priority/guaranteed pods, use IPPR to increase the memory limit, thus increasing the memory.high
- for low priority/burstable/best effort pods, either evict or terminate the pod if the containers are continually disobeying their memory.high.

I'm not currently seeing a reason why kubelet needs to be in charge of this. the best answer I have is it'd be faster, and thus potentially lower the likelihood the kernel steps in, but I'd want to see that  in practice before feeling certain that's enough of a problem to warrant changing the existing eviction semantics to keep state like this is proposing

we have an item on sig node agenda tomorrow, i'll voice my concerns there as well. Sorry for not catching this earlier, I initially +1'd you opening it because I didn't think thoroughly enough about it

sohankunkerkar:
> Some pieces of this feel a bit funky to me (changing eviction mgr semantics) but an overarching worry I have is this is actually possible with an external controller after we get eviction request API (or even today with triggering a graceful termination of a pod). An external controller could use the existing metrics exposed (like KEDA does today) and:
> 
> * for high priority/guaranteed pods, use IPPR to increase the memory limit, thus increasing the memory.high
> * for low priority/burstable/best effort pods, either evict or terminate the pod if the containers are continually disobeying their memory.high.
> 
> I'm not currently seeing a reason why kubelet needs to be in charge of this. the best answer I have is it'd be faster, and thus potentially lower the likelihood the kernel steps in, but I'd want to see that in practice before feeling certain that's enough of a problem to warrant changing the existing eviction semantics to keep state like this is proposing
> 
> we have an item on sig node agenda tomorrow, i'll voice my concerns there as well. Sorry for not catching this earlier, I initially +1'd you opening it because I didn't think thoroughly enough about it


Thanks for the thoughtful feedback Peter, and no worries about the timing. This is exactly the kind of discussion that helps make KEPs better. I have a few thoughts I wanted to share below:

**On the eviction manager semantics concern**: I agree this is different from the current signal-based eviction flow because it introduces stateful tracking across sync cycles and grace periods. I’ve been prototyping an `evictionCheck` interface to make these checks a first-class extension point in `synchronize()`, similar to how `localStorageEviction` already works today. The idea is to keep this logic cleanly separated instead of mixing it into the existing signal-based path.

**On the external controller idea**: I agree stage 2 (eviction/termination) could already be implemented externally using graceful termination. For the detection + stage 1 remediation side though, I could not find an existing controller that covers this. KEDA is horizontal only, Kedify PRA supports in-place resize but based on utilization thresholds rather than memory.events + PSI throttle rates, and VPA does not react to real-time throttling. So the controller would still need to be designed and built for this use case. The kubelet already has the stats pipeline, CRI access, and eviction path, so for alpha it seemed like the shorter path to validating the behavior.

**On IPPR vs stage 1**: IPPR being GA definitely helps, but it is solving a different problem. IPPR changes pod resources through the API, goes through admission/quota flows, and persists as part of pod state. Stage 1 is intentionally temporary and local to the node. It only writes `memory.high=max` through CRI and reverts on restart without changing pod spec or quotas. Also, for Guaranteed containers, kubelet skips `memory.high` entirely (request == limit), so IPPR wouldn't change memory.high for those.

**On latency**: I agree I do not yet have production data showing the latency difference is important. That is a fair point. My thinking is that alpha is the right place to validate this. We can measure how quickly the in-tree path reacts to throttling compared to what an external scrape/reconcile loop could realistically achieve.

**On graduation**: the current KEP already plans to integrate `EvictionRequest` at beta for PDB-aware eviction, and potentially use `InPlacePodResize` at GA as a “resize before evict” path. So the longer-term direction is compatible with the controller-based ideas you mentioned as well.

I think the main question is whether we want to ship the in-tree detection + remediation path behind a feature gate for alpha now, or wait for a separate external controller design and implementation first.

haircommander:
>  So the controller would still need to be designed and built for this use case. The kubelet already has the stats pipeline, CRI access, and eviction path, so for alpha it seemed like the shorter path to validating the behavior.

that's fine, it's not terribly hard to write a controller. doing so externally speeds up development time and reduces the restriction to tie with k8s releases.

>  It only writes memory.high=max through CRI and reverts on restart without changing pod spec or quotas. Also, for Guaranteed containers, kubelet skips memory.high entirely (request == limit), so IPPR wouldn't change memory.high for those.

good point, but the point still stands that burstable pods could be resized and get a different memory.high value. I think it's a bit weird to have some pods within the same qos class have memory.high and some not, especially because a restart will cause it to flap. better to persist in the api so a restart brings back to the old allocation. This is extra reason to me to go through api rather than sneak behind it

sohankunkerkar:
After discussing with @haircommander ,   we're pursuing an external operator for detection and remediation. The operator uses cadvisor pressure metrics (memory.events:high, PSI) for detection and InPlacePodResize for the first remediation step (raise the memory limit so kubelet recomputes memory.high). When IPPR doesn't resolve pressure (max attempts reached, admission rejection, or node capacity), the operator will escalate via the EvictionRequest API (KEP-4563) once it graduates, providing PDB-aware graceful eviction.

This covers the core detection + remediation flow but outside the kubelet. The stats API extension and ContainerMemoryPressure pod condition from this KEP can be revisited as separate proposals if needed. We'll evaluate the latency tradeoff with real numbers from the prototype and share results. The KEP stays open for now. If the external approach proves insufficient, we'll revisit with evidence.

ndixita:
I read the user agent proposal. It makes me think if we should allow setting memory.high opt-in for both burstable and guaranteed workloads. The reason being, it might only help guaranteed workloads (some of the usecase - will try to think of real examples and get back) to not be oom killed and rather throttle it and let the user agent act on the workload and scale it up as needed and then relax memory.high accordingly.

PR #6143, opened 2026-06-03
KEP-2570: update MemoryQoS for beta graduation in v1.37

QiWang19:
- One-line PR description: update MemoryQoS for beta graduation in v1.37

- Issue link: #2570 

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-node/2570.yaml
+++ b/keps/prod-readiness/sig-node/2570.yaml
@@ -2,4 +2,4 @@ kep-number: 2570
 alpha:
   approver: "@johnbelamaric"
 beta:
-  approver: "@johnbelamaric"
+  approver: "@kannon92"
--- a/keps/sig-node/2570-memory-qos/README.md
+++ b/keps/sig-node/2570-memory-qos/README.md
@@ -2,7 +2,7 @@
 <!-- toc -->
 - [Release Signoff Checklist](#release-signoff-checklist)
 - [Latest Update](#latest-update)
-  - [Why Alpha v3 Instead of Beta](#why-alpha-v3-instead-of-beta)
+  - [Previous Status (v1.36)](#previous-status-v136)
   - [Previous Status (v1.28)](#previous-status-v128)
 - [Summary](#summary)
 - [Motivation](#motivation)
@@ -13,6 +13,7 @@
     - [Alpha v1.27](#alpha-v127)
     - [Beta v1.28 - Cancelled](#beta-v128---cancelled)
     - [Alpha v1.36](#alpha-v136)
+    - [Beta v1.37](#beta-v137)
   - [User Stories (Optional)](#user-stories-optional)
     - [Memory Sensitive Workload](#memory-sensitive-workload)
     - [Node Availability](#node-availability)
@@ -68,17 +69,17 @@
 - [x] (R) Production readiness review completed
 - [x] (R) Production readiness review approved
 - [x] "Implementation History" section is up-to-date for milestone
-- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
-- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+- [x] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [x] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
 
 ## Latest Update
 
-Targeting Alpha-3 in v1.36. The livelock issue that blocked Beta in v1.28 has been resolved—kernel 5.9+ contains a fix ([commit b3ff929](https://github.com/torvalds/linux/commit/b3ff92916af3b458712110bb83976a23471c12fa)) that prevents indefinite throttling at `memory.high`. See [Alpha v1.36](#alpha-v136) for details.
+Targeting Beta in v1.37. Rollback cleanup is now fully implemented. Cgroup v2 memory knobs (`memory.min`, `memory.low`, `memory.high`) are properly cleared when the MemoryQoS feature gate is disabled. Node e2e tests cover memory protection, throttling, and rollback cleanup. Benchmark testing validates memory.high throttle behavior, tiered memory protection, and rollback safety. See [Beta v1.37](#beta-v137) for details.
 
-### Why Alpha v3 Instead of Beta
+### Previous Status (v1.36)
 
-This release is Alpha v3 rather than Beta based on PRR feedback:
+v1.36 shipped as Alpha v3 rather than Beta based on PRR feedback:
 
 1. **New KubeletConfiguration field requires Alpha iteration**: Adding `memoryReservationPolicy` to KubeletConfiguration is an API-level change. Per Kubernetes policy, features adding API fields must start in Alpha. This field provides independent control over `memory.min` protection, addressing concerns about validating memory.min behavior separately.
 
@@ -366,6 +367,21 @@ Linux kernel 5.9+ is required for correct `memory.high` behavior. See [Prerequis
 | CRI-O | 1.22 |
 
 
+#### Beta v1.37
+**Status:** Beta graduation criteria met. Rollback cleanup is fully implemented. Benchmark testing validates node stability under sustained memory pressure. No regressions reported from Alpha v3 (v1.36) users.
+
+**Changes from Alpha v1.36:**
+- No changes to memory.high formula or `memoryReservationPolicy` mapping
+- Rollback cleanup fully implemented: when the MemoryQoS feature gate is disabled, cgroup v2 memory knobs are cleaned up as follows:
+  - `memory.min` on the kubepods root cgroup is reset to 0 at kubelet startup ([kubernetes/kubernetes#138903](https://github.com/kubernetes/kubernetes/pull/138903))
+  - `memory.low` on the Burstable QoS cgroup is reset to 0 at kubelet startup ([kubernetes/kubernetes#138903](https://github.com/kubernetes/kubernetes/pull/138903))
+  - Per-container `memory.high` is set to `max` in the container resource config so that it is cleared when the container runtime applies the config. This takes effect on newly created containers, restarted containers, and existing containers updated via InPlacePodResize ([KEP-1287](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/1287-in-place-update-pod-resources/README.md)). For already-running containers with no restart or resize, stale `memory.high` values persist until the next container restart or InPlacePodResize update ([kubernetes/kubernetes#139377](https://github.com/kubernetes/kubernetes/pull/139377))
+- Rollback e2e test re-enabled (previously skipped due to systemd unit property side effects, see [kubernetes/kubernetes#138485](https://github.com/kubernetes/kubernetes/pull/138485))
+- Pod-level and container-level `memory.min`/`memory.low` values persist after feature disable but are effectively neutralized because parent cgroup protection is set to 0 (cgroup v2 memory protection is hierarchical — parent=0 wins)
+- Benchmark testing:
+  - Alpha v1.36: tiered reservation, `memory.high` throttle behavior, and kubelet overhead ([benchmark report](https://github.com/sohankunkerkar/kep-2570-memoryqos-benchmarks))
+  - Rollback safety and BestEffort `memory.high` fix validation ([beta benchmark report](https://github.com/QiWang19/kep-2570-memoryqos-benchmarks/blob/rollback-validation/README-rerun.md))
+
 ### User Stories (Optional)
 #### Memory Sensitive Workload
 Some workloads are sensitive to memory allocation and availability, slight delays may cause service outage. In this case, a mechanism is needed to ensure the quality of memory.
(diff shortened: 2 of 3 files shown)


QiWang19:
@sohankunkerkar PTAL

keps/sig-node/2570-memory-qos/kep.yaml
sohankunkerkar, line 34: You might need to add the metrics field here. 
```yaml
  metrics:
    - kubelet_memory_qos_node_memory_min_bytes
    - kubelet_memory_qos_node_memory_low_bytes
    - container_memory_events_high_total
    - container_memory_events_max_total

```

keps/sig-node/2570-memory-qos/README.md
sohankunkerkar, line 863: We need to also add `container_memory_events_high_total` (cadvisor PR google/cadvisor#3870). That metric tracks per-container `memory.high` throttle events. It's the primary signal for detecting memory.high degradation. And `container_memory_events_max_total` as well.

sohankunkerkar:
You need to also update https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/sig-node/2570.yaml

kannon92:
https://testgrid.k8s.io/sig-node-cri-o#ci-node-crio-swap-serial

I see this test perm failing on a few lanes right now.

```
E2eNode Suite: [It] [sig-node] MemoryQoS feature rollback should clear stale memory.high when MemoryQoS is disabled and container is resized
```

sohankunkerkar:
> https://testgrid.k8s.io/sig-node-cri-o#ci-node-crio-swap-serial
> 
> I see this test perm failing on a few lanes right now.
> 
> ```
> E2eNode Suite: [It] [sig-node] MemoryQoS feature rollback should clear stale memory.high when MemoryQoS is disabled and container is resized
> ```

The CI lane's CRI-O doesn't have cri-o/cri-o#9820, which added Unified map support to `UpdateContainerResources`. Without that, the resize path silently ignores the memory.high update. Let me go ahead and update the crio version there.

Here's the cri-o version update PR: https://github.com/kubernetes/test-infra/pull/37196

kannon92 (commented):
out of diff:

https://github.com/QiWang19/kubernetes-enhancements/blob/14bfa279cce47b6a2549d9b2c7c9232fa83897a7/keps/sig-node/2570-memory-qos/README.md#e2e-tests

Please add links to the e2e tests you have for MemoryQoS.

kannon92:
**AI Summary**
## PRR Review: KEP-2570 Memory QoS — Beta v1.37

**Reviewer:** @kannon92 (PRR approver per `keps/prod-readiness/sig-node/2570.yaml`)

### Overview

The PRR questionnaire is thorough and well-written overall. The rollback cleanup story is particularly strong — the hierarchical cgroup neutralization approach (parent=0 wins) is a clean solution for the memory.min/memory.low persistence gap. The failure modes section is comprehensive with 6 documented scenarios including detection, mitigation, and diagnostics for each.

### Blocking Issues

**1. e2e Tests — Missing links (required for Beta)**
The e2e test section still has Alpha-level language ("As part of alpha, we plan to add..."). For Beta, the template requires:
- Permalink(s) to the test source code in kubernetes/kubernetes
- Link to the periodic testgrid job filtered by test name
- Link to k8s-triage search

Please update with links in the format:
```
- [TestName](https://github.com/kubernetes/kubernetes/blob/.../test/e2e_node/...): [testgrid](https://testgrid.k8s.io/...), [triage](https://storage.googleapis.com/k8s-triage/index.html?test=MemoryQoS)
```

**2. Integration Tests — Needs Beta-level justification**
Current text is `n/a: plan to use node e2e tests`. For Beta, please either provide links to integration tests or update the justification to explicitly reference the e2e tests (with links from issue 1) explaining why node e2e tests provide sufficient coverage.

### Non-blocking Items

**3. Production feedback graduation criterion not yet met**
Beta Graduation Criteria includes "Production feedback from Alpha v3 users confirms no regressions" but the Beta v1.37 status says collection is "in progress." Please either document the feedback received or adjust the criterion.

**4. Release Signoff Checklist**
User-facing documentation and supporting documentation items are unchecked. Please confirm these are tracked for the release.

### Sections Reviewed — All Complete ✅
- Feature Enablement and Rollback
- Rollout, Upgrade and Rollback Planning
- Monitoring Requirements
- Dependencies
- Scalability
- Troubleshooting
- kep.yaml (feature-gates, metrics, disable-supported)

**Status: Pending — please address blocking issues 1 and 2, then ping for re-review.**

keps/prod-readiness/sig-node/2570.yaml
```diff
@@ -2,4 +2,4 @@ kep-number: 2570
 alpha:
   approver: "@johnbelamaric"
 beta:
-  approver: "@johnbelamaric"
\ No newline at end of file
+  approver:
```
kannon92:
```suggestion
  approver: kannon92
```

sohankunkerkar:
@kannon92 could you PTAL?

kannon92 (commented):
approve from PRR side.
/approve

keps/sig-node/2570-memory-qos/README.md
```diff
@@ -366,6 +367,21 @@ Linux kernel 5.9+ is required for correct `memory.high` behavior. See [Prerequis
 | CRI-O | 1.22 |
 
 
+#### Beta v1.37
+**Status:** Beta graduation criteria met. Rollback cleanup is fully implemented. Benchmark testing validates node stability under sustained memory pressure. No regressions reported from Alpha v3 (v1.36) users.
```
ndixita, line 371: Do we have data around how would system workloads vs guaranteed pods behave if there's memory contention?

QiWang19: We have multi-pod pressure test data: https://github.com/QiWang19/kep-2570-memoryqos-benchmarks/blob/rollback-validation/README-rerun.md#5-multi-pod-memory-protection-under-pressure-tieredreservation-on
A Burstable aggressor ramped to ~13 GiB. `memory.high` throttling contained it within its cgroup, the node remained `Ready` and kubelet was responsive throughout the test, so no system-level memory pressure was triggered. The Guaranteed pod also remained stable.

And system workloads (kubelet, CRI-O) can be protected by the same kernel mechanism. When kubelet is configured with --system-reserved and --kube-reserved with --enforce-node-allocatable=kube-reserved,system-reserved, kubelet sets memory.min on the system cgroup slices.

QiWang19:
@ndixita @haircommander PTAL

keps/sig-node/2570-memory-qos/README.md
```diff
@@ -402,11 +418,12 @@ This might be a good place to talk about core concepts and how they relate.
...
+- **TieredReservation and Guaranteed pods with page-cache-heavy workloads**: When `memoryReservationPolicy: TieredReservation` is configured, Guaranteed pods have `memory.min` set equal to `memory.max` (since requests = limits). Per the [kernel cgroup v2 documentation](https://docs.kernel.org/admin-guide/cgroup-v2.html#memory-interface-files), `memory.min` specifies memory that "can never be reclaimed by the system" and "the cgroup's memory won't be reclaimed under any conditions." When `memory.min` equals `memory.max`, the kernel cannot reclaim page cache within the cgroup to make room for new allocations, causing OOM kills when the cgroup reaches `memory.max`. This does not affect Burstable pods because `memory.low` is soft protection and the kernel still reclaims page cache within the cgroup when it approaches `memory.max`. With the default `memoryReservationPolicy: None` (v1.36+), `memory.min` is set to 0 for all pods, so page cache is freely reclaimable and no additional OOM risk is introduced. Users who opt into `TieredReservation` should ensure Guaranteed pods with page-cache-heavy workloads size their memory limits to include headroom for page cache.
```
haircommander, line 426: this point is concerning. it feels like maybe memory.min should be some factor of memory.max to make sure the kernel has something to reclaim? I think this can be covered by normal bug fixing procedures though

keps/sig-node/2570-memory-qos/README.md
haircommander: still seems undone

haircommander (approved):
LGTM, couple of notes but I do feel the alpha developments have made us ready for beta

keps/sig-node/2570-memory-qos/README.md
QiWang19: Agree the concern is valid. Since `memoryReservationPolicy` is not turn on by default and the existing mitigation  `memory.low`, it can be a follow-up bug fix (https://github.com/kubernetes/kubernetes/issues/137880)
~~Fractional memory.min may not fully address this since Guaranteed pods don't have memory.high set to trigger reclaim, we will explore the right approach in the bug fix.~~
Tested a fix https://github.com/kubernetes/kubernetes/pull/139771, memory.min = 460 Mi (fractional, ~90% of 512Mi), memory.max = 512 Mi, pod survived, no oomkilled. it seems memory.min < memory.max for Guaranteed pods gives the kernel room to reclaim page cache.

keps/sig-node/2570-memory-qos/README.md
QiWang19: @haircommander (cadvisor PR https://github.com/google/cadvisor/pull/3870) was merged, and `container_memory_events_high_total`, `container_memory_events_max_total` are included in the metrics section. 
Could you clarify what's undone here?

haircommander: doesn't seem to be in https://github.com/kubernetes/enhancements/pull/6143/changes#diff-721cae6c37b35b64d9aa33348c94cffd855ebe11dfe72c4619e5fc066feda3ecR864-R881 which is not a huge deal

QiWang19:
/assign @mrunalp 
Could you approve?

mrunalp (approved):

petr-muller:

keps/sig-node/2570-memory-qos/README.md
```diff
@@ -427,8 +444,8 @@ The main risk of this proposal is throttling applications too early.
 We intend to mitigate this by (1) setting a `memory.high` that is closer to the
 limit and (2) only throttling when usage > request.
 
-**Blast Radius Alpha v1.36:** in Alpha v1.36, the `MemoryQoS ` feature gate is disabled by default. There is no impact on existing clusters unless the feature is explicitly opted in. Mitigations for cluster with `MemoryQoS ` turned on:
-- Operators can set `memoryReservationPolicy: None` to disable memory.min/memory.low protection or `memoryThrottlingFactor: 1.0` to disable early throttling
+**Blast Radius:** In Alpha (v1.36), the `MemoryQoS` feature gate is disabled by default. In Beta (v1.37), the feature gate is enabled by default. On upgrade to v1.37, `memory.high` is set for all Burstable and BestEffort containers. Memory protection (`memory.min`/`memory.low`) is not applied unless `memoryReservationPolicy` is set to `TieredReservation` (default is `None`). Mitigations:
```
ajaysundark, line 447: After 1.37 upgrade users will start seeing performance difference for their burstable workloads. If swap is not enabled (this is not a prerequisite for the feature), the reclaim may not be effective in freeing up memory. Is there a reason why this behavior is not backed by `memoryReservationPolicy` similar to memory.min/low?

sohankunkerkar: Good question. The split was intentional in v1.36. The `MemoryQoS` feature gate enables `memory.high` throttling, while `memoryReservationPolicy` only controls `memory.min/memory.low` reservation. The thinking was to let operators enable throttling without also forcing reservation, since reservation has a different blast radius. Putting memory.high behind `memoryReservationPolicy` would make `None` effectively disable the core throttling behavior rather than just disabling reservation.
 
That said, your upgrade concern makes sense. Beta default-ON means Burstable workloads start getting memory.high. For operators who want to avoid early throttling, the mitigation is `memoryThrottlingFactor: 1.0`, which collapses `memory.high` to the limit boundary and removes the pre-limit throttle gap. Without swap, reclaim at `memory.high` is also less effective, especially for anonymous/heap-heavy workloads, so we should call this out more clearly in the upgrade guidance.

ndixita: I think change in behavior for existing workloads should be opt-in rather than by default.

ajaysundark: I agree with Dixi here on the impact. This will be breaking change for a wide surface affecting all burstable and best effort workloads.

Default should try to maintain existing behavior as the performance will be very observable - for instance, for large database workloads they'd see performance variation at 10% memory capacity remaining (that could be GBs of memory in reality). 

With swap disallowed for best effort pods in K8s, even in presence of swap this would have serious implications (worst case OOMs) even with 10% memory capacity remaining.

was this discussed in sig-node? 

cc @acurtiz @SergeyKanzhelev @haircommander

QiWang19: When comparing `memory.high` with no swap to having no `memory.high` and no swap, the configuration with memory.high works better or neutral because it introduces throttling before an OOM kill occurs at `memory.max`, which MemoryQoS does not change. I understand this is why memory.high is configured by default.
memoryQoS setting memory.high has been in alpha since v1.22, it was a design decisoin made earlier. As Sohan mentioned above,  there is a mitigation path `memoryThrottlingFactor: 1.0`.

ndixita: A variety of workloads run on K8s today, and we should empower operators to decide on the memoryThrottlingFactor based on their specific workload requirements. For large database workloads example, 10% memory capacity can represent gigabytes of headroom; if these workloads are actively throttled due to memory.high defaulting to 0.9, it will cause significant and observable performance degradation.

While having mitigation paths is beneficial, requiring operators to proactively modify their Kubernetes configuration to restore original performance levels for existing workloads effectively makes this a breaking change. Opt-out by default, and let operators decide whether they want to use the feature or not seems safer.

QiWang19: I think about two options:

1.  Change `memoryThrottlingFactor` default from 0.9 to `nil`, memory.high throttling is effectively off by default in beta. Operators opt-in by setting e.g. 0.9.
2. Keep 0.9 default, add a new knob to kubelet configuration, but keeps the 0.9 default for those who opt in.

haircommander: I'm a fan of 1

ajaysundark: +1 to (1) off by default.

ndixita: Thank you for addressing the concern. @QiWang19 @haircommander option 1 sounds good to me.

PR #6144, opened 2026-06-03
KEP-5541: Update PersistentVolumeClaimUnusedSinceTime KEP for beta in 1.37

RomanBednar:
- One-line PR description: Update KEP-5541 for beta graduation in v1.37 and add PRR approver

  - Issue link: https://github.com/kubernetes/enhancements/issues/5541

  - Other comments: 
    - Updates KEP-5541 (Report Last Used Time on a PVC) for beta graduation:
      - Updated KEP design to use conditions-based approach (`Unused` condition on PVC status)
      - Updated kep.yaml with stage: beta and milestone v1.37
      - Added beta PRR approver to prod-readiness file

### DIFF
--- a/keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
+++ b/keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
@@ -44,7 +44,7 @@ Items marked with (R) are required _prior to targeting to a milestone / release_
 - [x] (R) KEP approvers have approved the KEP status as `implementable`
 - [x] (R) Design details are appropriately documented
 - [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
-  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [x] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
 - [x] (R) Graduation criteria is in place
@@ -61,10 +61,11 @@ Items marked with (R) are required _prior to targeting to a milestone / release_
 
 ## Summary
 
-This KEP proposes adding a new `UnusedSince` status field on a `PersistentVolumeClaim`
-to determine when a PVC was last used by a Pod. This enables cluster
+This KEP proposes adding a new `Unused` condition on `PersistentVolumeClaimStatus`
+to determine whether a PVC is currently in use by a Pod. This enables cluster
 administrators to identify unused PVCs and implement cleanup policies for
-storage that is no longer in use.
+storage that is no longer in use. The `lastTransitionTime` of the condition
+indicates when the PVC last changed between used and unused states.
 
 ## Motivation
 
@@ -79,8 +80,8 @@ this the ideal place to track usage.
 
 ### Goals
 
-- Add an `UnusedSince` timestamp field to `PersistentVolumeClaimStatus`
-- Update this field with a timestamp when a PVC is last used by a Pod
+- Add an `Unused` condition type to `PersistentVolumeClaim` status conditions
+- Set the condition to reflect whether the PVC is currently in use by any non-terminal Pod
 
 ### Non-Goals
 
@@ -90,12 +91,19 @@ this the ideal place to track usage.
 
 ## Proposal
 
-Add a new field `UnusedSince` of type `metav1.Time` to the `PersistentVolumeClaimStatus`
-struct. This field will be updated by the PVC protection controller (in kube-controller-manager)
-when the PVC transitions from being in use to not being in use.
+Add a new condition type `Unused` to the `PersistentVolumeClaim` status conditions.
+This condition is managed by the PVC protection controller (in kube-controller-manager)
+and reflects whether the PVC is currently referenced by any non-terminal Pod.
 
-The definition of a PVC not being in use is when the last Pod referencing it has
-been deleted or reached a terminal state.
+When the last Pod referencing a PVC is deleted or reaches a terminal state, the
+condition is set to `Status=True` with `Reason="NoPodsUsingPVC"` and
+`Message="No pods are currently referencing this PVC"`. When a new Pod starts
+referencing the PVC, the condition is set to `Status=False` with
+`Reason="PodUsingPVC"` and `Message="A pod is currently referencing this PVC"`.
+
+The `lastTransitionTime` of the condition indicates when the PVC last changed
+between used and unused states, providing the equivalent of an "unused since"
+timestamp.
 
 ### User Stories (Optional)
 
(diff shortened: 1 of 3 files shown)


kannon92:
/assign @ameukam 

for PRR shadow.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -272,90 +284,80 @@ occurs.
...
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-<!--
-Describe manual testing that was done and the outcomes.
-Longer term, we may want to require automated upgrade/rollback tests, but we
-are missing a bunch of machinery and tooling and can't do that now.
--->
-Yet to be completed for beta.
+Not tested yet.
```
ameukam: Because, this KEP is targeting Beta for 1.37, would it possible to have a test to validate the downgrade path at least ?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -365,64 +367,44 @@ No.
...
-
 - [ ] Metrics
-  - Metric name:
-  - [Optional] Aggregation method:
-  - Components exposing the metric:
-- [ ] Other (treat as last resort)
-  - Details:
+  - No metric
+- [x] Other (treat as last resort)
+  - Rate of PVC status update errors in kube-controller-manager
```
ameukam: Without an actual Prometheus metric, it would be hard to monitor the SLO defined in this document. Are you referring to a generic KCM worqueue metric ?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -365,64 +367,44 @@ No.
...
 ### Scalability
 
 ###### Will enabling / using this feature result in any new API calls?
 
-Yes. A new PATCH call for PersistentVolumeClaimStatus. Estimated throughput would be
-one PATCH call per PVC when transitioning from in use to not in use. The originating
-component is PVC Protection Controller (in kube-controller-manager).
+Yes. A new UpdateStatus call for PersistentVolumeClaim. Estimated throughput would be
+one UpdateStatus call per PVC when transitioning between in use and not in use states.
+The originating component is PVC Protection Controller (in kube-controller-manager).
```
ameukam, line 438: In a configuration where there is a large number (5000?) of PVCs, will the controller immediately attempt to issue 5000 API Calls to append the `Unused` condition to PVCs that lack it ?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -61,10 +61,11 @@ Items marked with (R) are required _prior to targeting to a milestone / release_
 
 ## Summary
 
-This KEP proposes adding a new `UnusedSince` status field on a `PersistentVolumeClaim`
-to determine when a PVC was last used by a Pod. This enables cluster
+This KEP proposes adding a new `Unused` condition on `PersistentVolumeClaimStatus`
```
kannon92, line 64: Putting a note here for myself.

This is not alpha -> beta change but this reflects what went into alpha.

This PR is updating both the KEP based on alpha and promoting to beta. This is fine because the actual API merged in alpha.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: Yes, you can test this with alpha in 1.36. AFAIK we don't require automation but we still expect some verification of this. You can see https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/3939-allow-replacement-when-fully-terminated#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested for examples of how other keps have done this.

Please do this and report back.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -201,69 +216,66 @@ to implement this enhancement.
...
 
 - Feature implemented successfully behind a feature gate
 - Unit tests added to test out feature enablement/disablement, and passing
-- Initial e2e tests completed and enabled
 
 #### Beta
 
-TBD
+- Feature implemented and stable in alpha for one release
+- Initial e2e tests completed and enabled
```
kannon92, line 249: Ideally there needs to be e2e tests before this feature graduates to beta.

I would block promoting this feature until there are e2e tests in CI.

We can merge this KEP but I want to make sure that there is testing on this feature before we promote to beta.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -365,64 +367,44 @@ No.
 
 ###### How can an operator determine if the feature is in use by workloads?
 
-Operators can check if the PVCs have `status.UnusedSince` field populated.
+Operators can check if PVCs have an `Unused` condition in their `.status.conditions`.
```
kannon92, line 393: I'm really struggling with there being no metrics for this. 

> This enables cluster administrators to identify unused PVCs and implement cleanup policies for storage that is no longer in use.

If we have no metrics, a cluster admin has to check every PVC across namespaces. And a imagine a cluster admin is supporting 100s of clusters. This sounds really painful.

How do you envision someone supporting this feature in a multicluster environment? I could imagine that maybe you create an alert or some sorta of notification based on LastUsedTime so that one can know to delete PVC. But this seems difficult.

RomanBednar: I think this is what @ameukam is suggesting above: https://github.com/kubernetes/enhancements/pull/6144#discussion_r3389012681

I picked this up in Beta, and I think Alpha assumed that this will be used mostly with some custom tools, as you mentioned. But metric would certainly help with validating the SLO. I'll add it to this KEP.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Indeed, this is already in progress. We can add e2e and graduate to beta in same PR: https://github.com/kubernetes/kubernetes/pull/139620

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Correct - the implementation changed last minute, so this is also an update to reflect it.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: That's correct - I intend to run a scale test in beta to see how this behaves and how much time it takes for larger number of PVCs.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: I was referring to KCM logs, but you're right that a metric would help with SLO monitoring. We could add a `pvc_protection_controller_unused_condition_syncs_total{result}` counter that tracks both successful and failed condition updates.

Do you want this added to beta KEP?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Absolutely, I intend to do a manual upgrade/downgrade testing. But this can be done after KEP freeze, correct?

RomanBednar:
@ameukam @kannon92 I've addressed your comments and updated KEP to include the metric. Thank you for reviews and let me know if we're good for KEP freeze and merge.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: Usually we require this to be done for the beta KEP.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: I don't see this as related.

I'm mostly trying to understand how you intend the above cluster admin story to be resolved.

Right now a cluster admin would need to check every PVC across namespaces. I was thinking that an alert or a notification when a PVC has been unsused for certain amount of time would be necessary to achieve your goals.

I guess you could do this with an operator or controller that scans all namespaces. But it isn't clear to me if that is next steps for this or not?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: This answer concerns me.

Following [our policy](https://github.com/kubernetes/enhancements/tree/master/keps/sig-architecture/5241-beta-featuregate-promotion-requirements) on beta requirements I would expect this test to be performed and answered in this KEP before we aim to promote to beta.

We can proceed with this KEP as long as you commit to performing the scale test before turning this feature gate on. This should be a blocker for beta.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: I think the intention was that admin will create a script to check every PVC across namespaces. As an ad-hoc maintenance check. Providing an alert was not intended, but could be a follow-up - I'm just not sure if there's demand for this. Maybe this would be rather annoying to most users, or maybe some clusters are kept intentionally with some percentage of unused volumes...not sure really.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Similar situation as above, this is a dept from Alpha which I did not work on. I do commit to performing the scale tests - if this does not look good we still have the option to slip by not graduating feature in k/k I believe.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Well...usually I would do it in alpha, but I inherited this feature in beta. If this is considered a blocker the feature would slip, unless we don't go the usual route. 

The promotion can always be prevented before merging k/k feature gate change as well...

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
jsafrane: > In a configuration where there is a large number (5000?) of PVCs, will the controller immediately attempt to issue 5000 API Calls to append the Unused condition to PVCs that lack it ?

Yes, 5000 updates will be initiated. I would expect that the controller has its own kubeclient and it will throttle itself. The rollout may be slower, but steady. And I'd expect a scale test would confirm that (after merging this KEP and before flipping the feature gate to Beta).

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
jsafrane: > Usually we require this to be done for the beta KEP.

I think the requirement was to test it before flipping the actual feature gate to beta. Does it need to be done before a KEP targetting beta is merged?

kannon92: Usually yes.
Will the KEP be updated before the beta work is merged? Saying that this test was complete.

If that is a ack then I am fine to blcok beta promotion on this. So both the test is done and the KEP will be updated to reflect that this test is complete.

IMO its easier to perform this test now though rather than adding the KEP update after enhancement freeze.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -201,69 +216,66 @@ to implement this enhancement.
...
-- Initial e2e tests completed and enabled
 
 #### Beta
 
-TBD
+- Feature implemented and stable in alpha for one release
+- Initial e2e tests completed and enabled
+- Comprehensive unit test coverage
+- E2E tests added and passing
+- PRR completed and approved
```
kannon92, line 252: Can you add scale testing for beta?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
gnufied: >If we have no metrics, a cluster admin has to check every PVC across namespaces. And a imagine a cluster admin is supporting 100s of clusters. This sounds really painful.

@kannon92 so this entire feature is basically an observational feature.  This feature is intended to let cluster-admins decide their own staleness threshold and create alerts and metrics.  

We are providing data that cluster-admin can use to surface their own alerts. Say a valid metric to emit here could be - number of unused PVCs in the cluster. But that begs the question, unused for how long? May be we can hardcode a default like unused for 10 days but that seems wrong.

You are bringing a valid point that, how will a cluster-admin which operates 100s of clusters will observe this feature but in core k8s, we **can't decide** how long a PVC has to be unused to be considered really unused from cluster-admin perspective. For some clusters, it may be 10 days for some others it may be 30 days. 

Another way this can be used is - by a kubectl plugin which lists unused PVCs in a cluster and you specify number of days the PVCs must be unused as a CLI parameter.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -272,90 +284,81 @@ occurs.
...
 
-<!--
-What signals should users be paying attention to when the feature is young
-that might indicate a serious problem?
--->
+Monitor the `pvc_protection_controller_unused_condition_syncs_total` metric
+(labeled by `result`: `success` or `error`). An elevated error rate for this
+metric indicates failures when updating the `Unused` condition on PVCs and
+could inform a rollback decision. Additionally, check kube-controller-manager
+logs for PVC status update errors.
```
gnufied, line 357: >pvc_protection_controller_unused_condition_syncs_total

This metric seems unnecessary. We handle these updates via existing controller and we already have workqueue metrics syncs.

gnufied: I think this can be nailed down during implementation.  I will withdraw my objection for now. I do think workqueue metrics should have been sufficient.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: So is the idea that we introduce this condition and the community would build kubectl plugin or do whatever they need for this?

I just don't see this KEP as is addressing the user story in the KEP. But I agree that unusued is open to interpretation. 

Will you provide documentation for cluster-admins on how to best leverage this feature for this user story?

gnufied: >I just don't see this KEP as is addressing the user story in the KEP. But I agree that unusued is open to interpretation.

I think the KEP meets the goal. User stories are optional . I just don't believe this has to be a metric from core k8s. Metrics are not free IMO and each timeseries we create has a cost, we should be careful and emit metrics that are meaningful. 

> Will you provide documentation for cluster-admins on how to best leverage this feature for this user story?

Yes. But again updating kube-state-metrics or creating a `kubectl` plugin is out of scope for this KEP.

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -201,69 +216,66 @@ to implement this enhancement.
...
 - Feature gate enable/disable behavior
 
-- [test name](https://github.com/kubernetes/kubernetes/blob/2334b8469e1983c525c0c6382125710093a25883/test/e2e/...): [SIG ...](https://testgrid.k8s.io/sig-...?include-filter-by-regex=MyCoolFeature), [triage search](https://storage.googleapis.com/k8s-triage/index.html?test=MyCoolFeature)
-
 ### Graduation Criteria
 
 #### Alpha
 
 - Feature implemented successfully behind a feature gate
 - Unit tests added to test out feature enablement/disablement, and passing
```
kannon92, line 244: I think I want to block on this requirement. Why was e2e tests moved to beta?

Ideally I want you all to prove that these tests are stable and high quality before promoting this feature to beta. 

There are no blockers from testing alpha features in CI and we have lanes to make sure that alpha feature don't break release.

kannon92 (commented):
/hold

There are three things that I want addressed before beta.

1. Scalability testing of PVCs
2. Upgrade / downgrade testing
3. E2E tests running this feature.

To me, I think 3 is a hard no for 1.37.

I don't have confidence in this feature because there are no e2e tests in our CI with this feature.

https://github.com/kubernetes/kubernetes/pull/139620/changes#r3415459595

For 1.37, let's do 1, 2 and 3. If that is done I see this being a great candidate for beta on in 1.38. 

Also, please merge the KEP update in 1.37 so main can reflect the real design. You can remove the PRR request for beta and focus on getting the updates.

gnufied:
For posterity or PRR reviewers approve the feature graduation,  I performed upgrade, enable and downgrade tests with 500 Pods and PVCs running in the clusters.

1. The Conditions will be left over when feature is disabled on downgrade (after upgrade).
2. When I enabled the feature with 500 PVCs, there was some churn in KCM but it pretty quickly settled down and I noticed no leftover memory usage or anything.
3. Similarly when I mass deleted all the 500 pods that were referring those PVCs, there was some churn in KCM which quickly updated all PVCs, but KCM settled down pretty quickly.

kannon92:
For 1, I am curious on how chatty the updates for last used would be. ie if I have lots of pods that get deleted will the last used be updated quickly? And then new pods are used and it sets back.

For other projects we have introduced a knob to control how often updates go for reporting usage (Kueue is the example I think of)

RomanBednar:
I tested with 1000 PVCs on a single-node cluster. When all pods were mass-deleted, all 1000 PVCs transitioned to Unused=True within 50 seconds (~20 PVC/s). When we immediately recreated pods after deletion (overlapping events), workqueue deduplication meant only 7.6% of PVCs saw an intermediate state - the system settled in 10 seconds. KCM memory impact was +3 MB. We see no need for an additional throttle mechanism.

| Scenario                                  | PVCs | Duration | Throughput                 | KCM Memory |
|-------------------------------------------|------|----------|----------------------------|------------|
| Feature gate enable                       | 1000 | 47s      | ~21 PVC/s                  | +3 MB      |
| Mass pod deletion                         | 1000 | 50s      | ~20 PVC/s                  | +3 MB      |
| Flip-flop (delete → immediately recreate) | 1000 | 10s      | ~100 PVC/s (92.4% deduped) | negligible |

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: Sounds good.

kannon92:
Can you update the KEP with the scalability tests you performed and your findings?

RomanBednar:
@kannon92 The gist of it is already there: https://github.com/kubernetes/enhancements/compare/b5ddffda501300353ab396253e11b56db4f94a44..45f8bae48eb69f8aab2f6d5a9ef55d4a2675f3d7

RomanBednar:
@kannon92 The criteria mentioned are met - even though they are not required before merging a KEP, which was confirmed by two other members on Slack (@BenTheElder, @jsafrane). 

1. Scalability testing of PVCs - done, added to KEP
2. Upgrade / downgrade testing - done, added to KEP
3. E2E tests running this feature - done, needs a merge: https://github.com/kubernetes/kubernetes/pull/139784

E2E tests results, all passing:  https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/139784/pull-kubernetes-e2e-kind-alpha-beta-features/2066881433938432000

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
jsafrane: BTW, the PV / PVC protection controller already did the same when it graduated to enabled-by-default. It  added a finalizer to all PVs and PVCs in the cluster. I don't remember any complaints.

jsafrane:
I am approving the KEP as a plan for beta promotion. I am going to block the feature gate promotion until all beta requirements were met. Especially super green e2e tests.
/approve

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
```diff
@@ -201,69 +216,66 @@ to implement this enhancement.
...
-- Initial e2e tests completed and enabled
 
 #### Beta
 
-TBD
+- Feature implemented and stable in alpha for one release
```
kannon92, line 248: What do you mean by "stable in alpha for one release"?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
kannon92: Why are we moving this to a beta requirement? 

Previous release we agree to add this in alpha.

kannon92 (commented):
Thank you for addressing the beta items.

/approve

gnufied:
@kannon92 can you remove the hold?

keps/sig-storage/5541-pvc-last-used-time-status-field/README.md
RomanBednar: Maybe a bad wording, the message should be: it was in alpha, it did not break anything, and nobody complained :)

petr-muller:

PR #6145, opened 2026-06-03
KEP-961: update metadata for maxUnavailable for STS

soltysh:
- One-line PR description: This PR updates metadata, and KEP history, since we're turning the feature on by default in 1.37. 

- Issue link: https://github.com/kubernetes/enhancements/issues/961

/assign @helayoty

### DIFF
--- a/keps/sig-apps/961-maxunavailable-for-statefulset/kep.yaml
+++ b/keps/sig-apps/961-maxunavailable-for-statefulset/kep.yaml
@@ -28,7 +28,7 @@ stage: beta
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.35"
+latest-milestone: "v1.37"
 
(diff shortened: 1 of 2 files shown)


keps/sig-apps/961-maxunavailable-for-statefulset/kep.yaml
```diff
@@ -28,12 +28,12 @@ stage: beta
...
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.35"
+latest-milestone: "v1.37"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
   alpha: "v1.24"
-  beta: "v1.35"
+  beta: "v1.37"
```
helayoty: I don't think we need to update this since the feature moved to beta in v1.36.

soltysh: Good point, updating.

helayoty (approved):
/lgtm

Issue #5825, opened 2026-01-21
CRI List Streaming


bitoku:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Server-side streaming for CRI List Operations
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5825-cri-pagination 
- Discussion Link: https://youtu.be/zsLII23v-dk?si=rQeeNZi9uDgHwEkq&t=2207 
- PRs by stage and milestone:
  - [x] Alpha - v1.36
    - [x] KEP (`k/enhancements`) update PR(s):
      - https://github.com/kubernetes/enhancements/pull/5826
    - [x] Code (`k/k`) update PR(s):
      - https://github.com/kubernetes/kubernetes/pull/136987 
    - [x] Docs (`k/website`) update PR(s):
      - https://github.com/kubernetes/website/pull/54433

  - [ ] Beta - v1.38
    - [ ] KEP (`k/enhancements`) update PR(s):
      - [ ] https://github.com/kubernetes/enhancements/pull/6146
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update(s):
    - [ ] Runtime update(s):
      - [ ] https://github.com/cri-o/cri-o/pull/9761
      - [ ] https://github.com/containerd/containerd/pull/13187
      - [ ] https://github.com/kubernetes-sigs/cri-tools/pull/2099

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

kannon92:
@haircommander This needs a lead-opt-in label.

jmickey:
Hello @bitoku 👋, v1.36 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Wednesday 4th February 2026 (AoE) / Thursday 5th February 2026, 12:00 UTC**.

This enhancement is targeting stage `alpha` for v1.36 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
[...]

PR #6146, opened 2026-06-03
KEP-5825: update CRI List Streaming for beta graduation in v1.38

bitoku:
Assisted-by: Claude

  
- One-line PR description: Update KEP-5825 for beta graduation

- Issue link: https://github.com/kubernetes/enhancements/issues/5825

- Other comments:
  - k/k change in Alpha
      - https://github.com/kubernetes/kubernetes/pull/136987

### DIFF
--- a/keps/sig-node/5825-cri-pagination/README.md
+++ b/keps/sig-node/5825-cri-pagination/README.md
@@ -369,7 +369,8 @@ kubelet does not have integration tests.
 - Feature gate enabled by default
 - Both containerd and CRI-O implement streaming RPCs
 - E2E tests passing
-- Metrics for streaming usage added
+- Existing kubelet CRI metrics (`kubelet_runtime_operations_errors_total`,
+  `kubelet_runtime_operations_duration_seconds`) cover streaming transparently
 - Documentation updated
 
 #### GA
@@ -423,10 +424,18 @@ Streaming resumes if the runtime supports it. No state is persisted.
 ###### Are there any tests for feature enablement/disablement?
 
 No dedicated enablement/disablement tests are planned.
-The feature is stateless and toggling requires a kubelet restart, so the e2e tests that cover the [behavior matrix](#behavior-matrix)
+The only internal state is the cached fallback flag, which is reset on
+kubelet restart. Toggling the feature gate requires a kubelet restart,
+so the e2e tests that cover the [behavior matrix](#behavior-matrix)
 (including the fallback to unary RPCs) provide sufficient coverage.
 
-To be discussed further before beta graduation.
+Specifically, the following tests verify enablement/disablement behavior:
+- Unit tests confirm that when `CRIListStreaming` is disabled, kubelet uses
+  only unary RPCs and never attempts streaming calls.
+- CRI-proxy e2e tests confirm that when the feature gate is enabled, kubelet
+  calls the streaming RPCs.
+- CRI-proxy e2e tests confirm that when the runtime returns `UNIMPLEMENTED`,
+  kubelet falls back to unary RPCs automatically.
 
 ### Rollout, Upgrade and Rollback Planning
 
@@ -436,17 +445,25 @@ Rollout/rollback cannot impact running workloads. Streaming only affects how con
 
 ###### What specific metrics should inform a rollback?
 
-`kubelet_cri_list_streaming_failure_total`, Type: `Counter`, Label: `operation`
+`kubelet_runtime_operations_errors_total{operation_type="list_containers"}` (and equivalent
+`list_podsandbox`, `list_images`, etc.)
 
-This counter increments when a streaming RPC encounters an error during `Recv()`, causing kubelet to discard any items received so far. A sustained increase may indicate runtime instability.
+A sustained increase in CRI list operation errors may indicate runtime
+instability with streaming RPCs. This existing metric already captures
+mid-stream failures because the error propagates through the CRI client
+to the instrumented service layer.
 
-`kubelet_cri_list_streaming_fallback_total`, Type: `Counter`, Label: `operation`
+`kubelet_runtime_operations_duration_seconds{operation_type="list_containers"}` (and equivalent)
 
-This counter increments when kubelet falls back to the unary RPC because the runtime returned `UNIMPLEMENTED` for the streaming RPC.
+An unexpected increase in list operation latency after enabling the feature
+gate may indicate streaming overhead.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-To be tested during Alpha phase.
+Upgrade and downgrade do not require a specific path. Kubelet must be
+restarted to toggle the feature gate, which resets its internal state.
+When enabled, kubelet attempts streaming RPCs first and falls back to
+unary RPCs if the runtime returns `UNIMPLEMENTED`.
 
 ###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
 
@@ -460,8 +477,11 @@ Not applicable. It's not a workload-level feature.
 
 ###### How can someone using this feature know that it is working for their instance?
 
-- `kubelet_cri_list_streaming_fallback_total` is **not** incrementing: the runtime supports streaming and kubelet is using the streaming RPCs.
-- `kubelet_cri_list_streaming_failure_total` is **not** incrementing: streams are completing successfully without errors.
+- `kubelet_runtime_operations_errors_total` for list operations (`list_containers`,
+  `list_podsandbox`, etc.) is **not** incrementing: CRI list calls are completing
+  successfully, whether via streaming or unary RPCs.
+- `kubelet_runtime_operations_duration_seconds` for list operations is **not**
+  increasing: streaming is not adding latency overhead.
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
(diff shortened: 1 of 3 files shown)


keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -436,17 +445,25 @@ Rollout/rollback cannot impact running workloads. Streaming only affects how con
...
+to the instrumented service layer.
 
-`kubelet_cri_list_streaming_fallback_total`, Type: `Counter`, Label: `operation`
+`kubelet_runtime_operations_duration_seconds{operation_type="list_containers"}` (and equivalent)
 
-This counter increments when kubelet falls back to the unary RPC because the runtime returned `UNIMPLEMENTED` for the streaming RPC.
+An unexpected increase in list operation latency after enabling the feature
+gate may indicate streaming overhead.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
```
kannon92, line 461: Can you do something like https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/3085-pod-conditions-for-starting-completition-of-sandbox-creation#were-upgrade-and-rollback-tested-was-the-upgrade-downgrade-upgrade-path-tested?

We want to confirm that you are testing upgrades of this feature even if manually.

keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -460,34 +477,56 @@ Not applicable. It's not a workload-level feature.
...
 - `crio_operations_latency_seconds_total` (CRI-O)
 
-To be discussed further before beta graduation.
+Operators can use the existing kubelet-side metrics to monitor streaming health:
+- `kubelet_runtime_operations_errors_total` for list operations near zero: CRI list
+  calls completing successfully.
+- `kubelet_runtime_operations_duration_seconds` for list operations stable: no
+  latency regression from streaming.
 
 ###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
```
kannon92, line 498: We have https://github.com/kubernetes/community/blob/main/sig-scalability/slos/pod_startup_latency.md#pod-startup-latency-slislo-details but I'm not sure if CRI list will impact pod startup?

I guess it shouldn't but maybe we should include that as a SLO as increased pod startup time would be concerning operationally.

keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -369,7 +369,8 @@ kubelet does not have integration tests.
 - Feature gate enabled by default
 - Both containerd and CRI-O implement streaming RPCs
```
kannon92, line 370: Do we have containerd support for this?

bitoku: Not yet, the PR is pending.
https://github.com/containerd/containerd/pull/13187

keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -460,34 +477,56 @@ Not applicable. It's not a workload-level feature.
...
+below the instrumented service layer where kubelet metrics are recorded.
+Adding them would require either introducing metrics registration into
+the `cri-client` staging package (which currently has no metrics
+dependency) or injecting a callback interface from kubelet into the CRI
+client. Neither is justified because:
+
+- Mid-stream failures already propagate as errors to the instrumented
+  layer and are captured by `kubelet_runtime_operations_errors_total`.
+- Fallback to unary is cached after the first occurrence (via
+  `atomic.Bool`) and only happens a few times at startup. Kubelet logs
```
kannon92, line 526: Logs are a poor way to support a feature rollout.

I did an AI review and it has this to say:

>
   1. No streaming-specific observability — this is the biggest concern

    The alpha KEP originally promised dedicated streaming metrics
    (kubelet_cri_list_streaming_failure_total,
    kubelet_cri_list_streaming_fallback_total). The beta update removes these and
    argues existing metrics are sufficient. The justification (avoiding metrics in
    the cri-client staging package) is reasonable from a code-cleanliness
    perspective, but operationally this means:

    - An operator cannot tell whether streaming is active or the kubelet silently 
    fell back to unary. The only signal is a kubelet Info-level log line. At
    scale, log-based detection is fragile — operators rely on metrics and
    dashboards.
    - You cannot distinguish a streaming mid-stream failure from any other CRI 
    error. kubelet_runtime_operations_errors_total is a catch-all. If streaming
    introduces a new failure mode, it's invisible in the existing metric without
    label enrichment.
    - Recommend at minimum: a log-based metric or a label on the existing metric
    (path=streaming|unary) to let operators confirm code path. If the
    architectural argument holds (no metrics in cri-client), consider emitting a
    kubelet-level event or condition that's scrapeable.

I still thinking something more than logs to detect fall back is important. How would you detect if this feature is misheaving on a 10000 node cluster?

keps/sig-node/5825-cri-pagination/README.md
kannon92: So the requirement for beta is to have both implement this? Should we push this feature to 1.38?

I'm not as clear on CRI support for containerd skew. Even if this merges will we have a production release in time for 1.37?

kannon92 (commented):
Out of diff:

Can you fill out the checkbox for https://github.com/bitoku/kubernetes-enhancements/blob/def4672e67e92e65dc510ec5def8738c0a9df4bd/keps/sig-node/5825-cri-pagination/README.md#release-signoff-checklist

keps/sig-node/5825-cri-pagination/README.md
bitoku: > So the requirement for beta is to have both implement this?

Yes. 

> Should we push this feature to 1.38?
I'm not as clear on CRI support for containerd skew. Even if this merges will we have a production release in time for 1.37?

I was thinking we could merge the KEP while we're waiting for the PR merged.
However, as you said, containerd 2.4 release is scheduled to August, which is prior to 1.37 Code Freeze (end of July). We can't write test with the new containerd version, right?

I guess we may want to push to 1.38.

ref: https://containerd.io/releases/#current-state-of-containerd-releases

cc @haircommander @samuelkarp

keps/sig-node/5825-cri-pagination/README.md
bitoku: There's some difficulty having the streaming specific metrics because cri-client is a reusable client library with no metrics. Registering metrics via init() would inject kubelet-specific side effects into every importer (tests, third-party tools).

I'm exploring the possible solutions for it, and found we could do something like
https://github.com/kubernetes/kubernetes/pull/139500
It doesn't affect other importers and can emit metrics from kubelet.

I'd like to hear opinions about it.

keps/sig-node/5825-cri-pagination/README.md
haircommander: if the PR is staged for a release candidate before test freeze then I think it's acceptable. But we also could defer if needed, I just think this feature is gaining in importance as we look at extending the number of pods per node cc @SergeyKanzhelev

keps/sig-node/5825-cri-pagination/README.md
haircommander: I think this approach is reasonable and similar to other instrumented services

keps/sig-node/5825-cri-pagination/README.md
kannon92: I disagree. We won't be able to have testing for containerd before code freeze. THere is also no test freeze anymore.

So we would be effectively going beta without testing containerd in CI.

keps/sig-node/5825-cri-pagination/README.md
kannon92: cc @SergeyKanzhelev

keps/sig-node/5825-cri-pagination/README.md
SergeyKanzhelev: I think we need to wait for containerd support. It is important feature, but not urgent. I am a bit sad we didn't merge the PR to containerd in-time for 2.3

bitoku: Sure, thank you all for taking a look. I'll keep the PR open to discuss more about the design, but I'll update the KEP issue to tell it'll be delayed to 1.38.

bitoku:
/hold 
delay it to 1.38

kannon92:
@bitoku 
looking at this for PRR and this seems really close. Please let me know when this is ready for review.

SergeyKanzhelev:
/unhold

does this one needs to be rebased?

keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -368,8 +375,12 @@ kubelet does not have integration tests.
...
 - Both containerd and CRI-O implement streaming RPCs
-- E2E tests passing
-- Metrics for streaming usage added
+- Streaming and fallback e2e tests passing in Kubernetes CI with supporting
+  versions of both runtimes
+- Manual upgrade->downgrade->upgrade and feature gate rollback results recorded
+- `kubelet_cri_streaming_enabled` metric available
+- Existing kubelet CRI metrics (`kubelet_runtime_operations_errors_total`,
+  `kubelet_runtime_operations_duration_seconds`) cover streaming transparently
 - Documentation updated
```
SergeyKanzhelev, line 384: let's add critest for it as a graduation criteria.

keps/sig-node/5825-cri-pagination/README.md
bitoku: Completed manual upgrade->downgrade->upgrade test

keps/sig-node/5825-cri-pagination/README.md
bitoku: added `kubelet_pod_start_sli_duration_seconds` and other two metrics that can be used with `kubelet_cri_streaming_enabled` to see if there's any issue in streaming APIs

bitoku:
@kannon92 The PR is ready for PRR. Could you please do PRR?

kannon92 (commented):
/approve

keps/sig-node/5825-cri-pagination/README.md
```diff
@@ -460,34 +501,57 @@ Not applicable. It's not a workload-level feature.
...
+applies: in a default Kubernetes installation, the 99th percentile per cluster-day
+is at most 5 seconds for schedulable stateless pods, excluding image pulls and
+init container execution. Streaming must not regress this SLO.
 
-To be discussed further before beta graduation.
+Operators can use the existing kubelet-side metrics to monitor streaming health:
+- `kubelet_runtime_operations_errors_total` for list operations near zero: CRI list
+  calls completing successfully.
+- `kubelet_runtime_operations_duration_seconds` for list operations stable: no
+  latency regression from streaming.
```
haircommander, line 522: we also could mention kubelet memory/cpu usage to make sure it's in line. I'm sure it is, but that is another existing metric that could be effected

haircommander (commented):
LGTM, after one note

keps/sig-node/5825-cri-pagination/README.md
mikebrow, line 522: For consideration, we'll be adding runtime side config/tuning and that should work with kubelet tuning to get the best bang for the buck performance out of this feature.

SergeyKanzhelev (approved):
/approve
/lgtm

Issue #6060, opened 2026-05-05
API Server Authentication to Webhooks


benjaminapetersen:
-  KEP Goodies and Messy Bits:
    - Size: large
    - 1.37 work
       - Size: medium/large 
       - KEP merged in 1.37, probably no impl
    - KEP Issue: https://github.com/kubernetes/enhancements/issues/6060
      - [Preliminary Google Doc Design Discussion](https://docs.google.com/document/d/1Rr_J3_g_pO671h4CLd9szFiA8cpcHcnhHWeOx6L0HIU/edit?usp=sharing)
    - KEP PR:  
      - https://github.com/kubernetes/enhancements/pull/6156
      - https://github.com/kubernetes/enhancements/pull/6259
    - KEP authors:
       - pmengelbert
       - benjaminapetersen
    - KEP implementers:
       - pmengelbert
       - benjaminapetersen
    - KEP reviewers:
       - ?
    - KEP (related past KEPs closed):
       - https://github.com/kubernetes/enhancements/pull/658 
       - https://github.com/kubernetes/enhancements/pull/2512
- When last we left our heroes:
  -  From: 
     - [SIG-Auth NEEDS KEP Project Board](https://github.com/orgs/kubernetes/projects/262)
     - [DEPRECATED: Google Doc: Needs KEP, design during 1.37](https://docs.google.com/document/d/1sY8fRyRtk4eG9R439z5ao5i9bFuuxilS03XaNlqoni0/edit?tab=t.0#heading=h.grody9a56o1o)
- Request for new staging repo details (per the [task list described here](https://github.com/kubernetes/kubernetes/blob/master/staging/README.md))
   - [x] Discussion for [new staging repo k8s.io/webhookauth in kubernetes/org](https://groups.google.com/g/kubernetes-sig-architecture)
   - [x] [Thread started on kubernetes-sig-architecture mailing list](https://groups.google.com/g/kubernetes-sig-architecture/c/wrO1AvzZKaM) 
   - [x] Thread started in [sig-architecture slack channel](https://kubernetes.slack.com/archives/C5P3FE08M/p1784317065867799)
   - [ ] new `k8s.io/webhookauth` repository created
   - [ ] `importrestrictions.yaml` file
   - [ ] all mandatory template files added to new `k8s.io/webhookauth`
   - [ ] ensure .github/PULL_REQUEST_TEMPLATE.md and CONTRIBUTING.md files mention that PRs are not directly accepted to the repo
[...]

PR #6156, opened 2026-06-04
KEP-6060: API Server Authentication to Webhooks

pmengelbert:
- One-line PR description: Add API Server Authentication to Webhooks KEP
- Issue link: #6060

### DIFF
--- a/keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
+++ b/keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
@@ -6,24 +6,46 @@
 - [Motivation](#motivation)
   - [Goals](#goals)
   - [Non-Goals](#non-goals)
+- [Terms](#terms)
+  - [Token](#token)
+  - [Token Acquisition Service Account](#token-acquisition-service-account)
+  - [Webhook Authentication Client](#webhook-authentication-client)
+  - [Aggregated API Servers and <code>kube-apiserver</code>](#aggregated-api-servers-and-kube-apiserver)
 - [Proposal](#proposal)
-  - [Webhook Service Account Tokens](#webhook-authentication-tokens)
-  - [Token Acquisition](#token-acquisition)
-    - [Kube-apiserver](#kube-apiserver)
-    - [Aggregated API Servers](#aggregated-api-servers)
-  - [Authorization Checks](#authorization-checks)
-  - [Audience](#audience)
-  - [Token Caching and Rotation](#token-caching-and-rotation)
-  - [Webhook Verification](#webhook-verification)
-  - [User Stories](#user-stories)
-    - [Story 1: Kube-apiserver authenticates to an admission webhook](#story-1-kube-apiserver-authenticates-to-an-admission-webhook)
-    - [Story 2: Aggregated API server authenticates to an admission webhook](#story-2-aggregated-api-server-authenticates-to-an-admission-webhook)
+  - [Sequence Diagrams](#sequence-diagrams)
+    - [Flow 1: <code>kube-apiserver</code> as Webhook Client](#flow-1-kube-apiserver-as-webhook-client)
+      - [JWT payload](#jwt-payload)
+    - [Flow 2: <code>kube-apiserver</code> denies a suspicious request](#flow-2-kube-apiserver-denies-a-suspicious-request)
+    - [Flow 3: Aggregated API Server denied/accepted by webhook](#flow-3-aggregated-api-server-deniedaccepted-by-webhook)
+      - [JWT payload (both requests)](#jwt-payload-both-requests)
   - [Risks and Mitigations](#risks-and-mitigations)
+    - [Broad access to webhooks](#broad-access-to-webhooks)
+    - [Token replay across webhooks](#token-replay-across-webhooks)
+    - [Token replay across API groups](#token-replay-across-api-groups)
+    - [Service account compromise](#service-account-compromise)
+    - [Increased authorization load](#increased-authorization-load)
+    - [Potential Deadlock](#potential-deadlock)
 - [Design Details](#design-details)
-  - [New Private Claims](#new-private-claims)
-  - [BoundObjectRef for APIService](#boundobjectref-for-apiservice)
-  - [RBAC Configuration](#rbac-configuration)
-  - [Kube-apiserver Service Account Lifecycle](#kube-apiserver-service-account-lifecycle)
+  - [Changes to <code>TokenRequest</code> API](#changes-to-tokenrequest-api)
+    - [Changes to <code>TokenRequestSpec</code>](#changes-to-tokenrequestspec)
+    - [Changes to <code>TokenRequest</code> handler](#changes-to-tokenrequest-handler)
+    - [Added claim: <code>&quot;webhook-authentication.k8s.io/allowedAPIGroup&quot;</code>](#added-claim-webhook-authenticationk8sioallowedapigroup)
+  - [Token Acquisition (from the client perspective)](#token-acquisition-from-the-client-perspective)
+    - [All webhook authentication clients:](#all-webhook-authentication-clients)
+    - [<code>kube-apiserver</code>:](#kube-apiserver)
+      - [Dedicated service account for token acquisition](#dedicated-service-account-for-token-acquisition)
+      - [Other details](#other-details)
+    - [Aggregated API Servers:](#aggregated-api-servers)
+  - [Authorization Checks](#authorization-checks)
+    - [RBAC Example](#rbac-example)
+    - [Synthetic resource for authorization checks](#synthetic-resource-for-authorization-checks)
+    - [Recommendations for permissions](#recommendations-for-permissions)
+  - [New types of BoundObjectRef](#new-types-of-boundobjectref)
+  - [Audience](#audience)
+  - [New JWT Private Claims](#new-jwt-private-claims)
+  - [Token Verification](#token-verification)
+  - [Token Review](#token-review)
+  - [Token Caching and Rotation](#token-caching-and-rotation)
   - [Test Plan](#test-plan)
       - [Prerequisite testing updates](#prerequisite-testing-updates)
       - [Unit tests](#unit-tests)
@@ -45,6 +67,11 @@
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
+  - [<code>APIService</code> as <code>BoundObjectRef</code>](#apiservice-as-boundobjectref)
+  - [Client Certificates (mTLS)](#client-certificates-mtls)
+  - [Designated ServiceAccount (&quot;Magic SA&quot;)](#designated-serviceaccount-magic-sa)
+  - [ServiceAccount Token with Identity in Private Claims](#serviceaccount-token-with-identity-in-private-claims)
+  - [AdmissionReview Delegation](#admissionreview-delegation)
 <!-- /toc -->
 
 ## Release Signoff Checklist
@@ -52,17 +79,17 @@
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
 - [x] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [x] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
   - [ ] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
+- [ x (R) Graduation criteria is in place
   - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
-- [ ] (R) Production readiness review completed
-- [ ] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
+- [x] (R) Production readiness review completed
+- [x] (R) Production readiness review approved
+- [x] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation---e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
@@ -73,9 +100,9 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 ## Summary
 
-Today, the kube-apiserver does not authenticate itself to admission webhooks
-by default. Any entity with service network access can send requests to a webhook
-endpoint and impersonate the kube-apiserver.
+Today, `kube-apiserver` does not authenticate itself to admission
+webhooks by default. Any entity with service network access can
+send requests to a webhook endpoint and impersonate `kube-apiserver`.
 [CVE-2025-1974](https://nvd.nist.gov/vuln/detail/CVE-2025-1974) demonstrated
 real-world consequences of this class of vulnerability.
 
@@ -87,14 +114,14 @@ admission webhook. Second, `kube-apiserver` will be updated to dispense
 those tokens to authenticated and authorized principals. Third, a token
 verification library will be introduced for use by webhook maintainers.
 
-This KEP augments the private claims of a service account token (JWT) to support
-three new types of bound object, which will be included in the `TokenRequest`
-made by the [webhook authentication client](#webhook-authentication-client). The
-bound object may be one of `APIService`, `ValidatingWebhookConfiguration`,
-or `MutatingWebhookConfiguration`.
+This KEP augments the `TokenRequest` API to allow for the client to request
+claims in the JWT token to which `kube-apiserver` attests. In addition, tokens
+intended for admission webhooks will be bound to an webhook configuration
+object (either validating or mutating). The specifics of these changes are
+discussed in detail in the [design details](#design-details) section.
 
-A list of [terms](#terms) is provided below to prevent awkward sentence
-constructions and for disambiguation.
+A list of [terms](#terms) is provided below for disambiguation, and to
+prevent awkward sentence constructions.
 
 ## Motivation
 
@@ -109,43 +136,54 @@ through `--admission-control-config-file`), but they require manual credential
 management and an API server restart to change. That opt-in mechanism is
 unopinionated as to the method of authentication (mTLS / token / basic auth),
 creating a large burden on webhook maintainers to support verification of
-client identity by all three methods. More broadly, the burden is greatest
+client identity by all three methods. The burden is greatest
 when the actor setting up the API Server (or aggregated API server) and the
 actor setting up the webhook are not the same, as is usually the case with
 "off-the-shelf", community OSS webhooks.
 
-An opinionated, on-by-default solution is needed to reduce the friction
-to adoption. This KEP is designed to make it possible to transition in
-phases. First, [webhook authentication client](#webhook-authentication-client)
-libraries are configured to use them by default (except in cases where
-it would break an existing authentication setup). At this stage, webhooks
-may not yet have been updated to verify the tokens. Webhooks can instead
-silently ignore them. In the second phase, once credential issuance is GA and
-webhook maintainers can reasonably expect a credential to be present, webhook
-maintainers can use the provided library to opt-in to token verification. Over
-time, we expect this to make the landscape as a whole more secure.
+There are existing out-of tree solutions, such as
+[generic-admission-server](https://github.com/openshift/generic-admission-server).
+However, they require manual setup. This KEP posits that an
+ opinionated, on-by-default solution is needed to reduce the friction
+to adoption. It is designed to make it possible to transition in phases. First,
+[webhook authentication client](#webhook-authentication-client) libraries are
+configured to use them by default (except in cases where it would break an
+existing authentication setup). At this stage, webhooks may not yet have been
+updated to verify the tokens. Webhooks can instead silently ignore them. In
+the second phase, once credential issuance is GA and webhook maintainers can
+reasonably expect a credential to be present, webhook maintainers can use
+the provided library to opt-in to token verification. Over time, we expect
+this to make the landscape as a whole more secure.
 
 In addition to `kube-apiserver`, aggregated API servers often need to contact
 webhooks. Yet, they should should not have broad access to ask arbitrary
-questions to webhooks. A design is needed to make it easy for aggregated API
-servers to query webhooks about resources it controls, but which prevents
-a malicious aggregated API server from requesting policy information about
-resources it does not control.
+questions. A design is needed to make it easy for aggregated API servers to
+query webhooks about resources it controls, but which prevents a malicious
+or compromised aggregated API server from requesting policy information
+about resources it does not control.
 
 The scope of this KEP is limited to authenticating to admission webhooks.
-Authentication webhooks, authorization webhooks, and audit webhooks do
-not share the same practical barriers to authentication experienced by
-admission webhooks. For one, those webhooks are not dynamically deployed at
-runtime, and they already require a `kube-apiserver` restart to change their
-configuration. Furthermore, because the actor setting up `kube-apiserver`
-and the actor setting up the webhook are the same in the vast majority of
-cases, it is much more reasonable to expect that such an actor would use
-the already available solution: they are in control of both the method of
-authentication used by the client and the verification methods used by the
-webhook. This leaves a slight gap, requiring that all deployed aggregated API
-servers that communicate with these webhooks must also have access to the
-necessary credentials. The gap is acknowledged but considered out-of-scope
-to keep the implementation practical for the most common use-cases.
+Authentication webhooks, authorization webhooks, and audit webhooks do not
+share the same practical barriers to authentication experienced by admission
+webhooks. Those webhooks are not dynamically deployed at runtime, and don't
+have the same logistical barriers as admission webhooks: the actor setting
+up `kube-apiserver` and the actor setting up the webhook are the same in the
+vast majority of cases. Therefore, it is much more reasonable to expect that
+such an actor would use the already available solution: they are in control
+of both the method of authentication used by the client and the verification
+methods used by the webhook. This leaves a slight gap for audit webhooks,
+requiring that all deployed aggregated API servers that communicate with these
+webhooks must also have access to the necessary credentials. To close this gap,
+aggregated API servers can deploy their own audit webhook. `TokenReview` and
+`SubjectAccessReview` make this a non-issue for everything but audit webhooks.
+
+Conversion webhooks are likewise out of scope because they pertain to CRDs,
+which will be handled exclusively by `kube-apiserver` and do not share the
+same set of complications caused by allowing access from aggregated API servers.
+
+Credentials issued for authentication to webhooks must be reviewable by
+`kube-apiserver`, but webhooks should be able to verify them independently. No
+API calls should be required of webhooks to verify tokens.
 
 ### Goals
 
@@ -170,6 +208,10 @@ to keep the implementation practical for the most common use-cases.
 * The design is backward compatible: existing kubeconfig-based webhook
   authentication setups continue to work without modification.
 * Defining the webhook-side verification go library.
+* Supporting both `client-go` and `controller-runtime`.
+* Tokens must be reviewable by `TokenReview`, although the webhook
+  library **WILL NOT** use `TokenReview` by default to avoid unnecessary
+  round trips.
 
 ### Non-Goals
 
@@ -183,13 +225,20 @@ to keep the implementation practical for the most common use-cases.
 
 ## Terms
 
+### Token
+Unless otherwise indicted, the term **token** will be used to exclusively
+refer to service account tokens with two qualities: a webhook binding and
+a claim indicating which `APIGroup`s the token's bearer may inquire about.
+When discussing any other token, or service account tokens without both of these
+two qualities, they will be clearly distinguished by the surrounding context.
+
 ### Token Acquisition Service Account
 The service account named in tokens for webhook authentication will be termed
 the **Token Acquisition Service Account**. This is distinct from the identity
 that the principal requesting the token uses to authenticate itself to the
 Kubernetes API Server (which may or may not be a service account). The Token
-Acquisition Service Account must have `attest` permissions on the `APIService`
-object named in the `TokenRequest` ().
+Acquisition Service Account must have `attest` permissions on the `APIGroup`
+named in the `TokenRequest`.
 
 ### Webhook Authentication Client
 Because both `kube-apiserver` and aggregated API servers will attempt
@@ -205,54 +254,77 @@ terms **`kube-apiserver`** and **Kubernetes API Server** will be used
 interchangeably. When referring specifically to an **Aggregated API Server**,
 the full term will always be used.
 
-### Webhook Authentication Bound Object Types
-The term **webhook authentication bound object types** will be used to refer
-to the three newly-added types considered valid as a `BoundObjectRef` in a
-`TokenRequest` where distinguishing between them is not important. Specifically,
-the new types recognized will be `APIService`, `ValidatingWebhookConfiguration`,
-and `MutatingWebhookConfiguration`.
-
 ## Proposal
 
-[Webhook authentication clients](#webhook-authentication-client) may
-request service account tokens with a narrow scope, indicating to the
-webhook that it is only valid for its audience and for `AdmissionReview`
-requests about resources with a particular combination of `APIGroup` and
-`APIVersion` (i.e. an `APIService`). Because the number of per-webhook,
-per-`APIService` tokens can quickly get out of hand for `kube-apiserver`,
-tokens may alternatively be requested that are valid per-webhook, but which
-have no indication of which `APIService` the token may be used for. Because
-the authorization scope of such tokens is larger, broader permissions
-are required to obtain them. Per-webhook tokens are intended for use by
-`kube-apiserver`, whereas per-webhook per-`APIService` tokens are intended
-for use by aggregated API Servers.
+[Webhook authentication clients](#webhook-authentication-client) may request
+service account tokens with a narrow scope, indicating to the webhook that
+it is only valid for its audience and for `AdmissionReview` requests about
+resources with a particular `APIGroup`. Because the number of per-webhook,
+per-`APIGroup` tokens can quickly get out of hand for `kube-apiserver`, tokens
+may alternatively be requested that are valid per-webhook, but which authorize
+the bearer to ask questions about **any resource**. Because the authorization
+scope of such tokens is larger, broader permissions are required to obtain
+them. Tokens that authorize the bearer to make `AdmissionReview` requests about
+**any resource** are intended for use only with `kube-apiserver`, although
+there is no enforcement of this recommendation. Tokens restricted by `APIGroup`
+are intended for use by aggregated API Servers to prevent giving them more
+access than is needed. A fuller description of the permission model for token
+acquisition is described in the [design details](#authorization-checks) section.
 
-The scoping of service account tokens to a particular usage is accomplished
-by adding three new types of private claim to the JWT body. Corresponding
-to each of these is a new type of valid `BoundObjectRef` in the body of a
-`TokenRequest`. This KEP makes `APIService`, `ValidatingWebhookConfiguration`,
-and `MutatingWebhookConfiguration` valid types for the `BoundObjectRef`.
+The `TokenRequest` API will be expanded to accommodate the scoping of
+service account tokens to a particular usage. A brief description of
+those expansions is in order. To obtain a [token](#token), the webhook
+authentication client will make a `TokenRequest` on a service account. For a
+client to obtain a token, it must meet four conditions. First, it must request
+claims indicating which `APIGroups` it intends to query the webhook about;
+these claims must be coherent with the resources the webhook expects. Further
+details are described in the [design details](#changes-to-tokenrequestspec)
+section. Second, it must specify either a `ValidatingWebhookConfiguration`
+or a `MutatingWebhookConfiguration` as the `BoundObjectRef`. Third, it must
+specify an audience that is coherent for that `*WebhookConfiguration`. The
+exact specification of the derivation of the audience is deferred until
+implementation time, and is at the moment subject to change. Fourth and
+finally, the `ServiceAccount` for which the `TokenRequest` is being made
+must have sufficient permission to obtain the token. This is accomplished by
+means of a synthetic authorization check at token issuance, and is described
+in greater detail in the [design details](#authorization-checks) section.
 
-When a per-token webhook is required, as will be the case when the webhook
-authentication client is `kube-apiserver`, the bound object will typically be a
-`ValidatingWebhookConfiguration` or a `MutatingWebhookConfiguration`. Selection
-between the two is of course dependent on which type of webhook `kube-apiserver`
-wishes to contact. The token with one of these two bound object types
-authorizes its bearer to ask *any question* of a single webhook.
+When the `TokenRequest` caller wants a token authorizing the bearer to inquire
+about resources in any `APIGroup`, it will request that `kube-apiserver` attest
+to a claim on the `"*" APIGroup`. This requires that the token acquisition
+service account have broader permissions, described further in the [design
+details](#authorization-checks) section.
 
-Aggregated API servers should not have such broad access to ask questions
-of webhooks. One programmed to maliciously request policy information about
-resources it does not control should be prevented from doing so.
+This broad permission should only be granted to `kube-apiserver`, and its use by
+principals representing aggregated API servers is strongly discouraged. Instead,
+aggregated API servers should request that `kube-apiserver` attest to the
+`APIGroup` corresponding to the server's `APIService`(s) (there may be multiple
+`APIService`s to express multiple `APIVersion`s of a single `APIGroup`). This
+indicates to the webhook that it should deny `AdmissionReview` requests
+that do not pertain to objects within that `APIGroup`. This is recommended
+to prevent a potentially malicious aggregated API server from exposing a
+webhook's policy information or compromising it in some other way.
 
-When the webhook authentication client is an aggregated API server, the
-bound object should be an `APIService`. This indicates to the webhook that
-it should deny `AdmissionReview` requests that pertain to objects within that
-`APIService`'s `APIGroup` and `APIVersion`. This is recommended to prevent a
-potentially malicious aggregated API server from exposing a webhook's policy
-information or compromising it in some other way.
+The `TokenRequest` handler will be updated to accommodate the new kinds
+of bound object. When one of them is used, it will trigger additional
+authorization checks (described [below](#authorization-checks)), and checks
+on the existence of the bound objects. Likewise, it will be augmented to
+perform authorization checks on the permissions of the service account for
+which the token is requested.
 
-Following is more detail about requests for token acquisition, authorization
-checks by `kube-apiserver` at issuance, and webhook verification.
+Webhook libraries will be updated to optionally (and eventually always)
+require a bearer token. The webhook then verifies these tokens by taking
+the following steps:
+
+1. Verify the token's signature via the OIDC discovery endpoint.
+1. Verify that the token's audience matches the expected audience. This audience
+   is derived deterministically from the webhook configuration. Several
+   alternatives have been discussed including the url, but the tradeoffs
+   are still being evaluated.
+1. Verify that the resource named in the `AdmissionReview` request body is
+   a member of the `APIGroup`(s) named in the private claims. The `*`
+   `APIGroup` is a superset of all `APIGroup`s. When the `APIGroup` is `*`,
+   this check will always succeed.
 
 ### Sequence Diagrams
 
@@ -263,15 +335,18 @@ token verification.
 
 #### Flow 1: `kube-apiserver` as Webhook Client
 
-In this example, a user attempts to create a deployment named "ninja-turtles"
-(user request omitted from diagram). This requires review by the mutating
-admission webhook "mutagen-capsule". In order to authenticate itself to the
-webhook, `kube-apiserer` makes a `TokenRequest` for its dedicated service
-account `kube-system:webhook-auth`, bound to the `MutatingWebhookConfiguration`
-for the "mutagen-capsule" webhook. It binds the token to the webhook, rather
-than the `"v1.apps" APIService`, to avoid the burden of maintaining too
-many tokens. The `kube-system:webhook-auth` service account has `"attest"`
-permissions on the synthetic `"*"` `APIService`.
+In this example, a user attempts to create a deployment named
+"ninja-turtles" (user request omitted from diagram). This requires
+review by the mutating admission webhook "mutagen-capsule". In order to
+authenticate itself to the webhook, `kube-apisever` makes a `TokenRequest`
+for its dedicated service account `kube-system:webhook-auth`, bound to
+the `MutatingWebhookConfiguration` for the "mutagen-capsule" webhook. Note
+that the name for this dedicated service account remains an open question
+([details](#dedicated-service-account-for-token-acquisition)). It requests the
+`"*" APIGroup`, rather than the `"apps" APIGroup`, to avoid the burden of
+maintaining too many tokens, and because `kube-apiserver` is a privileged
+actor. The `kube-system:webhook-auth` service account has `"attest"`
+permissions on the synthetic `APIGroup` resource with the `"*"` name.
 
 ```mermaid
 sequenceDiagram
@@ -281,17 +356,17 @@ sequenceDiagram
 
     KAS->>KAS: Check token cache
     alt Cache miss or token expired
-        KAS->>TokenReq: TokenRequest for "kube-system:webhook-auth"<br/>BoundObjectRef: MutatingWebhookConfiguration "mutagen-capsule"<br/>Audience: "https://mutagen-capsule.default.svc/admission/review"
+        KAS->>TokenReq: TokenRequest for "kube-system:webhook-auth"<br/>BoundObjectRef: MutatingWebhookConfiguration "mutagen-capsule"<br/>Audience: "https://mutagen-capsule.default.svc/admission/review"<br/>AttestationClaims: { "webhook-authentication.k8s.io/allowedAPIGroup": ["*"] }
 
         TokenReq->>TokenReq: Authenticate client
         Note over TokenReq: Authorization
 
         TokenReq->>TokenReq: Can TokenRequest client  principal (`kube-system` as client)<br/>"create" serviceaccounts/token for kube-system:webhook-auth?
         TokenReq->>TokenReq: Does MutatingWebhookConfiguration<br/>"mutagen-capsule" exist?
-        TokenReq->>TokenReq: Does kube-system:webhook-auth have<br/>"attest" on APIService "*"?
+        TokenReq->>TokenReq: Does kube-system:webhook-auth have<br/>"attest" on APIGroup "*"?
 
         TokenReq-->>KAS: Token issued (see payload below)
-        Note over KAS: Cache JWT
+        KAS->>KAS: Cache JWT
     end
 
     Note over KAS,Webhook: AdmissionReview
@@ -300,7 +375,7 @@ sequenceDiagram
     Note over Webhook: Verify JWT
     Webhook->>Webhook: Check signature<br/>(OIDC discovery)<br/>
     Webhook->>Webhook: Check audience
-    Webhook->>Webhook: Check bound objects<br/>Am I a mutating admission webhook?
+    Webhook->>Webhook: Check private claims<br/>Authorized for `"apps" APIGroup`?
 
     Webhook-->>KAS: AdmissionReview response
 ```
@@ -313,20 +388,25 @@ sequenceDiagram
   <...>
   "kubernetes.io": {
       "mutatingWebhookConfiguration": {
-          "name": "mutagen-capusle",
+          "name": "mutagen-capsule",
           "uid": "<uid>"
-      }
+      },
+    "attestationClaims": {
+        "webhook-authentication.k8s.io/allowedAPIGroup": ["*"]
+    }
   }
 }
 ```
 
 #### Flow 2: `kube-apiserver` denies a suspicious request
 
-In this example, a compromised aggregated API server attempts to probe a
-validating webhook, "splinter-validate", for policy information. It wants
-to spam the webhook with `AdmissionReview` requests in an attempt to find
-principals that can read `Secret`s. To do so, it requests a JWT bound to the
-"splinter-validate" `ValidatingWebhookConfiguration`.
+In this example, a compromised aggregated API server attempts to probe
+a validating webhook, "splinter-validate", for policy information. It
+wants to spam the webhook with `AdmissionReview` requests in an attempt
+to find principals that can write `Secret`s. To do so, it requests a JWT
+bound to the "splinter-validate" `ValidatingWebhookConfiguration`,
+requesting that `kube-apiserver` attest to the claim
+`{"webhook-authentication.k8s.io/allowedAPIGroup": ["*"]}`.
 
 ```mermaid
 sequenceDiagram
@@ -342,13 +422,13 @@ sequenceDiagram
     AAS->>AAS: Check JWT cache
     alt Cache miss
         Note over KAS,AAS: Token Acquisition
-        AAS->>KAS: TokenRequest for "turtles-webhook-auth"<br/>BoundObjectRef: ValidatingWebhookConfiguration "splinter-validate"<br/>Audience: https://splinter-validate.default.svc/admission/review
+        AAS->>KAS: TokenRequest for "turtles-webhook-auth"<br/>BoundObjectRef: ValidatingWebhookConfiguration "splinter-validate"<br/>Audience: https://splinter-validate.default.svc/admission/review<br/>AttestationClaims: { "webhook-authentication.k8s.io/allowedAPIGroup": ["*"] }
 
         KAS->>KAS: Authenticate client
         Note over KAS: Authorization
         KAS->>KAS: Can AAS principal "create"<br/>serviceaccounts/token for turtles-webhook-auth?
         KAS->>KAS: Does MutatingWebhookConfiguration<br/>"mutagen-capsule" exist?
-        KAS->>KAS: Does kube-system:webhook-auth have<br/>"attest" on APIService "*"?
+        KAS->>KAS: Does turtles-webhook-auth<br/>have "attest" on APIGroup "*"?
 
         KAS-->>AAS: 403 Forbidden
     end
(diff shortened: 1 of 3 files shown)


keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+purposes, the term **Webhook Authentication Token (WAT)** will be used. However,
+it is important to understand that these are ServiceAccount tokens in every
+sense; but their use is constrained by newly added private claims.
+
+### Webhook Token Acquisition Service Account The service account named in
+WATs will be termed the **Token Acquisition Service Account (WTASA)**. This
+is distinct from the service account the the principal requesting the token
+might be using to authenticate itself to the Kubernetes API Server. The
+Token Acquisition Service Account must have `attest` permissions on the
+`APIService` object named in the `TokenRequest`.
```
liggitt: Historically, tokens only identified, and authorization checks were done by whatever received the token. This is proposing doing the authorization check at token issuance time so the webhook merely has to look for a particular claim in the token?

If we're doing an authorization check on the service account at the time of issuance, would making the bound object and authorization check be related to the Service / ValidatingWebhookConfiguration / MutatingWebhookConfiguration / CustomResourceDefinition the token is going to be used against make more sense than an APIService? That would let us verify the requested audience was coherent with it as well.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+token's `kubernetes.io` private claims include the name and UID of the bound
+APIService, which encodes the API group and version of the resources the
+caller is authorized to consult the webhook about.
+
+The webhook may verify these tokens by taking the following steps:
+
+1. Verify the token's signature via the OIDC discovery endpoint.
+1. Verify that the token's audience matches the expected audience. This audience
+   is derived deterministically from the webhook name, and is in the format is
+   in the format `k8s.io:admission:<webhook-name>`, where `<webhook-name>`
```
liggitt: why the unusual prefix? Isn't audience typically the host / URL where the token will be presented?

Is there a reason not to use `https://$url/with/path` or `https://$servicename.$servicenamespace.svc:port/with/path` as the audience? Webhooks already are required to have valid serving certificates for those hosts, so no new knowledge should be required to make the webhooks able to validate a similar audience in the token.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+   is the "inner" name of the webhook (i.e. the name in the inner list
+   of webhooks).
+1. Verify that the `APIGroup` and `APIVersion` encoded in the token's bound
+   APIService match the `APIGroup` and `Version` of the resource in the body
+   of the `AdmissionReview` request.
```
liggitt: Does this mean a server that serves lots of different API groups/versions (like kube-apiserver) combined with a webhook that intercepts lots of different groups/versions (as a generic label protection or something) would need a distinct token for every API group/version type it sent to the webhook? That's not ideal.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+
+The webhook may verify these tokens by taking the following steps:
+
+1. Verify the token's signature via the OIDC discovery endpoint.
+1. Verify that the token's audience matches the expected audience. This audience
+   is derived deterministically from the webhook name, and is in the format is
+   in the format `k8s.io:admission:<webhook-name>`, where `<webhook-name>`
+   is the "inner" name of the webhook (i.e. the name in the inner list
+   of webhooks).
+1. Verify that the `APIGroup` and `APIVersion` encoded in the token's bound
```
liggitt: A request to `$group/v2` of an API can be converted to `$group/v1` and sent to a webhook that asked to intercept `$group/v1`. See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-matchpolicy

The object sent to the webhook would be in `$group/v1`, the `kind` and `resource` in the AdmissionReview would be `$group/v1` and the `requestKind` and `requestResource` in the AdmissionReview would be `$group/v2`.

What APIService / version would you expect to be in the token sent to a webhook for that?

(I commented elsewhere about the implications of tying the token to a particular presented $group/$version ... we may want to tie it to the target webhook or service)

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+
+## Terms
+
+### Webhook Authentication Token
+This KEP introduces a flow for authenticating `kube-apiserver` and Aggregated
+API Servers to admission webhooks using ServiceAccount tokens specialized
+for this use. To distinguish between ServiceAccount tokens used for other
+purposes, the term **Webhook Authentication Token (WAT)** will be used. However,
+it is important to understand that these are ServiceAccount tokens in every
+sense; but their use is constrained by newly added private claims.
```
liggitt: Maybe it's just me, but I found defining new terms / acronyms for things we already have a bit confusing.

Also, be careful with language like "constrained" unless you really mean these tokens would be invalid to use as service account tokens in other ways because of these claims (which would make them sort of *not* normal service account tokens)

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+
+### Webhook Token Acquisition Service Account The service account named in
```
liggitt:
```suggestion
### Webhook Token Acquisition Service Account
```

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+WATs will be termed the **Token Acquisition Service Account (WTASA)**. This
+is distinct from the service account the the principal requesting the token
```
liggitt:
```suggestion
is distinct from the identity (which may or may not be a service account) the principal requesting the token
```

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+
+#### All webhook authentication clients
+
+When a [webhook authentication client](#webhook-authentication-client) needs
+to call an admission webhook about a given resource, it issues a `TokenRequest`
+for its [webhook token acquisition service account](#webhook-token-acquisition-service-account)
+to the Kubernetes API Server. The request includes:
+
+1. A `BoundObjectRef` pointing to the APIService corresponding to the resource
+   being admitted (e.g., `v1.networking.k8s.io`).
```
liggitt: We don't want to have to mint a token for every webhook invocation, a token should be able to be retained and renewed/rotated after some percentage of its lifetime is used. I was expecting each client to only have to maintain one token per webhook, not one token per resource `$group/$version` per webhook.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+
+The webhook authentication client will only receive the token the authorization
+checks (described in a separate section below) succeed.
+
+#### kube-apiserver
+
+In the case of `kube-apiserver`, the [webhook token acquisition service
+account](#webhook-token-acqcuisition-service-account) will be a discoverable service
+account automatically created in the boostrapping process. The name will be
+randomized to discourage its abuse by other webhook authentication clients.
```
liggitt: We don't randomize built-in service account names for other things ... this seems more confusing than helpful

Rather than randomizing the kube-apiserver serviceaccount, let's just make sure we have workable user stories / examples / documentation / defaults for how:
1. aggregated servers configure their own distinct service account
2. permissions are granted to the server's service account (what does the server admin do, what does the webhook configuration author do, what knowledge does each of those actors have, how does that combine to route permissions to the right spots?)

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+   token only authorizes the caller to consult the webhook about resources in
+   the API group and version named in the token.
+
+### User Stories
+
+#### Story 1: Kube-apiserver authenticates to an admission webhook
+
+A user creates a Pod. The kube-apiserver needs to consult a validating
+admission webhook. It requests a WAT from itself for its dedicated service
+account, bound to APIService `v1.networking.k8s.io` with an audience derived from the
```
liggitt: if it's presenting a pod (Group: "", Version: "v1"), why is it getting a token bound to APIService `v1.networking.k8s.io`?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+a WAT from the Kubernetes API Server. Each aggregated API server should
+have a dedicated service account for this purpose, as it must be named in
+the token request. The request flow is:
+
+1. The aggregated API server authenticates to the kube-apiserver using
+   whatever credential it is configured with. That principal must be authorized
+   to `create serviceaccount/token` in the relevant namespace.
+2. It sends a `TokenRequest` for its dedicated service account, with a
+   `BoundObjectRef` pointing to the APIService it serves (e.g.,
+   `v1.example.com`) and the appropriate audience.
```
liggitt: similar comment about token reuse and wanting to maintain one reusable token per client per webhook, not per resource $group/$version

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,926 @@
...
+2. Does the referenced APIService object actually exist?
+3. Does the [webhook token acquisition service
+   account](#webhook-token-acquisition-service-account) in the request have the
+   `attest` permission on the referenced APIService? This is verified via a
+   SubjectAccessReview-style check (an in-process `authorizer.Authorize()`
+   call) against the service account's identity.
+
+The `attest` verb has precedent (it is already used in Kubernetes for
+ClusterTrustBundle signer attestation). To illustrate the permission model,
+the following RBAC configuration is given as an example. To paraphrase Donald
```
liggitt: let's make sure it's possible to set up permissions correctly given the knowledge of these two personas:
* webhook author / webhook config manifest author who knows the things they are intercepting, but not server identities
* aggregated server admin who knows the identity they are giving the server, but not the webhooks that will need to be called by the server

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: > This is proposing doing the authorization check at token issuance time so the webhook merely has to look for a particular claim in the token?

Yes. In order to avoid the webhook talking back to `kube-apiserver` for `TokenReview` or `SubjectAccessReview`. This is all done to prevent aggregated API servers from probing policy information by asking questions about resources it does not control.

The purpose of using the `APIService` as the bound object is to make it possible for the Kubernetes API Server to check whether the webhook token acquisition service account has permission to talk to the webhook, and whether it may ask do so *in order to ask questions about this APIService*.

With the `{Validating,Mutating}WebhookConfiguration` as the bound object, I can't think of a way to perform the same authorization check as with the `APIService`, without the webhook having to call back to `kube-apiserver`. At token issuance, `kube-apiserver` does not know what question is being asked; it knows the audience and the bound object.

Because it is proposed that the audience be derived from the webhook config, the audience from the `TokenRequest` is used to correlate the `APIService` (from the bound object) and the webhook config corresponding to the audience. From there, it can be determined whether or not the webhook is even relevant to that APIService. Furthermore, `kube-apiserver` can check whether or not the token acquisition service account has the requisite permissions to ask questions *about resources in that APIService*. Without the bound `APIService`, `kube-apiserver` cannot do that check at credential issuance (it doesn't know the question being asked). We can't assume that the webhook author will have knowledge of which API Services it must support, so the webhook must instead answer the question (of whether or not the principal may ask questions about resources in a particular APIService) by making a SAR request to `kube-apiserver`. We want to avoid the webhook calling back to the Kubernetes API server for SAR and TokenReview, for obvious reasons.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: That is an unfortunate consequence.

In practice, aggregated API Servers will typically have one, and almost always fewer than five, relevant `APIService` objects. `kube-apiserver` is a distinct case (could have thousands of APIServices), and I share your concern about the explosion of tokens there.

We could permit `kube-apiserver` to use a "wildcard" APIService as the bound object. During bootstrapping, the requisite permissions (i.e. "attest" on "*" or an equivalent rule) would be set up for kube-apiserver's token acquisition service account. Then `kube-apiserver` only has to maintain one token per webhook, while aggregated API servers should be able to manage with one per webhook/apiservice combination.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: > Maybe it's just me, but I found defining new terms / acronyms for things we already have a bit confusing.

Not my intention to confuse. It helped with the writing to have those terms laid out, so that I have a consistent way to refer to a specific use of something in a particular context. It also helped with conciseness (ex. "token acquisition service account", vs "the service account token named in the `TokenRequest` for a token to authenticate to webhooks"). The latter is hard to work into more complex sentences.

> Also, be careful with language like "constrained" [. . .]

Noted. I'll give this another pass.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: After discussing further with @enj, we propose the following:

Support 3 types of bindings:
- `MutatingWebhookConfiguration`
- `ValidatingWebhookConfiguration`
- `APIService`

When either of the webhook configurations is used as the bound object, the permissions required for token issuance are `"attest"` on `"*"`, because in essence that is what is being permitted by such a token.

Tokens bound to `APIService` won't require such broad permission, and will be narrowly scoped to that `APIService`.

How does this strike you?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Note, rather than using the wildcard `APIService`, we think it's better to allow bindings to either an `APIService` or one of the two `WebhookConfiguration` types (see comment linked below):

https://github.com/kubernetes/enhancements/pull/6156#discussion_r3415121914

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: In this case:
- Aggregated API Servers (likely) don't need to worry about this, since they don't serve custom resources. Any conversion (or request to convert) will have to be done by the aggregated API server itself, so it can just request the token post-conversion.
- kube-apiserver won't need to worry about this either, if it can request (as proposed [in my comment above](https://github.com/kubernetes/enhancements/pull/6156#discussion_r3415121914)) tokens bound to one of the `WebhookConfiguration` types.

Alternatively, we can ignore the `APIVersion` and decide that we only care about the `APIGroup`. I don't think it's ever particularly relevant.

If we *do* decide we care about the version, then this is covered by the webhook: it will check whether or not the (APIGroup, APIVersion) match between those stated in the token and those from the `AdmissionReview` request body.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: > We don't want to have to mint a token for every webhook invocation, a token should be able to be retained and renewed/rotated after some percentage of its lifetime is used.

Yes, the tokens will be cached until their expiration, at which time they will be refreshed. We're in agreement on that. I'll make sure it gets mentioned here.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Agreed. I don't think we can really do anything to prevent people from abusing it, so there's no need to complicate things. I'll update the user stories to include these.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Mistake, thanks for catching. I meant to use a different resource (to avoid the weirdness of having an empty string APIGroup in the example).

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: This will be resolved if we are in agreement on this:

https://github.com/kubernetes/enhancements/pull/6156#discussion_r3415121914

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: This is covered by: 
https://github.com/kubernetes/enhancements/pull/6156#discussion_r3415121914

Basically, webhook config manifest authors can set up a service account with `"attest"` on the `"*"` `APIService` (or equivalent authz check); that allows `TokenRequest`s for that service account to be bound to a `MutatingWebhookConfiguration` or a `ValidatingWebhookConfiguration`.

Aggregated API Server admins will set up a service account with `"attest"` on the relevant `APIService` (or equivalent).

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: The reason for this is that the webhook author may not know the URL where the webhook is deployed, and there will be more plumbing needed on the webhook side to make that information available to the token verification routine.

That said, we don't want this to be a sticking point, so we'll accept your suggestion if you don't agree with the above reasoning.

liggitt: Audience as a URL is more typical ... and since webhooks already are required to have valid serving certificates for their host, it seems more likely to me they can know or be told the host/URL more easily

benjaminapetersen:
/hold 

@pmengelbert lost me in the commit history 😄

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1169 @@
...
+  apiGroup: rbac.authorization.k8s.io
+```
+
+### Token Verification
+
+The webhook may verify these tokens by taking the following steps:
+
+1. Verify the token's signature via the OIDC discovery endpoint.
+1. Verify that the token's audience matches the expected audience. This audience
+   is derived deterministically from the webhook url, and is in the format
```
ahmedtd: Nit --- this is doubled up "is in the format is in the format"

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1169 @@
...
+   is in the format `https://<url>/with/path`, where `<url>` matches that
```
ahmedtd: While this is a good default, I think there are probably setups where the webhook backend expects a different audience?  It might need to be configurable in the Mutating/ValidatingAdmissionWebhook object.

In general, it's the relying party that's in control of which audience value(s) are expected.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1169 @@
...
+
+The `BoundObjectRef` described in 1a are typical of an Aggregated API
+Server, whereas those in 1b are typical for `kube-apiserver`. The webhook
+authentication client will only receive the token if the authorization checks
+(described in a separate section below) succeed.
+
+#### `kube-apiserver`:
+In the case of `kube-apiserver`, the [token acquisition service
+account](#token-acqcuisition-service-account) will be a service with a
+well-known name, `kube-system:webhook-auth`, which is automatically created
```
ahmedtd: I would prefer if we give kube-apiserver *one* identity that it can use for *all* purposes (contacting webhooks, contacting peer replicas, uploading metrics, etc).

If that needs to be a service account, then that's fine, but it should probably be something like `system:serviceaccount:kube-system:kube-apiserver`.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Thanks. This and other things have since been updated. The version I just pushed is much closer to final.

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
```diff
@@ -1,21 +1,26 @@
-title: Per-Pod PID Limit
-kep-number: 6063
+title: KEP Template
```
enj: Fix

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
```diff
@@ -1,21 +1,26 @@
...
+  - "@pmengelbert"
+  - "@benjaminapetersen"
+owning-sig: sig-auth
 participating-sigs:
-  - sig-node
-status: implementable
-creation-date: 2026-05-06
+  - sig-auth
+  - sig-api-machinery
+status: provisional
```
enj: Fix

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
```diff
@@ -1,21 +1,26 @@
...
+  - "@liggitt"
 approvers:
-  - "@mrunalp"
+  - "@enj"
+  - "@deads2k"
+  - "@liggitt"
 
 see-also:
-
 replaces:
```
enj, line 25: List the old ones here.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+Any entity with service network access can send requests to an admission webhook
+endpoint. If the webhook does not authenticate the caller, an attacker can
+probe for policy information, trigger unintended side effects, or exploit
+the webhook's own privileges within the cluster.
+
+Opt-in mechanisms for authenticating the kube-apiserver to webhooks exist
+(client certs, bearer tokens, or basic auth via a kubeconfig file configured
+through `--admission-control-config-file`), but they require manual credential
+management and an API server restart to change. That opt-in mechanism is
+unopinionated as to the method of authentication (mTLS / token / basic auth),
```
enj, line 137: https://github.com/openshift/generic-admission-server should be mentioned.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+servers to query webhooks about resources it controls, but which prevents
+a malicious aggregated API server from requesting policy information about
+resources it does not control.
+
+The scope of this KEP is limited to authenticating to admission webhooks.
+Authentication webhooks, authorization webhooks, and audit webhooks do
+not share the same practical barriers to authentication experienced by
+admission webhooks. For one, those webhooks are not dynamically deployed at
+runtime, and they already require a `kube-apiserver` restart to change their
+configuration. Furthermore, because the actor setting up `kube-apiserver`
```
enj: The authz webhooks config file is dynamically reloaded from disk.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+and the actor setting up the webhook are the same in the vast majority of
+cases, it is much more reasonable to expect that such an actor would use
+the already available solution: they are in control of both the method of
+authentication used by the client and the verification methods used by the
+webhook. This leaves a slight gap, requiring that all deployed aggregated API
+servers that communicate with these webhooks must also have access to the
+necessary credentials. The gap is acknowledged but considered out-of-scope
+to keep the implementation practical for the most common use-cases.
```
enj: Mention that token review and SAR make this a non-issue for everything but audit webhooks.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+  contact the webhook.
+* Tokens may alternatively be scoped per-webhook (by audience).
+* The design is backward compatible: existing kubeconfig-based webhook
+  authentication setups continue to work without modification.
+* Defining the webhook-side verification go library.
+
+### Non-Goals
+
+* Authentication to non-admission webhooks (authentication webhooks,
+  authorization webhooks, audit webhooks).
```
enj, line 219: Lets add conversion webhooks with an explanation of why they are also out of scope.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+## Terms
+
+### Token Acquisition Service Account
+The service account named in tokens for webhook authentication will be termed
+the **Token Acquisition Service Account**. This is distinct from the identity
+that the principal requesting the token uses to authenticate itself to the
+Kubernetes API Server (which may or may not be a service account). The Token
+Acquisition Service Account must have `attest` permissions on the `APIService`
+object named in the `TokenRequest` ().
```
enj: Typo ?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+The `TokenRequest` handler will be updated to accommodate the three new
+kinds of bound object. When one of them is used, it will trigger additional
+authorization checks (described in another section below).
+
+Webhook libraries will be updated to optionally (and eventually always) require a bearer token. The webhook then verifies these tokens by taking the following steps:
+
+1. Verify the token's signature via the OIDC discovery endpoint.
+1. Verify that the token's audience matches the expected audience. This audience
+   is derived deterministically from the webhook url, and is in the format
+   `https://<url>/with/path`, where `<url>` matches the one specified in
```
enj: Open question about what exactly we want to use here / does this need a config knob and how are overlaps handled.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+   types](#webhook-authentication-bound-object-types), and
+     a. When the bound object is a `ValidatingWebhookConfiguration`, reject
+        the request if the webhook is not a validating admission webhook.
+     b. When the bound object is a `MutatingWebhookConfiguration`, reject
+        the request if the webhook is not a mutating admission webhook.
+     c. When the bound object is an `APIService`, reject the request if
+        the resource named in the `AdmissionReview` request body is not a member
+        of the `APIGroup` and `APIVersion` corresponding to that `APIService`.
+
+### Sequence Diagrams
```
enj, line 329: Somewhere you need a discussion about the deadlock scenario.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+  }
+}
+```
+
+#### Flow 2: `kube-apiserver` denies a suspicious request
+
+In this example, a compromised aggregated API server attempts to probe a
+validating webhook, "splinter-validate", for policy information. It wants
+to spam the webhook with `AdmissionReview` requests in an attempt to find
+principals that can read `Secret`s. To do so, it requests a JWT bound to the
```
enj:
```suggestion
principals that can write `Secret`s. To do so, it requests a JWT bound to the
```

Admission doesn't do anything with read requests.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+      "apiService": {
+          "name": "v1.ninja.turtles.ai",
+          "uid": "<uid>"
+      }
+  }
+}
+```
+
+
+### Risks and Mitigations
```
enj, line 524: Mention that "we don't just allow any SA to do this because that is too broad"

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+   `ValidatingWebhookConfiguration`, or a `MutatingWebhookConfiguration`)
+   actually exist?
+1. Service account check:
+     a. when the bound object is an `APIService`, does the [token acquisition
+        service account](#token-acquisition-service-account) have the
+        `attest` permission on that `APIService`?
+     b. When the bound object is one of
+        `{Validating,Mutating}WebhookConfiguration`, does the [token acquisition
+        service account](#token-acuisition-service-account) have `attest`
+        permissions on the wildcard (`"*"`) `APIService`?
```
enj: Ordering of checks and error messages must be careful not to leak state about what objects exist on the cluster.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: let-me-create-webhook-authentication-tokens
+rules:
+  - apiGroups: [""]
+    resources: ["serviceaccount/token"]
+    resourceName: webhook-token-acquisition-service-account
```
enj:
```suggestion
    resourceNames: ["webhook-token-acquisition-service-account"]
```

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+requested must have `"attest"` on the synthetic `"*"` `APIService`. This is not
+a real `APIService`, but is used to express the elevated permissions required
+to obtain a token bound to a *webhook*, for which there is no restriction
+on the `APIService` that the bearer of the token may ask admission questions
+about. This was introduced to this KEP after it was pointed out that caching
+tokens per webhook+APIService combination. The number of tokens in that case
+would be a burden for `kube-apiserver` in particular.
+
+As such, the KEP authors recommend that **only `kube-apiserver`** should be
+granted `"attest"` on `"*"`. Aggregated API servers should instead be granted
```
enj:
```suggestion
As such, the KEP authors recommend that **only `kube-apiserver`'s** service account should be
granted `"attest"` on `"*"`. Aggregated API servers should instead be granted
```

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+    }
+  }
+}
+```
+
+Where the `mutatingWebhookConfiguration` or `validatingWebhookConfiguration`
+does not match the type of webhook receiving the token, the request should
+be rejected.
+
+### Token Verification
```
enj, line 906: Token review semantics need to be described as well.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+
+### Token Caching and Rotation
+
+When the bound object is an `APIService`, serviceaccount tokens for
+authentication to webhooks are cached per combination of webhook and
+`APIService`. When the bound object is a `ValidatingWebhookConfiguration` or
+`MutatingWebhookConfiguration`, the token will be cached per-webhook. When
+a cached token has expired, the next webhook call for that combination
+triggers a new `TokenRequest`. Tokens will expire after 10 minutes,
```
enj: This should just say that it will always be set to 10 mins because anything less is a validation error.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+  is requested on expiry.
+- Feature gate toggling: verify behavior with the gate on and off.
+
+##### e2e tests
+
+- An aggregated API server authenticates to an admission webhook using a
+  JWT bound to each of the three new types.
+- A webhook rejects a request where the APIService claims do not match the
+  resource in the AdmissionReview body.
+- A webhook rejects a request where the bound object does not match the webhook type.
```
enj: Move to int, we only test aggregated API servers via int tests today.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `APIServerWebhookAuthenticationTokenIssuance`
+  - Components depending on the feature gate:
+    - `kube-apiserver`
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `APIServerWebhookAuthenticationTokenVerification`
+  - Components depending on the feature gate:
+    - `kube-apiserver`
```
enj, line 1042: I think we landed on not needing both feature gates because we can ensure that an old API server will never honor these incorrectly.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+New metrics to add:
+- `apiserver_webhook_authentication_token_request_total`: counter of
+  `TokenReview` requests containing a `BoundObjectRef` to one of the webhook
+  authentication bound object types.
+- `apiserver_webhook_authentication_token_request_duration_seconds`:
+  histogram of token request latency, for tokens used to authenticate to
+  webhooks.
+- `apiserver_webhook_authentication_token_cache_hit_total`: counter of
+  cache hits when looking up cached webhook authentication tokens.
```
enj: We need both hit / miss.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+of unique webhook+APIService combinations.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+Yes. Each aggregated API server will have a dedicated service account for
+token requests. `kube-apiserver` will have an additional service account
+for the same purpose. Additional RBAC roles and bindings will be needed.
+
+The JWT itself gains a new field in its private claims (`webhookAuthentication`)
+but this is not stored in etcd.
```
enj: Typo?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+`kube-apiserver`, one JWT per webhook+APIService combination for aggregated API
+servers). CPU impact from token signing is negligible and amortized by caching.
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No. This feature does not affect nodes.
+
+### Troubleshooting
+
+###### How does this feature react if the API server and/or etcd is unavailable?
```
enj, line 1216: This can break aggregated API servers that don't have the token they need to call the webhook.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+###### What are other known failure modes?
+
+- `kube-apiserver` token service account is deleted
+  - Detection: `apiserver_webhook_authentication_token_request_total` with
+    failure label increases.
+  - Mitigations: `kube-apiserver` will need to be restarted, in order for
+    the service account to be recreated as part of the boostrapping process.
+  - Diagnostics: `kube-apiserver` logs will show token request failures.
+  - Testing: Integration tests cover SA deletion and recreation.
```
enj, line 1238: KAS should automatically recreate it if it does not exist.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+
+### Troubleshooting
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+If `kube-apiserver` is unavailable, no webhook calls are made and the feature
+is moot. If etcd is unavailable, the dedicated service account and APIService
+objects cannot be read, and token issuance will fail. Webhook calls will
+proceed without a token (or fail, depending on the webhook's configuration).
+
```
enj, line 1229: We need to be more specific on when we will call a webhook without a token.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+1. Check `apiserver_webhook_authentication_token_request_total` for token
+   request failures.
+2. Check `apiserver_admission_webhook_rejection_count` for webhook
+   rejections.
+3. Check `apiserver_admission_webhook_latency_seconds` for increased
+   latency.
+4. Verify the dedicated SA exists and has the correct RBAC permissions.
```
enj: Include the relevant `kubectl` commands.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+* `kube-apiserver` authorizes webhook authentication clients at token issuance,
+  and will refuse to provide a token to unauthorized principals.
+* Setting up the requisite permissions for token acquisition should be simple
+  and easy.
+* The token is scoped to a subset of resources about which its bearer may
+  contact the webhook.
+* Tokens may alternatively be scoped per-webhook (by audience).
+* The design is backward compatible: existing kubeconfig-based webhook
+  authentication setups continue to work without modification.
+* Defining the webhook-side verification go library.
```
enj, line 210: Must support both client-go and controller-runtime.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -0,0 +1,1128 @@
...
+are two successful flows and one flow that fails at the time of webhook
+token verification.
+
+#### Flow 1: `kube-apiserver` as Webhook Client
+
+In this example, a user attempts to create a deployment named "ninja-turtles"
+(user request omitted from diagram). This requires review by the mutating
+admission webhook "mutagen-capsule". In order to authenticate itself to the
+webhook, `kube-apiserer` makes a `TokenRequest` for its dedicated service
+account `kube-system:webhook-auth`, bound to the `MutatingWebhookConfiguration`
```
enj: Open question about the exact name of this SA and if we plan to use it for other things in the future to represent KAS.

enj (changes requested):
First pass.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: [GK](https://github.com/open-policy-agent/gatekeeper/blob/9db1ea8104bf38de97646e7e68da1a12effbb55c/deploy/gatekeeper.yaml#L5703-L5742) inner webhook names as a potential choice.

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
pmengelbert: Clarify what needs fixing here. It's just SIG Auth?
Do I mark status: implementable yet?

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
pmengelbert: done (I think, double check please).

keps/sig-auth/6060-api-server-authentication-to-webhooks/kep.yaml
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Can you remind me why it was decided that it was fine to leave them out? I'm not remembering exactly.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Risks and Mitigations seems like the proper place for it.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: I understand the error messages part, but how does the ordering of checks affect this?

"Ordering of checks and error messages" is ambiguous. Do you mean
1. (Ordering of checks) and (error messages), or  
1. (Ordering of checks) and (ordering of error messages)?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: mentioned that alternatives are being evaluated. done.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: MIddle-click monstrosity.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: You have the recording 🙂

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: addressed in rewrite

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: please respond

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: added to design details section

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: removed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: incorporated this in the rewrite

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done, but not sure if correct.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -185,6 +209,12 @@ to keep the implementation practical for the most common use-cases.
 * The design is backward compatible: existing kubeconfig-based webhook
   authentication setups continue to work without modification.
 * Defining the webhook-side verification go library.
+* Supporting both `client-go` and `controller-runtime`.
+* Tokens must be reviewable by `TokenReview`, although the webhook
+  library **WILL NOT** use `TokenReview` by default to avoid unnecessary
+  round trips.
+* The token claims must be reviewable by `SubjectAccessReview`, but the
+  webhook library will not use `SubjectAccessReview` by default.
```
enj: What does this mean?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -124,43 +136,55 @@ through `--admission-control-config-file`), but they require manual credential
...
+up `kube-apiserver` and the actor setting up the webhook are the same in the
+vast majority of cases. Therefore, it is much more reasonable to expect that
+such an actor would use the already available solution: they are in control
+of both the method of authentication used by the client and the verification
+methods used by the webhook. This leaves a slight gap, requiring that all
+deployed aggregated API servers that communicate with these webhooks must
+also have access to the necessary credentials. The gap is acknowledged
+but considered out-of-scope to keep the implementation practical for the
+most common use-cases. `TokenReview` and `SubjectAccessReview` make this a
+non-issue for everything but audit webhooks.
```
enj: .

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -220,73 +257,76 @@ terms **`kube-apiserver`** and **Kubernetes API Server** will be used
...
+The `TokenRequest` API will be expanded to accommodate the scoping of service
+account tokens to a particular usage. A brief description of those expansions
+is in order. To obtain a [token](#token), the webhook authentication
+client will make a `TokenRequest` on a service account. For a client to
+obtain a token, it must meet four conditions. First, it must request claims
+indicating which `APIGroups` it intends to query the webhook about. Further
+details are described in the [design details](#changes-to-tokenrequestspec)
+section. Second, it must specify either a `ValidatingWebhookConfiguration`
+or a `MutatingWebhookConfiguration` as the `BoundObjectRef`. Third, it must
+specify an audience that is coherent for that `*WebhookConfiguration`. The
```
enj, line 284: And the `APIGroups` much be coherent with the webhook as well.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -220,73 +257,76 @@ terms **`kube-apiserver`** and **Kubernetes API Server** will be used
...
-     b. When the bound object is a `MutatingWebhookConfiguration`, reject
-        the request if the webhook is not a mutating admission webhook.
-     c. When the bound object is an `APIService`, reject the request if
-        the resource named in the `AdmissionReview` request body is not a member
-        of the `APIGroup` and `APIVersion` corresponding to that `APIService`.
+   is derived deterministically from the webhook configuration. Several
+   alternatives have been discussed including the url, but the tradeoffs
+   are still being evaluated.
+1. Verify that the JWT is bound to one (and only one) of
+   `ValidatingWebhookConfiguration` or `MutatingWebhookConfiguration`.
```
enj: Drop this -> this becomes a detail of how the credential was issued.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -482,27 +539,118 @@ was minted.
...
 #### Increased authorization load
 
-Each  request triggers an additional authorization check (the `attest`
+Each request triggers an additional authorization check (the `attest`
 verification). This is mitigated by caching: tokens are cached for their
 lifetime, so the authorization check is amortized over many webhook calls.
 
+#### Potential Deadlock
+
+If an admission webhook is configured to intercept `TokenReview` requests,
```
enj: `TokenRequest`

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -482,27 +539,118 @@ was minted.
...
+**and** the webhook requires an authentication token, there will be a deadlock
+with no way to proceed. This will be mitigated by designing authentication in
+such a way that the webhook authentication client attempts to proceed without
```
enj: No for this case we just accept the deadlock until the webhook config is fixed.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -482,27 +539,118 @@ was minted.
...
+### Changes to `TokenRequest` API
+
+#### Changes to `TokenRequestSpec`
+
+`TokenRequestSpec` will be augmented to allow for the specification of
+`AttestationClaims`, with a type of `map[string][]string`. This is modeled
+after the `UserInfo.Extra` field, which has served well for its purpose and
+proven flexible.
+
+The keys to this map represent  the names of claims to which the client wishes
```
enj: Too many spaces

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -482,27 +539,118 @@ was minted.
...
+`kube-apiserver` to attest. The values represent the specifics of the claim. The
+keys will be well-known. Arbitrary keys may not be used. Keys not recognized
+by `kube-apiserver` will result in a rejection of the `TokenRequest`. The
+addition of keys recognized by `kube-apiserver` will be subject to API review.
+The meaning of the value or values will be well-defined.  The same value
+may change meaning when it corresponds to a different key, but the meaning
+within the purview of each key (claim name) must be well-defined.
+
+Claims will be namespaced by a prefixed domain. That is, they will be of
+the form: `<FQDN>/<name>`.
```
enj: They are already nested under kubernetes.io so I don't think we need the FQDN?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: Now explicitly noted as an open question.

The meeting with leads on this KEP resulted in the streamlining of some design details. It's a big enough change to warrant a re-read when you are able.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -482,27 +539,118 @@ was minted.
...
+improperly formed.
+
+The special wildcard value `"*"` is recognized here to mean the token
+will be valid for **all `APIGroup`s**. Any other value represents a
+single `APIGroup` for which the resulting token will be valid. When not a
+wildcard, the `APIGroup` must correspond to at least one real `APIService`
+matched on the `APIService`'s `APIGroup` field (there may be multiple,
+owing to different versions of the same `APIGroip`, but only one is
+required). `"webhook-authentication.k8s.io/allowedAPIGroup"` values that
+match only to deleted or nonexistent `APIServices` will be rejected.
```
enj: ?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -512,60 +660,77 @@ to call an admission webhook about a given resource, it issues a `TokenRequest`
...
 are described in the next section.
 
-We expect each aggregated API server to have its own dedicated service account
-for obtaining tokens it will use to authenticate to webhooks. Reuse of these
-service accounts across multiple aggregated API servers is discouraged.
+We expect each aggregated API server to have its own dedicated service
+account for obtaining tokens it will use to authenticate to webhooks. Reuse
+of these service accounts across multiple aggregated API servers is strongly
+discouraged.  Reuse of `kube-apiserver`'s service account for by aggregated
+API servers is not only discouraged; it is furthermore condemned.
```
enj: 😆

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -623,18 +801,22 @@ roleRef:
...
 metadata:
   name: let-the-webhook-token-acquisition-service-account-request-tokens-bound-to-an-api-service
 rules:
-  - apiGroups: ["apiregistration.k8s.io"]
-    resources: ["apiservices"]
-    resourceName: "v1.example.com"
+  - apiGroups: ["webhook-authentication.k8s.io"]
+    resources: ["apigroups"]
+    resourceNames: ["jungle.panda"]
     verbs: ["attest"]
```
enj, line 808: Will need to bike shed on the specifics here.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -739,29 +917,35 @@ The webhook may verify these tokens by taking the following steps:
...
+`TokenReview`. Nevertheless, it must be possible for bearers or receivers
+of these specialized service account tokens to perform token validation via
+`TokenReview`. Therefore, object-existence checks will need to be added for
+the cases when the private claims contain a `"mutatingWebhookConfiguration"`
+or `"validatingWebhookConfiguration"` field.
+
+`TokenReview` will also validate the existence of at least one `APIService`
+whose `APIGroup` field matches exactly the value corresponding to the
+`"webhook-authentication.k8s.io/allowedAPIGroup"` key in the JWT's
+`"attestationClaims"` private claim.
```
enj, line 933: Not sure we want to do that.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -974,6 +1153,8 @@ New metrics to add:
   webhooks.
 - `apiserver_webhook_authentication_token_cache_hit_total`: counter of
   cache hits when looking up cached webhook authentication tokens.
+- `apiserver_webhook_authentication_token_cache_miss_total`: counter of
+  cache misses when looking up cached webhook authentication tokens.
```
enj: One metric with a label like `rest_client_transport_create_calls_total`

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
```diff
@@ -1074,13 +1260,15 @@ proceed without a token (or fail, depending on the webhook's configuration).
...
    rejections.
-3. Check `apiserver_admission_webhook_latency_seconds` for increased
+1. Check `apiserver_admission_webhook_latency_seconds` for increased
    latency.
-4. Verify the dedicated SA exists and has the correct RBAC permissions.
-5. If the problem cannot be resolved, disable the
-   `APIServerWebhookAuthenticationTokenIssuance` feature gate and restart
+1. Verify the dedicated SA exists and has the correct permissions (possibly
+   RBAC): `kubectl get clusterrolebinding sa-can-attest-to-star && kubectl
+   get clusterrole can-attest-to-star`
```
enj: `kubectl auth can-i educate dolphins --as=system:serviceaccount:... -n ...`

enj (changes requested):
Another pass.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: More detail around authz bits.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: Did not see it in my pass.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: Was it?

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
enj: As in, don't allow an actor to probe for the existence of the reference via error messages.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: noted for followup

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: done

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: removed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: agreed

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed.

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: double check this

keps/sig-auth/6060-api-server-authentication-to-webhooks/README.md
pmengelbert: fixed, but double check.

deads2k:
For sig auth and prr lgtm 

/approve

Issue #5978, opened 2026-03-24
DRA: ClusterResourceClaimTemplate


johnbelamaric:
### Enhancement Description

One of the challenges with using DRA is that it requires the creation of a separate ResourceClaimTemplate object in addition to the standard workload definition. Often, the ResourceClaimTemplate represents a common pattern that could be used across many workloads. But since it is a namespaced object, a copy must be populated in every namespace that might need to use that pattern. Instead, a cluster-scoped object could be used to allow administrators to capture common and complex patterns once, and let them be re-used across namespaces.

/assign @everpeace 
/cc @pohly @klueska @andreyvelich 

- One-line enhancement description (can be used as a release note): ClusterResourceClaimTemplates can be defined by the administrator once in the cluster scope, and used by workloads to create namespaced ResourceClaims.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6159
- Discussion Link: 
- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6159
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

johnbelamaric:
@liggitt Ideally, I'd like to be able to use the existing Pod and PodGroup fields that refer to a `ResourceClaimTemplateName` and have the controller check for a ClusterResourceClaimTemplate of that name if a namespace-local ResourceClaimTemplate does not exist. Is that a pattern with precedence in our APIs?

everpeace:
@johnbelamaric Thank you for opening the issue!! This enhancement focuses on something like Day2-ish operational issues. But, I feel this is an important UX/Admin-experience improvement.

For the API design, my initial API idea is:
- keeping `ResourceClaimTemplateName` field for backward compatibility
- introducing a new `ResourceClaimTemplateRef` field in the neighbor of `ResourceClaimTemplateName` so that it can refer to both cluster-scoped and namespace-scoped ones like this:
  - ```yaml
    # Case 1: namespace-scoped
    resourceClaimTemplateRef:
      kind: ResourceClaimTemplate
      name: namespaced-resource-claim-template
    # Case 2: cluster-scoped
    resourceClaimTemplateRef:
      kind: ClusterResourceClaimTemplate
[...]

PR #6159, opened 2026-06-04
KEP-5978: DRA ClusterResourceClaimTemplate

troychiu:
- One-line PR description: ClusterResourceClaimTemplates can be defined by the administrator once in the cluster scope, and used by workloads to create namespaced ResourceClaims.

- Issue link: https://github.com/kubernetes/enhancements/issues/5978

- Other comments: Gemini helped create this PR

### DIFF
--- a/keps/sig-node/5978-cluster-resource-claim-template/README.md
+++ b/keps/sig-node/5978-cluster-resource-claim-template/README.md
@@ -67,6 +67,8 @@ SIG Architecture for cross-cutting KEPs).
 -->
 # KEP-5978: DRA: ClusterResourceClaimTemplate
 
+**NOTE:** This KEP was withdrawn in favor of [the Out-of-Tree Synchronization Controller alternative](#out-of-tree-alternative).
+
 <!--
 This is the title of your KEP. Keep it short, simple, and descriptive. A good
 title can help communicate what the KEP is and should be considered as part of
@@ -1094,6 +1096,7 @@ Major milestones might include:
 -->
 
 - **2026-06-02**: Initial KEP drafted and proposed targeting v1.37 Alpha.
+- **2026-06-16**: KEP withdrawn in favor of an out-of-tree controller solution.
 
 ## Drawbacks
 
@@ -1121,6 +1124,29 @@ information to express the idea and why it was not acceptable.
     silently override/shadow the cluster-scoped template, changing Pod
     provisioning paths without explicit audit trails. Dedicated fields make
     references completely explicit and easily auditable.
+
+<a id="out-of-tree-alternative"></a>
+- **Out-of-Tree Synchronization Controller / CRD**: Create an out-of-tree
+  controller (for example, in a `kubernetes-sigs` repository) that automatically
+  generates namespaced `ResourceClaimTemplate` objects in target namespaces. This
+  could be driven either by a `ClusterResourceClaimTemplate` CRD or by publishing
+  master `ResourceClaimTemplate` objects in a centralized controller namespace
+  using a special naming convention or annotation mapping. Pods would then
+  refer to the generated namespaced templates by name.
+  - *Trade-offs / Motivation for Withdrawal*: While this approach involves certain
+    trade-offs compared to an in-tree API:
+    - **Race Conditions**: When a new namespace is created, Pods referencing the
+      template might be rejected or fail to deploy if they are created before the
+      out-of-tree controller finishes copying the template into the new namespace.
+    - **Resource Proliferation and Overhead**: Duplicating identical template
+      objects across tens or hundreds of namespaces creates unnecessary `etcd`
+      storage overhead and additional API server load.
+
+    Despite these trade-offs, the SIG decided to withdraw this KEP in favor of this
+    out-of-tree controller approach. The alternative is considered good enough to
+    fulfill the target use cases, and implementing an in-tree solution does not
+    justify the required engineering effort and the expansion of the core
+    Kubernetes API surface.
 
 ## Infrastructure Needed (Optional)
 
--- a/keps/sig-node/5978-cluster-resource-claim-template/README.md
+++ b/keps/sig-node/5978-cluster-resource-claim-template/README.md
@@ -1,0 +1,1102 @@
+<!--
+**Note:** When your KEP is complete, all of these comment blocks should be removed.
+
+Follow the guidelines of the [documentation style guide].
+In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+
+To get started with this template:
+
+- [ ] **Pick a hosting SIG.**
+  Make sure that the problem space is something the SIG is interested in taking
+  up. KEPs should not be checked in without a sponsoring SIG.
+- [ ] **Create an issue in kubernetes/enhancements**
+  When filing an enhancement tracking issue, please make sure to complete all
+  fields in that template. One of the fields asks for a link to the KEP. You
+  can leave that blank until this KEP is filed, and then go back to the
+  enhancement and add the link.
+- [ ] **Make a copy of this template directory.**
+  Copy this template into the owning SIG's directory and name it
+  `NNNN-short-descriptive-title`, where `NNNN` is the issue number (with no
+  leading-zero padding) assigned to your enhancement above.
+- [ ] **Fill out as much of the kep.yaml file as you can.**
+  At minimum, you should fill in the "Title", "Authors", "Owning-sig",
+  "Status", and date-related fields.
+- [ ] **Fill out this file as best you can.**
+  At minimum, you should fill in the "Summary" and "Motivation" sections.
+  These should be easy if you've preflighted the idea of the KEP with the
+  appropriate SIG(s).
+- [ ] **Create a PR for this KEP.**
+  Assign it to people in the SIG who are sponsoring this process.
+- [ ] **Merge early and iterate.**
+  Avoid getting hung up on specific details and instead aim to get the goals of
+  the KEP clarified and merged quickly. The best way to do this is to just
+  start with the high-level sections and fill out details incrementally in
+  subsequent PRs.
+
+Just because a KEP is merged does not mean it is complete or approved. Any KEP
+marked as `provisional` is a working document and subject to change. You can
+denote sections that are under active debate as follows:
+
+```
+<<[UNRESOLVED optional short context or usernames ]>>
+Stuff that is being argued.
+<<[/UNRESOLVED]>>
+```
+
+When editing KEPS, aim for tightly-scoped, single-topic PRs to keep discussions
+focused. If you disagree with what is already in a document, open a new PR
+with suggested changes.
+
+One KEP corresponds to one "feature" or "enhancement" for its whole lifecycle.
+You do not need a new KEP to move from beta to GA, for example. If
+new details emerge that belong in the KEP, edit the KEP. Once a feature has become
+"implemented", major changes should get new KEPs.
+
+The canonical place for the latest set of instructions (and the likely source
+of this file) is [here](/keps/NNNN-kep-template/README.md).
+
+**Note:** Any PRs to move a KEP to `implementable`, or significant changes once
+it is marked `implementable`, must be approved by each of the KEP approvers.
+If none of those approvers are still appropriate, then changes to that list
+should be approved by the remaining approvers and/or the owning SIG (or
+SIG Architecture for cross-cutting KEPs).
+-->
+# KEP-5978: DRA: ClusterResourceClaimTemplate
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1: Exposing and Consuming Specialized Hardware Cluster-Wide](#story-1-exposing-and-consuming-specialized-hardware-cluster-wide)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [<code>ResourceClaim</code> Controller Changes](#resourceclaim-controller-changes)
+  - [Admission &amp; Validation](#admission--validation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+<!--
+**ACTION REQUIRED:** In order to merge code into a release, there must be an
+issue in [kubernetes/enhancements] referencing this KEP and targeting a release
+milestone **before the [Enhancement Freeze](https://git.k8s.io/sig-release/releases)
+of the targeted release**.
+
+For enhancements that make changes to code or processes/procedures in core
+Kubernetes—i.e., [kubernetes/kubernetes], we require the following Release
+Signoff checklist to be completed.
+
+Check these off as they are completed for the Release Team to track. These
+checklist items _must_ be updated for the enhancement to be released.
+-->
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+<!--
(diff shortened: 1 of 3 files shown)


johnbelamaric (commented):
Nice, thanks!

troychiu:
cc @mortent @pohly for review. Thank you.

keps/sig-node/5978-cluster-resource-claim-template/kep.yaml
```diff
@@ -0,0 +1,42 @@
+title: "DRA: ClusterResourceClaimTemplate"
+kep-number: 5978
+authors:
+  - "@troychiu"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scheduling
```
macsko: Why SIG Scheduling is a participating SIG here? Is there any feedback needed from scheduling perspective?

johnbelamaric: I think you're right, we can remove it.

troychiu: Updated. Thank you!

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Currently, `ResourceClaimTemplate` is namespace-scoped. Administrators managing
+clusters with specialized hardware (GPUs, FPGAs, high-performance network
+devices) must deploy and synchronize identical `ResourceClaimTemplate` resources
```
nojnhuh: Are cluster administrators really doing this? ResourceClaims and ResourceClaimTemplates are inherent properties of a workload, not a cluster. ResourceSlices are the authoritative place to look for what's actually available to request.

If I'm a workload author, why would I prefer to use only a name mapping to an opaque set of resources that may or may not change from cluster to cluster instead of a literal ResourceClaimTemplate managed with the rest of the workload if they're functionally equivalent?

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+solve this by allowing a single central template definition.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+- Introduce a new cluster-scoped API resource: `ClusterResourceClaimTemplate`.
```
nojnhuh: Could these goals be rephrased to focus on the higher-level problems being targeted? Now they're mostly describing implementation details.

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+We propose introducing a new cluster-scoped API resource,
+`ClusterResourceClaimTemplate`, which serves as a cluster-wide template for
+generating `ResourceClaims`. Along with this, we propose adding a new
+`clusterResourceClaimTemplateName` field to `PodResourceClaim` alongside the
```
nojnhuh: Can you include whether or not this field will be added to the `spec.resourceClaims` field for PodGroups as well?

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
+
+We propose introducing a new cluster-scoped API resource,
+`ClusterResourceClaimTemplate`, to allow cluster administrators to define
```
nojnhuh: nit: API objects should not be formatted as `code`.
https://kubernetes.io/docs/contribute/style/style-guide/#code-style-inline-code

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+We propose introducing a new cluster-scoped API resource,
+`ClusterResourceClaimTemplate`, which serves as a cluster-wide template for
+generating `ResourceClaims`. Along with this, we propose adding a new
+`clusterResourceClaimTemplateName` field to `PodResourceClaim` alongside the
+existing `resourceClaimName` and `resourceClaimTemplateName` fields. 
+
+When `clusterResourceClaimTemplateName` is populated in a Pod specification, the
+`ResourceClaim` controller fetches the referenced `ClusterResourceClaimTemplate`
+and generates a corresponding `ResourceClaim` in the Pod's namespace,
+maintaining standard resource ownership and garbage collection semantics.
```
nojnhuh, line 256: Have you thought about how this might look for shareable ResourceClaims? One way to do that would be to reference a ClusterResourceClaimTemplate in a PodGroup. Without that PodGroup boundary, something like a ClusterResourceClaim could replicate once per namespace, but that might not be granular enough in all cases.

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+```
+
+### `ResourceClaim` Controller Changes
+
+`ResourceClaimController` will be updated to:
+- Fetch `ClusterResourceClaimTemplate` resources using a cluster-level template
+  informer.
+- When a `PodResourceClaim` references a `ClusterResourceClaimTemplateName`,
+  create a namespaced `ResourceClaim` in the Pod's namespace using
+  `template.Spec.Spec`, and attach standard OwnerReferences mapping it to the
```
nojnhuh: nit: there is no `template` field defined above (like there isn't for ResourceClaimTemplate), so this should only be `spec.spec`, or `Spec.Spec` if we prefer the Go names.

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+The data can be easily read from:
+https://testgrid.k8s.io/sig-testing-canaries#ci-kubernetes-coverage-unit
+
+This can inform certain test coverage improvements that we want to do before
+extending the production code to implement this enhancement.
+-->
+
+- `k8s.io/kubernetes/pkg/apis/core/validation`: `85.5%`
+- `k8s.io/kubernetes/pkg/apis/resource/validation`: `96.6%`
+- `k8s.io/kubernetes/pkg/controller/resourceclaim`: `75.0%`
```
nojnhuh, line 500: `k8s.io/dynamic-resource-allocation/resourceclaim` might also be good to mention here.

keps/sig-node/5978-cluster-resource-claim-template/README.md
johnbelamaric: This is a very reasonable question, and I don't have a data-driven answer, though it has come up a couple of times (e.g., https://github.com/kubernetes/enhancements/issues/5491#issuecomment-4560832119 and https://github.com/kubernetes/kubernetes/issues/128440).

That said, I think you could ask the same about DeviceClass, too, though - or a number of other cluster-scoped resources. One of the complaints about DRA is complexity, and so we're trying to provide another mechanism to push complexity away from the end user.

nojnhuh: I can see the value in having a kind of library of ResourceClaims to show common usage patterns, but I wonder if documenting those better would provide most of that benefit without any new APIs or functional surface area. Another alternative could be a CRD+controller like ComputeDomain that generates ResourceClaims to handle common use cases with a simpler API.

troychiu: I am not sure if my understanding is correct, but I imagine some workload authors often use platforms built on top of Kubernetes and don't interact with Kubernetes directly. This KEP provides an easy way for platform administrators to manage templates across the entire cluster, as the motivation mentioned.

nojnhuh: Whether or not there's some higher level abstraction in play doesn't affect the point about ResourceClaims being properties of the workload and not the cluster at the Kubernetes layer. If someone or something is managing lots of workloads across many namespaces, then they must already have some templating or automation set up that ResourceClaims should neatly fit into? Otherwise we could make the same convenience argument for any namespace-scoped object.

johnbelamaric: It's a reasonable challenge to the idea. @pohly @SergeyKanzhelev @mrunalp @everpeace @thameem-abbas any input?

pohly: I tend to agree with @nojnhuh.

ClusterResourceClaimTemplate makes sense when it will be used as an additional abstraction between cluster admin and users:
- Admin creates ClusterResourceClaimTemplate with a certain well-known name and a guarantee about what it does. That same ClusterResourceClaimTemplate will be present in different clusters.
- Users reference that well-known name when deploying workloads, making them more portable across different environments.

DeviceClass is used like this ("if you want a GPU X, use DeviceClass A").

But when ClusterResourceClaimTemplate and workload get created by the same person, it becomes just a redundant way of describing a workload: perhaps a bit easier to understand when looking at a deployed workload, but not really needed and (IMHO) not justifying the effort and additional Kubernetes API surface.

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+
+Consider the following in developing an upgrade/downgrade strategy for this
+enhancement:
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to maintain previous behavior?
+- What changes (in invocations, configurations, API use, etc.) is an existing
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+- **Upgrade**: Enabling the `ClusterResourceClaimTemplates` feature gate
```
johnbelamaric, line 628: feature gate and api enablement are actually independent, you would need both

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1102 @@
...
+###### How can an operator determine if the feature is in use by workloads?
+
+<!--
+Ideally, this should be a metric. Operations against the Kubernetes API (e.g.,
+checking if there are objects with field X set) may be a last resort. Avoid
+logs or events for this purpose.
+-->
+
+An operator can list `ClusterResourceClaimTemplate` objects or query
+`kube-apiserver` request metrics for `clusterresourceclaimtemplates`.
```
johnbelamaric, line 835: or kube-state-metrics

johnbelamaric (commented):
PRR looks ok, but needs sig approval

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Thanks for catching this. Yes I think we should add this field to podgroup as well. Updated.

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Updated. Thank you!

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Updated. Thank you

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Sure. Updated. Thank you

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Yes that will be better. Updated. Thank you

keps/sig-node/5978-cluster-resource-claim-template/README.md
troychiu: Yes, my idea is to treat ClusterResourceClaimTemplate similarly to ResourceClaimTemplate, but as a cluster-scoped API. We should therefore allow PodGroup to reference a ClusterResourceClaimTemplate. By referencing the template in a PodGroup, the controller would generate a single ResourceClaim owned by the PodGroup, which all member pods could then share.

I just updated the KEP to integrate ClusterResourceClaimTemplate with PodGroup. Please let me know what you think.

keps/sig-node/5978-cluster-resource-claim-template/kep.yaml
```diff
@@ -1,27 +1,16 @@
...
+# participating-sigs:
 status: implementable
-creation-date: 2026-04-27
+creation-date: "2026-06-02"
 reviewers:
-  - "@msau42"
-  - "@jsafrane"
-  - "@tallclair"
-  - "@ndixita"
+  - "@mortent"
```
everpeace, line 11: Would you mind adding me as a reviewer??

keps/sig-node/5978-cluster-resource-claim-template/kep.yaml
```diff
@@ -31,24 +20,22 @@ stage: alpha
...
 
 # The following PRR answers are required at alpha release
 # List the feature gate name and the components for which it must be enabled
 feature-gates:
-  - name: InPlacePodVerticalScalingMemoryBackedVolumes
+  - name: ClusterResourceClaimTemplates
     components:
       - kube-apiserver
-      - kubelet
+      - kube-controller-manager
```
everpeace, line 38: Shouldn't we add `kube-sheduler` to the affected components?

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1134 @@
...
+Because a ClusterResourceClaimTemplate is a template containing a declarative
+`ResourceClaimSpec` and does not grant any special capabilities on its own, a
+user could always copy the template's content and create an equivalent
+namespaced ResourceClaimTemplate or inline ResourceClaim. Enforcing access
+control on the actual underlying hardware or scheduling resources must be done
+at the resource driver level or via namespace resource quotas/limits.
+
+During admission, the API server will perform standard structural validation on
+the `clusterResourceClaimTemplateName` field in both `PodResourceClaim` and `PodGroupResourceClaim` (e.g. checking that only one of the
+template/claim name fields is populated).
```
everpeace, line 450: Should we perform name formatting validation for this field? I supposed [`+k8s:format=k8s-long-name`](https://kubernetes.io/docs/reference/using-api/declarative-validation/#tag-format)

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1134 @@
...
+  cluster required to make on upgrade, in order to make use of the enhancement?
+-->
+
+- **Upgrade**: Enabling the `ClusterResourceClaimTemplates` feature gate
+  registers the ClusterResourceClaimTemplate API. The feature is opt-in and
+  does not affect existing workloads.
+- **Downgrade**: Disabling the feature gate will prevent the creation of new
+  ClusterResourceClaimTemplate objects, and Pod or PodGroup creation requests referencing
+  them will be rejected. Already-generated, namespaced ResourceClaims remain
+  unaffected and continue to function as standard claims.
```
everpeace, line 634: > Pod or PodGroup creation requests referencing them will be rejected

This means that any deployment/statefulset rollout, or new pod creation initiated by evictions/preemptions will be rejected. Is this intended?

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1134 @@
...
+}
+```
+
+### ResourceClaim Controller Changes
+
+`ResourceClaimController` will be updated to:
+- Fetch ClusterResourceClaimTemplate resources using a cluster-level template
+  informer.
+- When a `PodResourceClaim` references a `ClusterResourceClaimTemplateName`,
+  create a namespaced ResourceClaim in the Pod's namespace using the ClusterResourceClaimTemplate's
```
everpeace, line 426: Should we discuss the naming convention for generated `ResourceClaim`s??

What if there is a namespaced `ResourceClaimTemplate` with the same name of the referenced `ClusterResourceClaimTemplate`?? 

Technically, the same naming convention of `ResourceClaimTemplate` will work as expected. But, because this might confuse users a bit, it might be better to explore the different naming convention?

Even if we apply the same naming convention, I think we should put some signal for users to know its origin(`ResourceClaimTemplate` or `ClusterResourceTemplate`) on the generated `ResourceClaim` at least.

everpeace (commented):
@troychiu Thanks for your write-up!! I put several comments. PTAL 🙇

mrunalp (approved):

keps/sig-node/5978-cluster-resource-claim-template/README.md
```diff
@@ -0,0 +1,1134 @@
...
+<!--
+Why should this KEP _not_ be implemented?
+-->
+
+Introducing ClusterResourceClaimTemplate adds another API resource to
+`resource.k8s.io` and expands the `PodResourceClaim` API structure. However,
+this is justified by the significant simplification it brings to cluster-scoped
+DRA configuration and resource control.
+
+## Alternatives
```
nojnhuh, line 1112: The main alternative I see to this is doing this as a separate CRD/controller out-of-tree rather than in-tree. If we're sure this has to be in-tree, can we describe why here?

Presumably, a controller could take an instance of some ClusterResourceClaim CRD and generate a ResourceClaimTemplate with a specific name in whichever namespaces. Then Pods can refer to the generated ResourceClaimTemplate by its name. That might not be as air-tight as a Pod referring directly to a ClusterResourceClaim, but I wonder if that's good enough. Does that otherwise fulfill the use cases being targeted here?

keps/sig-node/5978-cluster-resource-claim-template/README.md
everpeace: > This is a very reasonable question, and I don't have a data-driven answer, though it has come up a couple of times (e.g., https://github.com/kubernetes/enhancements/issues/5491#issuecomment-4560832119 and https://github.com/kubernetes/kubernetes/issues/128440).
> 
> That said, I think you could ask the same about DeviceClass, too, though - or a number of other cluster-scoped resources. One of the complaints about DRA is complexity, and so we're trying to provide another mechanism to push complexity away from the end user.

Yes, my original motivation for proposing this was to reduce the complexity of defining _complex_ and _common_ `ResourceClaim`, e.g., topology aligned `gpu/nic` pair or `gpu/core/nic` triples, for end users and cluster admins. These might be common patterns for high-performance AI/ML workloads.

> If someone or something is managing lots of workloads across many namespaces, then they must already have some templating or automation set up that ResourceClaims should neatly fit into? 

Hmm, that makes sense to me, either.

> DeviceClass is used like this ("if you want a GPU X, use DeviceClass A").

If I understood correctly, composite resources(I suppose `gpu/nic` pair) cannot be defined with `DeviceClass`. Am I correct?

> but not really needed and (IMHO) not justifying the effort and additional Kubernetes API surface.

If we won't support `ClusterResourceClaimTemplate` mechanism in the upstream, we might be better having an opportunity, which we document or share(blog?) best practices for how to reduce the toils for defining _common_ and _complex_ `ResourceClaim(Template)`s in the cluster in the near future, probably after all the related KEPs will be out (list type attribute, derived attribute, numaNode attribute, etc.)?

_updated: The nojnhuh's proposed alternative (https://github.com/kubernetes/enhancements/pull/6159#discussion_r3411235764) looks good_

keps/sig-node/5978-cluster-resource-claim-template/README.md
everpeace: > Presumably, a controller could take an instance of some ClusterResourceClaim CRD and generate a ResourceClaimTemplate with a specific name in whichever namespaces.

Nice alternative! Although this behavior is very simple, do you think it is worth providing/managing in the kubernetes community (I imagine this could be provided via `kubernetes-sigs/clusterresourceclaimte-controller` or something)??

keps/sig-node/5978-cluster-resource-claim-template/kep.yaml
troychiu: If I understand correctly, there are no changes to the scheduler. Am I missing anything?

keps/sig-node/5978-cluster-resource-claim-template/kep.yaml
troychiu: Sure. Added!

keps/sig-node/5978-cluster-resource-claim-template/README.md
pohly: > If I understood correctly, composite resources(I suppose gpu/nic pair) cannot be defined with DeviceClass. Am I correct?

Correct. I mentioned DeviceClass as an example where a separate, cluster-scoped type makes sense because it is more than syntactic sugar, not because it is a viable alternative for this KEP.

> The nojnhuh's proposed alternative (https://github.com/kubernetes/enhancements/pull/6159#discussion_r3411235764) looks good

That discussion hasn't settled down yet, but in the meantime do we agree that we don't proceed with this KEP for 1.37 (and potentially never)?

keps/sig-node/5978-cluster-resource-claim-template/README.md
johnbelamaric: I think that's reasonable. @troychiu WDYT? Instead of doing this in-tree, we create a separate controller in k-sigs? We would therefore withdraw this KEP. Perhaps mark it as `withdrawn` and point to the new repo (which we will have to create), and still merge it (that way the history is more discoverable).

As for implementation, I wouldn't bother with a CRD. What I would do is publish RCTs in the namespace of the controller with a special naming convention, and have it copy them to different namespaces. Maybe there would be a mapping of naming convention to sets of namespaces, too.

johnbelamaric: In fact there is another common pattern mentioned here - a CRD that generates appropriate RCs/RCTs. We could perhaps consider a set of libraries and/or controllers in a k-sigs repo to help support these kind of solutions.

troychiu: I think that makes sense. Maintaining an out-of-tree controller looks good enough. I have marked this KEP as `withdrawn`.

johnbelamaric: Can you add:

1) A description of the out-of-tree alternative (just a paragraph or two)
2) Something at the top that says something like "NOTE: This KEP was withdrawn in favor of `<link to alternative>`."

troychiu: Updated!

johnbelamaric (commented):
/approve
/lgtm

Thank you @troychiu 

@mrunalp @haircommander please note we are withdrawing this KEP; the merge is just to document this.

PR #6161, opened 2026-06-05
KEP-5683: Node Lifecycle Conditions

rthallisey:
- One-line PR description: Add several conditions to the Node, that will provide important node lifecycle information to core controllers

- Issue link: https://github.com/kubernetes/enhancements/issues/5683

- Other comments:
  - This design is meant to provide the building blocks for additional lifecycle improvements
  - Locking/ownership and a coordination API will be solved later as part of https://github.com/kubernetes/enhancements/pull/5769 
  - Use cases [document](https://docs.google.com/document/d/1EINvuVzEoRra0CKH6uQnOcQJVbd7ZnxT1bcyySN-r7c/edit?usp=sharing)

### DIFF
--- a/keps/sig-node/5683-lifecycle-conditions/README.md
+++ b/keps/sig-node/5683-lifecycle-conditions/README.md
@@ -1,0 +1,705 @@
+# KEP-NNNN: Node Lifecycle Conditions
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: DaemonSet Controller and GNS Kubelet Disagree](#story-1-daemonset-controller-and-gns-kubelet-disagree)
+    - [Story 2: Jobs Stuck When Nodes Become Unreachable](#story-2-jobs-stuck-when-nodes-become-unreachable)
+    - [Story 3: Broken Nodes Can Consume Rollout Budget](#story-3-broken-nodes-can-consume-rollout-budget)
+    - [Story 4: DaemonSet Rollouts Aren't Reporting the Node's State](#story-4-daemonset-rollouts-arent-reporting-the-nodes-state)
+    - [Story 5: Taints are Insufficient for Signaling Node Drain](#story-5-taints-are-insufficient-for-signaling-node-drain)
+    - [Story 6: Reactive vs Proactive Drain](#story-6-reactive-vs-proactive-drain)
+    - [Story 7: Coordinating between Controllers and Admins](#story-7-coordinating-between-controllers-and-admins)
+    - [Story 8: Coordinating between the Autoscaler and Scheduler during Scale-in](#story-8-coordinating-between-the-autoscaler-and-scheduler-during-scale-in)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [GracefulNodeShutdownInProgress Condition](#gracefulnodeshutdowninprogress-condition)
+  - [Lifecycle Conditions](#lifecycle-conditions)
+  - [Writer Ownership](#writer-ownership)
+  - [Feature Gate](#feature-gate)
+  - [Future Extension Points](#future-extension-points)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha - Introduce Well-Known Conditions](#alpha---introduce-well-known-conditions)
+    - [Alpha2 - Consume Conditions in Controllers](#alpha2---consume-conditions-in-controllers)
+    - [Beta](#beta)
+    - [GA](#ga)
+    - [Deprecation](#deprecation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [x] (R) Enhancement issue in release milestone, which links to KEP dir in
+  [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and
+  SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for
+    [Conformance Tests]
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints] must be hit by [Conformance Tests] within one
+    minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for
+  publication to [kubernetes.io]
+- [ ] Supporting documentation, for example additional design documents, links
+  to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[all GA Endpoints]: https://github.com/kubernetes/community/pull/1806
+[Conformance Tests]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Kubernetes has several independent components that need to understand a Node's
+lifecycle state to contextualize workload availability. Today, that state is
+inferred from a mix of Node readiness, taints, Pod state, controller status,
+labels, annotations, and provider-specific APIs. These signals are useful for
+their individual purposes, but they do not provide a shared Kubernetes-owned
+place for an admin to publish lifecycle state on the Node.
+
+This KEP introduces well-known "lifecycle conditions" to Nodes. These
+conditions provide an observable lifecycle signal that core controllers and
+ecosystem tooling can consume without each building its own interpretation of
+Node, Pod, and controller state. The pattern follows
+[KEP-5394](../5394-psi-node-conditions/README.md), which
+introduces well-known Node conditions as an observable signal and then updates
+core behavior to react to those condition values.
+
+The first version is intentionally narrow. It does not define a general node
+maintenance protocol. It establishes the pattern: a well-known condition
+consumable on the Node, providing a foundation for future work.
+
+## Motivation
+
+Node lifecycle state is a cross-cutting concern. Kubelet, node lifecycle
+controller, workload controllers, the scheduler, autoscalers, operators, admins,
+and external maintenance systems can all observe pieces of the Node's lifecycle,
+but there is no well-known place where an admin can publish "this Node is in x
+lifecycle state", which would explain certain Pod behaviors.
+
+The absence of a shared lifecycle signal leads to duplicated and fragile
+interpretation. One controller may infer lifecycle state from Node readiness.
+Others may look at taints. Some controllers watch Pods, map them back to Nodes,
+inspect Node conditions, and apply its own ignore policy. There are operators
+that publish annotations or labels that only its own components understand.
+These approaches work locally, but they do not create reusable lifecycle data
+that can be used cooperatively with core Kubernetes controllers and they do not
+provide a stable endpoint for admins or users.
+
+Kubernetes already uses well-known Node conditions for shared Node state such as
+`Ready`, `MemoryPressure`, `DiskPressure`, `PIDPressure`, and
+`NetworkUnavailable`. This KEP extends that model by proposing new Node
+conditions for Node Lifecycle. First, publish well-known Node lifecycle signals,
+then make core controllers use those signals for better status reporting.
+
+### Goals
+
+- Add a well-known `GracefulNodeShutdownInProgress` Node condition type.
+- Add well-known lifecycle Node condition types for drain and maintenance:
+  `DrainInProgress`, `Drained`, `MaintenancePlanned`, and `MaintenanceInProgress`.
+- Define condition semantics that let Kubernetes and admins communicate
+  lifecycle state relevant to workload availability.
+- Establish a Kubernetes-owned lifecycle signal that future core controllers and
+  ecosystem tools can consume.
+- Provide an asynchronous building block that can be extended without changing
+  the initial condition.
+
+### Non-Goals
+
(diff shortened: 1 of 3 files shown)


johnbelamaric:
/stage alpha
/wg node-lifecycle

@haircommander @mrunalp this is another one we'd like to consider for 1.37 (not DRA related, this is from node lifecycle WG)

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,549 @@
...
+progressing through the shutdown sequence. `LifecycleMaintenance` reports that
+the Node is in a known maintenance state that may explain workload
+unavailability. `LifecycleDrain` reports that node is undergoing disruption.
+This makes the lifecycle state observable on the Node and gives core controllers
+a common signal to consume.
+
+The new Node conditions will have different management models. Kubelet sets
+`GracefulNodeShutdownInProgress=True` while Graceful Node Shutdown is active and
+sets it to `False` when shutdown is no longer in progress. An admin, or an
+admin-authorized maintenance controller, sets the Lifecycle Condition bool to
```
mortent: This mentions "the Lifecycle Condition", but it is actually two separate condition types here right?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,549 @@
...
+- `status=True`: an admin has published a lifecycle state that may explain
+  workload unavailability on this Node.
+- `status=False`: no admin-published lifecycle state is currently active.
+- `status=Unknown`: the admin-published lifecycle state cannot be determined.
+
+When `status=True`, the condition `reason` contains the lifecycle value. The
+initial lifecycle values for `LifecycleDrain` are:
+
+- `Draining`: the Node is actively being drained.
+- `Drained`: the Node has reached the drain criteria selected by the actor
```
mortent: What is the transition away from this state? Is the assumption that the node will always be terminated after the drain is complete? If not, at what point and at which trigger should the `status` become `False`? Maybe this isn't critical as long as we rely on admins to set the conditions, but I think it is useful to consider what it means that a Node is no longer in the "has been drained" state.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,549 @@
...
+sets it to `False` when shutdown is no longer in progress. An admin, or an
+admin-authorized maintenance controller, sets the Lifecycle Condition bool to
+True and provides the lifecycle value in the condition `reason`. The initial
+Lifecycle Condition reasons are `Draining`, `Drained`, `MaintenancePlanned`, and
+`MaintenanceInProgress`. Clearing the lifecycle state is done by setting the
+bool to False or removing the condition.
+
+### Use Case Coverage
+
+The three conditions are intended to cover the use cases described in
```
mortent: The new conditions seems useful to address these use-cases, but the descriptions here feels a bit vague in terms of how these will be used by controllers and if the conditions will be sufficient to fix the issues described in the doc. Do we have a plan for exactly how controllers will consume them?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,549 @@
...
+
+### GracefulNodeShutdownInProgress Condition
+
+Add a new `NodeConditionType` constant:
+
+```go
+const (
+        // GracefulNodeShutdownInProgress reports whether kubelet is actively
+        // progressing through Graceful Node Shutdown on this Node.
+        GracefulNodeShutdownInProgress NodeConditionType = "GracefulNodeShutdownInProgress"
```
mortent, line 338: How will this interact with the Lost Graceful Shutdown State gap that has been identified? Will the Kubelet use this condition to discover whether a Graceful Node Shutdown is in progress? And if not, is there a chance that the node condition and the Kubelet gets out of sync?

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: For this KEP, the Admin owns the transition, so the KEP intentionally does not define where the Node goes next after `Drained`.  The conditions should become `False` or be removed when that actor considers the lifecycle state no longer active.

I agree it would be useful to model the next state explicitly.  I’m treating that as follow-up work for the broader lifecycle management proposal in (https://github.com/kubernetes/enhancements/pull/5769)

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Correct.  The "Lifecycle Conditions" are `lifecycleDrain` and `lifecycleMaintenance`.  I'll make the writing more clear.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: My goal in this KEP is to describe the shared signal and keep the scope focused on defining the Node  conditions. The individual controller changes are intentionally out of scope here, but these conditions are the prerequisite signal those fixes need.

Each use case is tracked by its own issue/design item, where the consuming controller behavior can be reviewed independently and assigned to the right owners. I can update this section and attach those issues.  Does that work for you @mortent ?

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: I’m going to change the wording here because the current text implies changing kubelet behavior that this KEP should not design.  For this KEP, `GracefulNodeShutdownInProgress` is only defining the condition.  How this condition is used to solve the Lost Graceful Shutdown State gap will be a follow up design in the bug.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Made some updates @mortent, after our discussion from the WG.  Can you see if that satisfies your concern?

mortent: Yeah, I think this gives a clearer vision for how the additional condition types will be used to solve actual problems.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,670 @@
...
+Kubernetes already uses well-known Node conditions for shared Node state such as
+`Ready`, `MemoryPressure`, `DiskPressure`, `PIDPressure`, and
+`NetworkUnavailable`. This KEP extends that model by proposing new Node
+conditions for Node Lifecycle. First, publish well-known Node lifecycle signals,
+then make core controllers use those signals for better status reporting.
+
+### Goals
+
+- Add a well-known `GracefulNodeShutdownInProgress` Node condition type.
+- Add a well-known `LifecycleDrain` Node condition type.
```
johnbelamaric: Please see https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties and revise these specific values to meet that guidance.

In particular, Reason generally should not define a state machine. Controllers may respond to particular conditions, but we don't want to program in the state machine.

rthallisey: Reworked the proposal to follow convention.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in
+  [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for
```
soltysh, line 72: Please stick with the [template](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/README.md), we have tooling build around it, which helps us ensuring the document has all the required parts.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+inferred from a mix of Node readiness, taints, Pod state, controller status,
+labels, annotations, and provider-specific APIs. These signals are useful for
+their individual purposes, but they do not provide a shared Kubernetes-owned
+place for an admin to publish lifecycle state on the Node.
+
+This KEP introduces well-known "lifecycle conditions" to Nodes. These
+conditions provide an observable lifecycle signal that core controllers and
+ecosystem tooling can consume without each building its own interpretation of
+Node, Pod, and controller state. The pattern follows
+[KEP-5394](https://github.com/kubernetes/enhancements/pull/5395), which
```
soltysh:
```suggestion
[KEP-5394](../5394-psi-node-conditions/README.md), which
```

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+
+- `kube-controller-manager`
+- `kube-apiserver`
+
+When the feature gate is disabled:
+
+- Kubernetes controllers do not recognize `GracefulNodeShutdownInProgress`,
+  `DrainInProgress`, `Drained`, `MaintenancePlanned`, or
+  `MaintenanceInProgress` as well-known conditions for core controller
+  consumption.
```
soltysh: Based on your last non-goal "Design how every workload controller consumes these conditions." this statement is unexpected. Unless I'm missing something? You're only defining the conditions, but there is no mention how controllers are consuming them, so whatever the value of FG is on or off, the controllers' behavior should not change.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+
+The use cases and the bugs associated with them should be addressed separately.
+Those designs would leverage this KEP to fill the gap around node state that
+control plane components don't currently have.
+
+This KEP is a precursor to
+[Specialized Lifecycle Management](https://github.com/kubernetes/enhancements/issues/5683).
+It creates the well-known Node conditions that can be used to influence core
+Kubernetes controller functionality. SLM can later support condition ownership,
+coordination, and define a synchronous state model behind an API.
```
soltysh: This statement is recursive 😅 Inside the document you're talking about the same document 😉 Maybe you've meant that the next stages of development (next alpha, etc) will focus on building on top of this current work?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+#### Integration tests
+
+- Verify that the Node has `GracefulNodeShutdownInProgress`,
+  `DrainInProgress`, `Drained`, `MaintenancePlanned`, and
+  `MaintenanceInProgress` conditions.
+
+#### e2e tests
+
+For alpha, e2e coverage should verify that the condition is exposed on the Node.
+The core controller tests will be tracked in their own issues.
```
soltysh: Hmmm... I'm having a hard time figuring out how you can test it. Theoretically you're covering the idea of introducing new conditions, but don't set them anywhere, just yet. What is there to test in that case?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+
+#### Beta
+
+- Define DaemonSet behavior when Node is Graceful Node Shutdown state.
+- Define DaemonSet and Job behavior when Node is undergoing maintenance
+- Define kubelet recovery behavior for lost Graceful Node Shutdown state.
+- Address condition ownership, locking, and coordination behind a new API.
+- Gather feedback from rollout tooling and large-cluster operators.
+- Decide whether additional lifecycle condition types should be standardized.
+- Feature gate defaults to enabled.
```
soltysh, line 505: Since you're proposing expanding controllers with consuming these newely added conditions, I'd expect this to happen during second alpha. This should give you enough time to flesh out all the edge cases. The reason this should not happen during beta, is that at beta we enable a FG on by default, which makes it broadly available. At this point in time, I'd expect to at most resolve bugs, not define the actual behavior. 

So my proposal is:
- alpha1: introduce conditions
- alpha2: consume conditions in controllers
- beta: on by default + eventual bug fixes, and minor improvements
- ga: only stabilitzation.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+
+This KEP is additive. Anything using these condition types can retain their
+current meaning.
+
+On upgrade, clusters that enable the feature gate may see new
+`GracefulNodeShutdownInProgress`, `DrainInProgress`, `Drained`,
+`MaintenancePlanned`, and `MaintenanceInProgress` conditions on Nodes.
+
+On downgrade or feature disablement, core controllers do not use the conditions.
+Existing admin-written conditions may remain on Nodes.
```
soltysh: Core controllers don't consume it initially, don't they?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,688 @@
...
+### Version Skew Strategy
+
+If a future controller-manager version supports consuming these conditions but
+the current controller-manager does not, the conditions are not consumed.
+
+If `kube-controller-manager` supports the feature but the feature gate is not
+enabled, behavior remains unchanged.
+
+If multiple controller-manager versions run during an upgrade, controllers may
+temporarily differ in whether they consume these conditions.
```
soltysh: Here again the question regarding if the controllers are consuming these conditions.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: :upside_down_face: wrong link.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Agreed.  There's no new behavior introduced so there probably isn't an e2e test I can add.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Correct, updated.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: I removed this.  This will return in alpha2, as recommended.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Removed this.  Agreed, the behavior is not changing in this KEP.

keps/sig-node/5683-lifecycle-conditions/README.md
rthallisey: Thanks.  Moved over to the new template.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,705 @@
...
+
+As a user or controller observing drain progress,
+`DrainInProgress=True` or `Drained=True` gives users a clear
+Node status signal that is separate from scheduling policy and can attest to
+lifecycle progress.
+
+There are many solutions that can be implemented for this one, so will focus on
+`kubectl drain` for simplicity. When running `kubectl drain`, the Node will receive
+condition `DrainInProgress=True`. When `kubectl drain` completes,
+`Drained=True` can be set. It will be up to the admin to clear the
```
soltysh, line 261: 👍

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,705 @@
...
+- `MaintenanceInProgress` condition type is added.
+- Initial reason cause categories are documented.
+- `NodeLifecycleConditions` feature gate is added.
+- Unit tests cover condition reading and feature-gate behavior.
+
+#### Alpha2 - Consume Conditions in Controllers
+
+- Define DaemonSet behavior when Node is Graceful Node Shutdown state.
+- Define DaemonSet and Job behavior when Node is undergoing maintenance
+- Define kubelet recovery behavior for lost Graceful Node Shutdown state.
```
soltysh, line 499: 👍 I assume you'll be updating the test sections above accordingly at that stage.

soltysh (approved):
/approve
the PRR for alpha

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -0,0 +1,705 @@
...
+For this KEP, the new Node conditions are admin managed. An admin, or an
+admin-authorized maintenance controller, sets the condition status to `True`
+when appropriate. Clearing the lifecycle state is done by setting the condition
+status to `False` or removing the condition.
+
+### User Stories
+
+The conditions in this proposal are intended to provide a shared signal for the use cases
+described in
+[[public] Node Lifecycle Use Cases](https://docs.google.com/document/d/1EINvuVzEoRra0CKH6uQnOcQJVbd7ZnxT1bcyySN-r7c/edit?usp=sharing)
```
mrunalp, line 175: Can we inline the relevant use cases directly into the KEP?

mrunalp: I see that you have use cases further down.

mrunalp (approved):

johnbelamaric (commented):
/approve
/lgtm

petr-muller:

Issue #6162, opened 2026-06-05
KEP-4322: Remove "Push Model via Credentials in Secret" from Cluster Access


kahirokunn:
### Enhancement Description
- One-line enhancement description (can be used as a release note): Remove the not-recommended direct credentials-in-Secret push model from the KEP-4322 text because Secret-based credential retrieval can be represented through Access Provider Plugins.
- Kubernetes Enhancement Proposal: [KEP-4322: ClusterProfile API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/4322-cluster-inventory)
- Discussion Link:
  - [PR #6136 comment proposing removal](https://github.com/kubernetes/enhancements/pull/6136#issuecomment-4619571213)
  - [PR #6136 maintainer agreement to handle this in a new PR](https://github.com/kubernetes/enhancements/pull/6136#issuecomment-4622025371)
  - [PR #6136 follow-up SGTM](https://github.com/kubernetes/enhancements/pull/6136#issuecomment-4622535786)
  - [cluster-inventory-api discussion about the credentials-in-Secret push model](https://github.com/kubernetes-sigs/cluster-inventory-api/pull/18)
- PRs by stage and milestone:
  - [ ] Alpha - v1.34 KEP cleanup, follow-up to existing KEP-4322 tracking
    - [ ] KEP (`k/enhancements`) update PR(s): TBD, follow-up to [PR #6136](https://github.com/kubernetes/enhancements/pull/6136)
    - [ ] Code (`k/k`) update PR(s): N/A
    - [ ] Docs (`k/website`) update PR(s): N/A
### Existing Tracking
- Existing KEP-4322 tracking issue: [#4322](https://github.com/kubernetes/enhancements/issues/4322)
- KEP-5339 tracking issue: [#5339](https://github.com/kubernetes/enhancements/issues/5339)
- KEP-5339 into KEP-4322 merge tracking issue: [#6135](https://github.com/kubernetes/enhancements/issues/6135)
- KEP-5339 into KEP-4322 merge PR: [#6136](https://github.com/kubernetes/enhancements/pull/6136)
- Secret reader plugin docs: [plugins/secretreader/cmd/plugin/README.md](https://github.com/kubernetes-sigs/cluster-inventory-api/blob/main/plugins/secretreader/cmd/plugin/README.md)
[...]

PR #6136, opened 2026-06-01
 KEP-4322: Merge KEP-5339 into KEP-4322

mikeshng:
- One-line PR description: [KEP-5339](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials) (Plugin for Credentials in ClusterProfile) content has been moved into [KEP-4322](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/4322-cluster-inventory) (ClusterProfile API) since the base API and the credentials-related fields are expected to graduate together. [KEP-5339](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/5339-clusterprofile-plugin-credentials) is marked as replaced.

- Issue link: https://github.com/kubernetes/enhancements/issues/6135

- Other comments: https://github.com/kubernetes-sigs/cluster-inventory-api/issues/58

PR #6163, opened 2026-06-05
KEP-4322: Move "Push Model via Credentials in Secret" to Alternatives

kahirokunn:
- One-line PR description: Move the `Push Model via Credentials in Secret (Not Recommended)` section, including the `Secret format` subsection, from Cluster Access to Alternatives with the rationale, and update the Cluster Access approach count.

- Issue link: https://github.com/kubernetes/enhancements/issues/6162

### DIFF
--- a/keps/sig-multicluster/4322-cluster-inventory/README.md
+++ b/keps/sig-multicluster/4322-cluster-inventory/README.md
@@ -110,8 +110,6 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Cluster Access](#cluster-access)
     - [Pull Model with Work API](#pull-model-with-work-api)
     - [Push Model with Identity Federation (Recommended)](#push-model-with-identity-federation-recommended)
-    - [Push Model via Credentials in Secret (Not Recommended)](#push-model-via-credentials-in-secret-not-recommended)
-      - [Secret format](#secret-format)
     - [Push Model via Access Provider Plugins](#push-model-via-access-provider-plugins)
   - [Access Provider Plugin Design](#access-provider-plugin-design)
     - [External Access Provider Plugin Mechanism](#external-access-provider-plugin-mechanism)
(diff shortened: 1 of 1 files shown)


kahirokunn:
@mikeshng I would appreciate it if you could review it when you have a moment. 🙏

mikeshng (approved):
The rationale for removing this is clearly explained in https://github.com/kubernetes/enhancements/issues/6162, and we seem to have already reached community consensus on the change.

/lgtm

as a cluster-inventory-api maintainer.

CC @skitt @JeremyOT @lauralorenz

mikeshng (commented):
I think @lauralorenz 's comment in https://github.com/kubernetes/enhancements/pull/6237#issuecomment-4971698024 should to here instead.

> Can we move the push model to Alternatives Considered instead of removing it entirely so that we have somewhere to point to in the future? ex when existing current tools that do that style want to adopt ClusterProfile. And give any more rationale about why it is not recommended if there was anything new discussed since it was first added.

kahirokunn:
Moved the section, including the Secret format details, to Alternatives instead of deleting it.

On why it's not recommended, the main development since it was first added is the access provider plugin mechanism from KEP-5339, merged in #6136. The secretreader and kubeconfig-secretreader plugins in cluster-inventory-api cover the same use case through the standard exec protocol, and there is more discussion in kubernetes-sigs/cluster-inventory-api#18.

kahirokunn:
I would appreciate it if you could review this again. Thank you 🙏

mikeshng (approved):
/lgtm

Thank you for addressing this feedback from the SIG-MC community.

kahirokunn:
Thank you too 🙏

mikeshng:
CC @skitt Thank you!

skitt:
/approve

Thanks!

kahirokunn:
Thank you too 🙌

Issue #6164, opened 2026-06-05
KEP: Eliminating Internal API Types


jpbetz:
### Enhancement Description

- One-line enhancement description (can be used as a release note): By eliminating needless conversions to the internal type we can reduce kube-apiserver peak memory by up to 46% for our highest-traffic APIs
- Kubernetes Enhancement Proposal: [6165](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/6164-internal-type-elimination/README.md)
- Discussion Link: [discussed in the May 27, 2026 SIG API Machinery meeting](https://docs.google.com/document/d/1x9RNaaysyO0gXHIr1y50QFbiL1x8OWnk2v3XnrdkT5Y/edit?tab=t.0#bookmark=id.gfkogozta1d)
- PRs by stage and milestone:
- [x] Alpha - N/A This is a internal only change to how we manage go source code and so it is not possible to feature gate it. So, like we did for [master/keps/sig-api-machinery/5589-gogo-dependency](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5589-gogo-dependency?rgh-link-date=2026-06-09T13%3A01%3A46.000Z) we are making this as stable on the first release because the nature of the change is that it is going straight to production once merged with no way to turn it off. 
- [x] Beta - N/ A Same rationale as Alpha
- [ ] Stable - v1.37
  - [ ] KEP (`k/enhancements`) update PR(s):
    - https://github.com/kubernetes/enhancements/pull/6165
    - https://github.com/kubernetes/enhancements/pull/6202
  - [ ] Code (`k/k`) update PR(s):
    - [x] Pod / Pod List migration
      -  https://github.com/kubernetes/kubernetes/pull/139851
      - https://github.com/kubernetes/kubernetes/pull/139852
      - https://github.com/kubernetes/kubernetes/pull/139853
      - https://github.com/kubernetes/kubernetes/pull/139854
      - https://github.com/kubernetes/kubernetes/pull/139855
      - https://github.com/kubernetes/kubernetes/pull/139856
      - https://github.com/kubernetes/kubernetes/pull/139857
      - https://github.com/kubernetes/kubernetes/pull/139858
    - [ ] Other migrations
      - core: https://github.com/kubernetes/kubernetes/pull/140368
      - endpoint slice: https://github.com/kubernetes/kubernetes/pull/140331
    - [ ] conversion-gen optimizations and guardrails
[...]

PR #6165, opened 2026-06-05
KEP-6164: Add internal version elimination KEP

jpbetz:
KEP issue: https://github.com/kubernetes/enhancements/issues/6164

This was [discussed in the May 27, 2026 SIG API Machinery meeting](https://docs.google.com/document/d/1x9RNaaysyO0gXHIr1y50QFbiL1x8OWnk2v3XnrdkT5Y/edit?tab=t.0#bookmark=id.gfkogozta1d) where there was consensus to pursue this approach.

### DIFF
--- a/keps/sig-api-machinery/6164-internal-type-elimination/README.md
+++ b/keps/sig-api-machinery/6164-internal-type-elimination/README.md
@@ -10,6 +10,7 @@
 - [Proposal](#proposal)
   - [Migration steps](#migration-steps)
   - [Migration order](#migration-order)
+    - [Commitment](#commitment)
   - [User Stories](#user-stories)
   - [Notes/Constraints/Caveats](#notesconstraintscaveats)
   - [Risks and Mitigations](#risks-and-mitigations)
@@ -153,13 +154,17 @@ measured an internal↔v1 round-trip conversion of a representative `ClusterRole
 - Preserve the idea of a hub type.
 - Eliminate the internal-to-versioned conversion cost for migrated resources.
 - Define a simple, incremental, per-resource migration.
-- Focus on the migration of high-traffic APIs to harvest the performance and scale benefits.
+- Sequence by traffic to harvest the performance and scale benefits early.
+- Converge on a *single* conversion pattern (a versioned hub) across all
+  built-in APIs, rather than carrying two patterns indefinitely.
 
 
 ### Non-Goals
 
-- It is not a goal to migrate all APIs. Low-traffic groups may keep their
-  internal type indefinitely. This does not make such types any worse than they are today.
+- It is not a goal to migrate all APIs in a single release. The migration is
+  sequenced over several releases (see [Migration order](#migration-order)), but
+  the end state is a single conversion pattern across *all* built-in APIs, not a
+  permanent split between migrated and unmigrated groups.
 - Changing the conversion-gen, deepcopy-gen, or defaulter-gen tooling contracts
   (they already handle aliases correctly; see Design Details).
 
@@ -172,26 +177,53 @@ measured an internal↔v1 round-trip conversion of a representative `ClusterRole
    deepcopy/conversion.
 1. **Switch to a Versioned hub:** Drop the `__internal` registration for the
    resource and set its in-memory/hub version to the storage version.
-1. **Cleanup (can long after the main migration).** Migrate all call sites from the internal
-   types to the versioned types and delete the internal package.
+1. **Update call sites:** Migrate call sites (admission, registry, controllers)
+   from the internal type to the versioned type. This follows each resource's
+   hub switch promptly; it is not deferred indefinitely.
+1. **Delete the internal package (can be later):** Once no call sites remain,
+   delete the internal type and its generated code. This final removal can land
+   well after the hub switch.
 
 ### Migration order
 
-We will **prioritize the highest-traffic APIs**:
+We sequence by traffic to harvest performance benefits early, while converging
+on a single pattern for *all* built-in APIs:
 
-1. **Demonstrate mechanism** Prove the mechanism on a small,
-   single-version, structurally-identical group. `rbac.authorization.k8s.io` is
-   the reference PoC; `coordination` (`Lease`) and `storage` are similar.
+Group 1:
+
+1. **Demonstrate mechanism** Prove the mechanism on a small set of APIs, e.g.
+   group `pod` (difficult in terms of scale), CRD (difficult because the internal API complex)
+   and an easy group like `rbac.authorization.k8s.io` to demonstrate the
+   migration mechanism.
+
+Group 2:
+
 1. **High-traffic easy APIs** These carry most
    apiserver traffic and alias cleanly today: `coordination/Lease` (kubelet
    heartbeats), `discovery/EndpointSlice`, `events.k8s.io/Event`,
    `core/ConfigMap`, `core/Endpoints`, `core/Node`, `core/ServiceAccount`.
+1. **Structurally-identical APIs.** All other groups whose internal
+   and versioned types are already identical are aliased and switched to a
+   versioned hub. These are low-risk and are part of the committed migration,
+   not just the high-traffic subset.
+
+Group 3:
+
 1. **High-traffic complex APIs** `core/Pod`,
    `core/Secret` have hand-written conversions and/or internal-only
-   fields and cannot be aliased as-is. In such cases, internal type will first need
-   to modified to match the hub type.
-   and are sequenced after the clean cases.
-1. **The long tail.** Remaining groups migrate opportunistically and so may never migrate.
+   fields and cannot be aliased as-is. In such cases, the internal type will
+   first need to be modified to match the hub type, and so these are sequenced
+   after the clean cases.
+1. **The long tail.** Groups that are not yet structurally identical have their
+   internal type reshaped to match the hub and are then migrated.
+
+#### Commitment
+
+Our criteria success is the successful migration of the `pod` (most impactful)
+and `CRD` (most difficult) APIs. If we fail to migate these within 2 releases we
+will add back the internal APIs for all migrated APIs. If these are successful, we
+ commit to completing the remaining migrations over a 3-release
+(~1 year) window**.
 
 ### User Stories
 
@@ -213,8 +245,14 @@ kube-apiserver runs with reduced peak load and GC churn.
 ### Risks and Mitigations
 
 
-- **Regressions:** *The existing round-trip
-  fuzz harness will be a huge help in preventing this.
+- **Regressions:** The existing round-trip fuzz and serialization-compatibility
+  harnesses (see [Test Plan](#test-plan)) are the main defense. A migration that
+  silently changed serialized output would fail the compatibility fixtures, and a
+  conversion that dropped data would fail the round-trip fuzz tests.
+- **Defaulting drift:** A few groups (e.g. `autoscaling/v1`) defer some defaulting
+  into the internal-conversion step. Collapsing that conversion could skip such
+  defaulting; these groups are identified and handled in the long-tail reshaping
+  step rather than aliased blindly.
 - **Reviewability:** Touching many groups risks large,
   hard-to-review PRs. We will limit PRs to a single migration each.
 
@@ -278,37 +316,66 @@ functions off of the internal type.
 
 ### Test Plan
 
-- **Round-trip fuzz:** This migration depends heavily on the existing roundtrip fuzz tests
-  to ensure the changes are safe.
-- **Conversion benchmarks:** A per-resource conversion benchmark will be included
-  in each migration PR.
+The best tests to ensure correctness are tests we already have today:
+
+- **Serialization compatibility fixtures.** Because a migration is
+  invisible on the wire and in storage, we require the migration to incur
+  zero fixture changes. These test are critical to manage upgrade/rollback risk.
+- **Round-trip fuzz tests.** Harness:
+  `staging/src/k8s.io/apimachinery/pkg/api/apitesting/roundtrip`. This offers some
+  assurances that migration does not drop or corrupt data. It complements the
+  compatibility fixture tests which offer the strongest accurance that the
+  internal type removal has not changed normal codepaths.
+- **Conversion benchmarks.** A per-resource conversion benchmark (see
+  [Benchmarks](#benchmarks)) is included in each migration PR to confirm the
+  expected allocation/latency reduction and guard against regressions.
 
 ### Graduation Criteria
 
-Because there is no single runtime feature gate (see PRR), graduation is defined
-by mechanism maturity and migration coverage rather than alpha/beta/GA of a
-toggle.
+This enhancement has no runtime feature gate (see PRR) so each resource
+migration is unconditionally in effect the moment it merges, and there is no
+higher maturity than "unconditionally on". This is unavoidable due to the nature
+of the code change.
 
 #### Alpha
 
-Alpha is not meaningful for this as all changes are production impacting.
+Not possible for this change. SIG Leads agreed that we should communicate the nature of this
+change by marking it as sable on the first release. This matches how we've handled
+other internal changes.
 
 #### Beta
 
-This feature will stay in beta for the migration of all high-traffic APIs.
+Same retionale as Alpha.
 
 #### Stable
 
-- Migration of all high-traffic APIs is complete.
-- All new APIs use a versioned hub type.
+Phase 1:
+
+- The migration mechanism is proven when a few low risk APIs and then `pod` (most impactful) and
+  `CRD` (most difficult) are migrated.
+- Because the change cannot be gated, these migrations are production-impacting
+  on merge.
+
+Phase 2:
+
+- We're committed to the migration, and the remaining structurally-identical and
+  will complete all remaining migration work over a 3-release window.
+
+Phase 3:
+
+- All built-in APIs use a versioned hub type.
+- Internal packages for migrated groups are deleted.
+- All new APIs use a versioned hub type by default.
 
 ### Upgrade / Downgrade Strategy
 
-None. This is an internal-only change.
+None required. The change is purely in-memory and stored and wire encodings are
+unchanged.
 
 ### Version Skew Strategy
 
-None. This is an internal-only change.
+None. The internal/hub type is server-local and never appears on the wire or in
+storage, so mixed-version control planes are unaffected.
 
 ## Production Readiness Review Questionnaire
 
@@ -317,8 +384,11 @@ None. This is an internal-only change.
 ###### How can this feature be enabled / disabled in a live cluster?
 
 - [x] Other
-  - **Describe the mechanism:** This is non-user-facing source code change that cannot be gated.
-  - **Will enabling / disabling require downtime of the control plane?** No.
+  - **Describe the mechanism:** This is a non-user-facing source-code change that
+    cannot be gated and cannot be disabled at runtime. It takes effect when an
+    apiserver built with the migration runs.
+  - **Will enabling / disabling require downtime of the control plane?** N/A. It
+    cannot be enabled or disabled and in effect for any apiserver with the merged code.
 
 ###### Does enabling the feature change any default behavior?
 
@@ -326,25 +396,38 @@ No. This is an internal-only change.
 
 ###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
 
-No. But it is an internal-only code change with rigiorious correctness and performance testing planned.
+No, this change cannot be disabled. There is no feature gate and the migration is
+a source-code change. Correctness and performance are validated by round-trip
+fuzz tests and conversion benchmarks. To reverse a specific migration, the
+migration PR must be reverted (and the revert would then be cherry-picked to release
+branches).
 
 ###### What happens if we reenable the feature if it was previously rolled back?
 
-No change. This is an internal-only change.
+Not applicable as there is no enablement toggle to reconcile. Re-landing a
+reverted migration (with any needed fixes) would be how we re-enable the feature.
 
 ###### Are there any tests for feature enablement/disablement?
 
-No, but the feature is internal-only and not feature-gatable.
+Compaibility fixture tests do ensure cross-version compatibility, providing insurance that the
+API behavior is the same both before and after the change. This is the closest
+we can hope to get to enablement/disablement testing.
 
 ### Rollout, Upgrade and Rollback Planning
 
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
-No. This is an internal-only change.
+A error in the migration could in theory lead to incorrect conversions that could result
+data that is incorrectly served or stored. Mistakes around defaulting are risk (@deads2k
+pointed this out). Since defaults are populated in the read path, a rollback could resolve
+some of the potential errors this migration could cause.
 
 ###### What specific metrics should inform a rollback?
 
-None. This is an internal-only change.
+A regression in `apiserver_request_duration_seconds` or in the apiserver memory
+profile for the migrated resource, or any round-trip / conformance test failure
+attributable to a migration. There is no flag to flip; a rollback means
+reverting the migration PR.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
--- a/keps/sig-api-machinery/6164-internal-type-elimination/README.md
+++ b/keps/sig-api-machinery/6164-internal-type-elimination/README.md
@@ -74,7 +74,7 @@ the project.
 
 Today many APIs have progressed to a stable `v1` version and the alpha and beta types are no
 longer served by the API. For such APIs the internal type, the versioned type and the preferred
-storage type are alls the same type. 
+storage type are identical. 
 
 As a first step, a type alias (`type Pod = corev1.Pod`) can be used to tell
 go that the internal type is the *same* Go type, so conversion collapses to a no-op.
@@ -89,10 +89,11 @@ of the kube-apiserver is heavily influenced by the allocations performed
 by conversion.
 
 https://github.com/kubernetes/kubernetes/issues/139026 showed that if conversion
-operations are streamed we reduce peak memory by up to 46%. By
-eliminating conversion entirely we expect to do even better.
+operations are streamed we reduce peak memory by up to 46% plus eliminate the GC
+churn caused by the allocations performed during conversion. By eliminating
+conversion entirely we expect to do even better.
 
-The internal type exists for three historical reasons (see [kubernetes/kubernetes#138097][kk-138097]):
+The internal type exists for a few historical reasons (see [kubernetes/kubernetes#138097][#138097]):
 
 1. **A single conversion hub.** N versioned types convert to/from one internal
    type (2N conversion functions) instead of pairwise (N^2).
@@ -109,7 +110,8 @@ All three have weakened:
    onto **versioned** types via `validation-gen`, further weakening the
    rationale for an internal type.
 3. No known built-in API actually requires a server-only type; in practice the
-   internal type is maintained to mirror the newest/preferred versioned type.
+   internal type is best maintained as a exact copy or near copy of the
+   newest/preferred versioned type.
 
 Meanwhile the internal type imposes ongoing costs:
 
@@ -302,11 +304,11 @@ This feature will stay in beta for the migration of all high-traffic APIs.
 
 ### Upgrade / Downgrade Strategy
 
-N/A
+None. This is an internal-only change.
 
 ### Version Skew Strategy
 
-N/A
+None. This is an internal-only change.
 
 ## Production Readiness Review Questionnaire
 
@@ -320,7 +322,7 @@ N/A
 
 ###### Does enabling the feature change any default behavior?
 
-No.
+No. This is an internal-only change.
 
 ###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
 
@@ -328,39 +330,40 @@ No. But it is an internal-only code change with rigiorious correctness and perfo
 
 ###### What happens if we reenable the feature if it was previously rolled back?
 
-N/A
+No change. This is an internal-only change.
 
 ###### Are there any tests for feature enablement/disablement?
 
-N/A
+No, but the feature is internal-only and not feature-gatable.
 
 ### Rollout, Upgrade and Rollback Planning
 
 ###### How can a rollout or rollback fail? Can it impact already running workloads?
 
-N/A
+No. This is an internal-only change.
 
 ###### What specific metrics should inform a rollback?
 
-N/A
+None. This is an internal-only change.
 
 ###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
 
-N/A
+No. 
 
 ###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
 
-N/A
+No.
 
 ### Monitoring Requirements
 
 ###### How can an operator determine if the feature is in use by workloads?
 
-N/A
+It will always be in effect Kubernetes 1.37+ for migrated APIs.
 
 ###### How can someone using this feature know that it is working for their instance?
 
-N/A
+There is no user-facing changes. The performance improvements may be observable by performance
+sensitive workloads.
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
 
--- a/keps/sig-api-machinery/6164-internal-type-elimination/README.md
+++ b/keps/sig-api-machinery/6164-internal-type-elimination/README.md
@@ -17,10 +17,10 @@
   - [Phase 1: Make the internal types memory-identical to the stable served version](#phase-1-make-the-internal-types-memory-identical-to-the-stable-served-version)
     - [Step 1:](#step-1)
     - [Step 2:](#step-2)
-  - [Phease 2: Remove internal types](#phease-2-remove-internal-types)
-    - [Step 1](#step-1-1)
-    - [Step 2](#step-2-1)
-    - [Step 3](#step-3)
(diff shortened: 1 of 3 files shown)


jpbetz:
@michaelasp I'm filing this one for 1.37 based on our discussions and have listed you as co-author

jpbetz:
@deads2k @liggitt @thockin I realize this is a late KEP submission for 1.37, but given we had some prior agreement in this direction in the last SIG API Machinery meeting, and given the DRAM industry shortage, I'll like to merge this so we can start making progress in 1.37 under a tracking KEP.

jpbetz:
@yedou37 Would you like to be involved in this approach? This builds on your analysis of streaming conversion.

michaelasp:
This is also pertinent to the graceful degradation workstream we have with parallel efforts like #5647 

One large problem we've seen is that during bulk operations is the spike memory usage. ~40-50% usage reduction would be a huge boost. A lot of these operations are infrequent and only happen during times such as apiserver restarts and can lead to occurrences such as a stable state -> apiserver restart -> relist causing oom crashloops. Reducing the cost of these bulk operations through these approaches + streaming is super important and would be a good first step in ensuring that the apiserver doesn't fall into these tricky hard to recover scenarios.

yedou37:
@jpbetz Sure, happy to be involved! The idea is to eliminate conversion so we no longer allocate external-version objects, which would likely benefit all relevant paths, especially LIST requests, and in fact all encoding formats, since they currently require a full conversion before encoding.

jpbetz:
@yedou37 Great, I'll list you as an author on the KEP since you did the initial research on streaming conversion that identified the impact here.  Once the KEP is ready we'll coordinate how we order the migrations high traffic types.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+## Proposal
+
+### Migration steps
+
+1. **Alias:** Replace the internal struct definitions with aliases to
+   the storage version, reconcile any internal-only methods, and regenerate
+   deepcopy/conversion.
+1. **Switch to a Versioned hub:** Drop the `__internal` registration for the
+   resource and set its in-memory/hub version to the storage version.
+1. **Cleanup (can long after the main migration).** Migrate all call sites from the internal
```
deads2k: I don't want to stay half-pregnant.  How soon could we update all call sites?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+   types to the versioned types and delete the internal package.
```
deads2k: The deletion could be later.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+- Preserve the idea of a hub type.
+- Eliminate the internal-to-versioned conversion cost for migrated resources.
+- Define a simple, incremental, per-resource migration.
+- Focus on the migration of high-traffic APIs to harvest the performance and scale benefits.
+
+
+### Non-Goals
+
+- It is not a goal to migrate all APIs. Low-traffic groups may keep their
+  internal type indefinitely. This does not make such types any worse than they are today.
```
deads2k: Acknowledge that this does mean we might end up with two different patterns.  That seems painful.  Why shouldn't our goal be to re-standardize over N releases?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+   the storage version, reconcile any internal-only methods, and regenerate
+   deepcopy/conversion.
+1. **Switch to a Versioned hub:** Drop the `__internal` registration for the
+   resource and set its in-memory/hub version to the storage version.
+1. **Cleanup (can long after the main migration).** Migrate all call sites from the internal
+   types to the versioned types and delete the internal package.
+
+### Migration order
+
+We will **prioritize the highest-traffic APIs**:
```
deads2k: Can we also do all the types where internal and external types are identical?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+1. **High-traffic easy APIs** These carry most
+   apiserver traffic and alias cleanly today: `coordination/Lease` (kubelet
+   heartbeats), `discovery/EndpointSlice`, `events.k8s.io/Event`,
+   `core/ConfigMap`, `core/Endpoints`, `core/Node`, `core/ServiceAccount`.
+1. **High-traffic complex APIs** `core/Pod`,
+   `core/Secret` have hand-written conversions and/or internal-only
+   fields and cannot be aliased as-is. In such cases, internal type will first need
+   to modified to match the hub type.
+   and are sequenced after the clean cases.
+1. **The long tail.** Remaining groups migrate opportunistically and so may never migrate.
```
deads2k: I get that it's less work, but five years from now, would you feel good about telling newcomers to learn two different ways of conversion?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+- **Round-trip fuzz:** This migration depends heavily on the existing roundtrip fuzz tests
+  to ensure the changes are safe.
+- **Conversion benchmarks:** A per-resource conversion benchmark will be included
+  in each migration PR.
+
+### Graduation Criteria
+
+Because there is no single runtime feature gate (see PRR), graduation is defined
+by mechanism maturity and migration coverage rather than alpha/beta/GA of a
+toggle.
```
deads2k: Since this cannot be disabled, I think this is moving to stable immediately.  There is no higher maturity than, "unconditionally on".

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Other
+  - **Describe the mechanism:** This is non-user-facing source code change that cannot be gated.
+  - **Will enabling / disabling require downtime of the control plane?** No.
```
deads2k: I think a more honest answer here is, "no, it cannot be disabled".

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+No. This is an internal-only change.
+
+###### What specific metrics should inform a rollback?
+
+None. This is an internal-only change.
```
deads2k: In places that talk about disablement, "this change cannot be disabled", provides greater clarity than "internal-only change".  We've had plenty of internal-only changes that are featuregated.  This one just can't be.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,441 @@
...
+
+###### Are there any tests for feature enablement/disablement?
+
+No, but the feature is internal-only and not feature-gatable.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+No. This is an internal-only change.
```
deads2k: Bold.  You don't think it's possible we'll overlook some defaulting case that can only be corrected by rolling back the version?

deads2k (changes requested):

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Touché.  I've updated this to commit to a go/no-go decision within 2 releases. Either we've migrated some easy cases + pod + CRD or we'll revert the change. If we're a "go" we'll complete the migration in 3 release from this point.  I'm willing to own the reverts if we can make it happen.  Note that the revert only requires we swap the typerefs back to full types, so it's conceptually low risk.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Good point. I'll split this out.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Agreed. See above.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: SG. I'll combine this with high traffic easy APIs.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: SG. See above.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: SGTM. I've switched this to a straight-to-stable with the migration progression spelled out.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: That fair. If we mess up bad enough it would be possible to change what is stored and that would persist across rollback. I've expanded on this to explain the sorts of failures that can happen across version boundaries.  I've also made note that rollback could fix some things, including the default case you've mentioned.

jpbetz:
Thanks @deads2k, feedback applied

keps/sig-api-machinery/6164-internal-type-elimination/kep.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+see-also:
+  - "/keps/sig-api-machinery/5073-declarative-validation-with-validation-gen"
+
+# This is an internal code refactor what is immediately production impacting and
+# not feature gate-able.
+stage: alpha
+latest-milestone: "v1.37"
+milestone:
+  stable: "v1.37"
```
whtssub, line 25: This KEP is targeting stage `alpha`, is the milestone struct mistakenly set to `stable` ?

jpbetz: Fixed. This is a internal only change to how we manage go source code and so it is not possible to feature gate it. So, like we did for https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/5589-gogo-dependency we are making this as stable on the first release because the nature of the change is that it is going straight to production once merged with no way to turn it off.

whtssub: makes sense, thanks for updating it.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+### User Stories
+
+N/A.
+
+kube-apiserver runs with reduced peak load and GC churn.
+
+### Notes/Constraints/Caveats
+
+- **Only structurally-identical types can be aliased directly.** The internal
```
wojtek-t, line 203: So what's the plan for types that aren't structurally identical?

Taking Pod as an example, we have `SecurityContext` that has a different structure.
(There is also ServiceAccountName vs DeprecratedServiceAccount).

Basically anything that have a custom conversion function will not be immediate.

So what's the plan for them? We need to solve that to address group 1 :)

Are we going to go over the codebase and adjust the internal structure to match the hub one first?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+   type (2N conversion functions) instead of pairwise (N^2).
+2. **A single validation target.** Hand-written validation is written once,
+   against the internal type.
+3. **A server-only representation** that is never exposed to clients.
+
+All three have weakened:
+
+1. The hub does not have to be a *separate* type. The hub type can be one of the
+   versioned types CustomResourceDefinitions already operate handle CRD versions this
+   way. For CRD conversion, the storage version *is* the hub.
```
liggitt: this is true, but it comes at the cost of the CRD conversion *having* to support conversion to/from any pair of external versions

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+Eliminating needless conversions offers up to 46% reduction on peak memory for list
+operations served by the kube-apiserver and has long term maintenance benefits to
+the project.
+
+Today many APIs have progressed to a stable `v1` version and the alpha and beta types are no
+longer served by the API. For such APIs the internal type, the versioned type and the preferred
+storage type are identical. 
+
+As a first step, a type alias (`type Pod = corev1.Pod`) can be used to tell
+go that the internal type is the *same* Go type, so conversion collapses to a no-op.
```
liggitt: Will we also need to alias list types (e.g. core.PodList to v1.PodList)?

Is the expensive conversion we're trying to make cheaper:
1. list or individual item conversion? Our list types tend to make their `Items []...` fields non-pointers, so even if individual conversion is made cheap / no-op, won't we still have to allocate a new `Items []...` slice for the list unless we also alias list types
2. "safe" or "unsafe" conversion?
https://github.com/kubernetes/kubernetes/blob/9e570c4124691bd0d765e780d0225655f99a905e/staging/src/k8s.io/apimachinery/pkg/runtime/scheme.go#L472-L486
if we're wanting to make conversion from the shared watch cache to a version serializable for a single client, I'm not sure we can use unsafe conversion due to https://github.com/kubernetes/kubernetes/issues/82497. Are the performance gains still applicable if we're doing safe conversion that produces a list copy?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: > Are we going to go over the codebase and adjust the internal structure to match the hub one first?

Yes. I'm doing a sweep today to ensure this aspect viable. Given that API types are required to be isomorphic, I'm not expecting this to find any hard blockers, but for types like pod where we have a lot of code that depends on the internal types, there is a risk that the external types are less optimal for internal (e.g. selectors), so I need to identify those problem cases early and make sure we'lll be able to deal with them.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+   internal type is best maintained as a exact copy or near copy of the
+   newest/preferred versioned type.
+
+Meanwhile the internal type imposes ongoing costs:
+
+- **Per-request CPU and allocations.** Every request converts at least twice
+  (decode -> internal and internal -> encode), storage adds two more
+  (internal -> storage on write, storage -> internal on read), and every watch event
+  converts once *per watcher*. Each conversion allocates a fresh destination
+  object and runs a generated field copy.
```
liggitt, line 137: Ownership of objects resulting from conversion calls is super important to understand if we're going to make unsafe/noop conversion possible in more places. There are spots that may be assuming ownership of the resulting objects that may need at minimum a DeepCopy even if no conversion is happening / needed.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+- **Maintenance burden.** Each group maintains an extra full set of Go types
+  plus generated deepcopy and conversion functions for them, and every new
+  field must round-trip losslessly through the internal type across *all*
+  supported versions. The drift between internal and versioned types is enough
```
liggitt, line 141: the round-tripping requirement doesn't go away, it just now needs to test round-tripping between every pair of external versions instead of from each version to the hub... I think this KEP makes the round-tripping cost higher to maintain / test (for APIs with multiple external versions, which ~most APIs have and will have)

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+### Benchmarks
+
+https://github.com/kubernetes/kubernetes/issues/139026 shows that even without
+eliminating conversion, that streaming the operation can reduce peak memory by
+by up to 46%.
+
+To validate that eliminating conversion performs as expected, we aliased the
+`rbac.authorization.k8s.io` internal types to their `v1` counterparts and
+measured an internal↔v1 round-trip conversion of a representative `ClusterRole`:
```
liggitt: was this using safe or unsafe conversion?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,524 @@
...
+- **Only structurally-identical types can be aliased directly.** The internal
+  type must be field-for-field identical to the versioned type.
+- **Defaulting:** Defaulting already runs on the *versioned* object
+  at decode time, before conversion to the hub. Aliasing and the hub switch do
+  not move the defaulting boundary. However, a few groups
+  (e.g. `autoscaling/v1`) intentionally defer some defaulting into the
+  internal-conversion step.
+- **Admission:** Admission controllers are written against internal types.
+
+### Risks and Mitigations
```
liggitt, line 212: The *primary* benefit of an unchanging "internal" hub type is during development of a new type when we're making schema changes across versions of an external type. That lets code (internal admission, hand-written validation, etc) stay unchanging as we add / promote new external versions.

Eliminating that means all use of the hub type has to change all at once. That will make PRs to promote / evolve an in-development API schema bigger / messier. I suspect the effect will be to put a damper on schema changes to types in development. Maybe that's what we want, but I didn't see much discussion of what we're losing in this proposal.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: For CRDs the problem with webhook conversion is that if you do 2-step conversion through a hub you double the webhook calls.  We COULD have chosen to support that but it's really horrible to allow users to set up a CRD where that happens.

EDIT: And the conversion webhook author can still use a hub type to manage conversion.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: > Will we also need to alias list types (e.g. core.PodList to v1.PodList)?

I think we should, but it won't have much impact since `core.PodList` allocation skipped entirely for streaming list.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
liggitt: yeah, I don't disagree with the decision for CRD where external calls that require transmitting serialized versions of API objects are involved, just pointing out that CRD CR conversion not using an "internal" hub type isn't especially compelling for getting rid of one inside kube-apiserver

keps/sig-api-machinery/6164-internal-type-elimination/README.md
liggitt: > it won't have much impact since `core.PodList` allocation skipped entirely for streaming list.

but... we need to have a v1.PodList in hand *before* we start streaming serialization, right?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
deads2k: I had forgotten about this.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
deads2k: > it just now needs to test round-tripping between every pair of external versions instead of from each version to the hub

One of the external version *is* the hub, so wouldn't this end up a v1beta1 <-> v1, v1alpha1 <-> v1, but we never bother checking v1beta1 <-> v1alpha1, since we'd be converting that as v1beta1 -> v1 -> v1alpha1.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: I'll expand on this in the KEP. 

There are a few different cases.

- `ConvertToVersion` / `Convert` already perform a deep-copy for self-conversion. We're going to need to explicitly opt-out of deep-copy even after switching to a versioned hub.
- `UnsafeConvertToVersion` already shares data, the existing expectation is that owner of the converted object is the new owner, so this is low risk
- There is a unsafe convert in the PATCH handler we'll need to review
- Unsafe encode does a hack where is GVK stamps and then removes the stamp in a defer and will need careful review

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: > The primary benefit of an unchanging "internal" hub type is during development

I'll capture this as a risk/con.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: I've left some notes in the KEP. All except about 15 fields are identical.

- Most the fields in `PodSpec`. I've reviewed them and most are trivial and a few are inconvenient but addressable.
- `RawExtension` is used in versioned types but converted to `runtime.Object` for internal. This is a mechanical change.
- There are a couple annotation conversions to perserve.

The good news is that the problem fields are in the same type that we most want to optimize, so we'll get a good return on investment here.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Unsafe. I'll make note.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Not sure I follow. We've always had a hub. Today it's the internal type, after this change it's a versioned type. I don't follow why we'd be okay saying "A <-> Hub" + "B <-> Hub" == "A <-> B"  today and be unwilling to say the same with a versioned type as the hub.

liggitt: Sorry, you're right... we still only need to test all the `hub<->version` round-tripping.

I was reacting to "every new field must round-trip losslessly through the internal type across *all* supported versions" being called out as a maintenance burden, as if that goes away if we eliminate internal types. That requirement does not go away just because the hub type is external (and it eliminates some options we have of additional non-serialized bookkeeping fields in the internal hub type).

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Ah, you're right.

While [`streamingEncodeList`](https://github.com/serathius/kubernetes/blob/e7c743b2ebfaed1e3132027c0369ac25b14b6f47/staging/src/k8s.io/apimachinery/pkg/runtime/serializer/json/collections.go#L92) doesn't require the entire List type, [`streamEncodeCollection`](https://github.com/kubernetes/kubernetes/pull/129334/changes?show-viewed-files=false#diff-81d32728d6064bea047cc8cd99fe5346d15c7e8df3a12063f5fd8ec96bf9c5deR35) does.

jpbetz: Re: safe vs. unsafe.

I'm going to run a bench to see what the performance is using safe.

I'm not convinced the encoder version stamping is needed in the [streaming list encoding path](https://github.com/kubernetes/kubernetes/pull/129334/changes?show-viewed-files=false#diff-81d32728d6064bea047cc8cd99fe5346d15c7e8df3a12063f5fd8ec96bf9c5deR125), so there is a possibility we might tool the streaming list path to be unsafe, but I'd like to see the data first.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Gotcha. I'll rephrase

jpbetz:
I ran an experiment to see what happens if the internal PodSpec type is modified to be memory-identical to to the v1 PodSpec type. (No aliasing yet)

This compares unsafe conversion of a non-trivial pod (3 containers, 10 env, 10 mounts, 10 ports, ...) from the internal type to the v1 type.

| pods | master | memory-identical | improvement |
|---|---|---|---|
| 1 | 1,435 ns · 4,280 B · **26** allocs | 485 ns · 1,472 B · **5** allocs | 3.0× faster · 2.9× mem |
| 100 | 109,851 ns · 403,879 B · **2,105** allocs | 19,314 ns · 123,072 B · **5** allocs | 5.7× faster · 3.3× mem |
| 1000 | 859,427 ns · 4.00 MB · **21,005** allocs | 148,795 ns · 1.20 MB · **5** allocs | **5.8× faster · 3.3× mem** |

This suggests that even *before* we invest in aliases, there is a big performance boost for making the internal types memory-identical with the preferred version.

jpbetz:
@deads2k @liggitt Heads up that I've split this into two phases where the first one focuses on making the internal type memory identical with the preferred served type. That's a ton of work, it's where nearly all the performance/scale benefits appears to be, and getting that done makes the removal of internal types far lower risk, so I'd like to focus on that for 1.37.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+Make the internal type field-for-field identical, the main complexity is:
+
+- `core.PodSpec -> v1.PodSpec` (this is also the was we get the biggest scale impact by optimizing)
+- `runtime.Object -> RawExtension`
+- A few fields that are round-tripped through annotations
+
+
+#### Step 2:
+
+Introduce guardrails to prevent internal types from becoming needlessly differen
```
deads2k:
```suggestion
Introduce guardrails to prevent internal types from becoming needlessly different
```
nit

deads2k:
Bold refactor.  The cost/benefit looks good. Both phases look plausible. The DeepCopy looks convincing for ownership.

/lgtm
/hold  holding in case there are additional concerns from Jordan or Tim.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+- **Admission:** Admission controllers are written against internal types.
+
+### Risks and Mitigations
+
+- **Maintaining the hub type increases maintenance burden.** Today the hub type
+  is the internal type. With this change, the hub type will be one of the
+  versioned typed and will change type as the feature stabilizes and the storage
+  version changes. This is a risk that the extra work of switching the hub type
+  each time an API is changed will result in increased maintenance burden and/or
+  will discourage API changes. Handw-ritten validation and admission controllers
```
yedou37:
```suggestion
  will discourage API changes. Hand-written validation and admission controllers
```
nit

jpbetz:
@liggitt @thockin Any other major concerns?  Ok if we merge and iterate?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+   type (2N conversion functions) instead of pairwise (N^2).
+2. **A single validation target.** Hand-written validation is written once,
+   against the internal type.
+3. **A server-only representation** that is never exposed to clients.
+
+All three have weakened:
+
+1. The hub does not have to be a *separate* type. The hub type can be one of the
+   versioned types CustomResourceDefinitions already operate handle CRD versions this
+   way. For CRD conversion, the storage version *is* the hub.
```
liggitt:
```suggestion
   way. For CRD conversion, the request version *is* the hub.
```

[reference](https://github.com/kubernetes/kubernetes/blob/d57b170cda65d284030d6773f7b967f224a6158b/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L974-L975):
```go
			// a handler for a specific group-version of a custom resource uses that version as the in-memory representation
			HubGroupVersion: kind.GroupVersion(),
```

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+  not move the defaulting boundary. However, a few groups
+  (e.g. `autoscaling/v1`) intentionally defer some defaulting into the
+  internal-conversion step.
+- **Admission:** Admission controllers are written against internal types.
+
+### Risks and Mitigations
+
+- **Maintaining the hub type increases maintenance burden.** Today the hub type
+  is the internal type. With this change, the hub type will be one of the
+  versioned typed and will change type as the feature stabilizes and the storage
```
liggitt:
```suggestion
  versioned types and will change type as the feature stabilizes and the storage
```

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+
+We plan to modify conversion-gen to track differences between the types and
+require exemptions for differences.
+
+This is important to prevent accidental performance regressions. Today we
+already have many memory-identical types and are planning to make pod and other
+high traffic APIs memory-identical. If any of those types become different than
+the performance will regress.
+
+### Phease 2: Remove internal types
```
liggitt:
```suggestion
### Phase 2: Remove internal types
```

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+
+This is our goal for v.1.38+
```
liggitt:
```suggestion
This is our goal for v1.38+
```

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+
+- `ConvertToVersion` / `Convert`: Already perform a deep-copy for self-conversion. This is safe. We're going to need to explicitly opt-out of deep-copy even after switching to a versioned hub.
+- `UnsafeConvertToVersion`: Already shares data, the existing expectation is that owner of the converted object is the new owner, so this is low risk.
+- Unsafe convert in the PATCH handler.
+- Unsafe encode does a hack where is GVK stamps and then removes the stamp in a defer and will need careful review.
+
+Also:
+
+1. Get rid of any internal-only methods
+2. Disable deepcopy gen on the internal type
```
liggitt: once you type-alias, you lose the ability to *have* any internal methods, right?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
liggitt: Type aliasing stable versions as the internal version seems reasonable. At that point, we're likely past radical schema changes, churning API version promotions, etc.

I think that's a clearer and less disruptive plan that gets us the benefits we care about without losing the benefits of a decoupled internal type during alpha/beta development. Can we scope down to that or move the "completely remove the concept of internal types" to a "possible future work" section?

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -0,0 +1,543 @@
...
+- Stop registering the resource under `runtime.APIVersionInternal` and update the
+  encoding configuration (`DefaultResourceEncodingConfig`).
+- Update decode target and the field-management `HubGroupVersion` to the
+  versioned type.
+- Adjust the round-trip fuzz harness to work primarily off the new hub version
+
+#### Step 3
+
+Delete the internal type after moving all handwritten validation and any custom
+functions off of the internal type.
```
liggitt: I'm unclear on the value of these steps. I see most of the performance value in the memory equivalence and the maintenance value in the type aliasing.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: I'm OK with this. I started this with the goal of eliminating the runtime costs of these conversions. If we get down to where we're only performing defensive copies or zero-copy casts (where appropriate) then I've achieved that goal.  I'm fine with leaving the whole "do we need an internal type" topic for another day.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Yes. In my PoC that's why I had to drop deepcopy. I'll explain that better here.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
jpbetz: Moving to potential future work.

keps/sig-api-machinery/6164-internal-type-elimination/README.md
```diff
@@ -179,15 +181,16 @@ internal <-> v1 pod conversion allocs go from O(n) -> O(1)  if the internal type
...
 
 #### Commitment
 
 Our Phase 1 goal is to optimize away the conversion costs via memory-identical types.
 
-If Phase 1 is successful, we will expore Phase 2, starting with type aliasing.
+If Phase 1 is successful, we will proceed to Phase 2.
 We commit to either completing Phase 2 or reverting all code to retain the
-internal types within a 3-release (~1 year) window**.
+distinct internal types within a 3-release (~1 year) window**.
```
liggitt, line 193: I think you can soften this. We already do some type-aliasing of specific types to minimize maintenance costs... even if we didn't complete every last one, I don't think we'd need to revert types we successfully unified / aliased.

liggitt:
lgtm, I'm happy with the current state / immediate plan.

thockin (commented):
I just wanted to +1 this

PR #6170, opened 2026-06-08
KEP-5554: Update to target Alpha in v1.37

esotsal:
- One-line PR description: Update to target Alpha in v1.37
- Includes also updates on checkpoint format as per the good discussion done in this PR, using CPU Manager State generalization checkpoint file version v4.
- Issue link: https://github.com/kubernetes/enhancements/issues/5554

### DIFF
--- a/keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
+++ b/keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
@@ -348,7 +348,7 @@ As it was decided in [SIG Node meeting discussing static CPU policy support](htt
 * Must ensure reserved CPUs from [Kubelet CPU reservation](#kubelet-cpu-reservation) are kept during resize
 * Must work for all [Supported CPU Static Policy Options Combination matrix](#supported-cpu-static-combination-matrix)
 
-Introduction of the inner map items `Original` and `Resized` to checkpoint the corresponding CPUs for a running container ( `ContainerCPUs Checkpoint` for sort ), is meant to make internal upgrade and adaptation easier as we need to track the original CPU set allocated at admission and grant that set never be decreasing. Set must be explicit and recorded clearly in CPUManager checkpoint so it's always simple and safe to derive it. Final naming in the implementation can change a bit without changing concepts as expressed in this KEP.
+[CPU Manager State generalization checkpoint file version v4](https://github.com/kubernetes/kubernetes/pull/139102) `CPUManagerCheckpoint` struct to store exclusive cpu/pod assignments, will be used by `InPlacePodVerticalScalingExclusiveCPUs`, to checkpoint per-container data embedded in `CPUManagerCheckpoint`. More specifically introduction of the inner map item `Originals` to checkpoint the corresponding _historical data_ of original CPUs exclusively allocated to a _running container_ ( `ContainerCPUs Checkpoint` for sort ), is meant to make internal upgrade and adaptation easier as we need to track the original CPU set allocated at admission and grant that set never be decreasing. Set must be explicit and recorded clearly in CPUManager checkpoint so it's always simple and safe to derive it. Final naming in the implementation can change a bit without changing concepts as expressed in this KEP. The corresponding _current_ set of CPUs exclusively allocated to a _running container_ will be checkpointed to `Entries` field embedded in `CheckpointData` section in `CPUManageCheckpoint` structure keeping the same structure used as in `Entries` field in embedded `CPUManagerCheckpointV2` section.
 
 When the CPU Manager, under a static policy, generates a NUMA Topology hint for a Guaranteed pod undergoing an in place CPU pod-resize, it follows these rules to determine the new affinity:
 
@@ -434,41 +434,38 @@ First a one container QoS Guaranteed PoD ( gu-pod-1 ), managed by CPU Static pol
 For educational purposes, in the example we use gu-pod-1-uid, in a real scenario it would be the podUID of gu-pod-1
 
 ```
-{                                                                                                                                                                                             
-  "policyName": "static",                                                                                                                                                                     
-  "defaultCpuSet": "0,3-15",                                                                                                                                                                  
-  "entries": {                                                                           
-    "gu-pod-1-uid": {                                                                                                                                                 
-      "gu-container-1": {                                                                                                                                                                     
-        "original": "1-2",                                                                                                                                                                    
-        "resized": ""                                                                                                                                                                         
-      }                                                                                                                                                                                       
-    }                                                                                                                                                                                         
-  },                                                                                                                                                                                          
-  "checksum": 1759739034                                                                                                                                                                      
-}                                                                                                                                                                                             
+{
+  "policyName": "static",
+  "defaultCpuSet": "0,3-15",
+  "entries": {
+    "gu-pod-1-uid": {
+      "gu-container-1": "1-2"
+    }
+  },
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0,3-15\",\"entries\":{\"gu-pod-1-uid\":{\"gu-container-1\":\"1-2\"}},\"originals"\":{\"gu-pod-1-uid\":{\"gu-container-1\":{\"original\":\"1-2\"}}}}",
+  "dataChecksum": <computed data checksum>
+}
 ```
 
-After the Pods CPUs are increased by 2 CPUs, original is kept 1-2, resized updates with the new CPUset 1-4, and default CPU set with 0,5-15.
+After the Pods CPUs are increased by 2 CPUs, original is kept 1-2, entries updates with the new CPUset 1-4, and default CPU set with 0,5-15.
 
 ```
-{                                                                                                                                                                                             
-  "policyName": "static",                                                                                                                                                                     
-  "defaultCpuSet": "0,5-15",                                                                                                                                                                  
-  "entries": {                                                                                                                                                                                
-    "gu-pod-1-uid": {                                                                                                                                                 
-      "gu-container-1": {                                                                                                                                                                     
-        "original": "1-2",                                                                                                                                                                    
-        "resized": "1-4"                                                                                                                                                                      
-      }                                                                                                                                                                                       
-    }                                                                                                                                                                                         
-  },                                                                                                                                                                                          
-  "checksum": 1822250518                                                                                                                                                                      
-}                                                                                                                                                                                             
+{
+  "policyName": "static",
+  "defaultCpuSet": "0,5-15",
+  "entries": {
+    "gu-pod-1-uid": {
+      "gu-container-1": "1-4"
+    }
+  },
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0,5-15\",\"entries\":{\"gu-pod-1-uid\":{\"gu-container-1\":\"1-4\"}},\"originals"\":{\"gu-pod-1-uid\":{\"gu-container-1\":{\"original\":\"1-2\"}}}}",
+  "dataChecksum": <computed data checksum>
+}
 ```
 
-
-Next the Pods CPUs are decreased by 2 CPUs, original is kept 1-2, resized updates with the new CPUset 1-2, default CPU Set updates with 0,3-15.
+Next the Pods CPUs are decreased by 2 CPUs, original is kept 1-2, entries updates with the new CPUset 1-2, default CPU Set updates with 0,3-15.
 
 ```
 {
@@ -476,15 +473,15 @@ Next the Pods CPUs are decreased by 2 CPUs, original is kept 1-2, resized update
   "defaultCpuSet": "0,3-15",
   "entries": {
     "gu-pod-1-uid": {
-      "gu-container-1": {
-        "original": "1-2",
-        "resized": "1-2"
-      }
+      "gu-container-1": "1-2"
     }
   },
-  "checksum": 4009312802
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0,3-15\",\"entries\":{\"gu-pod-1-uid\":{\"gu-container-1\":\"1-2\"}},\"originals"\":{\"gu-pod-1-uid\":{\"gu-container-1\":{\"original\":\"1-2\"}}}}",
+  "dataChecksum": <computed data checksum>
 }
 ```
+
 Now, a QoS Guaranteed PoD ( gu-pod-2 ), managed by CPU Static policy with two containers with 4 CPUs is created. 
 
 For educational purposes, in the example we use gu-pod-2-uid, in a real scenario it would be the podUID of gu-pod-2
@@ -495,27 +492,20 @@ For educational purposes, in the example we use gu-pod-2-uid, in a real scenario
   "defaultCpuSet": "0,7-15",
   "entries": {
     "gu-pod-1-uid": {
-      "gu-container-1": {
-        "original": "1-2",
-        "resized": "1-2"
-      }
+      "gu-container-1": "1-2"
     },
     "gu-pod-2-uid": {
-      "gu-container-1": {
-        "original": "3-4",
-        "resized": ""
-      },
-      "gu-container-2": {
-        "original": "5-6",
-        "resized": ""
-      }
-    }
+      "gu-container-1": "3-4",
+      "gu-container-2": "5-6",
+    },
   },
-  "checksum": 3881359401
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0,7-15\",\"entries\":{\"gu-pod-1-uid\":{\"gu-container-1\":\"1-2\"},\"gu-pod-2-uid\":{\"gu-container-1\":\"3-4\",\"gu-container-2\":\"5-6\",}},\"originals"\":{\"gu-pod-1-uid\":{\"gu-container-1\":{\"original\":\"1-2\"}},\"gu-pod-2-uid\":{\"gu-container-1\":{\"original\":\"3-4\"},\"gu-container-2\":{\"original\":\"5-6\"},}}}",
+  "dataChecksum": <computed data checksum>
 }
 ```
 
-Next we delete the first created Pod and then we up size the first container of the second Pod, by two CPUs.
+Next we delete gu-pod-1-uid and then we up size gu-container-2 of gu-pod-2-uid, by two CPUs.
 
 ```
 {
@@ -523,39 +513,31 @@ Next we delete the first created Pod and then we up size the first container of 
   "defaultCpuSet": "0,2,8-15",
   "entries": {
     "gu-pod-2-uid": {
-      "gu-container-1": {
-        "original": "3-4",
-        "resized": "1,3-4,7"
-      },
-      "gu-container-2": {
-        "original": "5-6",
-        "resized": "5-6"
-      }
-    }
+      "gu-container-1": "3-4",
+      "gu-container-2": "1,5-7",
+    },
   },
-  "checksum": 287770558
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0,2,8-15\",\"entries\":{\"gu-pod-2-uid\":{\"gu-container-1\":\"3-4\",\"gu-container-2\":\"1,5-7\",}},\"originals"\":{\"gu-pod-2-uid\":{\"gu-container-1\":{\"original\":\"3-4\"},\"gu-container-2\":{\"original\":\"5-6\"},}}}",
+  "dataChecksum": <computed data checksum>
 }
 ```
 
-Last we increase the number of CPUs for the second container by 2 CPUs and reduce the number of CPUs for the first container by 2 CPUs.
+Last we up size gu-container-1 of gu-pod-2-uid by 2, and down size gu-container-2 of gu-pod-2-uid by 2 CPUs.
 
 ```
 {
   "policyName": "static",
-  "defaultCpuSet": "0,2,8-15",
+  "defaultCpuSet": "0-2,7,10-15",
   "entries": {
     "gu-pod-2-uid": {
-      "gu-container-1": {
-        "original": "3-4",
-        "resized": "3-4"
-      },
-      "gu-container-2": {
-        "original": "5-6",
-        "resized": "1,5-7"
-      }
-    }
+      "gu-container-1": "3-4,8-9",
+      "gu-container-2": "5-6",
+    },
   },
-  "checksum": 3738859028
+  "checksum": <computed checksum>,
+  "data": "{\"policyName\":\"static\",\"defaultCpuSet\":\"0-2,7,10-15\",\"entries\":{\"gu-pod-2-uid\":{\"gu-container-1\":\"3-4,8-9\",\"gu-container-2\":\"5-6\",}},\"originals"\":{\"gu-pod-2-uid\":{\"gu-container-1\":{\"original\":\"3-4\"},\"gu-container-2\":{\"original\":\"5-6\"},}}}",
+  "dataChecksum": <computed data checksum>
 }
 ```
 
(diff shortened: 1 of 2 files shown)


keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -671,7 +671,7 @@ type CPUManagerCheckpoint struct {
 ```
 
 In addition to above the following list of functions must be added : 
-* migrateV2CheckpointToV3 to ensure backwards compatibility and upgrade support
+* functions to ensure backwards compatibility and upgrade support
```
ffromani: backward and forward compatibility at load stage are now ensured by V4 format, and node draining/checkpoint file removal is now no longer necessary. The very important caveat now is however that if upgrading from an older version with unsupported/disabled feature, the checkpointed structures will be **empty**. The code need to deal with that.

We probably just need to rephrase the KEP - maybe just remove "ensure backwards compatibility"?

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -885,9 +885,9 @@ enhancement:
   cluster required to make on upgrade, in order to make use of the enhancement?
 -->
 
-Addition of CPUManager checkpoint version 3 doesn't impact an upgrade or downgrade, since CPU Manager current behavior is to create a new checkpoint upon kubelet start, otherwise it will require the Node to be drained, i.e. to delete /var/lib/kubelet/cpu_manager_state file. 
+CPUManager checkpoint modifications doesn't impact an upgrade or downgrade, since CPU Manager current behavior is to create a new checkpoint upon kubelet start, otherwise it will require the Node to be drained, i.e. to delete /var/lib/kubelet/cpu_manager_state file.
```
ffromani: I think V4 format resolves this problem transparently in the sense that we don't need to drain the node or remove the file anymore.

ffromani (commented):
we need minor partial updates to acknowledge the fact the V4 format now handles backward and forward compatibility much more transparently, but other than this LGTM

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski, line 681: The structure looks now a little bit different. The checkpoint is defined as:
type CPUManagerCheckpoint struct {
	// Backward compatibility
	CPUManagerCheckpointV2 `json:",inline"`

	// Data is a serialized CPUManagerCheckpointData
	Data string `json:"data"`
	// DataChecksum is a checksum of Data string
	DataChecksum checksum.Checksum `json:"dataChecksum"`

	// CheckpointData holds actual data, not serialized directly
	CheckpointData CPUManagerCheckpointData `json:"-"`
}

and the changes should be applied only to `CPUManagerCheckpointData` part:
```
type CPUManagerCheckpointData struct {
	PolicyName    string                       `json:"policyName"`
	DefaultCPUSet string                       `json:"defaultCpuSet"`
	Entries       map[string]map[string]string `json:"entries,omitempty"`
	PodEntries    PodCPUAssignments            `json:"podEntries,omitempty"`
}
```

Note, that if you want to keep compatibility with 1.36, then you should not change `Entries` signature in embedded V2 part - `CPUManagerCheckpointV2`, otherwise checkpoint file created in 1.37 won't be understandable by 1.36.
The new V4 part is not yet released, so you can modify the `Entries` field in it to:
`Entries       map[string]map[string]ContainerCPUs`
I assume the changes won't be implemented behind feature gate, so the signature of the field will be exactly the same with exclusive CPUs scaling enabled and disabled.

If you plan to keep "draining the node" requirement for upgrade/downgrade, then the `Entries` will be empty during upgrade/downgrade.
But if you remove this condition form KEP and documentation and allow upgrade/downgrade while pods are deployed on the node (as @ffromani suggested),
then you need to think how `Entries` field in embedded V2 should be filled - they should have all the currently allocated CPUs (`Resized` if not empty, otherwise `Original`)

```
type CPUManagerCheckpointData struct {
	PolicyName      string                              `json:"policyName"`
	DefaultCPUSet   string                              `json:"defaultCpuSet"`
	Entries         map[string]map[string]ContainerCPUs `json:"entries,omitempty"`
	PodEntries      PodCPUAssignments                   `json:"podEntries,omitempty"`
}
```

| file format | kubelet version | InPlacePod VerticalScaling ExclusiveCPUs | description |
| --- | --- | --- | --- |
| V4 | 1.37 | enabled | both `Original` and `Resized` sets read from file and used in CPUManager's assignments (scaling possible) |
| V4 | 1.37 | disabled | both `Original` and `Resized` sets read from file and used in CPUManager's assignments (but due to scaling disabled, they won't change over time) |
| V4 | 1.36- | disabled | embedded V2 data read from checkpoint, assignments in CPUManager set according to `Entries` in V2 part mapping |
| V2 / V3 | 1.37 | enabled | `Original` set based on `Entries` , `Resized` empty (scaling possible, will change `Resized`) |
| V2 / V3 | 1.37 | disabled | `Original` set based on `Entries` , `Resized` empty (but due to scaling disabled, they won't change over time) |

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski: yes, there will be no file format issue, so file reading will succeed.
However to keep forward compatibility the `Entries` field in embedded V2 section must be fill properly with currently allocated CPU sets (Original if there was no resize, Resized if there was a resize)

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Sounds reasonable, will update

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Not sure i understand , are you suggesting to use Entries ? 

 I thought plan was to create a separate Allocations instead for InPlacePodVerticalScalingExclusiveCPUs to have it separate. 
 
Can you please explain ?

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Let’s first clarify https://github.com/kubernetes/enhancements/pull/6170#discussion_r3381789545

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski: In checkpoint V4 with V2 embedded, the `Entries` appears twice:
1) once in embedded `CPUManagerCheckpointV2`
2) in 'CheckpointData' section in `CPUManagerCheckpointData` structure.

I was referring to the first place. It must be filled up properly, so if the checkpoint file is saved in new kubelet and loaded by the old kubelet, the assignments should be read properly. And this is the part that cannot be changed, so it will stay with type `map[string]map[string]string`

For implementing new format for `InPlacePodVerticalScalingExclusiveCPUs` providing distinction between `Original` and `Resize` in second place, you can use whatever you want:
* modify `Entries` there from `map[string]map[string]string` to `map[string]map[string]ContainerCPUs` - as 1.37 will be first release with V4, so you can modify that
* or introduce new `Allocations` as you mentioned.

Does it explain my intention?

esotsal: Thanks for the clarification, yes it explains the intention.

A consensus from the community is required regarding the question about  where to implement the new format for `InPlacePodVerticalScalingExclusiveCPUs` providing distinction between `Original` and `Resize`. 

Should we modify `Entries`  from `map[string]map[string]string` to `map[string]map[string]ContainerCPUs` - as 1.37 will be first release with V4, or introduce new `Allocations` and keep `Entries` untouched ?

@ffromani , @natasha41575 , @tallclair , @SergeyKanzhelev and KEP reviewers what do you think ? 

Should we take the decision now and update the KEP accordingly or do you think this is an implementation detail and can be handled in the implementation PR?

Personally I prefer to get a consensus in this PR, and update the KEP than taking the discussion in the implementation PR, but i  understand we have one week before the KEP freeze this to be decided so might be to short time for some.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Done, please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
ffromani: I don't think we should mutate V2 or V2 compatibility fields. The V2 format is frozen. Fields should only be added in the V4 format.
What changes with V4 is "only" that we don't need to drain the node and/or to delete the checkpoint file explicitly (which is a big UX win!). But the code needs to deal with V2->V4 upgrades, V4->V2 downgrade and V4->V2->V4 rollbacks. I think from the application level perspective there are no real changes.

esotsal: Thanks @ffromani for your response, so i understand that you are in favor to introduce new `Allocations` and keep `Entries` format untouched. Will wait until tomorrow EU timezone and if there are no objections will update the KEP parts to depict your comments and mark it as ready for final review.

lukaszwojciechowski: @esotsal So you plan to do something like this ?
```
type CPUManagerCheckpointData struct {
	PolicyName    string                              `json:"policyName"`
	DefaultCPUSet string                              `json:"defaultCpuSet"`
	Entries       map[string]map[string]string        `json:"entries,omitempty"`
	PodEntries    PodCPUAssignments                   `json:"podEntries,omitempty"`
	Allocations   map[string]map[string]ContainerCPUs `json:"allocations,omitempty"`
}
``` 

If so then what would be the source of truth for data about allocations `Entries` or `Allocations` ?

I think that what @ffromani had on mind was not changing the V2 part (`CPUManagerCheckpointV2`) which must remain unchanged for securing upgrade/downgrade/roundtrip scenarios.

But for `CPUManagerCheckpointData` part, we can apply modifications.

esotsal: Correct this is what i thought, with `InPlacePodVerticalScalingExclusiveCPUs` enabled, source of truth to be Allocations ( since only Allocations will be used ) , when `InPlacePodVerticalScalingExclusiveCPUs` is not enabled `Entries` will be the source of truth ( since Entries will be used ).

lukaszwojciechowski: Have you considered what will happen if the kubelet is restarted with change of feature gate `InPlacePodVerticalScalingExclusiveCPUs`?
or when `InPlacePodVerticalScalingExclusiveCPUs` will go to beta and will be enabled by default (upgrade from hopefully 1.37 to 1.38) ?

ffromani: >I think that what @ffromani had on mind was not changing the V2 part (CPUManagerCheckpointV2) which must remain unchanged for securing upgrade/downgrade/roundtrip scenarios.

Correct. let me re-read the rest of the thread and catch up.

esotsal: > Have you considered what will happen if the kubelet is restarted with change of feature gate `InPlacePodVerticalScalingExclusiveCPUs`? or when `InPlacePodVerticalScalingExclusiveCPUs` will go to beta and will be enabled by default (upgrade from hopefully 1.37 to 1.38) ?

How does having them separated impact those cases ? Please note my original question is what community thinks on the question. So please  share what your opinion is with whys and why not. From my side I am in favour to have it separated as done in PodLevelResources to separate concerns. Also don’t want this KEP to impact other features using Entries , so If there is a blocking issue having Allocations please explain, i might be wrong and missed something important.

lukaszwojciechowski: If you define both `Entries` and `Allocations` in `CPUManagerCheckpointData`, then you need to fill both of them during `storeState()`.
If you don't do this and fill only one of the fields depending on the `InPlacePodVerticalScalingExclusiveCPUs` feature gate setting like [here](https://github.com/kubernetes/kubernetes/pull/129719/changes#diff-abe55fe94cc242fa5db3d3884128129ef2c40bc7f86006f1e10454a6630c1dc4R271-R287), then some scenarios won't work:

#### feature disabled then restart kubelet with feature enabled

The checkpoint written by kubelet before restart contains only `Entries`. After restart it tries to read only `Allocations`. So all data about assignments is lost.

The scenario occurs also during upgrade from 1.37 with alpha with feature disabled and 1.38 beta with feature enabled by default

#### feature enabled then restart kubelet with feature disabled

The checkpoint written by kubelet before restart contains only `Allocations`. After restart it tries to read only `Entries`. So again all data about assignments is lost.

Moreover without making changes to checkpoint marshalling, the `Entries` in embedded V2 part won't be filled either, as with current implementation of checkpoint marshalling it makes a copy of `Entries` from `CPUManagerCheckpointData` and that one is empty.

ffromani: OK, so thing is: disregard V2 data. That format is frozen. We're only talking about V4/CPUManagerCheckpointData and 1.37 *onwards* when the alpha feature lands.

We need per-container extra data to track the exclusive CPUs resizing support.
Currently, `Entries` represents the *current* set of CPUs exclusively allocated to a *container*. I think there's a solid argument to keep it this way even when the changes proposed here lands. Any historical data (e.g - random example - previous CPUset, previous N cpusets, whatever) can be stored in `Allocations` with the necessary format.

I think this approach is both cleanly composable and _not_ awkward even taking into account the evolution of the format.

The alternative would be to mutate `Entries` to record structs:
```
Entries map[string]map[string]Whatever
```

with `Whatever` being a struct holding `entires` (current cpuset) plus whatever else needed. This approach may be cleaner (not clear cut decision IMO) but doesn't seem to improve the reliablity. Disabling the FG _will_ still cause stale/lost resize data if the node state mutates.

Therefore I have a preference towards introducing `Allocations` which IIRC aligns with our former conversations.

@esotsal @lukaszwojciechowski makes sense? am I missing anything?

lukaszwojciechowski: Yes, we are talking only about V4 part.

I see 3 possible approaches here (1 and 2 are the ones mentioned aboved by @ffromani )

1) keep `Entries`, add `Allocations`
It's ok, but we need to fill `Entries` field regardless if the feature is enabled or disabled:
to be able to load from checkpoint with feature disabled
to have `Entries` propagated also to embedded V2 part (unless we modify marshalling)
With this approach some data will be stored twice both in `Entries` add `Allocations`, that's why I prefer 2) or 3) over this approach.

2) change `Entries` to hold `Whatever`s
It's also ok. We just need to know hot to interpret `Whatever` when feature is disabled and how to set `Entries` in V2 part based on that.

3) keep `Entries`, add `Originals` (name can be changed)
Let `Entries` continue to contain currently allocated CPUSets no matter if they are original CPUSet or resized one.
Add new field `Originals` for keeping just the extra information that is required to implement feature.
Without the feature enabled, we can ignore the `Originals` and use `Entries` legacy way.
With this approach we keep modify `Entries` after resize, so they keep the current state for forward compatibility. (the current implementation in [PR](https://github.com/kubernetes/kubernetes/pull/129719) uses `Original` if there was no resize and `Resized` after resize)

ffromani: >     3. keep `Entries`, add `Originals` (name can be changed)
>        Let `Entries` continue to contain currently allocated CPUSets no matter if they are original CPUSet or resized one.
>        Add new field `Originals` for keeping just the extra information that is required to implement feature.
>        Without the feature enabled, we can ignore the `Originals` and use `Entries` legacy way.
>        With this approach we keep modify `Entries` after resize, so they keep the current state for forward compatibility. (the current implementation in [PR](https://github.com/kubernetes/kubernetes/pull/129719) uses `Original` if there was no resize and `Resized` after resize)

Well, this is indeed the approach I was suggesting and advocating for.  I apologize if I caused any confusion. IIRC this is also what we agreed in the past. @esotsal makes sense?

esotsal: @ffromani and @lukaszwojciechowski thanks for the comments, yes it makes sense. We proceed with option 3 then , I will update and push an update for this PR, as soon as possible ( today or by tomorrow )  to cover this.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Acknowledge @lukaszwojciechowski  thanks makes sense, will update to capture this.

lukaszwojciechowski: The above comment is partially outdated as the option 3 was chosen in [above comment](https://github.com/kubernetes/enhancements/pull/6170#discussion_r3403684470)
but the transition cases remain to be considered

esotsal: Correct, my intention was to acknowledge that will capture those in this KEP PR update as well,  to improve clarity, considering the consensus reached moving forward with option 3, as you mention.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Done, PR updated, please take another look. I hope i have covered your comment and the decision for way forward. If you agree please resolve this conversation. if i have misunderstood please modify the PR with a suggestion where is needed.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Done, PR updated, please take another look. I hope i have covered your comment and the decision for way forward. If you agree please resolve this conversation. if i have misunderstood please modify the PR with a suggestion where is needed.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Done, PR updated, please take another look. I hope i have covered your comment and the decision for way forward. If you agree please resolve this conversation. if i have misunderstood please modify the PR with a suggestion where is needed.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -651,27 +651,56 @@ As it was decided in [SIG Node meeting discussing static CPU policy support](htt
...
 * Use of existing CPU Manager checkpoint should be used for storing the “resized” CPUs after a successful resize.
 
-To satisfy this requirement, CPUs allocated upon creation of Guaranteed Pods need to be explicitly checkpointed, for this to happen CPU Manager State checkpoint file version 3 format need to be introduced in CPUManagerCheckpoint struct used to store cpu/pod assignments in a checkpoint v3 format.
+To satisfy this requirement, CPUs allocated upon creation of Guaranteed Pods need to be explicitly checkpointed, for this to happen [CPU Manager State generalization checkpoint file version v4](https://github.com/kubernetes/kubernetes/pull/139102) need to introduce CPUManagerCheckpoint struct to store cpu/pod assignments.
+For `InPlacePodVerticalScalingExclusiveCPUs`, per-container data embedded in `CPUManagerCheckpoint` are needed, to track the exclusive
+CPUs resizing support, more explicitly. 
 
+* the _current_ set of CPUs exclusively allocated to a _container_
+* extra information required by this feature such us any _historical data_ of original CPUs exclusively allocated to a _container_ (e.g original CPUset, previous CPUset or previous N CPUsets)
```
lukaszwojciechowski, line 641: `previous CPUset or previous N CPUsets` - that's something new.
I understand that this is just an example of hypothetical historical data.
Just being curious: Do you plan to use any other data than Original CPUSet in future?

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -651,27 +651,56 @@ As it was decided in [SIG Node meeting discussing static CPU policy support](htt
...
+
+To ensure there are no stale/lost resize data if the node state mutates during an upgrade downgrade and at the same time disabling
+`InPlacePodVerticalScalingExclusiveCPUs`, the needed per-container data will be embedded in `CheckpointData` section in `CPUManagerCheckpoint`
+structure, more specifically: 
+
+* `Entries` field will store the _current_ set of CPUs allocated to a _container_ keeping the same structure used
+   as in `Entries` field in embedded `CPUManagerCheckpointV2` section.
+* `Originals` field will store the _historical data_ of original CPUs exclusively allocated to a _container_
+   ( for alpha it will be the original CPUset only stored ) with the necessary format.
```
lukaszwojciechowski, line 650: `for alpha it will be the original CPUset only stored`, but the `ContainerCPUs` structure below contains also `Resized` field.
Do we need `Resized` field at all ? Isn't  having `Original` and current `Entries`  enough?

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -651,27 +651,56 @@ As it was decided in [SIG Node meeting discussing static CPU policy support](htt
...
+
+type CPUManagerCheckpointData struct {
        PolicyName    string                              `json:"policyName"`
        DefaultCPUSet string                              `json:"defaultCpuSet"`
-       Entries       map[string]map[string]ContainerCPUs `json:"entries,omitempty"`
+       // the _current_ set of CPUs exclusively allocated to a _container_
+       Entries       map[string]map[string]string        `json:"entries,omitempty"`
+       // extra information required by this feature such us any _historical data_ of original CPUs exclusively allocated to a _container_
+       Originals     map[string]map[string]ContainerCPUs `json:"entries,omitempty"`
        Checksum      checksum.Checksum                   `json:"checksum"`
```
lukaszwojciechowski: The `CPUManagerCheckpointData`   structure does not contain `Checksum` field.
To calculate the checksum, the structure is serialized to `Data` string and only then `DataChecksum` is evaluated.

So please remove the `Checksum` from the structure.
```suggestion
```

lukaszwojciechowski (commented):
From the CPUManager checkpoint perspective the changes look good.

There is only one small fix to be made (suggestion made in comment)
and 2 questions - not  blocking acceptance of the KEP

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Correct , it is an example , after implementation  PR is merged for alpha , we can update  the checkpoint example KEP in the beta round or when is suggested by sig node, to make it more accurate. I cannot answer about the future now from our discussion in this PR seems might be such a possibility. Francesco used the term an appropriate structure, perhaps @ffromani  can share his opinion as well. Does this answer your question?

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: You are correct , ’Resized’ is an unnecessary duplication, ’Entries’ is enough. I have kept it as an example perhaps it is better now to update the checkpoint and the example section to improve clarity and depict the consensus reached with option 3. I will do it now together with your suggestion.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, will include this in the next update push. In couple of hours will push the new commit.

esotsal: Done please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Done please take another look. Updated also the checkpoint example.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: For the naming, added the statement to highlight possible modifications in the implementation PR.

_Final naming in the implementation can change slightly without changing concepts as expressed in this KEP._

Please take another look, PR updated.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -434,57 +434,75 @@ First a one container QoS Guaranteed PoD ( gu-pod-1 ), managed by CPU Static pol
...
+      "gu-container-1": "1-2"
+    }
+  },
+  "originals": {
+    "gu-pod-1-uid": {
+      "gu-container-1": {
+       "original": "1-2",
+      }
+   }
+  },
```
lukaszwojciechowski: `originals` must not be added to the legacy embedded V2 part. It only appears in `Data` section of the checpoint.
```suggestion
```

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -434,57 +434,75 @@ First a one container QoS Guaranteed PoD ( gu-pod-1 ), managed by CPU Static pol
...
+      "gu-container-1": "1-4"
+    }
+  },
+  "originals": {
+    "gu-pod-1-uid": {
+      "gu-container-1": {
+       "original": "1-2",
+      }
+   }
+  },
```
lukaszwojciechowski: `originals` must not be added to the legacy embedded V2 part. It only appears in `Data` section of the checpoint.

```suggestion
```

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -434,57 +434,75 @@ First a one container QoS Guaranteed PoD ( gu-pod-1 ), managed by CPU Static pol
...
+  "originals": {
     "gu-pod-1-uid": {
       "gu-container-1": {
-        "original": "1-2",
-        "resized": "1-2"
+       "original": "1-2",
       }
-    }
+   }
```
lukaszwojciechowski:
```suggestion
```

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -494,68 +512,88 @@ For educational purposes, in the example we use gu-pod-2-uid, in a real scenario
...
+       "original": "3-4",
       },
       "gu-container-2": {
-        "original": "5-6",
-        "resized": ""
-      }
+       "original": "5-6",
+      },
     }
   },
```
lukaszwojciechowski:
```suggestion
```

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -494,68 +512,88 @@ For educational purposes, in the example we use gu-pod-2-uid, in a real scenario
...
+       "original": "3-4",
       },
       "gu-container-2": {
-        "original": "5-6",
-        "resized": "5-6"
-      }
+       "original": "5-6",
+      },
     }
   },
```
lukaszwojciechowski:
```suggestion
```

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
```diff
@@ -494,68 +512,88 @@ For educational purposes, in the example we use gu-pod-2-uid, in a real scenario
...
+       "original": "3-4",
       },
       "gu-container-2": {
-        "original": "5-6",
-        "resized": "1,5-7"
-      }
+       "original": "5-6",
+      },
     }
   },
```
lukaszwojciechowski:
```suggestion
```

lukaszwojciechowski (commented):
Looks very good.
The examples containing serialized checkpoint have `originals` also in the embedded V2 part. I added suggestions to remove it

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski: Thanks for the explanation
Having the `ContainerCPUs` in `Originals`, makes that expandable in future.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski: It's fine now, thanks!

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
lukaszwojciechowski: Looks good now, thanks!

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done, please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done. Please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done. Please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done please take another look.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Thanks, done. Please take another look.

lukaszwojciechowski (approved):
lgtm, thanks

esotsal:
> lgtm, thanks

Thanks @lukaszwojciechowski

ffromani (commented):
/lgtm

thanks for mentioning the exact naming can change. Good idea.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Based on https://github.com/kubernetes/enhancements/pull/6170#pullrequestreview-4496492108 , resolving this conversation.

keps/sig-node/5554-in-place-update-pod-resources-alongside-static-cpu-manager-policy/README.md
esotsal: Based on https://github.com/kubernetes/enhancements/pull/6170#pullrequestreview-4496492108 , resolving this conversation.

PR #6177, opened 2026-06-08
Target Volume Limit scaling feature for beta release

gnufied:
https://github.com/kubernetes/enhancements/issues/5030

/sig storage
/sig autoscaling

### DIFF
--- a/keps/sig-autoscaling/5030-attach-limit-autoscaler/README.md
+++ b/keps/sig-autoscaling/5030-attach-limit-autoscaler/README.md
@@ -27,6 +27,7 @@
       - [e2e tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha](#alpha)
+    - [Beta](#beta)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
   - [Version Skew Strategy](#version-skew-strategy)
 - [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
@@ -251,7 +252,7 @@ type CSIDriverSpec struct {
     // if set to true, it will cause scheduler to prevent pod placement
     // to nodes where no CSI driver is installed.
     //   Defaults: false
-    PreventPodPlacementWithoutDriver *bool
+    PreventPodSchedulingIfMissing *bool
 }
 ```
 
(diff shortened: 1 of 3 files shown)


keps/sig-autoscaling/5030-attach-limit-autoscaler/README.md
```diff
@@ -379,12 +380,15 @@ Please note other conditions are already tested via - https://github.com/kuberne
...
 #### Beta
 
-- Gather feedback from developers and surveys
-- Complete features A, B, C
-- Additional tests are in Testgrid and linked in KEP
+Cluster autoscaler is mainly tested via integration style tests that test the whole autoscaler at a higher level. For beta we are going to:
+
+- Further enhance existing tests to cover scale-down behaviours as surfaced during first phase of implementation.
+- We should already have e2es that validate behavior of scheduler with/without CSI driver. 
+- Update metrics for `failed_scale_ups_total` and `scaled_up_nodes_total` to include information about CSI drivers.
```
deads2k, line 389: Please confirm that this the complete featureset. We only want to enable things in beta that are complete and ready for production use since they will be on in production.

deads2k:
PRR lgtm, sig still needs to decide to promote it.

/approve

keps/sig-autoscaling/5030-attach-limit-autoscaler/README.md
gnufied: I do not think there is anything else we want to accomplish in this release. We have some existing issues that were identified during alpha to fix in Autoscaler, but the feature list is complete.

jackfrancis: Adding additional dimensions to existing CA metrics is acceptable alpha-->beta feature intake IMO. We are good here.

keps/sig-autoscaling/5030-attach-limit-autoscaler/README.md
```diff
@@ -507,10 +499,10 @@ No, the scheduler and autoscaler defaul behavior is the same. A CSI driver must
 ###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
 
 For CAS:
-- Yes. This will simply cause old behaviour to be restored, but `PreventPodPlacementWithoutDriver` should be disabled (if enabled manually) in `CSIDriver` object before disabling this feature in CAS.
+- Yes. This will simply cause old behaviour to be restored, but `PreventPodSchedulingIfMissing` should be disabled (if enabled manually) in `CSIDriver` object before disabling this feature in CAS.
```
jackfrancis, line 502: @gnufied what is the backstory behind this change?

gnufied: So, the backstory is - in kube-scheduler we have a plugin that counts number of volumes that can be attached to a node, but nodes where no CSI driver is installed we previously allowed unlimited number of pods to be scheduled (we basically placed no scheduler restrictions). We specifically did this to not cause any problems in autoscaler.

Now we are closing that loophole, but rather than outright changing the behaviour of kube-scheduler, we are making this an opt-in via API in `CSIDriver` object.  

But this feature interacts with CAS, if someone enabled prevention of pods to nodes without CSI driver and they are running a version of CAS which has no CSI awareness or the feature we developed is disabled, they will break autoscaler for pods that use CSI volumes.  So if someone wants to enable `PreventPodSchedulingIfMissing`, they should ensure that they are using a autoscaler which has CSI limit awareness.

jackfrancis: Thank you.

jackfrancis (approved):
/lgtm

Issue #6178, opened 2026-06-08
Concurrent Watch Object Decode


Jefftree:
### Enhancement Description

- One-line enhancement description (can be used as a release note): The kube-apiserver decodes and converts watch events concurrently.
- Related Issue: https://github.com/kubernetes/kubernetes/issues/136950
- PRs by stage and milestone:
  - [ ] Beta (on by default) - v1.37
    - [x] KEP (`k/enhancements`) update PR(s):
      - https://github.com/kubernetes/enhancements/pull/6179
    - [x] Code (`k/k`) update PR(s):
      - [x] https://github.com/kubernetes/kubernetes/pull/139679	
    - [x] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

/sig api-machinery

whtssub:
Hello @Jefftree  👋, v1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `beta` for v1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [x] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

This enhancement has met all the PRR freeze requirements. However, as we look ahead to [enhancements freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#enhancements-freeze) on **Tuesday 16th June 2026 (AoE) / Wednesday 17th June 2026, 12:00 UTC**, this enhancement is currently `At risk for enhancements freeze`. Here's what still needs to be addressed:

[...]

PR #6179, opened 2026-06-08
KEP-6178: Concurrent Watch Object Decode, beta on by default in v1.37

Jefftree:
- One-line PR description: Add the Concurrent Watch Object Decode KEP, documenting the existing feature gate (Beta, disabled by default since v1.31) and targeting Beta on by default in v1.37.
- Issue link: https://github.com/kubernetes/enhancements/issues/6178
- Other comments:

Some data for reviewers:
- Built-in resources: `BenchmarkCacherInit` on 150k pods, concurrent decode alone ~40% faster, ~55% with EtcdRangeStream ([kubernetes/kubernetes#139619](https://github.com/kubernetes/kubernetes/pull/139619)).
- The CRD failure behind this: [kubernetes/kubernetes#136950](https://github.com/kubernetes/kubernetes/issues/136950). A large Kueue CRD took ~8min to convert serially, overran the 5min etcd compaction interval, and the cache never warmed.
- Webhook load stays bounded: init does one object per call, capped at [10 concurrent](https://github.com/kubernetes/kubernetes/blob/de391ca4a4dfe6288287088a8f8ea8bc2352d5b8/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go#L53), while a normal list already sends a whole page (default [500](https://github.com/kubernetes/kubernetes/blob/de391ca4a4dfe6288287088a8f8ea8bc2352d5b8/staging/src/k8s.io/client-go/tools/pager/pager.go#L31)) per ConversionReview.

### DIFF
--- a/keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
+++ b/keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
@@ -258,7 +258,8 @@ No regression in existing list and watch latency SLOs.
 
 ###### Are there any missing metrics that would be useful to improve observability of this feature?
 
-No.
+`apiserver_watch_decode_inflight`, a gauge of watch objects currently being
+decoded concurrently, to track how much of the per-stream decode budget is in use.
 
 ### Dependencies
 
--- a/keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
+++ b/keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
@@ -1,0 +1,370 @@
+# KEP-6178: Concurrent Watch Object Decode
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Choosing the concurrency level](#choosing-the-concurrency-level)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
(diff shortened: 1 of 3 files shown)


keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
```diff
@@ -0,0 +1,370 @@
...
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+It applies to all watches once enabled. Confirm via the apiserver feature gate
+configuration.
+
+###### How can someone using this feature know that it is working for their instance?
+
+No feature-specific metric. Watch cache initialization for a large or
```
jpbetz, line 243: Would there be value in having (a) total active watch object decode goroutines, or (b)  queue depth/delay for watch decode?

(a) might be helpful if we end up with too many goroutines,  (b) might be helpful to diagnose if we don't have enough goroutines (10 seems like a rather arbitrary limit and might not always be the best value?)

jpbetz (commented):
Added a comment to think about what observability we might want to have in place to tune or diagnose this change, otherwise PRR LGTM

keps/sig-api-machinery/6178-concurrent-watch-object-decode/README.md
Jefftree: (a) sounds good. Though I would prefer instead of describing goroutines, I added a gauge for `apiserver_watch_decode_inflight`.

(b) exists already via logs. when decoder is falling behind and crosses a threshold, https://github.com/kubernetes/kubernetes/blob/de391ca4a4dfe6288287088a8f8ea8bc2352d5b8/staging/src/k8s.io/apiserver/pkg/storage/etcd3/block_logger.go.

mimowo:
cc @mykysha who is working on the regression test in Kueue to observe the improvements: https://github.com/kubernetes-sigs/kueue/pull/11680

jpbetz (approved):
Thanks!

/approve
/lgtm

mimowo:
For visibility, graduation of the feature helps projects based on CRDs like Kueue. In particular the previous behavior caused us an incident as reported here: https://github.com/kubernetes/kubernetes/issues/136950 where Kueue could not boot up. 

We also do benchmarking for the newly graduated feature and they demonstrate we can reach  3.4x more workloads; more details in the https://github.com/kubernetes-sigs/kueue/pull/11680#issuecomment-4933626711. Notably this would be more than enough to successfully boot up Kueue at the scale we hit the issue, and so the incident would have been prevented - we estimated it would require 8min to process all the workloads on bootup, so missing the compaction window of 5min, but >3x improvement would have saved the deployment. 

These are very good results, thank you for the work 👍

PR #6180, opened 2026-06-08
KEP-2837: Add instrumentation section and update GA criteria

ndixita:
- One-line PR description: Add instrumentation and update GA criteria

- Issue link: https://github.com/kubernetes/enhancements/issues/2837

### DIFF
--- a/keps/sig-node/2837-pod-level-resource-spec/README.md
+++ b/keps/sig-node/2837-pod-level-resource-spec/README.md
@@ -39,6 +39,15 @@
     - [[Scoped for Beta] HPA](#scoped-for-beta-hpa)
     - [Cluster Autoscaler](#cluster-autoscaler)
     - [VPA](#vpa)
+  - [Instrumentation](#instrumentation)
+    - [Feature Adoption](#feature-adoption)
+      - [<code>kubelet_pod_level_resources_admission_total</code>](#kubelet_pod_level_resources_admission_total)
+    - [The Kubelet (Execution Phase)](#the-kubelet-execution-phase)
+      - [<code>kubelet_oom_kills_total</code>](#kubelet_oom_kills_total)
+      - [<code>kubelet_cpu_cfs_throttled_seconds_total</code>](#kubelet_cpu_cfs_throttled_seconds_total)
+    - [Kube-State-Metrics (KSM)](#kube-state-metrics-ksm)
+      - [<code>kube_pod_level_resource_spec</code>](#kube_pod_level_resource_spec)
+    - [Regression Monitoring (Existing Metrics)](#regression-monitoring-existing-metrics)
   - [Test Plan](#test-plan)
     - [Unit tests](#unit-tests)
     - [Integration tests](#integration-tests)
@@ -78,17 +87,17 @@
 Items marked with (R) are required *prior to targeting to a milestone / release*.
 
 - [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
-- [ ] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
+- [x] (R) KEP approvers have approved the KEP status as `implementable`
+- [x] (R) Design details are appropriately documented
 - [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
   - [ ] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
+- [x] (R) Graduation criteria is in place
   - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
 - [ ] (R) Production readiness review completed
 - [ ] (R) Production readiness review approved
-- [ ] "Implementation History" section is up-to-date for milestone
+- [x] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
 
@@ -1272,6 +1281,67 @@ type RecommendedPodResources struct {
 Note: Detailed KEP design is owned and being worked on by 
 sig-autoscaling: [#7571](https://github.com/kubernetes/autoscaler/issues/7571)
 
+### Instrumentation
+
+This section outlines the final list of metrics for the Pod-Level Resources feature, excluding Resource Manager extensions. These metrics are designed to provide deep observability into admission control, scheduling efficiency, and Kubelet-level execution.
+
+#### Feature Adoption
+These metrics track feature adoption, user intent, and validation friction within the control plane.
+
+##### `kubelet_pod_level_resources_admission_total`
+Total number of pods processed during Kubelet admission, categorized by resource configuration strategy.
+
+**Note:** This metric is **ALPHA** and temporary. It is intended to track feature adoption while the feature is new and is scheduled to be removed 2-3 releases after the Pod-Level Resources feature reaches General Availability (GA).
+
+Labels: 
+- `config_mode` - Possible values: `container_level`, `pod_level_only`, `pod_and_container_level`.
+- `status` - Possible values: `admitted`, `rejected`.
+- `qos_class` - Possible values: `guaranteed`, `burstable`, `best_effort`.
+
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
+Tracks operation failures and resource enforcement during the container lifecycle. Labels are kept low-cardinality to ensure node stability.
+
+##### `kubelet_oom_kills_total`
+Total number of OOM kills triggered. This generic metric uses labels to distinguish between pod-level and container-level events.
+
+Labels:
+- `scope` - Possible values: `pod` (kills triggered specifically because the shared pod-level memory pool was exhausted), `container` (standard container-level OOM kills).
+- `resource` - Always `memory`.
+
+This metric is recorded as a counter.
+
+##### `kubelet_cpu_cfs_throttled_seconds_total`
+Total time in seconds that containers were throttled due to exceeding CPU limits.
+
+Labels:
+- `scope` - Possible values: `pod` (throttling caused by the shared pod-level ceiling), `container` (throttling caused by an individual container's limit).
+
+This metric is recorded as a counter.
+
+#### Kube-State-Metrics (KSM)
+Exposed by the cluster-level state collector. This is the primary source for high-cardinality metadata used in observability joins.
+
+##### `kube_pod_level_resource_spec`
+Exposes the numeric values of the pod-level resources specified in the PodSpec.
+
+Labels:
+- `namespace`: The namespace of the pod.
+- `pod`: The name of the pod.
+- `uid`: The Kubernetes UID of the pod.
+- `node`: The node where the pod is running.
+- `resource`: The resource type (`cpu`, `memory`).
+- `type`: The spec type (`request`, `limit`).
+- `unit`: The unit of measurement (`core`, `bytes`).
+
+#### Regression Monitoring (Existing Metrics)
+While not new, the following metrics must be monitored to ensure no regressions in scheduling or node stability occur after adopting pod-level resource specifications.
+
+- `schedule_attempts_total{result="error|unschedulable"}`: Monitored to detect spikes in unschedulable pods due to potential bugs in the new resource requirement calculation.
+- `node_collector_evictions_total`: Monitored to ensure that the new pod eviction ranking logic based on pod-level requests behaves as expected and does not lead to an increase in unintended evictions.
+- `started_pods_errors_total` / `started_containers_errors_total`: Monitored to detect failures in the Kubelet's ability to create and configure the pod-level cgroups and container sandboxes.
+
 ### Test Plan
 
 [X] I/we understand the owners of the involved components may require updates to
@@ -1343,7 +1413,6 @@ feature gate and by setting the new `resources` fields in PodSpec at Pod level.
 
 #### GA (stable)
 
-* VPA integration of feature moved to beta.
 * No major bugs reported for 3 months.
 * Pod Level Resources Support With In Place Pod Vertical Scaling KEP is past alpha.
 * User feedback (ideally from at least two distinct users) is green
@@ -1671,9 +1740,9 @@ checking if there are objects with field X set) may be a last resort. Avoid
 logs or events for this purpose.
 -->
 
-This feature will be built into kubelet, API server and scheduler. In order to determine if the feature
-is being used by the workloads, check the  `resources` field at pod level in the spec. There's no special
-metric planned to track the usage of this feature at the moment.
+Operators can use the `kubelet_pod_level_resources_admission_total` metric to track the adoption of this feature. This metric is categorized by resource configuration strategy (`config_mode`), admission status, and QoS class. Additionally, `kube_pod_level_resource_spec` (Kube-State-Metrics) can be used to identify specific pods using the feature and their configured resource values.
+
+**Note:** This metric is **ALPHA** and temporary. It is intended to track feature adoption while the feature is new and is scheduled to be removed 2-3 releases after the Pod-Level Resources feature reaches General Availability (GA).
 
 ###### How can someone using this feature know that it is working for their instance?
 
@@ -1718,12 +1787,15 @@ Pick one more of these and delete the rest.
 
 - [X] Metrics
   - Metric name:
-  - `apiserver_rejected_requests` will indicate any failures (`Bad Request` code=400) related to translation of new `resources` field in PodSpec. 
+  - `kubelet_pod_level_resources_admission_total` (**ALPHA, Temporary**): Total number of pods processed during Kubelet admission, categorized by resource configuration strategy. Scheduled for removal 2-3 releases after GA.
+  - `kubelet_oom_kills_total`: Total number of OOM kills triggered. Use label `scope="pod"` to identify kills caused specifically by the shared pod-level limit.
+  - `kubelet_cpu_cfs_throttled_seconds_total`: Total time in seconds that containers were throttled. Use label `scope="pod"` to identify throttling caused by the shared pod-level ceiling.
+  - `kube_pod_level_resource_spec` (Kube-State-Metrics): Exposes the numeric values of pod-level resources specified in the PodSpec.
   - `schedule_attempts_total{result="error|unschedulable"}`
   - `node_collector_evictions_total`: to check if a pod level resource setting is causing to evict more pods than normal
   - `started_pods_errors_total`: exposed by kubelet to check if large number of pods are failing unusually
   - `started_containers_errors_total`: exposed by kubelet to check if large number of containers are failing unusually
-  - Components exposing the metric: apiserver, kubelet, scheduler
+  - Components exposing the metric: apiserver, kubelet, scheduler, kube-state-metrics
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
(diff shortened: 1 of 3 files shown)


keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
 sig-autoscaling: [#7571](https://github.com/kubernetes/autoscaler/issues/7571)
 
+### Instrumentation
+
+This section outlines the final list of metrics for the Pod-Level Resources feature, excluding Resource Manager extensions. These metrics are designed to provide deep observability into admission control, scheduling efficiency, and Kubelet-level execution.
+
+#### Admission & API Validation
+These metrics track feature adoption, user intent, and validation friction within the control plane.
+
+##### `pod_level_resources_admission_total`
```
tallclair:
```suggestion
##### `kubelet_pod_level_resources_admission_total`
```

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+Total number of pods processed during Kubelet admission, categorized by resource configuration strategy.
```
tallclair, line 1292: How do you anticipate using this? I wonder whether kube-state-metrics could meet the use case?

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+##### `pod_level_resources_defaulting_total`
+Total number of times pod-level resource fields were automatically populated by API server defaulting logic.
+
+Labels:
+- `resource` - Possible values: `cpu`, `memory`, `hugepages`.
+- `strategy` - Possible values: `request_from_container_requests`, `request_from_pod_limit`, `hugepage_limit_from_container_limits`.
+
+This metric is recorded as a counter.
+
+##### `pod_level_resources_validation_errors_total`
```
tallclair: Do we have any generic metrics tracking validation errors? I wonder whether we could count validation errors by field path (stripping out specific identifiers like index or key), or if the cardinality would be too high?

/cc @jpbetz @yongruilin

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+Total number of pods processed during Kubelet admission, categorized by resource configuration strategy.
+
+Labels: 
+- `config_mode` - Possible values: `container_level`, `pod_level_only`, `pod_and_container_level`.
+- `status` - Possible values: `admitted`, `rejected`.
+- `qos_class` - Possible values: `guaranteed`, `burstable`, `best_effort`.
+
+This metric is recorded as a counter.
+
+##### `pod_level_resources_defaulting_total`
```
tallclair: AFAIK we don't have any metrics measuring defaulting or validation decisions. I don't know if there's a technical reason for this or if we just haven't had a use case for it, but I'm hesitant to introduce the metrics here.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+Labels:
+- `reason` - Possible values: `sum_mismatch`, `negative_value`, `unsupported_resource`, `request_limit_mismatch`, `forbidden_field`.
+
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
+Tracks operation failures during the container lifecycle that are specific to the shared pod-level resource pool.
+
+##### `kubelet_pod_level_oom_kills_total`
+Total number of OOM kills triggered specifically because the shared pod-level memory pool was exhausted. This metric is crucial for identifying cases where a container was killed even if it was under its own individual limit, but the pod's aggregate limit was reached.
```
tallclair: How will this be measured / detected?

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+
+This metric is recorded as a counter.
+
+#### Resource State (State Metrics)
+These metrics expose the current state of Pod-level resource requests and limits, primarily for consumption by kube-state-metrics and observability dashboards.
+
+##### `kube_pod_level_resources_requests`
```
tallclair: I think drop the `level` on these. Either `kube_pod_resources_requests` or `kube_pod_spec_resources_requests` would be more consistent with the other metrics.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+Exposes the value of `pod.spec.resources.requests` for each resource type.
+
+Labels:
+- `pod` - Name of the pod.
+- `namespace` - Namespace of the pod.
+- `uid` - Kubernetes UID of the pod.
+- `resource` - The resource type (e.g., `cpu`, `memory`).
+- `unit` - The unit of the resource (e.g., `core`, `bytes`).
```
tallclair: The container level metrics also include a node label.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1279,77 @@ type RecommendedPodResources struct {
...
+#### Resource State (State Metrics)
+These metrics expose the current state of Pod-level resource requests and limits, primarily for consumption by kube-state-metrics and observability dashboards.
+
+##### `kube_pod_level_resources_requests`
+Exposes the value of `pod.spec.resources.requests` for each resource type.
+
+Labels:
+- `pod` - Name of the pod.
+- `namespace` - Namespace of the pod.
+- `uid` - Kubernetes UID of the pod.
```
natasha41575: This labels strike me as having extremely high cardinality. Or is there a reason we don't need to worry about that here?

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: This is to monitor the adoption and if there's any friction in the new validation.

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: Yeah, I think we can skip this.

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: If you agree that this is worth tracking, we could measure by:
1. fetching memory.events for all containers and pod cgroup using cadvisorStatsPRovider
2. pod limit ooms = pod ooms - sum(container ooms)
IMO it is worth tracking as pod-level resources allow resource sharing among the containers. So this metric could help understand if pod-level limit is set correctly or not.

keps/sig-node/2837-pod-level-resource-spec/README.md
yongruilin: Good question — we don't have a generic one today; the existing `apiserver_validation_*` metrics only track declarative-vs-handwritten parity.

I think the catch is cardinality: raw field paths are unbounded. We could strip subscripts (`spec.containers[].resources.limits[]`) or label by `field.Error.Origin` (the rule kind) instead.

But it feels more like declarative-validation-framework infra though, not really this KEP — maybe a separate issue under sig/api-machinery? Keeping `pod_level_resources_validation_errors_total` here for now.

jpbetz: Cardinality is already high just for the individual fields across all versions of all APIs. If then we try to multiply that by any other labels it get's completely out of control (even resources are risky this way... consider the CRD cases).  Better for feature owners to identify what is actually needed an add metrics for just those neesd.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1284,77 @@ type RecommendedPodResources struct {
 Note: Detailed KEP design is owned and being worked on by 
 sig-autoscaling: [#7571](https://github.com/kubernetes/autoscaler/issues/7571)
 
+### Instrumentation
+
+This section outlines the final list of metrics for the Pod-Level Resources feature, excluding Resource Manager extensions. These metrics are designed to provide deep observability into admission control, scheduling efficiency, and Kubelet-level execution.
+
+#### Admission & API Validation
+These metrics track feature adoption, user intent, and validation friction within the control plane.
```
jpbetz, line 1289: @richabanker how are metrics for feature adoption handled?  Do we keep them in alpha and then deprecate them after 2-3 release?  This would be my preference. If so let's make notes here about that plan.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1284,77 @@ type RecommendedPodResources struct {
...
+#### The Kubelet (Execution Phase)
+Tracks operation failures during the container lifecycle that are specific to the shared pod-level resource pool.
+
+##### `kubelet_pod_level_oom_kills_total`
+Total number of OOM kills triggered specifically because the shared pod-level memory pool was exhausted. This metric is crucial for identifying cases where a container was killed even if it was under its own individual limit, but the pod's aggregate limit was reached.
+
+This metric is recorded as a counter.
+
+#### Resource State (State Metrics)
+These metrics expose the current state of Pod-level resource requests and limits, primarily for consumption by kube-state-metrics and observability dashboards.
```
jpbetz: kube-state-metrics is informer based, not metric based.

Let's drop the metrics listed here for that purpose. If we need this information, we should open a PR against kube-state-metrics after this feature merges.

Specifically, I recommend dropping:
- `kube_pod_level_resource_requests`
- `kube_pod_level_resource_limits`

keps/sig-node/2837-pod-level-resource-spec/README.md
jpbetz: +1. See my comment above about kube-state-metrics. I don't think we need these metrics at all.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1284,77 @@ type RecommendedPodResources struct {
...
+
+Labels:
+- `reason` - Possible values: `sum_mismatch`, `negative_value`, `unsupported_resource`, `request_limit_mismatch`, `forbidden_field`.
+
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
+Tracks operation failures during the container lifecycle that are specific to the shared pod-level resource pool.
+
+##### `kubelet_pod_level_oom_kills_total`
```
jpbetz:
```suggestion
##### `kubelet_pod_oom_kills_total`
```

? (I don't think we need the feature name here, the fact that it's a pod oom is enough)

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1284,77 @@ type RecommendedPodResources struct {
...
+
+##### `pod_level_resources_validation_errors_total`
+Total number of pods rejected specifically due to validation errors in pod.spec.resources.
+
+Labels:
+- `reason` - Possible values: `sum_mismatch`, `negative_value`, `unsupported_resource`, `request_limit_mismatch`, `forbidden_field`.
+
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
```
jpbetz, line 1303: Recommend adding a "pod CPU throttling metric".  Maybe `pod_cpu_cfs_throttled_seconds_total` ?

keps/sig-node/2837-pod-level-resource-spec/README.md
jpbetz: +1. This should be the main adoption metric (ALPHA, removed in a future release).  I recommend dropping the other adoption related metrics listed below.

Let's make a note here that this metrics is ALPHA, temporary, and removed in 2-3 release.

keps/sig-node/2837-pod-level-resource-spec/README.md
richabanker: K8s already exposes `kubernetes_feature_enabled` metric showing whether a feature is enabled / disabled in a component, can that be used to track "feature adoption" in a way if all we care to know about is whether this feature was on/off ?

ndixita: We want to track how many pods have pod-level resources set. Even after enabling the feature gate, it is required to explicitly set resources at pod-level in the spec to use this functionality. Does that make sense?

richabanker: Then we would never want to deprecate this metric even when the feature graduates to GA right?

jpbetz: Can you track this with kube state metrics does have to be a metric?  If we can use kube state metrics that's usually better for this sort of thing

richabanker: Oh taking a step back, actually adding a permanent metric just to parse a resource's spec to capture presence/absence of a field will be an anti-pattern. We would generally want metrics if they help with determining operational health (health / availability / latency) of features , basically something actionable for cluster admins. 

And regarding KSM, agree that if we want to expose metrics about a resource's spec/status, KSM would be the way to go. But since pod.spec.resource is not natively tracked in existing KSM metrics today, you'd have to add a new metric in the KSM repo (similar to kube_pod_level_resources_requests being proposed in this PR) that can parse this new field to convert that to a metric. Currently it only exposes these metrics for pod resource https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md

Another way to use KSM for this would be to add a label onto pod metadata with the value you want to track, then configure KSM with `--metric-labels-allowlist=pods=[<label>]` which will make the existing kube_pod_labels KSM metric to show how many pods have this label set. Also cc @dgrisonnet to confirm if thats the right way to go about feature adoption metrics depending on resource spec.

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: Done

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: removed the metric

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: done

jpbetz (approved):
/approve
For PRR

keps/sig-node/2837-pod-level-resource-spec/README.md
tallclair: > Better for feature owners to identify what is actually needed an add metrics for just those neesd.

We could have an opt-in allowlist. So there's still one metric, but rather than every validation error across every field, it's only validation errors that a feature owner has asked to track. I don't know if this is a common enough need, but it would be lighter weight than adding a new metric for every validation we want to track.

Either way, out of scope for this KEP.

keps/sig-node/2837-pod-level-resource-spec/README.md
tallclair: Do we record any OOM metrics?

It does seem worth including these metrics. I think the best way to do it would be to expand the scope of the OOM watcher. Currently it only scans OOM events for OOMs in the system cgroup. Instead, we could evaluate each OOM event, and check the `VictimContainer` field to determine what type of cgroup (and probably also which constraint) triggered the OOM. We would need to further parse the cgroups paths to determine if it was a pod cgroup, container cgroup, or some other cgroup.

https://github.com/kubernetes/kubernetes/blob/9d6e94a40d9a4f6d1e69a77c9074cab5f32104bf/pkg/kubelet/oom/oom_watcher_linux.go#L74

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1280,45 @@ type RecommendedPodResources struct {
...
+- `config_mode` - Possible values: `container_level`, `pod_level_only`, `pod_and_container_level`.
+- `status` - Possible values: `admitted`, `rejected`.
+- `qos_class` - Possible values: `guaranteed`, `burstable`, `best_effort`.
+
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
+Tracks operation failures during the container lifecycle that are specific to the shared pod-level resource pool.
+
+##### `kubelet_pod_oom_kills_total`
```
tallclair: Per my comment below, let's rename this and include the scope as a label.
```suggestion
##### `kubelet_oom_kills_total`
```

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1280,45 @@ type RecommendedPodResources struct {
...
+Total number of OOM kills triggered specifically because the shared pod-level memory pool was exhausted. This metric is crucial for identifying cases where a container was killed even if it was under its own individual limit, but the pod's aggregate limit was reached.
+
+This metric is recorded as a counter.
+
+##### `pod_cpu_cfs_throttled_seconds_total`
+Total time in seconds that containers in a pod were throttled due to exceeding pod-level CPU limits. This metric helps identify pods that are consistently hitting their aggregate CPU limits.
```
tallclair: Similar question to the OOM events: how do we track this?

Are there any existing metrics that track CFS throttling? It looks like the cadvisor metrics do, but I can't remember the status of the cadvisor metrics.

Unlike with OOMs, these aren't discrete events, so we'd need to add polling for this. We're already polling and parsing this information for the stats provider. Let's be careful not to duplicate that work.

tallclair: Let's also mirror the oom_kill metric structure. So this would be a single metric, with labels for pod and container (not sure `other` makes sense here).

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1280,45 @@ type RecommendedPodResources struct {
 Note: Detailed KEP design is owned and being worked on by 
 sig-autoscaling: [#7571](https://github.com/kubernetes/autoscaler/issues/7571)
 
+### Instrumentation
+
+This section outlines the final list of metrics for the Pod-Level Resources feature, excluding Resource Manager extensions. These metrics are designed to provide deep observability into admission control, scheduling efficiency, and Kubelet-level execution.
+
+#### Admission & API Validation
```
tallclair:
```suggestion
#### Feature Adoption
```

keps/sig-node/2837-pod-level-resource-spec/README.md
dgrisonnet: I agree with @richabanker, these metrics would be an anti-pattern in k/k.

One of the problem of metrics tracking feature enablement upstream is that we don't have a centralized cluster to which this data would be reported. For downstream, this pattern makes a lot more sense, but for upstream the majority of users will bear the cost of these metrics without being able to do anything with them.

 With regards to KSM, I think it could be a great fit. The project already exposes metrics about container-level resources so adding pod-level would make sense 👍

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: we do have [container_oom_events_total](https://github.com/google/cadvisor/blob/195858077459e69455fd9621fcbaeaf377d69d0e/metrics/prometheus.go#L1879) in cadvisor which monitors  /dev/kmsg and when the event happens cadvisor increments the counter for oom events. I like the idea to extend oom watcher. Taking a note of it and we can discuss it further on the implementation PR. 

https://github.com/google/cadvisor/blob/195858077459e69455fd9621fcbaeaf377d69d0e/utils/oomparser/oomparser.go#L119

keps/sig-node/2837-pod-level-resource-spec/README.md
ndixita: changed to kubelet_cpu_cfs_throttled_seconds_total
cadvisor does track container_cpu_cfs_throttled_seconds_total so similarly we should be able to get the cpu.stat info from cadvisorapiv2.ContainerInfo object. 
https://github.com/kubernetes/kubernetes/blob/808a553dc88f0343a895e4bbf9917107500b6488/pkg/kubelet/stats/cadvisor_stats_provider.go#L426

I spoke with @jpbetz and the cadence here is to track the metric in the KEP, and if it is not implementable we can drop it later.

keps/sig-node/2837-pod-level-resource-spec/README.md
```diff
@@ -1272,6 +1281,67 @@ type RecommendedPodResources struct {
...
+This metric is recorded as a counter.
+
+#### The Kubelet (Execution Phase)
+Tracks operation failures and resource enforcement during the container lifecycle. Labels are kept low-cardinality to ensure node stability.
+
+##### `kubelet_oom_kills_total`
+Total number of OOM kills triggered. This generic metric uses labels to distinguish between pod-level and container-level events.
+
+Labels:
+- `scope` - Possible values: `pod` (kills triggered specifically because the shared pod-level memory pool was exhausted), `container` (standard container-level OOM kills).
```
tallclair, line 1310: I think we should add `other` for OOM kills that don't match a Kubelet-aware cgroup.

tallclair (commented):
/lgtm
/aprove

PR #6184, opened 2026-06-09
KEP 5419: Add instrumentation section and update GA criteria

ndixita:
- One-line PR description: Add instrumentation, swap support and update GA criteria

- Issue link: https://github.com/kubernetes/enhancements/issues/5419

### DIFF
--- a/keps/prod-readiness/sig-node/5419.yaml
+++ b/keps/prod-readiness/sig-node/5419.yaml
@@ -3,3 +3,5 @@ alpha:
   approver: "@soltysh"
 beta:
   approver: "@soltysh"
+stable:
+  approver: "@soltysh"
--- a/keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
+++ b/keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
@@ -20,13 +20,17 @@
       - [Goals of surfacing Pod Resource Requirements](#goals-of-surfacing-pod-resource-requirements)
       - [Implementation Details](#implementation-details-1)
       - [Notes for implementation](#notes-for-implementation)
+  - [Instrumentation](#instrumentation)
+    - [New Metrics (Pod-Level Specific)](#new-metrics-pod-level-specific)
+    - [Extended Metrics (From KEP-1287)](#extended-metrics-from-kep-1287)
+    - [Standardized Boolean Labels](#standardized-boolean-labels)
   - [Test Plan](#test-plan)
     - [Unit tests](#unit-tests)
     - [Integration tests](#integration-tests)
     - [e2e tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
     - [Phase 1: Alpha (target 1.35) [DONE]](#phase-1-alpha-target-135-done)
-    - [Phase 2:  Beta (target 1.36)](#phase-2--beta-target-136)
+    - [Phase 2:  Beta (target 1.36) [DONE]](#phase-2--beta-target-136-done)
     - [GA (stable)](#ga-stable)
   - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
       - [Upgrade](#upgrade)
@@ -58,10 +62,10 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
   - [ ] e2e Tests for all Beta API Operations (endpoints)
   - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
   - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
-- [ ] (R) Graduation criteria is in place
-  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
-- [ ] (R) Production readiness review completed
-- [ ] (R) Production readiness review approved
+- [X] (R) Graduation criteria is in place
+  - [X] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [X] (R) Production readiness review completed
+- [X] (R) Production readiness review approved
 - [X] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
@@ -423,7 +427,7 @@ PodStatus.Resources field is set in
 [generateAPIPodStatus](https://github.com/kubernetes/kubernetes/blob/a668924cb60901b413abc1fe7817bc7969167064/pkg/kubelet/kubelet_pods.go#L1459)
 method for now. 
 
-Note: We'll need to revisit this to enable controllers to utilize this field in 1.35. The motivation for defaulting PodStatus.Resources is to allow components, such as admission (e.g., quota controls) or any controllers that run before the pod starts, to use it for calculating pod-level resource totals, rather than relying on a component-helper.
+Note: The decision to set `PodStatus.Resources` in the Kubelet ensures that the reported values reflect the actual resources admitted and allocated on the node. This field is populated during pod status generation and serves as the source of truth for controllers to quickly determine effective resource requirements.
 
 2. Update the
 [PodRequestsAndLimitsReuse](https://github.com/kubernetes/kubernetes/blob/dfc9bf0953360201618ad52308ccb34fd8076dff/pkg/api/v1/resource/helpers.go#L64)
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
 implemented, will in a small way reduce the effort for both of those KEPs by
 providing a single place to update the pod resource calculation.
 
+### Instrumentation
+
+The Kubelet will record the following metrics. These are categorized by whether they are new additions for Pod-level resources or existing metrics (from KEP-1287) that have been extended.
+
+#### New Metrics (Pod-Level Specific)
+
+These metrics are introduced to provide dedicated observability into Pod-level resize intent and state.
+
+* **`kubelet_pod_requested_resizes_total`**
+  * *Description:* Tracks the total number of Pod-level resize attempts observed by the Kubelet.
+  * *Labels:* `container_resize`, `pod_level_resize`, `resource`, `requirement`, `operation`, `namespace`.
+
+* **`kubelet_pod_pending_resizes`**
+  * *Description:* Tracks the current count of pods that the Kubelet marks as pending.
+  * *Labels:* `reason`, `reason_detail`, `namespace`, `container_resize`, `pod_level_resize`.
+
+* **`kubelet_pod_in_progress_resizes`**
+  * *Description:* Tracks the total count of resize requests that the Kubelet marks as in progress.
+  * *Labels:* `namespace`, `container_resize`, `pod_level_resize`.
+
+* **`kubelet_pod_deferred_resize_accepted_total`**
+  * *Description:* Tracks the total number of resize requests that the Kubelet originally marked as deferred but later accepted.
+  * *Labels:* `accepted_reason`, `namespace`, `container_resize`, `pod_level_resize`.
+
+#### Extended Metrics (From KEP-1287)
+
+These metrics are inherited from KEP-1287 and have been enhanced with new labels to capture the interaction between container-level and pod-level resizes.
+
+* **`kubelet_container_requested_resizes_total`**
+  * *Description:* Tracks the total number of container-level resize attempts.
+  * *Labels (Enhanced):* Includes original labels (`resource`, `requirement`, `operation`, `namespace`) plus new `container_resize` and `pod_level_resize` boolean labels.
+
+* **`kubelet_pod_resize_duration_seconds`**
+  * *Description:* Tracks the duration of `doPodResizeAction` for actuating resizes.
+  * *Labels (Enhanced):* Includes original labels (`namespace`) plus new `container_resize` and `pod_level_resize` boolean labels.
+
+#### Standardized Boolean Labels
+
+To ensure orthogonality across all metrics, the following boolean labels are now applied to **both new and extended** metrics:
+
+* **`container_resize`**: `true` | `false`
+* **`pod_level_resize`**: `true` | `false`
+
 ### Test Plan
 
 [X] I/we understand the owners of the involved components may require updates to
@@ -465,12 +512,20 @@ necessary to implement this enhancement.
 #### Unit tests
 
 `k8s.io/kubernetes/pkg/kubelet/cm`: `20250618` - 18.4
+- Added tests for pod-level cgroup configuration and resizing in `container_manager_linux_test.go` and `node_container_manager_linux_test.go`.
+- Validated cgroup v2 specific logic for pod-level resource limits.
 
 `k8s.io/kubernetes/pkg/kubelet/kuberuntime`: `20250618` - 69.1
+- Added unit tests in `kuberuntime_sandbox_test.go` to verify `UpdatePodSandboxResources` calls.
+- Updated `kuberuntime_pod_test.go` to test pod-level resource tracking in internal checkpoints.
 
 `k8s.io/kubernetes/pkg/apis/core/validation` - `20250618` - 84.7
+- Added comprehensive validation tests for the new `Resources` field in `PodSpec` and `PodStatus` in `validation_test.go`.
+- Verified mutation rules and immutability (or lack thereof) for pod-level resources.
 
 `k8s.io/kubernetes/pkg/scheduler/framework` - 20250618 - 71.7 
+- Added unit tests for the `NodeResourcesFit` plugin to correctly account for pod-level requests in `pkg/scheduler/framework/plugins/noderesources/fit_test.go`.
+- Verified that pod-level resources take precedence over container-level aggregations when specified.
 
 #### Integration tests
 
@@ -484,15 +539,22 @@ covered by planned and implemented tests.
 
   * - [Pod Level Resources Resize](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/node/pod_level_resources_resize.go): [SIG Node](https://testgrid.k8s.io/sig-node-presubmits#pr-kubelet-e2e-podlevelresources-resize), [triage search](https://storage.googleapis.com/k8s-triage/index.html?ci=0&pr=1&sig=node&job=pr-kubelet-e2e-podlevelresources-resize)
 
+The following e2e scenarios were implemented and verified during Alpha and Beta:
+- **Basic Resize**: Updating `spec.resources` at the pod level and verifying cgroup updates on the node.
+- **Mixed Resize**: Simultaneously updating pod-level and container-level resources and verifying the atomic nature of the update.
+- **Restart Policy Interaction**: Verifying that containers are restarted (or not) based on their `resizePolicy` when pod-level limits change.
+- **Guaranteed Pods Resize**: Independent, simultaneous same-direction, and simultaneous opposite-direction resizing of CPU and Memory at both container-level and Pod-level (including multi-container net increase/decrease scenarios and restart policy evaluation).
+- **Burstable Pods Resize**: Scaling requests and limits independently or concurrently, including dynamically adding limits/requests across multiple containers, equivalent value resizing (e.g., `2m` -> `1m`), and unrestricted container handling.
+- **Memory Limit Decrease**: Actuating viable memory limit reductions while verifying that attempting to lower memory limits below active usage fails and reports a `PodResizeInProgress` error condition.
+- **Initial Creation Event Suppression**: Validating that initial Pod creation does not trigger spurious `ResizeCompleted` events.
+- **Correctness of Resources Values**: Validating the correctness of values reported in `Status.Resources`.
+- **Cgroup Settings Validation**: Verifying that cgroup settings are correctly applied to the Pod's cgroup.
+- **Scheduling and Admission**: Validating that the scheduler and admission controller correctly handle pod-level resources.
 
-Following scenarios need to be covered:
+Following scenarios are to be covered as part of GA graduation:
 
-* Cgroup settings when pod-level resources are set.
-* Validate scheduling and admission.
 * Validate the containers with no limits set are throttled on CPU when CPU usage reaches Pod level CPU limits.
-* Validate the containers with no limits set are OOMKilled when memory usage
-  reaches Pod level memory limits.
-* Test the correct values in TotalResourcesRequested.
+* Validate the containers with no limits set are OOMKilled when memory usage reaches Pod level memory limits.
 
 ### Graduation Criteria
 
@@ -508,8 +570,7 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
 * Unit test coverage.
 * E2E tests.
 
-#### Phase 2:  Beta (target 1.36)
-
+#### Phase 2:  Beta (target 1.36) [DONE]
 * Pod Level Resources Feature moved to beta.
 * The semantic of `UpdatePodSandboxResources` is clarified. And there is a way for container runtime to reject the resize of Pod resources via this method or by other means
 * Actual pod resource data may be cached in memory, which will be refreshed after
@@ -518,13 +579,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
 * Extend instrumentation from
   [KEP#1287](https://github.com/kubernetes/enhancements/blob/ef7e11d088086afd84d26c9249a4ca480df2d05a/keps/sig-node/1287-in-place-update-pod-resources/README.md)
   for Pod-level resource resize.
-* Revisit the decision of which component sets the defaults for PodStatus.Resources.
+* Decision finalized: `PodStatus.Resources` defaults are set by the Kubelet, not the API server, to ensure accuracy based on node-level admission and allocation. [DONE]
 
 #### GA (stable)
 
-* VPA Integration of In-Place Resize moved to beta.
 * No major bugs reported for 3 months.
 * UpdatePodSandboxResources is implemented by containerd & CRI-O
+* [PodLevelResources] Double-counting Pod Overhead [kubernetes/kubernetes#139627](https://github.com/kubernetes/kubernetes/issues/139627) (listed as blocker for GA)
+* [PodLevelResources] Inconsistency in computing QOSClass [kubernetes/kubernetes#135082](https://github.com/kubernetes/kubernetes/issues/135082)
+* [PodLevelResources] CPU Request in PodSpec Status.Resources becomes invalid [kubernetes/kubernetes#137628](https://github.com/kubernetes/kubernetes/issues/137628)
+* PodLevelResources: Why is pod created when resource requests exceed limits? (K8s 1.34.5) [kubernetes/kubernetes#138473](https://github.com/kubernetes/kubernetes/issues/138473)
 
 ### Upgrade / Downgrade Strategy
 
@@ -693,7 +757,7 @@ Testing plan:
 * Verify original test pod is still running
 
 Initial manual verification was completed following the Alpha release ([Results](https://docs.google.com/document/d/19dKnTxH34YjSzrQCMqmpNp9iJk4YfWXXf5ytqNvV4c0/edit?usp=sharing)).
-Comprehensive automated testing and E2E coverage are slated for implementation prior to GA graduation."
+Comprehensive automated testing and E2E coverage have been fully implemented and verified for GA graduation.
 
 ###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
 
@@ -734,7 +798,7 @@ Recall that end users cannot usually observe component logs or access metrics.
 
 * If the Kubelet supports InPlacePodLevelResourcesVerticalScaling, it will always
   set the Resources field in Pod status.
-* The ResizeStatus in the pod status should converge to the empty value, indicating the resize has completed.
+* The `PodResizePending` and `PodResizeInProgress` pod conditions should be cleared, indicating the resize has completed.
 * The Resources in the pod and container statuses should converge to the resized resources, or an approximation of it.
 
 ###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
@@ -762,7 +826,26 @@ Pick one more of these and delete the rest.
 -->
 
 - [X] Metrics
-  - Metric name:
+  - Metric name: `apiserver_request_total{resource="pods",subresource="resize"}`
+    - [X] apiserver
+  - Metric name: `kubelet_container_requested_resizes_total`
+    - [X] kubelet
+    - Labels: `resource`, `requirement`, `operation`, `namespace`, `container_resize`, `pod_level_resize`
+  - Metric name: `kubelet_pod_requested_resizes_total`
+    - [X] kubelet
+    - Labels: `resource`, `requirement`, `operation`, `namespace`, `container_resize`, `pod_level_resize`
+  - Metric name: `kubelet_pod_resize_duration_seconds`
+    - [X] kubelet
+    - Labels: `namespace`, `container_resize`, `pod_level_resize`
+  - Metric name: `kubelet_pod_pending_resizes`
+    - [X] kubelet
+    - Labels: `reason`, `reason_detail`, `namespace`, `container_resize`, `pod_level_resize`
+  - Metric name: `kubelet_pod_in_progress_resizes`
+    - [X] kubelet
+    - Labels: `namespace`, `container_resize`, `pod_level_resize`
+  - Metric name: `kubelet_pod_deferred_resize_accepted_total`
+    - [X] kubelet
+    - Labels: `accepted_reason`, `namespace`, `container_resize`, `pod_level_resize`
   - `apiserver_rejected_requests` will indicate any failures (`Bad Request` code=400) related to translation of new `resources` field in PodSpec. 
   - `schedule_attempts_total{result="error|unschedulable"}`
   - `node_collector_evictions_total`: to check if a pod level resource setting is causing to evict more pods than normal
@@ -772,11 +855,7 @@ Pick one more of these and delete the rest.
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
-<!--
-Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
-implementation difficulties, etc.).
--->
-No 
+No
 
 ### Dependencies
 
@@ -958,12 +1037,16 @@ For each of them, fill in the following information by copying the below templat
   - Mitigations: Disable the feature gate or upgrade the container runtime to a
     compatible version (e.g., latest containerd/CRI-O).
   - Diagnostics: Kubelet logs (search for `UpdatePodSandboxResources` errors) and
-    `kubectl get pod <name> -o yaml` to check `resizeStatus`.
+    `kubectl get pod <name> -o yaml` to check Pod conditions (`PodResizePending`, `PodResizeInProgress`).
 - **Cgroup update failure (OS level)**:
-  - Detection: Kubelet will emit an event indicating failure to update cgroups.
+  - Detection: Kubelet will emit an event indicating failure to update cgroups. The metric `kubelet_pod_in_progress_resizes` will fail to decrement for the target pod, potentially revealing a spike in actuation duration.
   - Mitigations: Revert the resize request in the Pod spec to a known-good value.
   - Diagnostics: Kubelet logs and `dmesg` on the node for potential OOM or cgroup
     permission issues.
+- **`kubectl describe node` shows 0 for pod-level resource pods (v1.34, v1.35)**:
+  - Detection: `kubectl describe node` reports 0 (0%) CPU/memory for pods that set resources only at pod level (`spec.resources`), not at container level. Ref: [kubernetes/kubectl#1807](https://github.com/kubernetes/kubectl/issues/1807) and [kubernetes/kubernetes#137158](https://github.com/kubernetes/kubernetes/issues/137158).
+  - Mitigations: Upgrade `kubectl` to version >= v1.36 where the fix has landed. Ref: [kubernetes/kubernetes#137394](https://github.com/kubernetes/kubernetes/pull/137394).
+  - Diagnostics: Check `kubectl get pod <name> -o jsonpath='{.spec.resources}'` to verify the actual values stored in etcd.
 
 
 ###### What steps should be taken if SLOs are not being met to determine the problem?
@@ -972,15 +1055,17 @@ For each of them, fill in the following information by copying the below templat
    on all components (apiserver, scheduler, kubelet).
 2. Check `apiserver_request_total{resource="pods", subresource="resize"}` to see
    if resize requests are being rejected at the API level.
-3. Inspect Kubelet logs for errors related to `UpdatePodSandboxResources` or
+3. Use `kubelet_pod_pending_resizes` with the `container_resize` and `pod_level_resize` labels to identify if resizes are pending due to container or pod-level limitations.
+4. Inspect Kubelet logs for errors related to `UpdatePodSandboxResources` or
   `ResourceCalculation`.
-4. Monitor `node_collector_evictions_total` to ensure pod-level limits aren't
+5. Monitor `node_collector_evictions_total` to ensure pod-level limits aren't
   causing unexpected evictions.
 
 ## Implementation History
 
 - **2025-06-18:** KEP draft split from (KEP#2387)[https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/2837-pod-level-resource-spec/README.md]
 - **2026-01-28:** KEP moved to beta for 1.36 release
+- **2026-06-10:** KEP moved to stable for 1.37 release
 
 ## Drawbacks
 
@@ -1000,6 +1085,13 @@ information to express the idea and why it was not acceptable.
 -->
 
 ## Future Work
+
+**VPA Integration**
+
+SIG Autoscaling is actively working on Vertical Pod Autoscaler (VPA) support for pod-level resources (see [design doc](https://docs.google.com/document/d/1nbUNuuW3YbKkea124nyBrUjw97TS3i1d019S8YxsFYA/edit?usp=sharing)). Since these changes are non-trivial, they will require significant time to reach completion. Integration with VPA is planned but is not a GA blocker for this KEP because:
+- VPA is out-of-tree and has a separate release cycle from Kubernetes core.
+- VPA doesn’t break when pod-level resources are present. While it might invalidate a pod if container requests/limits conflict with pod-level rules, it does not prevent the core feature from functioning safely.
+- In-place resize support for VPA will be tracked separately in the VPA repository.
 
 **Ephemeral containers with pod-level resources and IPPR**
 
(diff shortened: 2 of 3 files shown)


keps/prod-readiness/sig-node/5419.yaml
```diff
@@ -2,4 +2,6 @@ kep-number: 5419
 alpha:
   approver: "@soltysh"
 beta:
-  approver: "@soltysh"
\ No newline at end of file
+  approver: "@soltysh"
+stable:
+  approver: "@soltysh"
```
jpbetz, line 7:
```suggestion
  approver: "@jpbetz"
```

I can take this one

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -954,13 +978,13 @@ For each of them, fill in the following information by copying the below templat
 
 - **CRI Runtime doesn't support Pod Sandbox Resize**:
   - Detection: `PodStatus.Resize` will be stuck in `InProgress` and Kubelet logs will
-    show errors calling `UpdatePodSandboxResources`.
+    show errors calling `UpdatePodSandboxResources`. Additionally, `kubelet_pod_in_progress_resizes` will remain elevated for the affected pods, and `kubelet_pod_resize_duration_seconds` may show high latency or timeouts.
   - Mitigations: Disable the feature gate or upgrade the container runtime to a
     compatible version (e.g., latest containerd/CRI-O).
   - Diagnostics: Kubelet logs (search for `UpdatePodSandboxResources` errors) and
     `kubectl get pod <name> -o yaml` to check `resizeStatus`.
```
Priyankasaggu11929: `PodStatus.Resize` (or `ResizeStatus`) field is now [deprecated](https://github.com/kubernetes/kubernetes/blob/1b85ff009e30bd1f42c320697a0548f5591e6e6c/staging/src/k8s.io/api/core/v1/types.go#L5440)
 
> Deprecated: Resize status is moved to two pod conditions `PodResizePending` and `PodResizeInProgress`

Could you update all references to use the new recommended pod status conditions?

another reference that needs similar updating - https://github.com/kubernetes/enhancements/pull/6184/changes#diff-f329850177ce55e0ef3ed66fac40885c7dba8466c4774b1c4bdc35b85332154eL737

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
Priyankasaggu11929, line 1032: Another recent one to add:


- **`kubectl describe node` shows `0` for pod-level resource pods (v1.34, v1.35)**:
  - Detection: `kubectl describe node` reports `0 (0%)` CPU/memory for pods that set resources only at pod level (`spec.resources`), not at container level.
    Ref: [kubernetes/kubectl#1807](https://github.com/kubernetes/kubectl/issues/1807) and [kubernetes/kubernetes#137158](https://github.com/kubernetes/kubernetes/issues/137158).
  - Mitigations: Upgrade kubectl to version >= v1.36 where the fix has landed.
    Ref: [kubernetes/kubernetes#137394](https://github.com/kubernetes/kubernetes/pull/137394)
  - Diagnostics: Check `kubectl get pod <name> -o jsonpath='{.spec.resources}'` to verify the actual values stored in etcd.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -772,11 +790,17 @@ Pick one more of these and delete the rest.
...
+- `container`: Only container resources are changing.
+- `pod`: Only pod-level resources are changing.
+- `mixed`: Both container-level and pod-level resources are changing in the same request.
+
+**Why this is better than a binary "either/or" scope:**
+- **Atomicity Visibility:** Because the KEP treats the entire request as an atomic operation, using a `mixed` value allows operators to see the aggregate success rate of complex, combined resize requests.
+- **Lower Cardinality:** We are not adding redundant metrics; we are simply adding a dimension to existing ones. This keeps monitoring overhead low and dashboards clean.
+- **Troubleshooting Path:** If a spike in infeasible resizes occurs, an operator can immediately filter by `target_type=mixed` to see if the contention is specifically caused by the interaction between pod-level and container-level resource requirements.
 
 ### Dependencies
```
Priyankasaggu11929, line 860: There's a [dependency](https://github.com/kubernetes/kubernetes/blob/master/pkg/features/kube_features.go#L2371) on the `NodeDeclaredFeatures` feature gate.

If `NodeDeclaredFeatures` is disabled, all three components (apiserver, scheduler, kubelet) will fail to start.

please update this section to call this out.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
Priyankasaggu11929, line 678: could you please update the list of e2e and unit tests that were added during alpha/beta graduations for documentation and future reference purposes?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -520,7 +519,7 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
   for Pod-level resource resize.
 * Revisit the decision of which component sets the defaults for PodStatus.Resources.
 
-#### GA (stable)
+#### GA (stable) [DONE]
 
 * VPA Integration of In-Place Resize moved to beta.
 * No major bugs reported for 3 months.
```
Priyankasaggu11929, line 586: Following k/k bugs are still open:

- https://github.com/kubernetes/kubernetes/issues/139627   (listed as [blocker for GA](https://github.com/kubernetes/kubernetes/issues/139627#issuecomment-4674977843))

- https://github.com/kubernetes/kubernetes/issues/135082
- https://github.com/kubernetes/kubernetes/issues/137628
- https://github.com/kubernetes/kubernetes/issues/138473

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -520,7 +519,7 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
   for Pod-level resource resize.
 * Revisit the decision of which component sets the defaults for PodStatus.Resources.
 
-#### GA (stable)
+#### GA (stable) [DONE]
 
 * VPA Integration of In-Place Resize moved to beta.
```
Priyankasaggu11929: Is this work tracked as part of [AEP-7571](https://github.com/kubernetes/autoscaler/pull/8586) or some other AEP?

Confirming since the AEP-7571 is not merged (and IIUC not code implemented as well yet?)

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -520,7 +519,7 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
   for Pod-level resource resize.
 * Revisit the decision of which component sets the defaults for PodStatus.Resources.
```
Priyankasaggu11929: > Revisit the decision of which component sets the defaults for PodStatus.Resources.

Could you clarify this decision before GA?


There is also a open TODO as well [in the code](https://github.com/kubernetes/kubernetes/blob/5f3b845f8f5a361455ff47adbe8b11d0de471cad/pkg/kubelet/kubelet_pods.go#L2201-L2204) and [above in KEP](https://github.com/kubernetes/enhancements/pull/6184/changes#diff-f329850177ce55e0ef3ed66fac40885c7dba8466c4774b1c4bdc35b85332154eL426)

Priyankasaggu11929 (commented):
PRR shadow review

keps/prod-readiness/sig-node/5419.yaml
soltysh: Thanks for the offer, I'm already handling this one, so let's leave it as is 😉

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
soltysh: This part deserves an update as well.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
soltysh: This is in-cluster, or even specifically in-process dependency, which will fail during startup. If you try to enable a FG which depends on a different one we will fail starting the component. This question is strictly to describe dependencies on external services. An example would be CRI or metrics server, etc. So @ndixita you can ignore the above comment for PRR.

soltysh (commented):
Several comments from Priyanka stand and there's one regarding upgrade from me as well.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: This is being removed as a GA blocker. https://docs.google.com/document/d/1nbUNuuW3YbKkea124nyBrUjw97TS3i1d019S8YxsFYA/edit?usp=sharing We discussed the plan with VPA folks and it has been agreed upon to remove this as a GA blocker as:

- VPA is out-of-tree and has a separate release cycle unlike in-tree HPA 
- VPA doesn’t break if pod-level resources exist. It could invalidate the pod if container requests/limits don’t comply with the pod-level resources requests and limits rules.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: These bugs are related to pod-level resources KEP-2837 which we plan to resolve in the development cycle

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Fixed, PTAL

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
Priyankasaggu11929: @ndixita, 

https://github.com/kubernetes/kubernetes/issues/139627 is marked as GA blocker by  @SergeyKanzhelev 

could you please confirm on this one.

Although, one thing (just for understanding sake) - how much of the above bug in the pod level
resources/status fields will affect the "in place resize" logic?

@ndixita - can we get a SIG review/approval as well on this? Thanks

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
Priyankasaggu11929: please update to reflect it in the KEP as well (to strike it from GA criteria with a link to same doc) for documentation purposes please? !

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
Priyankasaggu11929: @ndixita, i think the recent changes for addressing the review are not synced in the PR. PTAL

ndixita: sorry, pushing the changes now

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Sorry, and yes it is a GA blocker. Let me add these to the list of GA work items then, and these are all small bugs.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
soltysh: What about these tests? Seems like they cover several important areas of the functionality, are they implemented?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
soltysh: https://github.com/kubernetes/kubernetes/issues/139627 makes me very uncomfortable approving the PRR knowing we have a rather recent blocker and I don't see any WIP PR attempting to fix it. I'd love to hear from @SergeyKanzhelev before I will approve this.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -518,13 +533,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
...
   [KEP#1287](https://github.com/kubernetes/enhancements/blob/ef7e11d088086afd84d26c9249a4ca480df2d05a/keps/sig-node/1287-in-place-update-pod-resources/README.md)
   for Pod-level resource resize.
-* Revisit the decision of which component sets the defaults for PodStatus.Resources.
+* Decision finalized: `PodStatus.Resources` defaults are set by the Kubelet, not the API server, to ensure accuracy based on node-level admission and allocation. [DONE]
 
 #### GA (stable)
 
-* VPA Integration of In-Place Resize moved to beta.
 * No major bugs reported for 3 months.
 * UpdatePodSandboxResources is implemented by containerd & CRI-O
```
soltysh, line 587: I see this is done:
- https://github.com/search?q=repo%3Acontainerd%2Fcontainerd%20UpdatePodSandboxResources&type=code
- https://github.com/search?q=repo%3Acri-o%2Fcri-o%20UpdatePodSandboxResources&type=code

soltysh (commented):
The two issues that stand are:
1. the GA blocker issue
2. missing tests

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -1001,6 +1050,13 @@ information to express the idea and why it was not acceptable.
 
 ## Future Work
 
+**VPA Integration**
+
+SIG Autoscaling is actively working on Vertical Pod Autoscaler (VPA) support for pod-level resources (see [design doc](https://docs.google.com/document/d/1nbUNuuW3YbKkea124nyBrUjw97TS3i1d019S8YxsFYA/edit?usp=sharing)). Since these changes are non-trivial, they will require significant time to reach completion. Integration with VPA is planned but is not a GA blocker for this KEP because:
```
adrianmoisey, line 1091: I'd prefer if we linked to the AEP and not a google doc.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -1001,6 +1050,13 @@ information to express the idea and why it was not acceptable.
...
+- VPA is out-of-tree and has a separate release cycle from Kubernetes core.
+- VPA doesn’t break when pod-level resources are present. While it might invalidate a pod if container requests/limits conflict with pod-level rules, it does not prevent the core feature from functioning safely.
```
adrianmoisey, line 1093: > VPA doesn’t break when pod-level resources are present

Has this been tested?

keps/sig-node/5419-pod-level-resources-in-place-resize/kep.yaml
```diff
@@ -58,3 +60,11 @@ metrics:
 - node_collector_evictions_total
 - started_pods_errors_total
 - started_containers_errors_total
+- kubelet_pod_level_requested_resizes_total
+- kubelet_pod_level_resize_duration_seconds
```
natasha41575, line 64: do we really need a separate metrics for this? We could just extend container-level resize to have labels:
- has_container_resize: true/false
- has_pod_level_resize: true/false

natasha41575: never mind, I see enhancing existing metrics is proposed below. I think the file is in need of an update?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -772,11 +809,17 @@ Pick one more of these and delete the rest.
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
-<!--
-Describe the metrics themselves and the reasons why they weren't added (e.g., cost,
-implementation difficulties, etc.).
--->
-No 
+The metrics inherited from KEP-1287 will be enhanced with a `target_type` label to account for the atomic nature of Pod-level resource updates. This instrumentation strategy shifts from a simple binary scope to an encompassing one that captures unified transactions.
```
natasha41575: Can you list the exact metrics that this applies to? For example, I don't know whether this strategy makes sense for `kubelet_container_requested_resizes_total`; I imagine we likely want to introduce a new metric `kubelet_pod_requested_resizes_total` rather than enhancing `kubelet_container_requested_resizes_total` ?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -772,11 +809,17 @@ Pick one more of these and delete the rest.
...
+
+`target_type` values:
+- `container`: Only container resources are changing.
+- `pod`: Only pod-level resources are changing.
+- `mixed`: Both container-level and pod-level resources are changing in the same request.
```
natasha41575: I wonder if this is the right model. If an operator wants an alert on any request that involves pod-level resize, they would have to query both `pod` and `mixed`. 

An alternative labelling strategy would be to have two separate labels:

- `container_resize`: true/false
- `pod_level_resize`: true/false

If mixed, set both to true.

I feel this alternative is a bit cleaner but I can see merits to both. It might be good to get feedback from sig instrumentation here.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -772,11 +809,17 @@ Pick one more of these and delete the rest.
...
+
+**Why this is better than a binary "either/or" scope:**
+- **Atomicity Visibility:** Because the KEP treats the entire request as an atomic operation, using a `mixed` value allows operators to see the aggregate success rate of complex, combined resize requests.
+- **Lower Cardinality:** We are not adding redundant metrics; we are simply adding a dimension to existing ones. This keeps monitoring overhead low and dashboards clean.
+- **Troubleshooting Path:** If a spike in infeasible resizes occurs, an operator can immediately filter by `target_type=mixed` to see if the contention is specifically caused by the interaction between pod-level and container-level resource requirements.
```
natasha41575: I don't quite follow this reasoning. The feasibility checks don't really have anything to do with interaction with pod and container level resources ?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -484,6 +492,14 @@ covered by planned and implemented tests.
 
   * - [Pod Level Resources Resize](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/common/node/pod_level_resources_resize.go): [SIG Node](https://testgrid.k8s.io/sig-node-presubmits#pr-kubelet-e2e-podlevelresources-resize), [triage search](https://storage.googleapis.com/k8s-triage/index.html?ci=0&pr=1&sig=node&job=pr-kubelet-e2e-podlevelresources-resize)
 
+The following e2e scenarios were implemented and verified during Alpha and Beta:
```
natasha41575, line 542: Do we have an e2e test that covers shrinking pod-level memory limits below usage?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -953,17 +996,21 @@ For each of them, fill in the following information by copying the below templat
 -->
 
 - **CRI Runtime doesn't support Pod Sandbox Resize**:
-  - Detection: `PodStatus.Resize` will be stuck in `InProgress` and Kubelet logs will
-    show errors calling `UpdatePodSandboxResources`.
+  - Detection: `PodResizeInProgress` condition will be stuck in `True` and Kubelet logs will
+    show errors calling `UpdatePodSandboxResources`. Additionally, `kubelet_pod_in_progress_resizes` will remain elevated for the affected pods, and `kubelet_pod_resize_duration_seconds` may show high latency or timeouts.
```
natasha41575: > `kubelet_pod_resize_duration_seconds` may show high latency or timeouts

`kubelet_pod_resize_duration_seconds` tracks the latency of [doPodResizeAction](https://github.com/kubernetes/kubernetes/blob/e7c90e9b46be42a1980342991c3c87ae626c69e4/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L820). If the runtime blocks calls to `UpdatePodSandboxResources`, it would be surfaced in this metric as a high error rate through its `success=false` label. Because it returns early, the reported latency might actually be lower in this scenario.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -734,7 +752,7 @@ Recall that end users cannot usually observe component logs or access metrics.
 
 * If the Kubelet supports InPlacePodLevelResourcesVerticalScaling, it will always
   set the Resources field in Pod status.
-* The ResizeStatus in the pod status should converge to the empty value, indicating the resize has completed.
+* The `PodResizePending` and `PodResizeInProgress` pod conditions should be false, indicating the resize has completed.
```
natasha41575:
```suggestion
* The `PodResizePending` and `PodResizeInProgress` pod conditions should be cleared, indicating the resize has completed.
```

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
natasha41575: Another reason I am in favor of this alternative is that it is more extensible. I'd like to extend the existing resize metrics for [memory-backed volumes resize](https://github.com/kubernetes/kubernetes/pull/139425), and the alternative would allow me to add a third label `mem_backed_volume_resize` without ambiguity about what "mixed" means.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Reverted. PTAL

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Yes.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Reverted

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Cleaned the section and incorporated your feedback. PTAL

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -518,13 +579,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
 * Extend instrumentation from
   [KEP#1287](https://github.com/kubernetes/enhancements/blob/ef7e11d088086afd84d26c9249a4ca480df2d05a/keps/sig-node/1287-in-place-update-pod-resources/README.md)
   for Pod-level resource resize.
-* Revisit the decision of which component sets the defaults for PodStatus.Resources.
+* Decision finalized: `PodStatus.Resources` defaults are set by the Kubelet, not the API server, to ensure accuracy based on node-level admission and allocation. [DONE]
 
 #### GA (stable)
 
-* VPA Integration of In-Place Resize moved to beta.
```
adrianmoisey, line 525: sig-autoscaling lead here: this is fine, I don't see VPA as a blocked (and it shouldn't have been added here in the first place)

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
...
+  * *Description:* Tracks the total number of resize requests that the Kubelet originally marked as deferred but later accepted.
+  * *Labels:* `accepted_reason`, `namespace`, `container_resize`, `pod_level_resize`.
+
+#### Extended Metrics (From KEP-1287)
+
+These metrics are inherited from KEP-1287 and have been enhanced with new labels to capture the interaction between container-level and pod-level resizes.
+
+* **`kubelet_container_requested_resizes_total`**
+  * *Description:* Tracks the total number of container-level resize attempts.
+  * *Labels (Enhanced):* Includes original labels (`resource`, `requirement`, `operation`, `namespace`) plus new `container_resize` and `pod_level_resize` boolean labels.
```
natasha41575, line 493: I don't think this should have container_resize and pod_level_resize labels. This metric is specific to container resize requests.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
...
+
+The Kubelet will record the following metrics. These are categorized by whether they are new additions for Pod-level resources or existing metrics (from KEP-1287) that have been extended.
+
+#### New Metrics (Pod-Level Specific)
+
+These metrics are introduced to provide dedicated observability into Pod-level resize intent and state.
+
+* **`kubelet_pod_requested_resizes_total`**
+  * *Description:* Tracks the total number of Pod-level resize attempts observed by the Kubelet.
+  * *Labels:* `container_resize`, `pod_level_resize`, `resource`, `requirement`, `operation`, `namespace`.
```
natasha41575, line 473: Similarly, I don't think the container_resize and pod_level_resize labels should be added here. 

It's worth adding a description of what this metric tracks. If this metric is analogous to kubelet_container_request_resizes_total, then it means for example that a single resize request that changes all mem and cpu requests & limits will be recorded 4 times as each change is recorded separately. It would be good to include this detail in the KEP.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
...
+  * *Description:* Tracks the current count of pods that the Kubelet marks as pending.
+  * *Labels:* `reason`, `reason_detail`, `namespace`, `container_resize`, `pod_level_resize`.
+
+* **`kubelet_pod_in_progress_resizes`**
+  * *Description:* Tracks the total count of resize requests that the Kubelet marks as in progress.
+  * *Labels:* `namespace`, `container_resize`, `pod_level_resize`.
+
+* **`kubelet_pod_deferred_resize_accepted_total`**
+  * *Description:* Tracks the total number of resize requests that the Kubelet originally marked as deferred but later accepted.
+  * *Labels:* `accepted_reason`, `namespace`, `container_resize`, `pod_level_resize`.
```
natasha41575, line 485: These metrics are included in KEP-1287. Did you mean to have them in the below section?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
...
+
+These metrics are inherited from KEP-1287 and have been enhanced with new labels to capture the interaction between container-level and pod-level resizes.
+
+* **`kubelet_container_requested_resizes_total`**
+  * *Description:* Tracks the total number of container-level resize attempts.
+  * *Labels (Enhanced):* Includes original labels (`resource`, `requirement`, `operation`, `namespace`) plus new `container_resize` and `pod_level_resize` boolean labels.
+
+* **`kubelet_pod_resize_duration_seconds`**
+  * *Description:* Tracks the duration of `doPodResizeAction` for actuating resizes.
+  * *Labels (Enhanced):* Includes original labels (`namespace`) plus new `container_resize` and `pod_level_resize` boolean labels.
```
natasha41575, line 497: It looks like you are looking at the description of the metrics from KEP 1287. In implementation we made a couple of changes, including removing the `namespace` label from all metrics, changing this metric to milliseconds, and adding a success label. I would reference the official metrics documentation rather than KEP 1287 which is slightly outdated.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -456,6 +460,49 @@ KEPs.  The first change doesn’t present any user visible change, and if
...
+
+#### Standardized Boolean Labels
+
+To ensure orthogonality across all metrics, the following boolean labels are now applied to **both new and extended** metrics:
+
+* **`container_resize`**: `true` | `false`
+* **`pod_level_resize`**: `true` | `false`
```
natasha41575, line 504: nit / bikeshedding

maybe better names are `container_resources` and `pod_resources` ?

then for mem-backed volumes I could add another one something like `memory_volumes`

cc @tallclair for opinions

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
natasha41575: for container-resize, we also have some tests covering resource quota, prioritized resizes, limitRanger, and complex interaction between multiple `Deferred` resizes.

these tests live here: https://github.com/kubernetes/kubernetes/blob/master/test/e2e/node/pod_resize.go

it may be worth evaluating of any of them would benefit from including coverage for PLR resize

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -958,12 +1037,16 @@ For each of them, fill in the following information by copying the below templat
   - Mitigations: Disable the feature gate or upgrade the container runtime to a
     compatible version (e.g., latest containerd/CRI-O).
   - Diagnostics: Kubelet logs (search for `UpdatePodSandboxResources` errors) and
-    `kubectl get pod <name> -o yaml` to check `resizeStatus`.
+    `kubectl get pod <name> -o yaml` to check Pod conditions (`PodResizePending`, `PodResizeInProgress`).
 - **Cgroup update failure (OS level)**:
-  - Detection: Kubelet will emit an event indicating failure to update cgroups.
+  - Detection: Kubelet will emit an event indicating failure to update cgroups. The metric `kubelet_pod_in_progress_resizes` will fail to decrement for the target pod, potentially revealing a spike in actuation duration.
```
natasha41575, line 1042: I think it still worth it to mention that this can be detected by increases in error cases from the `kubelet_pod_resize_duration_milliseconds` metric (label is `success` with value `false` in this case)

natasha41575: iirc we also emit an event when an in-progress resize errors out, which could be another way to detect this

natasha41575 (commented):
kep.yaml still needs to be updated I think? ref https://github.com/kubernetes/enhancements/pull/6184#discussion_r3422348203

I may not be able to do another pass of this before kep freeze. If you end up requesting an extension I can take another look tomorrow, but otherwise I'm happy to defer IPPR-side approval to @tallclair.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Yes it has been tested, however if VPA results in aggregated container requests being greater than pod level requests, it results in validation failures. I can document that as well in this KEP

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: cc: @soltysh

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
SergeyKanzhelev: I am ok to fix the bug during development and I feel like this can be well tested. But the graduation criteria needs to be reworded.

That's said my bigger concern is lack of production usage reports - we just adding metrics and seeing first bug reports from testing. It will be great to find more real world usage examples.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Yes, but in case the container resizes by VPA result in container requests being greater than the pod-level request. This throws an error as the pod would be invalid.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Updated, 2 tests are not implemented. It is easy to implement the OOMKill related scenario. I need to vet if the cpu throttling scenario is possible to implement or not.

soltysh (approved):
With the remaining bits addressed
/approve
the PRR

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
adrianmoisey: Would this be addressed before merging? 
We have an open AEP that can be linked to

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -1001,6 +1086,13 @@ information to express the idea and why it was not acceptable.
 
 ## Future Work
 
+**VPA Integration**
+
+SIG Autoscaling is actively working on Vertical Pod Autoscaler (VPA) support for pod-level resources (see [design doc](https://docs.google.com/document/d/1nbUNuuW3YbKkea124nyBrUjw97TS3i1d019S8YxsFYA/edit?usp=sharing)). Since these changes are non-trivial, they will require significant time to reach completion. Integration with VPA is planned but is not a GA blocker for this KEP because:
```
adrianmoisey, line 1091:
```suggestion
SIG Autoscaling is actively working on Vertical Pod Autoscaler (VPA) support for pod-level resources (see [design doc](https://github.com/kubernetes/autoscaler/pull/8586)). Since these changes are non-trivial, they will require significant time to reach completion. Integration with VPA is planned but is not a GA blocker for this KEP because:
```

natasha41575:
@ndixita are you abandoning this PR or still planning to work on the metrics part?

a bunch of my comments are marked resolved but not addressed

natasha41575:
/hold

https://github.com/kubernetes/enhancements/pull/6184#discussion_r3423708684 and https://github.com/kubernetes/enhancements/pull/6184#issuecomment-4731379620

ndixita:
@natasha41575 @soltysh @adrianmoisey thank you for reviewing. I plan to resolve the bugs and improve coverage in 1.37 and postponing the GA to 1.38.

SergeyKanzhelev:
> @natasha41575 @soltysh @adrianmoisey thank you for reviewing. I plan to resolve the bugs and improve coverage in 1.37 and postponing the GA to 1.38.

We typically track KEP for the same stage if substantial work is happening on it. I would suggest to change the stage and merge this update to KEP. I am not sure if we need an exception for this, we can figure this part out.

soltysh:
@ndixita any plans to refresh this? I see https://github.com/kubernetes/enhancements/issues/5419 got included in 1.38 release.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Took a note of this, and will add the tests during implementation phase

ndixita:
> The two issues that stand are:
> 
> 1. the GA blocker issue
> 2. missing tests

https://github.com/kubernetes/kubernetes/pull/140481 GA blocker issue had lgtm in last cycle but missed the merge. As mentioned by Sergey, this can be merged in the implementation phase before we flip the feature gate. So we will merge it soon.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: VPA integration is WIP - https://github.com/kubernetes/autoscaler/pull/9988 and it is not a GA blocker

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: Done

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: 3/4 bugs were fixed in 1.37. Overhead double counting PR is out and already had lgtm in 1.37.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -58,10 +66,10 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
...
-  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
-- [ ] (R) Production readiness review completed
-- [ ] (R) Production readiness review approved
+- [X] (R) Graduation criteria is in place
+  - [X] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) 
+- [X] (R) Production readiness review completed
+- [X] (R) Production readiness review approved
 - [X] "Implementation History" section is up-to-date for milestone
 - [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
 - [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
```
soltysh: These two probably deserve  ✔️ as well.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -352,6 +360,122 @@ in-place resizing:
    to determine if caching is required and, if so, what caching strategy is most
    appropriate.
```
soltysh: Was the above (caching) implemented for beta or not? I believe it's worth updating that with the current state of matters.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -352,6 +360,122 @@ in-place resizing:
...
 
+#### Node Swap Support
+
+As defined in [KEP-2837 (Node Swap Support)](../2837-pod-level-resource-spec/README.md#node-swap-support), when `MemorySwap.SwapBehavior` is set to `LimitedSwap` on Linux nodes with cgroup v2 and `pod.spec.resources.requests.memory` is specified:
+
+* **Pod Cgroup `memory.swap.max`:** For an eligible `Burstable` pod (`PodMemoryRequest > 0` and `PodMemoryRequest != PodMemoryLimit`), the kubelet configures the pod's total swap budget on the Pod cgroup:
+  $$\text{PodSwapLimit} = \left( \frac{\text{PodMemoryRequest}}{\text{NodeTotalMemory}} \right) \times \text{TotalPodsSwapAvailable}$$
+* **Container Cgroup `memory.swap.max`:**
+  * Containers with neither `requests.memory` nor `limits.memory` set receive a container-level `memory.swap.max` calculated from the **unallocated pod memory request** ($\text{UnallocatedPodMemoryRequest} = \text{PodMemoryRequest} - \sum_{c} \text{ContainerMemoryRequest}_c$):
```
soltysh: Unrelated to PRR, but gh fails to render these equations, except for the first one. I'm guessing a formatting error might be the problem there

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -352,6 +360,122 @@ in-place resizing:
...
+  2. **Swap recalculation on container restart:** When `resizePolicy: RestartContainer` is specified, the resize is admitted, the container is restarted, and its swap limit (`memory.swap.max`) is recalculated from the new memory request upon recreation.
+  3. **No-op swap resizes proceed in-place:** Resizes that leave swap allocations unchanged—such as CPU-only resizes, resizes on containers with `NoSwap` (`memory.swap.max = 0`), or memory limit changes where `request != limit` continues to hold—are allowed in-place without requiring a container restart for swap.
+* **Extension to Pod-Level Resources in this KEP:** For IPPR with Pod-Level Resources, the Kubelet applies these exact same container-level policies whenever a resize to `pod.spec.resources` affects the pod's swap budget ($\text{PodSwapLimit}$) or any container's effective swap allocation.
+
```
soltysh: Was the above implemented as part of beta or it's something that's being work on for GA?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -518,13 +719,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
 * Extend instrumentation from
   [KEP#1287](https://github.com/kubernetes/enhancements/blob/ef7e11d088086afd84d26c9249a4ca480df2d05a/keps/sig-node/1287-in-place-update-pod-resources/README.md)
   for Pod-level resource resize.
-* Revisit the decision of which component sets the defaults for PodStatus.Resources.
+* Decision finalized: `PodStatus.Resources` defaults are set by the Kubelet, not the API server, to ensure accuracy based on node-level admission and allocation. [DONE]
```
soltysh: There's a question above regarding caching for beta, which I can't find information about.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -518,13 +719,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
...
 
-#### GA (stable)
+#### GA (stable) (target 1.38)
 
-* VPA Integration of In-Place Resize moved to beta.
 * No major bugs reported for 3 months.
-* UpdatePodSandboxResources is implemented by containerd & CRI-O
+* `UpdatePodSandboxResources` is implemented by containerd & CRI-O. [DONE]
+* Resolve Pod Overhead double-counting when pod-level resources are specified ([kubernetes/kubernetes#139627](https://github.com/kubernetes/kubernetes/issues/139627)).
```
soltysh: ~This was [previously mentioned](https://github.com/kubernetes/enhancements/pull/6184#discussion_r3397874840) as blocker for GA, what is the current status of it?~
Let's point to https://github.com/kubernetes/kubernetes/pull/140481 which is solving that one.

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -518,13 +719,16 @@ Support the basic functionality for kubelet to translate pod-level requests/limi
...
+* Resolve inconsistency in computing `PodQOSClass` for pod-level resources ([kubernetes/kubernetes#135082](https://github.com/kubernetes/kubernetes/issues/135082)). [DONE]
+* Clarify and resolve `Status.Resources` CPU request reporting and container request vs. limit validation with pod-level resources ([kubernetes/kubernetes#137628](https://github.com/kubernetes/kubernetes/issues/137628), [kubernetes/kubernetes#138473](https://github.com/kubernetes/kubernetes/issues/138473)).
```
soltysh: I see both of these closed, although the latter has an open PR referencing it, what's the state there?

soltysh (commented):
/approve cancel
the PRR bits, there are a few questions I'd like to get answered, and ideally try to get approval from sig-node asap

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
```diff
@@ -352,6 +360,122 @@ in-place resizing:
    to determine if caching is required and, if so, what caching strategy is most
    appropriate.
 
+#### Node Swap Support
```
soltysh: I just learned about #5359 (KEP doc in https://github.com/kubernetes/enhancements/pull/5569), will that other KEP affect this feature in any way?

keps/sig-node/5419-pod-level-resources-in-place-resize/README.md
ndixita: This is being worked on as a part of GA. container-level resize already support this

PR #6190, opened 2026-06-10
verify-node-approvers

SergeyKanzhelev:
- One-line PR description:

Keep OWNERS and kep.yaml in sync for SIG Node approvers scaling

@haircommander and @wojtek-t for the feedback on approach

Note: some AI was used

### DIFF
--- a/docs/kepval.md
+++ b/docs/kepval.md
@@ -12,3 +12,14 @@ Enhancement Proposal) is valid.
 ## Development
 
 1. Run the tests with `go test -cover ./...`
+
+## SIG Node assigned reviewers/approvers
+
+A `kep.yaml` may annotate individual entries with the inline comments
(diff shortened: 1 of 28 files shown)


haircommander:
I'm curious if pkg is the right place to put this or maybe a subrepo of hack? unless hack is just scripts...

SergeyKanzhelev:
> I'm curious if pkg is the right place to put this or maybe a subrepo of hack? unless hack is just scripts...

Other enforcements are in pkg/ as well. So looks like the right place.

> You can then trigger verification by writing `/verify-owners` in a comment.

This seems to be non-blocking. Those are test-only OWNER files. I think I will replace with other existing aliases than...

SergeyKanzhelev:
@johnbelamaric @wojtek-t what is the ownership model for these validations? Is what I put together looks OK?

johnbelamaric:
nice

/approve

PR #6193, opened 2026-06-11
KEP-4794: update Implementation History, etc

danwinship:
Updates to https://github.com/kubernetes/enhancements/issues/4974, mostly in the "Implementation History" section, but also removing one UNRESOLVED that's been resolved

### DIFF
--- a/keps/sig-network/4974-deprecate-endpoints/README.md
+++ b/keps/sig-network/4974-deprecate-endpoints/README.md
@@ -60,17 +60,17 @@ Items marked with (R) are required *prior to targeting to a milestone / release*
 
 - [X] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
 - [X] (R) KEP approvers have approved the KEP status as `implementable`
-- [ ] (R) Design details are appropriately documented
-- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
-  - [ ] e2e Tests for all Beta API Operations (endpoints)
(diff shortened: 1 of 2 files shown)


keps/sig-network/4974-deprecate-endpoints/README.md
```diff
@@ -262,10 +268,12 @@ Or perhaps the EndpointSlice controller could delete Endpoints objects
 that were more than 24 hours out of date with respect to their
 EndpointSlices?
 
-In all cases, we should probably not automatically delete Endpoints
-that don't looke like they were originally created by the Endpoints
-controller. (That is, we should not delete Endpoints unless they
-correspond to a Service with a selector.)
+In all cases, we should only automatically delete Endpoints that were
```
aojea, line 271: There is a particular situation because the Endpoints controller does not reconcile on EndpointsUpdates on the contrary to the EndpointSlice controller.  

https://github.com/kubernetes/kubernetes/blob/8e7227488643695ffedb8acc00e89d9f50928115/pkg/controller/endpoint/endpoints_controller.go#L294-L301

So I do not know if there can be some super edge case of someone taking over the endpoint and making the service dummy or something ... anyway , previous attempt with SSA here for reference https://github.com/kubernetes/kubernetes/pull/104126

danwinship: from the linked bug:

> The main problem is that there is no way to differentiate between
the endpoints generated by the controller and the ones created or
modified by the users.

That's fixed now; the controller sets a label on the ones it creates, just like the EnpdointSlice controller does.

(We had talked about using SSA/ManagedFields instead, but decided that didn't make sense, because that's more about owning _specific fields_ in an object, whereas here we own the entire object.)

PR #6153, opened 2026-06-03
KEP-5547: Update job integration to use workloadbuilder

helayoty:
- One-line PR description: Update Job integration Kep for alpha2 to include the https://github.com/kubernetes/enhancements/issues/6089 and `minCount` mutability in https://github.com/kubernetes/enhancements/issues/4671 .

- Issue link: https://github.com/kubernetes/enhancements/issues/5547

- Other comments:

PR #6194, opened 2026-06-12
KEP-4671: Change preemption behavior for basic policy in subsequent cycles

macsko:
- One-line PR description: Update Gang Scheduling KEP to always prioritize binding over preemption for basic policy.

- Issue link: #4671

- Other comments: This change seems necessary for Job integration, where Job will defer to the `Basic` policy when the gang logic is not desired. To preserve the backwards compatibility from user perspective, the `Basic` policy should be as much similar to the pod-by-pod flow as possible. Context: https://github.com/kubernetes/enhancements/pull/6153#discussion_r3396404182

### DIFF
--- a/keps/sig-scheduling/4671-gang-scheduling/README.md
+++ b/keps/sig-scheduling/4671-gang-scheduling/README.md
@@ -1208,7 +1208,9 @@ The list and configuration of plugins used by this algorithm will be the same as
 
      * If this is a subsequent scheduling attempt (i.e., some pod group members were already scheduled
        when the cycle started), but some pods were unschedulable, the scheduler will attempt
-       [Workload-aware Preemption](#workload-aware-preemption).
+       [Workload-aware Preemption](#workload-aware-preemption). In such cases, `PlacementFeasible`
+       in the `GangScheduling` plugin will return a new `PartialSuccess` status, which informs the framework
+       that the scheduler should prioritize preemption over binding for that group.
 
        * If the preemption can accommodate more pods, it will be actuated (as described below)
          and all considered pods, including the schedulable ones, will be moved back to the scheduling queue.
@@ -1219,6 +1221,14 @@ The list and configuration of plugins used by this algorithm will be the same as
        This aligns with real applications, where attempting preemption to secure capacity for as many additional,
        remaining pods as possible is preferable to binding only a schedulable subset of pods,
        if soon after that we will schedule another set of pods unblocked by preemption.
+
+       To preserve standard pod-by-pod behavior under the `Basic` policy, the scheduler will always prioritize
+       binding over preemption for that policy. As long as any incoming pod is schedulable,
+       it will proceed directly to binding - any necessary preemption will be deferred to a subsequent cycle.
+       Attempting preemption immediately would delay the binding of schedulable pods,
+       making the `Basic` policy less compatible with the pod-by-pod behavior.
+       The scheduling cycle code will interpret a `Success` status returned by the `PlacementFeasible`
+       extension point (default behavior for `Basic` policy) as an indication that binding is prioritized over preemption.
 
    * If `schedulableCount < minCount`, the cycle fails. The scheduler attempts
      [Workload-aware Preemption](#workload-aware-preemption) to free sufficient space for the `PodGroup` through disruption.


macsko:
@dom4ha @sanposhiho @wojtek-t

macsko:
/hold
For reviews

wojtek-t:
/lgtm

Thanks for adjusting this!

dom4ha:
> As long as any incoming pod is schedulable, it will proceed directly to binding 

/lgtm

soltysh (approved):
lgtm, coming from https://github.com/kubernetes/enhancements/pull/6153

keps/sig-scheduling/4671-gang-scheduling/README.md
```diff
@@ -1220,6 +1220,15 @@ The list and configuration of plugins used by this algorithm will be the same as
        remaining pods as possible is preferable to binding only a schedulable subset of pods,
        if soon after that we will schedule another set of pods unblocked by preemption.
 
+       To preserve standard pod-by-pod behavior under the `Basic` policy, the scheduler will always prioritize
+       binding over preemption for that policy. As long as any incoming pod is schedulable,
+       it will proceed directly to binding - any necessary preemption will be deferred to a subsequent cycle.
+       Attempting preemption immediately would delay the binding of schedulable pods,
+       making the `Basic` policy less compatible with the pod-by-pod behavior.
+       The scheduling cycle code will directly inspect the `PodGroup` object for the `Basic` policy
+       to trigger this behavior. This coupling is deemed acceptable for now, though a new extension point
```
sanposhiho: I don't want to just say it is acceptable for now. I mean I am actually ok for now, but I want us to discuss a clear path at the same time so that we can eventually eliminate it, if that makes sense.

This discussion feels similar to what was discussed in CPG KEP, around the greedy-mode switching. 
https://github.com/kubernetes/enhancements/pull/6017#discussion_r3324991703

sanposhiho: Like, this changed part only talks about basic policy, but what about gang that minCount is already satisfied in a earlier scheduling cycle?

macsko: > I don't want to just say it is acceptable for now. I mean I am actually ok for now, but I want us to discuss a clear path at the same time so that we can eventually eliminate it, if that makes sense.

For basic policy, one could argue that if we say in the API that is specifies the "standard Kubernetes scheduling behavior", it should be fine to access the information about this policy from the framework (which implements this standard behavior).

> This discussion feels similar to what was discussed in CPG KEP, around the greedy-mode switching.

Greedy mode switching is different and could be achievable using PlacementFeasible, by running it before scheduling any pod to check, whether the constraints are initially met or not.

However, for this preemption case that I'm adjusting in this PR, we AFAIK we cannot use any existing extension points to implement this. It's because we need to somehow know that algorithm's success for basic policy is different than algorithm's success for gang policy when `len(schedulablePods) < len(unscheduledPods)`. We would need a new extension point that would decide whether the given outcome is:

- OK - schedulable pods should go to binding (basic policy behavior)
- Acceptable, but the preemption is preferred - pods should go to WAP, but bind if the preemption won't help (gang policy behavior when `len(schedulablePods) < len(unscheduledPods)`)
- Unschedulable (or whatever) - try WAP, move back to the queue

> Like, this changed part only talks about basic policy, but what about gang that minCount is already satisfied in a earlier scheduling cycle?

It's above the change in the KEP:
https://github.com/kubernetes/enhancements/blob/82ec31498e9df304d971cc8ac702ad6d43065ca7/keps/sig-scheduling/4671-gang-scheduling/README.md?plain=1#L1209-L1221

macsko: > Greedy mode switching is different and could be achievable using PlacementFeasible, by running it before scheduling any pod to check, whether the constraints are initially met or not.

Probably, with this being implemented, we could return Unschedulable for basic policy from the PlacementFeasible called before scheduling and in this way distinguish basic and gang for the preemption behavior. However, this wouldn't be so clean that basic policy plugin would return unschedulable, even if some basic pods are already scheduled for the group, just to bypass preemption.

wojtek-t: What we can consider though is to have a new result (like `PartiallySchedulable` or whatever) that we would return in such case and the framework would be able to interpret correctly.
This seems cleaner then abusining `Unschedulable`

sanposhiho: > For basic policy, one could argue that if we say in the API that is specifies the "standard Kubernetes scheduling behavior", it should be fine to access the information about this policy from the framework

Well, what is the "standard Kubernetes scheduling behavior"? If that is something defined on the core APIs, all scheduling constraints like pod affinity etc are also them (but obviously we don't want to make framework move specifically for pod affinity)
We've been trying to decouple gang behavior from the framework with many new ext points, and it doesn't make sense to me that `Basic`  can be implemented directly on the framework side.

---

> It's above the change in the KEP:

I mean, I know it is mentioned. To clarify my question - If we want to `prioritize binding over preemption` for Basic policy, why don't we want to do the same with gang that already has running pods satisfying minCount? I am regarding the basic policy essentially as the gang policy with minCount=0. But, you are trying to make a difference between basic vs minCount=0 gang.

macsko: > I am regarding the basic policy essentially as the gang policy with minCount=0. But, you are trying to make a difference between basic vs minCount=0 gang.

This case just finally shows that `basic != gang with minCount=1`. Basic is intended to preserve as much from pod-by-pod flow as possible to allow, for example, Job, to default on basic policy when gang behavior is not desired (i.e., user wants older flow). For gangs, however, we can apply policies that we think are more desired. In particular, when adding more pods to a gang with satisfied `minCount`, we can prioritize executing preemption to fit more pods, because we already satisfied the policy and doing a single preemption path can minimize potential disruptions, and as a result bind more pods in a batch.

macsko: > What we can consider though is to have a new result (like `PartiallySchedulable` or whatever) that we would return in such case and the framework would be able to interpret correctly. This seems cleaner then abusining `Unschedulable`

With your comment, I see that we could just introduce this new status to the PlacementFeasible (let's say `PartialSuccess`) and return it from the gang scheduling plugin as long as not all pods from a gang are scheduled. For basic, it would still always default to `Success`. Then, we could execute preemption when the status is a `PartialSuccess`, but move to binding when it's a `Success`. If i'm not missing something, it should be fairly easy to implement.

wojtek-t: Yes - it's simpler than I thought. Let's just do that from the very beginning.

macsko: Updated the KEP to reflect the new idea

wojtek-t: > For gangs, however, we can apply policies that we think are more desired. In particular, when adding more pods to a gang with satisfied minCount, we can prioritize executing preemption to fit more pods, because we already satisfied the policy and doing a single preemption path can minimize potential disruptions, and as a result bind more pods in a batch.

+1 - as often adding more pods to gang causes disruption (e.g. restart) of running pods.

wojtek-t:
/lgtm

It's much cleaner now.

sanposhiho:
/lgtm
/approve

lookin great now

petr-muller:

PR #6197, opened 2026-06-12
Update sig-node KEP approvers/reviewers from project board

SergeyKanzhelev:
- One-line PR description:

Add missing sig-node-assigned-approver and sig-node-assigned-reviewer annotations to kep.yaml files based on the SIG Node project board. Remove duplicate entries.

See triaged: https://github.com/orgs/kubernetes/projects/265/

/sig node
/assign @mrunalp @dchen1107

PR #6198, opened 2026-06-12
[WIP] Add hack/sync-sig-node-with-project.sh to sync project with keps

SergeyKanzhelev:
Fetches issues from a GitHub Projects v2 board, finds matching kep.yaml files under keps/sig-node/, and compares approvers/reviewers annotated with sig-node-assigned comments. Supports --update-keps to add missing handles to kep.yaml files.

Special note: AI was used

/sig node

I do not know if hack/ is the best place for this script or we should keep it someplace else. Community repo maybe?

@haircommander 

This script was used to fix up approvers from the project: https://github.com/kubernetes/enhancements/pull/6197

### DIFF
--- a/hack/sync-sig-node-with-project.sh
+++ b/hack/sync-sig-node-with-project.sh
@@ -1,0 +1,425 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


haircommander:
I haven't reviewed the bash thoroughly yet, but one note is I vibecoded a version of this in python which may be easier to maintain over time https://github.com/haircommander/sig-node-tools though it has some bugs I haven't worked through yet (many that you revealed in your validation package work)

PR #6201, opened 2026-06-15
Ensure SIG Node apporvers

SergeyKanzhelev:
- One-line PR description:

/kind cleanup
/sig node
/assign @haircommander 

Some AI helped

### DIFF
--- a/test/node_approvers_test.go
+++ b/test/node_approvers_test.go
@@ -52,3 +52,44 @@ func TestNodeApprovers(t *testing.T) {
 		)
 	}
 }
+
+// TestNodeTechLeadApprovers verifies that every KEP under keps/sig-node lists an
+// acceptable approver per the stage-dependent rules: alpha-stage KEPs must list a
+// sig-node-tech-leads member (and must not use the # sig-node-assigned-approver
+// marker), while non-alpha KEPs must list a sig-node-tech-leads member or an
+// approver marked # sig-node-assigned-approver. It runs over the real
+// keps/sig-node tree using the repo-root OWNERS_ALIASES and is exercised
+// automatically in CI via hack/test-go.sh.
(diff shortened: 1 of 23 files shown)


hack/verify-node-approvers.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright The Kubernetes Authors.
+# Copyright 2026 The Kubernetes Authors.
```
johnbelamaric: nit: best practice is now to leave the year off

test/node_approvers_test.go
```diff
@@ -52,3 +52,44 @@ func TestNodeApprovers(t *testing.T) {
 		)
 	}
 }
+
+// TestNodeTechLeadApprovers verifies that every KEP under keps/sig-node lists an
+// acceptable approver per the stage-dependent rules: alpha-stage KEPs must list a
+// sig-node-tech-leads member (and must not use the # sig-node-assigned-approver
+// marker), while non-alpha KEPs must list a sig-node-tech-leads member or an
+// approver marked # sig-node-assigned-approver. It runs over the real
+// keps/sig-node tree using the repo-root OWNERS_ALIASES and is exercised
```
johnbelamaric, line 61: Is it using OWNERS_ALIASES from the **PR** or from the committed repo?

johnbelamaric: I guess it doesn't really matter?

SergeyKanzhelev: from the PR. But I agree, it should not matter. It will be a highly exceptional case when it will become a problem

hack/verify-node-approvers.sh
SergeyKanzhelev: fixed and enabled boilerplate validation

PR #6204, opened 2026-06-17
Mark lifecycle conditions as implementable

rthallisey:
The discussions had around https://github.com/kubernetes/enhancements/pull/6161 were with the assumption that it is `implementable`.  It was oversight on my part to mark it incorrectly so addressing that now.


mrunalp (approved):
/lgtm

PR #6205, opened 2026-06-18
KEP-6035: Update exec session identity propagation for API review

nispriha:
- One-line PR description:  Update implementation history, submitting for API review.

- Issue link: https://github.com/kubernetes/enhancements/issues/6035

- Other comments: The original KEP PR #6036 was merged before API review. Opening this follow-up to trigger the review.

### DIFF
--- a/keps/sig-node/6035-exec-session-identity/README.md
+++ b/keps/sig-node/6035-exec-session-identity/README.md
@@ -514,6 +514,7 @@ For each of them, fill in the following information by copying the below templat
 - 2026-05-18: KEP-6090 created (CRI exec environment variable injection primitive)
 - 2026-05-19: Proposal presented at SIG Node meeting ([recording](https://www.youtube.com/watch?v=bl8OX2nlEfM&t=900s), discussion starts at 15:00)
 - 2026-06-15: KEP-6035 and KEP-6090 merged into this KEP per reviewer feedback; CRI primitive and audit ID propagation are presented as a single coherent proposal
+- 2026-06-18: Submitted for API review
 
 ## Drawbacks
 


keps/sig-node/6035-exec-session-identity/README.md
liggitt, line 120: > The audit ID (a UUID such as `f4a3b2c1-...`) is a per-request identifier assigned by the API server

FYI, this is assigned *if absent*, but can also be set via header on the incoming request

xref https://github.com/kubernetes/kubernetes/issues/127801 and https://github.com/kubernetes/kubernetes/issues/101597

That may not be relevant to this or change this KEP, but I wanted to mention it

liggitt, line 134: Can't the kube-apiserver audit logs be configured to contain pod exec options to give this visibility?

liggitt, line 138: This is a *really* buried lede ... I didn't expect a "let exec requests be correlated with audit id" feature to propose giving end users full control over exec env ... 

Do we _really_ have to propagate the audit ID into a process-visible env var to get audit logging?

Ideally, there would be an audit or trace thread making it possible to correlate *every* request from kube-apiserver → kubelet → cri-api. I think we already do things to make this chain possible from kube-apiserver to webhooks and etcd.

I can see how we could ensure all the `kube-apiserver → kubelet` requests could propagate audit ID as a header. On the kubelet side, I can see how we could propagate audit ID into grpc attributes or otel trace attributes. On the CRI side, expecting an audit or trace ID and correlating that with an exec session ID that was observable would be needed. @dashpole might have insight about chaining these things together.

containerd already appears to generate and log an exec ID (https://github.com/containerd/containerd/blob/b6aca07dd50288a598baadab38dcc8656c5e5327/internal/cri/server/container_execsync.go#L157-L158). It looks like that exec ID is propagated down to runc as the process ID. Is that already persisted somewhere that is observable / correlatable with the process? It seems like it must be if containerd / runc can resume streaming on restart. @samuelkarp might have insight on the wiring here and what ends up persisted / observable... I went a little cross-eyed trying to follow the exec spawning abstractions.

SergeyKanzhelev: > Ideally, there would be an audit or trace thread making it possible to correlate _every_ request from kube-apiserver → kubelet → cri-api. I

My understanding is that the goal is to make correlation available inside the executable, not in cri-api. Even if we decided to propagate using trace context or a baggage, those audit id will end up in env variables. And from there on it will be more and more requests to put something there. So it is easier to make it explicit rather than hidden contract.

This KEP started as CRI API-only proposal to allow env vars in exec calls, Audit ID is just one use case. I bet there will be way more.

liggitt, line 118: This gives full control over the exec env to the end user (which allows all sorts of surprising malarkey when execing into a container that was thought to be pretty locked down by not including bash/env utils)

It also overlays a *particular* env key for use with official auditing/tracing, in ways that user-specified envs and control-plane-defined envs are indistinguishable from each other when they get to the kubelet or cri... smushing those together seems risky.

samuelkarp: > This KEP started as CRI API-only proposal to allow env vars in exec calls

Is the current goal for arbitrary env vars or for specific audit values?

liggitt: cc @kubernetes/sig-auth-api-reviews @tallclair  for intersection with audit surface area (https://github.com/kubernetes/community/blob/main/sig-auth/README.md#audit-logging)

SergeyKanzhelev: I was thinking of scenarios where some secret would be passed using the env var like this. So the value of that env variable will not show up in a regular container logs - I made comment about it somewhere. There may be other use cases to use arbitrary environment variables.

Is the control over env vars so much worse than the control over the command line? Are you worrying about the ability to audit what was called easier by simply looking that a specific executable was called and there were no override of it using env vars?

The alternative is to prefix env vars with something to make it impossible to abuse. But if we go the prefix route, making it not a generic env vars, but rather some ambient context propagation, we may be better of using [baggage](https://opentelemetry.io/docs/concepts/signals/baggage/#what-should-otel-baggage-be-used-for) that [renders](https://opentelemetry.io/docs/specs/otel/context/env-carriers/) into the [prefixed](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/30983a333b775817cf8ccb5db5207ab4b0a5b526/propagators/ot/ot_propagator.go#L23) env vars.

SergeyKanzhelev: The Audit ID scenarios is a key driver, the idea was to allow arbitrary env vars. I commented here: https://github.com/kubernetes/enhancements/pull/6205#discussion_r3437515382 that if arbitrary env vars is not the right way to go, we may be better off using baggage

sigtrapd: That would work in an `ExecSyncRequest` where the only command executed inside the pod would be the one present in `PodExecOptions`. But in a typical Exec session (`ExecRequest`) , where the streaming endpoints of the kubelet come into play, the connection gets upgraded to a web socket/spdy and whatever commands are executed inside the session are not within the visibility of the audit mechanisms in the API server. The runtime would be the only place where this is visible reliably.

sigtrapd: There are two questions here worth addressing separately.

On whether the existing containerd exec ID chain is sufficient:
- The control plane and the runtime operate at different observability planes. The API server audit log captures what was requested. The runtime is the only place where what was actually executed is visible, and these require different correlation mechanisms.
- The containerd exec ID is persisted in runc's state directory on disk. A security agent hooking execve observes a PID and an argv. Mapping that PID back to the exec ID requires a privileged real-time lookup against containerd's internal state, synchronized to the exact moment the process execs. The env var eliminates that coupling, making the correlation identifier visible at the syscall boundary itself, readable from mm->env_start. An OTel/gRPC attribute chain covers the session level at the CRI layer. It does not cross the kernel boundary.
- Also, the file linked (container_execsync.go) is the ExecSync path. Interactive kubectl exec uses the streaming Exec path with a different handler.

On arbitrary env vars:
- The audit ID is control-plane-initiated. The kubectl exec --env mechanism is a separate consumer of the same CRI primitive. Legitimate use cases beyond audit correlation include injecting identity-aware policy context readable by an eBPF agent at execve time to enforce per-user runtime policies, exec'ing into distroless containers with no shell or env utility, and propagating trace IDs into short-lived processes. User-supplied values can be subject to the same RBAC controls that govern exec access itself.

sigtrapd: @liggitt The concern about distinguishability is valid. But this field does not introduce a new attack surface. A user with exec access can already inject arbitrary env vars today by prepending `env KEY=VALUE` to the exec command. This just provides a structured, runtime-agnostic way to do the same thing.

On the smushing concern, the API server enforces a hard override for `KUBERNETES_EXEC_AUDIT_ID` regardless of what the user supplies, as defined in the precedence order section of this KEP. For the broader question of distinguishing user-supplied from control-plane-supplied vars at the kubelet/CRI layer, env vars are flat by nature and there is no in-band way to tag their origin. A naming prefix is the only viable signaling mechanism here. Validation that rejects user-supplied vars with the `KUBERNETES_ ` prefix is something we can look at as a follow-on if the concern warrants it.

On the baggage alternative, it would still render into prefixed env vars at the process boundary through the OTel carrier. The outcome is the same, just with additional indirection through the OTel propagation machinery that every consumer of this would need to understand.

SergeyKanzhelev: > Validation that rejects user-supplied vars with the KUBERNETES_  prefix is a reasonable follow-on hardening step if the concern warrants it.
> ...
> The end state is the same, with additional indirection through the OTel propagation machinery

The baggage is a good alternative if we are removing the feature of generic environment variable and limiting this to infra-supplied values. In this case, baggage may be a better alternative.

I see value in explicit API for env vars, without making them prefixed. The audit of exec calls will be harder, but not that much harder.

sigtrapd:
**BLUF:** Interactive `kubectl exec` sessions are invisible to audit logs and OTel tracing at the kernel boundary; an env var injected by the API server (`KUBERNETES_EXEC_AUDIT_ID`) is the only viable cross-boundary correlation mechanism. The main unresolved concern — user-supplied vars being indistinguishable from control-plane vars — is addressed by rejecting `KUBERNETES_*` entries in `PodExecOptions.Env` at API server validation (Option A). The `nodes/proxy` bypass cannot be closed without identity-based kubelet enforcement (out of scope); it is documented as a known trust boundary. Seeking consensus on this direction to unblock the 1.38 alpha graduation requirement.

Thanks for the thorough review. Summarising the open concerns and where we've landed after working through them.

**On whether audit log configuration already provides the needed visibility (@liggitt)**

This only works for `ExecSyncRequest` (lifecycle hooks and probes), where the single command is visible in `PodExecOptions`. For interactive `kubectl exec`, the connection upgrades to SPDY/WebSocket and whatever commands are typed in the session are never sent back through the API server. The audit log records only the initial `connect` event. The runtime is the only place where in-session process activity is observable, and the correlation gap exists at that boundary.

**On whether env vars at the process level are necessary vs. OTel/baggage/trace attributes (@liggitt)**

An OTel/gRPC attribute chain can propagate the audit ID from the API server through to the CRI layer. It does not cross the kernel boundary. Runtime security agents (Falco, Tetragon) hook `execve`/`sched_process_exec` and observe the new process's environment at the syscall boundary via `mm->env_start`. They do not read gRPC metadata. There is no mechanism to surface a gRPC attribute into the process environment without the runtime writing it there explicitly, which is exactly what `ExecRequest.envs` provides. The env var is the cross-kernel-boundary mechanism; OTel is orthogonal to it.

On the containerd exec ID: the linked file (`container_execsync.go`) is the `ExecSync` path, not the interactive exec path. The interactive exec path generates a separate exec ID persisted in runc's state directory on disk. Mapping a PID observed at `execve` time back to that exec ID requires a privileged real-time lookup against containerd's internal state at the exact moment of the syscall. The env var eliminates that coupling. The correlation identifier is readable directly from the process's environment at the syscall boundary.

**On giving end users full control over exec env, and user-supplied vars being indistinguishable from control-plane vars (@liggitt, @SergeyKanzhelev)**

This is the most substantive concern. We want to address it concretely.

A previous reply argued that `PodExecOptions.Env` does not introduce a new capability surface. A user with `pods/exec` permission can already inject arbitrary env vars today by prepending `env KEY=VALUE` to the exec command, so this field changes the mechanism, not the capability ceiling. That argument stands: no new attack surface is introduced by formalising something that was already possible. What it leaves unaddressed is the "smushing" concern. User-specified and control-plane-defined env vars arrive at the kubelet and CRI through the same channel with no structural way to tell them apart. A caller could supply `KUBERNETES_EXEC_AUDIT_ID=fake` in `PodExecOptions.Env` and it would be indistinguishable from the value the API server injected, at least from the kubelet's perspective.

We considered three approaches to resolve this:

**Option A: Enforce the `KUBERNETES_` prefix restriction at the API server.** API server validation rejects any `PodExecOptions.Env` entry whose `Name` starts with `KUBERNETES_`. This covers the standard `pods/exec` path, is auditable via admission controllers, and makes the naming convention enforceable at the entry point where user input is validated. Its limitation is that there is a direct-to-kubelet path via `nodes/proxy` that bypasses API server validation entirely. A caller with `nodes/proxy` permission can craft arbitrary query params including `?env=KUBERNETES_EXEC_AUDIT_ID=fake` directly in the URL, and this reaches the kubelet without going through API server validation.

**Option B: Enforce the `KUBERNETES_` prefix restriction in the kubelet.** The kubelet strips any `?env=KUBERNETES_*` param from the incoming request before populating `ExecRequest.envs`. This would catch the `nodes/proxy` path too. However, the API server's own legitimate injection of `KUBERNETES_EXEC_AUDIT_ID` also travels as a `?env=` query param on the kubelet request. The kubelet cannot distinguish a param injected by the API server from one injected by a `nodes/proxy` caller, because `UpgradeAwareHandler` forwards the client's raw query string verbatim to the kubelet. Kubelet-side prefix restriction would therefore also block the API server's own injection.

**Option C: Identity-based enforcement in the kubelet.** Accept `KUBERNETES_*` params only from requests authenticated as the API server's identity. This resolves the ambiguity but is operationally fragile and out of scope here.

Since the `nodes/proxy` gap is not solvable without Option C, the gap exists regardless of where the prefix restriction is placed. **We are going with Option A.** The `nodes/proxy` bypass is documented as a known trust boundary in the Risks section. This is consistent with how it already operates for every other exec parameter today, since a `nodes/proxy` caller can already craft arbitrary `?command=` params. The feature's tamper-proofing scope remains explicitly deferred, as stated in the KEP.

The KEP update will add the `KUBERNETES_` prefix restriction as an alpha graduation requirement in the validation section, and extend the Risks section to document the `nodes/proxy` trust boundary explicitly.

We would like to reach consensus on this direction before finalising the implementation for 1.38. Happy to discuss further here or in a SIG Node / SIG API Machinery slot if that would help move things forward.

CC: @haircommander @nispriha @ngopalak-redhat @SergeyKanzhelev

keps/sig-node/6035-exec-session-identity/README.md
dashpole: https://github.com/kubernetes/enhancements/pull/5916 was meant to describe how a trace context is propagated from the APIServer's request into controller logic, which would include the kubelet, or downstream plugins like CRI, device plugins, etc. But the author seems to have abandoned the proposal.

wendy-ha18:
@nispriha just a very minor note that as you still work on this KEP this release cycle, could you please help to update `kep.yaml` from `latest-milestone: v1.37` to ``latest-milestone: v1.38`. It will help you meet deadline for KEP readiness [today](https://github.com/kubernetes/enhancements/issues/6035#issuecomment-5749500093), no need to have PR merge :pray:

sigtrapd:
@wendy-ha18 
I have made the changes. Please let me know if there's anything else needed for the readiness freeze.

wendy-ha18:
Thank @sigtrapd for the quick update, this KEP has met the requirement for `KEP readiness` deadline, you can target the next deadline for Enhancement Freeze in [here](https://github.com/kubernetes/enhancements/issues/6035#issuecomment-5749500093). Let's me know if you need any helps!

PR #6217, opened 2026-06-30
KEP-6063: Update minimum per-pod PID limit from 1024 to 128

BhargaviGudi:
## Summary

Lowers the minimum allowed value for `spec.resources.limits.pid` from 1024 to 128, based on
empirical PID usage measurements on an OpenShift cluster.

KEP: https://github.com/kubernetes/enhancements/pull/6064
Discussion regrading minimum PID limit: https://github.com/kubernetes/enhancements/pull/6064#discussion_r3398676323 

## Justification

Created a multi-container pod (`multi-container-pid-test` in namespace `pid-measure-test`) with 7 containers: 3× busybox, nginx, alpine, redis, and curl.  The pod was configured with exec liveness/readiness probes on all containers and PreStop hooks that spawn background processes. All seven containers ran successfully.

Total PIDs used by the pod were measured from the pod cgroup on the worker node:

## PID Usage Testing Results

  Tested multi-container pods with exec probes and PreStop hooks on OpenShift cluster.

  ### Test Configuration
  - **multi-container-pid-test**: 7 containers with exec liveness/readiness probes + PreStop hooks
  - **baseline-multi-container**: Same 7 containers, no probes (baseline)
  - **intensive-exec-probes**: Single container with aggressive exec probes (5s intervals)

  ### Results (cgroup measurements)

  | Pod Configuration | Containers | pids.current | pids.peak |
  |------------------|-----------|--------------|-----------|
  | Multi-container + exec probes | 7 | 36 | **47** |
  | Baseline (no probes) | 7 | 26 | **30** |
  | Intensive exec probes | 1 | 12 | **20** |

  ### Exec Probe/Hook Overhead
  - Baseline: 30 PIDs peak
  - With exec probes/hooks: 47 PIDs peak
  - **Overhead: ~17 PIDs**
  
  ### Pod-level cgroup (authoritative):
```
  pids.current: 36
  pids.peak: 47
  pids.max: 4096
```
A minimum of 128 provides adequate headroom: ~2.7× over the observed maximum (47 PIDs peak) for a complex multi-container pod with exec probes and PreStop hooks, and ~5× for baseline multi-container workloads (25-30 PIDs). This is sufficient for any standard workload while still allowing meaningful restriction of PID-intensive or untrusted processes.

### DIFF
--- a/keps/sig-node/6063-pod-pid-limit/README.md
+++ b/keps/sig-node/6063-pod-pid-limit/README.md
@@ -142,9 +142,8 @@ spec:
 
 
 ### Valid Values
-Following the same constraints as node-level `podPidsLimit`:
 
- - **Minimum**: 1024 (conservative lower bound)
+ - **Minimum**: 128 (based on measured PID usage of multi-container pods with exec probes and PreStop hooks, observed peak 47 PIDs)
  - **Maximum**: 16384 (maximum supported in managed environments)
  - **Default**: If not specified, inherits kubelet's `podPidsLimit`
 
@@ -232,7 +231,7 @@ Node Declared Features: This feature integrates with the Node Declared Features 
 
 2. **Interaction with existing PID exhaustion protections**
    - Risk: A Pod could set a very low PID limit, making itself non-functional.
-   - Mitigation: API validation enforces a minimum of 1024 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (1024), the pod has sufficient PIDs to function.
+   - Mitigation: API validation enforces a minimum of 128 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (128), the pod has sufficient PIDs to function.
 
 3. **Version skew: apiserver accepts but kubelet ignores**
    - Risk: If only apiserver enables the gate, users think their limit is enforced but kubelet applies the node default.
@@ -257,7 +256,7 @@ When the feature gate is disabled:
 #### API Validation
 Add `pid` to the list of valid resource types
 
-Validate that `pid` values are integers within the allowed range (1024-16384)
+Validate that `pid` values are integers within the allowed range (128-16384)
 
 #### Kubelet Implementation
 Parse Pod specification: The kubelet reads `spec.resources.limits.pid` from the Pod spec during Pod admission.
@@ -267,7 +266,7 @@ Limit enforcement: When creating the Pod cgroup, the kubelet sets effective PID 
 - If Pod `spec.resources.limits.pid` is set: `min(podPidsLimit, pod.spec.resources.limits.pid)`
 - If Pod `spec.resources.limits.pid` is not set: `podPidsLimit` (current behavior)
 
-Validation: The kubelet validates that the PID limit is within acceptable bounds (1024 to 16384).
+Validation: The kubelet validates that the PID limit is within acceptable bounds (128 to 16384).
 
 cgroupsv2 requirement: The kubelet checks the cgroup version and only applies Pod-level limits on cgroupsv2 systems. On cgroupsv1 systems, the kubelet rejects pods that specify `spec.resources.limits.pid` during admission.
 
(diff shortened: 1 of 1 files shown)


SergeyKanzhelev:
> with 7 containers: 3× busybox (sleep), nginx, alpine (sleep), redis, and curl (sleep).

Any exec probes or exec pre-stop hooks? exec into container takes like surprisingly a lot of PIDs.

BhargaviGudi:
> > with 7 containers: 3× busybox (sleep), nginx, alpine (sleep), redis, and curl (sleep).
> 
> Any exec probes or exec pre-stop hooks? exec into container takes like surprisingly a lot of PIDs.

Tested with exec probes and PreStop hooks - added details to description.
Peak usage with 7 containers + exec probes + hooks: **47 PIDs** (2.7× below 128 minimum).
Exec overhead: ~17 PIDs compared to baseline.

SergeyKanzhelev:
> Exec overhead: ~17 PIDs compared to baseline.

17 is not divisible by 7. Was it perhaps 3 PIDs per probe as an absolute worst case than? If so, `7 containers x 2 probe types (readiness and liveness) x 3 processes per probe = 42`. This is still lower than 100, so sounds good. Thank you for checking.

/lgtm 
/assign @mrunalp

BhargaviGudi:
> 17 is not divisible by 7. Was it perhaps 3 PIDs per probe as an absolute worst case than? If so, `7 containers x 2 probe types (readiness and liveness) x 3 processes per probe = 42`. This is still lower than 100, so sounds good. Thank you for checking.

The 17 is the measured overhead (47 peak - 30 baseline), not a per-container calculation.
I ran additional testing with 7 identical containers (all busybox), each with 2 exec probes (liveness + readiness):
Per-container: Each has 5 PIDs
  Pod total (cgroup):
  - pids.current: 14 PIDs (steady state)
  - pids.peak: 32 PIDs (when probes execute)

Analysis:
  - Each container: ~2 PIDs base (sleep process + container runtime)
  - During probe execution: adds ~3-4 PIDs per probe
  - 7 containers × 2 probes × ~2 PIDs per probe = ~28 PIDs
  - Actual peak: 32 PIDs - Very close!  - lower than 100

SergeyKanzhelev:
yep, we are on the same page.

BhargaviGudi:
@haircommander @kannon92 @mrunalp Could you please help me to review this PR? Thanks

BhargaviGudi:
@mrunalp Could you please help to review this PR? Thanks

BhargaviGudi:
The implementation PR (kubernetes/kubernetes#139277) has already been updated to use 128 as the minimum PID limit — all constants, validation, comments, and test cases reflect this change.

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -142,9 +142,8 @@ spec:
 
 
 ### Valid Values
-Following the same constraints as node-level `podPidsLimit`:
 
- - **Minimum**: 1024 (conservative lower bound)
+ - **Minimum**: 128
```
harche: Could we carry a short version of the justification into the doc itself? The measurement data lives only in this PR description, and future readers of the KEP will see a bare 128 with no basis.
```suggestion
 - **Minimum**: 128 (based on measured PID usage of multi-container pods with exec probes and PreStop hooks, observed peak 47 PIDs)
```

keps/sig-node/6063-pod-pid-limit/README.md
```diff
@@ -232,7 +231,7 @@ Node Declared Features: This feature integrates with the Node Declared Features
 
 2. **Interaction with existing PID exhaustion protections**
    - Risk: A Pod could set a very low PID limit, making itself non-functional.
-   - Mitigation: API validation enforces a minimum of 1024 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (1024), the pod has sufficient PIDs to function.
+   - Mitigation: API validation enforces a minimum of 128 and a maximum of 16384, so extremely low or high values are rejected. Even at the minimum (128), the pod has sufficient PIDs to function.
```
harche: One caveat worth capturing here: the pids controller counts every thread, not just processes. From the kernel cgroup v2 documentation, "Note that PIDs used in this controller refer to TIDs, process IDs as used by the kernel" (https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#pid), and thread creation via clone() fails with EAGAIN at the limit. The measured workloads here are thread light, while common runtimes are not: a default Tomcat connector alone allocates up to 200 request processing threads (https://tomcat.apache.org/tomcat-10.1-doc/config/http.html) and the Go runtime budgets up to 10000 OS threads by default (https://pkg.go.dev/runtime/debug#SetMaxThreads). So 128 is sufficient for typical workloads rather than universally sufficient, and it may be worth saying so.
```suggestion
   - Mitigation: API validation enforces a minimum of 128 and a maximum of 16384, so extremely low or high values are rejected. The minimum (128) provides sufficient PIDs for typical workloads. Note that the pids cgroup controller counts threads, so thread-heavy runtimes (for example a default Tomcat connector alone can create up to 200 request threads) should set a higher limit.
```

harche (commented):
Verified against the implementation: the new bounds match both enforcement points, API validation (https://github.com/kubernetes/kubernetes/blob/f050421a6693073df402fde0cc9f3568087c2d1e/pkg/apis/core/validation/validation.go#L8279-L8280) and kubelet admission (https://github.com/kubernetes/kubernetes/blob/f050421a6693073df402fde0cc9f3568087c2d1e/pkg/kubelet/lifecycle/features_linux.go#L50-L51), and all five occurrences of 1024 in the doc are covered.

One sequencing note: this overlaps with #6258, both PRs delete the same Valid Values intro line and several lines touched here get the `pid` to `pids` rename there. Whichever merges second needs a rebase, and if this one goes second its reapplied lines should pick up the `pids` spelling.

keps/sig-node/6063-pod-pid-limit/README.md
mrunalp: I think we can add this to KEP as well as documentation but still allow a lower value like 128 as it is the lower bound and not a default.

mrunalp (approved):

keps/sig-node/6063-pod-pid-limit/README.md
BhargaviGudi: Reverted this change

PR #6221, opened 2026-07-09
KEP-4322: Require ClusterProperty resources in ClusterProfile status properties

kahirokunn:
- One-line PR description: Require ClusterProperty resources to be reflected in ClusterProfile
`.status.properties` with their names and values preserved

- Issue link: https://github.com/kubernetes/enhancements/issues/4322

- Other comments:
There are already multiple cluster managers implementing the ClusterInventory API, each with its own
rules for populating properties. This creates a risk that ClusterProfile API consumers cannot be reused
across cluster managers even though they should be. To keep consumers portable regardless of which
cluster manager is used, this PR changes the wording so that ClusterProperty resources MUST be reflected
in the ClusterProfile's `.status.properties`, preserving their names and values as-is. Cluster managers
can still add their own custom-named properties as long as they do no

### DIFF
--- a/keps/sig-multicluster/4322-cluster-inventory/README.md
+++ b/keps/sig-multicluster/4322-cluster-inventory/README.md
@@ -604,11 +604,13 @@ minimum kubelet version, maximum kubelet version, and enabled featureset version
 
 #### Properties
 
-Name/value pairs to represent properties of the clusters. It could be a
-collection of ClusterProperty resources, but could also be info based on
-other implementations. The name of the cluster property can be predefined
-name from ClusterProperty resources and is allowed to be customized by
-different cluster managers.
+Name/value pairs to represent properties of the clusters. Cluster managers
+SHOULD include ClusterProperty resources defined on the member cluster in the
+ClusterProfile's `.status.properties`. They MAY also include properties from
+other implementations. Properties derived from ClusterProperty resources MUST
+preserve the name and value of the ClusterProperty as-is. Cluster managers MAY
+additionally include their own custom-named properties, as long as they do not
+conflict with names derived from ClusterProperty resources.
 
 #### Conditions
 


keps/sig-multicluster/4322-cluster-inventory/README.md
```diff
@@ -604,11 +604,13 @@ minimum kubelet version, maximum kubelet version, and enabled featureset version
 
 #### Properties
 
-Name/value pairs to represent properties of the clusters. It could be a
-collection of ClusterProperty resources, but could also be info based on
-other implementations. The name of the cluster property can be predefined
-name from ClusterProperty resources and is allowed to be customized by
-different cluster managers.
+Name/value pairs to represent properties of the clusters. ClusterProperty
+resources defined on the member cluster MUST be included in the
```
mikeshng: I am ok with this but it does goes against what what said in https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75#issuecomment-4917369795 regarding hiding and exposing data. 

Do we have to say MUST? Can we go with strongly recommended instead to keep things flexible a bit while still having the benefit of producer and consumer expecting a certain behavior?

kahirokunn: I thought that it might also be worth considering logic such as not propagating clusterproperties that match specific domain keywords or regular expressions, so I changed it to "SHOULD."

mikeshng (approved):
/lgtm

Thank you for making the adjustments based on the comment discussion!

kahirokunn:
Thank you too 😊

mikeshng (commented):
CC @skitt @lauralorenz @JeremyOT 

I feel like this is a non controversial update to the ClusterProfile KEP. If needed, we will also bring this up to the SIG-MC meeting to get further community feedback. Thanks!

skitt:
This is indeed very sensible, thanks @kahirokunn!

/approve

PR #6222, opened 2026-07-09
KEP-5710: Update with PodGroupPreemptionPolicy feature gate

Argh4k:
- One-line PR description:

Update KEP with the description of PodGroupPreemptionPolicy to match the implementation.

- Issue link: https://github.com/kubernetes/enhancements/issues/5710

- Other comments:
This is something that we missed in Alpha implementation and because of that the field will lag one release cycle behind, requiring a separate feature gate to not slow down the graduation of GenericWorkload feature gate.

### DIFF
--- a/keps/sig-scheduling/5710-workload-aware-preemption/kep.yaml
+++ b/keps/sig-scheduling/5710-workload-aware-preemption/kep.yaml
@@ -41,6 +41,10 @@ feature-gates:
       - kube-apiserver
       - kube-scheduler
       - kube-controller-manager
+  - name: PodGroupPreemptionPolicy
+    components:
+      - kube-apiserver
+      - kube-scheduler
 disable-supported: true
 
 # The following PRR answers are required at beta release
--- a/keps/sig-scheduling/5710-workload-aware-preemption/README.md
+++ b/keps/sig-scheduling/5710-workload-aware-preemption/README.md
@@ -487,9 +487,18 @@ Note that, for workload-aware preemption we will support the `preemptionPolicy` 
 of requestion `PriorityClass` - namely both currently existing modes: `PreemptLowerPriority`
 and `Never`.
 
+This field was not added as a part of the Alpha release of Workload Aware Preemption,
+thus it will lag one release behind the other fields. This means that it cannot be covered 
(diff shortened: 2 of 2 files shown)


keps/sig-scheduling/5710-workload-aware-preemption/kep.yaml
kannon92, line 23: Should we update the stage and the beta graduation?

It sounds like this KEP needs to stay in alpha.

macsko: Not really. The essence of this KEP goes to beta, with a subtle subfeature being 1 release off. It's a bit weird, but I think tracking this as alpha would be even worse

kannon92: So would we track the beta graduate of PodGroupdPreemptoinPolicy?

Argh4k: I think in the ideal world we would have a separate KEP for only this change, but this seems like so minuscule that a separate KEP for it seems to be an overkill. Basically we only change the place from which we take the preemptionpolicy (from pods within podgroup to pod group object) but later, the logic stays the same. 

@liggitt can you help there? How should we track this change?

macsko: DRA structured parameters did something similar that they needed to add a functionality behind a feature gate (`DRASchedulerFilterTimeout`), tracked differently than the main feature (DRA was going to GA, but they needed to introduce a timeout on Filter):
https://github.com/kubernetes/enhancements/blob/481e3710995d9996e27c7364b14d9ab870b65e74/keps/sig-node/4381-dra-structured-parameters/kep.yaml#L39-L47

Tracking specifically the graduation of such a small feature would give an unnecessary overhead that is not needed here.

([DynamicResourceAllocation versioning](https://github.com/kubernetes/kubernetes/blob/0fe889a76062e58098cb7582ee7aa91cf2649812/pkg/features/kube_features.go#L1392-L1396),[ DRASchedulerFilterTimeout versioning](https://github.com/kubernetes/kubernetes/blob/0fe889a76062e58098cb7582ee7aa91cf2649812/pkg/features/kube_features.go#L1359-L1361))

Issue #6232, opened 2026-07-13
Tolerate benign per-NUMA memory drift in the Memory Manager


AI-Armless:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Tolerate a benign, bounded per-NUMA memory drift in the Memory Manager `Static` policy so kubelet does not fail to start after a reboot.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6233
- Discussion Link: https://github.com/kubernetes/kubernetes/issues/131253
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [x] KEP (`k/enhancements`) update PR(s): kubernetes/enhancements#6233
    - [ ] Code (`k/k`) update PR(s): kubernetes/kubernetes#140473
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

/sig node

ffromani:
thanks. This is 1.38 material at earliest

AI-Armless:
thanks! bumped the KEP milestone to 1.38 (alpha).

ffromani:
@AI-Armless hey! would you have capacity to work on this KEP for the 1.38 cycle which is starting as we speak?

AI-Armless:
> [@AI-Armless](https://github.com/AI-Armless) hey! would you have capacity to work on this KEP for the 1.38 cycle which is starting as we speak?

yep, capacity is there for 1.38. i'll drive it through the cycle.

rebasing the KEP on current master now and re-running the verifiers, then i'll ping for the first review pass. code side (#140473) is up too, just needs an ok-to-test to get moving.

AI-Armless:
> [@AI-Armless](https://github.com/AI-Armless) hey! would you have capacity to work on this KEP for the 1.38 cycle which is starting as we speak?

yep, capacity is there for 1.38. i'll drive it through the cycle.

rebasing the KEP on current master now and re-running the verifiers, then i'll ping for the first review pass. code side (#140473) is up too, just needs an ok-to-test to get moving.

tico88612:
Hello @AI-Armless :wave:, 1.38 Enhancements team here.

[...]

PR #6233, opened 2026-07-13
KEP-6232: Tolerate benign per-NUMA memory drift in the Memory Manager

AI-Armless:
This is the initial (provisional) KEP for tolerating a benign, bounded per-NUMA
memory drift in the kubelet Memory Manager `Static` policy (and the delegating
`BestEffort` policy), so kubelet does not crash-loop after a reboot when a node's
per-NUMA `MemTotal` shifts benignly and the node otherwise stays `NotReady` until
an operator deletes the checkpoint by hand.

The background section works through the two independent causes, both confirmed
against real node data on the tracking issue:

- KASLR relocation of the kernel image (MB-scale, needs more than one NUMA node), and
- variable boot-time reserved-memory freeing (KiB-scale, hits even single-NUMA nodes).

The design is in three parts: a bounded tolerance + conservation + re-baseline
(already implemented in the linked code PR), an auto-detected bound derived from
the running kernel image size in `/proc/iomem`, and an operator policy option to
disable or override it.

- Enhancement issue: #6232
- Related code PR (part 1, the bounded tolerance): kubernetes/kubernetes#140473
- Root-cause discussion: kubernetes/kubernetes#131253

/sig node
/cc @ffromani

### DIFF
--- a/keps/sig-node/6232-memory-manager-drift-tolerance/README.md
+++ b/keps/sig-node/6232-memory-manager-drift-tolerance/README.md
@@ -1,0 +1,431 @@
+# KEP-6232: Tolerate benign per-NUMA memory drift in the Memory Manager
+
+## Table of Contents
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Background: why per-node MemTotal moves across a reboot](#background-why-per-node-memtotal-moves-across-a-reboot)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Part 1: bounded tolerance, conservation and re-baseline](#part-1-bounded-tolerance-conservation-and-re-baseline)
+  - [Part 2: auto-detecting the bound](#part-2-auto-detecting-the-bound)
+  - [Part 3: operator option](#part-3-operator-option)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+- [Infrastructure Needed](#infrastructure-needed)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+- [ ] (R) Graduation criteria is in place
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website]
+- [ ] Supporting documentation
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+The Memory Manager `Static` policy records the total memory of each NUMA node in
+its checkpoint and, on kubelet start, requires the recorded per-node totals to
+match the current machine exactly. That total (cAdvisor's
+`/sys/.../nodeN/meminfo` `MemTotal`) is not stable across reboots, so kubelet
+crash-loops on a benign change with `the expected machine state is different
+from the real one`, and the node stays `NotReady` until an operator deletes the
+checkpoint by hand.
+
+This KEP makes the policy tolerate a bounded, benign per-node memory drift and
+re-baseline onto the current machine, while still failing on genuine hardware or
+configuration changes. The bound is auto-detected from the running kernel image
+size, and an operator option can disable or override it.
+
+## Motivation
+
+### Background: why per-node MemTotal moves across a reboot
+
+Two independent effects, at very different scales, both change the per-node
+total that the policy checks.
+
+**1. KASLR relocation of the kernel image (major, MB-scale).** The physical half
+of `CONFIG_RANDOMIZE_BASE` (KASLR) chooses a random, 2 MiB-aligned base for the
+kernel image across all usable RAM on every boot (`find_random_phys_addr()` in
+`arch/x86/boot/compressed/kaslr.c` scans the whole e820/EFI map, not just node
+0). The image is `memblock_reserve`d at that address (`__pa_symbol(_text)`), and
+per-node `MemTotal` is the sum of `zone_managed_pages` = present - reserved, so
+the reservation subtracts from whichever NUMA node owns those physical addresses.
+Between boots the image lands on a different node: one node gains ~the image
+size, another loses it, and the sum across nodes stays constant. The magnitude
+equals the resident image size (`text+rodata+data+bss`; `.init.*` is freed before
+the policy reads meminfo), typically tens of MiB, up to ~130 MiB for a large
+distro kernel. This needs more than one NUMA node to be visible.
+
+Observed in #131253: a bare-metal node moves exactly 12130 pages off node0 /
+12131 onto node1 (~47.4 MiB), and that machine's dmesg reports a ~49 MiB resident
+image (essentially the whole image relocating); a VM case shows +41.67 MiB /
+-41.67 MiB with a bit-for-bit constant sum.
+
+**2. Variable boot-time reserved-memory freeing (minor, KiB-scale).**
+Independently of KASLR, the amount of memory the kernel frees during boot varies
+slightly between boots, so the *total* drifts by tens of KiB, which hits even
+single-NUMA nodes. Observed on a production single-NUMA node: kubelet
+crash-looped with `TotalMemSize1=65839165440` vs `TotalMemSize2=65839153152`, a
+12288-byte (3-page) difference with `systemReserved` unchanged.
+
+Both effects have existed since KASLR became default-on across distros
+(~2017-2018), which predates the Memory Manager (alpha 1.21, 2021). The strict
+per-node equality check never accounted for them, which is why the failure
+surfaces on ordinary reboots.
+
+### Goals
+
+- The `Static`/`BestEffort` policies start after a reboot when the only change is
+  a benign per-node memory drift, without operator intervention.
+- Genuine changes still fail the start: a change to `systemReserved`/reserved
+  memory, a hugepage change, an added/removed memory bank (GiB-scale), or an
+  assignment that no longer fits.
+- The tolerated bound is principled (derived from the running kernel), not a
+  hand-picked constant.
+- Operators can opt out of, or override, the behavior.
+
+### Non-Goals
+
+- Changing how memory is allocated to pods, or the checkpoint format.
+- Eliminating the fluctuation itself (a kernel/firmware concern).
+- Covering the `None` policy (it does not validate machine state).
+
+## Proposal
+
+Three pieces, matching the shape outlined on the tracking issue:
+
+1. **Tolerate a bounded drift and re-baseline** (implemented in
+   kubernetes/kubernetes#140473).
+2. **Auto-detect the bound** from the kernel image size.
+3. **An operator option** to disable the tolerance or set an explicit bound.
+
+### User Stories
+
+- *Routine reboot:* a node reboots (or finishes an OS update) and kubelet comes
+  back `Ready` without anyone deleting `/var/lib/kubelet/memory_manager_state`.
+- *Strict environment:* an operator who wants the previous strict behavior, or a
+  specific bound, sets a policy option.
+
+### Risks and Mitigations
+
+- *Tolerating a real change.* Mitigated by keeping `systemReserved`, hugepages,
+  the assignment structure and the group-reserved sum exact, by bounding the
+  drift, and by the conservation check (an assignment that no longer fits still
+  fails the start).
+- *Wrong auto-detected bound.* Mitigated by a generous grace and a conservative
+  fallback when `/proc/iomem` is unreadable; the operator option is the final
+  override.
+
+## Design Details
+
+### Part 1: bounded tolerance, conservation and re-baseline
+
+In `validateState`, when `areMachineStatesEqual` fails, accept the difference iff
+the states differ only within the tolerated drift: identical topology (nodes,
+NUMA grouping, assignment count), identical `SystemReserved`, hugepage totals
+exact, and the regular-memory `TotalMemSize`/`Allocatable` within the bound per
+node. Per-node `Reserved` is not compared - a drift can legitimately reshuffle a
+cross-NUMA assignment's split while the group total is unchanged, and the
+assignments are re-derived from the persisted blocks. `updateExpectedMachineState`
+returns an error when a recorded assignment no longer fits, so a reduction that
+would under-serve a pod still fails regardless of the bound. On success the policy
+re-baselines with `SetMachineState(expected)`. Implemented and tested in #140473.
+
+### Part 2: auto-detecting the bound
+
+At policy construction, derive the bound from the running kernel image: read the
+`Kernel code`/`Kernel data`/`Kernel bss` lines of `/proc/iomem`, take the physical
+span from the lowest start to the highest end (the image is contiguous; this
+captures rodata/alignment gaps), and add a grace for the KiB-scale secondary drift
+and rounding. The span gives both the movable size and (via the code start and
+`/sys/devices/system/memory/block_size_bytes`) the current node.
+
+Fallbacks: `/proc/iomem` hides its addresses without `CAP_SYS_ADMIN` (a
+KASLR-leak mitigation); kubelet has the capability, but if the addresses read as
+zero, or the file is absent (non-Linux), fall back to a conservative default. The
+read is in kubelet, not cAdvisor: the value is boot-static so cAdvisor's
+collection loop adds nothing, and cAdvisor exposes only `nodeN/meminfo MemTotal`
+today, so it is net-new either way.
+
+On the reported single-NUMA node this yields ~119 MiB (55 MiB image + 64 MiB
+grace), which tolerates the observed 12 KiB drift by ~4 orders of magnitude and
+is tighter than a fixed 256 MiB, so it detects real changes sooner.
+
+### Part 3: operator option
+
+Add a `memoryManagerPolicyOptions` map (mirroring `cpuManagerPolicyOptions` /
+`topologyManagerPolicyOptions`), gated behind the `MemoryManagerDriftTolerance`
+feature gate, with an option to disable the tolerance (restore strict behavior)
+or set an explicit byte bound. The default keeps the auto-detected behavior.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes
+necessary to implement this enhancement.
+
+##### Prerequisite testing updates
+
+None. The existing Memory Manager policy tests already cover the `validateState`
+start-up path this enhancement extends.
+
+##### Unit tests
+
+Part 1 is implemented with unit tests in kubernetes/kubernetes#140473; parts 2-3
+add tests in the same package.
+
+- `k8s.io/kubernetes/pkg/kubelet/cm/memorymanager`: `2026-07-13` - covered by the
+  existing policy suite plus the cases below.
+
+Cases (added / planned):
+- policy: a small drift with no assignments; a small drift where assignments still
+  fit; a reduction that no longer fits (start fails); a drift above the bound
+  (start fails); a cross-NUMA assignment whose split shifts under drift; a manager
+  restart with a drifted `machineInfo`.
+- autodetect: parse the kernel span from a captured `/proc/iomem` (including the
+  rodata gap); reject zeroed / absent addresses; fall back on a missing file.
+- option: strict mode rejects any drift; an explicit bound is honored.
+
+##### Integration tests
+
+None planned. The behavior is kubelet node-local at start-up with no control-plane
+configuration to exercise; unit tests and node e2e cover it.
+
+##### e2e tests
+
+- A node-e2e that writes a Memory Manager checkpoint, restarts kubelet with a
+  slightly different per-NUMA `MemTotal`, and asserts the node returns `Ready`
+  instead of crash-looping.
+- Optionally document the `nokaslr` boot option as the confirming experiment for
+  the KASLR factor (diagnostic only, not a fix).
+
+### Graduation Criteria
+
+- *Alpha:* parts 1-2 behind the Memory Manager machinery; the option (part 3)
+  behind the `MemoryManagerDriftTolerance` feature gate, default off.
+- *Beta:* option on by default; e2e_node coverage; no open correctness issues.
+- *GA:* soak across releases with no regressions.
+
+### Upgrade / Downgrade Strategy
+
+No configuration change is required to keep working: a kubelet with the feature
+tolerates benign drift automatically, and an operator who wants the previous
+strict behavior sets the opt-out option. The checkpoint format is unchanged, so
+there is no state migration. On downgrade the kubelet reverts to strict equality
+(and the pre-existing reboot failure can reappear).
+
+### Version Skew Strategy
+
+This is a node-local kubelet behavior with no control-plane coordination and no
+change to the checkpoint format, CRI, CNI or CSI. Nodes without the feature keep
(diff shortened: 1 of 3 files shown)


AI-Armless:
gentle ping @ffromani, no rush. anything needed from my side before a first pass here, or someone to loop in for ok-to-test? happy to bring it to a sig-node meeting if that's the easier route.

ffromani:
/ok-to-test

@AI-Armless would you have capacity to push this work in the 1.38 cycle?

(cc @Tal-or )

AI-Armless:
> /ok-to-test
> 
> @AI-Armless would you have capacity to push this work in the 1.38 cycle?
> 
> (cc @Tal-or )

thanks @ffromani. good for 1.38 on my end.

rebased on master, kep now implementable, toc + metadata + node_approvers all green. main change since the first draft: the whole behavior sits behind the MemoryManagerDriftTolerance gate now, option included. @ffromani @Tal-or ptal when you get a chance. @johnbelamaric for the PRR (keps/prod-readiness/sig-node/6232.yaml) whenever there's a minute.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+its checkpoint and, on kubelet start, requires the recorded per-node totals to
+match the current machine exactly. That total (cAdvisor's
+`/sys/.../nodeN/meminfo` `MemTotal`) is not stable across reboots, so kubelet
+crash-loops on a benign change with `the expected machine state is different
+from the real one`, and the node stays `NotReady` until an operator deletes the
+checkpoint by hand.
+
+This KEP makes the policy tolerate a bounded, benign per-node memory drift and
+re-baseline onto the current machine, while still failing on genuine hardware or
+configuration changes. The behavior is gated by the `MemoryManagerDriftTolerance`
```
ffromani: this is of course correct but note that FG come intrinsically with an expiration date attached, while the KEP embodies the design so should be timeless. Therefore, here and in the rest of the document, please refer to FGs only when required by the template.
IOW, a good rule of thumb: would the text mentioning the FG still be accurate and relevant 10 releases from now, when (hopefully!) the feature is GA and the FG is removed per due process?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+kernel image across all usable RAM on every boot (`find_random_phys_addr()` in
+`arch/x86/boot/compressed/kaslr.c` scans the whole e820/EFI map, not just node
+0). The image is `memblock_reserve`d at that address (`__pa_symbol(_text)`), and
+per-node `MemTotal` is the sum of `zone_managed_pages` = present - reserved, so
+the reservation subtracts from whichever NUMA node owns those physical addresses.
+Between boots the image lands on a different node: one node gains ~the image
+size, another loses it, and the sum across nodes stays constant. The magnitude
+equals the resident image size (`text+rodata+data+bss`; `.init.*` is freed before
+the policy reads meminfo), typically tens of MiB, up to ~130 MiB for a large
+distro kernel. This needs more than one NUMA node to be visible.
```
ffromani, line 105: if you mention linux kernel data structures and behavior, please pin them to a specific kernel version (e.g. 7.2.3 - random version).
In addition, please mention right in the Summary that this feature is linux-specific and will devolve into a no-operation (NOP) on windows. Likewise, we need to list explicitly the archs (e.g. x86_64 only) this feature is about, right upfront in the summary

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Proposal
+
+Three pieces, matching the shape outlined on the tracking issue, all behind the
+`MemoryManagerDriftTolerance` feature gate (alpha: disabled by default). With the
+gate disabled the policy keeps today's strict per-node equality check.
+
+1. **Tolerate a bounded drift and re-baseline** (implemented in
+   kubernetes/kubernetes#140473).
+2. **Auto-detect the bound** from the kernel image size.
```
ffromani: can we also autodetect if KASLR is enabled, or is it sufficient to assume it is enabled?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+3. **An operator option** to disable the tolerance or set an explicit bound.
```
ffromani: kubelet option or memory manager option? (new concept, but natural extension of cpu and topology manager options)

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+### User Stories
+
+- *Routine reboot:* a node reboots (or finishes an OS update) and kubelet comes
+  back `Ready` without anyone deleting `/var/lib/kubelet/memory_manager_state`.
+- *Strict environment:* an operator who wants the previous strict behavior leaves
+  the feature gate off, or sets the policy option to disable the tolerance or pin
+  a specific bound.
```
ffromani: this makes sense, but it seems it's the same story from two different angles.
Plus, the user stories should be told from the perspective of a persona:
"As cluster administrator, if I set "Static" as memory manager option, when I reboot the nodes..."

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+### Risks and Mitigations
+
+- *Tolerating a real change.* Mitigated by keeping `systemReserved`, hugepages,
+  the assignment structure and the group-reserved sum exact, by bounding the
+  drift, and by the conservation check (an assignment that no longer fits still
+  fails the start).
+- *Wrong auto-detected bound.* Mitigated by a generous grace and a conservative
+  fallback when `/proc/iomem` is unreadable; the operator option is the final
+  override.
```
ffromani: on which cases we anticipate `/proc/iomem` to be unreadable?
would we need special permissions to read that file? If so, can we use other files which require no extra permissions?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Design Details
+
+### Part 1: bounded tolerance, conservation and re-baseline
+
+In `validateState`, when `areMachineStatesEqual` fails and the
+`MemoryManagerDriftTolerance` gate is enabled, accept the difference iff the
```
ffromani: if you refer to kubelet code, please pin it to a specific kubelet version

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+NUMA grouping, assignment count), identical `SystemReserved`, hugepage totals
+exact, and the regular-memory `TotalMemSize`/`Allocatable` within the bound per
+node. Per-node `Reserved` is not compared - a drift can legitimately reshuffle a
+cross-NUMA assignment's split while the group total is unchanged, and the
+assignments are re-derived from the persisted blocks. `updateExpectedMachineState`
+returns an error when a recorded assignment no longer fits, so a reduction that
+would under-serve a pod still fails regardless of the bound. On success the policy
+re-baselines with `SetMachineState(expected)`. With the gate disabled the existing
+error is returned unchanged. Implemented and tested in kubernetes/kubernetes#140473;
+the gate is added to that change as part of the alpha work.
```
ffromani: this helps, but the KEP is about discussing and documenting the design, so an existing implementation is not a strong reference; quite the contrary: depending on the community conversation the KEP process enables, we may need to implement differently.
Having a reference PR however helps to ground the conversation and works as running demo.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+### Part 2: auto-detecting the bound
+
+At policy construction, derive the bound from the running kernel image: read the
```
ffromani: at memory manager initialization?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+`Kernel code`/`Kernel data`/`Kernel bss` lines of `/proc/iomem`, take the physical
```
ffromani, line 205: ok, retracting my previous concern `/proc/iomem` is world-readable by default. We will however ensure that adding this new direct dependency on procfs is fine

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+span from the lowest start to the highest end (the image is contiguous; this
+captures rodata/alignment gaps), and add a grace for the KiB-scale secondary drift
+and rounding. The span gives both the movable size and (via the code start and
+`/sys/devices/system/memory/block_size_bytes`) the current node.
+
+Fallbacks: `/proc/iomem` hides its addresses without `CAP_SYS_ADMIN` (a
+KASLR-leak mitigation); kubelet has the capability, but if the addresses read as
+zero, or the file is absent (non-Linux), fall back to a conservative default. The
```
ffromani: would the default depend on the kernel size? If so, this will change over time. Maybe would it be safer to fail hard if we can't autodetect and let the admin set a value?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+`the expected machine state is different from the real one`, or nodes going
+`NotReady` after reboot under the `Static` memory manager policy.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Not yet; this is alpha. The checkpoint format is unchanged, so no state is
+migrated in either direction: enabling the gate only changes how an existing
+checkpoint is validated at start, and disabling it restores the strict check on
+the same file. The enable -> disable -> enable path will be exercised manually on
+a node before beta.
```
ffromani: I think there's no need to test this path as the feature makes no changes to API or disk files

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+It is node-level, not workload-level. Its effect is visible in kubelet logs (a
+re-baseline message when a benign drift is tolerated) and in the node staying
+`Ready` after a reboot.
```
ffromani: we can probably add a metric to report the max tolerated drift and the actual drift. These are easy to add and would inform the operator the feature is being used, albeit not by workloads.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,477 @@
...
+
+- [x] Other (treatment): a benign drift is logged as tolerated and re-baselined
+  at start; a drift beyond the bound still fails with the existing error. A
+  counter (`memory_manager_drift_tolerated_total`, beta) will make this
+  observable via metrics.
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+Node startup succeeds after a benign reboot drift, with no change to allocation
+behavior.
```
ffromani: I think we should expect no SLO change for this feature, as it affects only the start path with negligible overhead

ffromani (commented):
initial review, mostly tightening/process, nothing big

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: done, the gate is now only in graduation criteria, PRR and kep.yaml; the design text doesn't mention it.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: pinned to Linux 6.12 on x86-64 (and 4.6 for the /proc/iomem address hiding). summary now says linux-only: validated on x86-64, arm64 expected via the same /proc/iomem entries and to be confirmed during alpha, no-op on windows and everything else.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: no need: the image size bounds the shift whether KASLR is on or not, with nokaslr the image never moves and the tolerance just isn't exercised. reading nokaslr from /proc/cmdline would add a dependency without changing any decision. added a sentence in part 2.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: memory manager policy option: a new memoryManagerPolicyOptions field on KubeletConfiguration, same shape as cpu/topology manager options. clarified in the proposal.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: rewritten as three personas: reboot (Ready without touching the state file), real capacity loss (must still refuse to start), strictly controlled fleet (exact comparison or a pinned bound).

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: world-readable, addresses zeroed without CAP_SYS_ADMIN since 4.6, kubelet has it; entries absent on non-linux. and per your later comment there's no guessing anymore when it can't be read: exact comparison stays, log line points at the option.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: pinned to kubelet v1.37.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: reworded: #140473 and #142121 are reference implementations to ground the discussion, the KEP text is what counts.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: yes, when the memory manager is initialized and the static policy is created. reworded.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: kubelet and cadvisor already read procfs (/proc/meminfo, /proc/cpuinfo), noted in the text.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: agreed. the bound follows the running kernel by design (recomputed on every start), and when it can't be read there's no constant fallback anymore: tolerance stays off, exact comparison, log line pointing at memory-drift-tolerance. dropped the 256Mi default from the KEP and from the PR.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: agreed, changed to not applicable: no API and no on-disk changes.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: added two kubelet gauges for alpha: memory_manager_drift_tolerance_bytes (bound in effect, 0 = exact comparison) and memory_manager_memory_drift_bytes{numa_node} (drift seen at the last start). in the PR as well.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: agreed, changed to "no SLO change expected".

AI-Armless:
@ffromani all addressed in the last commit, ptal. code PRs updated to match (#140473, #142121).

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,521 @@
...
+12288-byte (3-page) difference with `systemReserved` unchanged.
+
+Both effects have existed since KASLR became default-on across distros
+(~2017-2018), which predates the Memory Manager (alpha 1.21, 2021). The strict
+per-node equality check never accounted for them, which is why the failure
+surfaces on ordinary reboots.
+
+### Goals
+
+- The `Static`/`BestEffort` policies start after a reboot when the only change is
```
Tal-or: Worth mentioning explicitly that those are MemoryManager policies.
Also `BestEffort` is a Windows only option and this KEP is about Linux platform only.

keps/sig-node/6232-memory-manager-drift-tolerance/kep.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - "@AI-Armless"
+owning-sig: sig-node
+participating-sigs: []
+status: implementable
+creation-date: "2026-07-13"
+reviewers:
+  - "@ffromani"
+  - "@Tal-or"
+approvers:
+  - "@mrunalp"
```
Tal-or, line 13: I think the approver should match the one from the PRR

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,521 @@
...
+kubernetes/kubernetes#140473 (parts 1-2) and kubernetes/kubernetes#142121
+(part 3) are reference implementations that ground the discussion; the design
+in this document is what counts.
+
+### User Stories
+
+- As a cluster administrator running the `Static` memory manager policy, when
+  my nodes reboot (a kernel update, a power event) I want kubelet to come back
+  `Ready` on its own, without someone deleting
+  `/var/lib/kubelet/memory_manager_state` on every node.
```
Tal-or: From the user story it sounds like this is an operation that is always required.
It would be more accurate to specify that the issue happens from time to time and that's the undesried workaround.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,521 @@
...
+(the bound could not be derived, or the option turned it off) the existing error
+is returned unchanged.
+
+### Part 2: deriving the bound
+
+When the memory manager is initialized at kubelet start (the static policy
+object is created), derive the bound from the running kernel image: read the
+`Kernel code`/`Kernel data`/`Kernel bss` lines of `/proc/iomem`, take the physical
+span from the lowest start to the highest end (the image is contiguous; this
+captures rodata/alignment gaps), and add a 64 MiB grace for the KiB-scale
```
Tal-or, line 207: Do we have any solid evidence that 64MiB grace is enough?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+On the reported single-NUMA node this yields ~119 MiB (55 MiB image + 64 MiB
+grace), which tolerates the observed 12 KiB drift by ~4 orders of magnitude and
+stays far below a memory bank, so a real change is still caught.
+
+### Part 3: operator option
+
+Add a `memoryManagerPolicyOptions` map to the kubelet configuration (mirroring
+`cpuManagerPolicyOptions` / `topologyManagerPolicyOptions`) with one option,
+`memory-drift-tolerance`: `auto` (default: the bound derived from the kernel
```
Tal-or, line 242: is that ok to have this option `auto` by default for a feature which is in alpha stage?

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,521 @@
...
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `MemoryManagerDriftTolerance`
+  - Components depending on the feature gate: kubelet
+
+###### Does enabling the feature change any default behavior?
+
+Yes. The `Static` policy (and `BestEffort`, which delegates to the same
```
Tal-or: `BestEffort` is available only for Windows

Tal-or (commented):
Looking good overall 
Small comments from my side

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: done: "memory manager Static policy", and BestEffort is now listed under non-goals as windows-only.

keps/sig-node/6232-memory-manager-drift-tolerance/kep.yaml
AI-Armless: they're different roles by design: kep.yaml approvers is the sig-node approver (mrunalp, a sig-node tech lead, which test/node_approvers_test.go requires for an alpha KEP), the PRR approver (johnbelamaric) lives in keps/prod-readiness/sig-node/6232.yaml. happy to switch to whoever sig-node assigns.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: reworded: it happens on some reboots and not others, and deleting the state file is the workaround, not a routine step.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: not solid, no: 12 KiB on our node, tens of KiB in the #131253 reports, and the mechanism is KiB to low-MiB, so 64Mi is about three orders of magnitude of headroom. that's what the observed-drift metric is for during alpha; if the field numbers come in higher the grace changes before beta. spelled this out in the text.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: the option's default only kicks in once the feature is enabled, which in alpha means the gate is on; nobody gets auto by upgrading. once enabled, auto is the behavior this KEP is about, off keeps the exact comparison. added a sentence.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: dropped, static only.

AI-Armless:
@Tal-or thanks, all addressed in the last commit.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,533 @@
...
+1. **Tolerate a bounded drift and re-baseline.**
+2. **Derive the bound** from the kernel image size when the memory manager is
+   initialized; when it cannot be derived, keep the exact comparison.
+3. **A memory manager policy option** (`memoryManagerPolicyOptions`, a new
+   `KubeletConfiguration` field that mirrors `cpuManagerPolicyOptions` and
+   `topologyManagerPolicyOptions`) to disable the tolerance or pin the bound.
+
+kubernetes/kubernetes#140473 (parts 1-2) and kubernetes/kubernetes#142121
+(part 3) are reference implementations that ground the discussion; the design
+in this document is what counts.
```
ffromani: nit: this is implicit in the KEP process. I'd just delete it

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
ffromani: I agree this is acceptable for alpha. Gathering this feedback and refining would be a beta graduation criteria

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
```diff
@@ -0,0 +1,533 @@
...
+
+Add a `memoryManagerPolicyOptions` map to the kubelet configuration (mirroring
+`cpuManagerPolicyOptions` / `topologyManagerPolicyOptions`) with one option,
+`memory-drift-tolerance`: `auto` (default: the bound derived from the kernel
+image, or the exact comparison when it cannot be derived), `off` (the exact
+comparison) or an explicit quantity such as `128Mi`. Unknown options and values
+are rejected at kubelet start, as for the other managers. This is how an
+operator keeps the exact comparison, or sets the bound on a platform where it
+cannot be derived. `auto` is the default only once the feature itself is
+enabled; an administrator who has not enabled it sees no change.
```
ffromani: for compatibility reasons, the existing behavior should be the default, regardless from FG state. Therefore we will need a double opt-in mechanism like basically each every feature we introduced:
- enable the FG
- THEN set the `memory-drift-tolerance=auto` (or non-`off` in general)

As a template, we should mimic `TopologyManagerPolicyOptions` rather than `CPUManagerPolicyOptions` (in hindsight, the perma-alpha and perma-beta cpumanagerpolicyoptions created more head scratchers than they solved. I for one I'd like to move away from that model)

ffromani (commented):
we just need to turn the default to align to all the other KEP work, and then from my side LGTM

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: done, deleted.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: added to the beta criteria: the grace confirmed or refined from the observed-drift metric gathered during alpha.

keps/sig-node/6232-memory-manager-drift-tolerance/README.md
AI-Armless: done. unset = exact comparison regardless of the gate; enabling is the gate plus memory-drift-tolerance=auto (or a quantity), off spells it out. reshaped the option after topologymanager's policy_options.go (PolicyOptions struct, NewPolicyOptions, CheckPolicyOptionAvailable with alpha/beta/stable sets), with one difference I'd like you to confirm: the option's availability is tied to the MemoryManagerDriftTolerance gate itself, no separate alpha/beta option gates, so it graduates with the feature. code updated in #140473: the option is the on-switch now, so it moved into that PR and #142121 is closed.

AI-Armless:
@ffromani default flipped, all three addressed in the last commit. ptal

AI-Armless:
@johnbelamaric PRR request for the alpha in 1.38: the questionnaire is filled out and keps/prod-readiness/sig-node/6232.yaml lists you. both sig-node reviewers' comments are addressed, ffromani said he's ready to lgtm now that the default flip landed. enhancements freeze is the 29th, so whenever you have a minute this week. thanks!

ffromani:
/lgtm

@AI-Armless thanks for addressing all the comments. We're good from my side now.

AI-Armless:
/assign @mrunalp
@mrunalp could you take a look for sig-node approval? ffromani lgtm'd, enhancements freeze is on the 29th.

PR #6235, opened 2026-07-14
KEP-4929: Add ExecAction timeout for lifecycle hooks

googs1025:
- One-line PR description: Add an optional timeoutSeconds field for lifecycle hook ExecAction, preserving the existing no-timeout behavior by default.

- Issue link: https://github.com/kubernetes/enhancements/issues/4929

- Other comments:
Follow-up to the lifecycle ExecAction timeout discussion in https://github.com/kubernetes/kubernetes/issues/128250. This PR adds the initial KEP-4929 proposal for bounding PostStart and PreStop exec lifecycle hooks with an optional timeoutSeconds field. The model keeps unset or zero timeoutSeconds as the current no-timeout behavior, applies positive values only to lifecycle hook exec actions, and avoids extending the kubelet /run HTTP API. Because ExecAction is shared with probes, the KEP explicitly keeps probe timeout behavior on Probe.timeoutSeconds as covered by KEP-1972, and proposes rejecting non-zero probe.exec.timeoutSeconds values. The proposal also calls out feature-gate behavior, version skew, downgrade considerations, and initial validation/test expectations.

### DIFF
--- a/keps/sig-node/4929-exec-action-timeout/README.md
+++ b/keps/sig-node/4929-exec-action-timeout/README.md
@@ -1,0 +1,588 @@
+# KEP-4929: ExecAction timeout for lifecycle hooks
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [API](#api)
+  - [Execution semantics](#execution-semantics)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Validation](#validation)
+  - [Kubelet behavior](#kubelet-behavior)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+    - [Upgrade](#upgrade)
+    - [Downgrade](#downgrade)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Set a default lifecycle exec timeout](#set-a-default-lifecycle-exec-timeout)
+  - [Add timeout support to the kubelet <code>/run</code> API](#add-timeout-support-to-the-kubelet-run-api)
+  - [Require users to wrap commands with image-specific timeout tooling](#require-users-to-wrap-commands-with-image-specific-timeout-tooling)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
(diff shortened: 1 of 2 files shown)


keps/sig-node/4929-exec-action-timeout/README.md
```diff
@@ -0,0 +1,588 @@
...
+  `preStop`.
+- Preserve existing behavior when `timeoutSeconds` is unset or set to `0`.
+- Reject negative timeout values.
+- Implement the feature behind a `PodLifecycleExecActionTimeout` feature gate.
+- Avoid changes to the kubelet `/run` HTTP API.
+
+### Non-Goals
+
+- This KEP does not define default timeouts for lifecycle exec hooks.
+- This KEP does not change probe timeout behavior. Exec probes already use
```
googs1025, line 115: 1. `ExecAction` is shared by lifecycle hooks and probes. This KEP intentionally keeps probe timeout behavior unchanged because exec probes already use `Probe.timeoutSeconds` and are
  covered by KEP-1972. The proposal currently says kube-apiserver should reject non-zero `probe.exec.timeoutSeconds` and direct users to `probe.timeoutSeconds`.

googs1025:
cc @tallclair @haircommander @kannon92

keps/sig-node/4929-exec-action-timeout/README.md
```diff
@@ -0,0 +1,588 @@
...
+
+### Goals
+
+- Add an optional `timeoutSeconds` field to `ExecAction`.
+- Support the field for lifecycle hook exec actions in both `postStart` and
+  `preStop`.
+- Preserve existing behavior when `timeoutSeconds` is unset or set to `0`.
+- Reject negative timeout values.
+- Implement the feature behind a `PodLifecycleExecActionTimeout` feature gate.
+- Avoid changes to the kubelet `/run` HTTP API.
```
toVersus: The Goals section isn't the place to write implementation details, so I don't think detailed specifications need to be included here.

```suggestion
- Allow users to set an optional timeout value for exec actions in lifecycle hooks to prevent arbitrary command execution from hanging indefinitely
```

keps/sig-node/4929-exec-action-timeout/README.md
```diff
@@ -0,0 +1,588 @@
...
+
+When kubelet executes a lifecycle hook with `exec.timeoutSeconds`:
+
+- `timeoutSeconds` unset or `0`: pass `0` to `RunInContainer`, preserving the
+  current no-timeout behavior.
+- `timeoutSeconds` greater than `0`: pass that duration to `RunInContainer`.
+- If the exec action times out, the lifecycle hook fails and kubelet handles
+  that failure the same way it handles other lifecycle exec failures.
+
+For `preStop`, the hook remains subject to the existing pod termination flow and
```
toVersus, line 201: This could be misunderstood by some users (e.g., setting `timeoutSeconds` to 60 seconds while leaving `terminationGracePeriodSeconds` at the default 30 seconds), so I think it needs to be clearly documented in the official documentation as well. I understand that hard validation on the API server side, comparing `timeoutSeconds` on the preStop hook against `terminationGracePeriodSeconds`, isn't possible, since `terminationGracePeriodSeconds` can be overridden at pod termination phase.

keps/sig-node/4929-exec-action-timeout/README.md
```diff
@@ -0,0 +1,588 @@
...
+### User Stories (Optional)
+
+#### Story 1
+
+As a cluster user, I want to set a timeout on a `postStart` exec hook so that a
+stuck startup script cannot block the kubelet pod worker indefinitely.
+
+#### Story 2
+
+As a workload author, I want to set a timeout on a `preStop` exec hook so that
```
toVersus: I can't think of a concrete use case off the top of my head, but this is about cases where it's fine for the preStop hook to be cut off partway through once `timeoutSeconds` elapses, while still wanting the application to shut down gracefully within the `terminationGracePeriodSeconds` window,  right?

keps/sig-node/4929-exec-action-timeout/README.md
```diff
@@ -0,0 +1,588 @@
...
+
+### Kubelet behavior
+
+The kubelet lifecycle handler currently calls:
+
+```go
+hr.commandRunner.RunInContainer(ctx, containerID, handler.Exec.Command, 0)
+```
+
+The implementation will derive a timeout from `handler.Exec.TimeoutSeconds`.
```
toVersus, line 282: Just out of curiosity, in the exec action timeout case, does this message actually indicate that it was a timeout?
https://github.com/kubernetes/kubernetes/blob/17db279a2dfd593dae61b627d4564908f13047c5/pkg/kubelet/lifecycle/handlers.go#L83

keps/sig-node/4929-exec-action-timeout/README.md
googs1025: Added notes that preStop exec timeout does not extend the pod termination grace period, and called out that this interaction needs user-facing documentation.

keps/sig-node/4929-exec-action-timeout/README.md
googs1025: Yes, that is the intended case. I updated the story to describe best-effort cleanup/drain work with a smaller budget inside the pod termination grace period.

keps/sig-node/4929-exec-action-timeout/README.md
googs1025: Updated the Goals section to keep it at the user-goal level and moved the implementation details to the proposal/design sections.

keps/sig-node/4929-exec-action-timeout/README.md
googs1025: The current message depends on the RunInContainer error. I updated the KEP to require timeout failures to be distinguishable in the failure message / kubelet log.

googs1025:
I forgot to add you  😄 Thanks!
/assign @toVersus

googs1025:
@tallclair /PTAL

Issue #6236, opened 2026-07-14
KEP-4322: Define cluster inventory by ClusterProfile namespace


kahirokunn:
Follow-up to https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75. The SIG-MC call on July 07, 2026 [reconfirmed namespace scope](https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75#issuecomment-4917369795), and the next step is documenting the current behavior and rationale. This issue tracks the KEP-4322 side of that.

The CRD has been [namespace-scoped since the first implementation](https://github.com/kubernetes-sigs/cluster-inventory-api/blob/04603eea/apis/v1alpha1/clusterprofile_types.go#L126) (April 2024), but the KEP still hedges:

- "Cluster Inventory" is just ["a conceptual term referring to a collection of clusters"](https://github.com/kubernetes/enhancements/blob/48e6d949/keps/sig-multicluster/4322-cluster-inventory/README.md#terminology)
- one ClusterProfile per member cluster is only [recommended](https://github.com/kubernetes/enhancements/blob/48e6d949/keps/sig-multicluster/4322-cluster-inventory/README.md#uniqueness-of-the-clusterprofile-object)
- [cluster name uniqueness](https://github.com/kubernetes/enhancements/blob/48e6d949/keps/sig-multicluster/4322-cluster-inventory/README.md#cluster-name) is global, with no defined boundary
- [namespace scoping itself is a recommendation](https://github.com/kubernetes/enhancements/blob/48e6d949/keps/sig-multicluster/4322-cluster-inventory/README.md#how-should-we-organize-clusterprofile-objects-on-a-hub-cluster), not a statement of fact

Nobody has relied on this looseness in the years since, but consumers pay for it anyway: they can't assume what a namespace of ClusterProfiles means or whether duplicates can appear, so every manager/consumer pair ends up working it out case by case.

Proposal: make the namespace the boundary.

- A **cluster inventory** is the set of ClusterProfile objects in one namespace. A hub cluster can host multiple inventories.
- Within an inventory, a member cluster is represented by at most one ClusterProfile (MUST), whichever cluster manager created it. The same cluster may appear in other inventories.
- Cluster name uniqueness is per inventory, not global.

Cross-inventory identity and deduplication stay out of scope — a consumer reading multiple namespaces is consuming multiple inventories.

PR #6237, opened 2026-07-14
KEP-4322: Define inventories by ClusterProfile namespace

kahirokunn:
- One-line PR description: Define inventories by ClusterProfile namespace

- Issue link: https://github.com/kubernetes/enhancements/issues/6236

- Other comments:

Follow-up to https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75 (namespace scope [reconfirmed](https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75#issuecomment-4917369795) on the July 07, 2026 SIG-MC call): the ClusterProfile objects in one namespace form one cluster inventory, and the uniqueness rules are scoped to it.

Per https://github.com/kubernetes/enhancements/pull/6237#discussion_r3592200728, cluster inventories are defined independently of ClusterSet.


lauralorenz:
Can we move the push model to Alternatives Considered instead of removing it entirely so that we have somewhere to point to in the future? ex when existing current tools that do that style want to adopt ClusterProfile. And give any more rationale about why it is not recommended if there was anything new discussed since it was first added.

MrFreezeex (approved):
This change makes sense to me, thanks! 

I read this and the PR on the website and I am not entirely sure how much it actually help with your intiial problem with the ArgoCD consumer?

Are you going to have a have a configurable list of inventory/namespace that the user can configure there and let him handle the conflict wrt to what they configured somehow?

mikeshng (commented):
The SIG-MC community is thinking about decoupling the ClusterProfile from ClusterSet and fully support the namespace scoping of ClusterProfile to do grouping.

mikeshng (commented):
What if instead of clusterSet we just replace it with clusterInventory as defined in this KEP? We will also need https://github.com/kubernetes/enhancements/pull/6239 but instead of multiple clusterSet it will be multiple clusterInventory. So the clusterProperty will look something like this

```
apiVersion: about.k8s.io/v1beta1
kind: ClusterProperty
metadata:
  name: clusterinventory.k8s.io
spec:
  value: prod # or multiple values
```

That way we have a real definition of clusterInventory and we solve the clusterProfile should be able to belong to multiple groups problem.

kahirokunn:
> What if instead of clusterSet we just replace it with clusterInventory as defined in this KEP? We will also need #6239 but instead of multiple clusterSet it will be multiple clusterInventory. So the clusterProperty will look something like this
> 
> ```
> apiVersion: about.k8s.io/v1beta1
> kind: ClusterProperty
> metadata:
>   name: clusterinventory.k8s.io
> spec:
>   value: prod # or multiple values
> ```
> 
> That way we have a real definition of clusterInventory and we solve the clusterProfile should be able to belong to multiple groups problem.

@mikeshng mikeshng Does this look about right? I've made the changes. If you're satisfied, I'd be grateful if you could leave a review. Thank you 🙏

keps/sig-multicluster/4322-cluster-inventory/README.md
```diff
@@ -1024,8 +1033,8 @@ status:
   version:
     kubernetes: 1.28.0
   properties:
-   - name: clusterset.k8s.io
-     value: some-clusterset
+   - name: clusterinventory.k8s.io
```
mikeshng: Do we need a new KEP to clearly define this first? I feel like this KEP depends on it.

@lauralorenz It's related to what we discussed yesterday about moving away from `clusterSet`. We are considering defining a new ClusterProperty, `clusterInventory`, that is similar to clusterSet but allows a cluster to belong to multiple inventories.

mikeshng (commented):
We will need to bring this idea to the SIG-MC community for discussion. There are still lots of clusterSet mentions in the KEP. We might need to drastically change those references if the community is ok with this approach.

kahirokunn:
> We will need to bring this idea to the SIG-MC community for discussion. There are still lots of clusterSet mentions in the KEP. We might need to drastically change those references if the community is ok with this approach.

Should we remove all information about ClusterSets from the ClusterInventoryAPI in this PR? I, personally, am in favor of this.

keps/sig-multicluster/4322-cluster-inventory/README.md
kahirokunn: I am a bit concerned that the About API is described as being for ClusterSet.
https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid

I feel that it is probably not appropriate to add it to KEP-2149, so I included it in KEP-4322, but should I create a new KEP for this?
By the way, just for reference, do you have any thoughts on the specific use cases for clusterinventory.k8s.io?

mikeshng:
~~Yes, we should remove/update all references of clusterSet. Assuming the community is ok with it. This is a big change and we will need community consensus first for sure.~~

See thread in https://github.com/kubernetes/enhancements/pull/6237#discussion_r3588565052

keps/sig-multicluster/4322-cluster-inventory/README.md
mikeshng: I agree. Maybe we just define the new inventory property here in this KEP update.

The main use case is basically replaces how this KEP/ClusterProfile leverages clusterSet but allows for multiple inventories (vs a cluster cannot belong to multiple clusterSets). This is a big change and will require community agreement.

MrFreezeex: Hmm I am a bit confused as what is the main use case for this property is 🤔. If a namespace is an inventory do we need also a property to indicate this?

If we want to do this does the value represent all the namespace names somehow? Can they be different?

mikeshng: @MrFreezeex That's a great point! Maybe we don't need to do that, since it would introduce additional complexity around the namespaces on the hub vs what's stored in ClusterProperty. They can differ, and we'd have to add even more complexity to explain those differences. Initially, I was thinking of replacing all ClusterSet references with the proposed ClusterInventor. But it might be simpler to leave the ClusterSet references as they are and explicitly state that a ClusterProfile cannot belong to multiple ClusterSets, while it can belong to multiple inventories.

MrFreezeex: For what it's worth on the MCS-API side we may also want to evolve the clusterset notion. I don't know if our target should be that a cluster could be part of multiple clusterset or to completely remove that notion somehow 🤔.

For instance for Cilium we don't really have any notion of clusterset it's more a matter that a cluster can be connected to a number of remote clusters but we don't have to connect to all the clusters (and we don't have any notion of transitivity too). The main relevant requirements is that each connected cluster have a distinct id and name though which could loosely map to a clusterset somehow. It looks like that it might be a somewhat similar set of requirements as what you have in OCM somehow?

We probably want to have some similar notions for MCS-API/cluster profile as for instance in (again) Cilium, we could imagine building an operator inside a hub cluster that "pilot" clusters connections by consuming ClusterProfile (something I just imagined, may not be built in the short/medium term).

This doesn't mean we need to solve this (or all of this) as part of this PR, for instance defining the notion of inventory per namespace and later on tackling this question of where we want to go with clusterset could probably be fine!

mikeshng: +1 @kahirokunn let's focus this PR on what you had before which is clearly defining a namespace == a clusterInventory without talking about the ClusterProperty clusterinventory.k8s.io. To address @lauralorenz 's concern regarding clusterSet being everywhere in the KEP. I think we need a section on ClusterSet vs the new NS/ClusterInventory.

kahirokunn: Thanks. I agree that a section comparing Cluster Inventory and ClusterSet would be useful. However, I would like to clarify why I hesitate to define a normative one-to-one relationship between them.

The current text says that an inventory is considered a ClusterSet when all of its member clusters adhere to namespace sameness. However, namespace sameness is only one part of the MCS API definition of a
ClusterSet. Satisfying namespace sameness alone does not establish the other ClusterSet properties, such as mutual trust, shared ownership, symmetric and transitive membership, and agreement among member
clusters.

This concern is based on implementation experience rather than only a hypothetical use case. I have built an operator on a hub cluster that consumes ClusterProfile objects and configures connectivity between
clusters. The underlying implementations use Cilium Cluster Mesh and Istio ambient mode multicluster.

Consider the following scenario:

1. The number of logical cluster groups grows dynamically.
2. Clusters within each group are connected automatically.
3. Different groups may also be connected according to explicit policies, conceptually similar to AWS PrivateLink.

If every such group must map to a separate inventory namespace, creating a group also requires creating a namespace and updating the controller’s RBAC. Granting the controller cluster-wide access avoids that
RBAC churn, but it also gives the controller visibility into unrelated inventories and weakens the intended isolation boundary.

There is also a problem when the same physical cluster participates in multiple logical groups. Mapping each group to a separate inventory namespace would require representing the same cluster in multiple
inventories, while this KEP intentionally does not define cross-inventory identity or deduplication.

One possible approach would be to deploy a separate connectivity operator for each ClusterSet. However, that feels more like a workaround imposed by the one-namespace-per-ClusterSet model than a desirable
architecture.

In the implementation I described, the controller only needs access to one inventory. Logical groups and automatic connection boundaries can be represented using labels, ClusterProperties, or a consumer-owned
membership resource. This allows the groups to change dynamically without creating namespaces or updating RBAC, and it allows one cluster to participate in multiple groupings.

This also raises the question of whether these groups should be called ClusterSets at all. For example, connections between groups may not be transitive, and memberships may overlap. If this KEP is referring
specifically to an MCS API ClusterSet, I think it should say so explicitly and retain the exact MCS API semantics. If the intent is to define a more general grouping concept, it may require a separate KEP, an
additional custom resource, or another API that models membership explicitly.

Is there a concrete consumer for which inventory-scoped selection and consumer-owned membership information are insufficient, and which specifically requires this KEP to guarantee that one inventory namespace
forms one MCS API ClusterSet?

Until such a requirement is identified, I think KEP-4322 should define Cluster Inventory independently from ClusterSet membership and leave the grouping semantics to consumers or the relevant MCS implementation.

WDYT?

kahirokunn: I've made the changes. Would you mind checking them over? Thank you 🙏

MrFreezeex: The last changes looks good to me!

Just one question though, I guess that you would practically want that most producers support projecting their ClusterProfile into multiple namespace (like supporting the one inventory per consumer mode ~)? Is that common today and would producer needs anything from some sig-multicluster API to help do that somehow? It's probably not required to fully solve those questions for this KEP change (if not already solved) but could be still good to know our options!

MrFreezeex: Also note that I didn't meant to suggest to make clusterset and namespace/inventory equivalent.

But if ClusterProfile move to being independent of ClusterSet it would probably be one more argument to evolve what a ClusterSet is as it's one of the required property for About API and ClusterProfile is one of the main use case for About API (or maybe even the only one so far even? As you may know I do plan to use it more in network related thing like https://github.com/kubernetes/enhancements/pull/5940 or even cilium network policies though).

It's more a thought about something we would need to discuss later/not in this PR anyway ~.

mikeshng (approved):
/lgtm

@kahirokunn thanks again for driving this discussion. I feel like we are almost at the finish line.

Your PR changes aligned exactly with what was discussed in the SIG-MC community meeting.
This is still a fundamental change (a correct one but still big), so I think we should share this with the community and get more eyes on it.

keps/sig-multicluster/4322-cluster-inventory/README.md
kahirokunn: > The last changes looks good to me!
> 
> Just one question though, I guess that you would practically want that most producers support projecting their ClusterProfile into multiple namespace (like supporting the one inventory per consumer mode ~)? Is that common today and would producer needs anything from some sig-multicluster API to help do that somehow? It's probably not required to fully solve those questions for this KEP change (if not already solved) but could be still good to know our options!

Ideally yes, but in practice this works even without producer-side support for multi-namespace projection.

I recently surveyed the existing implementations (not merged yet):
https://github.com/kahirokunn/sig-multicluster-site/blob/fe1ca13e4030752ffb4eefa591a6f2e512484c40/site-src/implementations/cluster-inventory-api-implementations.md#implementations-1

Today there are two publishing models:

- **Shared inventory** (GKE Fleet, KubeFleet): all ClusterProfiles are published into a single namespace (`fleet-cluster-inventory` by default for GKE Fleet, `fleet-system` for KubeFleet). These producers cannot create a minimal per-consumer inventory. Consumers such as Argo CD or Kueue can still be used by deploying them against that shared namespace, but then every consumer sees the full inventory, which I don't think is desirable from a least-privilege standpoint.
- **Projected inventories** (OCM): each namespace's inventory is built from the `ManagedClusterSetBinding` objects in that namespace, and the same cluster can be projected into multiple inventories. OCM is currently the only producer that can project ClusterProfiles into multiple namespaces, i.e. the only one that supports the one-inventory-per-consumer mode today.

Either way, the per-consumer mode is already implementable with what this KEP defines. The projection mechanism (deciding which clusters land in which namespace) is producer-specific for now, e.g. OCM's `ManagedClusterSetBinding`.

kahirokunn:
@mikeshng 
Thank you for the review!
It would be a huge help if you could bring this up as an agenda item at our next meeting.
Thank you 🙌

mikeshng:
>  It would be a huge help if you could bring this up as an agenda item at our next meeting.

Absolutely!

keps/sig-multicluster/4322-cluster-inventory/README.md
```diff
@@ -461,31 +467,37 @@ manager.
...
 
-#### Uniqueness of the ClusterProfile object
-While there are no strict requirements, we recommend that there is only one ClusterProfile object representing any member cluster
-on a hub cluster. 
+![two consumer-scoped inventories on one hub cluster sharing a member cluster](./consumer-scoped-inventories.svg)
 
-However, a ClusterProfile object can only be in one ClusterSet since the namespace sameness property is transitive, therefore 
-it can only be in the namespace of that clusterSet if it is in a ClusterSet.
+#### Uniqueness of the ClusterProfile object
+Within a cluster inventory, each member cluster SHOULD be represented by at most one
```
mikeshng, line 498: SIG-MC meeting feedback (July 21, 2026)

The rationale for the `SHOULD` requirement is understood: nothing prevents multiple ClusterProfile resources with different names from representing the same logical cluster.

However, it would be valuable to define a standard mechanism for detecting or warning about duplicate representations of the same logical cluster within a namespace, rather than relying solely on resource names. For example, a well defined label (such as `managed-by`) or another Kubernetes standard label or annotation could provide a consistent way to identify resources that refer to the same logical cluster.

This would help prevent conflicts in scenarios where multiple ClusterProfile resources target the same cluster.

For example, during the meeting a real world use case was discussed in which two ClusterProfile resources in the same namespace represented the same logical cluster. Each resource contained different configuration, and a controller reconciled those resources against the same target cluster. As a result, the controller competed to apply different state, leading to conflicts and race conditions.

keps/sig-multicluster/4322-cluster-inventory/README.md
```diff
@@ -461,31 +467,37 @@ manager.
...
-A cluster inventory may or may not represent a ClusterSet. A cluster inventory is considered a [clusterSet](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#terminology)
-if all its member clusters adhere to the [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md) principle.
-Note that a cluster can only be in one ClusterSet while there is not such restriction for a cluster inventory.
+A cluster inventory is defined by its namespace and is independent of ClusterSet.
+This KEP does not define a mapping between cluster inventories and [clusterSets](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/1645-multi-cluster-services-api#terminology):
+a clusterSet requires properties beyond [namespace sameness](https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md),
+such as mutual trust, shared ownership, and symmetric and transitive membership, and those properties
+are established by an MCS implementation rather than by this API.
```
mikeshng, line 474: SIG-MC meeting feedback (July 21, 2026) The proposed decoupling of certain ClusterSet responsibilities in this KEP was well received by the SIG-MC community, which welcomes this change.

kahirokunn: That's great to hear! Thanks for confirming!

keps/sig-multicluster/4322-cluster-inventory/README.md
kahirokunn: Thanks for relaying the meeting feedback @mikeshng!

I've updated the KEP to define this. Instead of introducing a new label, it builds on the existing well-known id.k8s.io property from KEP-2149: ClusterProfiles in the same namespace with the same id.k8s.io value claim the same cluster identity. The platform admin should resolve such duplicates by deleting all but one.

Until then, consumers surface a warning and act only on the object with the oldest creationTimestamp. The thinking is that a newly created ClusterProfile carrying the same id.k8s.io as an existing one is most likely a mistake, so preferring the oldest object keeps whatever is already running against the existing one undisturbed. If two or more objects share the oldest timestamp there is no safe pick, so consumers act on none of them.

This should also cover the case from the meeting: only one object is actionable at a time, so controllers won't compete over the same target cluster.

Once we reach agreement on this in the KEP, I'm planning to implement a utility in kubernetes-sigs/cluster-inventory-api to support this behavior, so consumers don't each have to reimplement the dedup logic.

mikeshng: SGTM! Just want to say:

> Once we reach agreement on this in the KEP, I'm planning to implement a utility in kubernetes-sigs/cluster-inventory-api to support this behavior, so consumers don't each have to reimplement the dedup logic.

We might not need to do that. There will be some corner cases where consumers will want to have dupes for their specific reasons. As long as the KEP/API is clearly stated, I think it should be fine. The sig-mc api projects (mcs-api, about-api) tends to shy away from providing actual default implementation.

kahirokunn: Sounds good. If any concrete issues come up, I'll revisit it then 💪

kahirokunn:
just fixed conflict

mikeshng:
CC @LogicalShark

LogicalShark:
Thanks for the CC! LGTM and I will start looking at updating GCP's fleet clusterprofile syncer with the new ID field and potentially multi-namespace support

mikeshng:
Thanks @LogicalShark for the review! Maybe wait until the PR merged first before making any changes. We are still seeking broader consensus, and your approval representing GCP fleet was definitely a significant step forward!

skitt:
I think the KEP-4322 changes here make sense — they clarify cluster inventories a lot, but without affecting any clusterset assumptions. It’s on the agenda for today’s SIG call so I’ll wait but I think this is ready to merge.

keps/sig-multicluster/2149-clusterid/README.md
```diff
@@ -274,28 +274,29 @@ nitty-gritty.
...
+[RFC-1123](https://tools.ietf.org/html/rfc1123) DNS subdomain, and should be
+less than 128 characters in total.
 
 While it is a member of a ClusterSet, a cluster must also have an additional
 `clusterset.k8s.io ClusterProperty` which describes its current membership. This
 property must be present as long as the cluster's membership in a
 ClusterSet lasts, and removed when the cluster is no longer a member.
 
 More detail and examples of the uniqueness, lifespan, immutability, and content
-requirements for both the `cluster.clusterset.k8s.io ClusterProperty` and
```
mikeshng, line 296: I don't think we can make this change. Especially not in this PR. At best, it will be another PR but even that will be contentious. 

I was looking at the 4322 changes only and completely missed this in my previous code review. 

I will discuss this today in the SIG-MC meeting.

skitt:
The 2149 changes are too invasive. We can’t change `cluster.clusterset.k8s.io` like this: this was chosen specifically to reflect the fact that the identifier is only supposed to be unique within a given clusterset, for reasons explained in the KEP; in particular, a desire to allow human-readable values, and most significantly, previous bad experiences with supposedly globally-unique identifiers that weren’t. That last reason means that even if we could try to *add* a globally-unique `id.k8s.io`, relying on it in practice is likely to cause issues; KubeFed in particular tried this in the past and ran into problems.

I recommend dropping the changes to 2149 and adjusting 4322 to suit. If something like `id.k8s.io` really would make life easier, we can try to come up with a definition, but that should be a new KEP (and a reminder that we need to finish defining the “property ladder”).

keps/sig-multicluster/4322-cluster-inventory/README.md
```diff
@@ -1014,6 +1053,7 @@ metadata:
  namespace: some-cluster-inventory
  labels:
    x-k8s.io/cluster-manager: some-cluster-manager
+   multicluster.x-k8s.io/inventory-member-id: cluster-us-east
```
mikeshng, line 1057: Nice. This label and together with the behavior described in this PR, addresses the final piece of review feedback from the SIG-MC community. It provides consumers with a reliable way to determine whether two or more ClusterProfile resources within the same namespace (inventory) represent the same logical cluster.

/lgtm

mikeshng (approved):

keps/sig-multicluster/4322-cluster-inventory/README.md
kahirokunn: Thanks!

I force-pushed one scope clarification: consumer-side deduplication now applies to the set of ClusterProfile objects selected by the consumer's configuration, rather than automatically to every ClusterProfile in an inventory.

For example, an inventory may contain ten ClusterProfiles while a consumer selects only a subset through a label selector or explicit object references. ClusterProfiles outside that selection do not affect the consumer's operation, so requiring them to participate in that consumer's deduplication would be unnecessarily broad.

The inventory-level rule remains unchanged: when multiple ClusterProfiles in the same inventory have the same non-empty inventory member ID, the platform administrator should remove all but one. A consumer surfaces a warning when its selected set contains more than one of them.

```diff
-A consumer SHOULD deduplicate only among objects in the inventories it is
-configured to read. The consumer SHOULD act on only the matching object with
-the oldest `creationTimestamp` from its configured inventories;
+A consumer SHOULD deduplicate only among ClusterProfile objects selected by
+its configuration, such as inventory namespaces, label selectors, or object
+references. Among objects with the same value in that set, the consumer
+SHOULD act only on the object with the oldest `creationTimestamp`;
```

```diff
-administrator SHOULD delete all but one, and consumers SHOULD surface a
-warning until the issue is resolved.
+administrator SHOULD delete all but one. Consumers that select more than one
+of these objects SHOULD surface a warning until the issue is resolved.
```

keps/sig-multicluster/2149-clusterid/README.md
kahirokunn: Thank you. fixed.

keps/sig-multicluster/4322-cluster-inventory/README.md
kahirokunn: I noticed one more case and force-pushed a refinement: deduplication and the warning now apply only when the consumer's actions on the duplicate objects would conflict.

The previous wording asked consumers to deduplicate and warn whenever their selection contained duplicates, even when acting on all of them is intentional and harmless. For example, a GitOps tool may have two Applications that deliberately reference two ClusterProfile objects claiming the same member cluster, each deploying a disjoint set of resources. In pseudo-code:

```yaml
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ClusterProfile
metadata:
  name: prod-team-a
  namespace: argocd
  labels:
    multicluster.x-k8s.io/inventory-member-id: prod-cluster
---
apiVersion: multicluster.x-k8s.io/v1alpha1
kind: ClusterProfile
metadata:
  name: prod-team-b
  namespace: argocd
  labels:
    multicluster.x-k8s.io/inventory-member-id: prod-cluster
---
kind: Application
metadata:
  name: team-a-app
spec:
  clusterProfileRef:
    name: prod-team-a
  source:
    path: team-a/  # disjoint from team-b/
---
kind: Application
metadata:
  name: team-b-app
spec:
  clusterProfileRef:
    name: prod-team-b
  source:
    path: team-b/
```

Nothing competes there, so dropping one object or surfacing a warning adds no value.

```diff
-A consumer SHOULD deduplicate only among ClusterProfile objects selected by
-its configuration, such as inventory namespaces, label selectors, or object
-references.
+A consumer SHOULD deduplicate only among ClusterProfile objects selected by
+its configuration, such as inventory namespaces, label selectors, or object
+references, and only when its actions on those objects would conflict.
```

```diff
-Consumers that select more than one
-of these objects SHOULD surface a warning until the issue is resolved.
+Consumers that select more than one
+of these objects, and whose actions on all of them would conflict, SHOULD
+surface a warning until the issue is resolved.
```

kahirokunn: This document has also been updated to reflect the latest information.
https://github.com/kubernetes-sigs/sig-multicluster-site/pull/65

mikeshng (approved):
/lgtm

@skitt I think @kahirokunn has addressed all our concerns (discussed in the SIG-MC call) in the latest revision. Could you PTAL? Thanks!

skitt:
Thanks @kahirokunn!

/approve

Issue #6238, opened 2026-07-14
KEP-2149, KEP-4322: Allow a cluster to belong to multiple ClusterSets


kahirokunn:
## Enhancement Description

- One-line enhancement description: Remove the single-ClusterSet-membership assumption from KEP-2149 and KEP-4322 so a cluster can belong to multiple ClusterSets at the same time.
- Kubernetes Enhancement Proposal: [KEP-2149: ClusterId for ClusterSet identification](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/2149-clusterid), [KEP-4322: ClusterProfile API](https://github.com/kubernetes/enhancements/tree/master/keps/sig-multicluster/4322-cluster-inventory)
- Discussion Link:
  - https://github.com/kubernetes-sigs/cluster-inventory-api/issues/75
  - https://github.com/kubernetes-sigs/cluster-inventory-api/issues/76
  - https://github.com/kubernetes-sigs/about-api/issues/29
- PRs by stage and milestone:
  - [ ] KEP (`k/enhancements`) update PR(s): TBD

## Context

KEP-2149 lists a cluster joining multiple ClusterSets as an explicit non-goal, and KEP-4322 states that a cluster can only be in one ClusterSet. Real cluster managers don't work that way: OCM's ManagedClusterSet can place one cluster in multiple sets, and the duplicate-cluster problems in cluster-inventory-api#75 partly come from squeezing multi-set membership through a single-membership API.

A cluster can't carry two ClusterProperty resources with the same name (property name = resource name), so multiple memberships have to be encoded in a single `clusterset.k8s.io` value.

## Proposed Change

- KEP-2149: the `clusterset.k8s.io` property value becomes a comma-separated list of every ClusterSet the cluster belongs to; move multi-membership from Non-Goals to Goals.
- KEP-4322: ClusterSet membership is expressed by the `clusterset.k8s.io` property; drop the requirement to label the inventory namespace with `clusterset.multicluster.x-k8s.io`; cluster managers SHOULD mirror each membership as a `clusterset.multicluster.x-k8s.io/<name>: "true"` label so ClusterProfiles stay selectable with standard label selectors (precedent: `node-role.kubernetes.io/<role>`).

PR #6239, opened 2026-07-14
KEP-2149, KEP-4322: Allow a cluster to belong to multiple ClusterSets

kahirokunn:
- One-line PR description: Remove the single-ClusterSet-membership assumption from KEP-2149 and KEP-4322, encoding multiple memberships as a comma-separated `clusterset.k8s.io` value

- Issue link: https://github.com/kubernetes/enhancements/issues/6238

- Other comments:
One commit per KEP:

- KEP-2149: the `clusterset.k8s.io` property value becomes a comma-separated list of every ClusterSet the cluster belongs to; multi-membership moves from Non-Goals to Goals.
- KEP-4322: ClusterSet membership is expressed by the `clusterset.k8s.io` property; the requirement to label the inventory namespace with `clusterset.multicluster.x-k8s.io` is dropped; cluster managers SHOULD mirror each membership as a `clusterset.multicluster.x-k8s.io/<name>: "true"` label so ClusterProfiles stay selectable with standard label selectors.

### DIFF
--- a/keps/sig-multicluster/2149-clusterid/README.md
+++ b/keps/sig-multicluster/2149-clusterid/README.md
@@ -245,6 +245,8 @@ and managed as Kubernetes resources
     the same hostnames
   * Facilitate enrichment of log / event / metrics data with cluster id / set
     coordinates
+* Support expressing that a cluster belongs to multiple ClusterSets
+  simultaneously
 
 ### Non-Goals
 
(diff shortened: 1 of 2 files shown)


mikeshng:
CC @lauralorenz

mikeshng (commented):
This update might not be needed if we do https://github.com/kubernetes/enhancements/pull/6237#pullrequestreview-4697386801

mikeshng (commented):
Instead of multiple clusterSet, how about multiple clusterInventory instead: https://github.com/kubernetes/enhancements/pull/6237#pullrequestreview-4700154007

kahirokunn:
That sounds good. I've fixed 🙏

mikeshng (commented):
I think we can close this since the SIG-MC decided to not to go with this direction.

PR #6245, opened 2026-07-15
KEP: Surge capacity on eviction instead of blocking on PodDisruptionBudget

monta-riahi:
## What this PR does

Adds a **provisional** KEP proposing an opt-in behavior where an exhausted PodDisruptionBudget triggers a *surge* of replacement capacity before an eviction completes, instead of blocking the eviction outright with `429 TooManyRequests`.

Today API-initiated eviction is a binary decision against the PDB: if `DisruptionsAllowed == 0` the eviction is rejected indefinitely. For a single-replica Deployment with `minAvailable: 1` this makes the workload undrainable without an outage. This KEP proposes bringing up a replacement first (like `maxSurge` in a rolling update) and releasing the eviction once the replacement is Ready.

The document is intentionally `provisional`: it frames the problem, lays out the three cooperating pieces (signal → surge → release), and enumerates the open design questions for SIG discussion (where surge intent lives, who owns the surge, non-surgeable workloads, and the async eviction contract). It is expected to compose with KEP-4212 Declarative Node Maintenance.

/sig node
/sig apps
/kind kep

### DIFF
--- a/keps/sig-node/NNNN-surge-on-eviction/README.md
+++ b/keps/sig-node/NNNN-surge-on-eviction/README.md
@@ -1,0 +1,257 @@
+# KEP-NNNN: Surge capacity on eviction instead of blocking on PodDisruptionBudget
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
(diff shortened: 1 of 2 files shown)


monta-riahi:
Description from kubernetes/kubernetes #140599, 

> When K8s currently evicts a pod, the pod gets terminated/deleted and only then the controller will schedule a new replacement pod, this means that deployments with one replica will experience outage, unlike when rolling them out which will surge a new pod first and only then terminate the old one.

> This is rather poor behavior, and it would be much nicer if K8s can be configured to block eviction until a new pod is scheduled (Surging the configured replicas count if needed) instead. e.g. Instead of a PodDisruptionBudget just outright blocking the eviction, it triggers a capacity surge to support the requested budget, allowing the eviction to continue once the new pods are available.

Issue #6247, opened 2026-07-18
Kubelet systemd watchdog diagnostic guardrails


googs1025:
/sig node
/kind feature
/stage alpha

## Summary

Improve kubelet's Linux systemd watchdog integration so operators can identify why watchdog heartbeats stopped before a systemd-triggered kubelet restart.

## Motivation

Today kubelet may stop notifying systemd because a watchdog health checker fails, a checker stalls, `SdNotify()` returns an error, or `SdNotify()` blocks. Operators often only see the eventual watchdog-triggered restart, without enough pre-restart evidence to identify the source.

This enhancement is intended to provide bounded execution and default-visible diagnostics for the existing kubelet systemd watchdog path.

## Scope

- Per-check timeout diagnostics for kubelet watchdog health checkers.
- Timeout diagnostics around `SdNotify()`.
- Default-visible structured logs for `SdNotify()` returned errors and timeout paths.
- Retry summary logging when watchdog notification retries are exhausted.
- Unit tests for success, checker error, checker timeout, notify error, notify timeout, retry, watchdog-disabled, and non-Linux compatibility paths.

## Non-Goals

- No kubelet config API changes.
- No metrics, events, or NodeCondition in the initial scope.
- No non-Linux behavior changes.
- No redesign of kubelet health checking.
- No broader node self-healing behavior.

## Related Issue

- kubernetes/kubernetes#135449

## Notes

The initial design intent is to keep this focused on operational diagnostics and internal kubelet watchdog guardrails. If SIG Node prefers this to proceed as a focused implementation PR instead of a KEP, this issue can be closed or repurposed based on reviewer guidance.

SergeyKanzhelev:
> /label lead-opted-in /milestone v1.38

@haircommander did you get signal that this will be worked on?

googs1025:
> > /label lead-opted-in /milestone v1.38
> 
> [@haircommander](https://github.com/haircommander) did you get signal that this will be worked on?

I’m actively driving this work. The KEP is open as kubernetes/enhancements#6248, and I’ve incorporated the initial review feedback. I’ll continue working with SIG Node on the deeper review and follow up with the implementation plan for v1.38.

SergeyKanzhelev:
great, looking forward for it

lasomethingsomething:
Hello  @googs1025 :wave:, 1.38 Enhancements team here.

[...]

PR #6248, opened 2026-07-19
KEP-6247: Kubelet watchdog diagnostics

googs1025:
- One-line PR description: Add an initial KEP proposal for kubelet systemd watchdog diagnostic guardrails, guarded by the KubeletWatchdogDiagnostics feature gate.

- Issue link: https://github.com/kubernetes/enhancements/issues/6247

- Other comments:
Follow-up to the kubelet systemd watchdog diagnostics discussion in https://github.com/kubernetes/kubernetes/issues/135449. This PR adds the initial KEP-6247 proposal for improving diagnostic evidence before systemd watchdog-triggered kubelet restarts. The proposal covers bounded watchdog health-check execution, bounded SdNotify calls, default-visible SdNotify failure logging, structured watchdog failure logs, and retry summaries when notification attempts are exhausted. The scope is intentionally kubelet-local: it does not add Kubernetes API surface, kubelet configuration fields, metrics, events, or NodeConditions. Because bounded waits can affect watchdog failure-path behavior, the KEP proposes the kubelet feature gate KubeletWatchdogDiagnostics for alpha rollout and rollback. The proposal also calls out timeout budget constraints, single-flight handling for blocked operations, version skew, downgrade considerations, and initial unit-test expectations.

### DIFF
--- a/keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
+++ b/keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
@@ -43,6 +43,7 @@
 - [Alternatives](#alternatives)
   - [Only raise <code>SdNotify()</code> error log visibility](#only-raise-sdnotify-error-log-visibility)
   - [Add kubelet configuration for watchdog timeout budgets](#add-kubelet-configuration-for-watchdog-timeout-budgets)
+  - [Only wrap <code>SdNotify()</code> outside the source library](#only-wrap-sdnotify-outside-the-source-library)
   - [Add metrics or events](#add-metrics-or-events)
 - [Infrastructure Needed](#infrastructure-needed)
 <!-- /toc -->
@@ -119,6 +120,8 @@ notify call.
 - Identify `SdNotify()` calls that exceed their timeout budget.
 - Summarize exhausted notification retry attempts.
 - Preserve existing successful watchdog heartbeat behavior.
+- Keep the watchdog loop non-blocking and reliable when an individual health
+  checker or systemd notification operation stalls.
 - Avoid new Kubernetes API, kubelet configuration, metric, event, or
   NodeCondition surface in the initial implementation.
 
@@ -127,9 +130,12 @@ notify call.
 - This KEP does not add kubelet configuration fields for watchdog timeout
   budgets.
 - This KEP does not add metrics, events, or NodeConditions.
+- This KEP does not initially add Node Problem Detector integration.
 - This KEP does not redesign kubelet health checking.
 - This KEP does not define node self-healing behavior beyond existing systemd
   watchdog behavior.
+- This KEP does not dynamically disable watchdog health checks that fail or
+  time out.
 - This KEP does not change the non-Linux watchdog implementation.
 - This KEP does not guarantee forced cancellation of health checker or
   `SdNotify()` internals that block in non-cancellable code.
@@ -156,32 +162,55 @@ The alpha implementation is guarded by the `KubeletWatchdogDiagnostics`
 kubelet feature gate. The gate gives SIG Node and operators a rollback switch
 while timeout budgets and log volume are validated.
 
+During implementation, kubelet should evaluate whether the underlying
+`SdNotify()` path can accept cancellation directly. If the current source
+library cannot support context-aware notification calls, the implementation may
+need a small library contribution or kubelet-local adapter. This KEP does not
+require a broad redesign of the notification library, but it should not assume
+that a goroutine wrapper is the only viable implementation.
+
 ### User Stories
 
 #### Story 1: Triage watchdog-triggered kubelet restarts
 
-As a cluster operator, I want kubelet to leave clear diagnostic evidence before
-a systemd watchdog-triggered restart so that I can understand why the node
-stopped sending watchdog heartbeats without needing to reproduce the incident.
+A cluster operator receives an alert that kubelet restarted on a node, for
+example from existing process restart monitoring or systemd journal entries. The
+operator checks the normal kubelet logs around the restart timestamp and finds a
+structured watchdog diagnostic entry. The entry indicates whether the last
+watchdog iteration failed because a health checker returned an error, a checker
+timed out, `SdNotify()` returned an error, or notification retries were
+exhausted. The operator uses that evidence to decide whether to investigate
+kubelet health, node-local systemd notification delivery, or a failing watchdog
+checker.
 
 #### Story 2: Preserve evidence at default log levels
 
-As an on-call SRE, I want watchdog failure evidence to appear in the normal
-kubelet logs collected from production nodes so that an incident can be
-investigated after the fact, even when verbose logging was not enabled.
+An on-call SRE investigates a watchdog-triggered kubelet restart after the node
+has already recovered. The SRE only has the default kubelet logs collected from
+the affected node and cannot reproduce the failure with higher verbosity. The
+SRE finds default-visible watchdog failure logs with stable fields such as the
+operation, elapsed time, timeout budget, retry attempt, and returned error when
+available. Those fields let the SRE preserve the incident evidence and avoid
+classifying the restart only as an unexplained kubelet process restart.
 
 #### Story 3: Separate node environment issues from kubelet health issues
 
-As a node operator, I want to distinguish a kubelet-internal health problem from
-a node-local systemd notification problem so that I can route the fix to the
-right owner instead of treating every watchdog restart as the same failure.
+A node operator compares kubelet logs with systemd journal entries after a
+watchdog restart. If kubelet reports a watchdog health-check failure, the
+operator investigates kubelet internals and the failing checker. If kubelet
+reports `SdNotify()` errors, notification timeouts, or exhausted notification
+retries, the operator investigates node-local systemd notification delivery
+instead. The diagnostic distinction reduces the chance of misdiagnosing a node
+environment problem as a kubelet health problem, or the reverse.
 
 #### Story 4: Investigate repeated node restarts at fleet scale
 
-As a platform engineer responsible for many nodes, I want watchdog restart
-diagnostics to use consistent fields so that I can group similar failures across
-nodes and identify whether an incident is isolated to one machine, one node
-image, or a broader kubelet behavior.
+A platform engineer sees repeated kubelet watchdog restarts across multiple
+nodes from existing restart-count monitoring. The engineer queries collected
+kubelet logs for the structured watchdog operation fields and groups failures by
+checker name, notify error, timeout, and retry exhaustion. This helps determine
+whether the incident is isolated to one machine, correlated with a node image or
+systemd configuration, or likely caused by broader kubelet behavior.
 
 ### Notes/Constraints/Caveats
 
@@ -212,6 +241,10 @@ notify call for the current iteration.
 - Wrapping blocking calls can leave goroutines running after timeout. The
   implementation must treat timeouts as diagnostic boundaries and avoid
   unbounded goroutine accumulation.
+- Dynamically disabling a failing or timed-out watchdog health checker could
+  make the watchdog loop appear healthy while hiding a real kubelet health
+  problem. The initial implementation should keep failed checks visible and
+  continue treating them as failed for the current watchdog iteration.
 - Default-visible notify error logs could be noisy in repeatedly failing
   environments. The implementation should log exceptional failure and timeout
   paths at default visibility while keeping success logs at high verbosity.
@@ -486,6 +519,13 @@ Metrics for watchdog checker timeout count or notify failure count could be
 useful in the future, but they are intentionally out of scope for the initial
 proposal to avoid expanding metric stability and label cardinality concerns.
 
+This KEP does not propose Node Problem Detector integration for alpha. The
+initial recommendation is to keep the diagnostic signal in kubelet logs because
+the first user story is local post-restart investigation using kubelet and
+systemd logs. If SIG Node finds that these diagnostics need a cluster-visible
+surface, a follow-up enhancement could evaluate whether NPD conditions, events,
+or another node-level reporting mechanism is appropriate.
+
 ### Dependencies
 
 ###### Does this feature depend on any specific services running in the cluster?
@@ -572,6 +612,16 @@ for important watchdog restart cases.
 This would provide operator control, but it would add kubelet configuration API
 surface and significantly increase review scope. The initial proposal keeps
 timeouts internal and derived from the existing watchdog interval.
+
+### Only wrap `SdNotify()` outside the source library
+
+A kubelet-local wrapper around `SdNotify()` would minimize changes to
+dependencies, but it may leave blocked notification calls running until the
+underlying operation returns. During implementation, kubelet should evaluate
+whether the notification source library can support cancellation directly. A
+small context-aware library change may be preferable if it avoids lingering
+blocked notification operations without expanding this KEP into a broad library
+redesign.
 
 ### Add metrics or events
 
--- a/keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
+++ b/keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
@@ -1,0 +1,585 @@
+# KEP-6247: Kubelet Systemd Watchdog Diagnostic Guardrails
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Triage watchdog-triggered kubelet restarts](#story-1-triage-watchdog-triggered-kubelet-restarts)
+    - [Story 2: Preserve evidence at default log levels](#story-2-preserve-evidence-at-default-log-levels)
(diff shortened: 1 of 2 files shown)


keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,585 @@
...
+does not leave enough diagnostic information before it stops sending
+heartbeats. This increases time to recovery for node operators and makes it hard
+to distinguish kubelet health-check problems from systemd notification problems.
+
+Bounded diagnostics provide root-cause hints before systemd restarts kubelet.
+Operators should be able to answer whether kubelet skipped a heartbeat because
+of a failed checker, a stalled checker, a returned notify error, or a blocked
+notify call.
+
+### Goals
```
SergeyKanzhelev, line 115: add a goal to keep watchdog checks non-blocking and reliable

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,585 @@
...
+- Identify failed watchdog health checkers by name.
+- Identify watchdog health checkers that exceed their timeout budget.
+- Identify `SdNotify()` calls that return errors at default-visible log levels.
+- Identify `SdNotify()` calls that exceed their timeout budget.
+- Summarize exhausted notification retry attempts.
+- Preserve existing successful watchdog heartbeat behavior.
+- Avoid new Kubernetes API, kubelet configuration, metric, event, or
+  NodeCondition surface in the initial implementation.
+
+### Non-Goals
```
SergeyKanzhelev, line 129: should this KEP also contribute to `SdNotify` source library to make it accept cancellation context?

SergeyKanzhelev: also what other things we can improve in that library? Let's not limit the exploration of the work scope

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
SergeyKanzhelev, line 115: Are there any self-healing from faulty checks? Or dynamic disablement of checks needed?

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,585 @@
...
+of a failed checker, a stalled checker, a returned notify error, or a blocked
+notify call.
+
+### Goals
+
+- Identify failed watchdog health checkers by name.
+- Identify watchdog health checkers that exceed their timeout budget.
+- Identify `SdNotify()` calls that return errors at default-visible log levels.
+- Identify `SdNotify()` calls that exceed their timeout budget.
+- Summarize exhausted notification retry attempts.
```
SergeyKanzhelev, line 121: would we need NPD to expose this information in a way that is better than "frequent restarts detected"?

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,585 @@
...
+- Emit a structured retry summary when notification retries are exhausted.
+
+This proposal is intentionally internal to kubelet. It does not add new API
+surface and does not change how users enable systemd watchdog.
+
+The alpha implementation is guarded by the `KubeletWatchdogDiagnostics`
+kubelet feature gate. The gate gives SIG Node and operators a rollback switch
+while timeout budgets and log volume are validated.
+
+### User Stories
```
SergeyKanzhelev, line 175: please formulate user stories in a form of "actions taken", not the "desired sentiment". I.e. - cluster operators detects kubelet restart (how?), looks in logs (what  level is expected?), and see the issue (what can it done with it and can it be misdiagnosed?). Formulated this way we may discover gaps in current story better.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Added one goal about keeping the watchdog loop non-blocking and reliable when one checker or notify call stalls. The design already has bounded wait and single-flight parts, so I made this explicit in Goals.

googs1025: Added this to non-goals and risks. For now I do not propose dynamically disabling bad checks, because it may hide a real kubelet health issue. A timed-out check is still treated as failed for that watchdog iteration.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Expanded this a bit. The KEP now says we should look at direct cancellation support in the notification library, and also mentions the tradeoff of only wrapping SdNotify in kubelet.

googs1025: Added text to say implementation should check if the SdNotify path can support cancellation directly. If not, we may need a small library change or a kubelet adapter. I tried to keep this from becoming a large library redesign.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Updated this as an alpha scope suggestion. For now the KEP keeps the signal in kubelet logs, since the first use case is local investigation after restart. If SIG Node thinks this needs a cluster-visible signal, a follow-up can look at NPD or another mechanism.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Rewrote the stories to be more action based: operator sees a restart, checks normal kubelet logs and the systemd journal time range, reads structured fields, then decides what to investigate next.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: I checked the current kubelet code. The watchdog path calls go-systemd directly here:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/watchdog/watchdog_linux.go#L49-L51

And the watchdog loop calls that interface here:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/watchdog/watchdog_linux.go#L147-L151

So one possible implementation is to change the kubelet `WatchdogClient` interface to accept context, then decide whether the default implementation uses a future go-systemd `SdNotifyContext` API or a kubelet-local adapter first.

SergeyKanzhelev (commented):
A few additonal notes, deeper review will be at 1.38 timeframe:

- I observed some process freezes that led to kubelet watchdog kills. I wonder if there are anything we can do to help with those? Maybe explore the threads dumps on watchdog kill as a good practice? At least documenting it.
- Need to check if logging to super-high latency disk may cause issues.
- I think the timeouts we use are a little off. Let's review how many sdnotify attempts kubelet will have during the watchdog interval to ensure we have 2 at least, maybe more.

googs1025:
Thanks for the notes. I updated the KEP to cover these parts:

- Added troubleshooting guidance to check systemd journal and available core/process stack evidence around watchdog-triggered kubelet termination. I kept kubelet-managed goroutine/thread dump generation out of alpha scope, but documented it as useful evidence when it is available.
- Added high latency or stalled journal/log storage as a limitation and risk. The diagnostic logs still depend on node-local logging making progress before systemd terminates kubelet.
- Tightened the timeout budget text for `SdNotify()`. The KEP now says the watchdog interval should keep room for at least two `SdNotify()` attempts when health checks pass, including bounded waits and retry backoff.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,664 @@
...
+
+- Identify failed watchdog health checkers by name.
+- Identify watchdog health checkers that exceed their timeout budget.
+- Identify `SdNotify()` calls that return errors at default-visible log levels.
+- Identify `SdNotify()` calls that exceed their timeout budget.
+- Summarize exhausted notification retry attempts.
+- Preserve existing successful watchdog heartbeat behavior.
+- Keep the watchdog loop non-blocking and reliable when an individual health
+  checker or systemd notification operation stalls, while preserving enough
+  watchdog interval budget for multiple `SdNotify()` attempts.
```
googs1025, line 125: Important part for the timeout concern: the goal is not only to make the loop bounded, but also to keep enough watchdog interval budget so kubelet still has multiple chances to send `SdNotify()`.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,664 @@
...
+the checker is still in flight and treat the checker as failed for that
+iteration. The same single-flight rule applies to `SdNotify()` attempts: if a
+previous notify call is still in flight, kubelet should not start another
+notify call for the current iteration.
+
+These diagnostics also depend on node-local logging making progress before the
+watchdog termination. Very high latency or stalled journal or log storage can
+delay or prevent diagnostic log persistence. Operators should correlate kubelet
+logs with systemd journal entries and, when available, core dumps or process
+stack evidence from the watchdog-triggered termination.
```
googs1025, line 243: This is for the high-latency disk question. The KEP should not promise logs always survive before watchdog kill, because journal or log storage can also be the slow part.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,664 @@
...
+### Bounded systemd notification
+
+Each `SdNotify()` attempt should be run with a timeout budget derived from the
+same iteration deadline used for health checkers. Retry backoff and retry
+attempts should fit inside the remaining iteration budget. The budget
+calculation must reserve room for at least two `SdNotify()` attempts within one
+watchdog notification interval when health checks pass. This includes the
+bounded wait for each attempt and any retry backoff between attempts. The exact
+constants should be finalized during implementation review, but the invariant is
+that one slow notification attempt must not consume the entire watchdog
```
googs1025, line 325: This is the main invariant for timeout values. Exact constants can be decided in implementation, but one slow notify call should not use the whole watchdog interval.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,664 @@
...
+Inspect kubelet logs for structured watchdog diagnostics with
+`operation=watchdog_health_check` or `operation=watchdog_notify`. Compare those
+logs with systemd journal entries and kubelet restart timestamps.
+
+If systemd reports that kubelet was terminated by the watchdog, inspect the
+systemd journal around the failure timestamp for the watchdog termination reason
+and signal. On systems configured to preserve cores or process dumps, inspect
+the kubelet core or stack evidence to determine whether the process was
+globally blocked, blocked in logging or journald paths, blocked in systemd
+notification, or blocked in unrelated kubelet work.
```
googs1025, line 616: This is for the process-freeze question. I documented checking journal/core/stack evidence, but kept automatic dump generation outside this KEP scope.

googs1025 (commented):
Thanks for the review.  @SergeyKanzhelev  Based on the feedback, I simplified the KEP scope to focus on diagnostic logging only.

  This revision:

  - adds default-visible structured logs for health checker errors and `SdNotify()` errors;
  - adds a structured summary when the existing notification retries are exhausted;
  - preserves the existing watchdog execution, timing, retry count, backoff, socket behavior, and non-Linux behavior;
  - removes the proposed feature gate;
  - defers timeout, cancellation, goroutine/single-flight handling, socket adapters, self-healing, metrics, events, NPD, and automatic thread dumps to future work;
  - updates the user stories, risks, PRR, troubleshooting guidance, and test plan accordingly.

  The intent is to keep this change small and low risk while providing useful failure evidence. Blocking checker or notification calls remain outside the scope of this KEP and can be addressed in a separate follow-up design.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,347 @@
...
+Upgrading kubelet adds failure diagnostics and retry summaries. Successful
+heartbeat behavior is unchanged. Downgrading kubelet removes the additional
+diagnostics. No state or configuration migration is involved.
+
+### Version Skew Strategy
+
+The change is kubelet-local and requires no coordination with other components.
+Nodes may run versions with and without the additional diagnostics.
+
+## Production Readiness Review Questionnaire
```
jpbetz, line 262: I'll review PRR. Please add `keps/prod-readiness/sig-node/6247.yaml` with `approver: "@jpbetz"` for alpha.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,347 @@
...
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+There is no independent feature gate. It is enabled by upgrading kubelet and
+removed by downgrading or reverting kubelet.
```
jpbetz, line 269: Alpha features must be disabled by default. This is a project wide requirement. Please add a feature gate.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,347 @@
...
+
+###### Can the feature be disabled once it has been enabled?
+
+Binary rollback removes the behavior; no persisted state or configuration
+migration is needed.
+
+###### What metrics should inform a rollback?
+
+No new metrics are proposed. Use existing kubelet restart signals, kubelet logs,
+and systemd journal entries.
```
jpbetz, line 284: PRR requires answers to these questions from the template: "What happens if we reenable the feature if it was previously rolled back?" and "Are there any tests for feature enablement/disablement?"

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
```diff
@@ -0,0 +1,347 @@
...
+watchdog timing or Kubernetes operation latency changes. Extra work is limited
+to failure-path log records and one retry summary.
+
+### Troubleshooting
+
+Inspect kubelet logs and the systemd journal around the restart. If no structured
+failure record exists, inspect available core dumps or process-stack evidence;
+the failure may have occurred in a blocking operation outside this KEP. High
+latency logging storage may delay or lose records, and repeated failures may
+still produce noisy logs despite rate limiting.
```
jpbetz, line 317: Please copy in the questions from the KEP template for this PRR section.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. I narrowed this KEP to diagnostic logging only. Blocking checker or notify calls are now out of scope, and I will cover bounded execution and watchdog budget in a separate KEP. Is this smaller scope okay for this KEP?

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. I moved the go-systemd cancellation and library exploration out of this KEP. This KEP only adds diagnostic logs. I will start a separate discussion before any library change. Is this scope okay?

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. I kept self-healing and dynamic disablement out of scope because they can hide a real kubelet health issue. This KEP reports the failure only. I can explore this in a follow-up KEP if needed.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. Added keps/prod-readiness/sig-node/6247.yaml with @jpbetz as the Alpha PRR approver in 062d332.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. Added the disabled-by-default Alpha KubeletWatchdogDiagnostics feature gate in the KEP and kep.yaml in 062d332.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. Added both template questions and answers for reenabling after rollback and feature enablement or disablement tests in 062d332.

keps/sig-node/6247-kubelet-watchdog-diagnostics/README.md
googs1025: Thanks. Added all Scalability questions from the KEP template with answers for this KEP in 062d332.

googs1025:
Hi @jpbetz and @SergeyKanzhelev, I addressed the review comments in commit 062d332.

The KEP now includes the disabled-by-default Alpha feature gate, PRR approver file, required PRR questions, and Scalability template answers. I also clarified the earlier scope discussion in the related threads.

When you have time, could you please take another look? Thank you.

jpbetz (approved):
/approve
For PRR

googs1025:
@SergeyKanzhelev /PTAL

PR #6253, opened 2026-07-20
KEP-6252: Hugepages Reservation

Tal-or:
- One-line PR description: Adding new KEP for supporting hugepages in kubelet's `--system-reserved` and `--kube-reserved `flags

- Issue link: https://github.com/kubernetes/enhancements/issues/6252

- Other comments:  See also https://github.com/kubernetes/kubernetes/issues/140544  and prior implementation https://github.com/kubernetes/kubernetes/pull/124357

/sig node

### DIFF
--- a/keps/sig-node/6252-system-reserved-hugepages/README.md
+++ b/keps/sig-node/6252-system-reserved-hugepages/README.md
@@ -1,0 +1,428 @@
+# KEP-6252: Hugepages Reservation
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: OVS-DPDK](#story-1-ovs-dpdk)
+    - [Story 2: Other system daemons](#story-2-other-system-daemons)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Flag Parsing](#flag-parsing)
+  - [Allocatable Computation](#allocatable-computation)
+  - [Memory Manager Integration](#memory-manager-integration)
+  - [Feature Gate](#feature-gate)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes extending kubelet's `--system-reserved` and `--kube-reserved`
+flags to accept hugepages resources (e.g., `hugepages-2Mi=512Mi`). When set,
+the specified hugepages are subtracted from the node's `Allocatable` capacity,
+preventing kubelet from offering hugepages that are already consumed by system
+daemons to pods.
+
+## Motivation
+
+System daemons such as OVS-DPDK allocate hugepages at startup for their own use
+(see [OVS-DPDK hugepages setup](https://docs.openvswitch.org/en/latest/intro/install/dpdk/#setup-hugepages)).
+When this happens, kubelet is unaware that those hugepages are taken and
+continues to report them as `Allocatable`. This leads to pods being scheduled
+with hugepages requests that cannot actually be fulfilled, causing allocation
+failures at runtime.
+
+Today, `--system-reserved` and `--kube-reserved` only accept `cpu`, `memory`,
+`pid`, and `ephemeral-storage`. There is no supported way to tell kubelet to
+exclude system-consumed hugepages from the allocatable pool.
+
+See also [kubernetes/kubernetes#140544](https://github.com/kubernetes/kubernetes/issues/140544)
+and [kubernetes/kubernetes#124357](https://github.com/kubernetes/kubernetes/pull/124357),
+which implemented this feature but went stale due to lack of a concrete use case
+at the time.
+
+### Goals
+
+- Allow `hugepages-<size>` resource types in `--system-reserved` and
+  `--kube-reserved` kubelet flags.
+- Subtract reserved hugepages from node `Allocatable` so the scheduler does not
+  over-commit hugepages that are consumed by system processes.
+- Enable `--reserved-memory` to work with hugepages end-to-end. The memory
+  manager's validation requires that `--reserved-memory` totals match
+  `system-reserved + kube-reserved + eviction-threshold` for each resource type.
+  Once hugepages are accepted in `--system-reserved` / `--kube-reserved`, this
+  validation will naturally pass.
+
+### Non-Goals
+
+- Changes to the Kubernetes scheduler.
+- Changes to the hugepages resource model (sizes, naming conventions).
+- Hugepages support in eviction thresholds.
+
+## Proposal
+
+Extend the kubelet flag validation to accept `hugepages-<size>` keys in
+`--system-reserved` and `--kube-reserved`, gated behind a new
+`SystemReservedHugepages` feature gate.
+
+The existing `Allocatable` computation in kubelet already iterates over all
+resource types in node capacity and subtracts matching entries from
+`system-reserved` and `kube-reserved`. The only change needed is to allow
+hugepages keys in the flag validation.
+
+### User Stories
+
+#### Story 1: OVS-DPDK
+
+As a cluster administrator running OVS-DPDK for high-performance networking,
+I configure OVS to use DPDK which requires allocating hugepages for packet
+buffers. I want to reserve those hugepages via `--system-reserved` so that
+kubelet does not offer them to pods, avoiding runtime allocation failures.
+
+Example:
+```
+--system-reserved=cpu=500m,memory=1Gi,hugepages-1Gi=2Gi
+```
+
+#### Story 2: Other system daemons
+
+As a cluster administrator running system daemons that require hugepages,
+I need to reserve those hugepages at the kubelet level to prevent resource
+contention with pods.
+
+### Notes/Constraints/Caveats (Optional)
+
+The `--reserved-memory` flag already shows hugepages in its help text example
+(`--reserved-memory 0:memory=1Gi,hugepages-1M=2Gi`), but in practice this
+cannot be used for hugepages today because the memory manager's validation
+requires matching values in `system-reserved + kube-reserved +
+eviction-threshold`, which don't accept hugepages. This KEP unblocks that
+path as a side effect.
+
+### Risks and Mitigations
+
+**Risk:** Misconfiguration — an administrator reserves more hugepages than
+available on the node.
+**Mitigation:** Kubelet already validates that reservations do not exceed node
+capacity via `validateNodeAllocatable()`. This validation applies to all
+resource types including hugepages once they are accepted.
+
+**Risk:** Feature interaction with Memory Manager — reserved hugepages must be
+consistent with `--reserved-memory` when the Memory Manager is enabled.
+**Mitigation:** The existing `validateReservedMemory()` check enforces this
+consistency. No additional validation is needed.
+
+## Design Details
+
+### Flag Parsing
+
+`--system-reserved` and `--kube-reserved` use `MapStringString` and accept
+arbitrary key-value pairs. The values are parsed by `parseResourceList()` in
+`cmd/kubelet/app/server.go`, which has a switch case that only accepts `cpu`,
+`memory`, `ephemeral-storage`, and `pid`. Any other resource type is rejected
+with `"cannot reserve %q resource"`. This switch case needs to be expanded to
+also accept `hugepages-<size>` keys when the `SystemReservedHugepages` feature
+gate is enabled.
+
+### Allocatable Computation
+
+`GetNodeAllocatableReservation()` in `pkg/kubelet/cm/node_container_manager_linux.go`
+iterates over all resource types in node capacity and sums `SystemReserved[k] +
+KubeReserved[k] + evictionReservation[k]`. The result is subtracted from
+`Capacity` to compute `Allocatable` on the node status. Once hugepages are
+accepted in the flags, they will be included in this subtraction automatically
+with no code changes needed in this path.
+
+### Memory Manager Integration
+
+`validateReservedMemory()` in `pkg/kubelet/cm/memorymanager/memory_manager.go`
+checks that `--reserved-memory` totals equal `nodeAllocatableReservation` for
+each memory-type resource. Once hugepages flow into `system-reserved` /
+`kube-reserved`, the hugepages portion of `nodeAllocatableReservation` will be
+non-zero, and the validation will pass for matching `--reserved-memory` values.
+
+### Feature Gate
+
+A new feature gate `SystemReservedHugepages` controls this behavior:
+- **Disabled (default in alpha):** Hugepages keys in `--system-reserved` and
+  `--kube-reserved` are rejected, preserving current behavior.
+- **Enabled:** Hugepages keys are accepted and processed.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes
+necessary to implement this enhancement.
+
+##### Prerequisite testing updates
+
+None.
+
+##### Unit tests
+
+- Extend `TestNodeAllocatableReservationForScheduling` in
+  `pkg/kubelet/cm/node_container_manager_linux_test.go` to include hugepages
+  resources in `SystemReserved` and `KubeReserved` and verify the allocatable
+  computation.
+- Add validation tests for flag parsing with hugepages keys, both with the
+  feature gate enabled and disabled.
+
+- `k8s.io/kubernetes/pkg/kubelet/cm`: `TBD`
+
+##### Integration tests
+
+- TBD
+
+##### e2e tests
+
+- TBD
+
+### Graduation Criteria
+
+#### Alpha
+
+- Feature implemented behind the `SystemReservedHugepages` feature gate.
+- Unit tests covering flag validation and allocatable computation.
+- Initial e2e tests completed and enabled.
+
+#### Beta
+
+- Gather feedback from developers and users.
+- Feature gate enabled by default.
+- Extend e2e test coverage.
+
+#### GA
+
+- Feature gate locked to enabled.
+- At least two releases since beta.
+- Real-world usage confirmed.
+
+### Upgrade / Downgrade Strategy
+
+No special upgrade steps required. The feature is opt-in via the
+`--system-reserved` / `--kube-reserved` flags. Existing clusters that do not
+set hugepages in these flags are unaffected.
+
+On downgrade, if hugepages were configured in `--system-reserved` or
+`--kube-reserved`, the kubelet will reject the configuration and fail to start.
+Administrators must remove hugepages entries from these flags before downgrading.
+
+### Version Skew Strategy
+
+This feature is kubelet-only and does not involve coordination with the control
+plane. The kubelet independently computes `Allocatable` and reports it on the
+node status. No version skew concerns exist.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `SystemReservedHugepages`
+  - Components depending on the feature gate: kubelet
+
+###### Does enabling the feature change any default behavior?
+
+No. The feature only takes effect when an administrator explicitly adds
+hugepages entries to `--system-reserved` or `--kube-reserved`. Without those
+entries, behavior is identical to today.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Disable the feature gate and remove hugepages entries from
+`--system-reserved` / `--kube-reserved`, then restart kubelet. The node's
+`Allocatable` will return to its previous values.
+
+###### What happens if we reenable the feature if it was previously rolled back?
+
+Hugepages entries in `--system-reserved` / `--kube-reserved` will be accepted
+again and subtracted from `Allocatable`. No state is persisted beyond the flag
+values.
+
+###### Are there any tests for feature enablement/disablement?
+
+Unit tests will verify that hugepages keys are rejected when the feature gate
+is disabled and accepted when enabled.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+A rollout cannot impact running workloads because the feature only changes
+`Allocatable` reporting. Already-running pods that have hugepages allocated will
+continue to function. The change only affects future scheduling decisions.
+
+A rollback (disabling the feature gate) while hugepages entries are still in
+`--system-reserved` will cause kubelet to reject the configuration on restart.
+Administrators must remove the entries first.
+
+###### What specific metrics should inform a rollback?
+
+If pods fail to schedule due to insufficient hugepages despite the node having
+enough total hugepages to accommodate both system daemons and pod requests,
+the reservation values may be misconfigured. Verify the `--system-reserved` /
+`--kube-reserved` hugepages values match the actual system daemon consumption.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Will be tested during beta.
+
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+Check the node's `status.allocatable` for hugepages resources and compare with
+`status.capacity`. If they differ, hugepages reservation is active.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] API .status
+  - Other field: `node.status.allocatable[hugepages-<size>]` reflects
+    `capacity - system-reserved - kube-reserved`.
+
+###### What are the reasonable SLOs (Service Level Objectives) for the enhancement?
+
+No new SLOs. This feature affects node status reporting, which is covered by
+existing kubelet SLOs.
+
+###### What are the SLIs (Service Level Indicators) an operator can use to determine the health of the service?
+
+- [x] Other (treat as last resort)
+  - Details: Compare `node.status.capacity[hugepages-<size>]` with
+    `node.status.allocatable[hugepages-<size>]`. The difference should equal
+    the configured reservation.
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+No. The existing node capacity and allocatable fields provide sufficient
+observability.
+
+### Dependencies
+
+###### Does this feature depend on any specific services running in the cluster?
+
+No.
+
+### Scalability
+
+###### Will enabling / using this feature result in any new API calls?
+
+No.
+
+###### Will enabling / using this feature result in introducing new API types?
+
+No.
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+No. The node status already includes hugepages in capacity and allocatable.
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+No.
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any components?
+
+No.
+
+###### Can enabling / using this feature result in resource exhaustion of some node resources (PIDs, sockets, inodes, etc.)?
+
+No.
+
+### Troubleshooting
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+Kubelet computes `Allocatable` locally, but it needs the API server to update
+the node status. If the API server is unavailable, the node status will not
+reflect the reserved hugepages until connectivity is restored. This is the same
+behavior as existing resource reservations (cpu, memory).
+
+###### What are other known failure modes?
+
+- Misconfigured hugepages reservation exceeding node capacity
+  - Detection: Kubelet will fail to start with a validation error.
+  - Mitigations: Fix the `--system-reserved` / `--kube-reserved` values.
+  - Diagnostics: Kubelet logs will contain the validation error message.
+  - Testing: Unit tests cover this case.
+
+###### What steps should be taken if SLOs are not being met to determine the problem?
+
+This feature does not introduce new SLOs. It modifies a static value on the
+node status at kubelet startup. If the node's `Allocatable` hugepages value
+is incorrect, check the `--system-reserved` and `--kube-reserved` flag values.
+
+## Implementation History
+
+- 2024-04-18: Prior implementation PR [kubernetes/kubernetes#124357](https://github.com/kubernetes/kubernetes/pull/124357) opened.
+- 2024-10-15: PR closed as stale.
+- 2026-07-20: KEP created.
+
+## Drawbacks
+
+This feature extends the accepted resource types for existing flags, which
+increases the surface area for misconfiguration. However, the existing
+validation mechanisms (`validateNodeAllocatable`) already handle this.
+
+## Alternatives
+
+N/A
--- a/keps/sig-node/6252-system-reserved-hugepages/kep.yaml
+++ b/keps/sig-node/6252-system-reserved-hugepages/kep.yaml
@@ -1,0 +1,33 @@
+title: Hugepages Reservation
+kep-number: 6252
+authors:
+  - "@Tal-or"
+owning-sig: sig-node
+participating-sigs: []
+status: provisional
+creation-date: 2026-07-20
+reviewers:
+  - "@ffromani"
+approvers:
+  - TBD
+
+see-also:
+  - "/keps/sig-node/1769-memory-manager"
+  - "/keps/sig-node/1539-hugepages"
+
+stage: alpha
+
+latest-milestone: "v1.38"
+
+milestone:
+  alpha: "v1.38"
+  beta: ""
+  stable: ""
+
+feature-gates:
+  - name: SystemReservedHugepages
+    components:
+      - kubelet
+disable-supported: true
+
+metrics: []


keps/sig-node/6252-system-reserved-hugepages/kep.yaml
```diff
@@ -0,0 +1,35 @@
+title: Hugepages Reservation
```
ffromani, line 1: you need to add a PRR file, example: https://github.com/kubernetes/enhancements/pull/6353/changes#diff-1b1d1fe5655237e41a49e5fc7e80c76a9e61ea4ff44cba666bb40dc10ada4685

ffromani: also, please make sure to use the very last KEP template (https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template)

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
```diff
@@ -0,0 +1,35 @@
...
+kep-number: 6252
+authors:
+  - "@Tal-or"
+owning-sig: sig-node
+participating-sigs: []
+status: provisional
+creation-date: 2026-07-20
+reviewers:
+  - "@ffromani"
```
ffromani: sig-node identified other reviewers that we will soon join the review

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - "/keps/sig-node/1539-hugepages"
+  - "/keps/sig-node/5894-node-system-partition"
+
+stage: alpha
+
+latest-milestone: "v1.38"
+
+milestone:
+  alpha: "v1.38"
+  beta: ""
```
ffromani: please set tentative releases. GA should usually be +2 releases from beta. E.g.:
alpha 1.38
beta 1.39
stable 1.41

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes extending kubelet's `--system-reserved` and `--kube-reserved`
+flags to accept hugepages resources (e.g., `hugepages-2Mi=512Mi`). When set,
+the specified hugepages are subtracted from the node's `Allocatable` capacity,
+preventing kubelet from offering hugepages that are already consumed by system
```
ffromani, line 76: You probably discuss in the rest of the document, but let's make sure to point if and how hugepages overlap with memory and treat that accordingly. E.g. linux kernel's `memory_hugetlb_accounting`

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+### Non-Goals
+
+- Changes to the Kubernetes scheduler.
+- Hugepages support in eviction thresholds.
+- Windows kubelet support. This work is Linux-only.
+
+## Proposal
+
+Extend `--system-reserved` and `--kube-reserved` to accept `hugepages-<size>`
+keys, gated by `SystemReservedHugepages`. Kubelet already subtracts matching
```
ffromani: nit: clarity: "gated by the `SystemReservedHugepages` Feature Gate`"

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+```
+--system-reserved=cpu=500m,memory=1Gi,hugepages-2Mi=512Mi
+--reserved-memory 0:hugepages-2Mi=256Mi --reserved-memory 1:hugepages-2Mi=256Mi
+```
+
+#### Story 2: Other system daemons
+
+As a cluster administrator running system daemons that require hugepages,
+I need to reserve those hugepages at the kubelet level to prevent resource
+contention with pods.
```
ffromani, line 151: I believe the story 1 is already enough for such a simple KEP, but do we have other more concrete use cases?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+against it.
+
+When [KEP-5894 Node System Partition](/keps/sig-node/5894-node-system-partition)
+is enabled, `--system-reserved` and `--kube-reserved` continue to cover host
+processes (kubelet, container runtime, OS services) - not system partition
+Pods. The hugepages reserved through this KEP are subtracted from the
+node-wide `Allocatable`. This KEP does not add any configuration under
+`systemPartition`, since that section deals with system Pods whereas this
+KEP reserves hugepages for host services. This KEP does not depend on or
+modify KEP-5894.
```
ffromani: could you add an example using the system partition? I'm not sure I understood correctly: will each partition need to set hugepages reserved? only the system partition? only  the non-system partition?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+consistent with `--reserved-memory` when the Memory Manager is enabled.
+**Mitigation:** The existing `validateReservedMemory()` check enforces this
+consistency. No additional validation is needed.
+
+**Risk:** Reviewers may treat this as a bug fix because hugepages are already
+GA, rather than as a feature.
+**Mitigation:** Kubelet flags are API, so this KEP treats the change as a
+feature. `SystemReservedHugepages` is rollout safety: when the gate is
+disabled, hugepages keys in `--system-reserved` and `--kube-reserved` are
+rejected, which is current behavior.
```
ffromani: I think this is more a meta-risk. We should track risks of this work assuming we do it, not track the risks of doing this work at all (which seems indeed a meta-risk)

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+gate is disabled, those keys continue to be rejected so existing clusters
+keep current behavior.
+
+Hugepage quantities should be divisible by the page size, matching pod
+admission (`IsHugePageResourceValueDivisible`). Non-divisible values should
+be rejected at parse time.
+
+The `--system-reserved` and `--kube-reserved` flag help text in
+`cmd/kubelet/app/options/options.go` should list hugepages as an accepted
+resource type.
```
ffromani, line 215: question: do we need to do any cross-validation among `--system-reserved` and `--kube-reserved` or are these independent from each other?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+
+`hardEvictionReservation` only handles `memory` and `ephemeral-storage`, so
+the eviction term for hugepages is always zero. For hugepages, reservation
+and absolute allocatable therefore differ from capacity by the same
+system-reserved + kube-reserved amount.
+
+`CapacityFromMachineInfo()` in `pkg/kubelet/cadvisor/util.go` already reports
+node-wide hugepage capacity. Once hugepages are present in the parsed
+`NodeConfig` reserved lists, they are included in both computations with no
+further changes on those paths and no scheduler changes.
```
ffromani: a call graph or a way to quickly grasp how these functions interact together would help

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+  and `GetNodeAllocatableAbsolute` include hugepages in `SystemReserved` and
+  `KubeReserved`. QoS manager tests: Guaranteed tier gets limited hugepages;
+  Burstable and BestEffort stay unbounded.
+- `k8s.io/kubernetes/pkg/kubelet/cm/memorymanager`: `validateReservedMemory`
+  with matching hugepage totals across `--reserved-memory` and
+  system/kube-reserved.
+
+##### Integration tests
+
+None planned at this time.
```
ffromani: ...because e2e test will cover the flows, I reckon (worth a mention).

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+
+##### e2e tests
+
+Extend `test/e2e_node/node_container_manager_test.go` as in
+[kubernetes/kubernetes#124357](https://github.com/kubernetes/kubernetes/pull/124357):
```
ffromani: (here and anywhere else in the doc) please quickly summarize the needed changes vs refer to a PR. The KEP must be self-sufficient and timeless; PRs come, go and get updated.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### Alpha
+
+- Feature implemented behind the `SystemReservedHugepages` feature gate.
+- Unit tests covering flag validation and allocatable computation.
+- Initial e2e tests completed and enabled.
+
+#### Beta
+
+- Gather feedback from developers and users.
```
ffromani: how? poll? lazy consensus? (no issues reported?)

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+- Feature gate enabled by default.
+- Extend e2e test coverage.
```
ffromani: fine! to cover what?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+
+#### GA
+
+- Feature gate locked to enabled.
+- At least two releases since beta.
```
ffromani: without no major bugs? any other criteria worth mentioning?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+- Real-world usage confirmed.
```
ffromani: how? can we pin a bit more precisely?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+Unit tests will verify that hugepages keys are rejected when the feature gate
+is disabled and accepted when enabled.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+A rollout changes future scheduling through `Allocatable`. Already-running
+pods that have hugepages allocated keep those mappings. The change does not
+evict or rewrite existing hugepage allocations.
```
ffromani: what if the node is pretty much at capacity, with all hugepages assigned to pods, and setting allocatable conflicts with the allocation?
IOW, say we have 10 hugepages, all allocated to pods, the admin somehow set 2 hugepages as reserved and restarts the kubelet. What should happen?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+###### What specific metrics should inform a rollback?
+
+If pods fail to schedule due to insufficient hugepages despite the node having
+enough total hugepages to accommodate both system daemons and pod requests,
+the reservation values may be misconfigured. Verify the `--system-reserved` /
+`--kube-reserved` hugepages values match the actual system daemon consumption.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+Will be tested during beta.
```
ffromani: any chance to anticipate to alpha? deferring the final call to PRR folks though

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Run system daemons as pods.** Give those workloads ordinary pod hugepage
+requests instead of system-reserved. This is not always possible: daemons
+such as OVS-DPDK are often started on the host, outside the pod lifecycle.
+
+**Placeholder pods.** Deploy a pod that requests the exact hugepage
+amount the system daemon consumes. This requires no kubelet changes and the
+scheduler accounts for the reservation correctly. It is however a hack: the
+pod serves no workload purpose, must be kept in sync with the daemon's actual
+consumption, and adds operational overhead.
```
ffromani, line 626: plus, should probably started first (or among the first) but there's no real direct/explicit control over the ordering on which kubelet restore pods

ffromani (commented):
initial pass
quite some comments, but all minors and a chunk of them are about the paperwork

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Not ones that I'm aware of unfortunately.

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
Tal-or: Done for both. Thanks!

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
Tal-or: Changed to correct reviewers.

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
Tal-or: Done.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: @ffromani according to my understanding `memory_hugetlb_accounting` is not enabled by default for k8s and also it will break the existing accounting mechanism since the runtime is setting separate cgroups controllers for memory and hugepages (based on the resources request) .

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Done.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: The KEP says:
"This KEP does not add any configuration under `systemPartition`"
An example to how the new partition for system pods looks like:
```yaml
apiVersion: kubelet.config.k8s.io/v1beta1
kind: KubeletConfiguration
systemPartition:
  memoryLimit: "4Gi"
  cpuset: "0-3"
  namespaces:
    - kube-system
```
and it is going to remain this way, i.e. no support for hugepages is planned for this specific partition.

> will each partition need to set hugepages reserved? only the system partition? only the non-system partition?

Only the non-system partition.
KEP was updated to reflect this info clearer.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Ack. removing this section

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: They are independent.

The only cross-validation is that `kube-reserved` + `system-reserved` + `eviction` <= node capacity.
This is done in https://github.com/kubernetes/kubernetes/blob/1d54bff3606e44a9d5178698816fc8af340e85a9/pkg/kubelet/cm/node_container_manager_linux.go#L317-L335 
for all resources.
Once hugepages are added to  `kube-reserved` and `system-reserved` it will perform the same validation with no extra changes.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Done.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Added a  mermaid call graph.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: inline the needed changes.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: I'll change to alpha

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Indeed.

keps/sig-node/6252-system-reserved-hugepages/README.md
ffromani: matches my recollection, but is worth mentioning it explicitly, even with a couple lines.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Nice catch, updated in the KEP, thanks

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Thanks! Added few lines under the "Cgroup Enforment" section.

keps/prod-readiness/sig-node/6252.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 6252
+alpha:
+  approver: "@dchen1107"
```
jpbetz, line 3:
```suggestion
  approver: "@jpbetz"
```

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
```diff
@@ -0,0 +1,36 @@
+title: Hugepages Reservation
+kep-number: 6252
+authors:
+  - "@Tal-or"
+owning-sig: sig-node
+participating-sigs: []
+status: provisional
```
jpbetz, line 7: PRR approval is only checked once the KEP is `implementable`. Is the status meant to flip in this PR, or in a follow-up before the Sep 22 PRR freeze?

jpbetz (commented):
I'm the PRR approver assigned to this KEP for 1.38 on the release board, so swapping the approver in the prod-readiness file. Questionnaire review to follow.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,636 @@
...
+A rollback that disables the feature gate while hugepages entries are still
+in `--system-reserved` or `--kube-reserved` causes kubelet to reject the
+configuration on restart. Administrators must remove the entries first.
+
+###### What specific metrics should inform a rollback?
+
+If pods fail to schedule due to insufficient hugepages despite the node having
+enough total hugepages to accommodate both system daemons and pod requests,
+the reservation values may be misconfigured. Verify the `--system-reserved` /
+`--kube-reserved` hugepages values match the actual system daemon consumption.
```
jpbetz, line 486: This is a good point. Would you add it to the known failure modes section below?

Could you also add a paragraph here about metrics?  The answer here is valuable but not exactly answering the question asked here.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,636 @@
...
+
+###### How does this feature react if the API server and/or etcd is unavailable?
+
+Kubelet computes `Allocatable` locally, but it needs the API server to update
+the node status. If the API server is unavailable, the node status will not
+reflect the reserved hugepages until connectivity is restored. This is the same
+behavior as existing resource reservations (cpu, memory).
+
+###### What are other known failure modes?
+
```
jpbetz: The failure mode described at lines 467-475, running pods holding more hugepages than the new allocatable so the `kubepods` hugetlb limit cannot be applied, is missing from this list. It is the one that can persist indefinitely since there is no hugepage eviction, so please add it with the four bullets. In particular: how an operator detects it beyond the once-a-minute warning event, and whether e2e_node covers enforcement failure with running hugepage pods.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,636 @@
...
+node status. No version skew concerns exist.
+
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `SystemReservedHugepages`
```
jpbetz, line 434: Does kubelet usually require a feature gate for capabilities turned on via command line flags?

I ask because for the control plane components we typically do not require a feature gate when an enhancement is enabled by a flag. The flag is set in the same way as the feature gate and by the same user (cluster admin/operator) and so making them set two flags is needless.

jpbetz:
@Tal-or Quick reminder that I've submitted PRR feedback. I can review again once addressed. Thanks!

keps/prod-readiness/sig-node/6252.yaml
Tal-or: Done.

keps/sig-node/6252-system-reserved-hugepages/kep.yaml
Tal-or: Changed to `implementable`, `provisional` was there just for the initial state of the KEP

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: I honestly don't know what is the usual standard unfortunately.

 >  The flag is set in the same way as the feature gate and by the same user (cluster admin/operator) and so making them set two flags is needless.

That makes complete sense. would you like me to remove the feature gate then?

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Moved to the right section and added few sentences about metrics.

Tal-or (commented):
@jpbetz sorry for the late response, I went for a long weekend.

Addressed all your comments PTAL when possible.
Thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+
+- Allow `hugepages-<size>` resource types in `--system-reserved` and
+  `--kube-reserved` kubelet flags.
+- Subtract reserved hugepages from node `Allocatable` so the scheduler does not
+  over-commit hugepages that are consumed by system processes.
+- Enable `--reserved-memory` to work with hugepages end-to-end. The memory
+  manager's validation requires that `--reserved-memory` totals match
+  `system-reserved + kube-reserved + eviction-threshold` for each resource type.
+  Once hugepages are accepted in `--system-reserved` / `--kube-reserved`, this
+  validation will naturally pass.
```
swatisehgal: Do we have to add code to handle hugepage validation as part of the work this cycle? The text here seems to read in a way that once the hugepages are passed into system-reserved + kube-reserved the vidation will all be taken care of. Perhaps a rephrase might help here.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+- Document the two-flag workflow: node-wide hugepage totals in
+  `--system-reserved` / `--kube-reserved`, and the per-NUMA split in
+  `--reserved-memory`.
+
+### Non-Goals
+
+- Changes to the Kubernetes scheduler.
+- Hugepages support in eviction thresholds.
```
swatisehgal, line 116: Why are we choosing to exclude this? Is this completely out of scope of something we plan to handle during Beta phase?

swatisehgal: We have done some work in this space: https://github.com/kubernetes/kubernetes/pull/138127 and I am curious how this proposal would work with the existing hugepage aware eviction mechnism.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+(`--reserved-memory 0:memory=1Gi,hugepages-1M=2Gi`), but in practice this
+cannot be used for hugepages today because the memory manager's validation
+requires matching values in `system-reserved + kube-reserved +
+eviction-threshold`, which don't accept hugepages. This KEP unblocks that
+path.
+
+The administrator must keep the per-NUMA `--reserved-memory` totals equal to
+`system-reserved + kube-reserved` for each resource type, just as with
+ordinary memory today. Eviction thresholds never include hugepages, so they
+do not contribute to that equality for hugepage resources.
```
swatisehgal: This might not be accurate, please refer to https://github.com/kubernetes/kubernetes/pull/138127 and determine if this KEP needs to account for eviction thresholds for hugepgaes.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+matching the existing QoS cgroup design for other resources.
+
+### Memory Manager Integration
+
+Per-NUMA hugepage reservation through `--reserved-memory` is already
+implemented end-to-end; this KEP unblocks it. `validateReservedMemory()`
+requires `sum(--reserved-memory)` per resource type to equal
+`system-reserved + kube-reserved + eviction-threshold`. Once hugepages are
+accepted in `--system-reserved` / `--kube-reserved`, matching
+`--reserved-memory` values pass that check.
```
swatisehgal: I would try to phrase this part carefully, I don't think it is right to say that the --reserved-memory works end to end considering hugepage could not be indicated in system-reserved and kube-reserved, perhaps you mean that the plumbing exists.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+  never pre-allocated
+  - Detection: `validateNodeAllocatable()` fails, or pods and system daemons
+    cannot obtain hugepages at runtime.
+  - Mitigations: Pre-allocate the kernel pool via sysfs so capacity covers
+    reserved plus pod demand.
+  - Diagnostics: Compare sysfs hugepage counts with `node.status.capacity`
+    and the reserved flag values.
+  - Testing: e2e_node configures host hugepages before applying reserved
+    flags.
+- Reservation value exceeds actual system daemon hugepage consumption
```
swatisehgal: I think we explicitly state in the KEP that these flags represent an administrator-declared reservation rather than accounting of actual hugepage consumption by host processes? Kubelet cannot infer how much hugepage memory OVS-DPDK or another host daemon is actually consuming, so the configured value remains an operational contract.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+`kubepods` via `enforceNodeAllocatableCgroups()`, and the system-reserved /
+kube-reserved cgroups via `enforceExistingCgroup()` - goes through this
+function. Those paths require no changes once hugepages are in the parsed
+`ResourceList`.
+
+The only issue is the QoS cgroup manager
+(`pkg/kubelet/cm/qos_container_manager_linux.go`), which runs `UpdateCgroups()`
+every minute and currently sets hugepage limits to unbounded on all QoS tiers,
+including Guaranteed. Because the Guaranteed tier maps to the `kubepods` root,
+this overwrites the limits that node-allocatable enforcement applied.
```
swatisehgal, line 315: This interaction seems important because this is what can overwrite the node-allocatable hugepage limit after initial enforcement. Could the e2e test explicitly allow the periodic QoS update to run (or otherwise exercise UpdateCgroups) and verify that the Guaranteed/kubepods hugetlb limit remains equal to the node allocatable value? Otherwise a regression in the periodic reconciliation path may not be caught.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+##### Unit tests
+
+- `k8s.io/kubernetes/cmd/kubelet/app`: `parseResourceList` with
+  `hugepages-<size>` keys, feature gate enabled and disabled, and quantities
+  that are not divisible by the page size.
+- `k8s.io/kubernetes/pkg/kubelet/cm`: extend
+  `TestNodeAllocatableReservationForScheduling` in
+  `node_container_manager_linux_test.go` so `GetNodeAllocatableReservation`
+  and `GetNodeAllocatableAbsolute` include hugepages in `SystemReserved` and
+  `KubeReserved`. QoS manager tests: Guaranteed tier gets limited hugepages;
```
swatisehgal, line 378: Let's ensure (during implementation) that we are able to handle the cases where only one ( system-reserved or kube-reserved) is specified.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+
+##### Integration tests
+
+None planned at this time, because e2e test will cover the flows.
+
+##### e2e tests
+
+Extend `test/e2e_node/node_container_manager_test.go`:
+
+- Configure 2Mi hugepages on the host.
```
swatisehgal: I belive we would need to validate both hugepages-2Mi and hugepages-1Gi. We need to ensure that reservation and allocatable calculation are independently applied per page size.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+from `--system-reserved` and `--kube-reserved` first. If those keys remain
+while the gate is disabled or the kubelet build rejects them, kubelet will
+fail to start.
+
+### Version Skew Strategy
+
+This feature is kubelet-only and does not involve coordination with the control
+plane. The kubelet independently computes `Allocatable` and reports it on the
+node status. No version skew concerns exist.
+
```
swatisehgal, line 471: While there is no control-plane/API coordination required, the kubelet configuration itself is not forward-compatible: an older kubelet that doesn't recognize the feature will reject these resource keys. Perhaps @jpbetz could review this as part of the PRR review?

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+## Production Readiness Review Questionnaire
+
+### Feature Enablement and Rollback
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `SystemReservedHugepages`
+  - Components depending on the feature gate: kubelet
```
jpbetz: Answering the question in the thread on this line: PRR does not require the gate. The hugepages keys are the switch, so the gate is a second lock on the same knob set by the same operator. Two things decide whether to drop it: (1) without a gate this ships as GA behavior in 1.38 with no alpha period, so the SIG Node approvers should explicitly be okay with that, and the graduation criteria and kep.yaml stages would need to change to match; (2) the QoS manager cgroup change (next comment) would have to be scoped to nodes with a reservation, since nothing else would turn it off. Given the freeze timing my suggestion is to keep the gate for now; it is harmless and the questionnaire is already consistent with it. Either way is approvable from the PRR side.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+
+###### Does enabling the feature change any default behavior?
+
+No. The feature only takes effect when an administrator explicitly adds
+hugepages entries to `--system-reserved` or `--kube-reserved`. Without those
+entries, behavior is identical to today.
```
jpbetz, line 486: Can you confirm that the QoS manager will not change default behavior? It's not clear to me from the KEP. Please consider updating the relevant KEP to make this clear if QoS changes will be entirely isolated to when a hugepage reservation is used.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+retries every minute and emits `FailedNodeAllocatableEnforcement` warning
+events until pods release hugepages. Unlike memory, there is no hugepage
+eviction mechanism to drive usage down. Administrators should drain
+hugepage-consuming pods from the node before introducing or increasing
+hugepage reservations, the same recommendation that applies when reducing
+memory allocatable.
+
+A rollback that disables the feature gate while hugepages entries are still
+in `--system-reserved` or `--kube-reserved` causes kubelet to reject the
+configuration on restart. Administrators must remove the entries first.
```
jpbetz, line 524: Let's also specifically cover these cases:

Case 1:

- Admin sets `--system-reserved=hugepages-2Mi=512Mi` and `--enforce-node-allocatable=pods,system-reserved`.
- Kubelet starts, subtracts 512Mi from allocatable, and writes `hugetlb.2MB.max=512Mi` on the system-reserved cgroup.
- OVS runs in that cgroup and grows to 600Mi of hugepages.
- Kernel refuses the extra pages. OVS fails?
- (I don't see anything in the KEP handling this in a safe way, yet)

Case 2:

- OVS is already using 600Mi of hugepages.
- Admin adds `--system-reserved=hugepages-2Mi=512Mi` with `system-reserved` enforcement on, then starts the kubelet.
- Kubelet writes `hugetlb.2MB.max=512Mi` on the system-reserved cgroup.
- Kernel rejects the write (EBUSY?).
- The `reserved-cgroup` path treats that error as fatal? Potentially a crash-loop?
- (I don't see this handled in the KEP either)

Also, any similar cases you can think of would be valuable here.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,658 @@
...
+#### Deprecation
+
+N/A — this feature extends existing flags; no deprecation is planned.
+
+### Upgrade / Downgrade Strategy
+
+No special upgrade steps required. The feature is opt-in via the
+`--system-reserved` / `--kube-reserved` flags. Existing clusters that do not
+set hugepages in these flags are unaffected.
+
```
swatisehgal, line 451: What happens to existing workloads that are requesting hugepages when a node's kubelet is upgraded to a version that supports this feature?

In particular, if the node already has running pods consuming hugepages, does the new kubelet immediately apply the system-reserved / kube-reserved hugepage values to the kubepods hugetlb cgroup and reduce the effective limit available to existing workloads?

If the new limit is below the workloads' current hugepage usage, I assume the kernel will reject the limit reduction and kubelet will retry as described in the configuration-change case. Could we explicitly document the expected upgrade behavior and plan to add tests for this scenario?

swatisehgal (commented):
Thanks for your work on this, I think we need some more work. The key point from my POV are:

1. Analyse the impact of hugepage aware eviction.
2. Strengthen the Node Allocatable enforcement and its interaction with the QoS cgroup manager. Adding some examples would really help as well.
3. Determine and capture what happens to existing workloads requesting hugepages when kubelet is upgraded to a version with this feature.

swatisehgal (commented):
Thinking of some of the work from NUMA-aware scheduling point of view, I was wondering if we currently have per NUMA allocated hugepages captured through podresource API (I think the answer is No but I am not 100% sure). I wonder if this something we should think of perhaps in the next iteration of this KEP (or even as a separate KEP)?

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: > The text here seems to read in a way that once the hugepages are passed into system-reserved + kube-reserved the vidation will all be taken care of.

That's correct and that's because the validation function of memory-manager was written in a way that sums  `system-reserved` + `kube-reserved`  for memory and hugepages and compare them against what was set by the admin, see:
https://github.com/kubernetes/kubernetes/blob/4b19e2b0244b00b56d6604deab3ff32b1c7e5706/pkg/kubelet/cm/memorymanager/memory_manager.go#L445 

IOW, the logic is already threre - added a reference to the code in the doc

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: The reference is about refining the existing memory eviction mechanism to correctly account for memory dedicated to hugepages.

The eviction mechanism should not count hugepages within the eviction threshold because memory dedicated to hugepages is accounted in a separate kernel pool - it is not part of the general memory pool that eviction monitors.

>  I am curious how this proposal would work with the existing hugepage aware eviction mechnism.

There is no eviction mechanism for hugepages since hugepages in opposed to standard memory cannot overcommit  (even for burstable pods) 
meaning `requests` = `limits` when it comes to hugepages, so pod cannot suddenly consumes more hugepages than its allocated limit.

More over the eviction thresholds mechanism exist to solve general `memory.available` pressure - when the threshold is hit, kubelet starts terminating pods to protect critical system processes. evicting pods that are using hugepages won't solve that problem because hugepages are not count as part of `memory.available` (they have their own accounting system), so kubelet would simply kill the pod but the memory pressure will remain.

keps/sig-node/6252-system-reserved-hugepages/README.md
jpbetz: Yes.

If questions come up about this later, We can reference previous cases such as `--config-dir` (KEP 3983) that also are flag only.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: see:  https://github.com/kubernetes/enhancements/pull/6253/changes#r4073710203

Tal-or: elaborated about that in the `### Notes/Constraints/Caveats (Optional)` section

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Thanks. Changed to and/or to include individual cases.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Ack added.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: > is not forward-compatible

I think you meant to say is not backward-compatible. So older kubelet version refuses to start until its the reserved hugepages will be remove from its kubelet configuration.

I believe it make sense it's merely a validation, I wouldn't expect a different behavior in that sense.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Added. thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: It is also meantioned in the `### Upgrade / Downgrade Strategy` that kubelet will failed to start of those values are remain in the config while dowgrading kubelet version.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: rephrased PTAL. thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: I agree, this section was over engineered and dose not provide an actual value to the proposal. removing it.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: First, it should be explicitly stated by the admin, there is no hugepages default reservation.  

>If the new limit is below the workloads' current hugepage usage, I assume the kernel will reject the limit reduction and kubelet will retry as described in the configuration-change case. Could we explicitly document the expected upgrade behavior and plan to add tests for this scenario?

I think this should act similarly as to increment of `SystemReservedCPUs` on a condense system where all CPUs are taken by pods - Kubelet will refuses to start some of the pods because now `Allocatable` pool for CPUs is smaller.

I'm unsure about this one though, so I will look into it and get back to you.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: FeatureGate and all its references has been removed. Thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Added a section about QoS thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: In case 1 OVS or whatever application that is running on the host under the specified cgroups in
`--system-reserved-cgroup` flag, would be responsible to handle the rejection coming from the kernel in its code.

This is actually a good thing because `--enforce-node-allocatable` is doing exactly what it's suppose to do - set a hard cap (by setting the cgroups) to avoid a specific application from over committing and breaking/putting pressure on the rest of the system.

keps/sig-node/6252-system-reserved-hugepages/README.md
```diff
@@ -0,0 +1,694 @@
...
+### Cgroup Enforcement
+
+Memory and hugepages use separate cgroup controllers: the `memory` controller
+for RAM, and the `hugetlb` controller for hugepages. There is a kernel mount
+option (`memory_hugetlb_accounting`) that routes hugepage accounting through
+the memory controller instead, but Kubernetes does not enable it. Doing so
+would conflict with container runtimes, which set the two controllers
+independently based on pod resource requests.
+
+`getCgroupConfigInternal()` in `pkg/kubelet/cm/node_container_manager_linux.go`
```
ffromani, line 303: if you refer to kubernetes implementation, please pin it to a specific kubernetes version so we have a stable reference.
The KEP should be timeless, but some KEPs are in nature incremental update so having a stable reference is the next best thing

ffromani (commented):
non-binding LGTM to signal all my comments have been addressed. Leaving the final call to the assigned reviewers. PTAL to inline comment about the last important improvement I can think of

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: You mean the component responsible for setting the cgroups enforcement per each QOS class?

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: In case 1 OVS or whatever application that is running on the host under the specified cgroups in
--system-reserved-cgroup flag, would be responsible to handle the rejection coming from the kernel in its code.

This is actually a good thing because --enforce-node-allocatable is doing exactly what it's suppose to do - set a hard cap (by setting the cgroups) to avoid a specific application from over committing and breaking/putting pressure on the rest of the system.

In case 2 it will follow the same pattern as for other resources:
https://github.com/kubernetes/kubernetes/blob/6d13e54a08b511582318f24ba0a5bb670214693e/pkg/kubelet/cm/node_container_manager_linux.go#L141-L148
TL;DR kubelet will failed to start but that's the same behavior for other resources as well.

Added those details in the KEP

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: FeatureGate and all its references has been removed. Thanks!

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Anyway, I've added a reference to ### Cgroup Enforcement where QoS cgroup manager set hugepages limit to be the new `Allocatable` (total - reserved) and this won't be enabled by default only when setting `--system-reserved` and/or `--kube-reserved` for hugepages.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: pinpoint to a specific version here and in other places on the KEP. Thanks!

Tal-or:
> Thinking of some of the work from NUMA-aware scheduling point of view, I was wondering if we currently have per NUMA allocated hugepages captured through podresource API (I think the answer is No but I am not 100% sure). I wonder if this something we should think of perhaps in the next iteration of this KEP (or even as a separate KEP)?

@swatisehgal we do: https://github.com/kubernetes/kubernetes/blob/bba12b9fa9e7e6f3767021b585b6cf660f1217be/staging/src/k8s.io/kubelet/pkg/apis/podresources/v1/api.proto#L57-L62 

the `ContainerMemory`  used to represent both standard memory and different hugepage sizes for the PodResourceAPI [resposne](https://github.com/kubernetes/kubernetes/blob/bba12b9fa9e7e6f3767021b585b6cf660f1217be/pkg/kubelet/cm/container_manager_linux.go#L1095-L1115).  

There are also unit-tests in the code exercising hugepages resources for this object.

keps/sig-node/6252-system-reserved-hugepages/README.md
Tal-or: Revised this section PTAL

Tal-or:
@jpbetz @swatisehgal 

Hey folks, thank you for the thorough review. 
I've addressed all of your open questions/comments PTAL

jpbetz:
/approve
For PRR

swatisehgal (commented):
/lgtm

from node-perspective

Tal-or (commented):
@jpbetz @swatisehgal @ffromani Thank you all for the through review 

Hey @dchen1107 PTAL 
PR got /lgtm from node-perspective and /approve from PRR perspective.

Need your final ack in case you're happy with the KEP

KEP freeze is coming on Tuesday.
Thanks!

PR #6257, opened 2026-07-23
Update the API in Node Allocatable DRA KEP to reflect the alpha2 implementation

pravk03:
- One-line PR description: Update the API in Node Allocatable DRA KEP to reflect the alpha2 implementation

- Issue link: https://github.com/kubernetes/enhancements/issues/5517

- Other comments:

https://github.com/kubernetes/kubernetes/pull/140009 deviates from the API proposed in the  KEP.

### DIFF
--- a/keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
+++ b/keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
@@ -357,7 +357,7 @@ To support unified accounting for node allocatable resources, this KEP proposes 
 
 #### Device API Extensions
 
-The new field `NodeAllocatableResourceMappings` within the
+The new field `NodeAllocatableResources` within the
 `ResourceSlice.Device` spec is used to define the node allocatable
 resource quantities.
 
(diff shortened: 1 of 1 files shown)


vitrixLab:
Thank you for updating the KEP to reflect the `alpha2 `implementation.

As someone following the evolution of DRA, it would be helpful to include a brief explanation of the design considerations behind these API changes.

In particular, the transition from `nodeAllocatableResourceMappings` to `nodeAllocatableResources` appears to represent more than a naming change: it introduces a broader resource accounting model that separates resource representation from the way allocations are translated into node resource consumption.

A short explanation of why the implementation diverged from the original proposal, and how this revised model better supports the long-term direction of DRA, would make the architectural intent clearer for contributors.

One area that may benefit from additional clarification is the relationship between existing node resource accounting (`node.status.allocatable`) and DRA-provided accounting through `ResourceSlice`.

Specifically, documenting how these representations are expected to coexist, which component is authoritative when calculating available capacity, and how temporary inconsistencies or synchronization concerns are handled would help contributors understand the intended architecture before diving into implementation details.

liggitt:
thanks for updating
/lgtm

PR #6260, opened 2026-07-27
KEP-5474: revise the CRI API for cgroup mount mode and move alpha to v1.38

chrishenzie:
- One-line PR description: Revises the CRI API surface for `cgroupOptions.mountMode` following API review feedback, and moves alpha to v1.38.

- Issue link: https://github.com/kubernetes/enhancements/issues/5474

- Other comments: See https://github.com/kubernetes/kubernetes/pull/139824 for API review feedback that motivated this change. The milestone moved because the implementation did not reach a merge-ready state by 1.37 code freeze.

Drafted with AI assistance. Content reviewed by me.

/sig node
/cc @divya063 @liggitt @dchen1107


vitrixLab:
Before diving into the implementation, 
I think the highest priority is validating the API contract itself. 

Since this revises the CRI surface for 

`cgroupOptions.mountMode`, 

I'd like to understand why this API shape is the right long-term abstraction, 
how it behaves across version skew and older runtimes, 
and what the compatibility and fallback semantics are. 

If those guarantees are well defined, the implementation review becomes much more straightforward.

SergeyKanzhelev (approved):
/lgtm

/assign @dchen1107 

(as per kep.yaml)

keps/sig-node/5474-enable-writable-cgroups/kep.yaml
```diff
@@ -19,13 +19,13 @@ stage: alpha
...
+latest-milestone: "v1.38"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
-  alpha: "v1.37"
-  beta: "v1.38"
-  stable: "v1.39"
+  alpha: "v1.38"
+  beta: "v1.39"
+  stable: "v1.40"
```
haircommander, line 28: this is an ambitious beta->stable timeline

chrishenzie: We can always adjust. Should we target 1.41 or just mark TBD?

haircommander: I think we'd need a good reason to go GA so fast so I think we should keep to 1.41 and adjust if something comes up

chrishenzie:
Closing in favor of https://github.com/kubernetes/enhancements/pull/6366, which addresses other design changes that came up while implementing the feature

Issue #6268, opened 2026-08-06
Stall and resume slow watch streams instead of terminating them (WatchCacheStallResume)


nami-ant:
### Enhancement Description

- **One-line enhancement description (can be used as a release note):** Park and resume slow watch streams from the watch cache history instead of force-closing them, so a slow or stalled watch client no longer delays or terminates healthy watchers (`WatchCacheStallResume` feature gate, alpha).
- **Kubernetes Enhancement Proposal:** `keps/sig-api-machinery/NNNN-watch-cache-stall-resume` (KEP PR to follow)
- **Discussion Link:** the implementation PR series (see below) and the sig-api-machinery watch-stability discussion
- **Primary contact (assignee):** @nami-ant
- **Responsible SIGs:** sig-api-machinery
- **Enhancement target (which target equals to which milestone):**
  - **Alpha release target (x.y):** 1.38
  - **Beta release target (x.y):** TBD
  - **Stable release target (x.y):** TBD
- [ ] **Alpha**
  - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6269
  - [ ] Code (`k/k`) update PR(s): kubernetes/kubernetes#141475 (benchmark), kubernetes/kubernetes#141448 (metric label), kubernetes/kubernetes#141449 (refactor), kubernetes/kubernetes#141450 (gate and mechanism), kubernetes/kubernetes#141451 (stalled watchers gauge and cull), kubernetes/kubernetes#141452 (integration test); split out of kubernetes/kubernetes#141228 (closed)
  - [ ] Docs (`k/website`) update PR(s): N/A for alpha

_Please keep this description up to date._

/sig api-machinery
/kind feature

jpbetz:
/label lead-opted-in
/milestone v1.38

While I agree that a KEP is not required. It did take me quite a bit of reading to understand the problem being solved here, so I think there is a lot of value to the community to writing one.

If we want to add a KEP in 1.38 for this, we need to have the PR open by end-of-day today (anywhere on earth EOD) to meet the KEP readiness deadline.

I'm happy to volunteer to be PRR reviewer for it.

If there is not enough time, since this is internal only changes, we can always track this as a provisional kep in the near term too.

wojtek-t:
There is a draft for it already in https://github.com/kubernetes/enhancements/pull/6269

tico88612:
Hello @nami-ant :wave:, 1.38 Enhancements team here.

[...]

PR #6269, opened 2026-08-06
KEP-6268: Stall and resume slow watch streams instead of terminating them

nami-ant:
- One-line PR description: Add provisional KEP-6268 for the `WatchCacheStallResume` feature gate.
- Issue link: kubernetes/enhancements#6268
- Other comments:

This adds the provisional KEP for **WatchCacheStallResume** (sig-api-machinery, alpha): when a watch's delivery buffer fills, park the watch and resume it from the watch cache history instead of force-closing it, so a slow or stalled watch client no longer delays or force-closes healthy watchers. A watch is terminated only when its resume position has aged out of the history (an honest `410 Gone`). Off by default.

Implementation PRs: kubernetes/kubernetes#141475 (benchmark), kubernetes/kubernetes#141448 (metric label), kubernetes/kubernetes#141449 (refactor), kubernetes/kubernetes#141450 (gate and mechanism), kubernetes/kubernetes#141451 (stalled watchers gauge and cull), kubernetes/kubernetes#141452 (integration test); split out of kubernetes/kubernetes#141228 (closed)

Draft pending KEP-approver / PRR review; happy to iterate on the design and PRR details here.

/sig api-machinery
/kind kep


jpbetz:
Is this targeting v1.38?

If so please bring it out of draft mode on github and add a `keps/prod-readiness/sig-api-machinery/6268.yaml` file. Thanks!

keps/sig-api-machinery/6268-watch-cache-stall-resume/README.md
```diff
@@ -0,0 +1,531 @@
...
+falls off the history, forces a re-list that adds load. With the gate on, a
+stalled agent is parked and imposes no cost on others; if it never recovers it
+receives an honest `410`.
+
+### Notes/Constraints/Caveats
+
+- The unit of tolerance shifts from "how many events fit in a fixed-size channel"
+  (a count) to "how far back the watch cache history reaches" (a time window,
+  `eventFreshDuration`). The history window, not the channel size, now bounds how
+  long a client may lag before it must re-list.
```
wojtek-t, line 180: As a followup once we prove the feature in production, we can consider simplify the heuristics about the watch channel sizes (to some extent also optimizing the memory usage).

keps/sig-api-machinery/6268-watch-cache-stall-resume/README.md
```diff
@@ -0,0 +1,531 @@
...
+When a watch client cannot keep up and its per-watcher delivery buffer fills, the
+apiserver blocks the shared watch-dispatch path for up to a small shared time
+budget and, if the buffer is still full, force-closes the watch as
+"unresponsive". This has two costs at scale: the dispatcher's wait delays
+delivery to every watcher of that resource, and a watcher that is merely *behind*
+(for example, one replaying its history right after a reconnect) is guaranteed to
+fill its buffer and be terminated — which triggers an immediate reconnect with a
+larger gap and repeats.
+
+This KEP proposes an alpha, opt-in `WatchCacheStallResume` feature gate for
```
wojtek-t, line 54: Let's clarify that "Alpha" and "opt-in" are only rollout stages.

Once we prove it in production this should become "the behavior" and the old one should be removed.

keps/sig-api-machinery/6268-watch-cache-stall-resume/README.md
```diff
@@ -0,0 +1,531 @@
...
+
+## Design Details
+
+The change is contained within
+`staging/src/k8s.io/apiserver/pkg/storage/cacher`. When the gate is enabled, each
+`cacheWatcher` gains stall state (a single-slot latch holding the first missed
+resource version, the starting position, and a last-progress timestamp) and a
+read-only view over the existing watch cache history; the `Cacher` gains a
+stalled-watchers sampler. Dispatch, on a failed non-blocking enqueue, latches the
+event's resource version instead of waiting/closing. The watcher's processing
```
wojtek-t, line 221: I might be missing something, but I think the following patter would be simpler to reason about and less-error prone:

1) When the event can't be dispatched to the `cacheWatcher` we:
  a) store its `resourceVersion` (to ensure we know where to resume from)
  b) remove the `cacheWatcher` from the list of registered watches in the Cacher (Cacher no longer tries to deliver any events themselves)

2) When the faulty `cacheWatcher` is done with draining its input channel and observes it simply reinitializes itself the same way as new watch is established - the operation is exactly the same - we need to replay the events from a given RV (in this case the one stored)

The benefit of this approach is that we minimize the number of new codepaths, we utilize the existing logic to guarantee we deliver events exactly once in order and the faulty `cacheWatcher` really is not affecting the other watchers at all in the meantime.

@serathius - for your thoughts too

wojtek-t (commented):
Thanks a lot for opening this KEP - this was on my list of things for a long time...

keps/sig-api-machinery/6268-watch-cache-stall-resume/kep.yaml
```diff
@@ -0,0 +1,38 @@
...
+creation-date: 2026-08-06
+reviewers:
+  - TBD
+approvers:
+  - TBD
+
+see-also: []
+replaces: []
+
+stage: alpha
```
wojtek-t, line 17: I'm happy to take the PRR of it.

keps/sig-api-machinery/6268-watch-cache-stall-resume/README.md
serathius: Introducing new codepaths can make the changes easier to review and validate, however I think there is performance limitations of this approach. In etcd we always preferred server side re-sync as it allows us batch resync of multiple watchers and reuse events.

While cost scanning of cyclic buffer is minimal, watch scalability really depends on the caching object serialization. Server driven resynchronization would allow us to reuse serializations and scale resync for large number of wachers, while client resync will need to serialize each object separatly.

Watchers resync was one of the main allocation sources on bad path discovered in https://github.com/kubernetes/kubernetes/issues/142223. Doing client side resync would not bring the same performance benefits, requiring us to commit to permanently caching objects to make them available to clients. While I think we might still go to that direction, it's also much more risky change.

To keep the scope of the proposal smaller and still benefit from caching serializations I think doing server side resync is better short term.

EDIT: by client side I meant goroutine directly handling Watch call, and server side the background goroutines distributing events.

wojtek-t: I think there is some misunderstanding - I didn't suggest doing that client-side.

`cacheWatcher` is internal concept in the watchcache and everything that I described above remains server-side (limited to watchcache).

keps/sig-api-machinery/6268-watch-cache-stall-resume/kep.yaml
jpbetz: Thanks @wojtek-t.

@nami-ant please add a `keps/prod-readiness/sig-api-machinery/6268.yaml` file to this PR and list @wojtek-t as the alpha reviewer. This needs to happen today (EOD is defines as being "anywhere on earth") to make the KEP readiness deadline for v1.38 

EDIT: This PR does not need to merge today, but it does need to be updated to contain the file.

keps/sig-api-machinery/6268-watch-cache-stall-resume/kep.yaml
```diff
@@ -0,0 +1,38 @@
+title: Stall and resume slow watch streams instead of terminating them
+kep-number: 6268
+authors:
+  - "@nami-ant"
+  - "@rpb-ant"
+owning-sig: sig-api-machinery
+status: provisional
```
jpbetz, line 7:
```suggestion
status: implementable
```

(Required for KEP readiness deadline today)

keps/sig-api-machinery/6268-watch-cache-stall-resume/kep.yaml
```diff
@@ -0,0 +1,38 @@
...
+creation-date: 2026-08-06
+reviewers:
+  - TBD
+approvers:
+  - TBD
```
jpbetz, line 12: Please add me as a reviewer and an approver for now. We'll expand shortly.

jpbetz (commented):
Please also move this out of draft for the KEP readiness deadline today

Issue #6272, opened 2026-08-12
Service External Failover Policy


defo89:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Introduce a new Service API field, `spec.externalFailoverPolicy`, for Services using `externalTrafficPolicy: Local`, allowing traffic that would otherwise be dropped or rejected due to no eligible endpoints to instead fail over through normal network routing.
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6273
- Discussion Link: 
- Related issue: kubernetes/kubernetes#139300
- PRs by stage and milestone:
  - [ ] Alpha - v1.37
    - [ ] KEP (`k/enhancements`) update PR(s): #6273
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

/sig network

PR #6273, opened 2026-08-12
KEP-6272: Service External Failover Policy

defo89:
- One-line PR description: Add initial KEP for Service External Failover Policy

- Issue link: #6272

- Other comments: Provisional KEP introducing a new opt-in Service API field
  `spec.externalFailoverPolicy: Passthrough` for `externalTrafficPolicy: Local`
  Services. When set, traffic that would otherwise be dropped or rejected solely
  because no eligible endpoints are reachable is instead allowed to fail over
  through normal network routing, so BGP/anycast/ECMP mechanisms can redirect it
  to another node, cluster, or location. Existing behavior is unchanged by
  default. Feature gate: `ServiceExternalFailoverPolicy`. Targets alpha v1.37,
  beta v1.39, stable v1.41.

/sig network
/cc @danwinship @aojea


keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
+# KEP-6272: Service External Failover Policy
```
danwinship, line 1: ah, sorry, I just realized I hadn't replied to your comments on https://github.com/kubernetes/kubernetes/issues/139300 and was about to do that, and noticed this.

keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
...
+short-circuit) is unchanged.
+
+## Motivation
+
+With `externalTrafficPolicy: Local`, kube-proxy currently installs DROP or
+REJECT rules for external Service destinations when there are no locally
+eligible endpoints.
+
+This is appropriate for deployments where an external load balancer stops
+sending traffic to a node before the absence of local endpoints matters.
```
danwinship, line 114: Not quite; if we could guarantee that the load balancer would stop sending connections to a node before the last endpoint on that node exited, then _it wouldn't matter_ what kube-proxy did in the no-endpoints case, because that case would never get hit. The DROP rule is installed to ensure that during the race where a load balancer is still sending packets to a node that can't accept them, it appears to the client that their packets just got lost, and so their TCP stack will automatically retry the connection, and (hopefully, eventually) get redirected to a node that can actually accept the connection.

keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
...
+However, in BGP/anycast deployments, endpoint state and route advertisement are
+updated independently. Whenever a node has no local endpoints, kube-proxy blocks
+traffic to the external VIP even when normal routing could still deliver it:
```
danwinship, line 117: Right. `externalTrafficPolicy: Local` makes a lot of assumptions, and one of those assumptions is "there's no way to bounce a connection from one node to another without masquerading it and breaking the source IP preservation".

FTR, that assumption is also incorrect for _all_ TCP services under [some configurations of Cilium](https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/#client-source-ip-preservation).

keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
...
+2. kube-proxy installs a DROP or REJECT rule.
+3. Traffic still arriving at the node is blocked by kube-proxy instead of being
+   handled by normal routing.
+
+This persists for as long as the block applies (no local endpoints for the DROP,
+no endpoints anywhere in the cluster for the REJECT) and is not removed by other
+means. The convergence window before a routing controller withdraws the route is
+one instance, but the block also applies when no controller withdraws the route
+at all, or when the VIP is anycast and still serving from another cluster or
+location.
```
danwinship, line 129: OK, so this is a very different thing. Kube-proxy rejects packets to the LB IP when there are no endpoints, because doing that is _correct_; the semantics of Service don't allow for the possibility that it could have additional endpoints that kubernetes doesn't know about.

Compare with [Multi-Cluster Services](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md), where in the case of a cross-cluster Service, the mcs-controller has to arrange for the EndpointSlices to be mirrored across the different clusters, so that each cluster knows about the endpoints in remote clusters.

keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
...
+
+The same issue exists when a Pod in the local cluster accesses the externally
+advertised VIP and the cluster has no endpoints at all: kube-proxy rejects that
+traffic (via the `FORWARD` hook) instead of letting it follow the external
+routing path to another cluster or location. (When other cluster-local
```
danwinship, line 134: So, while the naming of the feature completely obscures this fact, the _primary_ functionality of `externalTrafficPolicy: Local` is to guarantee that the client IP is preserved from end to end. What happens in your system if a pod in one cluster tries to connect to a loadbalancer VIP, but then that connection gets forwarded to another cluster?

If the connection gets masqueraded to the pod's node IP, then you're no longer correctly implementing `externalTrafficPolicy: Local`. If it _doesn't_ get masqueraded, then you have another "hidden" requirement here (that pod IPs must be routable across clusters without masquerading).

keps/sig-network/6272-service-external-failover-policy/README.md
```diff
@@ -0,0 +1,762 @@
...
+
+- When the cluster has no endpoints at all, kube-proxy currently installs a
+  filter `REJECT` for the external VIP that is reached from the `FORWARD` hook,
+  so Pod-originated traffic to the VIP is rejected. `Passthrough` suppresses
+  that reject and lets the packet follow normal node routing, so it can reach
+  another cluster or location.
+- When the local node has no eligible endpoint but other cluster-local
+  endpoints exist, kube-proxy already load-balances this traffic cluster-wide
+  ("up-and-out" simulation). It is served without loss, so `Passthrough` leaves
+  it unchanged.
```
danwinship, line 217: So both of the User Stories here involve the multi-cluster / anycast VIP case, and it seems like that's _really_ what this KEP is about; the stuff about not dropping packets when there are no local endpoints is just a low-level detail of this high-level goal, right?

`externalTrafficPolicy: Local` is kind of a mess, in that it completely intertwines a high-level goal ("preserve client IPs") with specific low-level implementation bits. I'm not sure how much you're using eTP:Local because you actually want the goal (source IP preservation) vs because you want some or all of the implementation bits (eg, no DNAT). But you clearly don't want _all_ of the eTP:Local implementation bits...

It seems like what you want may possibly be a new `externalTrafficPolicy`, or maybe a new `ipMode`, or maybe both. Or maybe it needs its own field, as you've done here. But "ExternalFailoverPolicy" doesn't feel like the right framing.

And of course, as always in these situations, we need to ask "would Gateway API be a better solution for this?" Fundamentally, `type: LoadBalancer` services are mostly about cloud load balancers anyway. If you are implementing the load-balancing part yourself, and the behavior that you want does not exactly match the existing way that `type: LoadBalancer` services work, then it might make more sense to just _not use `type: LoadBalancer`_. If you have a Gateway implementation, then the VIP would be owned by the Gateway, and you could arrange for traffic to that IP to be routed however you need it to be routed, and then eventually DNAT it to an endpoint pod IP yourself directly at the appropriate point.

aojea:
> * When set, traffic that would otherwise be dropped or rejected solely
>   because no eligible endpoints are reachable is instead allowed to fail over
>   through normal network routing, so BGP/anycast/ECMP mechanisms can redirect it
>   to another node, cluster, or location.

Is this behavior also applying to cilium, antrea, ovn-kubernetes, ... a KEP should provide consistency across implementations and AFAIK this is a specific kube-proxy behavior

keps/sig-network/6272-service-external-failover-policy/README.md
defo89: Thanks for the detailed review @danwinship. 

> It seems like what you want may possibly be a new externalTrafficPolicy, or maybe a new ipMode, or maybe both. Or maybe it needs its own field, as you've done here. But "ExternalFailoverPolicy" doesn't feel like the right framing.

I think you're right that a separate field is the wrong abstraction. Maybe proposing a new `externalTrafficPolicy` value would make more sense instead. So I'd like to check direction before rewriting.

`externalTrafficPolicy: PreferLocal` value would be same as `Local`, except it steps aside instead of blocking when no endpoints are available:

| | `Local` | `PreferLocal` |
|-|---------|---------------|
| Local endpoints → DNAT local | ✅ | ✅ |
| Source IP preserved | ✅ | ✅ |
| healthCheckNodePort | ✅ | ✅ |
| No local endpoints, external origin | DROP | passthrough to node networking |
| No endpoints anywhere | REJECT | passthrough to node networking |
| Pod/host short-circuit (other endpoints exist) | DNAT cluster-wide | DNAT cluster-wide |
| NodePort | same | same |

Proxy still owns the VIP. It just stops claiming "definitely unreachable" when it has no endpoints to offer.

>  I'm not sure how much you're using eTP:Local because you actually want the goal (source IP preservation)

We are using it extensively and would still like to preserve Source IP for proxy-served traffic. Passthrough path would be: proxy isn't involved, packet follows normal node egress (SNAT to node IP, same as any pod→external today). Receiving side sees a routable external IP.

> And of course, as always in these situations, we need to ask "would Gateway API be a better solution for this?"

Problem is the proxy datapath dropping/rejecting packets when no endpoints exist. Wouldn't same happen regardless of how the VIP is managed or which implementation runs?

> Compare with [Multi-Cluster Services](https://github.com/kubernetes/enhancements/blob/master/keps/sig-multicluster/1645-multi-cluster-services-api/README.md), where in the case of a cross-cluster Service, the mcs-controller has to arrange for the EndpointSlices to be mirrored across the different clusters, so that each cluster knows about the endpoints in remote clusters. 

MCS would require multi-cluster control plane and also assume that targets are K8s. In the proposed approach we could fail over to non-K8s infra too (examples: NTP, DNS, LDAP server).

defo89:
> > * When set, traffic that would otherwise be dropped or rejected solely
> >   because no eligible endpoints are reachable is instead allowed to fail over
> >   through normal network routing, so BGP/anycast/ECMP mechanisms can redirect it
> >   to another node, cluster, or location.
> 
> Is this behavior also applying to cilium, antrea, ovn-kubernetes, ... a KEP should provide consistency across implementations and AFAIK this is a specific kube-proxy behavior

Thanks for the good follow up question. I have spent some tokens to research the datapath behavior and provide code references for each implementation. 

The behavior of dropping/rejecting when eTP:Local has no local endpoints is not kube-proxy specific. Every major service proxy implementation does the same thing. But it's worth noting though that kube-proxy distinguishes between DROP (no local endpoints but cluster endpoints exist) and REJECT (no endpoints anywhere). Calico BPF, Cilium BPF, Antrea, and OVN-Kubernetes do not make this distinction just reject in both cases.

The proposed behavior (suppress drop/reject, let packet pass through node networking) would need implementation across all service proxies, but the underlying problem is consistent.

| Implementation | No local endpoints (eTP:Local) | Mechanism |
|---|---|---|
| **kube-proxy** | DROP (external origin) / REJECT (no endpoints anywhere) | iptables/nftables rules |
| **Calico (BPF)** | ICMP dest unreachable (port unreach) | `NAT_NO_BACKEND` → ICMP reply |
| **Cilium (BPF)** | ICMP dest unreachable (port unreach) | `DROP_NO_SERVICE` → `SERVICE_NO_BACKEND_RESPONSE` |
| **Antrea (OVS)** | TCP RST / ICMP dest unreachable (host admin prohibited) | OVS packet-in → controller sends reject |
| **OVN-Kubernetes** | Reject (OVN LB native `reject=true`) | OVN load balancer option |

### Code references

**Calico** — BPF NAT lookup returns `NAT_NO_BACKEND` when local count is 0, then `tc.c` sends ICMP port unreachable:
- [`felix/bpf-gpl/nat_lookup.h#L158-L160`](https://github.com/projectcalico/calico/blob/8dd8d361/felix/bpf-gpl/nat_lookup.h#L158-L160) — `if (count == 0) → NAT_NO_BACKEND`
- [`felix/bpf-gpl/tc.c#L538-L548`](https://github.com/projectcalico/calico/blob/8dd8d361/felix/bpf-gpl/tc.c#L538-L548) — ICMP dest unreach / port unreach reply
- [`confd/pkg/backends/calico/routes.go#L545-L554`](https://github.com/projectcalico/calico/blob/8dd8d361/confd/pkg/backends/calico/routes.go#L545-L554) — BGP route withdrawn when no local endpoints

**Cilium** — `lb4_local` returns `DROP_NO_SERVICE` when backend count is 0; nodeport handler sends ICMP or drops:
- [`bpf/lib/lb.h#L2450-L2458`](https://github.com/cilium/cilium/blob/c19c89e8/bpf/lib/lb.h#L2450-L2458) — `handle_nonroutable_endpoints_v4`: drops for eTP:Local
- [`bpf/lib/lb.h#L2386-L2391`](https://github.com/cilium/cilium/blob/c19c89e8/bpf/lib/lb.h#L2386-L2391) — `tail_no_service_ipv4`: ICMP dest unreach / port unreach
- [`bpf/lib/nodeport.h#L2798-L2802`](https://github.com/cilium/cilium/blob/c19c89e8/bpf/lib/nodeport.h#L2798-L2802) — decision path (`enable_no_service_endpoints_routable` controls whether DROP or ICMP)
- [`pkg/bgp/manager/reconciler/service.go#L583`](https://github.com/cilium/cilium/blob/c19c89e8/pkg/bgp/manager/reconciler/service.go#L583) — BGP route skipped when `eTP:Local && !hasLocalBackends`

**Antrea** — OVS marks packet with `SvcNoEpRegMark`, sends to controller, controller replies with TCP RST or ICMP reject:
- [`pkg/agent/openflow/pipeline.go#L2579`](https://github.com/antrea-io/antrea/blob/16392206/pkg/agent/openflow/pipeline.go#L2579) — loads `SvcNoEpRegMark` when endpoint group is empty
- [`pkg/agent/openflow/service.go#L135-L141`](https://github.com/antrea-io/antrea/blob/16392206/pkg/agent/openflow/service.go#L135-L141) — flow sends packet-in to controller
- [`pkg/agent/openflow/packetout.go#L41-L80`](https://github.com/antrea-io/antrea/blob/16392206/pkg/agent/openflow/packetout.go#L41-L80) — `SendRejectPacketOut`: TCP RST or ICMP host-admin-prohibited

**OVN-Kubernetes** — OVN native LB with `reject=true` (default):
- [`go-controller/pkg/ovn/controller/services/lb_config.go#L834`](https://github.com/ovn-org/ovn-kubernetes/blob/0522888b/go-controller/pkg/ovn/controller/services/lb_config.go#L834) — `lbOptions.Reject = true`
- [`go-controller/pkg/ovn/controller/services/loadbalancer.go#L349-L360`](https://github.com/ovn-org/ovn-kubernetes/blob/0522888b/go-controller/pkg/ovn/controller/services/loadbalancer.go#L349-L360) — passed as OVN LB option

keps/sig-network/6272-service-external-failover-policy/README.md
danwinship: > > And of course, as always in these situations, we need to ask "would Gateway API be a better solution for this?"
> 
> Problem is the proxy datapath dropping/rejecting packets when no endpoints exist. Wouldn't same happen regardless of how the VIP is managed or which implementation runs?

No, you wouldn't have a LoadBalancer Service, and kube-proxy would not be aware of the IP assigned by the Gateway

defo89: > > > And of course, as always in these situations, we need to ask "would Gateway API be a better solution for this?"
> > 
> > 
> > Problem is the proxy datapath dropping/rejecting packets when no endpoints exist. Wouldn't same happen regardless of how the VIP is managed or which implementation runs?
> 
> No, you wouldn't have a LoadBalancer Service, and kube-proxy would not be aware of the IP assigned by the Gateway

I looked into this more concretely. The Gateway API approach would mean: Gateway owns the VIP, Gateway implementation does the DNAT, BGP speaker advertises the Gateway address, kube-proxy is  not involved.

I have checked CIlium as example - it looks to me that in Gateway API implementation it creates `type: LoadBalancer` Services under the hood. Based on the research done in https://github.com/kubernetes/enhancements/pull/6273#issuecomment-5326619649 - same REJECT behavior would apply, right?

1. The Gateway reconciler calls [ensureService()](https://github.com/cilium/cilium/blob/main/operator/pkg/gateway-api/gateway_reconcile.go#L200) which creates/patches a Service for every Gateway.

 2. The translator constructs a [type: LoadBalancer](https://github.com/cilium/cilium/blob/main/operator/pkg/model/translation/gateway-api/translator.go#L274-L279) Service with [externalTrafficPolicy](https://github.com/cilium/cilium/blob/main/operator/pkg/model/translation/gateway-api/translator.go#L287-L296) set from config.

 3. That Service ends up in the BPF LB map → same [DROP_NO_SERVICE](https://github.com/cilium/cilium/blob/c19c89e8/bpf/lib/lb.h#L2450-L2458) / [ICMP reject](https://github.com/cilium/cilium/blob/c19c89e8/bpf/lib/lb.h#L2386-L2391) path when no backends exist.

If I read Calico implementation correctly, it embeds [Envoy Gateway](https://github.com/envoyproxy/gateway) as its Gateway API controller ([gateway_api.go](https://github.com/projectcalico/calico/blob/main/operator/pkg/render/gatewayapi/gateway_api.go)). Envoy Gateway defaults to [ServiceTypeLoadBalancer](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/kubernetes_helpers.go#L88-L91) + [ServiceExternalTrafficPolicyLocal](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/kubernetes_helpers.go#L98-L100), applied via [ExpectedServiceSpec](https://github.com/envoyproxy/gateway/blob/main/internal/infrastructure/kubernetes/resource/resource.go#L25-L48) for every Gateway proxy.

With regards to the above, a Gateway implementation could in theory bypass the Service model entirely, but the major ones that exist today does translate Gateway to LoadBalancer Service.

So, if I understood the references right, switching to Gateway API would not avoid the problem. Happy to be corrected of course. 

Coming back to the PreferLocal direction from my earlier comment — does a new externalTrafficPolicy value feel like the right level to solve this at?

Happy to rework the KEP around PreferLocal if the direction makes sense, or explore alternatives if there's a better solution.

defo89: Hi @danwinship, just following up on this when you have a chance.

defo89:
Hi @danwinship @aojea,

Following up with concrete evidence on the Gateway API question and why it's not the best fit for this problem.

I built a reproducer using Calico's Gateway API implementation (Envoy Gateway) on a kind cluster: GatewayClass → Gateway → HTTPRoute → backend.

Here is a [gist](https://gist.github.com/defo89/99fba7a7b575ccb3175df2164d4a82e1) with reproducer in local kind cluster.

Gateway API implementations create a `type: LoadBalancer` Service per Gateway — this is the Service that receives external traffic and forwards it to the envoy-proxy pod. Since it's a regular Service, kube-proxy (or its BPF replacement) manages it like any other — the VIP goes through the same iptables/nftables rules.

When application deployment scales to zero, the envoy-proxy pod is still running, so its LB Service still has endpoints. Traffic reaches envoy, which has no backends to forward to and returns 503.

The BGP advertisement with Gateway API looks interesting. I checked how two major BGP speakers handle this:

- Cilium — BGP reconciler has explicit [hasLocalProxy](https://github.com/cilium/cilium/blob/main/pkg/bgp/manager/reconciler/service.go) logic: 
```
		// For services with real backends, respect externalTrafficPolicy: Local by skipping
		// advertisement when there are no local endpoints.
```
Above is what we want for anycast failover to work.
```
		// For proxy-redirected services (Gateway API/Ingress), advertise if the local
		// proxy is handling traffic, since backends are managed by the proxy.
```
And here the route is never withdrawn when backends disappear, because the proxy pod is still always running. 

- Calico — BGP speaker ([advertiseThisService](https://github.com/projectcalico/calico/blob/master/confd/pkg/backends/calico/routes.go)) checks EndpointSlices for local ready endpoints. The envoy-proxy pod is a local endpoint of the LB Service, so it keeps advertising  even with zero backends.


For a BGP speaker to withdraw the route in the Gateway API model, it would need to look through the proxy to the HTTPRoute's backendRef EndpointSlices — something no implementation does today.

Based on above findings, referring to the question:
> would Gateway API be a better solution for this

I'd say no - the fix rather belongs at the service proxy level. Coming back to the PreferLocal direction from my [earlier comment](https://github.com/kubernetes/enhancements/pull/6273#discussion_r3802549601) — does a new eTP value still feel like the right abstraction?

Issue #6274, opened 2026-08-13
KEP-6274: GPU support for Pod-level Checkpoint/Restore


rst0git:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Add GPU support to Pod-level checkpoint/restore for Pods using device-plugin or DRA allocations.
- Kubernetes Enhancement Proposal: TBD
- Discussion Link: TBD
- PRs by stage and milestone:
  - [x] Alpha - v1.38
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6275
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

yindia:
follow

karimzakzouk:
Hello @rst0git :wave:, 1.38 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 22nd September 2026 (AoE) / Wednesday 23rd September 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.38 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [x] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/tree/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

Note that the PR is not required to be approved or merged by the PRR freeze deadline. Having the PRR questionnaire filled out by the deadline will help ensure that the PRR team has enough time to review your KEP before **enhancements freeze on Tuesday 29th September 2026 (AoE) / Wednesday 30th September 2026 12:00 UTC**. For more information on the PRR process, see [here](https://github.com/kubernetes/community/blob/master/sig-architecture/production-readiness.md#submitting-a-kep-for-production-readiness-approval).

[...]

PR #6275, opened 2026-08-13
KEP-6274: Add GPU support for Pod-level Checkpoint/Restore

rst0git:
- One-line PR description: Add the initial KEP for GPU support in Pod-level Checkpoint/Restore, covering device-plugin and DRA allocations.

- Issue link: https://github.com/kubernetes/enhancements/issues/6274

- Other comments: This KEP is provisional and does not target a Kubernetes release yet. SIG Node is the proposed owning SIG, with SIG Scheduling participating in the DRA aspects.


rst0git:
cc: @haircommander @andreyvelich @adrianreber @dfeigin-nv @Ronkahn21

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+5. For the NVIDIA implementation, the GPU integration performs the equivalent
+   of:
+   - `RUNNING -> LOCKED` with a bounded lock;
+   - `LOCKED -> CHECKPOINTED`, moving GPU memory to host memory and releasing
+     device resources;
+   - CPU process-tree checkpoint through CRIU;
+   - restoration of the source process's GPU state onto its original GPU; and
+   - `LOCKED -> RUNNING` before the kubelet resumes the Pod.
+6. The runtime returns success only after CPU and GPU artifacts are complete and
+   the source process has returned to a valid running state.
```
dfeigin-nv: If something goes wrong in the checkpoint process, will the source process go into Error state?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+- Checkpoint and restore a Pod using one full GPU on Linux.
+- Support GPUs allocated through both:
+  - the Kubernetes device-plugin framework; and
+  - DRA through a generated ResourceClaim from a ResourceClaimTemplate.
+- Preserve the KEP-5823 Alpha lifecycle: same-node restore, a new Pod UID, and a
+  source Pod that remains running.
+- Allocate and prepare a new device before calling CRI `RestorePod`.
+- Let the runtime map the original device to the new allocation and reject an
+  incompatible target without starting a fresh process.
+- Make checkpoint success atomic across CPU process state and GPU state.
```
dfeigin-nv: atomic how? the container is paused and gpu state is frozen? How is that guaranteed?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+### Security and privacy
+
+KEP-5823 authorization and path-confinement rules remain authoritative. This
+extension adds no user-supplied host path, device identifier, helper binary, or
+driver option. Kubelet supplies the checkpoint directory and device
+configuration from its own device-manager and DRA state.
+
+GPU memory and driver state are sensitive workload memory. They inherit the
+parent KEP's restrictive node-local storage, retention, and restore authorization
+requirements. Raw device IDs and vendor compatibility data remain in the
```
dfeigin-nv: Could you clarify exactly what “vendor compatibility data” includes—for example, GPU architecture/model, memory requirements, driver or checkpoint-format constraints

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+Before generating restore container configs, kubelet device admission has
+already called `Allocate` for the new Pod. `GenerateRunContainerOptions`
+retrieves that state and performs any device-plugin `PreStartContainer` call
+required before container creation. The resulting configuration can contain
+traditional Linux devices, mounts and environment, CDI device names, or both.
+
+The runtime treats the restore-time config as the new allocation. If the target
+is incompatible, `RestorePod` returns `FailedPrecondition`; kubelet emits
+`DeviceRestoreIncompatible` and applies sandbox retry backoff. Alpha does not
+ask the device manager to swap an allocation after that failure. The user or
```
dfeigin-nv: If incompatibility is persistent, that may repeatedly invoke an expensive restore against the same GPU while retaining the claim/allocation. Could this instead become a terminal condition, or otherwise suppress retries until the Pod or allocation changes

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+
+The runtime owns rollback between GPU state transitions because only it knows
+the vendor state machine. On a checkpoint error it must inspect the process
+state and:
+
+- unlock only from a state for which unlock is valid;
+- never attempt a normal transition from a vendor-reported failed state;
+- remove partial checkpoint output;
+- report whether the source was safely returned to running in its diagnostic
+  message.
```
dfeigin-nv, line 933: By message we mean text or something structured?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+This KEP grants no new Kubernetes API access to the runtime or GPU Operator.
+DRA API access remains with existing scheduler, kubelet, and DRA components; the
+core path must not copy the standalone agent's cluster-wide API lookup into a
+container runtime.
+
+GPU checkpointing amplifies the parent KEP's denial-of-service risk because
+device memory may be staged in host memory and checkpoint storage. The parent
+deadline, concurrency, storage-accounting, and cleanup controls apply to the
+combined CPU/GPU operation. GPU-specific work must remain bounded by the CRI
+context and stop after cancellation.
```
dfeigin-nv, line 1039: Can the CUDA driver transition, CRIU plugin, and external vendor helper all be reliably interrupted when the CRI context is cancelled?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+stack, and optional DRA driver. They then enable
+`PodLevelCheckpointRestoreDevices` on selected kubelets after prerequisite
+validation succeeds.
+
+Disabling the gate stops new device-bearing checkpoint and restore CRI calls.
+Existing source and restored Pods continue running. Existing archives and
+`PodCheckpoint` objects are not modified or deleted and can be used after the
+gate is re-enabled on a compatible node.
+
+A runtime or driver downgrade may make an existing GPU checkpoint unreadable.
```
dfeigin-nv, line 1345: If a runtime or driver change makes the archive unreadable, does the `PodCheckpoint `remain `Ready=True`? If so, please clarify that Ready only means the checkpoint was complete when created and does not guarantee that it is restorable with the node’s current runtime and driver versions.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
dfeigin-nv: I’m asking because a CRI deadline only bounds the RPC if the underlying CUDA/CRIU/helper operation honors cancellation. If any phase is not interruptible, what prevents it from continuing after the RPC times out, and how are the source GPU lock and temporary resources eventually recovered?

The PoC should make sure they can be faulted/interrupted/killed, and check the state of the processes and GPU afterwards.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+
+[KEP-5823] introduces Pod-level checkpoint and restore, but deliberately excludes
+devices attached through the Kubernetes device-plugin framework or Dynamic
+Resource Allocation (DRA), as well as device memory and driver state. This KEP
+extends that mechanism so a Pod using a GPU can be checkpointed and restored as
+one operation.
+
+The Alpha design supports same-node restore into a new Pod while the source Pod
+continues running. The new Pod receives a new GPU allocation through either the
+device-plugin path or DRA. Kubernetes orders allocation and device preparation
```
haircommander: if it would be at all simpler, I lean towards leaving out device-plugin

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+
+- Checkpoint and restore a Pod using one full GPU on Linux.
+- Support GPUs allocated through both:
+  - the Kubernetes device-plugin framework; and
+  - DRA through a generated ResourceClaim from a ResourceClaimTemplate.
+- Preserve the KEP-5823 Alpha lifecycle: same-node restore, a new Pod UID, and a
+  source Pod that remains running.
+- Allocate and prepare a new device before calling CRI `RestorePod`.
+- Let the runtime map the original device to the new allocation and reject an
+  incompatible target without starting a fresh process.
```
Ronkahn21: when the rejection happens? is before pod binding, if after what would happed to the pod that is rejected but bound

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+  models, GPU UUIDs, or CRIU plugin versions.
+- A custom scheduler plugin or a dependency on GPU Feature Discovery labels.
+- Requiring the NVIDIA GPU Operator. Preinstalled drivers and toolkit components
+  remain valid when they satisfy the same node/runtime contract.
+- AMD GPU support in Alpha. AMDGPU/ROCm maintainers should review the contract,
+  and a second implementation is a Beta goal.
+
+## Proposal
+
+Introduce a kubelet-only `PodLevelCheckpointRestoreDevices` feature gate layered
```
Ronkahn21: should it be PodLevelCheckpointRestoreGPUDevices? as this KEP is not aiming to solve any type device just GPU devices

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+CRI operation.
+
+The kubelet does not inspect vendor compatibility. On checkpoint, the runtime
+discovers the devices used by the existing sandbox and stores original device
+identity and vendor restore requirements in its opaque checkpoint directory. On
+restore, normal Kubernetes allocation runs first. The kubelet supplies complete
+restore-time `ContainerConfig` objects, including Linux devices, mounts,
+environment, annotations, and CDI device names, to `RestorePod`. The runtime
+uses those configs and its checkpoint inventory to build the old-to-new device
+mapping, validate compatibility, restore CPU and GPU state, and return restored
```
Ronkahn21: what is the behavior if the pod is not compatible

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
Ronkahn21: Why we mind if the pod is not resumed correctly if Checkpoint succeeded but the continuer failed to resume

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1352 @@
...
+containers in the non-running state required by KEP-5823.
+
+Alpha intentionally relies on the original workload's allocation constraints
+for scheduling. Runtime validation is authoritative. DRA compatibility selectors
+generated from a checkpoint are deferred until a driver-owned compatibility
+vocabulary and the interaction with KEP-5823 Pod-template equality are agreed.
+
+### Ownership and review boundaries
```
kannon92: I would remove this section. Don't see the value here.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
rst0git: Thanks! I've updated the KEP to be DRA-only a this stage.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
rst0git: Thank you Kevin! I've updated the KEP to drop this section.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
rst0git: Sure, I've updated the KEP to clarify it.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
rst0git: Thanks! I've updated the KEP to clarify this.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
rst0git: Thanks! I've updated the KEP to address this.

keps/sig-node/5823-pod-level-checkpoint-restore/README.md
carlory, line 808: Nit: spec.sourcePodName -> spec.sourcePod.name

carlory, line 633: Nit: spec.sourcePod.name?

kannon92:
can you fix the failing checks on this PR

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/kep.yaml
```diff
@@ -0,0 +1,45 @@
...
+  - "@dfeigin-nv"
+  - "@Ronkahn21"
+owning-sig: sig-node
+participating-sigs:
+  - sig-scheduling
+status: provisional
+creation-date: 2026-08-13
+reviewers:
+  - "@haircommander"
+approvers:
```
harche: Small one: this is what makes [`pull-enhancements-verify`](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/enhancements/6275/pull-enhancements-verify/2102453588822331392) / `-test` fail. With `stage: alpha` + `latest-milestone: "v1.38"`, the check needs a SIG Node tech lead here:

```
approver alpha-stage KEP must list at least one sig-node-tech-leads member as approver
```

Since the PR description says this is provisional with no release target yet, would it make sense to drop the milestone (and the PRR alpha entry) for now? Either way should turn CI green.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/kep.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+feature-gates:
+  - name: PodLevelCheckpointRestoreDevices
+    components:
+      - kube-apiserver
+      - kube-controller-manager
+      - kubelet
+disable-supported: true
+
+metrics:
```
harche: Nit: these metrics are all existing ones (from [KEP-5823](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5823-pod-level-checkpoint-restore), and `kubelet_runtime_operations_errors_total` already exists). As far as I can tell this KEP only proposes a `device_allocation` label. Maybe worth clarifying so reviewers don't read it as six new metrics?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+check for the `set` verb on the virtual
+`resourceclaims/template-provenance` subresource. Only the built-in
+ResourceClaim-controller role receives that permission. Updates cannot change
+or remove the value. Kubelet reads it from the allocated claim rather than
+gaining permission to read ResourceClaimTemplates.
+
+On restore, the ResourceClaim controller copies the matching
+`ResourceClaimTemplate` and adds system-owned constraints for the recorded DRA
+driver, exact `driverStackID`, and minimum capacities. The DRA driver publishes
+the stack attribute only for checkpoint-capable, exclusive full devices, so
```
harche: I think this assumption doesn't hold for the NVIDIA driver (or the example driver). MPS and time-slicing aren't separate devices in the ResourceSlice; they're chosen **per claim** through opaque config:

```go
type GpuConfig struct {
	metav1.TypeMeta `json:",inline"`
	Sharing         *GpuSharing `json:"sharing,omitempty"`
}
```
[gpuconfig.go#L29-L32](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/api/nvidia.com/resource/v1beta1/gpuconfig.go#L29-L32)

and applied at prepare time:

```go
// Apply MPS settings (if available and feature gate enabled).
if featuregates.Enabled(featuregates.MPSSupport) && config.IsMps() {
```
[device_state.go#L1340-L1341](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/device_state.go#L1340-L1341)

So the same `gpu-0` (with `driverStackID`) can be allocated to an exclusive claim or to an MPS/time-sliced one, and the selector matches both. dra-example-driver works the same way ([gpu.go#L357-L368](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/internal/profiles/gpu/gpu.go#L357-L368)).

Kubelet can't read opaque config, so maybe the runtime or DRA driver should explicitly reject a shared prepared device at checkpoint/restore time? Also curious how an MPS client would be handled, since its state depends on a separate MPS control daemon.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+
+`claimName` and `requestName` use the existing DNS-label validation for their
+source fields. `driver` uses the DRA driver-name validation and its 63-character
+limit. `resourceClaimTemplateUID` is required. `driverStackID` is a non-empty
+ASCII token matching `[A-Za-z0-9][A-Za-z0-9._:+-]{0,63}`, within the DRA
+string-attribute limit. `minimumCapacities` is a map list keyed by a fully
+qualified DRA capacity name, has at most 32 entries, and requires a positive
+quantity for every value.
+
+Alpha requires exactly one `memory` entry under the source driver's domain. Its
```
harche: With dynamic MIG enabled, I think the exact `memory` equality could fail. The NVIDIA driver advertises the full GPU's capacity as the **max across MIG profiles** instead of the NVML total:

```go
func (d *GpuInfo) PartCapacities() PartCapacityMap {
	if len(d.maxCapacities) > 0 {
		return d.maxCapacities
	}
	return d.fullGpuCapacity()
}
```
[partitions.go#L53-L58](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/partitions.go#L53-L58)

where the MIG memory is `MemorySizeMB * 1Mi` ([mig.go#L133](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/mig.go#L133)), while the non-MIG path uses NVML `memory.Total` ([nvlib.go#L494-L497](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/nvlib.go#L494-L497)). A runtime reporting the NVML total wouldn't match exactly. See also my note on [L611](https://github.com/kubernetes/enhancements/blob/e7d20c54b4fe6fd577790022140452bbb69bdbf5/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L611) for a possible simplification.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+9. On any failure, the runtime removes partial output as required by the CRI
+   contract, attempts to resume the source process only when the
+   device checkpoint API reports that resuming is valid, and reports the first
+   causal error. Kubelet records `DeviceCheckpointFailed` when it can classify
+   the error as device-related; otherwise it uses the parent `CheckpointFailed`
+   reason.
+10. After a successful CRI call, kubelet validates the response, associates its
+    single entry with the prevalidated claim, request, and allocation result,
+    and adds the driver and protected ResourceClaimTemplate UID from Kubernetes
+    state. It re-resolves the latest complete ResourceSlice generation and
```
harche: Would it be simpler for kubelet to **take** `driverStackID` and `memory` from the ResourceSlice (it already reads it here), rather than requiring the runtime to compute the same values and cross-checking them?

Today the runtime/CRIU plugin and the DRA driver have to implement the same ID algorithm independently. If one is upgraded without the other, every checkpoint fails. With the slice as the source:
- the runtime only needs to confirm that device state was captured (so we still never get a CPU-only "success"), and
- it keeps its own private fingerprint for the final check at restore.

That might also make most of the new `CheckpointPodResponse` fields unnecessary. What do you think?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+   and `PodLevelCheckpointRestoreDevices` is disabled, or if the Pod uses a
+   device-plugin allocation, the operation fails before CRI. With the gate
+   enabled, kubelet proves from Pod and DRA state that there is exactly one
+   Pod-owned generated claim, one supported exact request, one allocation
+   result across the Pod, and one non-shareable device. Direct, PodGroup,
+   workload-shared, administrative, `All`, and additional allocations fail
+   before CRI.
+3. Before invoking the expensive operation, kubelet validates the protected
+   ResourceClaimTemplate UID annotation and resolves the source allocation
+   against the latest complete ResourceSlice generation. The source device must
```
harche: FYI: kubelet doesn't read ResourceSlices today. Its only call is a cleanup `DeleteCollection` ([dra_plugin_manager.go#L253](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/kubelet/cm/dra/plugin/dra_plugin_manager.go#L253)), so this adds a new list/informer per node. Also, the node authorizer only allows listing slices scoped to the node's own name:

```go
case "watch", "list", "deletecollection":
	// only allow a scoped fieldSelector
	...
	if req.Field == resourceapi.ResourceSliceSelectorNodeName && req.Operator == selection.Equals && req.Value == nodeName {
```
[node_authorizer.go#L381-L385](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/plugin/pkg/auth/authorizer/node/node_authorizer.go#L381-L385)

So slices published with a `nodeSelector` / `allNodes` wouldn't be readable. The NVIDIA and example drivers are node-local, so that's fine, but it may be worth stating as an Alpha constraint. Pool generation "completeness" logic would also need to live in kubelet.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+   publish `driverStackID` and `memory`. When `DRAConsumableCapacity` is enabled,
+   `allowMultipleAllocations` must be false or unset. Missing, malformed, or
+   ambiguous provenance fails the checkpoint without calling the runtime.
+4. Kubelet suspends probes, resolves the sandbox ID, and calls CRI
+   `CheckpointPod` with that ID and a bounded deadline. The runtime enumerates
```
harche: Small alignment with the merged CRI: in `CheckpointPodRequest` the **caller** (kubelet) selects the containers:

```proto
// IDs of the containers to include in the checkpoint. The list must be
// non-empty, must not contain duplicates, and must contain exactly the
// running containers selected by the caller.
repeated string container_ids = 3;
```
[api.proto#L2189-L2194](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L2189-L2194)

So the "exactly one GPU container" rule could be defined against `container_ids`.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+message DeviceCapacityRequirement {
+    string name = 1;
+    string quantity = 2;
+}
+```
+
+- `CheckpointPodRequest` identifies the sandbox. The runtime enumerates the
+  complete container set required by KEP-5823 and inspects its live device
+  configuration.
+- checkpoint output is an opaque runtime-owned directory.
```
harche: Nit: per the merged CRI, kubelet owns this directory and the runtime only writes into it:

```proto
// Absolute path to an existing, empty directory where the runtime must
// write the checkpoint. The caller owns the directory, ...
string output_path = 2;
```
[api.proto#L2181-L2188](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L2181-L2188)

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+every visible GPU, topology and peer-access validation, and clear behavior when
+only a subset of targets can be allocated.
+
+### Runtime and CRI contract
+
+This KEP adds a bounded response field to the KEP-5823 `CheckpointPod` RPC. It
+does not add a new RPC or change `RestorePodRequest`:
+
+```protobuf
+message CheckpointPodResponse {
```
harche: Worth noting in Implementation: the response is currently discarded at the Go interface level, so this is an interface change across cri-api, cri-client, the instrumented wrapper and fakes, not just a proto field:

```go
CheckpointPod(ctx context.Context, options *runtimeapi.CheckpointPodRequest) error
```
[services.go#L92](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/services.go#L92)

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+constructs a complete old-to-new mapping and includes identity mappings for any
+device that does not change.
+
+Alpha restricts the visible set to one full GPU, making the mapping one-to-one.
+For the CUDA implementation, the target must have the required driver stack ID
+and enough memory, and must pass the private device and file checks. The runtime
+or CRIU GPU plugin passes the mapping to the CUDA Driver API restore operation.
+Kubernetes never receives or interprets the UUID pair.
+
+The runtime must resolve the new identity from the actual applied allocation,
```
harche: One thing I couldn't find addressed: device setup often injects **claim-specific env vars**, which will differ between source and restore. For example, dra-example-driver:

```go
fmt.Sprintf("%s_DEVICE_%s_RESOURCE_CLAIM=%s", strings.ToUpper(cdi.class), deviceEnvKey, claimUID),
```
[cdi.go#L101](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/cmd/dra-example-kubeletplugin/cdi.go#L101), plus `GPU_DEVICE_<id>=<device>` ([gpu.go#L349](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/internal/profiles/gpu/gpu.go#L349)).

The restored Pod has a new claim UID (and maybe `gpu-1` instead of `gpu-0`), but the merged CRI says:

```proto
// ... Image, command, args, working directory, environment, and
// process credentials describe the expected checkpointed process; they
// must not start or mutate it, and any mismatch the runtime can validate
// must fail the restore.
```
[api.proto#L2239-L2243](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L2239-L2243)

So a strict runtime would fail every DRA restore, and a lenient one leaves stale values in the process. Could the KEP say which device-injected env is exempt, and what the application should expect to see?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+   every container selected by the KEP-5823 Pod-level contract, pauses the
+   complete set before capturing any container, keeps the set paused until
+   capture finishes, and resumes it before completing the call.
+5. The runtime detects a supported GPU workload from its sandbox and container
+   state. One integration owner coordinates device transitions; CRIU and an
+   external helper must not both issue the CUDA checkpoint transitions.
+6. For the NVIDIA implementation, the GPU integration performs the equivalent
+   of:
+   - `RUNNING -> LOCKED` with a bounded lock;
+   - `LOCKED -> CHECKPOINTED`, moving GPU memory to host memory and releasing
```
harche: Question: when GPU memory is moved into host memory here, is it charged to the **container's** memory cgroup? If so, a container with, say, a 16Gi limit using a 40GB GPU could be OOM-killed mid-checkpoint (and similarly on restore). The risk section ([L340](https://github.com/kubernetes/enhancements/blob/e7d20c54b4fe6fd577790022140452bbb69bdbf5/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L340)) covers node-level pressure; maybe add a pre-check or document this failure mode?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+   before CRI.
+3. Before invoking the expensive operation, kubelet validates the protected
+   ResourceClaimTemplate UID annotation and resolves the source allocation
+   against the latest complete ResourceSlice generation. The source device must
+   publish `driverStackID` and `memory`. When `DRAConsumableCapacity` is enabled,
+   `allowMultipleAllocations` must be false or unset. Missing, malformed, or
+   ambiguous provenance fails the checkpoint without calling the runtime.
+4. Kubelet suspends probes, resolves the sandbox ID, and calls CRI
+   `CheckpointPod` with that ID and a bounded deadline. The runtime enumerates
+   every container selected by the KEP-5823 Pod-level contract, pauses the
```
harche: Related to [Open Question 2](https://github.com/kubernetes/enhancements/blob/e7d20c54b4fe6fd577790022140452bbb69bdbf5/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L1117-L1118): if the all-container pause uses the cgroup freezer, I wonder if the CUDA lock can still be serviced (it needs to act inside the process). If not, the GPU lock may effectively have to happen *before* the freeze, which would make the ordering a design constraint rather than an implementation detail. Would be great to see the PoC results on this.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+this status for a transient internal or host failure. After runtime cleanup
+completes, kubelet sets `Restoring=False` with reason
+`DeviceRestoreIncompatible`, emits a warning event when it first records that
+condition, and immediately suppresses further calls in memory. Once the
+condition is persisted, kubelet does not call `RestorePod` again for that Pod
+UID, including after restart. If kubelet fails between the CRI response and the
+status update, recovery may attempt restore again. Kubernetes provides
+at-least-once execution in that window; the runtime's all-or-nothing cleanup and
+idempotency requirements keep it safe, but the KEP does not claim exactly-once
+execution. The Pod remains `Pending`; this is a terminal restore outcome for
```
harche: I'm a bit worried about this case: e.g. a Deployment creates a restore Pod, the runtime says `FailedPrecondition`, and the Pod stays `Pending` forever while holding a GPU claim. The owning controller won't replace it, because Pending isn't a failure. Would moving the Pod to `Failed` be a better terminal state here? Also, [KEP-5823](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5823-pod-level-checkpoint-restore) always retries, so "never retry" is new kubelet behavior worth calling out explicitly.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+  transferring its allocation is not hidden inside this enhancement.
+- **GPU memory can create large host-memory and disk pressure.** Implementations
+  must honor the parent checkpoint timeout and concurrency controls, expose
+  duration and checkpoint-size metrics, fail cleanly on insufficient resources,
+  and avoid unbounded parallel GPU checkpoints.
+- **Checkpoint support can change with the node software stack.** The DRA
+  driver publishes a driver stack ID instead of asking Kubernetes to interpret
+  version strings. The runtime also checks the exact archive format and files,
+  and supported combinations remain part of the implementation's test matrix.
+- **Raw device identity can reveal node inventory.** GPU UUIDs and the detailed
```
harche: Minor wording suggestion: the NVIDIA driver already publishes `uuid`, `productName`, `architecture`, `driverVersion`, `cudaDriverVersion` and the PCI bus ID in every ResourceSlice ([deviceinfo.go#L168-L205](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/deviceinfo.go#L168-L205)). So keeping them out of `PodCheckpoint.status` is fine, but they're already visible in Kubernetes. Maybe soften this and the related Alternatives text?

That also suggests an alternative worth listing: build the exact-match filter from attributes drivers already publish (or a driver-published hash of them), which might need little or no driver change for Alpha.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+logical DRA claim and request, adds the DRA driver and
+`ResourceClaimTemplate` UID from trusted Kubernetes state, and stores the
+result in `PodCheckpoint.status`. The ResourceClaim controller records that UID
+when it creates the source claim, using the protected
+`resource.kubernetes.io/resource-claim-template-uid` annotation. An in-tree
+admission check rejects that reserved key in
+`ResourceClaimTemplate.spec.metadata.annotations`. On a Pod-owned ResourceClaim
+create that contains the key, admission performs a synthetic authorization
+check for the `set` verb on the virtual
+`resourceclaims/template-provenance` subresource. Only the built-in
```
harche: Could you share the threat this provenance mechanism protects against? It adds a reserved annotation, admission, a virtual `template-provenance` subresource for authz, and immutability rules. Meanwhile the Pod spec already pins the template name, the appended selector constrains the device, and the runtime does the final check. If the main concern is "template deleted and recreated", would a hash of the template spec be enough? If it stays, [sig-auth](https://github.com/kubernetes/community/tree/master/sig-auth) would probably want to review it.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+
+`DeviceClass.spec` is mutable and is deliberately resolved under current
+cluster policy when the restore claim is allocated. The controller retains the
+template's class reference but does not claim that the referenced class still
+has its source-time contents. A class change may make the Pod unschedulable or
+cause final runtime validation to fail. In this KEP, "only narrow" means that
+the controller's synthesized claim never removes or relaxes a field copied from
+the current matching template; it does not make `DeviceClass` immutable.
+
+This is a deliberate exception to the current `ResourceClaimTemplateSpec`
```
harche: +1 on calling this out. Today the controller copies the template as-is:

```go
Spec: template.Spec.Spec,
```
[controller.go#L1142](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/controller/resourceclaim/controller.go#L1142)

Since this changes the DRA API contract, it might be good to add [WG Device Management](https://github.com/kubernetes/community/tree/master/wg-device-management) / DRA approvers explicitly (the kep.yaml currently lists sig-scheduling, though the scheduler needs no change).

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+Alpha requires exactly one `memory` entry under the source driver's domain. Its
+value is derived from the source device's trusted ResourceSlice after kubelet
+requires exact agreement with the runtime response. Alpha also requires one
+`Exactly` request whose allocation mode is unset or `ExactCount`, whose
+effective count is one, and whose `adminAccess` is not true. Across the entire
+Pod there must be exactly one DRA allocation result and one checkpointable
+device. When `DRAConsumableCapacity` is enabled, that device's
+`allowMultipleAllocations` must be false or unset; when it is disabled, DRA
+cannot advertise shareable devices through that field. In both cases the DRA
+driver must omit `driverStackID` from shared or partitioned devices. Kubelet
```
harche: There's no common way across drivers to say "full, exclusive device". NVIDIA uses `type: gpu|mig|vfio` ([deviceinfo.go#L170-L172](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/deviceinfo.go#L170-L172)), while dra-example-driver uses `full` / `partitionable` booleans ([gpu.go#L157-L158](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/internal/profiles/gpu/gpu.go#L157-L158)). Would a standard `resource.kubernetes.io/...` attribute (or the existing `ResourceSlice.PartitionTypeAttribute`, [types.go#L200](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/apis/resource/types.go#L200)) be a more vendor-neutral fit?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+- device-bearing checkpoint cannot succeed after omitting required runtime
+  device state;
+- every supported checkpointed allocation has a complete bounded response
+  entry, and an ambiguous association is rejected;
+- restore receives complete container configs and either succeeds or returns a
+  defined gRPC failure;
+- partial checkpoint output and partially restored runtime objects are cleaned;
+- restore never falls back to a fresh process.
+
+The precise conformance mechanism remains unresolved because the generic CRI
```
harche: Suggestion: [dra-example-driver](https://github.com/kubernetes-sigs/dra-example-driver) could be a nice vendor-neutral conformance target. It already publishes `uuid`, `model`, `driverVersion`, `memory`, and full/partition devices, so adding a fake `driverStackID` would be a small patch. Its per-claim env vars would also exercise the env-mismatch case (see [L816](https://github.com/kubernetes/enhancements/blob/e7d20c54b4fe6fd577790022140452bbb69bdbf5/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L816)).

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+- timeout and cancellation leave no Ready checkpoint or partial runtime object;
+- feature-gate disablement rejects GPU operations and leaves CPU-only operations
+  working;
+- device-plugin GPU Pods are rejected before a checkpoint or restore CRI call;
+- containerd and CRI-O are exercised as their KEP-5823 implementations become
+  available.
+
+### Graduation Criteria
+
+#### Alpha
```
harche: Maybe add an explicit dependency here: on k/k master only the [KEP-5823](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/5823-pod-level-checkpoint-restore) CRI RPCs have merged ([api.proto#L175-L183](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L175-L183)) so far. There's no `PodCheckpoint` type or `spec.restoreFrom` yet, so this KEP's API changes depend on those landing first.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
```diff
@@ -0,0 +1,1825 @@
...
+   labels.
+3. Check the Pod and `PodCheckpoint` conditions and first warning event.
+4. Correlate kubelet, runtime, CRIU plugin, and GPU driver logs using the Pod UID
+   and operation time; do not publish GPU UUIDs in shared telemetry.
+5. Verify the node's tested version matrix, host memory, disk space, and target
+   device compatibility.
+6. Stop new device checkpoint operations or disable the device gate if source
+   health failures or resource exhaustion are increasing.
+
+## Drawbacks
```
harche: Nit: the template's `Implementation History` section seems to be missing. PRR reviewers also usually prefer the exact template wording for the PRR question headings.

rst0git:
> can you fix the failing checks on this PR

Thank you Kevin! We needed to add an approver for the CI checks to pass. I asked at the SIG Node meeting yesterday, and Peter suggested to add Mrunal. It should pass now.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche, line 209: The delete-and-recreate case ([L689-L695](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L689-L695)) makes sense. Could a simpler check do the same job, though? When the controller creates a claim from a template, the claim's spec is an exact copy of the template's spec ([controller.go#L1142](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/controller/resourceclaim/controller.go#L1142)), and it can never change afterwards ([validation.go#L129-L132](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/apis/resource/validation/validation.go#L129-L132)). Kubelet already fetches that claim ([manager.go#L297](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/kubelet/cm/dra/manager.go#L297)) and is allowed to read it ([graph.go#L418-L419](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/plugin/pkg/auth/authorizer/node/graph.go#L418-L419)). So at checkpoint time kubelet could simply save the claim's spec into `PodCheckpoint.status`, next to the `checkpointedPodTemplate` that [KEP-5823](https://github.com/kubernetes/enhancements/blob/eaaf899265d4120f4de8653e81fafae68c34d850/keps/sig-node/5823-pod-level-checkpoint-restore/README.md?plain=1#L701-L707) already saves there. At restore time, the controller (which already reads templates, [controller.go#L134](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/controller/resourceclaim/controller.go#L134)) would compare that saved spec with the current template: same content, restore goes ahead; different content, restore is refused. If storing the whole spec is too big, storing a hash of it would work too (like Deployments' `pod-template-hash`), though then the error can only say "something changed", not what.

That would remove three new mechanisms from the KEP: the reserved annotation, the admission check and the `template-provenance` permission check. Safety-wise, if someone creates a claim for their Pod by hand with a different spec (the controller accepts such claims without checking the spec, [controller.go#L1189-L1215](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/controller/resourceclaim/controller.go#L1189-L1215)), the saved spec won't match the template, so only that Pod's restore is refused; nothing else is affected. If the UID approach stays, [sig-auth](https://github.com/kubernetes/community/tree/master/sig-auth) would probably want to review the new synthetic authorization; the pattern is similar to the `signers` check used for CSR approval ([certauthorization.go#L66](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/certauthorization/certauthorization.go#L66)).

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Small suggestion for Alternatives: drivers already publish attributes like `driverVersion`, `cudaDriverVersion`, `architecture` and `productName` (NVIDIA [deviceinfo.go#L168-L205](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/deviceinfo.go#L168-L205)). An exact-match filter built from those (or a driver-published hash of them) might need little or no driver change for Alpha, and the runtime's private manifest check would still cover `libcuda.so` file identity. Might be worth listing as a considered alternative?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: There's no common way across drivers to say "full, exclusive device". NVIDIA uses `type: gpu|mig|vfio` ([deviceinfo.go#L170-L172](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/deviceinfo.go#L170-L172)), while dra-example-driver uses `full` / `partitionable` booleans ([gpu.go#L157-L158](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/internal/profiles/gpu/gpu.go#L157-L158)). Would a standard `resource.kubernetes.io/...` attribute (or the existing `ResourceSlice.PartitionTypeAttribute`, [types.go#L200](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/apis/resource/types.go#L200)) be a more vendor-neutral fit?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche, line 572: Kubelet will read the source device's ResourceSlice here (as noted in the PRR section). One thing I couldn't find: the node authorizer only lets kubelet read slices that name its own node. Listing requires the `spec.nodeName=<self>` field selector:

```go
case "watch", "list", "deletecollection":
	// only allow a scoped fieldSelector
	...
	if req.Field == resourceapi.ResourceSliceSelectorNodeName && req.Operator == selection.Equals && req.Value == nodeName {
```
[node_authorizer.go#L381-L385](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/plugin/pkg/auth/authorizer/node/node_authorizer.go#L381-L385)

and a `get` by name only works when the slice has `nodeName` set:

```go
// if we have a node, establish new edges
if len(nodeName) > 0 {
```
[graph.go#L596-L602](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/plugin/pkg/auth/authorizer/node/graph.go#L596-L602)

So a device published in a slice that uses `nodeSelector` or `allNodes` (instead of `nodeName`; a slice sets exactly one of these) wouldn't be readable here. IMO no real driver uses those today: the NVIDIA GPU driver, dra-example-driver, [DRANET](https://github.com/kubernetes-sigs/dranet) and [dra-driver-cpu](https://github.com/kubernetes-sigs/dra-driver-cpu) all publish node-local slices, and I only found `nodeSelector` / `allNodes` in the k/k e2e tests. But since the API supports it, it might be worth stating "the source device must be in a node-local ResourceSlice" as an Alpha assumption, so it's not a surprise later.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: The KEP says the runtime picks the containers, but in the merged CRI kubelet passes the list (`container_ids`) and the runtime must checkpoint exactly those:

```proto
// IDs of the containers to include in the checkpoint. The list must be
// non-empty, must not contain duplicates, and must contain exactly the
// running containers selected by the caller.
repeated string container_ids = 3;
```
[api.proto#L2189-L2194](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L2189-L2194)

So the "exactly one GPU container" rule could be defined against `container_ids`.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Related to [Open Question 2](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L1117-L1118): CRIU's own flow locks the GPU *before* freezing. With cgroup freeze it runs the `PAUSE_DEVICES` plugin hook (the CUDA lock) and only then `freeze_processes()`, and it has a separate path for containers that are already frozen:

```c
if (opts.freeze_cgroup && !compel_interrupt_only_mode) {
	ret = run_plugins(PAUSE_DEVICES, pid);
	...
	if (freeze_processes())
		goto err;
} else {
	...
	/*
	 * Call PAUSE_DEVICES after prepare_freezer_for_interrupt_only_mode()
	 * to be able to checkpoint containers in a frozen state.
	 */
	ret = run_plugins(PAUSE_DEVICES, pid);
```
[seize.c#L1068-L1084](https://github.com/checkpoint-restore/criu/blob/33482a1af18bd9d2868a7c8b22f34e48c91ccd33/criu/seize.c#L1068-L1084)

So if the runtime pauses all containers first (the KEP-5823 all-container pause), CRIU would take the frozen-container path. It would be great if the PoC confirms that path works for the GPU container while the rest of the Pod stays paused. If it only works in a particular order, the ordering becomes a design constraint rather than an implementation detail.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Question: when GPU memory is moved into host memory here, is it charged to the **container's** memory cgroup? If so, a container with, say, a 16Gi limit using a 40GB GPU could be OOM-killed mid-checkpoint (and similarly on restore). The risk section ([L340](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md#L340)) covers node-level pressure; maybe add a pre-check or document this failure mode?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Would it be simpler for kubelet to **take** `driverStackID` and `memory` from the ResourceSlice (it already reads it here), rather than requiring the runtime to compute the same values and cross-checking them?

Today the runtime/CRIU plugin and the DRA driver have to implement the same ID algorithm independently. If one is upgraded without the other, every checkpoint fails. With the slice as the source:
- the runtime only needs to confirm that device state was captured (so we still never get a CPU-only "success"), and
- it keeps its own private fingerprint for the final check at restore.

That might also make most of the new `CheckpointPodResponse` fields unnecessary. What do you think?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: +1 on calling this out. Today the controller copies the template as-is:

```go
Spec: template.Spec.Spec,
```
[controller.go#L1142](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/controller/resourceclaim/controller.go#L1142)

Since this changes the DRA API contract, it might be good to add [WG Device Management](https://github.com/kubernetes/community/tree/master/wg-device-management) / DRA approvers explicitly (the kep.yaml currently lists sig-scheduling, though the scheduler needs no change).

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche, line 816: [KEP-5823](https://github.com/kubernetes/enhancements/blob/eaaf899265d4120f4de8653e81fafae68c34d850/keps/sig-node/5823-pod-level-checkpoint-restore/README.md?plain=1#L354-L356) already notes that env vars persist through restore. The new piece here is that CDI env edits for the new claim will differ from the checkpointed ones. For example, dra-example-driver injects:

```go
fmt.Sprintf("%s_DEVICE_%s_RESOURCE_CLAIM=%s", strings.ToUpper(cdi.class), deviceEnvKey, claimUID),
```
[cdi.go#L101](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/cmd/dra-example-kubeletplugin/cdi.go#L101), plus `GPU_DEVICE_<id>=<device>` ([gpu.go#L349](https://github.com/kubernetes-sigs/dra-example-driver/blob/ad2f6494498cbc609cff3e71332274d715fa6a7b/internal/profiles/gpu/gpu.go#L349)).

The restored Pod has a new claim UID (and maybe `gpu-1` instead of `gpu-0`), and the runtime applies restore-time CDI edits ([L653-L654](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L653-L654)), but the process environment comes back from the checkpoint. For the Alpha target this is benign: the NVIDIA driver's per-claim CDI spec carries only constant env vars, `NVIDIA_VISIBLE_DEVICES=void` from nvcdi's common edits ([wrapper.go#L99](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/vendor/github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi/wrapper.go#L99), included via [cdi.go#L291](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/cmd/gpu-kubelet-plugin/cdi.go#L291)) and `NVIDIA_CTK_LIBCUDA_DIR` ([driver-nvml.go#L123-L127](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/2b0257ccc35d54968756fb0db10c6740ae923826/vendor/github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi/driver-nvml.go#L123-L127)), which are identical for a same-node restore; the per-device edits add none. But the contract is vendor-neutral. Could the KEP say whether runtimes should ignore those env edits or reject the restore when they differ?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche, line 850: [KEP-5823](https://github.com/kubernetes/enhancements/blob/eaaf899265d4120f4de8653e81fafae68c34d850/keps/sig-node/5823-pod-level-checkpoint-restore/README.md?plain=1#L444-L445) says kubelet owns this directory ("for storage accounting and path confinement"), and the merged CRI agrees:

```proto
// Absolute path to an existing, empty directory where the runtime must
// write the checkpoint. The caller owns the directory, ...
string output_path = 2;
```
[api.proto#L2181-L2188](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/staging/src/k8s.io/cri-api/pkg/apis/runtime/v1/api.proto#L2181-L2188)

Would "runtime-defined contents in a kubelet-owned directory" be more accurate than "runtime-owned directory"?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche, line 966: There seem to be two cases here. If the GPU mismatch is noticed at scheduling time, the Pod just waits and starts on its own once the driver publishes a matching GPU again ([L949-L952](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L949-L952)), like [KEP-5823](https://github.com/kubernetes/enhancements/blob/eaaf899265d4120f4de8653e81fafae68c34d850/keps/sig-node/5823-pod-level-checkpoint-restore/README.md?plain=1#L1227-L1232)'s "fix it and it recovers" model. But if the mismatch is only noticed by the runtime after the GPU was already allocated ([L1676-L1678](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L1676-L1678)), the Pod is never retried. So even after the admin puts the node stack back (your own mitigation at [L1680](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L1680)), the Pod stays stuck and has to be deleted and recreated ([L969-L974](https://github.com/kubernetes/enhancements/blob/0a2f78e3dfbbecadb2218b99fb150cb3c8f1ce9d/keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md?plain=1#L969-L974)).

I understand retries are expensive. Would an exponential backoff with a much higher cap for this case (say, up to 30-60 minutes, instead of the normal 5-minute [`MaxCrashLoopBackOff`](https://github.com/kubernetes/kubernetes/blob/d176e72c54c260cc4edc8b2aa44bc37e2850d552/pkg/kubelet/kubelet.go#L166-L167)) be an option, so a fixed node revives the Pod without a tight loop?

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Would [dra-example-driver](https://github.com/kubernetes-sigs/dra-example-driver) work as a vendor-neutral conformance target here? It already publishes `uuid`, `model`, `driverVersion` and `memory` (and optionally full/partition devices), so adding a fake `driverStackID` might be a small patch. Its per-claim CDI env vars could also exercise the restore-env case ([my comment on L816](https://github.com/kubernetes/enhancements/pull/6275#discussion_r4084301018)). Happy to help if that's useful.

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Small question: was the template's `Implementation History` section left out intentionally? Also, would it be worth keeping the PRR question headings in the exact template wording? PRR reviewers usually look for those.

harche (commented):
Thanks for putting this together

keps/sig-node/6274-pod-level-checkpoint-restore-gpu/README.md
harche: Correction to my "no real driver uses those today": I just found one. [CoHDI/composable-dra-driver](https://github.com/CoHDI/composable-dra-driver), the DRA driver for composable disaggregated infrastructure, publishes fabric-attached devices (GPUs included) in pools with a `nodeSelector` rather than `nodeName` ([manager.go#L599-L613](https://github.com/CoHDI/composable-dra-driver/blob/fa9cd04bc2728f49e940288bbe556387e948c4a4/pkg/manager/manager.go#L599-L613)), with `BindsToNode` and the KEP-5007 binding conditions on each device ([manager.go#L588-L591](https://github.com/CoHDI/composable-dra-driver/blob/fa9cd04bc2728f49e940288bbe556387e948c4a4/pkg/manager/manager.go#L588-L591)). The other drivers I checked (NVIDIA, AMD, Intel, Google TPU, HAMi, DRANET, SR-IOV, OVS-DPDK, dra-driver-cpu, IBM Power, dra-example-driver) all publish node-local slices.

So a GPU allocated through a composable-infrastructure driver is a plausible future target for this KEP, and its ResourceSlice wouldn't be readable by kubelet. That makes stating the node-local assumption for Alpha a bit more relevant than I first suggested.

kannon92:
/assign @haircommander @mrunalp 
for sig-node

/assign
for PRR

I will wait for sig review before I jump in on PRR side.

Issue #6285, opened 2026-08-18
Structured configuration for x509 authentication


liggitt:
### Structured x509 authentication configuration

- One-line enhancement description: Expanded x509 client certificate authentication options
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6286
- Discussion Link: [sig-auth 2026-08-12 meeting](https://docs.google.com/document/d/1woLGRoONE3EBVx-wTb4pvp4CI7tmLZ6lS26VTbosLKM/edit?tab=t.0#bookmark=id.ycmbkggfqax4)
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [ ] KEP (`k/enhancements`) update PR(s):
      - [ ] https://github.com/kubernetes/enhancements/pull/6286
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

tico88612:
Hello @liggitt :wave:, 1.38 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 22nd September 2026 (AoE) / Wednesday 23th September 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.38 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [ ] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [ ] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

For this KEP, we would just need to update the following:
- PRR questionnaire filled out.
- Needs PRR approval file

[...]

Issue #6283, opened 2026-08-18
Leveraging certificate APIs in the platform


stlaz:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Use the new Kubernetes certificate APIs to enhance the platform.
- Kubernetes Enhancement Proposal: 
- Discussion Link: SIG Auth tracks platform serving signer as a future item in its KEP board: https://github.com/orgs/kubernetes/projects/262/views/7
- PRs by stage and milestone:
  - [ ] Alpha - v1.xx
    - [ ] KEP (`k/enhancements`) update PR(s):
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

/sig auth

tico88612:
Hello @stlaz :wave:, 1.38 Enhancements team here.

This is a reminder of the upcoming [KEP readiness (formerly named PRR freeze)](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 22nd September 2026 (AoE) / Wednesday 23th September 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.38 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
- [x] PR open or merged with [kep.yaml](https://github.com/kubernetes/enhancements/blob/master/keps/NNNN-kep-template/kep.yaml) updated with the `stage`, `latest-milestone`, and `milestone` struct filled out.
- [x] PR open or merged with a [PRR approval file](https://github.com/kubernetes/enhancements/blob/master/keps/prod-readiness/template/nnnn.yaml) with the PRR approver listed for the stage the KEP is targeting.

[...]

PR #6286, opened 2026-08-18
KEP-6285: Initial draft

liggitt:
- One-line PR description: Initial draft of https://github.com/kubernetes/enhancements/issues/6285 design

- Issue link: https://github.com/kubernetes/enhancements/issues/6285

/sig auth


phoban01:
Story 1 is empty. Here is a use possible case:

> We run hosted control planes. Each worker node gets a hardware-attested X.509 SVID from an out-of-band identity agent before the kubelet starts. We cannot put that CA in `--client-ca-file` today, because any anchor in the bundle can assert any identity, `system:masters` included. Under this proposal the cluster CA sits in the first authenticator and the machine identity CA in a second, with `userValidationRules` setting the boundary on its username prefix and groups.

---

A question on the non-goal, "This KEP does not propose changing the mapping from X.509 attributes to user info".

That works where we control the issuer, since we can put a scoped name in the CN. It does not work where the operator cannot. An external PKI sets CN and O to suit itself, and an X.509 SVID carries its identity in the URI SAN. `userValidationRules` run against the final user object and return a boolean, so no rule can produce a username the subject does not already contain.

Is the non-goal for this iteration or for good? Not asking to widen the KEP but for understanding.

keps/sig-auth/6285-x509-authentication-configuration/README.md
```diff
@@ -0,0 +1,823 @@
...
+The content of the file referenced by `--client-ca-file` is merged into the `client-ca-file` key of the `extension-apiserver-authentication` ConfigMap in the `kube-system` namespace, which is read by extension API servers to configure their own client certificate authentication.
+
+When configured via the `x509` file stanza instead, all the content from authenticators without `userValidationRules` will still be used this way.
+
+However, for authenticators with `userValidationRules` specified, it is unsafe to merge the content into an undifferentiated trust root bundle which extension API servers will allow to validate a certificate identifying *any* arbitrary user.
+
+Therefore, we will do the following:
+1. Require any unconditional (no `userValidationRules`) authorities be placed in the first authenticator in the `x509` list.
+2. Publish any unconditional authenticator bundle into the `client-ca-file` key, as before.
+3. Serialize the remaining conditional `x509` authenticators as `v1.X509Authenticator` objects into `authentication-config-x509-<index>` keys
```
liggitt, line 198: There was a question about publishing the CA content to a ClusterTrustBundle object, which I think is probably a good idea. The CA content *and* the user validation rules are also needed...

keps/sig-auth/6285-x509-authentication-configuration/README.md
```diff
@@ -0,0 +1,823 @@
...
+
+When configured via the `x509` file stanza instead, trust roots from all configured X.509 authenticators will be aggregated together for this purpose.
+
+#### Verifying presented client certificates
+
+The file referenced by `--client-ca-file` populates the trust roots used by the singular X.509 authenticator to verify presented X.509 certificates.
+
+When configured via the `x509` file stanza instead, a distinct X.509 authenticator instance will be constructed for each item in the `x509` list, with its own trust bundle to use for verification. All of the X.509 authenticators will be unioned together, and the first to verify a presented client certificate will short-circuit further evaluation.
+
+#### Publishing to extension-apiserver-authentication ConfigMap
```
liggitt, line 187: note to self: rationalize / make coherent with the request-header authentication configmap bits of https://github.com/kubernetes/enhancements/issues/6283

keps/sig-auth/6285-x509-authentication-configuration/README.md
```diff
@@ -0,0 +1,823 @@
...
+Consider the following in developing a version skew strategy for this
+enhancement:
+- Does this enhancement involve coordinating behavior in the control plane and nodes?
+- How does an n-3 kubelet or kube-proxy without this feature available behave when this feature is used?
+- How does an n-1 kube-controller-manager or kube-scheduler without this feature available behave when this feature is used?
+- Will any other components on the node change? For example, changes to CSI,
+  CRI or CNI may require updating that component before the kubelet.
+-->
+
+## Production Readiness Review Questionnaire
```
jpbetz, line 446: I'll take PRR review for this one. Go ahead and add the PRR yaml file.

PR #6298, opened 2026-08-25
KEP-6277: Add KEP for integrating Workload APIs with StatefulSet

sairameshv:
Add initial alpha KEP for integrating the Workload-Aware Scheduling APIs (Workload + PodGroup) with the StatefulSet controller, enabling gang scheduling for stateful workloads

  
- One-line PR description:
Integration of Workload APIs(Workload Aware Scheduling) with the StatefulSet Controller

- Issue link:
https://github.com/kubernetes/enhancements/issues/6277

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-apps/6277.yaml
+++ b/keps/prod-readiness/sig-apps/6277.yaml
@@ -1,3 +1,3 @@
 kep-number: 6277
 alpha:
-  approver: TBD
+  approver: kannon92
--- a/keps/prod-readiness/sig-apps/6277.yaml
+++ b/keps/prod-readiness/sig-apps/6277.yaml
@@ -1,0 +1,3 @@
+kep-number: 6277
+alpha:
+  approver: TBD
--- a/keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
+++ b/keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
@@ -1,0 +1,1375 @@
+# KEP-6277: Workload API Integration with StatefulSet
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Distributed Database with Quorum Requirements](#story-1-distributed-database-with-quorum-requirements)
+    - [Story 2: Distributed Coordination Service](#story-2-distributed-coordination-service)
+    - [Story 3: Stateful AI/ML Inference Serving](#story-3-stateful-aiml-inference-serving)
+    - [Story 4: Topology-Aware GPU Placement for Distributed Training](#story-4-topology-aware-gpu-placement-for-distributed-training)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Workload Ownership Model](#workload-ownership-model)
+  - [Pod Management Policy Constraints](#pod-management-policy-constraints)
+  - [Field Mapping](#field-mapping)
+  - [StatefulSet with Parallel Pod Management](#statefulset-with-parallel-pod-management)
+  - [StatefulSet with OrderedReady](#statefulset-with-orderedready)
+  - [StatefulSet with RollingUpdate.Partition](#statefulset-with-rollingupdatepartition)
+  - [Lifecycle Management](#lifecycle-management)
+    - [Initial Creation Lifecycle](#initial-creation-lifecycle)
+    - [Scale Lifecycle](#scale-lifecycle)
+  - [Enabling Gang Scheduling](#enabling-gang-scheduling)
+  - [Admission Validation](#admission-validation)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Explicit gangScheduling API Field on StatefulSet](#alternative-1-explicit-gangscheduling-api-field-on-statefulset)
+  - [Alternative 2: Create Workload by Default for All StatefulSets](#alternative-2-create-workload-by-default-for-all-statefulsets)
+- [Infrastructure Needed](#infrastructure-needed)
+- [References](#references)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements]
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes integrating the Kubernetes Workload API (`scheduling.k8s.io/v1alpha3`) with the
+StatefulSet controller to enable gang scheduling for stateful workloads. The StatefulSet controller
+will be extended to automatically create and manage `Workload` and `PodGroup` objects that describe
+pod groups and their scheduling requirements, allowing the kube-scheduler to make gang scheduling
+decisions for StatefulSet pods.
+
+Gang scheduling ensures that all pods in a group are scheduled atomically—either all pods are placed
+or none are—preventing partial deployments that could cause split-brain scenarios, inconsistent
+quorum states, or resource waste in distributed stateful systems such as databases, coordination
+services, and stateful AI/ML serving infrastructure.
+
+This integration follows the patterns established by the Job controller integration ([KEP-5547](https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job))
+and uses the reusable `workloadbuilder` library and `scheduling.k8s.io/v1alpha3` building blocks from ([KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis)).
+A new `scheduling` field is added to `StatefulSetSpec` to expose scheduling policy configuration.
+
+Gang scheduling requires `podManagementPolicy: Parallel`, as the default `OrderedReady` policy's
+sequential pod creation fundamentally conflicts with gang scheduling's all-at-once model.
+StatefulSets using `OrderedReady` will receive a Workload with `Basic` (non-gang) policy,
+preserving existing behavior while still enabling workload-level observability.
+
+Additionally, this KEP integrates ResourceClaim support for StatefulSet PodGroups, building on
+[KEP-5729](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5729-resourceclaim-support-for-workloads).
+A `resourceClaims` field on `StatefulSetSchedulingConfiguration` allows users to specify
+ResourceClaims or ResourceClaimTemplates that are shared across all pods in a StatefulSet's
+PodGroup. This enables use cases such as topology-aware GPU placement, where all pods in a
+StatefulSet must share the same topological domain device without requiring manual ResourceClaim
+management.
+
+## Motivation
+
+StatefulSets are the primary Kubernetes primitive for deploying stateful applications—distributed
+databases (etcd, CockroachDB, Cassandra), coordination services (ZooKeeper), message queues
+(Kafka), and increasingly stateful AI/ML inference serving systems. These workloads have a
+fundamental property that the current scheduler does not account for: their pods are
+interdependent and a partial deployment is often worse than no deployment at all.
+
+Today, when a StatefulSet with `podManagementPolicy: Parallel` is created, all replica pods are
+submitted to the scheduler independently. The scheduler places pods one at a time as resources
+become available. This leads to several problems:
+
+1. **Split-brain and quorum failures**: A distributed database configured for 5 replicas may have
+   only 3 pods scheduled, which is enough to form a quorum but not enough for the intended
+   fault-tolerance level. The 2 remaining pods may be stuck pending indefinitely, leaving the
+   cluster in a degraded state that operators may not detect until a failure occurs.
+
+2. **Resource waste from partial placement**: If only some pods of a tightly-coupled stateful
+   workload can be placed, those running pods consume resources without providing useful work,
+   since the application cannot function until all members are present.
+
+3. **Deadlock scenarios**: In a resource-constrained cluster, two StatefulSets may each get partial
+   placement, with neither able to acquire enough resources to run. Gang scheduling prevents this
+   by ensuring resources are committed atomically.
+
+The Workload API was introduced in
+[KEP-4671](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4671-gang-scheduling)
+to provide a standardized mechanism for describing groups of pods that must be scheduled together.
+[KEP-5832](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5832-decouple-podgroup-api)
+decoupled `PodGroup` as a standalone runtime object, and
+[KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis)
+introduced reusable `scheduling.k8s.io/v1alpha3` building blocks and a shared `workloadbuilder`
+library for controller integration. Together these provide the `Workload` (static template),
+`PodGroup` (runtime scheduling unit), and scheduling policies (`Gang` and `Basic`) that this KEP
+consumes.
+
+The [Workload API Integration Design](https://docs.google.com/document/d/1-lBmKIlgcCggjBWYO0VkZ8t3NmvJp-LCsfP6IB0XKRY)
+document (by Heba Elayoty and Andrey Velichkevich) provides the cross-controller integration
+blueprint. This KEP focuses specifically on the StatefulSet controller integration, following the
+patterns established by the Job controller integration in KEP-5547 and aligning with the broader
+effort to integrate all true workload controllers with the Workload API.
+
+### Goals
+
+- Enable the StatefulSet controller to automatically create and manage `Workload` and `PodGroup`
+  objects when gang scheduling is requested via a new `scheduling` field on `StatefulSetSpec`.
+
+- Support gang scheduling for StatefulSets using `podManagementPolicy: Parallel`, where all replica
+  pods must be schedulable before any are bound to nodes.
+
+- Provide a `Basic` (non-gang) Workload policy for StatefulSets using `podManagementPolicy:
+  OrderedReady`, enabling workload-level observability without altering their sequential scheduling
+  behavior.
+
+- Handle the StatefulSet lifecycle events correctly—including initial creation, scaling (up/down),
+  and `RollingUpdate` with `partition`—with proper Workload creation, recreation, and cleanup.
+
+- Follow the ownership model defined in the Workload API: one Workload and one PodGroup per
+  StatefulSet, both owned via `ownerReference` with `controller=true` and
+  `blockOwnerDeletion=true`.
+
+- Populate `pod.spec.schedulingGroup.podGroupName` on all pods created by the StatefulSet
+  controller, linking each pod to its PodGroup for scheduler consumption.
+
+- Use the shared `workloadbuilder` library from KEP-6089 for Workload and PodGroup construction,
+  ensuring consistency with other controller integrations (Job, JobSet).
+
+- Enable per-PodGroup ResourceClaim sharing via the `resourceClaims` field on
+  `StatefulSetSchedulingConfiguration`, allowing users to specify ResourceClaims or
+  ResourceClaimTemplates that are shared across all pods in the StatefulSet's PodGroup. The
+  controller populates `PodGroupTemplate.resourceClaims` and creates pods with matching
+  `spec.resourceClaims` entries so that the ResourceClaim controller and scheduler handle
+  allocation and reservation at the PodGroup level (per
+  [KEP-5729](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5729-resourceclaim-support-for-workloads)).
+
+### Non-Goals
+
+- Modifying the `OrderedReady` pod management policy to support gang scheduling. The sequential
+  creation semantics of `OrderedReady` are fundamentally incompatible with gang scheduling, and
+  changing them is out of scope.
+
+- Introducing changes to the Workload API itself (`scheduling.k8s.io`). This KEP consumes the
+  API as defined in KEP-4671; any API changes are tracked separately.
+
+- Supporting cross-namespace or cross-controller Workload ownership. The Workload is always
+  co-located in the same namespace as the StatefulSet and owned exclusively by it.
+
+- Implementing workload-aware preemption for StatefulSets. Preemption behavior is handled by
+  [KEP-5710](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5710-workload-aware-preemption)
+  and applies uniformly to all Workload-aware controllers.
+
+- Topology-aware scheduling for StatefulSet pods. This is covered by
+  [KEP-5732](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5732-topology-aware-workload-scheduling).
+
+- Automatic detection of whether a StatefulSet "should" use gang scheduling via heuristics. The
+  user must explicitly opt in by setting `spec.scheduling` on the StatefulSet.
+
+- Mutable `spec.scheduling` post-creation (toggle on/off, flip gang/basic, or change topology
+  constraints); scheduling configuration is immutable for Alpha. For Beta, `gang.minCount` will
+  become user-configurable (and mutable), while the scheduling policy type and topology
+  constraints remain immutable.
+
+- User-configurable `gang.minCount` — for Alpha, `minCount` is always derived from
+  `spec.replicas` to keep the initial implementation simple. Making `minCount` configurable
+  (allowing `minCount < replicas` for partial-gang semantics) is planned for Beta as a future
+  extension (see [Beta graduation criteria](#beta)).
+
+- Per-replica ResourceClaims analogous to `volumeClaimTemplates`, where each StatefulSet replica
+  gets its own dedicated ResourceClaim. This KEP only supports per-PodGroup ResourceClaims
+  (shared across all pods in a PodGroup). Per-replica ResourceClaim semantics would require
+  additional design and are deferred to future work.
+
+## Proposal
+
+The StatefulSet controller in `kube-controller-manager` will be extended to create and manage
+`Workload` and `PodGroup` objects based on the StatefulSet's configuration. A new `scheduling`
+field is added to `StatefulSetSpec`, embedding the reusable `scheduling.k8s.io/v1alpha3` building
+blocks (scheduling policy, topology constraints, disruption mode).
+
+Following the decoupled model from KEP-5832 and the controller pattern from KEP-6089, the
+integration uses the `workloadbuilder` library:
+
+```go
+// Scheduling, if set, opts this StatefulSet into Workload-Aware Scheduling.
+// The controller compiles one Workload and one PodGroup per StatefulSet with
+// gang minCount == replicas (Alpha; user-configurable in Beta).
+// Requires podManagementPolicy: Parallel for gang;
+// OrderedReady StatefulSets receive Basic (non-gang) policy.
+//
+// +featureGate=WorkloadWithStatefulSet
+// +optional
+// +k8s:ifDisabled(WorkloadWithStatefulSet)=+k8s:forbidden
+// +k8s:optional
+// +k8s:update=NoSet
+// +k8s:update=NoUnset
+Scheduling *StatefulSetSchedulingConfiguration `json:"scheduling,omitempty"`
+```
+
+`StatefulSetSchedulingConfiguration` reuses the `scheduling.k8s.io/v1alpha3` building-block types:
+
+```go
+type StatefulSetSchedulingConfiguration struct {
+    // SchedulingPolicy selects gang or basic scheduling. For Parallel
+    // StatefulSets the user sets gang: {}; for OrderedReady, only basic: {}
+    // is valid. For Alpha, the controller derives minCount from spec.replicas
+    // and a user-supplied minCount is rejected at admission. User-configurable
+    // minCount is planned for Beta.
+    // +optional
+    // +k8s:optional
+    // +k8s:update=NoSet
+    // +k8s:update=NoUnset
+    SchedulingPolicy *WorkloadPodGroupSchedulingPolicy
+
+    // SchedulingConstraints carries topology placement rules (e.g., zone
+    // co-location). Only meaningful with gang policy — basic-policy
+    // StatefulSets ignore topology constraints at the scheduler level.
+    // +optional
+    // +k8s:optional
+    // +k8s:immutable
+    SchedulingConstraints *WorkloadPodGroupSchedulingConstraints
+
+    // DisruptionMode controls whether preemption evicts a single pod or
+    // the entire gang. Stateful workloads that cannot tolerate partial
+    // membership (e.g., quorum-based systems) should use "all".
+    // +optional
+    // +k8s:optional
+    // +k8s:immutable
+    DisruptionMode *WorkloadPodGroupDisruptionMode
+
+    // ResourceClaims defines ResourceClaims or ResourceClaimTemplates that
+    // are shared across all pods in the StatefulSet's PodGroup. Each entry
+    // maps to a PodGroupResourceClaim in the PodGroupTemplate and PodGroup.
+    // Pods consume these shared claims by defining a matching entry in their
+    // own spec.resourceClaims (same name and same ResourceClaimName or
+    // ResourceClaimTemplateName). The ResourceClaim controller creates one
+    // ResourceClaim per PodGroup for each ResourceClaimTemplate reference,
+    // and the claim is reserved for the PodGroup rather than individual pods.
+    //
+    // This field requires the DRAWorkloadResourceClaims feature gate
+    // (KEP-5729) to be enabled in kube-apiserver, kube-controller-manager,
+    // kube-scheduler, and kubelet.
+    //
+    // +optional
+    // +k8s:optional
+    // +k8s:immutable
+    // +featureGate=DRAWorkloadResourceClaims
+    ResourceClaims []PodGroupResourceClaim
+}
+```
+
+When the StatefulSet controller detects that `spec.scheduling` is non-nil, it:
+
+1. Creates a `Workload` object (static template defining the PodGroup template).
+2. Creates a `PodGroup` object (runtime scheduling unit, snapshot of the Workload's template).
+3. Waits for the PodGroup to be observed in the informer cache.
+4. Creates pods with `spec.schedulingGroup.podGroupName` referencing the PodGroup.
+5. The scheduler uses the PodGroup's gang policy to hold pods in the `Permit` phase until
+   `MinCount` pods are schedulable, then binds them simultaneously.
+
+For StatefulSets with `podManagementPolicy: OrderedReady`, the controller creates a Workload and
+PodGroup with `Basic` policy (no gang semantics), preserving the existing sequential creation
+behavior while enabling workload-level observability and future scheduling optimizations.
+
+### User Stories
+
+#### Story 1: Distributed Database with Quorum Requirements
+
+As a platform engineer deploying a 5-node CockroachDB cluster on Kubernetes using a StatefulSet, I
+want all 5 database pods to be scheduled atomically so that the cluster starts with full
+replication factor and never enters a degraded state due to partial scheduling. Without gang
+scheduling, 3 of 5 pods might get scheduled, forming a quorum but with reduced fault tolerance,
+while the remaining 2 pods stay pending indefinitely due to resource fragmentation.
+
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: cockroachdb
+spec:
+  serviceName: cockroachdb
+  replicas: 5
+  podManagementPolicy: Parallel
+  scheduling:
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (5)
+    disruptionMode:
+      all: {}
+  template:
+    spec:
+      containers:
+      - name: cockroachdb
+        image: cockroachdb/cockroach:v23.2
+        resources:
+          requests:
+            cpu: "4"
+            memory: 8Gi
+```
+
+#### Story 2: Distributed Coordination Service
+
+As a cluster administrator running a ZooKeeper ensemble, I need all 3 ZooKeeper pods to start
+together to form a quorum. A ZooKeeper ensemble with only 1 or 2 out of 3 members running cannot
+elect a leader and is completely non-functional. Gang scheduling prevents wasting resources on a
+partial deployment that cannot serve any requests.
+
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: zookeeper
+spec:
+  serviceName: zk-headless
+  replicas: 3
+  podManagementPolicy: Parallel
+  scheduling:
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (3)
+  template:
+    spec:
+      containers:
+      - name: zookeeper
+        image: zookeeper:3.9
+```
+
+#### Story 3: Stateful AI/ML Inference Serving
+
+As an ML engineer deploying a sharded model inference service where each shard holds a portion of a
+large language model, I need all shards to be co-scheduled because the model cannot serve inference
+requests until every shard is available. Partial deployment wastes expensive GPU resources on shards
+that cannot function independently.
+
+```yaml
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: llm-inference-shards
+spec:
+  serviceName: llm-shards
+  replicas: 8
+  podManagementPolicy: Parallel
+  scheduling:
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (8)
+  template:
+    spec:
+      containers:
+      - name: inference-shard
+        image: my-org/llm-inference:latest
+        resources:
+          requests:
+            nvidia.com/gpu: 1
+```
+
+#### Story 4: Topology-Aware GPU Placement for Distributed Training
+
+As an ML engineer deploying a distributed training workload using a StatefulSet where each pod
+holds a model shard, I need all pods to be placed within the same network topology domain (e.g.,
+under the same spine switch) to minimize inter-shard communication latency. I want to express this
+as a shared ResourceClaim that represents the topological domain, so that all pods in my
+StatefulSet are co-located without managing ResourceClaims separately from the StatefulSet.
+
+```yaml
+apiVersion: resource.k8s.io/v1
+kind: ResourceClaimTemplate
+metadata:
+  name: topology-domain
+spec:
+  spec:
+    devices:
+      requests:
+      - name: domain
+        exactly:
+          deviceClassName: network-topology
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: training-shards
+spec:
+  serviceName: training-shards
+  replicas: 4
+  podManagementPolicy: Parallel
+  scheduling:
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (4)
+    resourceClaims:
+    - name: topo
+      resourceClaimTemplateName: topology-domain
+  template:
+    spec:
+      containers:
+      - name: trainer
+        image: my-org/distributed-trainer:latest
+        resources:
+          requests:
+            nvidia.com/gpu: 1
+          claims:
+          - name: topo
+      resourceClaims:
+      - name: topo
+        resourceClaimTemplateName: topology-domain
+```
+
+The StatefulSet controller creates a PodGroup with a `resourceClaims` entry referencing the
+`topology-domain` ResourceClaimTemplate. The ResourceClaim controller generates one ResourceClaim
+for the PodGroup, and the scheduler allocates a topological domain device once, shared by all 4
+pods. Each pod's `spec.resourceClaims` matches the PodGroup's claim, so the ResourceClaim is
+reserved for the PodGroup (not per-pod), and all pods are placed within the same domain.
+
+### Notes/Constraints/Caveats
+
+1. **OrderedReady incompatibility**: Gang scheduling is fundamentally incompatible with
+   `podManagementPolicy: OrderedReady` because `OrderedReady` creates pods sequentially (waiting
+   for each to become Ready before creating the next), which contradicts gang scheduling's
+   requirement that all pods be created and evaluated simultaneously. Admission validation MUST
+   reject a StatefulSet that specifies both `OrderedReady` and a gang scheduling policy in
+   `spec.scheduling.schedulingPolicy.gang`.
+
+2. **User-set `gang.minCount` is rejected for Alpha**: For Alpha, `minCount` is always derived
+   from `spec.replicas` by the controller, and a user-supplied `minCount` value is rejected at
+   admission. A gang is satisfiable once `minCount` pods can be scheduled, so allowing
+   `minCount < replicas` would enable useful partial-gang semantics (e.g., a 5-replica database
+   that can start with 3 members). Making `minCount` user-configurable is planned for Beta
+   (see [Beta graduation criteria](#beta)).
+
+3. **Mutable minCount on PodGroup**: Per KEP-5832, `minCount` on the standalone `PodGroup` object
+   is mutable. This means simple scaling operations (changing `spec.replicas`) can be handled by
+   updating the PodGroup's `minCount` in place, without the full drain-and-recreate pattern
+   described in the original design document. However, operations that change the PodGroup
+   structure (e.g., partition changes creating new PodGroups) still require PodGroup recreation.
+
+4. **schedulingGroup immutability on pods**: The `spec.schedulingGroup.podGroupName` field on pods
+   is immutable after creation. This means operations like removing a `RollingUpdate.partition`
+   that change which PodGroup a pod belongs to require pod recreation—there is no way to reassign
+   a running pod to a different PodGroup.
+
+5. **Finalizer requirement**: The PodGroup must have a finalizer to prevent premature deletion
+   while pods still reference it (KEP-5832 deletion protection). Without this, deleting a
+   PodGroup would orphan its pods, leaving them pointing to a non-existent PodGroup and causing
+   scheduler issues. The finalizer is added at PodGroup creation and removed only when all
+   referencing pods reach a terminal phase.
+
+6. **Feature gate dependencies**: This feature depends on the `GenericWorkload` feature gate
+   (introduced in KEP-4671, consolidated in 1.37) being enabled in both `kube-controller-manager`
+   and `kube-scheduler`. The `resourceClaims` field on `StatefulSetSchedulingConfiguration`
+   additionally requires the `DRAWorkloadResourceClaims` feature gate (KEP-5729) to be enabled
+   in kube-apiserver, kube-controller-manager, kube-scheduler, and kubelet. When
+   `DRAWorkloadResourceClaims` is disabled, the `resourceClaims` field is ignored and the
+   StatefulSet controller does not populate `resourceClaims` on the PodGroupTemplate or PodGroup.
+
+7. **ResourceClaim lifecycle tied to PodGroup**: ResourceClaims generated from
+   ResourceClaimTemplates referenced by a PodGroup are owned by the PodGroup (via
+   `ownerReferences`). When the PodGroup is deleted, the generated ResourceClaim is deallocated
+   and garbage-collected. For StatefulSets, this means that PodGroup recreation (e.g., due to
+   partition changes) causes the old ResourceClaim to be deleted and a new one generated for the
+   new PodGroup.
+
+8. **Maximum 4 ResourceClaims per PodGroup**: KEP-5729 limits the `resourceClaims` list to 4
+   entries per PodGroupTemplate/PodGroup. This limit applies to the StatefulSet's
+   `spec.scheduling.resourceClaims` field as well.
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| Partition changes require PodGroup recreation | Changing `RollingUpdate.partition` requires PodGroup recreation and pod recreation due to immutable `schedulingGroup` on pods | Document this behavior clearly; partition changes are infrequent and already involve rolling pod updates |
+| Workload object leak if StatefulSet controller crashes mid-creation | Orphaned Workload objects consume scheduler resources | OwnerReference with `blockOwnerDeletion=true` ensures garbage collection; finalizer prevents premature cleanup |
+| Users configure gang scheduling with `OrderedReady` policy | Unclear failure mode, pods created sequentially cannot satisfy gang | Admission validation rejects this combination at creation time with a clear error message |
+| Scheduler overhead from additional Workload objects | Increased memory and processing in kube-scheduler | Workloads are only created when explicitly requested via `spec.scheduling`; no Workload objects for StatefulSets that don't opt in |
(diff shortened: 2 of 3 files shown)


keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1141 @@
...
+1. **OrderedReady incompatibility**: Gang scheduling is fundamentally incompatible with
+   `podManagementPolicy: OrderedReady` because `OrderedReady` creates pods sequentially (waiting
+   for each to become Ready before creating the next), which contradicts gang scheduling's
+   requirement that all pods be created and evaluated simultaneously. Admission validation MUST
+   reject a StatefulSet that specifies both `OrderedReady` and a gang scheduling policy in
+   `spec.scheduling.schedulingPolicy.gang`.
+
+1. **User-set `gang.minCount` is rejected**: `minCount` is always derived from `spec.replicas`
+   by the controller. If `minCount > replicas`,
+   the gang can never be satisfied; if `minCount < replicas`, true partial-gang semantics require
```
kannon92: This isn't really true.

A gang can be scheduled if the `minCount` is satisfied.

I'm fine to introduce this for alpha but I think we should consider allowing folks to change `minCount` for beta.

Or we call it out as a future extension..

keps/prod-readiness/sig-apps/6277.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 6277
+alpha:
+  approver: TBD
```
kannon92:
```suggestion
  approver: kannon92
```

Fine to take this.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+    // and the claim is reserved for the PodGroup rather than individual pods.
+    //
+    // This field requires the DRAWorkloadResourceClaims feature gate
+    // (KEP-5729) to be enabled in kube-apiserver, kube-controller-manager,
+    // kube-scheduler, and kubelet.
+    //
+    // +optional
+    // +k8s:optional
+    // +k8s:immutable
+    // +featureGate=DRAWorkloadResourceClaims
```
kannon92, line 313: please make sure these fields match what we define in the job controller. You are missing quite a few items.

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/batch/v1/types.go#L547

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+When the StatefulSet controller detects that `spec.scheduling` is non-nil, it:
+
+1. Creates a `Workload` object (static template defining the PodGroup template).
+2. Creates a `PodGroup` object (runtime scheduling unit, snapshot of the Workload's template).
+3. Waits for the PodGroup to be observed in the informer cache.
+4. Creates pods with `spec.schedulingGroup.podGroupName` referencing the PodGroup.
+5. The scheduler uses the PodGroup's gang policy to hold pods in the `Permit` phase until
+   `MinCount` pods are schedulable, then binds them simultaneously.
+
+For StatefulSets with `podManagementPolicy: OrderedReady`, the controller creates a Workload and
```
kannon92, line 327: Would it make sense to just not set scheduling?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+   from `spec.replicas` by the controller, and a user-supplied `minCount` value is rejected at
+   admission. A gang is satisfiable once `minCount` pods can be scheduled, so allowing
+   `minCount < replicas` would enable useful partial-gang semantics (e.g., a 5-replica database
+   that can start with 3 members). Making `minCount` user-configurable is planned for Beta
+   (see [Beta graduation criteria](#beta)).
+
+3. **Mutable minCount on PodGroup**: Per KEP-5832, `minCount` on the standalone `PodGroup` object
+   is mutable. This means simple scaling operations (changing `spec.replicas`) can be handled by
+   updating the PodGroup's `minCount` in place, without the full drain-and-recreate pattern
+   described in the original design document. However, operations that change the PodGroup
```
kannon92: Can you remove this? The KEP is the source of truth for the design.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+   StatefulSet controller does not populate `resourceClaims` on the PodGroupTemplate or PodGroup.
+
+7. **ResourceClaim lifecycle tied to PodGroup**: ResourceClaims generated from
+   ResourceClaimTemplates referenced by a PodGroup are owned by the PodGroup (via
+   `ownerReferences`). When the PodGroup is deleted, the generated ResourceClaim is deallocated
+   and garbage-collected. For StatefulSets, this means that PodGroup recreation (e.g., due to
+   partition changes) causes the old ResourceClaim to be deleted and a new one generated for the
+   new PodGroup.
+
+8. **Maximum 4 ResourceClaims per PodGroup**: KEP-5729 limits the `resourceClaims` list to 4
```
kannon92, line 524: Again, I think you need to update the API to reflect this.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+  and PodGroup are fully removed, maintaining consistent cleanup ordering.
+- The PodGroup is created with a finalizer (KEP-5832 deletion protection) that is removed only
+  when all referencing pods reach a terminal phase.
+
+### Pod Management Policy Constraints
+
+| Pod Management Policy | Workload Policy | Gang Scheduling | Rationale |
+|----------------------|-----------------|-----------------|-----------|
+| `Parallel` | `Gang` | Yes | All pods created simultaneously—natural fit for gang semantics |
+| `OrderedReady` | `Basic` | No | Sequential pod creation conflicts with gang's all-at-once model |
```
kannon92, line 579: Is there a reason to allow disruption of the podgroup once scheduled?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1375 @@
...
+  template:
+    spec:
+      containers:
+      - name: zookeeper
+        image: zookeeper:3.9
+```
+
+#### Story 3: Stateful AI/ML Inference Serving
+
+As an ML engineer deploying a sharded model inference service where each shard holds a portion of a
```
kannon92, line 394: LWS composes Statefulsets. I'm curious on how this composing would work with LWS. We want to make sure that statefulsets never create PodGroups/Workloads if no scheduling spec is set.

kannon92:
Did a quick skim. There are a few inconsistencies but overall the KEP is looking really solid.

/cc @soltysh @mm4tt @helayoty @janetkuo

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Hmm, I agree that there is a no-op with this policy but to have a uniform view of every scheduling opted Statefulset has a Workload object, I feel we should have this.

But, I'm open for suggestions if we should leave this untouched for Alpha and think about it in Beta.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Sorry, I didn't get the context here. Do you mean to say if we should expose the `DisruptionMode` at all for StsSchedulingSpec?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Yes, Added explicitly in the caveat section

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
kannon92: I meant for `OrderedReady`. It makes sense that it has to be basic but can you have `DisruptionMode` of `None` or `All`.

sairameshv: Ah okay. Yes, for `OrderedReady` with `Basic` policy, the `DisruptionMode` may not make sense if it is `All` as there won't be any gang. `Single` is the default and expected mode/behavior

kannon92: Yea I realize this may be a separate thing. Basic would disable podgroup preemption but curious if we need something like "schedule one-by-one but if pods are preempted preempt as a gang".

Anyway this is probably out of scope but curious if there would be a need for this for sts/deployment?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
kannon92: I'd follow what Job does so we are consistent.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Added a note in the Pod Management Policy Constraints section that it is a no-op i.e. pods are always disrupted independently

keps/sig-apps/6277-integrate-workload-with-statefulset/kep.yaml
```diff
@@ -0,0 +1,50 @@
...
+participating-sigs:
+  - sig-scheduling
+status: provisional
+creation-date: 2026-08-25
+reviewers:
+  - "@kannon92"
+  - "@helayoty"
+  - "@soltysh"
+approvers:
+  - "@soltysh"
```
janetkuo, line 15: Please add me as well

sairameshv: Sure

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+- [ ] Supporting documentation
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP proposes integrating the Kubernetes Workload API (`scheduling.k8s.io/v1alpha3`) with the
```
mm4tt, line 82: NIT: Workload / PodGroup API has been promoted to v1beta1 in 1.37.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    spec:
+      containers:
+      - name: inference-shard
+        image: my-org/llm-inference:latest
+        resources:
+          requests:
+            nvidia.com/gpu: 1
+```
+
+#### Story 4: Topology-Aware GPU Placement for Distributed Training
```
mm4tt, line 421: This example is misleading / incorrect.

KEP-5729 (KEP-5729: DRA: ResourceClaim Support for Workloads) simply provides lifecycle management for sharing a single `ResourceClaim` across a `PodGroup` (avoiding the 256 `status.reservedFor` pod limit and generating one claim per PodGroup from a template). Influencing pod placement across topology domains (like racks or spine switches) is explicitly a non-goal of KEP-5729 and is the responsibility of KEP-5732 (`schedulingConstraints.topology`).

I'd suggest replacing this user story with a more standard DRA resource sharing scenario (e.g., sharing a pooled hardware device or shared memory across pods in the gang), and adding another user story for label based topology through KEP-5732 (`schedulingConstraints.topology`).

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+Gang scheduling requires `podManagementPolicy: Parallel`, as the default `OrderedReady` policy's
+sequential pod creation fundamentally conflicts with gang scheduling's all-at-once model.
+StatefulSets using `OrderedReady` will receive a Workload with `Basic` (non-gang) policy,
+preserving existing behavior while still enabling workload-level observability.
+
+Additionally, this KEP integrates ResourceClaim support for StatefulSet PodGroups, building on
+[KEP-5729](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5729-resourceclaim-support-for-workloads).
+A `resourceClaims` field on `StatefulSetSchedulingConfiguration` allows users to specify
+ResourceClaims or ResourceClaimTemplates that are shared across all pods in a StatefulSet's
+PodGroup. This enables use cases such as topology-aware GPU placement, where all pods in a
```
mm4tt, line 106: Same comment as in the in "Story 4" - this is incorrect. This integration will not enable topology aware scheduling (TAS) through DRA, but rather through KEP-5732 (`schedulingConstraints.topology`).

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+
+## Summary
+
+This KEP proposes integrating the Kubernetes Workload API (`scheduling.k8s.io/v1alpha3`) with the
+StatefulSet controller to enable gang scheduling for stateful workloads. The StatefulSet controller
+will be extended to automatically create and manage `Workload` and `PodGroup` objects that describe
+pod groups and their scheduling requirements, allowing the kube-scheduler to make gang scheduling
+decisions for StatefulSet pods.
+
+Gang scheduling ensures that all pods in a group are scheduled atomically—either all pods are placed
```
mm4tt, line 88: Integration of StatefulSet with Workload/PodGroup is not only about gang scheduling (`policy: gang/basic`), but also provides the following capabilities:

- `disruptionMode: single/all` (KEP-5710) — where `single` is the current behavior (treating StatefulSet pods as independent preemption units) and `all` treats the entire StatefulSet as an atomic preemption unit (i.e., gang preemption).
- TAS (Topology-Aware Scheduling) — KEP-5732 (`schedulingConstraints.topology`). The ability to specify topology constraints, e.g., "all pods of this StatefulSet should be scheduled in a single rack/zone".
- `resourceClaims` for PodGroups — KEP-5729. This is already mentioned, but the conclusion is inaccurate: it does not provide topology-aware scheduling through DRA on its own (see my other comment). Its primary purpose is allowing pods in a PodGroup to share a single `ResourceClaim`.

I would recommend rewriting this section to clearly distinguish between these capabilities and clarify the actual benefits this integration brings.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+to provide a standardized mechanism for describing groups of pods that must be scheduled together.
+[KEP-5832](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5832-decouple-podgroup-api)
+decoupled `PodGroup` as a standalone runtime object, and
+[KEP-6089](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis)
+introduced reusable `scheduling.k8s.io/v1alpha3` building blocks and a shared `workloadbuilder`
+library for controller integration. Together these provide the `Workload` (static template),
+`PodGroup` (runtime scheduling unit), and scheduling policies (`Gang` and `Basic`) that this KEP
+consumes.
+
+The [Workload API Integration Design](https://docs.google.com/document/d/1-lBmKIlgcCggjBWYO0VkZ8t3NmvJp-LCsfP6IB0XKRY)
```
mm4tt, line 146: The link doesn't work

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+document (by Heba Elayoty and Andrey Velichkevich) provides the cross-controller integration
+blueprint. This KEP focuses specifically on the StatefulSet controller integration, following the
+patterns established by the Job controller integration in KEP-5547 and aligning with the broader
+effort to integrate all true workload controllers with the Workload API.
+
+### Goals
+
+- Enable the StatefulSet controller to automatically create and manage `Workload` and `PodGroup`
+  objects when gang scheduling is requested via a new `scheduling` field on `StatefulSetSpec`.
```
mm4tt, line 155: Again, this is not only about gang scheduling, but "workload aware scheduling (WAS) capabilities" in general

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+
+- Supporting cross-namespace or cross-controller Workload ownership. The Workload is always
+  co-located in the same namespace as the StatefulSet and owned exclusively by it.
+
+- Implementing workload-aware preemption for StatefulSets. Preemption behavior is handled by
+  [KEP-5710](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5710-workload-aware-preemption)
+  and applies uniformly to all Workload-aware controllers.
+
+- Topology-aware scheduling for StatefulSet pods. This is covered by
+  [KEP-5732](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5732-topology-aware-workload-scheduling).
```
mm4tt, line 202: I disagree with listing Workload-Aware Preemption (WAP) and Topology-Aware Scheduling (TAS) as Non-Goals here. 

By that logic, Gang Scheduling itself would be a non-goal since its scheduling behavior is handled by KEP-4671. The entire purpose of this controller-integration KEP is to expose the WAS API surface (`schedulingPolicy`, `disruptionMode`, `schedulingConstraints`) on the StatefulSet, validate it, and compile it into `Workload`/`PodGroup` templates via `workloadbuilder`.

Supporting and validating these fields on the StatefulSet API is firmly **in scope**. What is out of scope is implementing or changing the underlying scheduler logic/plugins (which is handled upstream by KEP-5710 and KEP-5732).

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+// Requires podManagementPolicy: Parallel for gang;
+// OrderedReady StatefulSets receive Basic (non-gang) policy.
+//
+// +featureGate=WorkloadWithStatefulSet
+// +optional
+// +k8s:ifDisabled(WorkloadWithStatefulSet)=+k8s:forbidden
+// +k8s:optional
+// +k8s:update=NoSet
+// +k8s:update=NoUnset
+Scheduling *StatefulSetSchedulingConfiguration `json:"scheduling,omitempty"`
```
mm4tt, line 245: What are the defaults if scheduling is set but empty, i.e. `statefulset.spec.scheduling: {}` ?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+type StatefulSetSchedulingConfiguration struct {
+    // SchedulingPolicy defines the scheduling policy for this StatefulSet.
+    // Exactly one of Basic or Gang must be set.
+    // For Parallel StatefulSets the user sets gang: {}; for OrderedReady,
+    // only basic: {} is valid. For Alpha, the controller derives minCount
+    // from spec.replicas and a user-supplied minCount is rejected at
+    // admission. User-configurable minCount is planned for Beta.
+    // This field is immutable after creation: the policy may not be added or
+    // removed. The policy variant (basic/gang) is frozen by hand-written
+    // validation; only schedulingPolicy.gang.minCount may be changed.
```
mm4tt, line 260: > only schedulingPolicy.gang.minCount may be changed.

But this is not supported in alpha given the "For Alpha, the controller derives minCount
    // from spec.replicas and a user-supplied minCount is rejected " above, right?

Given that, should we just remove this add add in beta?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    //
+    // +optional
+    // +k8s:optional
+    // +k8s:update=NoSet
```
mm4tt, line 264: If we don't allow setting/changin MinCount in alpha, then wouldn't be simpler to just use `+immutable` here for alpha? We can relax the validation in further releases once we support setting/mutating MinCount.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    // This field is immutable after creation.
+    //
+    // +optional
+    // +k8s:optional
+    // +k8s:immutable
+    SchedulingConstraints *schedulingv1alpha3.WorkloadPodGroupSchedulingConstraints `json:"schedulingConstraints,omitempty" protobuf:"bytes,2,opt,name=schedulingConstraints"`
+
+    // DisruptionMode defines the mode in which the StatefulSet's pods can be
+    // disrupted. One of Single, All.
+    // Only meaningful with Gang policy — for Basic policy this field is a
```
mm4tt, line 279: To be precise, for basic policy it's illegal to set this to All. This is validated by the workloadbuilder - https://github.com/kubernetes/kubernetes/blob/45f110b546702be3bcbf8a1dff948a7902590938/staging/src/k8s.io/component-helpers/scheduling/schedulingv1/workloadbuilder/validate.go#L362

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+}
+```
+
+When the StatefulSet controller detects that `spec.scheduling` is non-nil, it:
+
+1. Creates a `Workload` object (static template defining the PodGroup template).
+2. Creates a `PodGroup` object (runtime scheduling unit, snapshot of the Workload's template).
+3. Waits for the PodGroup to be observed in the informer cache.
+4. Creates pods with `spec.schedulingGroup.podGroupName` referencing the PodGroup.
+5. The scheduler uses the PodGroup's gang policy to hold pods in the `Permit` phase until
```
mm4tt, line 324: > hold pods in the `Permit` phase 

This is no longer true and a lower level detail. The point is that scheduler uses PodGroup's gang policy (and other settings: topology, disruptionMode, ...) and guarantees all-or-nothing scheduling.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+   `MinCount` pods are schedulable, then binds them simultaneously.
+
+For StatefulSets with `podManagementPolicy: OrderedReady`, the controller creates a Workload and
```
mm4tt, line 327: What does it mean? I guess we should say that validation will reject statefulset if it tries to speficy `spec.scheduling.policy = gang {}` and `podManagementPolicy: OrderedReady`, right?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (3)
+  template:
+    spec:
+      containers:
+      - name: zookeeper
+        image: zookeeper:3.9
+```
+
+#### Story 3: Stateful AI/ML Inference Serving
```
mm4tt, line 392: It seems that stories 1-3 are essentially the same functional capability - "all or nothing scheduling". I would recommend providing examples for other feature like "disruptionMode: all" or "topology".

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    apiGroup: apps
+    kind: StatefulSet
+    name: my-ordered-statefulset
+  podGroupTemplates:
+  - name: my-ordered-statefulset
+    schedulingPolicy:
+      basic: {}
+```
+
+### StatefulSet with RollingUpdate.Partition
```
mm4tt, line 750: A few thoughts on the `RollingUpdate.partition` design:

1. Proper support for partitions and canary rollouts in combination with Workload-Aware Scheduling is tricky and will require further discussion and design exploration.
2. The proposed approach of splitting a StatefulSet into multiple PodGroups (`partition-old` and `partition-new`) is problematic if it forces pod recreations just to switch an immutable `pod.spec.schedulingGroup.podGroupName` (e.g. when partitions are shifted or removed). We should aim for a design where pods are never recreated purely for scheduling-group reassignment.
3. We should rethink the assumption that each partition must map 1:1 to a `PodGroup`. A `PodGroup` represents a "scheduling configuration" unit for the workload as a whole (e.g. quorum, topology, disruption), whereas a partition represents revision rollout stages. It seems plausible that this could be handled much more cleanly with a **single PodGroup** across the entire StatefulSet using mutable `minCount` (or preserving the gang across revisions), rather than tearing down and recreating PodGroups.


Given the complexity, open questions, and potential deadlocks with sequential pod updates, I'd strongly suggest **dropping partition support for Alpha**. We should simply validate and forbid setting `RollingUpdate.partition > 0` when `spec.scheduling` is configured, and defer full multi-partition semantics to Beta once the base single-PodGroup lifecycle is proven.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+When a Parallel StatefulSet is created with gang scheduling enabled:
+
+1. The StatefulSet controller detects `spec.scheduling` is non-nil.
+2. The controller creates the `Workload` object (static template with PodGroup template).
+3. The controller creates the `PodGroup` object (runtime unit, snapshot of Workload template).
+4. The controller waits for the PodGroup to be observed in the informer cache.
+5. The controller creates all pod templates with `spec.schedulingGroup.podGroupName` pointing to
+   the PodGroup.
+6. Pods enter the scheduler queue.
+7. The scheduler's `PreFilter` plugin looks up each pod's PodGroup to determine gang requirements.
```
mm4tt, line 804: This is incorrect, as gang scheduling does not use the PreFilter extension point. Furthermore, this is a low-level implementation detail that can be omitted here (same for Permit and other scheduler low level details)

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+are completely unaffected. There is no heuristic or automatic opt-in.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Disabling the gate has two effects depending on the component:
+
+- **kube-apiserver**: Clears `spec.scheduling` on new creates; preserves the stored value on
+  existing objects (standard alpha field-gating).
+- **kube-controller-manager**: Stops compiling Workload and PodGroup objects. Existing scheduling
+  objects become inert but remain in the cluster because they are owned by the StatefulSet.
```
mm4tt, line 1136: This is incorrect. As long as the `PodGroup` objects exist in the cluster and pods reference them via `spec.schedulingGroup.podGroupName`, the scheduler will continue enforcing WAS semantics like gang scheduling. Disabling the feature gate on `kube-controller-manager` only stops the controller from syncing or managing them, which can actually cause desynchronization (e.g. if replicas scale while `minCount` on the PodGroup is no longer updated), but the existing scheduling objects remain active from the scheduler's perspective.

mm4tt (commented):
Thanks!

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: That's actually partially true - yes this will enable TAS through the constraints.topology, but this will also allow to simulate TAS via DRA if a device structure reflects the needs.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+  creation semantics of `OrderedReady` are fundamentally incompatible with gang scheduling, and
+  changing them is out of scope.
+
+- Introducing changes to the Workload API itself (`scheduling.k8s.io`). This KEP consumes the
+  API as defined in KEP-4671; any API changes are tracked separately.
+
+- Supporting cross-namespace or cross-controller Workload ownership. The Workload is always
+  co-located in the same namespace as the StatefulSet and owned exclusively by it.
+
+- Implementing workload-aware preemption for StatefulSets. Preemption behavior is handled by
```
wojtek-t, line 197: This is a bit misleading point. In fact, via integration with Workload/PodGroup APIs you will get support for workload-aware preemption kind-of for free.
So I wouldn't say it's a non-goal.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: +1 - this is similar to what I wrote above about workload-aware-preemption.

You should probably change these point to something like:
"- Changes in the scheduling or preemption logic. We will rely on the feature surface provided by the scheduler."
or something like that.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+  and applies uniformly to all Workload-aware controllers.
+
+- Topology-aware scheduling for StatefulSet pods. This is covered by
+  [KEP-5732](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5732-topology-aware-workload-scheduling).
+
+- Automatic detection of whether a StatefulSet "should" use gang scheduling via heuristics. The
+  user must explicitly opt in by setting `spec.scheduling` on the StatefulSet.
+
+- Mutable `spec.scheduling` post-creation (toggle on/off, flip gang/basic, or change topology
+  constraints); scheduling configuration is immutable for Alpha. For Beta, `gang.minCount` will
```
wojtek-t, line 208: You have a separate point about gang.minCount below. Let's remove it from here and focus this point purely on gang/basic flipping (which we will not support even for GA).

The point below about gang.minCount we will revisit for beta.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: [I'm assuming these are all empty then, but it should be clarified.]

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    // +featureGate=DRAWorkloadResourceClaims
+    ResourceClaims []schedulingv1alpha3.WorkloadPodGroupResourceClaim `json:"resourceClaims,omitempty" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,4,rep,name=resourceClaims"`
+}
+```
+
+When the StatefulSet controller detects that `spec.scheduling` is non-nil, it:
+
+1. Creates a `Workload` object (static template defining the PodGroup template).
+2. Creates a `PodGroup` object (runtime scheduling unit, snapshot of the Workload's template).
+3. Waits for the PodGroup to be observed in the informer cache.
```
wojtek-t, line 322: Why?
The fact that it was observed in controllers cache doesn't even guarantee that it was observed by scheduler too, so that doesn't really bring any guarantees.

So it's completely enough to just get the (synchronous) response from the PodGroup creation API call and proceed from that.

[Given there are no guarantees of order delivery across object types, scheduler already supports observing pods before their corresponding PodGroup - in such case it just holds them before the PodGroup is observed to realize what scheduling mode should be used.]

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: +1 - let's remove this point. This is irrelevant for this KEP - you just rely on the functionality provided by the scheduler.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+4. Creates pods with `spec.schedulingGroup.podGroupName` referencing the PodGroup.
+5. The scheduler uses the PodGroup's gang policy to hold pods in the `Permit` phase until
+   `MinCount` pods are schedulable, then binds them simultaneously.
+
+For StatefulSets with `podManagementPolicy: OrderedReady`, the controller creates a Workload and
+PodGroup with `Basic` policy (no gang semantics), preserving the existing sequential creation
+behavior while enabling workload-level observability and future scheduling optimizations.
```
wojtek-t, line 329: The fact that you can't specify gang for `OrderedReady` mode doesn't mean you just get observability. You still can use TAS or PodGroup-level DRA in such case, which already provide value from the beginning.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+  name: cockroachdb
+spec:
+  serviceName: cockroachdb
+  replicas: 5
+  podManagementPolicy: Parallel
+  scheduling:
+    schedulingPolicy:
+      gang: {}          # minCount derived from spec.replicas (5)
+    disruptionMode:
+      all: {}
```
wojtek-t, line 354: This is very arguable to me.

The whole point of redundancy is to prevent a situation when a single pod failure is bringing down the whole DB.
So let's imagine a case when a higher priority workload comes and needs to be scheduled. If it's enough to preempt a single pod of my DB in such case, I would actually prefer it compared to preempting the whole DB and making it a complete outage for my users.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: +1, but let me clarify this comment a bit (as it may not be obvious for people not involved in those effort)

1. Providing PodGroup-level ResourceClaim will guarantee you that pods will be scheduled within the same domain (defined by RC).
2. However, it does NOT guarantee you that if such placement exists we will give it to you. Scheduler TAS algorithm still doesn't support DRA, so what happens is that scheduler will be scheduling pods one-by-one and the placement of the first pod will determine the domain for the rest of them. If we made the wrong decision for the first pod, it will not be corrected.
Given it's a gang, we will not actuate the (wrong) scheduling decision, but we will either not try to find a different place for it immediately.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+   `spec.scheduling.schedulingPolicy.gang`.
+
+2. **User-set `gang.minCount` is rejected for Alpha**: For Alpha, `minCount` is always derived
+   from `spec.replicas` by the controller, and a user-supplied `minCount` value is rejected at
+   admission. A gang is satisfiable once `minCount` pods can be scheduled, so allowing
+   `minCount < replicas` would enable useful partial-gang semantics (e.g., a 5-replica database
+   that can start with 3 members). Making `minCount` user-configurable is planned for Beta
+   (see [Beta graduation criteria](#beta)).
+
+3. **Mutable minCount on PodGroup**: Per KEP-5832, `minCount` on the standalone `PodGroup` object
```
wojtek-t, line 493: So what exactly we will do as part of this KEP (if anything)?

The consequences of this point are unclear.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+   is mutable. This means simple scaling operations (changing `spec.replicas`) can be handled by
+   updating the PodGroup's `minCount` in place. However, operations that change the PodGroup
+   structure (e.g., partition changes creating new PodGroups) still require PodGroup recreation.
+
+4. **schedulingGroup immutability on pods**: The `spec.schedulingGroup.podGroupName` field on pods
+   is immutable after creation. This means operations like removing a `RollingUpdate.partition`
```
wojtek-t, line 499: Comment to myself: understand how the design approaches partitions and get back here.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+   `ownerReferences`). When the PodGroup is deleted, the generated ResourceClaim is deallocated
+   and garbage-collected. For StatefulSets, this means that PodGroup recreation (e.g., due to
+   partition changes) causes the old ResourceClaim to be deleted and a new one generated for the
+   new PodGroup.
+
+8. **Maximum 4 ResourceClaims per PodGroup**: KEP-5729 limits the `resourceClaims` list to 4
+   entries per PodGroupTemplate/PodGroup. This limit applies to the StatefulSet's
+   `spec.scheduling.resourceClaims` field as well.
+
+9. **Higher-level controllers composing StatefulSets**: Controllers such as LeaderWorkerSet (LWS)
```
wojtek-t, line 528: I don't agree this is how it should work.
I can easily imagine a situation where (as a higher-level controller) I want to create a StatefulSet and e.g. define TAS requirements on it, but still make it part of my higher-level Workload & (Composite)PodGroup structure.

So even if we say that this is based on existence of `spec.scheduling`, there should be an ability to "attach" it to existing Workload object (and not manage our own) and to existing (Composite)PodGroup structure. This is exactly what the annotations were designed for:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/6089-was-controller-apis#the-solution-downward-mapping-annotations
and statefulset controller should respect them.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+   `spec.scheduling` on the inner StatefulSet specs. Higher-level controllers that manage their
+   own gang scheduling across multiple StatefulSets should not set `spec.scheduling` on the
+   composed StatefulSets to avoid conflicting or redundant PodGroups.
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| Partition changes require PodGroup recreation | Changing `RollingUpdate.partition` requires PodGroup recreation and pod recreation due to immutable `schedulingGroup` on pods | Document this behavior clearly; partition changes are infrequent and already involve rolling pod updates |
+| Workload object leak if StatefulSet controller crashes mid-creation | Orphaned Workload objects consume scheduler resources | OwnerReference with `blockOwnerDeletion=true` ensures garbage collection; finalizer prevents premature cleanup |
```
wojtek-t, line 542: 1. Workload doesn't consume resources on its own
2. PodGroups can, but only via the attached ResourceClaims
3. I still don't understand what you're trying to say here. As described in the job integration KEP, before creating a Workload (and same for PodGroup) controller should appropriate check if one already exists and in such case reuse it - the pattern has already been established here:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#workload-and-podgroup-discovery

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+
+## Design Details
+
+### Workload Ownership Model
+
+Following the ownership hierarchy from KEP-5832, the StatefulSet controller directly owns the
+Workload, PodGroup, and Pods:
+
+```
+StatefulSet (apps/v1)
```
wojtek-t, line 557: Please don't reinvent it - the same model as defined by the Job should be used:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#ownerreferences-relationship

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+- Setting `blockOwnerDeletion=true` ensures the StatefulSet's deletion blocks until the Workload
+  and PodGroup are fully removed, maintaining consistent cleanup ordering.
+- The PodGroup is created with a finalizer (KEP-5832 deletion protection) that is removed only
+  when all referencing pods reach a terminal phase.
+
+### Pod Management Policy Constraints
+
+| Pod Management Policy | Workload Policy | Gang Scheduling | Rationale |
+|----------------------|-----------------|-----------------|-----------|
+| `Parallel` | `Gang` | Yes | All pods created simultaneously—natural fit for gang semantics |
```
wojtek-t, line 578: Even for Parallel, I should be able to specify `Basic` too - because I want to use other constraints (e.g. TAS).

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+| `OrderedReady` | `Basic` | No | Sequential pod creation conflicts with gang's all-at-once model |
+
+Gang scheduling for StatefulSet **requires** `podManagementPolicy: Parallel`. The `OrderedReady`
+policy creates pods sequentially, waiting for each to become Ready before creating the next, which
+fundamentally conflicts with gang scheduling's requirement that all pods exist simultaneously for
+the scheduler to evaluate them as a group.
+
+**DisruptionMode with Basic policy**: `DisruptionMode` is only meaningful with Gang policy, where
+it controls whether preemption evicts a single pod or the entire gang. For OrderedReady
+StatefulSets with Basic policy, `DisruptionMode` is a no-op — pods are always disrupted
```
wojtek-t, line 588: Let's not say it's no-op - instead we should extend the validation and reject if someone sets DisruptionMode to All for basic policy.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+    resourceClaims:            # (requires DRAWorkloadResourceClaims gate)
+    - name: shared-device
+      resourceClaimTemplateName: my-device-template
+---
+# PodGroup (runtime scheduling unit)
+apiVersion: scheduling.k8s.io/v1alpha3
+kind: PodGroup
+metadata:
+  name: my-statefulset-podgroup
+  ownerReferences:
```
wojtek-t, line 653: Please adjust to what was already designed in Job controller:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#ownerreferences-relationship

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
wojtek-t: For me, it's less about complexity, but more about usecases - do we really already have important enough usecase that clearly shows we need two podgroups in such case?

I believe that the most common usecase for partition is canarying of a new rollout. And the question here is - are two PodGroups in such case really a desired behavior?
Imagine a usecase of 16 pods within StatefulSet and now I'm rolling a new version and creating a partition of 4. If I previously claimed that my whole workload forms a gang (so all 16 pods are in a gang), do they now really form two separate gangs? It means that original 12 instances can work without the new 4 replacements, which contradicts the fact that the gang originally had a size of 16.

Now - I can imagine hypothetical usecase where having separate PodGroups would make sense in case of a TAS or in the context of ResourceClaims. But that effectively would mean that the partition is a unit and as a result StatefulSet as a whole (even in the steady state after dropping the partition e.g. after finished rollout) consists of more than one scheduling unit.

So I actually strongly believe that having more than one PodGroup for a single StatefulSet is a much complex usecase and is not really coupled with the partition feature. At the same time, I can see a clear usecases where having single PodGroup even in partitioned case makes sense.

So I actually agree with Matt's conclusion above - let's focuse on clear usecases and:
1) At least for Alpha (but I would argue that for the whole lifetime of this KEP), assume that we have 1:1 mapping between PodGroup and StatefulSet as a whole.
2) Design the case where we would support multiple PodGroups per StatefulSet separately based on very clear user-oriented usecases (so far none of these are defined in the KEP).

@soltysh @janetkuo - for your thoughts too

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+solution is pod recreation.
+
+### Lifecycle Management
+
+#### Initial Creation Lifecycle
+
+When a Parallel StatefulSet is created with gang scheduling enabled:
+
+1. The StatefulSet controller detects `spec.scheduling` is non-nil.
+2. The controller creates the `Workload` object (static template with PodGroup template).
```
wojtek-t, line 798: Please take a look a Job controller integration and adjust to support higher-level controller the way we solved that with annotations:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#controller-workflow

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+3. The controller creates the `PodGroup` object (runtime unit, snapshot of Workload template).
+4. The controller waits for the PodGroup to be observed in the informer cache.
```
wojtek-t, line 800: Please remove - as commented above this is not needed.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+5. The controller creates all pod templates with `spec.schedulingGroup.podGroupName` pointing to
+   the PodGroup.
+6. Pods enter the scheduler queue.
```
wojtek-t, line 803: Please collapse all points below into a single "Scheduler schedules the pods as a single PodGroup".

The details of what is happening there are irrelevant for this KEP.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1406 @@
...
+`minCount` becomes user-configurable, scaling `replicas` will not automatically change a
+user-supplied `minCount`. For example, a user may set `replicas: 5` with `minCount: 3` and later
+scale to `replicas: 7` while keeping `minCount: 3`.
+
+**ResourceClaims during scaling**: ResourceClaims shared at the PodGroup level are unaffected by
+simple scaling operations. Scale-up adds new pods that reference the existing PodGroup and its
+already-allocated ResourceClaims — no new ResourceClaims are generated. Scale-down deletes pods
+but the PodGroup's ResourceClaims remain allocated as long as the PodGroup exists.
+
+**Note**: Operations that change the PodGroup *structure* (e.g., RollingUpdate.partition changes
```
wojtek-t, line 866: based on results of discsussion in this comment thread:
https://github.com/kubernetes/enhancements/pull/6298/changes#r3958492202

it should may need to be adjusted (same in some other places).

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Sure, The intention was to only highlight the gang-scheduling capability in the Summary and not over shadow the other capabilities. Re-written.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Addressed

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Agreed

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Noted. The intention was not to have any changes in the scheduler logic and noted in a bullet point with clarity.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: An empty value (`scheduling: {}`) is valid and resolves to the `Basic` policy with no constraints, no disruption mode, and no shared claims. Added a comment

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Clarified the comment mentioning it is for beta

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Added the `immutable` at one level higher i.e. at the `Scheduling` field

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Makes sense. Removed the reference of the cache

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Removed

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Yes. I assume those are covered under the term "scheduling optimizations"
Anyway, I've re-worded this section providing some more clarity

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Yes. There would be a validation to reject this combination

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Sure, Addressed

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Yes, I agree this is valid for an ensemble with a point of redundancy. But for a service/system like a large model sharded across multiple replicas of a StatefulSet as mentioned in the updated `Story 2`, it makes sense to preempt the whole group instead of a single pod as it doesn't serve the purpose while the other pods keep using the GPU resources sitting idle.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: @wojtek-t , Thanks for the clarification. Your point about simulating TAS via DRA is understood and agreed, but I have kept it out now as it describes the scheduler behavior owned by KEP-5729/5732. So, the story is now purely about claim sharing capability as the other story covers the TAS

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Thanks for the catch, the caveat is gone now. It wasn't a constraint on this design but a property describing the easy path it enables during scaling.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Thanks for pointing this out. Updated the point along with the reference

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack, The pattern should work here too. Removed this point

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Ack

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
sairameshv: Makes sense, Updated

sairameshv:
@mm4tt , @wojtek-t Thanks for the thorough reviews. I've addressed most of your review comments, Could you make some time to take a re-look at the updated changes and leave your feedback?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+the StatefulSet controller fighting the parent for ownership, which the earlier revision of this
+KEP incorrectly ruled out by telling parent controllers not to set `spec.scheduling` at all.
+
+### OwnerReferences Relationship
+
+This KEP uses the relationship defined by
+[KEP-5547](https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5547-integrate-workload-with-job#ownerreferences-relationship)
+with `StatefulSet` substituted for `Job`:
+
+```mermaid
```
kannon92, line 731: This is not rendering correctly.

kannon92 (commented):
PRR is looking really good.

I'm going to wait for final sig approval to make sure I review the finished product.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+Following the decoupled model from KEP-5832 and the controller pattern from KEP-6089, the
+integration uses the `workloadbuilder` library:
+
+```go
+// Scheduling, if set, opts this StatefulSet into Workload-Aware Scheduling:
+// the controller compiles it into one Workload and one PodGroup, and every pod
+// the StatefulSet creates references that PodGroup.
+// An empty value (scheduling: {}) is valid and resolves to the Basic policy
+// with no constraints, no disruption mode, and no shared claims.
+// Gang scheduling requires podManagementPolicy: Parallel.
```
atiratree, line 274: The parallel is more complicated during rollouts/updates. What is the expected behavior when using `maxUnavailable`?

Do the new revision pods belong to the same PodGroup as the old ones? What if the scheduling constraints change for the new template?

`partition` support should also be mentioned here.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+    // +optional
+    // +k8s:optional
+    DisruptionMode *schedulingv1alpha3.WorkloadPodGroupDisruptionMode `json:"disruptionMode,omitempty" protobuf:"bytes,3,opt,name=disruptionMode"`
+
+    // ResourceClaims defines which ResourceClaims may be shared among Pods in
+    // the StatefulSet. Pods consume the devices allocated to a PodGroup's
+    // claim by defining a claim in its own Spec.ResourceClaims that matches
+    // the PodGroup's claim exactly. The claim must have the same name and
+    // refer to the same ResourceClaim or ResourceClaimTemplate.
+    // At most 4 claims may be set, matching the limit on the resulting
```
atiratree, line 336: What happens when we add a support for dedicated ResourceClaims to StatefulSets? For example if we have a conflict between the PodGroup and StatefulSet new definition? How is something similar to PersistentVolumeClaimRetentionPolicy going to look like?

keps/sig-apps/6277-integrate-workload-with-statefulset/kep.yaml
```diff
@@ -0,0 +1,52 @@
+title: Workload API Integration with StatefulSet
+kep-number: 6277
+authors:
+  - "@sairameshv"
+owning-sig: sig-apps
+participating-sigs:
+  - sig-scheduling
```
atiratree, line 7: shouldn't sig-storage be consulted as well? For example for:

> Volume bindings conflict with group placement

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+   individually satisfy `minCount` while the application's actual quorum requirement spans both,
+   and it makes topology constraints ambiguous (must both halves land in the same rack?).
+
+3. **The single-PodGroup model does not need it.** Pod count does not change during a rolling
+   update, so `minCount` does not change either. Pods keep the same `podGroupName` across revisions
+   and the group's scheduling configuration is stable throughout.
+
+What still needs design before partitions can be supported (Beta):
+
+- **Gang semantics during a rolling update.** A gang is satisfied at bind time; a rolling update
```
atiratree, line 1060: as mentioned, this happens even without a partition

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+StatefulSet into WAS in Alpha, and — because `spec.scheduling` is immutable in Alpha — cannot
+temporarily opt out to perform one. This is called out in
+[Risks and Mitigations](#risks-and-mitigations) and in the Beta graduation criteria.
+
+Note that the underlying tension is not created by partitions: even a plain
+`RollingUpdate` with `partition: 0` recreates gang members one at a time. Alpha allows this
+(rejecting rolling updates outright would make the feature unusable) and documents the stall risk;
+Beta is where the semantics get pinned down.
+
+### Lifecycle Management
```
atiratree, line 1083: Lifecycle management should include rolling out of new revisions.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+    SchedulingConstraints *schedulingv1alpha3.WorkloadPodGroupSchedulingConstraints `json:"schedulingConstraints,omitempty" protobuf:"bytes,2,opt,name=schedulingConstraints"`
+
+    // DisruptionMode defines the mode in which the StatefulSet's pods can be
+    // disrupted. Exactly one of Single or All must be set.
+    // Single (the scheduler default when unset) treats each pod as an
+    // independent preemption victim; All makes the whole StatefulSet a single
+    // atomic preemption unit.
+    // All is only valid with the Gang policy: workloadbuilder rejects
+    // `all` combined with `basic`, since a Basic group is scheduled
+    // independently and all-or-nothing disruption is meaningless for it.
```
atiratree, line 325:
```suggestion
    // DisruptionMode defines the mode in which the StatefulSet's pods can be
    // disrupted by an external entity. It does not redefine disruption or unavailability patterns by the StatefulSet controller. Exactly one of Single or All must be set.
    // Single (the scheduler default when unset) treats each pod as an
    // independent disruption (e.g. preemption) victim; All makes the whole StatefulSet a single
    // atomic disruption (e.g. preemption) unit.
    // All is only valid with the Gang policy: workloadbuilder rejects
    // `all` combined with `basic`, since a Basic group is scheduled
    // independently and all-or-nothing disruption is meaningless for it.
```

Or similar wording?
AFAIU this is the current state. And expected future state?

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+   reconciliation. A consequence worth stating: the controller never blindly creates, it discovers
+   first and creates only what is missing, which is what makes crash recovery and re-enablement
+   after a rollback safe.
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| `rollingUpdate.partition` cannot be used with `spec.scheduling` in Alpha | Partitioned canary rollouts are unavailable to any StatefulSet using WAS, and immutability means one that already opted in cannot opt out to run one | Rejected by validation on create and update with a clear message. See [RollingUpdate.Partition](#rollingupdatepartition-deferred-to-beta) |
+| Rolling update of a gang-scheduled StatefulSet stalls | A pod recreated with a revision that no longer fits (bigger requests, new topology domain) cannot be placed, and the gang stays unsatisfied | Visible in PodGroup status and Pending-pod signals; rolling-update semantics under gang are a Beta design item |
```
atiratree, line 632: This seems like an important item to solve in alpha. Updates are essential to sts, so it seems we should know the outcome for this sooner than later.

keps/sig-apps/6277-integrate-workload-with-statefulset/README.md
```diff
@@ -0,0 +1,1876 @@
...
+**Note on Alpha behavior**: The examples above reflect Alpha, where `minCount` is always derived
+from `spec.replicas` — scaling replicas automatically updates `minCount` to match. In Beta, when
+`minCount` becomes user-configurable, scaling `replicas` will not automatically change a
+user-supplied `minCount`. For example, a user may set `replicas: 5` with `minCount: 3` and later
+scale to `replicas: 7` while keeping `minCount: 3`.
+
+**ResourceClaims during scaling**: ResourceClaims shared at the PodGroup level are unaffected by
+simple scaling operations. Scale-up adds new pods that reference the existing PodGroup and its
+already-allocated ResourceClaims — no new ResourceClaims are generated. Scale-down deletes pods
+but the PodGroup's ResourceClaims remain allocated as long as the PodGroup exists.
```
atiratree, line 1153: What happens when the pods resource different ResourceClaims during an update?

PR #6309, opened 2026-08-31
KEP-5683: Kubectl drain reporting node conditions

rthallisey:
One-line PR description: Write several conditions to the Node during `kubectl drain`, that will provide context to the node's lifecycle state

Issue link: https://github.com/kubernetes/enhancements/issues/5683

Other comments:
This KEP is not meant to handle locking/ownership for conditions or concurrent writers.  That will be addressed in a follow up proposal.

Use cases [document](https://docs.google.com/document/d/1EINvuVzEoRra0CKH6uQnOcQJVbd7ZnxT1bcyySN-r7c/edit?usp=sharing)

### DIFF
--- a/keps/sig-node/5683-lifecycle-conditions/README.md
+++ b/keps/sig-node/5683-lifecycle-conditions/README.md
@@ -1,4 +1,4 @@
-# KEP-NNNN: Node Lifecycle Conditions
+# KEP-5683: Node Lifecycle Conditions
 
 <!-- toc -->
 - [Release Signoff Checklist](#release-signoff-checklist)
@@ -21,6 +21,9 @@
   - [GracefulNodeShutdownInProgress Condition](#gracefulnodeshutdowninprogress-condition)
   - [Lifecycle Conditions](#lifecycle-conditions)
   - [Writer Ownership](#writer-ownership)
+  - [Kubectl Drain Condition Writer](#kubectl-drain-condition-writer)
+    - [Permissions](#permissions)
+    - [Signal Handling](#signal-handling)
   - [Feature Gate](#feature-gate)
   - [Future Extension Points](#future-extension-points)
   - [Test Plan](#test-plan)
@@ -30,7 +33,7 @@
       - [e2e tests](#e2e-tests)
   - [Graduation Criteria](#graduation-criteria)
     - [Alpha - Introduce Well-Known Conditions](#alpha---introduce-well-known-conditions)
-    - [Alpha2 - Consume Conditions in Controllers](#alpha2---consume-conditions-in-controllers)
+    - [Alpha2 - Publish Drain Conditions from kubectl](#alpha2---publish-drain-conditions-from-kubectl)
     - [Beta](#beta)
     - [GA](#ga)
     - [Deprecation](#deprecation)
@@ -100,7 +103,10 @@ core behavior to react to those condition values.
 
 The first version is intentionally narrow. It does not define a general node
 maintenance protocol. It establishes the pattern: a well-known condition
-consumable on the Node, providing a foundation for future work.
+consumable on the Node, providing a foundation for future work. The second
+alpha introduces `kubectl drain` the first in-tree writer of `DrainInProgress`
+and `Drained`, establishing how a lifecycle tool publishes these observations
+without changing drain behavior.
 
 ## Motivation
 
@@ -123,7 +129,8 @@ Kubernetes already uses well-known Node conditions for shared Node state such as
 `Ready`, `MemoryPressure`, `DiskPressure`, `PIDPressure`, and
 `NetworkUnavailable`. This KEP extends that model by proposing new Node
 conditions for Node Lifecycle. First, publish well-known Node lifecycle signals,
-then make core controllers use those signals for better status reporting.
+then follow up KEPs can make core controllers use those signals for better
+status reporting.
 
 ### Goals
 
@@ -136,6 +143,7 @@ then make core controllers use those signals for better status reporting.
   ecosystem tools can consume.
 - Provide an asynchronous building block that can be extended without changing
   the initial condition.
+- alpha2: Publish `kubectl drain` progress through `DrainInProgress` and `Drained`.
 
 ### Non-Goals
 
@@ -167,6 +175,12 @@ For this KEP, the new Node conditions are admin managed. An admin, or an
 admin-authorized maintenance controller, sets the condition status to `True`
 when appropriate. Clearing the lifecycle state is done by setting the condition
 status to `False` or removing the condition.
+
+In Alpha2, `kubectl drain` becomes the first in-tree lifecycle condition
+writer. With condition reporting enabled, it sets `DrainInProgress=True` when
+it begins processing a Node and sets `Drained=True` when the drain criteria
+selected by that invocation have been met. Reporting is informational and does
+not change the existing drain operation.
 
 ### User Stories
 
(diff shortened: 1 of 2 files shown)


keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -429,6 +450,74 @@ CamelCase.
...
+
+Starting a new drain resets `Drained=False` before processing the Node.
+Concurrent drain actors are not coordinated. Any authorized actor can patch
+the lifecycle conditions, and the latest successful update determines the
+observed value.
+
+Kubectl uses a strategic merge patch containing only `DrainInProgress` and
+`Drained`. `NodeCondition` is a merge list keyed by condition type, so unrelated
+Node conditions and status fields are preserved. A reporting failure is
+printed as a warning and does not change the drain command's result.
```
atiratree, line 493: Could we consider leveraging server-side-apply for the kubectl. This could allow us to do a simple ownership model based on SSA to ensure that two kubectl drain invocations do not step on each other's toes.

Proposal:
- Before the drain we could check if we can take over the ownership of the drain process / condition(s)
- If the condition(s) is in KubectlDrainCompleted, KubectlDrainFailed or KubectlDrainInterrupted state, we can forcefully take the ownership and mark it as KubectlDrainStarted.
- If it is KubectlDrainStarted, we would let the management of the drain to the first original instance and wait until we observe KubectlDrainCompleted. If our drain is still progressing (e.g. different drain end conditions) we can flip the condition back again to KubectlDrainStarted and take ownership

The only edge case is when the original process is forcefully killed and we are stuck in KubectlDrainStarted. 

One way of solving this is to instruct kubectl to give periodic updates to the condition and if they are not received within a certain time frame (e.g. 10 minutes), we can override the condition ownership. Fortunately, node conditions are special and have a lastHeartbeatTime that we can use: https://github.com/kubernetes/kubernetes/blob/d888ea8ae424fd5dac4535a9ef013b12cb0cb375/staging/src/k8s.io/api/core/v1/types.go#L7346-L7348

@rthallisey WDYT?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -100,7 +103,10 @@ core behavior to react to those condition values.
 
 The first version is intentionally narrow. It does not define a general node
 maintenance protocol. It establishes the pattern: a well-known condition
-consumable on the Node, providing a foundation for future work.
+consumable on the Node, providing a foundation for future work. The second
+alpha introduces `kubectl drain` the first in-tree writer of `DrainInProgress`
```
janetkuo, line 107:
```suggestion
alpha introduces `kubectl drain` as the first in-tree writer of `DrainInProgress`
```

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -429,6 +450,74 @@ CamelCase.
...
+| `KubectlDrainFailed` | Kubectl could not meet its selected drain criteria because the operation failed or timed out. |
+| `KubectlDrainInterrupted` | Kubectl handled an interrupt before its selected drain criteria were met. |
+
+The two conditions are updated together:
+
+| Event | `DrainInProgress` | `Drained` | Reason |
+|---|---|---|---|
+| Kubectl starts processing the Node | `True` | `False` | `KubectlDrainStarted` |
+| Selected criteria are met | `False` | `True` | `KubectlDrainCompleted` |
+| Drain fails or times out | `False` | `False` | `KubectlDrainFailed` |
```
janetkuo, line 478: Once `kubectl drain` sets `Drained=True`, what clears `Drained=True` when the operator finishes maintenance and runs `kubectl uncordon <node>`?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -429,6 +450,74 @@ CamelCase.
...
+| Kubectl handles an interrupt | `False` | `False` | `KubectlDrainInterrupted` |
+
+For a multi-Node invocation, kubectl publishes `DrainInProgress=True`
+immediately before processing each individual Node. It does not publish that
+condition for every selected Node when the initial cordon operation begins.
+
+Starting a new drain resets `Drained=False` before processing the Node.
```
janetkuo: In `k8s.io/kubectl/pkg/cmd/drain`, `deleteOrEvictPodsSimple` runs `GetPodsForDeletion()` to validate pod filters (`--ignore-daemonsets`, `--delete-emptydir-data`, `--force`) before calling `DeleteOrEvictPods()`. Can we specify that `KubectlDrainStarted` (`DrainInProgress=True, Drained=False`) is only patched *after* `GetPodsForDeletion()` succeeds?

Otherwise, running `kubectl drain` with a missing filter flag (e.g. forgetting `--ignore-daemonsets`) on an already-drained Node (`Drained=True`) will fail pre-flight validation immediately and overwrite `Drained=True` with `Drained=False (KubectlDrainFailed)` without evicting any Pods.

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -429,6 +450,74 @@ CamelCase.
...
+Concurrent drain actors are not coordinated. Any authorized actor can patch
+the lifecycle conditions, and the latest successful update determines the
+observed value.
+
+Kubectl uses a strategic merge patch containing only `DrainInProgress` and
+`Drained`. `NodeCondition` is a merge list keyed by condition type, so unrelated
```
janetkuo, line 491: Would you also specify how `lastHeartbeatTime`, `lastTransitionTime`, and `message` are populated in this patch, as well as `--dry-run=client` / `--dry-run=server` behavior?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -545,39 +671,46 @@ enabled, behavior remains unchanged.
...
-No scheduling or rollout behavior changes. Enabling the feature adds an
-observability signal.
+No scheduling or rollout behavior changes. During Alpha2, kubectl reporting is
+opt-in. Enabling it adds best-effort Node status writes.
 
 ###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
 
-Yes.
+Yes. Disable the feature gate and stop passing
+`--report-node-conditions`. Existing conditions remain until an authorized
```
janetkuo: Disabling the `NodeLifecycleConditions` feature gate on `kube-apiserver` does not prevent `kubectl drain --report-node-conditions` from patching `DrainInProgress`/`Drained` onto `Node.status.conditions` (since `Node.status.conditions` accepts arbitrary `NodeConditionType` strings on any supported API server, as noted in the Version Skew section above).

Also, `kep.yaml` adds `kubectl` under the `NodeLifecycleConditions` feature gate (`components: [kube-apiserver, kube-controller-manager, kubectl]`), whereas `kubectl` uses `KUBECTL_*` environment variables (`cmdutil.FeatureGate`) rather than `--feature-gates`, and the Feature Gate section above (line 525) omits `kubectl`. Could we align `kep.yaml` and the README on whether `kubectl` uses a `KUBECTL_*` env var gate in alpha or relies solely on the `--report-node-conditions` flag?

keps/sig-node/5683-lifecycle-conditions/README.md
```diff
@@ -665,12 +802,24 @@ etcd availability returns.
...
+    Pods remaining on the Node.
+  - Mitigation: an authorized administrator corrects or clears the condition.
+  - Testing: handled interrupts are tested; uncatchable process or host failure
+    cannot be cleaned up automatically.
+- The kubectl user lacks `nodes/status` permission:
+  - Detection: kubectl prints a forbidden warning and audit logs record the
+    denied request.
+  - Mitigation: grant the permission or run without condition reporting.
+- Concurrent writers publish conflicting observations:
+  - Detection: inspect condition timestamps, reasons, messages, and audit logs.
```
omerap12, line 815: Does it make sense that the writer will include some sort of identifier in the condition message to make concurrent-writer debugging possible?

omerap12 (commented):
Comment from PRR shadow. Overall LGTM.
(The metrics section is empty - I think that's fine though).

keps/sig-node/5683-lifecycle-conditions/README.md
janetkuo, line 485: Would it accidentally overwrite the drain condition on an already drained Node in some cases?

Issue #6122, opened 2026-05-28
Support CPU Configurable Scaling Delay


Chunxia202410:
### Enhancement Description

- One-line enhancement description (can be used as a release note):   Support Configurable CPU Scale Down Delay
- Kubernetes Enhancement Proposal:  https://github.com/kubernetes/enhancements/pull/6123
- Discussion Link:   [SIG Node meeting](https://www.youtube.com/watch?v=1nsOSmgYh4w&list=PL69nYSiGNLP1wJPj5DYWXjiArF-MJ5fNG&index=3)
- Primary contact (assignee): @Chunxia202410 , @lukaszwojciechowski , @radoslawc , @jain-ashish-sam, @subh007 , @shiya0705
- Responsible SIGs: sig-node
- Enhancement target (which target equals to which milestone):
  - [ ] Alpha release target (x.y): 1.38
  - [ ] Beta release target (x.y): 1.39
  - [ ] Stable release target (x.y): 1.40
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [X] KEP (`k/enhancements`) update PR(s): 
      - https://github.com/kubernetes/enhancements/pull/6123
      - https://github.com/kubernetes/enhancements/pull/6314
    - [ ] Code (`k/k`) update PR(s): 
      - https://github.com/kubernetes/kubernetes/pull/136015
      - https://github.com/kubernetes/kubernetes/pull/140627
    - [ ] Docs (`k/website`) update PR(s): https://github.com/kubernetes/website/pull/56235
    - [ ] Blog (`k/website`) update PR(s): https://github.com/kubernetes/website/pull/56234

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

whtssub:
Hello @radoslawc  👋, v1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9 June 2026 (AoE) / Wednesday 10 June 2026 12:00 UTC**.

This enhancement is targeting stage `alpha` for v1.37 (correct me, if otherwise)

Here's where this enhancement currently stands:

- [x] PR open or merged with the KEP's [PRR questionnaire](https://github.com/kubernetes/enhancements/tree/master/keps/NNNN-kep-template#production-readiness-review-questionnaire) filled out.
[...]

PR #6314, opened 2026-09-02
KEP-6122: Update to target Alpha in v1.38

Chunxia202410:
- One-line PR description: Update to target Alpha in v1.38

- Issue link: https://github.com/kubernetes/enhancements/issues/6122

- Other comments:

### DIFF
--- a/keps/sig-node/6122-configurable-scaling-delay-with-pod-resource-exposure/kep.yaml
+++ b/keps/sig-node/6122-configurable-scaling-delay-with-pod-resource-exposure/kep.yaml
@@ -41,11 +41,11 @@ stage: alpha
 # The most recent milestone for which work toward delivery of this KEP has been
 # done. This can be the current (upcoming) milestone, if it is being actively
 # worked on.
-latest-milestone: "v1.37"
+latest-milestone: "v1.38"
 
 # The milestone at which this feature was, or is targeted to be, at each stage.
 milestone:
-  alpha: "v1.37"
+  alpha: "v1.38"
   beta: "v1.39"
   stable: "v1.40"
 


ffromani:
> /ok-to-test /cc @dchen1107 /cc @ffromani /cc @pravk03 /cc @natasha41575 /cc @kad /cc @liggitt /cc @SergeyKanzhelev

please make sure to add the PRR metadata: https://github.com/kubernetes/enhancements/pull/6353/changes#diff-1b1d1fe5655237e41a49e5fc7e80c76a9e61ea4ff44cba666bb40dc10ada4685

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+-->
```
ffromani, line 127: nit: please remove these guiding comments (here and in the reset of the document) to make the review easier

lukaszwojciechowski:
> > /ok-to-test /cc @dchen1107 /cc @ffromani /cc @pravk03 /cc @natasha41575 /cc @kad /cc @liggitt /cc @SergeyKanzhelev
> 
> please make sure to add the PRR metadata: https://github.com/kubernetes/enhancements/pull/6353/changes#diff-1b1d1fe5655237e41a49e5fc7e80c76a9e61ea4ff44cba666bb40dc10ada4685

@ffromani do we need that?
The prod-readiness/6122.yaml exists since 1.37 cycle and is up to date.
Again @kannon92 was assigned for PRR

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+> **Note:** This KEP introduces one new feature gate, `InPlacePodVerticalScalingExclusiveCPUsScaleDownDelay`, which requires `InPlacePodVerticalScalingExclusiveCPUs` to be enabled as well. The dependency is enforced at kubelet startup by the [feature gate dependency functionality](https://github.com/kubernetes/kubernetes/pull/133697).
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+* Allow pods to optionally specify a scale-down grace period via the `scaleDownGracePeriodSeconds` field in the Pod spec.
```
ffromani, line 161: the all-or-nothing characterization of the past iteration, with the setting being per-node and not per-workload, was indeed one of the main concerns of the past proposal.
I just want to clarify that from my side _any_ workload-specific (per-pod) way to opt-in in the behavior is fine; adding a pod field is totally an option, but not a hard requirement from my side.
For example, just thinking aloud: a special-purpose extended resource COULD also be an alternative worth exploring.
(NOTE: I didn't read "discarded alternatives" yet nor I remember if we already discussed and discarded this approach previously)
EDIT: I’m **NOT** asking to revisit the pod-field direction in this PR; it addresses my concern. I only want to keep the requirement separate from the specific mechanism, and I’m happy to help evaluate alternatives in a later iteration if useful.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+* Add new CPU manager policies.
+* Allow containers to specify which CPUs to remove during scale-down.
+* **Per-container grace periods:** `scaleDownGracePeriodSeconds` is a pod-level field, so a single value applies to every container in the pod, even though cpusets are assigned per container. Allowing a different grace period per container is out of scope.
+* **Scale-down delay for PodLevelResourceManagers:** The scale-down delay introduced by this KEP does not apply to resources managed by `PodLevelResourceManagers` (KEP-5526). `PodLevelResourceManagers` does not support In-Place scaling for pod-level resources. Additionally, KEP-5554 (which enables In-Place scaling of exclusive CPU at the container level) excludes scaling for pods that define pod-level resources.
```
ffromani, line 175: I think this means that KEP 5554 intentionally deferred integration with pod-level-resources tracked in KEP 5526 (and I agree with the decision for scope and gradual rollout reasons), and that this KEP builds on that decisions and focuses on container-level resources only?
Anyway, a rephrasing to clarify a bit more would help

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+nitty-gritty.
+-->
+
+This proposal introduces a new pod-level field, `scaleDownGracePeriodSeconds`, which specifies the minimum delay before applying an updated cpuset when a container has its CPU allocation scaled down. This field is opt-in; for pods that do not specify it (or set it to 0), the new cpuset is applied at the next cpuset actuation, preserving the existing behavior.
+
+During the delay window, the container continues to use the current cpuset for at least the configured grace period. This allows latency-sensitive workloads to monitor and prepare for the upcoming CPUSet change before CPU(s) are removed from the container. The minimum delay must be guaranteed at all times, including during kubelet restarts.
+
+This proposal does not require a handshake (acknowledgment) from the workload. The workload declares its needed preparation time via `scaleDownGracePeriodSeconds`, and the kubelet guarantees to honor this time window without requiring explicit feedback from the application.
+
+### Use Cases
```
ffromani, line 194: the use case helps. You may want (and I think the KEP template requires) to add user stories told from the user perspective: "as workload author, I want..."

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
+# KEP-6122: Configurable Scaling Delay
```
ffromani, line 1: please make sure to use the latest KEP template, as sections are continuosly updated and fine tuned per previous cycles' feedback

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+
+**Workload may not complete preparations within the delay window:** The kubelet guarantees only that the cpuset will not be applied before `scaleDownGracePeriodSeconds` has elapsed. There is no synchronization mechanism between the workload and kubelet — if the workload fails to complete its preparations (e.g., workload migration, draining tasks) within the delay window, the cpuset change is applied anyway. This is an inherent limitation of keeping kubelet independent from workload state.
+
+**Mitigation:** Application developers should configure `scaleDownGracePeriodSeconds` based on the worst-case preparation time required by their latency-sensitive workloads. Workloads should be designed to handle premature CPU removal gracefully (e.g., by quickly migrating tasks to remaining CPUs or tolerating brief performance degradation). Early notification of the upcoming change, which gives a workload the maximum possible preparation time, is provided by [KEP-6369](https://github.com/kubernetes/enhancements/issues/6369).
```
ffromani, line 223: I'm trying to think if there's a way to detect and report (through a metric maybe?) if the workload missed the grace period or not
EDIT: I realize this may read as "reintroduce the handshake" - not that I meant, and I agree about avoiding the workload handshake which would complicate things very significantly. Perhaps we can exposes if the kubelet actuates before the deadline - this is a clear indicator something is wrong?
perhaps we can record the actual actuation times, maybe using histograms?

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+1. In the Allocate CPU stage:
+   Allocate a new CPUSet based on the container's assignments and defaultCPUset in the checkpoint (all references to "checkpoint" below refer to cpu_manager_state).
+
+   When a container scales down (if the CPU number of assignments in the checkpoint > CPU request and limit for the container in the pod spec):
+   - If `scaleDownGracePeriodSeconds` is unset or 0, the container's scale down behavior is the same as before, after reallocating the new cpuset:
+     - The removed_CPUs (assignments in checkpoint - new cpuset) are added to the defaultCPUset
+     - The container's assignments are updated to the new cpuset
+     - CPU manager checkpoint is updated immediately with new assignments and defaultCPUSet.
+   - If `scaleDownGracePeriodSeconds` is greater than 0, after reallocating the new cpuset:
+     - A scale_delay_timer is started with the pod's `scaleDownGracePeriodSeconds` as its duration.
```
ffromani: the other main concern I remember from the previous iteration is how we track time, especially across possible kubelet restarts. I'm worried we can introduce issues related to the possible clock skew, lack of clock monothonic source and so forth.
I think the design should outline how a robust implementation can be made.
I'd be fine to be pointed out to a later section if these topics are already being discussed.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,1243 @@
...
+
+##### Kubelet Restart
+
+A pending scale-down is persisted in the CPU Manager checkpoint, so that a kubelet restart neither restarts the delay nor recomputes the target cpuset. For each container with a pending scale-down the checkpoint holds:
+
+- the **preAssignments**, so the cpuset already exposed to the workload is the one eventually applied,
+- the **time at which the new cpuset may be applied**, expressed as monotonic time since boot, not counting time spent suspended,
+- the **node's boot ID**, read locally from cAdvisor (the same value the kubelet publishes in `node.status.nodeInfo.bootID`).
+
+Wall-clock time is deliberately not used: serializing a `time.Time` discards its monotonic reading and leaves only a wall clock, which is exposed to NTP steps, manual changes and timezone handling. Monotonic time since boot advances steadily by kernel guarantee, but it is comparable only within a single boot — which is what the boot ID establishes. How the value is represented in the checkpoint is an implementation detail.
```
ffromani, line 390: ok, at glance this seems sufficiently robust but I'd need at least another careful pass to this portion

ffromani (commented):
Initial review: this is a remarkable improvement over the past iteration and addresses pretty much all the concerns I raised. I will spend more time on the provided diagrams and on the state persistence and clock concerns.
My comments are about possible alternatives, no major problems from my side yet.

ffromani:
> @ffromani do we need that? The prod-readiness/6122.yaml exists since 1.37 cycle and is up to date. Again @kannon92 was assigned for PRR

Then you should be good. I didn't see in the current diff and pointed out. Better safe than sorry, I thought.

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: Done,
comments removed

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: We analyzed few possible directions allowing per-pod opt-in including hook approach discussed during review in 1.37 cycle.
The chosen architecture candidate was the best among feasible ones (satisfying opt-in requirement and allowing to handle kubelet restart issues). The other strong candidate was the mentioned hook approach. The reasons for rejecting it are described in Alternative 7

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: Both KEPs (#5526 and #5554) were developed in parallel and they both made assumptions that they do not cover a common scenario of scaling exclusive CPUs managed by `PodLevelResourceManagers`.
As this KEP is based on #5554 it also does not cover scale down delay for scaling exclusive CPUs at pod-level.
However the opt-in feature defined in this KEP is pod-level not container-level, so there is no potential conflict for expanding the scale down delay feature for pod-level bubble and containers when scaling of exclusive CPUs at pod-level will be available.

This non-goal was rephrased. I hope it's clearer now.

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: Document updated to latest KEP template.
User story rephrased from workload developer perspective
Two additional user stories added:
* for scaling up - no need for delay
* for workloads not requiring delay - no opt-in

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: Done,
Latest template used.

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: That's exactly what is written in section `Are there any missing metrics that would be useful to have to improve observability of this feature?` - such histograms are proposed, but to be added at beta stage:

> The counters and gauge described in [Metrics](#metrics) are added in Alpha and give an aggregate view of how many scale-down delays start and complete. A histogram of the delay between a scale-down being accepted and the new cpuset being applied would additionally let an operator check the timing SLO without inspecting individual pods; it is not implemented yet and remains a candidate for Beta.

lukaszwojciechowski:
> Initial review: this is a remarkable improvement over the past iteration and addresses pretty much all the concerns I raised. I will spend more time on the provided diagrams and on the state persistence and clock concerns. My comments are about possible alternatives, no major problems from my side yet.

Thanks @ffromani for the initial review.
All comments were addressed and changes are pushed in separate commit for easier review.

Chunxia202410:
/cc @liggitt 
/cc @pohly 
The scale-down delay part remains in this KEP. Since it adds a new field, `scaleDownGracePeriodSeconds`, to the Pod spec, it requires API review. If you have a chance, could you please take a look at this KEP? Thank you very much!

liggitt:
> /cc @liggitt /cc @pohly The scale-down delay part remains in this KEP. Since it adds a new field, `scaleDownGracePeriodSeconds`, to the Pod spec, it requires API review. If you have a chance, could you please take a look at this KEP? Thank you very much!

It would make sense for a field controlling when the scale change is done to live in the same type / place the triggers / bounds of the scale are specified. What triggers scale-up / scale-down / changes in cpu allocation today? Is it something in the Pod spec, or something in some other API object, or something opaque/internal to the kubelet?

lukaszwojciechowski:
> > /cc @liggitt /cc @pohly The scale-down delay part remains in this KEP. Since it adds a new field, `scaleDownGracePeriodSeconds`, to the Pod spec, it requires API review. If you have a chance, could you please take a look at this KEP? Thank you very much!
> 
> It would make sense for a field controlling when the scale change is done to live in the same type / place the triggers / bounds of the scale are specified. What triggers scale-up / scale-down / changes in cpu allocation today? Is it something in the Pod spec, or something in some other API object, or something opaque/internal to the kubelet?

Scaling is done by patching pod spec and adjusting cpu limits and requests for containers. See documentation: https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/
This is a regular `InPlacePodVerticalScaling` operation as described in #1287 
On top of that #5554 enables scaling of exclusive CPUs

So the fields that are changed are defined in resources sections of each container.
However the scaling operation is done on pod object, that's why we define pod-level field. #5554 and so #6122 currently does not work together with `PodLevelResourceManagers`, but integration of the feature and scaling exclusive CPUs on pod-level is also considered as a future work. So a pod-level option is consistent with that plans and would allow in future to implement pod-level scaling as well

keps/sig-node/6122-configurable-scaling-delay/README.md
ffromani: fair enough. From my perspective no flags to raise nor complaints and surely this is a clean approach.

keps/sig-node/6122-configurable-scaling-delay/README.md
ffromani: Fair enough. The best mitigation is to emphasize this side clearly in the docs.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,834 @@
...
+
+## Design Details
+
+### Implementation
+
+The overview of the design:
+![alt text](scale_delay_overview.png)
+
+The basic flow shown in the diagram is as follows:
+- During allocation of new CPUSets, changes are not applied to assignments immediately but kept in preAssignments, and `scale_delay_timers` are started.
```
ffromani, line 175: the only remaining concern from the previous iteration is here about the timers.
I'm fine adding conceptually (at design stage) one timer per pending update, but at implementation stage we need to be careful to make the code efficient and not necessarily require one actual timer per pending update (as in for example `time.Ticker`). IOW, let's make sure the design doesn't lock in a specific, possibly inefficient implementation. The current wording seems good, just want to make sure this point is accounted in the future.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,834 @@
...
+- The preAssignments and the time at which the cpuset may be applied are recorded in the CPU manager checkpoint.
+- The new CPUSets kept in preAssignments are the values [KEP-6369](https://github.com/kubernetes/enhancements/issues/6369) exposes to the container.
+- During cpuset actuation, containers scaling down (with active `scale_delay_timer`) are skipped if the pod's `scaleDownGracePeriodSeconds` has not yet elapsed. For those where the time has passed, preAssignments are written to assignments and actuated in containers.
+- The SyncPod performs two actions after there is no active `scale_delay_timer` on the pod, and actuated states equal the allocated ones:
```
ffromani, line 179: question just to be sure I got the design 100% correct: the idea is that SyncPod will query the cpumanager through container manager and mark the resize completed and update actual resources once cpumanager reports it completed actuation, right?
IOW, we do not want cpumanager to signal/call back SyncPod, correct?
If the above is correct I'd be completely fine with the design

ffromani (commented):
/lgtm

from my side. A couple of clarifications/future concerns but I'm happy with all the changes and from kubelet/resource managers side I'm good moving forward

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: The _timer_ as described in a document is a concept.
We do not plan to implement it with `time.Ticker` or any similar heavy solutions.
Each resize in progress with delayed scale down will just keep a single timestamp (uptime_now() + delay) - the same timestamp that will be stored in CPUManager checkpoint for resuming delay after kubelet restart.
Reconcile loop of CPUManager that iterates over all running pods and containers and verify if resources are actuated will additionally verify that uptime_now() is past the timestamp. If so, the resize will be actuated, if not it will be verified again in next reconclie loop iteration.

So no additional timers or goroutines will be required.

As the KEP document does not bind to any specific implementation, we can clear any possible concerns during review of code PRs

ffromani: perfect, that cleanly address my remaining concern. From my POV the KEP LGTM pending possible wording/editorial changes. Will review again

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: > question just to be sure I got the design 100% correct: the idea is that SyncPod will query the cpumanager through container manager and mark the resize completed and update actual resources once cpumanager reports it completed actuation, right?

yes, that's exactly the design

> IOW, we do not want cpumanager to signal/call back SyncPod, correct?

well, CPUManager won't signal SyncPod directly, but when last container of a pod in resize is completed, CPUManager will emit a PLEG event and that will trigger SyncPod to verify the status of pod and clear pod resize InProgress status.

The mechanism is described in KEP's section `Resize Complete State`: https://github.com/kubernetes/enhancements/pull/6314/changes#diff-743112a5c0297c8404a9b03fdfa4820ec812d02e9108272900bd64f826031385R345

The diagrams `flow_as_is` `flow_to_be` show that additional event:
https://github.com/kubernetes/enhancements/pull/6314/changes#diff-743112a5c0297c8404a9b03fdfa4820ec812d02e9108272900bd64f826031385R194-R200

natasha41575:
> > > /cc @liggitt /cc @pohly The scale-down delay part remains in this KEP. Since it adds a new field, `scaleDownGracePeriodSeconds`, to the Pod spec, it requires API review. If you have a chance, could you please take a look at this KEP? Thank you very much!
> > 
> > 
> > It would make sense for a field controlling when the scale change is done to live in the same type / place the triggers / bounds of the scale are specified. What triggers scale-up / scale-down / changes in cpu allocation today? Is it something in the Pod spec, or something in some other API object, or something opaque/internal to the kubelet?
> 
> Scaling is done by patching pod spec and adjusting cpu limits and requests for containers. See documentation: https://kubernetes.io/docs/tasks/configure-pod-container/resize-container-resources/ This is a regular `InPlacePodVerticalScaling` operation as described in #1287 On top of that #5554 enables scaling of exclusive CPUs
> 
> So the fields that are changed are defined in resources sections of each container. However the scaling operation is done on pod object, that's why we define pod-level field. #5554 and so #6122 currently does not work together with `PodLevelResourceManagers`, but integration of the feature and scaling exclusive CPUs on pod-level is also considered as a future work. So a pod-level option is consistent with that plans and would allow in future to implement pod-level scaling as well

I have reservations about adding scaleDownGracePeriodSeconds directly to v1.PodSpec.

The field sounds general, but it only applies under very narrow conditions: static CPU manager policy, Guaranteed QoS, integer cores, and CPU scale-down only. Paying the permanent API cost of a top-level PodSpec field for this highly specific use case feels disproportionate.

I recognize that a node-level flag was rejected in the alternatives, but have we considered keeping this as an opt-in pod annotation until there is a concrete, generalized design that also covers memory and other potentially other resources?

lukaszwojciechowski:
Thank you @natasha41575 for joining review and sharing your opinion
I'll try to address your concerns.

> I have reservations about adding scaleDownGracePeriodSeconds directly to v1.PodSpec.
> 
> The field sounds general, but it only applies under very narrow conditions: static CPU manager policy, Guaranteed QoS, integer cores, and CPU scale-down only. Paying the permanent API cost of a top-level PodSpec field for this highly specific use case feels disproportionate.
> 

It is an optional pointer, so it costs nothing on pods that do not set it, and as an Alpha field behind a gate it is removable if the feature does not graduate. That does not make it free, but the commitment is not permanent yet.

> I recognize that a node-level flag was rejected in the alternatives, but have we considered keeping this as an opt-in pod annotation until there is a concrete, generalized design that also covers memory and other potentially other resources?

Yes we have, but it was rejected (probably I should add that into alternatives section)

### Generally guidance from sig-architecture say we should not use annotations:

[API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#label-selector-and-annotation-conventions):

> [!TIP]
> In early versions of Kubernetes, some in-development features represented new API fields as annotations, generally with the form `something.alpha.kubernetes.io/name` or `something.beta.kubernetes.io/name` (depending on our confidence in it). This pattern is deprecated. Some such annotations may still exist, but no new annotations may be defined. New API fields are now developed as regular fields.


[API change guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#new-field-in-existing-api-version), on annotations as the alpha mechanism:

> [!TIP]
> Previously, annotations were used for experimental alpha features, but are no longer recommended for several reasons:
> * They expose the cluster to "time-bomb" data added as unstructured annotations against an earlier API server (https://issue.k8s.io/30819)
> * They cannot be migrated to first-class fields in the same API version
>
> The preferred approach adds an alpha field to the existing object, and ensures it is disabled by default


Same approach is found in recent KEPs, e.g.
KEP-5836 rejected a node annotation for its preemption policy on the grounds that "core Kubernetes components (such as the Scheduler and Kubelet) should generally not depend on unstructured annotations to drive critical runtime decisions. Annotations are meant for metadata and not for policy enforcement."

### But except for general reasons, here is why this KEP rejected annotations

Four properties of this design depend on the value being a validated, immutable, gated field. An annotation gives up all four.

**1. Validation.** The value is range-limited to `[0, 10]` seconds by kube-apiserver. The cap is not cosmetic: CPUs being released are not returned to the shared pool until the delay expires, so it bounds how long one pod can hold CPUs it no longer requests while another pod waits for them. As an annotation, that bound moves from an API guarantee to per-kubelet string parsing, and a malformed value silently means "no delay" instead of being rejected at admission.

**2. Immutability.** A pending scale-down — the target cpuset and the time at which it may be applied — is persisted in the CPU Manager checkpoint, so that a kubelet restart neither re-arms the delay nor recomputes the target. That is only sound because the grace period cannot change for the lifetime of the pod. Annotations are freely patchable, so a deadline persisted at admission could end up governed by a value the pod no longer carries. This is precisely the argument the KEP already uses to reject the node-level option in Alternative 5, and it applies to annotations verbatim.

**3. Feature gating and version skew.** The skew story starts with kube-apiserver rejecting the field while the gate is disabled, which is what makes "an accepted pod was admitted by a cluster that understands `scaleDownGracePeriodSeconds`" true. Every apiserver accepts every annotation, including ones that predate this feature entirely. A pod could then carry a grace period in a cluster where nothing honors it, with no error surfaced anywhere.

**4. Node Declared Features.** The scheduler keeps pods that ask for a grace period off nodes that cannot honor it, and the kubelet refuses to admit such a pod on a node that does not declare the feature. That inference runs over `PodInfo{Spec, Status}` — the framework does not pass `ObjectMeta` to a feature, so an annotation is invisible to it. Moving to an annotation would mean either giving up the placement guarantee and the admission check, or changing the Node Declared Features framework itself.

There is also the migration cost. An annotation that later becomes a field means supporting both for several releases, and Kubernetes has moved away from annotation-first APIs for exactly that reason.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+
+* **Why Rejected**: The delay would be applied below the layer at which the kubelet accounts for CPUs, so the kubelet would not know that it exists. The consequences, in the order of their severity:
+  1. **The CPUs are released to the shared pool immediately.** The CPU Manager returns the removed CPUs to `defaultCPUSet` at allocation time, when the resize is admitted, whereas the hook is only reached when the reconcile loop actuates the cpuset later. In between, the CPUs can be handed to a newly created container, whose cpuset is set at creation and may well happen before the next reconcile, or to the shared-pool containers reconciled in the same pass, while the scaled-down container is still running on them. The plugin cannot prevent this, because it has no channel to the CPU Manager state.
```
tallclair, line 826: This is a problem we need to resolve regardless. See https://github.com/kubernetes/kubernetes/issues/142363

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  2. **The resize completes prematurely from the kubelet's point of view.** The kubelet clears `PodResizeInProgress` once the actuated resources match the allocated ones, and the actuated resources are recorded as soon as the runtime accepts the cpu.weight update, which is a separate `UpdateContainerResources` call that carries no cpuset. The actual resources reported in the pod status come from the runtime as well. The pod status and the scheduler would therefore see the resize as complete, and the CPUs as free, before the cpuset has changed. This is exactly the situation that [Actual Resources Update](#actual-resources-update) exists to prevent.
```
tallclair, line 827: My proposal is to have the plugin return a transient error. The Kubelet just needs to handle retry correctly, and report that the resize is in-progress until it completes successfully.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  3. **The plugin request timeout is global.** It is a single runtime setting that applies to every hook, of every plugin, for every pod on the node. Raising it to cover the grace period extends the time any misbehaving plugin can hold up any container operation. Exceeding it is not reported as an error either: the runtime closes the plugin and proceeds without it, so a delay that runs too long silently applies the cpuset at once and disconnects the plugin.
```
tallclair, line 828: There's a proposal for per-plugin timeouts: https://github.com/containerd/nri/issues/287

But with the async approach, I don't think you need it.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  4. **There is no correct reaction to a plugin restart.** If the plugin exits while holding a request, the runtime treats the lost connection the same way and applies the cpuset immediately. When the plugin reconnects it is synchronized with the list of running containers, but nothing tells it which resizes were pending or how much of their grace period had elapsed. Every plugin restart, and the window until it reconnects, drops the guarantee.
```
tallclair, line 829: As long as it reports the error back to Kubelet, Kubelet should keep retrying the request. You could have the plugin checkpoint in-progress resizes, or just reset its clock. Or let the application signal the plugin when it's ready.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  5. **The CPU Manager reconcile loop is blocked.** The kubelet calls `UpdateContainerResources` synchronously from its single reconcile goroutine, so while the hook holds one reply no other container is actuated, including scale-ups of other pods and the cpuset of containers that have just restarted.
```
tallclair, line 830: I don't think it's an issue with the async approach, but I'd like to fix it anyway. I think with modern cgroups v2 implementations (linux 6.7+ IIRC) we should be able to just use exclusive CPUs and not manage the default set. The cpuset updating should be moved to SyncPod either way.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  6. **All NRI events on the node are blocked.** The NRI adaptation in the runtime serializes every hook under one mutex. A plugin holding `UpdateContainer` also holds up `RunPodSandbox`, `CreateContainer`, `StartContainer` and `StopContainer` for every pod and every plugin on the node for the whole grace period, so no pod can start or terminate in that window.
```
tallclair, line 831: Proposed fix in https://github.com/containerd/nri/issues/165, but also resolved by the async approach.

keps/sig-node/6122-configurable-scaling-delay/README.md
```diff
@@ -0,0 +1,845 @@
...
+  7. **It does not avoid an API change.** Applying the delay only to the pods that ask for it, and keeping those pods off nodes that cannot provide it through [Node Declared Features](#node-declared-features-integration), requires the pod to declare its grace period in a validated, immutable field. That is the field this KEP adds; the plugin would only move the enforcement to a place where it cannot be enforced.
```
tallclair, line 832: With an out-of-tree NRI plugin, configuration should be an annotation on the pod. You can enforce immutability with a VAP if needed.

keps/sig-node/6122-configurable-scaling-delay/README.md
lukaszwojciechowski: It's a known problem. I replied in the issue

ffromani: +1 to resolve this issue anyway as soon as possible (best case scenario in the first KEP 5554 iteration)

keps/sig-node/6122-configurable-scaling-delay/README.md
ffromani: I'm open to improve the reconcile loop (until we can get rid of it) anyway regardless if we take the NRI route or the in-tree route. Handling transient errors in the reconcile loop especially on scale down is something we can do anyway. As much as I'd like to explore modern cgroups v2 facilities, I'm under the impression the scope of that work would require a separate KEP.

keps/sig-node/6122-configurable-scaling-delay/README.md
Chunxia202410: Thanks @tallclair, the transient-error approach is much better than the "hold the request" design we had in mind. 

However, three concerns I still have with the NRI route:

**1. Two owners for one decision.** In this design, the kubelet drives the retry loop and reports the resize status, but the timer lives in the plugin. The only message between them is a plain error with no meaning attached. This split creates problems that don't exist in-tree:

- When the kubelet gets an error, it cannot tell "grace period still running, try later" apart from "the plugin has a bug". Telling them apart needs a well-defined "please retry" error in the NRI API.
- If the plugin crashes and never comes back, the kubelet would retry forever. So the kubelet needs its own escape policy — which means the delay logic now lives in two places.
- If the resize is cancelled mid-grace, or the container restarts mid-grace, the kubelet already handles these in its own state machine. A plugin only sees a stream of update requests, so it has to rebuild the same logic itself, and clean up its stale records.

**2. Silent failure when the feature is missing.** With an annotation, every cluster accepts the pod — including clusters with no plugin installed. The pod then runs with no delay at all, and nothing reports an error anywhere. For a feature whose job is to protect performance, this "silently does nothing" mode is the scariest one. A real field fails loudly instead: the apiserver rejects it when the feature gate is off, and with Node Declared Features the scheduler keeps the pod away from nodes that cannot provide the delay.

**3. Bigger version matrix.** In-tree, the feature depends on two components: kubelet and the runtime. With a plugin it depends on three: the plugin version, the runtime's NRI support (which differs across containerd/CRI-O releases), and the kubelet. And since the runtime just proceeds when a plugin fails, a mismatch on any of these axes does not produce an error — it silently turns the feature off.

Looking forward to hearing your thoughts.

PR #6317, opened 2026-09-02
KEP-6313: Hermetic Pods

aojea:
- One-line PR description: Add new KEP to add spec.networkMode (Cluster|Host|None) so a Pod can opt out of the cluster network and run with only a loopback interface.

- Issue link: https://github.com/kubernetes/enhancements/issues/6313

- Other comments:

### DIFF
--- a/keps/sig-network/6313-hermetic-network-isolated-pods/README.md
+++ b/keps/sig-network/6313-hermetic-network-isolated-pods/README.md
@@ -1,0 +1,1047 @@
+# KEP-6313: Hermetic Pods: network-isolated pod sandboxes
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: DRA-provisioned accelerator fabric](#story-1-dra-provisioned-accelerator-fabric)
+    - [Story 2: Disconnected batch job](#story-2-disconnected-batch-job)
+    - [Story 3: VM sandboxes and device passthrough](#story-3-vm-sandboxes-and-device-passthrough)
+    - [Story 4: Local-only multi-container pods](#story-4-local-only-multi-container-pods)
+    - [Story 5: Agentic orchestrators](#story-5-agentic-orchestrators)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Defaulting and <code>hostNetwork</code> mirroring](#defaulting-and-hostnetwork-mirroring)
+  - [Validation](#validation)
+  - [CRI Integration](#cri-integration)
+  - [Kubelet Behavior](#kubelet-behavior)
+  - [Control Plane Behavior](#control-plane-behavior)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+    - [Deprecation](#deprecation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP adds **Hermetic (Network Isolated) Pods**, Pods that opt out of the cluster network.
+A hermetic pod gets a network namespace with only a loopback interface and
+is not reachable from the cluster.
+
+## Motivation
+
+Kubernetes was built on the early assumption that every Pod is attached to a
+single flat cluster network and is routable from every other Pod. This
+assumption has always generated friction with workloads that need advanced
+routing, multi-tenancy or meshes. Those workloads get around the "default
+network" with out-of-band mechanisms: per-node runtime configuration, custom
+runtime classes, chained plugins, or annotations.
+
+SIG Network has already tried to solve this inside the project. The
+[Multi-Network effort (KEP-3698)] concluded it was too complex: problems like
+network probes or DNS across multiple networks had no clear solution, portability
+depended on very specific (often private) plugin implementations, and the
+user stories were fragmented — some wanted multiple NICs, others VPC-like
+networks — each backed by a relatively small number of users. A follow-up
+proposal to simplify the low level API ([KEP-4410]) required offloading the
+network to another component that will make it incompatible with existing
+deployments and hard to roll out.
+
+What has worked is composing on top of stable core primitives instead of
+extending the core network model:
+
+- [Kubernetes Network Drivers][knd-paper] defined a new architecture, based
+  on DRA, to provide a native mechanism to plumb additional network devices.
+  [DRANET] is a project that demonstrated the successful use of this
+  architecture to solve the challenges of the new AI/ML workloads and RDMA
+  networks.
+- [KEP-4962] (GA) added a standard way to report network data in DRA. All
+  drivers report their network devices with a common language, and
+  integrators can use that data to build network integrations out of core
+  (for example, their own EndpointSlice controllers), keeping core stable.
+- [NRI] showed how network applications can integrate directly with the
+  runtime. This removes the limitations of the CNI model, which composes
+  plugins only through serial Unix-style chaining and file system based configuration
+  that breaks the separation of concerns (see the [service mesh CNI discussion]).
+
+This friction is more critical these days. New workloads, like the agentic
+ones, require strict isolation, and their density and scale make the traditional
+out-of-band mechanisms obsolete. At the same time, DRA network drivers, VM sandboxes
+that provide their own networking (e.g. Kata/Firecracker/... with VFIO passthrough),
+disconnected jobs, ... have been waiting for a solution for years.
+
+The missing piece is a core primitive to NOT attach the default network.
+Decomposing the Pod network attachment solves this problem for both
+maintainers and integrators: integrators get an isolated sandbox they can
+build on with the mechanisms above, and the project stops carrying the toil
+of workarounds and requirements that the core network model was never meant
+to express.
+
+[Multi-Network effort (KEP-3698)]: https://github.com/kubernetes/enhancements/issues/3698
+[KEP-4410]: https://github.com/kubernetes/enhancements/issues/4410
+[knd-paper]: https://ieeexplore.ieee.org/document/11146291/
+[DRANET]: https://github.com/google/dranet
+[KEP-4962]: https://github.com/kubernetes/enhancements/issues/4962
+[NRI]: https://github.com/containerd/nri
+[service mesh CNI discussion]: https://github.com/kubernetes/kubernetes/issues/130594
+
+### Goals
+
+- Allow users to declare that a Pod must not be attached to the cluster
+  network.
+- Provide an API that is backwards compatible with the current network model.
+- Fail predictably: reject network-dependent features (cluster DNS, service
+  links, network probes, host ports) at admission time instead of failing at
+  runtime, and never run an isolated pod attached to the network on nodes
+  that do not support the feature.
+- Define the contract between the kubelet and the container runtimes.
+- Keep isolated pods out of service discovery: Services, Endpoints,
+  EndpointSlices and service environment variables.
+
+### Non-Goals
+
+- Designing a multi-network or secondary-interface attachment framework.
+  `networkMode` is intentionally about the absence of automatic attachment;
+  it provides a clean surface for delegation (e.g., DRA drivers), but how
+  interfaces are attached afterwards and its lifecycle is out of scope.
+- Isolating anything other than networking. Storage, IPC, PID, devices and
+  host access are governed by existing fields.
+- Deprecating or removing `hostNetwork`. It is a GA v1 field and remains
+  supported forever; `networkMode: Host` is its enum alias.
+- Changing how kube-proxy, NetworkPolicy or DNS work for non-hermetic pods.
+- Skipping service account token mounting for hermetic pods. Tokens may still
+  be consumed by non-network means.
+
+## Proposal
+
+Add a new optional enum field to `PodSpec`:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: hermetic-example
+spec:
+  networkMode: None
+  containers:
+  - name: worker
+    image: registry.k8s.io/e2e-test-images/agnhost:2.53
+    command: ["sleep", "3600"]
+```
+
+Semantics of `networkMode: None`:
+
+1. The API server defaults `dnsPolicy` to `None` (without requiring
+   `dnsConfig.nameservers`) and `enableServiceLinks` to `false`, and validates
+   that no network-dependent feature is requested (see [Validation](#validation)).
+2. The kubelet uses a new CRI field to tell the runtime to create the sandbox
+   network namespace with only the loopback interface and to skip all the
+   network plumbing (CNI is never invoked on Linux runtimes).
+3. The pod runs with empty `status.podIP`/`status.podIPs` and becomes `Ready`
+   without waiting for an IP. `status.hostIP` is still reported.
+4. The endpoints and endpointslice controllers never select the pod into
+   Endpoints/EndpointSlices, even if a Service selector matches it.
+5. The kubelet does not inject the `KUBERNETES_SERVICE_HOST`/`KUBERNETES_PORT`
+   environment variables. It also maps the pod hostname to `127.0.0.1` and
+   `::1` in the managed `/etc/hosts`, so `localhost` and `$(hostname)`
+   resolve inside the sandbox.
+6. If the container runtime does not report support for isolated sandboxes,
+   the kubelet rejects the pod at admission (**fail closed**) instead of
+   silently attaching it to the cluster network.
+
+`networkMode: Host` and `networkMode: Cluster` do not change any behavior:
+`Host` is the explicit form of `hostNetwork: true` and `Cluster` is the
+current default.
+
+### User Stories
+
+#### Story 1: DRA-provisioned accelerator fabric
+
+As an operator of an AI training cluster, I allocate RDMA/accelerator NICs to
+pods through DRA drivers. I want the pod sandbox to start empty: no
+interfaces, routes or iptables rules from the default network plugin. The DRA
+driver is then the only owner of the pod connectivity. Today I have to
+coordinate teardown and priority with the CNI configuration on every node.
+
+#### Story 2: Disconnected batch job
+
+As a security engineer, I run jobs that process sensitive data mounted from
+volumes, and I must prove to auditors that the workload has no network path
+in or out. A deny-all NetworkPolicy depends on the network plugin enforcing
+it, and the pod still gets a routable IP. With `networkMode: None` there is
+nothing to enforce: the only interface is loopback.
+
+#### Story 3: VM sandboxes and device passthrough
+
+As a confidential-computing platform owner using VM sandboxes, networking
+is provided inside the VM (e.g., via VFIO passthrough of a physical function).
+The node-level network plumbing is useless and potentially harmful. I want
+the runtime to skip it entirely, driven by a core API field rather
+than a runtime-specific annotation.
+
+#### Story 4: Local-only multi-container pods
+
+As a CI system author, I run pods whose containers only talk over `localhost`
+(hermetic builds and tests). I want to guarantee that no external network
+dependency can leak into the build, and I do not want these pods to consume
+cluster IP space or appear in service discovery.
+
+#### Story 5: Agentic orchestrators
+
+As an operator of an agent orchestrator (e.g. Substrate), I run untrusted,
+often machine-generated code in sandboxed pods at very high density and
+churn. Each agent must start with zero network access by default; the
+orchestrator then grants connectivity explicitly through its own channels
+(a local proxy over `localhost`, or interfaces injected by a driver). I
+cannot depend on NetworkPolicy enforcement for this guarantee, and at this
+scale the per-pod network plumbing and IPAM are too much overhead: the pods never
+use the cluster network, but they still consume IP space, endpoints
+processing, and setup/teardown time on every sandbox.
+
+### Notes/Constraints/Caveats
+
+**Naming.** This was extensively discussed during the initial reviews
+([kubernetes/kubernetes#141604]):
+
+- "Hermetic" is only the colloquial name of the feature. It describes the
+  contract well: a sealed environment with no external network attachments,
+  like hermetic builds in Bazel or Nix. It is deliberately **not** an API
+  value: reviewers pointed out that "hermetic" can also imply storage, host
+  access or compute isolation, and this KEP does not provide those.
+- "CNI" never appears in the API. CNI is an implementation detail of Linux
+  container runtimes; it does not exist in the Kubernetes API vocabulary.
+- `None` follows existing ecosystem semantics (`docker run --network
+  none`) and the existing `dnsPolicy: None`.
+
+**The pod still has a network namespace.** The pod keeps loopback
+(`127.0.0.1`, `::1`) and the containers in the pod communicate over
+`localhost`. This is why names like `noNetwork` were rejected: they are
+technically inaccurate.
+
+**Kubernetes API access.** A hermetic pod cannot reach the kube-apiserver (or
+anything else). The documentation in [services-networking concepts] assumes
+that every pod can reach the API and other pods. Those documents must be
+updated to describe `networkMode: None` as an explicit, opt-in exception.
+Pods that do not set the field keep all the documented guarantees.
+
+**Downward API.** `status.podIP`/`status.podIPs` resolve to empty values for
+hermetic pods. `status.hostIP` remains populated.
+
+[kubernetes/kubernetes#141604]: https://github.com/kubernetes/kubernetes/pull/141604
+[services-networking concepts]: https://kubernetes.io/docs/concepts/services-networking/
+
+### Risks and Mitigations
+
+**Fail-open on version skew (security risk).** The scheduler keeps hermetic
+pods away from unsupported nodes using [Node Declared Features (KEP-5328)],
+but pods can bypass the scheduler (`spec.nodeName`, DaemonSets, static pods).
+An old runtime or kubelet that does not understand the field could silently
+attach an isolated pod to the cluster network. The kubelet **fails closed**
+instead: if `networkMode` is `None` and the runtime does not report
+support for isolated sandboxes (via CRI `RuntimeFeatures`), the kubelet
+rejects the pod with `PodAdmitFailed`. See
+[Version Skew Strategy](#version-skew-strategy).
+
+[Node Declared Features (KEP-5328)]: https://github.com/kubernetes/enhancements/issues/5328
+
+**Ecosystem assumptions that every pod has an IP.** Controllers, meshes and
+operators that read `status.podIP` will see an empty value. This is the same
+state as a pod whose sandbox is still being created, so well-behaved
+consumers already tolerate it. The endpoints controllers get an explicit
+check anyway.
+
+**API confusion with `hostNetwork`.** Two fields describing the same thing is
+not great, but `hostNetwork` is v1 and cannot go away. Defaulting keeps both in
+sync and validation rejects conflicts; a reader of either field gets the
+same answer.
+
+**Scope creep towards multi-network.** An enum invites new values
+(`bridge`-like modes, named networks). Not in this KEP. Any future value
+needs its own KEP and cannot weaken the conformance guarantees of the
+default mode.
+
+## Design Details
+
+### API Changes
+
+New enum type and `PodSpec` field:
+
+```go
+// PodNetworkMode describes how the pod sandbox network is provisioned.
+// +enum
+type PodNetworkMode string
+
+const (
+	// PodNetworkModeCluster attaches the pod to the cluster-wide pod
+	// network. The container runtime performs its configured network
+	// plumbing and the pod is assigned routable pod IPs. This is the
+	// default and matches the historical behavior of Kubernetes.
+	PodNetworkModeCluster PodNetworkMode = "Cluster"
+
+	// PodNetworkModeHost shares the host's network namespace with the pod.
+	// It is equivalent to setting hostNetwork: true; the two fields are
+	// kept in sync by API defaulting.
+	PodNetworkModeHost PodNetworkMode = "Host"
+
+	// PodNetworkModeNone provides the pod with a dedicated network
+	// namespace containing only the loopback interface. The container
+	// runtime performs no network plumbing and the pod is assigned no IP.
+	// The pod is never selected into Services and receives no cluster DNS
+	// configuration or service environment variables.
+	PodNetworkModeNone PodNetworkMode = "None"
+)
+```
+
+```go
+type PodSpec struct {
+	// ...
+
+	// networkMode selects how the pod sandbox network is provisioned.
+	// "Cluster" (default) attaches the pod to the cluster network,
+	// "Host" shares the host network namespace (equivalent to
+	// hostNetwork: true and kept in sync with it), and "None" gives the
+	// pod an isolated network namespace with only a loopback interface
+	// and no automatic network plumbing.
+	// When "None" is selected, dnsPolicy must be "None",
+	// enableServiceLinks must be false, and probes, lifecycle handlers
+	// and hostPorts that require networking are forbidden.
+	// This field is immutable.
+	// +featureGate=PodNetworkMode
+	// +optional
+	NetworkMode *PodNetworkMode `json:"networkMode,omitempty" protobuf:"bytes,45,opt,name=networkMode,casttype=PodNetworkMode"`
+}
+```
+
+**Why an enum and not a boolean.** The first proposal was `hermetic *bool`
+(see [Alternatives](#alternatives)). It got a lot of pushback and it makes
+it hard to think about in conjunction with `hostNetwork`, since they will be
+two mutually exclusive booleans that will need cross-field validation
+in both directions (`if hermetic { hostNetwork must be false }`, and the
+reverse). The [API conventions] discourage exactly this. With a single enum
+a pod has exactly one network mode, the invalid combinations cannot be
+written, and `hostNetwork` is absorbed into it as the `Host` value.
+
+[API conventions]: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types
+
+### Defaulting and `hostNetwork` mirroring
+
+`hostNetwork` is a GA v1 field and cannot be removed. Mirroring an enum onto
+a legacy boolean is the best we can do without breaking v1 compatibility. Defaulting
+in `pkg/apis/core/v1/defaults.go` keeps both fields in sync, for Pods and
+for every workload `PodTemplateSpec`:
+
+```go
+func SetDefaults_PodSpec(obj *v1.PodSpec) {
+	// NetworkMode <-> HostNetwork mirroring.
+	if utilfeature.DefaultFeatureGate.Enabled(features.PodNetworkMode) {
+		if obj.NetworkMode == nil {
+			// Old clients: derive the mode from the legacy field.
+			if obj.HostNetwork {
+				obj.NetworkMode = ptr.To(v1.PodNetworkModeHost)
+			} else {
+				obj.NetworkMode = ptr.To(v1.PodNetworkModeCluster)
+			}
+		} else if *obj.NetworkMode == v1.PodNetworkModeHost {
+			// New clients: keep the legacy field in sync for older
+			// controllers and clients that read hostNetwork directly.
+			obj.HostNetwork = true
+		}
+	}
+
+	if obj.DNSPolicy == "" {
+		if obj.NetworkMode != nil && *obj.NetworkMode == v1.PodNetworkModeNone {
+			obj.DNSPolicy = v1.DNSNone
+		} else {
+			obj.DNSPolicy = v1.DNSClusterFirst
+		}
+	}
+	if obj.NetworkMode != nil && *obj.NetworkMode == v1.PodNetworkModeNone &&
+		obj.EnableServiceLinks == nil {
+		obj.EnableServiceLinks = ptr.To(false)
+	}
+	// ... existing defaulting ...
+}
+```
+
+Properties of this strategy:
+
+- An old client submitting `hostNetwork: true` observes `networkMode: Host`
+  after defaulting; behavior is unchanged.
+- A new client submitting `networkMode: Host` produces `hostNetwork: true` in
+  the stored object, so an n-1 controller reading only the raw boolean makes
+  the correct decision.
+- A client cannot send the conflict "`networkMode: Host` with
+  `hostNetwork: false`". `hostNetwork` is a plain boolean, so the API server
+  cannot tell an explicit `false` apart from an unset field. In both cases
+  defaulting sets `hostNetwork` to `true` to match `networkMode: Host`.
+- The remaining conflicts (`networkMode: Cluster` or `None` combined with an
+  explicit `hostNetwork: true`) are rejected by validation.
+- When the `PodNetworkMode` feature gate is disabled, the field is stripped
+  from new objects (standard `dropDisabledFields` handling), no mirroring
+  occurs, and behavior is exactly as today. Objects that already carry the
+  field keep it, following the API compatibility rules for disabled gates.
+
+### Validation
+
+Implemented in `pkg/apis/core/validation/validation.go` and invoked from
+`ValidatePodSpec` (covering Pods and all workload templates):
+
+```go
+// validatePodNetworkMode validates spec.networkMode and its interactions
+// with network-dependent fields.
+func validatePodNetworkMode(spec *core.PodSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if spec.NetworkMode == nil {
+		return allErrs
+	}
+	modePath := fldPath.Child("networkMode")
+
+	switch *spec.NetworkMode {
+	case core.PodNetworkModeCluster:
+		if spec.HostNetwork {
+			allErrs = append(allErrs, field.Invalid(modePath, *spec.NetworkMode,
+				`must be "Host" when hostNetwork is true`))
+		}
+	case core.PodNetworkModeHost:
+		if !spec.HostNetwork {
+			// Defaulting keeps these in sync; reaching this branch means
+			// an internal client bypassed defaulting.
+			allErrs = append(allErrs, field.Invalid(modePath, *spec.NetworkMode,
+				"hostNetwork must be true when networkMode is \"Host\""))
+		}
+	case core.PodNetworkModeNone:
+		allErrs = append(allErrs, validateIsolatedPodSpec(spec, fldPath)...)
+	default:
+		allErrs = append(allErrs, field.NotSupported(modePath, *spec.NetworkMode,
+			[]core.PodNetworkMode{core.PodNetworkModeCluster, core.PodNetworkModeHost, core.PodNetworkModeNone}))
+	}
+	return allErrs
+}
+
+// validateIsolatedPodSpec enforces that a networkMode "None" pod requests no
+// feature that depends on pod or cluster networking.
+func validateIsolatedPodSpec(spec *core.PodSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if spec.HostNetwork {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networkMode"),
+			core.PodNetworkModeNone, "must not be \"None\" when hostNetwork is true"))
+	}
+	if spec.DNSPolicy != core.DNSNone {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("dnsPolicy"),
+			spec.DNSPolicy, `must be "None" when networkMode is "None"`))
+	}
+	if spec.EnableServiceLinks == nil || *spec.EnableServiceLinks {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("enableServiceLinks"),
+			spec.EnableServiceLinks, `must be false when networkMode is "None"`))
+	}
+
+	podshelper.VisitContainersWithPath(spec, fldPath, func(c *core.Container, cFldPath *field.Path) bool {
+		// Network probes (httpGet, tcpSocket, grpc) cannot succeed without a
+		// network interface; only exec probes are permitted.
+		checkProbe := func(probe *core.Probe, probePath *field.Path) {
+			if probe == nil {
+				return
+			}
+			if probe.HTTPGet != nil {
+				allErrs = append(allErrs, field.Forbidden(probePath.Child("httpGet"),
+					`may not be set when networkMode is "None"`))
+			}
+			if probe.TCPSocket != nil {
+				allErrs = append(allErrs, field.Forbidden(probePath.Child("tcpSocket"),
+					`may not be set when networkMode is "None"`))
+			}
+			if probe.GRPC != nil {
+				allErrs = append(allErrs, field.Forbidden(probePath.Child("grpc"),
+					`may not be set when networkMode is "None"`))
+			}
+		}
+		checkProbe(c.LivenessProbe, cFldPath.Child("livenessProbe"))
+		checkProbe(c.ReadinessProbe, cFldPath.Child("readinessProbe"))
+		checkProbe(c.StartupProbe, cFldPath.Child("startupProbe"))
+
+		if c.Lifecycle != nil {
+			checkHandler := func(h *core.LifecycleHandler, hPath *field.Path) {
+				if h == nil {
+					return
+				}
+				if h.HTTPGet != nil {
+					allErrs = append(allErrs, field.Forbidden(hPath.Child("httpGet"),
+						`may not be set when networkMode is "None"`))
+				}
+				if h.TCPSocket != nil {
+					allErrs = append(allErrs, field.Forbidden(hPath.Child("tcpSocket"),
+						`may not be set when networkMode is "None"`))
+				}
+			}
+			checkHandler(c.Lifecycle.PostStart, cFldPath.Child("lifecycle", "postStart"))
+			checkHandler(c.Lifecycle.PreStop, cFldPath.Child("lifecycle", "preStop"))
+		}
+
+		// hostPort mappings require an attached network interface.
+		portsPath := cFldPath.Child("ports")
+		for i, port := range c.Ports {
+			if port.HostPort > 0 {
+				allErrs = append(allErrs, field.Forbidden(portsPath.Index(i).Child("hostPort"),
+					`may not be set when networkMode is "None"`))
+			}
+		}
+		return true
+	})
+
+	return allErrs
+}
+```
+
+Additional validation adjustments:
+
+- Only `exec` probes and `exec` lifecycle handlers pass validation.
+  `httpGet`, `tcpSocket` and `grpc` are rejected at admission.
+- Today, `dnsPolicy: None` requires the user to provide
+  `dnsConfig.nameservers`. Pods with `networkMode: None` are an exception: they can
+  have no nameservers at all. The kubelet then writes an empty
+  `/etc/resolv.conf`, unless the user provides a `dnsConfig` (for example,
+  for a resolver listening on loopback).
+- `networkMode` cannot be changed after the pod is created, like the other
+  fields that define the sandbox.
+
+### CRI Integration
+
+A new typed field in `PodSandboxConfig` (`k8s.io/cri-api`) carries the
+sandbox network mode to the runtime:
+
+```proto
+enum PodSandboxNetworkMode {
+    // POD attaches the sandbox to the pod network. The runtime performs its
+    // configured network plumbing (e.g., invokes CNI plugins on Linux).
+    // This is the default and matches historical behavior.
+    POD = 0;
+    // NONE creates the sandbox network namespace with only the loopback
+    // interface configured and skips all network plumbing. The runtime MUST
+    // NOT attach any external interface and MUST report an empty IP in
+    // PodSandboxStatus.
+    NONE = 1;
+}
+
+message PodSandboxConfig {
+    // ... existing fields ...
+
+    // network_mode instructs the runtime how to provision networking for
+    // the sandbox network namespace. Host-network sandboxes continue to be
+    // requested via NamespaceOption (NamespaceMode NODE) and use POD here.
+    PodSandboxNetworkMode network_mode = 10;
+}
+```
+
+Capability discovery, so the kubelet can fail closed:
+
+```proto
+message RuntimeFeatures {
+    // ... existing fields ...
+
+    // network_mode_none is set to true if the runtime supports creating
+    // sandboxes with PodSandboxNetworkMode NONE.
+    bool network_mode_none = 4;
+}
+```
+
+Runtime obligations for `NONE` sandboxes:
+
+- create the netns according to `NamespaceOption` as usual;
+- bring up `lo` (`127.0.0.1/8`, `::1/128`), the equivalent of
+  `ip link set lo up`;
+- no plumbing on setup and none on teardown;
+- report no IPs in `PodSandboxStatus.network`.
+
+A reference containerd implementation exists ([aojea/containerd `cniless`
+branch]). Upstream containerd and CRI-O implementations are a beta graduation
+requirement.
+
+[aojea/containerd `cniless` branch]: https://github.com/aojea/containerd/tree/cniless
+
+### Kubelet Behavior
+
+- Admission is fail-closed. If the pod has `networkMode: None` and the
+  runtime does not report `network_mode_none` in `RuntimeFeatures`, the
+  kubelet rejects the pod with `PodAdmitFailed`, reason
+  `UnsupportedNetworkMode`. Same rejection if the kubelet's `PodNetworkMode`
+  gate is off. Attaching an intentionally isolated pod to the cluster
+  network is worse than not running it.
+- `generatePodSandboxConfig` sets `network_mode: NONE`.
+  `determinePodSandboxIPs` returns no IPs, and `PodSandboxChanged` must not
+  recreate the sandbox because the IP is missing.
+- `status.podIP`/`status.podIPs` stay empty. Readiness is computed from
+  container state only. `status.hostIP` is reported as usual.
+- The managed `/etc/hosts` is still mounted, with the pod hostname (and
+  `hostname.subdomain` if set) mapped to `127.0.0.1` and `::1` to avoid
+  breaking applications that depend on that.
+- No `KUBERNETES_SERVICE_HOST`/`KUBERNETES_PORT`, no service links
+  (`enableServiceLinks` is already forced to `false`). The kubelet also
+  skips the service-informer sync wait for these pods: one less cluster
+  dependency in the startup path, and it shows in startup latency.
+- `dnsPolicy: None` with no `dnsConfig` produces an empty `resolv.conf`. A
+  user-supplied `dnsConfig` is honored: a resolver on loopback inside the
+  pod, or nameservers reachable over driver-injected interfaces.
+
+### Control Plane Behavior
+
+- `ShouldPodBeInEndpoints` (in `k8s.io/endpointslice/util`, shared by the
+  Endpoints and EndpointSlice controllers) returns `false` for
+  `networkMode: None` pods, so they are never published even if a Service
+  selector matches.
+- The kubelet publishes support for `networkMode: None` through
+  [Node Declared Features (KEP-5328)] (GA since v1.37), so the scheduler
+  keeps hermetic pods away from nodes without a supporting kubelet and
+  runtime. This KEP only registers a new feature in that existing mechanism.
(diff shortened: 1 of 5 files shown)


keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
+# KEP-6313: Hermetic Pods: network-isolated pod sandboxes
```
danwinship: To me, "hermetic" does not imply "is not connected to the network", and _does_ imply lots of other things you don't mean (like "_can't_ be attached to _any_ network, ever" and "can't be modified after it's created").

"Network isolated" is also not quite right, because the pod is not actually "isolated", it's just "initially not connected".

Probably more discussion below when we get to the actual API...

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP adds **Hermetic (Network Isolated) Pods**, Pods that opt out of the cluster network.
```
danwinship, line 75:
```suggestion
This KEP adds **Hermetic (Network Isolated) Pods**, Pods that opt out of the pod network.
```

I feel like "cluster network" is sometimes slightly ambiguous (eg, including nodes as well).

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+that provide their own networking (e.g. Kata/Firecracker/... with VFIO passthrough),
+disconnected jobs, ... have been waiting for a solution for years.
+
+The missing piece is a core primitive to NOT attach the default network.
+Decomposing the Pod network attachment solves this problem for both
+maintainers and integrators: integrators get an isolated sandbox they can
+build on with the mechanisms above, and the project stops carrying the toil
+of workarounds and requirements that the core network model was never meant
+to express.
+
```
danwinship, line 127: The ability to not attach the default network is _one_ missing piece, but I wouldn't call it "_the_" missing piece. There are still other issues.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+[knd-paper]: https://ieeexplore.ieee.org/document/11146291/
+[DRANET]: https://github.com/google/dranet
+[KEP-4962]: https://github.com/kubernetes/enhancements/issues/4962
+[NRI]: https://github.com/containerd/nri
+[service mesh CNI discussion]: https://github.com/kubernetes/kubernetes/issues/130594
+
+### Goals
+
+- Allow users to declare that a Pod must not be attached to the cluster
+  network.
```
danwinship, line 139:
```suggestion
- Allow users to declare that a Pod must not be attached to the cluster-default
  pod network.
```

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- Provide an API that is backwards compatible with the current network model.
```
danwinship, line 140: I'm not sure what this means specifically. Of course we require APIs to be backward compatible...

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- Fail predictably: reject network-dependent features (cluster DNS, service
+  links, network probes, host ports) at admission time instead of failing at
+  runtime, and never run an isolated pod attached to the network on nodes
+  that do not support the feature.
```
danwinship, line 144: "an isolated pod attached to the network"???

and what are "nodes that do not support the feature"? Are you just talking about version skew?

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- Define the contract between the kubelet and the container runtimes.
```
danwinship, line 145:
```suggestion
- Define the contract between the kubelet and the container runtimes for isolated pods.
```

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- Keep isolated pods out of service discovery: Services, Endpoints,
+  EndpointSlices and service environment variables.
```
danwinship, line 147: Maybe rather than enumerating specific features here the Goal should just be to fully define the semantics for isolated/hermetic pods.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+```
+
+Semantics of `networkMode: None`:
+
+1. The API server defaults `dnsPolicy` to `None` (without requiring
+   `dnsConfig.nameservers`) and `enableServiceLinks` to `false`, and validates
+   that no network-dependent feature is requested (see [Validation](#validation)).
+2. The kubelet uses a new CRI field to tell the runtime to create the sandbox
+   network namespace with only the loopback interface and to skip all the
+   network plumbing (CNI is never invoked on Linux runtimes).
```
danwinship, line 187: FWIW, some/most/all? Windows runtimes use CNI too. Maybe just


```suggestion
   network plumbing (e.g., runtimes that normally invoke CNI plugins would skip that step).
```

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+3. The pod runs with empty `status.podIP`/`status.podIPs` and becomes `Ready`
+   without waiting for an IP. `status.hostIP` is still reported.
+4. The endpoints and endpointslice controllers never select the pod into
+   Endpoints/EndpointSlices, even if a Service selector matches it.
+5. The kubelet does not inject the `KUBERNETES_SERVICE_HOST`/`KUBERNETES_PORT`
+   environment variables. It also maps the pod hostname to `127.0.0.1` and
```
danwinship, line 193: This feels overly opinionated. We explicitly assume that some hermetic pods will set up non-standard networking. The apiserver IP _may_ be reachable via those networks, and if so, then having `KUBERNETES_SERVICE_HOST` set would be useful. If the user is setting up a hermetic pod that _doesn't_ have access to the apiserver, then presumably they won't even look at `KUBERNETES_SERVICE_HOST`, so it won't matter whether it's set or not.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+   `::1` in the managed `/etc/hosts`, so `localhost` and `$(hostname)`
+   resolve inside the sandbox.
+6. If the container runtime does not report support for isolated sandboxes,
+   the kubelet rejects the pod at admission (**fail closed**) instead of
```
danwinship, line 197: Is this worded right? "kubelet" can't reject it at admission time, it can only put the pod into a Failed state after it gets scheduled to the node, right?

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+   silently attaching it to the cluster network.
+
+`networkMode: Host` and `networkMode: Cluster` do not change any behavior:
+`Host` is the explicit form of `hostNetwork: true` and `Cluster` is the
+current default.
```
danwinship, line 202: OK, bikeshedding...

`networkMode: None` kind of implies "_can't_ have networking" as opposed to merely "doesn't get the default network".

I know I complained about boolean fields before, but what about `podNetwork: [ true | false ]` (where `hostNetwork: true` implies `podNetwork: false`).

Or maybe `defaultPodNetwork: [ true | false ]`.

Or `podNetwork: [ Default | None ]` with maybe the possibility of other options in the future.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+
+### User Stories
+
+#### Story 1: DRA-provisioned accelerator fabric
+
+As an operator of an AI training cluster, I allocate RDMA/accelerator NICs to
+pods through DRA drivers. I want the pod sandbox to start empty: no
+interfaces, routes or iptables rules from the default network plugin. The DRA
```
danwinship, line 210: Why do you want that?

This is just saying "I want the feature as defined". It doesn't explain why the feature as defined is a good idea.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+driver is then the only owner of the pod connectivity. Today I have to
+coordinate teardown and priority with the CNI configuration on every node.
+
+#### Story 2: Disconnected batch job
+
+As a security engineer, I run jobs that process sensitive data mounted from
+volumes, and I must prove to auditors that the workload has no network path
+in or out. A deny-all NetworkPolicy depends on the network plugin enforcing
+it, and the pod still gets a routable IP. With `networkMode: None` there is
+nothing to enforce: the only interface is loopback.
```
danwinship, line 220: Hm... so for this use case you really do want "the pod can't _ever_ get external network connectivity", not just "we don't give the pod external network connectivity by default". I wonder if there's a case for expressing that in the API. (Though I'm not sure how the runtime could manage to enforce "no networking ever" while still allowing `lo` to work...)

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+The node-level network plumbing is useless and potentially harmful. I want
+the runtime to skip it entirely, driven by a core API field rather
+than a runtime-specific annotation.
+
+#### Story 4: Local-only multi-container pods
+
+As a CI system author, I run pods whose containers only talk over `localhost`
+(hermetic builds and tests). I want to guarantee that no external network
+dependency can leak into the build, and I do not want these pods to consume
+cluster IP space or appear in service discovery.
```
danwinship, line 235: (So this is different from story 2, because you aren't really worried that users/pods/attackers are going to try to subvert the isolation; you assume that if the pod starts out isolated, it will stay that way.)

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+### Notes/Constraints/Caveats
+
+**Naming.** This was extensively discussed during the initial reviews
+([kubernetes/kubernetes#141604]):
+
+- "Hermetic" is only the colloquial name of the feature. It describes the
+  contract well: a sealed environment with no external network attachments,
+  like hermetic builds in Bazel or Nix. It is deliberately **not** an API
+  value: reviewers pointed out that "hermetic" can also imply storage, host
+  access or compute isolation, and this KEP does not provide those.
```
danwinship, line 258: You can't say "it describes the contract well" and then agree that it could also imply all sorts of incorrect things. IMO it _doesn't_ describe the contract well. I'm glad you took "hermetic" out of the API but I don't even think it should be part of the name of the KEP.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- "CNI" never appears in the API. CNI is an implementation detail of Linux
+  container runtimes; it does not exist in the Kubernetes API vocabulary.
```
danwinship, line 260: 👍

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- `None` follows existing ecosystem semantics (`docker run --network
+  none`) and the existing `dnsPolicy: None`.
+
+**The pod still has a network namespace.** The pod keeps loopback
+(`127.0.0.1`, `::1`) and the containers in the pod communicate over
+`localhost`. This is why names like `noNetwork` were rejected: they are
+technically inaccurate.
+
+**Kubernetes API access.** A hermetic pod cannot reach the kube-apiserver (or
```
danwinship, line 269:
```suggestion
**Kubernetes API access.** A hermetic pod cannot necessarily reach the kube-apiserver (or
```

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+[kubernetes/kubernetes#141604]: https://github.com/kubernetes/kubernetes/pull/141604
+[services-networking concepts]: https://kubernetes.io/docs/concepts/services-networking/
+
+### Risks and Mitigations
+
+**Fail-open on version skew (security risk).** The scheduler keeps hermetic
+pods away from unsupported nodes using [Node Declared Features (KEP-5328)],
+but pods can bypass the scheduler (`spec.nodeName`, DaemonSets, static pods).
+An old runtime or kubelet that does not understand the field could silently
+attach an isolated pod to the cluster network. The kubelet **fails closed**
```
danwinship, line 287: I think this violates skew policies. AIUI you have to do the rollout such that in the release where `networkMode` becomes beta, the oldest skewed kubelet supported by that apiserver knows how to handle it.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+rejects the pod with `PodAdmitFailed`. See
+[Version Skew Strategy](#version-skew-strategy).
+
+[Node Declared Features (KEP-5328)]: https://github.com/kubernetes/enhancements/issues/5328
+
+**Ecosystem assumptions that every pod has an IP.** Controllers, meshes and
+operators that read `status.podIP` will see an empty value. This is the same
+state as a pod whose sandbox is still being created, so well-behaved
+consumers already tolerate it. The endpoints controllers get an explicit
+check anyway.
```
danwinship, line 299: Pods that are still being created won't be `Ready`. It's definitely possible that there are ecosystem components that will do bad things if they see `Ready` pods with no IP. Of course, this only affects people who are trying to use the new feature, and the mitigation for them is "stop using the new feature until your third-party components have been upgraded".

(EDIT: but see below about if the pod network implementation doesn't handle this...)

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+func validateIsolatedPodSpec(spec *core.PodSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if spec.HostNetwork {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networkMode"),
+			core.PodNetworkModeNone, "must not be \"None\" when hostNetwork is true"))
+	}
+	if spec.DNSPolicy != core.DNSNone {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("dnsPolicy"),
+			spec.DNSPolicy, `must be "None" when networkMode is "None"`))
```
danwinship, line 480: Disagree. It should _default_ to None, but we shouldn't require it. The pod may be set up in some way that the DNS config still works.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+	}
+	if spec.EnableServiceLinks == nil || *spec.EnableServiceLinks {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("enableServiceLinks"),
+			spec.EnableServiceLinks, `must be false when networkMode is "None"`))
```
danwinship, line 484: ditto

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+	}
+
+	podshelper.VisitContainersWithPath(spec, fldPath, func(c *core.Container, cFldPath *field.Path) bool {
+		// Network probes (httpGet, tcpSocket, grpc) cannot succeed without a
+		// network interface; only exec probes are permitted.
```
danwinship, line 489: Agree with this; this is about _kubelet_'s behavior, and we don't (currently) define any way for kubelet to perform http/tcp/grpc probes when there is no `podIP`. The pod can't do anything clever to work around that.

Of course, if we later define some way for kubelet to do network probes without using the pod network, then we could relax this validation.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+						`may not be set when networkMode is "None"`))
+				}
+				if h.TCPSocket != nil {
+					allErrs = append(allErrs, field.Forbidden(hPath.Child("tcpSocket"),
+						`may not be set when networkMode is "None"`))
+				}
+			}
+			checkHandler(c.Lifecycle.PostStart, cFldPath.Child("lifecycle", "postStart"))
+			checkHandler(c.Lifecycle.PreStop, cFldPath.Child("lifecycle", "preStop"))
+		}
```
danwinship, line 527: (This is a lot of distinct checks and I worry that we'll introduce some additional probe in the future and forget to update the validation here... but I'm not sure there's anything we can do?)

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+
+		// hostPort mappings require an attached network interface.
```
danwinship, line 529: Agree. `hostPort` is implicitly defined in terms of `podIP`, so no `podIP` means no `hostPort`s.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+- `test/e2e_node/hermetic_test.go` (node e2e): sandbox creation without
+  network plumbing, fail-closed admission against a runtime without support,
+  kubelet restart idempotency (no sandbox recreation due to missing IP).
+- Until upstream runtime releases exist, CI runs against kind with the
+  patched containerd reference implementation; a periodic job will be added
+  under sig-network.
+
+- [test name](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/network/hermetic.go): links to [SIG Network testgrid](https://testgrid.k8s.io/sig-network-misc?include-filter-by-regex=PodNetworkMode) and [triage](https://storage.googleapis.com/k8s-triage/index.html?test=PodNetworkMode) to be added once the job exists.
+
+### Graduation Criteria
```
danwinship, line 709: We need some criteria about pod network implementations ("CNI plugins").

I would not be surprised if some of them freak out when they see `Ready` non-`hostNetwork` pods that they were never told about. (eg, when trying to apply NetworkPolicy). We may want some sort of criteria like "_N_ pod network implementations have been shown to work fine (including not excessively logging errors) in the presence of `networkMode: None` pods".

Especially... some implementations may blindly assume that `net.ParseIP(pod.spec.podIP)` will always return a non-`nil` result for any `Ready` non-`hostNetwork` pod, and thus it may be possible for any user who can create Pods to crash such a pod network implementation just by creating a `networkMode: None` pod. (Related: should `networkMode` be protected by PSA?)

danwinship (commented):
Generally +1

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+  networkMode: None
+  containers:
+  - name: worker
+    image: registry.k8s.io/e2e-test-images/agnhost:2.53
+    command: ["sleep", "3600"]
+```
+
+Semantics of `networkMode: None`:
+
+1. The API server defaults `dnsPolicy` to `None` (without requiring
```
bowei, line 182: default or enforce to None? Would be good to make this very explicit.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+   `dnsConfig.nameservers`) and `enableServiceLinks` to `false`, and validates
+   that no network-dependent feature is requested (see [Validation](#validation)).
+2. The kubelet uses a new CRI field to tell the runtime to create the sandbox
+   network namespace with only the loopback interface and to skip all the
+   network plumbing (CNI is never invoked on Linux runtimes).
+3. The pod runs with empty `status.podIP`/`status.podIPs` and becomes `Ready`
```
bowei, line 188: This will probably break some controllers in the ecosystem. Would be good to see how we can make it obvious/mitigate issues.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1047 @@
...
+   without waiting for an IP. `status.hostIP` is still reported.
+4. The endpoints and endpointslice controllers never select the pod into
```
bowei, line 190: Need to also add something about NetworkPolicy (and friends)..?

bowei (commented):
Intriguing proposal towards addressing some of the user stories that have come up over the years wrt to use of "other" networks.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: yeah, been struggling with the name too, pod network sounds more accurate indeed

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: yes, this was one feedback from node folks, there is the chance that an old kubelet does not understand the new field and gives the pod a network, see Line 769 `## Version Skew Strategy`

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: Will relax the sentence, to `One of the missing pieces` ... it was more introductory than authoritative

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: ack

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: It comes from the context on some comments about this model breaking current network model, so tries to address that comment ... this proposal is additive and backwards compatible, what worked before keeps working the same way

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: ack

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: I went back and forth between high level or granular, and based on the comments on the doc and the PR I submitted I have found that high level was interpreted very differently depending on the background of the person, so opted for granular.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: there is a sentence about in Line 295

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: We do not mutate, if is not set  we default if is set to an invalid value we fail in validation, see below on  Validation section, line 478 of the doc

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: ~adding some variables that kubernetes know are invalid for that Pod can create hard to debug situations. All kubernetes sdks default to those values, from Kubernetes POV adding them will be confusing ... and if there is going to be an owner of those "external" additional network they should be the ones populating those variables, since most likely the value will be different and you can not change the env variable of a process that is already running IIUIC~

Oh, you mean just defaulting to false `enableServiceLinks` and allow the user to decide if they want to inject them and take the risk? ... that sounds fair, need to check if `enableServiceLinks` allow that, IIRC it was not a pointer to a bool

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: There is an admission phase in the kubelet ... it may be confusing  the wording and you are conflating it with apiserver admission? I will make it more clear

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: ack

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: ack

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: Reworked: instead of never injecting the vars, for `networkMode: None` pods the KUBERNETES_* variables will follow `enableServiceLinks` (defaulted to false for these pods, overridable), consistent with how dnsPolicy is handled.

The reason for not injecting them by default is debuggability: all the SDKs' in-cluster config prefers KUBERNETES_SERVICE_HOST, and they fail fast with a clear error when it is unset, instead of hanging on dial timeouts.  Also, if an external driver provides connectivity, the apiserver may be reachable at a different address over that network; the owner of that connectivity is the right party to set the variables (explicit container env takes precedence over injection), since env vars cannot be changed after the process has started.

A side benefit: with nothing to inject by default, the kubelet no longer needs the service informer to have synced to start these pods removing one control-plane dependency from the isolated-pod startup path.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: what about `networkMode: [ Default | Host | None ]`, that is your last option, plus `Host`, so a pod has exactly one network mode and we avoid invalid combinations. The original proposal was a bool and had a lot of pushback. In addition, absorbing `hostNetwork` into `networkMode` makes the API easier to think about (hostNetwork moves to be a mirror field and deprecated, and there can be only one `networkMode` for a pod 

The name `podNetwork` looks like an antipattern, it is  already nested inside the Pod so is redundant, other fields follow the same model,   dnsConfig (PodDNSConfig), securityContext (PodSecurityContext), os (PodOS).

"None kind of implies can't have networking" that is fair, we need to make clear it follows the same semantics as `docker` and `podman` `--network none`. Having consistency across ecosystems seems a good thing to avoid fragmentation

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: The current CNI plugins add overhead to any interface or networking in the Pod so it impact performance, this is a common case with RDMA devices ... I had to implement a `disableEBPF` option in DRANET just because of that... I need to rewrite this to make it more clear the user story

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: Hm... I get your argument, though I dislike tying the `KUBERNETES_SERVICE_HOST` behavior to `enableServiceLinks`, because `enableServiceLinks` is dumb and we kind of want everyone to set it to `false`...

I don't currently have a better suggestion though.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: > "None kind of implies can't have networking" that is fair, we need to make clear it follows the same semantics as `docker` and `podman` `--network none`.

Right, but for `docker --network none`, I think there's a very strong assumption that if you say that then you actually do want _no_ networking (like in your 2nd and 4th user stories). And even if you do use `docker --network none` and then manually add NICs to the container, there's still "no networking" _as far as docker is concerned_ because docker won't know about those NICs. Whereas in the case of this KEP, the expectation is that some users will add NICs via DRA or future Multi-Network APIs, and Kubernetes _will_ know about those. So it's not "no networking", it's just "no cluster-default pod network". Hence `defaultPodNetwork: false`, which makes it clearer what behavior the pod is requesting and what kubelet/runtime will actually do differently.

> The name podNetwork looks like an antipattern, it is already nested inside the Pod so is redundant

I don't think this is like the other examples you gave; the "pod" in `podNetwork` / `defaultPodNetwork` doesn't refer to the `v1.Pod` that contains the field, it's referring to an external thing which is called a "pod network". (It's like, if a `Service` had to refer to a `ServiceAccount`, we would call the field `serviceAccount` or `serviceAccountName`.)

> In addition, absorbing `hostNetwork` into `networkMode` makes the API easier to think about

I think that depends on how you're thinking about the field. If it's a field for configuring networking in the pod in general like docker's `--network`, then yes. But if it's a field for configuring how the pod network works specifically, then no.

danwinship: (Note: I'm bumping up against terminology that the Multi-Network subproject has discussed but I don't think ever fully agreed upon. In particular, "default network", "primary network", etc.)

danwinship: > > In addition, absorbing `hostNetwork` into `networkMode` makes the API easier to think about
> 
> I think that depends on how you're thinking about the field.

Hm... actually, we (or at least "I"?) tend to refer to "host-network pods" and "pod-network pods", and from that point of view, "no-network pods" is clearly a third orthogonal type with respect to expectations of how different APIs will work with the pod... (eg, "NetworkPolicy `podSelector`s select pod-network pods, but not host-network or no-network pods.)

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: Yeah, I think more generally, this KEP marks the point where we need to clearly define how core APIs interact with non-core networking.

(Note: "CDPN" = "cluster-default pod network")

- Can `pod.status.podIPs` ever refer to IPs that are not on the CDPN (other than the `hostIPs`, for `hostNetwork` pods)? (Multi-Network considered this use case important at one point, but more recent work doesn't seem to care. I think we want to commit to "NO" here. Note that if we say "YES" then some of the questions below have to be split into separate "CDPN" and "`pod.status.podIPs`" questions.)
- Can `hostPort` ever forward to pod IPs other than `pod.status.podIPs`? (I argued "NO" above.)
- Are pods that are connected to neither the CDPN nor the host network namespace _required_ to be able to reach service IPs? (Obviously "NO" in the case where there's actually no networking, probably "NO" even if there is; consider SR-IOV/ipvlan interfaces.)
- Are pods that are connected to neither the cluster-default pod network nor the host network namespace _allowed_ to be able to reach service IPs? (Probably "YES"; consider eBPF-connect-intercept-based proxying as in Cilium.)
- Are pods that are not connected to the cluster-default pod network _required_ to be able to make use of the default pod DNS config? (Again, obviously "NO" when there's no networking, and probably "NO" when there is.)
- Are pods that are not connected to the cluster-default pod network _allowed_ to be able to make use of the default pod DNS config? (Obviously "YES"; the admin can choose to create the DNS config in such a way that it will work for everyone.)
- How does Service interact with non-default networking? (As of 1.37, I think a Service's `selector` specifically selects `pod.status.podIPs`, which are assumed to be reachable over the CDPN, meaning it basically ignores `networkMode: None` pods. Future Multi-Network APIs may change this.)
- How does NetworkPolicy interact with non-default networking? (As of 1.37, I think `podSelector`/`namespaceSelector` specifically select traffic which goes to/from `pod.status.podIPs` over the CDPN, and is not expected to apply to traffic to/from other IPs or not travelling over the CDPN, meaning it basically ignores `networkMode: None` pods. This behavior may not be consistent between implementations. Future Multi-Network APIs may change this.)
- How does the EndpointSlice _controller_ interact with non-default networking? (As of 1.37, it only ever creates slices using `pod.status.podIPs`, meaning it basically ignores `networkMode: None` pods. Future Multi-Network APIs may change this.)
- How does the EndpointSlice _API_ interact with non-default networking? Specifically, does having a `targetRef` pointing to a Pod imply that `addresses` is the `pod.status.podIPs` of the Pod / that the `addresses` are reachable over the CDPN? Is that implied for `addresses[0]` but not `addresses[1:]`? Is it implied for `addressType: IPv4` and `IPv6` slices, but not necessarily for future address types?

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: > hereas in the case of this KEP, the expectation is that some users will add NICs via DRA or future Multi-Network APIs, and Kubernetes will know about those

yep, but not leaking their problems in to the core, same as CRDs , you can build ON kubernetes but kubernetes does not absorb your problem, it just let you fix it providing the substrate

> - Designing a multi-network or secondary-interface attachment framework.
  `networkMode` is intentionally about the absence of automatic attachment;
  it provides a clean surface for delegation (e.g., DRA drivers), but how
  interfaces are attached afterwards and its lifecycle is out of scope.

> docker --network none

that is the exact behavior we are targeting here, same as you describe it

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: yeah, I need to flesh out clearly the details on the section but you are right, this just changes the graduation steps I guess

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
aojea: @danwinship do you feel we should also enforce the pod.status.podIPs to not be mutable if the pod has `networkMode: None` ? or do we leave open as today and delegate the problem to RBAC?

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: What I'm saying is that "None" sounds like it means "none", not like it means "do-it-yourself".

Looking at other `None` values in  the API...

- `None` means "none":
  - `admissionregistrationv1.SideEffectClassNone`: asserts that calling a webhook will have no side effects
  - `v1.AzureDataDiskCachingNone`: creates a volume with no caching (neither `ReadOnly` nor `ReadWrite`)
  - `v1.ClusterIPNone`: creates a service with no `ClusterIP` (neither manually nor automatically assigned)
  - `v1.MountPropagationNone`: creates a mount with no mount propagation (neither `HostToContainer` nor `Bidirectional`)
  - `v1.ServiceAffinityNone`: creates a service with no session affinity
  - `storagev1.NoneFSGroupPolicy`: indicates that a CSI driver does not support ownership/permission changes
- `None` means ... "not none":
  - `v1.DNSNone`: still uses `pod.spec.dnsConfig`, just tells kubelet to not add anything extra to it.
  - `resourcev1.DeviceTaintEffectNone`: means k8s itself doesn't do anything with the taint but "DRA drivers can use it to communicate exceptions to admins or other controllers"

So I guess the last 2 examples are similar to `networkMode: None`. But also, `DNSPolicy` is hardly the sort of API that you want to be copying; it has a value called `Default` that _isn't_ the default value for the field!

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: I think if we say that `pod.status.podIPs` can only contain the CDPN IPs (which I think is probably a good idea) _and_ that being `networkMode: Cluster` is the _only_ way to attach a pod to the CDPN (which seems very reasonable), then it logically follows that it would never make sense to modify the podIPs of a `networkMode: None` pod, and so it would make sense to enforce that.

keps/sig-network/6313-hermetic-network-isolated-pods/README.md
danwinship: Oh, another mode possibility: in ovn-kubernetes, pods attached to a (primary) UserDefinedNetwork end up with a "vestigial" connection to the cluster-default pod network, which can only be used for kubelet probes, DNS, and apiserver connectivity. As implemented, this is very hacky, but there may be some less-hacky version of it that could be useful.

There had also been discussion somewhere at one point about having kubelet set up a private IPv6-only link-local per-node pod network for k8s-internal uses, though that never got beyond the extremely-hand-wave-y stage.

aojea: On the enum over `defaultPodNetwork: [true|false]`. Enum is an extension point and a future KEP could allow things like: `networkMode: ovn.org/user-defined-network`, `networkMode: networking.gke.io/blue`. This is one of the main problem from Multi-Network.
So, we avoid defining what a "pod network" is (L2 vs L3, OVS vs eBPF vs netfilter vs DPDK...) and the field defines who provides the pod's networking, not what the network is. Your UDN example fits here too.

Taking this future extension into account, `None` now makes more sense because it really means none, nothing will provision any network , neither Kubernetes or a referenced owner. 

Today the field has the expectations of `do-it-yourself`, but if we put this into a future extension perspective, then is the same as `docker/podman --network none`. I think is easier to inherit those widely understood semantics than search for a new word: we were never consistent and we can leverage the inertia.

On the name: touche on serviceAccountName. But coming back to the argument of defining the provider instead of the abstract concept of network,  `podNetwork: Host` (or a future `podNetwork: ovn.org/x`) reads wrong, because the field isn't selecting the pod network — it selects how the pod's networking is provided.

danwinship: > Taking this future extension into account, `None` now makes more sense because it really means none, nothing will provision any network , neither Kubernetes or a referenced owner.
> 
> Today the field has the expectations of `do-it-yourself`

This is the part I completely disagree with. If you do-it-yourself and end up with some network, then it is not the case that "it really means none". If you provision a network yourself, then it is not the case that "nothing will provision any network". If you use DRA to do it and end up with a ResourceClaimStatus pointing back to the object that did the provisioning, then it is not the case that "nothing will provision any network, neither Kubernetes or a referenced owner".

It feels like maybe when you say "network" you're just implicitly talking only about the default pod network and so `networkMode: None` just means "there is no default pod network" and so there is no contradiction with somebody adding _some other network_ to the pod. But to me, `networkMode: None` doesn't sound like it means "there is no default pod network", it sounds like it means "there is _no networking at all_". And since that's not actually what it's supposed to mean, I don't like that name.

> the same as `docker/podman --network none`. I think is easier to inherit those widely understood semantics than search for a new word

Again, I think the widely-understood semantics of `docker --network none` are "no networking", _not_ "do-it-yourself networking". Yes, you _can_ do DIY networking on top of `docker --network none`, but [they don't document that as a possibility](https://docs.docker.com/engine/network/drivers/none/), which is a clear contrast to your feature here, where more than half of your user stories involve DIY networking.

> `podNetwork: Host` (or a future `podNetwork: ovn.org/x`) reads wrong, because the field isn't selecting the pod network — it selects how the pod's networking is provided.

Right, `hostNetwork` pods would have `podNetwork: false` or `podNetwork: None` or whatever.

And while `podNetwork: ovn.org/user-defined-networks` doesn't make sense, `podNetwork: pepsi-udn` (ie, referencing a specific network, not a type of network) would. (If we were going to support that sort of thing. Which we probably aren't.)

danwinship: while booleans are bad (m'kay?), `podNetwork: [true|false]` is a pretty obvious parallel to `hostNetwork: [true|false]`. Your pod is connected to the host network if and only if `hostNetwork` is `true`. Your pod is connected to the (cluster-default) pod network if and only if `podNetwork` is `true`. If neither is `true` then you're not connected to either network. If both are `true` then your object is invalid. Nice and easy to understand and reason about...

also, parallelism of `hostNetwork`/`hostIPs` and `podNetwork`/`podIPs`...

danwinship: Or `defaultNetwork: [ Pod | Host | None ]` (or `defaultNetworking`?)

aojea: > parallelism of hostNetwork/hostIPs and podNetwork/podIPs...

The parallel holds in the names but not in the semantics. `hostIPs` is set for every pod, and `hostNetwork: true` means `podIPs = hostIPs`:

| | `hostNetwork` | `podNetwork` | `status.hostIPs` | `status.podIPs` |
|---|---|---|---|---|
| Cluster-default pod network | false | true | node IPs | pod IPs (≠ hostIPs) |
| Host network | true | false | node IPs | node IPs (= hostIPs) |
| None | false | false | node IPs | empty |

So `podNetwork: false` means two different things for `podIPs` depending on the other field

Ok, I think we are converging and IIUIC we are trending towards consolidate the `default network` concept across the APIs [ and `status.podIPs` only means cluster-default-pod-network IPs](https://github.com/kubernetes/enhancements/pull/6317#discussion_r3926470353)

Anything else builds on top of `None`.

| Option | Pros | Cons |
|---|---|---|
| `podNetwork: [true\|false]` (or `defaultPodNetwork`) | Simple "on the pod network if true", mirroring `hostNetwork` | Two booleans encode a 3-state mode and API conventions: bools "trend towards a small set of mutually exclusive options" and can't grow. The default must be `true` unless `hostNetwork: true`, so it's a `*bool` with a default derived from another field,. The `hostIPs`/`podIPs` parallel holds in the names but not the semantics (see above). |
| `networkMode: [Default\|Host\|None]` (current text) | One field, one mode; extensible; absorbs `hostNetwork` as a mirror | `None` reads as "no networking at all", nothing in the field name means it refers the default network; "mode" is vague |
| `defaultNetwork: [Pod\|Host\|None]` | `None` unambiguously means "no default network"; each value maps 1:1 to a `podIPs` behavior; extensible; `default*` field names have precedent (`defaultMode`, `defaultBackend`, LimitRange `default`/`defaultRequest`); "cluster-wide default network" is an established vocabulary already used in the ecosystem | `defaultNetwork: Host` reads slightly oddly |

I'm +1 on `defaultNetwork: [Pod|Host|None]`

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+   pods default to a configuration with no network dependencies. Users may
+   override these defaults when an external integration (a DRA driver, VM
+   networking) provides connectivity by other means, in line with the goal
+   of offering a delegation contract to external integrations.
+   Configurations that cannot work without a pod IP (network probes, host
+   ports) are rejected by validation (see [Validation](#validation)).
+2. The kubelet uses a new CRI field to tell the runtime to create the sandbox
+   network namespace with only the loopback interface and to skip all the
+   network plumbing (e.g., runtimes that normally invoke CNI plugins would
+   skip that step).
```
EdDev, line 203: The proposal focuses on the default network, and attempts to still allow secondary networks through other delegation mechanisms.
However, the current ecosystem is heavily dependent of CNI delegation (e.g. through Multus). Some scenarios have been already detected which ask to keep the CNI delegation in place in parallel to the new DRA method (e.g. https://github.com/kubevirt/enhancements/pull/428).

Virtualization solutions (e.g. Kubevirt) commonly do not wire the default pod network to the VM, however in many cases it will wire a secondary network.

I think there is a challenge  here on how to express that the default pod network should not be attached  to the pod and still preserve the existing mechanics to attach secondary networks. Assuming that this can be done only with the new DRA mechanics is limiting, as we have observed that it is still needed at runtime.

The limitation and implications should either be acknowledge or a mechanism to resolve them should be provided. The field name `defaultNetwork` is a bit misleading as it implicitly effect secondary networks as we know them today.

aojea: > The limitation and implications should either be acknowledge or a mechanism to resolve them should be provided. The field name defaultNetwork is a bit misleading as it implicitly effect secondary networks as we know them today.

This allow users to say "I want a pod without network", it is a field and is opt-in, existing things keep working as they are working today and projects can also implement admission policies for not allowing this new type.

Projects and integrations have DRA that is a core API to provide secondary networks. We provide a core and native mechanism to support this pattern, it's unfortunate the ecosystem build onto an undocumented and unsupported hook for this, but this tries to amend that and provide a path and supported path forward

EdDev: > "I want a pod without network"

The field name implies it is related to default pod network only. Projects like Kubevirt who use mostly secondary networks may incorrectly interpret it as if secondary are still available.

> Projects and integrations have DRA that is a core API to provide secondary networks.

DRA by itself, AFAIK is not yet capable to answer all needs for secondary networks, without adding functionality/format which is not yet fully agreed on. E.g. requesting a specific MAC, MTU and other pod-interface specifics.

I think it is totally fine to frame this enhancement to "pod without network", secondaries included. And mention that extension to secondary network can be done through native DRA (only?). Adding a note about supporting CNI delegation, that is no longer possible, explicitly will make the acknowledgement to the exact implications this option has. I would also try to find a clearer name for the field, per my previous note. But I do acknowledge that the CNI delegation technique everyone depends on today is not formal from Kubernetes point of view.

aojea: > The field name implies it is related to default pod network only.

see this thread with @danwinship https://github.com/kubernetes/enhancements/pull/6317#discussion_r3925924364 . Kubernetes does not have explicit concept of network, but was clear after all the multinetwork discussion is that there is a strong convention that everything is connected to the host network (probes, dns, ...)


> E.g. requesting a specific MAC, MTU and other pod-interface specifics.

I think we are mixing things, kubernetes was never aware of these things, and really does not need to ... why should be aware now? 

Is it possible you are conflating things here, maybe the references to secondary networks is what confuses things ... this is not defining multi network or anything related, that is an anti goal, this is allowing people that wants to totally handle the Pod network to guarantee that Kubernetes will not touch it ... but this is a consequence, not the end goal

EdDev: I am pointing out that this proposal will also disable common secondary networks implicitly.
I would claim that the community depends on the non-formal established patterns that uses CNI delegation (e.g. Multus) and IMO the KEP should make this clear. Calling this explicit is what I feedbacked about.

The details about secondary networks needs, not being fully supported by the native DRA options, was an attempt to explain why new modern solutions are not yet there. The KEP is describing that secondary networks are out-of-scope and can still be supported, through DRA. I challenged that point in relation to how secondary networks are handled today (outside Kubernetes core).

Considering that the original network requirements/assumption from Kubernetes is to always connect the pod network, and this proposal suggestions to change it, I find it reasonable to attend explicitly to existing solutions that depend on it to support secondary networks.

danwinship: OK, so there are 2 parts to this:

First, independently of this KEP, we do totally want to make it possible to eventually replace Multus with something less hacky, _probably_ DRA+NRI based.

Second, more specifically, in this KEP, the official semantics of `defaultNetwork: None` are "the pod is not connected to the default pod network". You're right that the parenthetical comment here "(e.g., runtimes that normally invoke CNI plugins would skip that step)" would exclude Multus-based secondary networks as well, but that wasn't the intent, and it would be correct for a runtime that uses Multus to implement `defaultNetwork: None` by still invoking Multus to create secondary networks, while skipping the attachment of the default network.

Of course, there's no way for the runtime to do that right now, because containerd/cri-o don't actually know about Multus, so they have no way of knowing that the CNI plugin they're using creates both primary and secondary networks, and even if they did, they have no way of communicating to Multus that they want it to only create the secondaries. And that's an example of what I was saying before about Multus being hacky; none of the other parts of the system currently understand what Multus is doing, which makes it hard to extend/change the behavior around it. (Whereas with DRA-based secondaries, it's much more explicit at the k8s API level what is going on.)

Anyway, at the k8s API level, nothing would change to support this; the user expresses `defaultNetwork: None`, and then kubelet sets the `network_mode` in the `PodSandboxConfig` accordingly when invoking CRI. (@aojea: maybe rename that to better match the renamed pod field?). What we need is some CNI-level API so that cri-o and Multus can still do the right thing after that.

aojea:
/assign @soltysh 
for PRR

keps/sig-network/6313-network-isolated-pods/README.md
aojea: @danwinship @EdDev to close on this, I think there are multiple alternatives to keep multus being invoked with these new flavor of pods:
- Using OCI hooks via CDI or NRI
- Using some crio magic
- ...

I have some code here to demonstrate https://github.com/aojea/network-device-plugin so if the solution is acceptable there should not be any problem of adapting multus to work .

We can also a criteria to support multus in order to graduate to beta if that helps

danwinship: > Using OCI hooks via CDI or NRI

To clarify this a bit (Antonio and I were talking about this yesterday), right now Multus hooks into the CNI calls used for primary pod network setup because at the time Multus was originally written that was really the only option for doing additional stuff inside the pod network namespace at pod creation time. But if someone was writing Multus for the first time today, it would clearly make more sense to do it via NRI hooks, independently of the primary pod network CNI calls. So maybe the right fix is to make Multus work that way.

keps/sig-network/6313-network-isolated-pods/kep.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v1.38"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v1.38"
+  beta: "v1.39"
```
danwinship: Need to figure out the skew requirements. As mentioned above, my understanding is that when adding new fields that kubelet needs to be aware of, you end up needing 3 releases between alpha and beta.

keps/sig-network/6313-network-isolated-pods/kep.yaml
```diff
@@ -0,0 +1,56 @@
...
+  stable: "v1.41"
+
+# The following PRR answers are required at alpha release
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: PodDefaultNetwork
+    components:
+      - kube-apiserver
+      - kube-controller-manager
```
danwinship: I don't think kcm needs to care? [The EndpointSlice controller already ignores pods with no `podIPs`](https://github.com/kubernetes/kubernetes/blob/v1.37.1/staging/src/k8s.io/endpointslice/util/controller_utils.go#L180-L190).

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+This KEP adds **network-isolated Pods**: Pods that opt out of the default network.
+An isolated pod gets a network namespace with only a loopback interface and is
+not reachable from the cluster.
```
danwinship:
```suggestion
not reachable from the cluster unless other network connectivity is added to it somehow.
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+4. The endpoints and endpointslice controllers never select the pod into
+   Endpoints/EndpointSlices, even if a Service selector matches it.
+5. For isolated pods, `enableServiceLinks` also governs the `KUBERNETES_*`
+   master-service variables, which are injected unconditionally for all
+   other pods: with the defaulted `false` no service environment variables
+   are injected, and an explicit `enableServiceLinks: true` restores the
+   standard injection for pods that obtain connectivity by other means.
+   The kubelet maps the pod hostname to `127.0.0.1` and `::1` in the
+   managed `/etc/hosts`, so `localhost` and `$(hostname)` resolve inside
+   the sandbox.
```
danwinship: #5 is two separate points (service links, hostname)

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+metadata:
+  name: network-isolated-example
+spec:
+  defaultNetwork: None
+  containers:
+  - name: worker
+    image: registry.k8s.io/e2e-test-images/agnhost:2.53
+    command: ["sleep", "3600"]
+```
+
```
danwinship: maybe explain the defaulting of `defaultNetwork` here?

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation—e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
```
soltysh: Nit: can you ensure the right checkboxes are ✔️

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+
+What has worked is composing on top of stable core primitives instead of
+extending the core network model:
+
+- [Kubernetes Network Drivers][knd-paper] defined a new architecture, based
+  on DRA, to provide a native mechanism to plumb additional network devices.
+  [DRANET] is a project that demonstrated the successful use of this
+  architecture to solve the challenges of the new AI/ML workloads and RDMA
+  networks.
+- [KEP-4962] (GA) added a standard way to report network data in DRA. All
```
soltysh: Is this the correct link? I'm looking at https://github.com/kubernetes/enhancements/issues/4962 and it's a closed, not implemented feature, and you're mentioning it GA-ed.

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+## Design Details
+
+### API Changes
+
+New enum type and `PodSpec` field:
+
+```go
+// PodDefaultNetwork describes how the pod is connected to the default
+// network, the network that Kubernetes connects every pod to unless the
+// pod opts out.
```
danwinship:
```suggestion
// PodDefaultNetwork describes the default networking state of the pod.
```

`defaultNetwork: Host` describes something more than just the pod's relationship to "the network that Kubernetes connects every pod to unless the pod opts out".

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+// +enum
+type PodDefaultNetwork string
+
+const (
+	// PodDefaultNetworkPod gives the pod its own network namespace attached
+	// to the default network. The container runtime performs its configured
```
danwinship:
```suggestion
	// PodDefaultNetworkPod gives the pod its own network namespace and
	// attaches it to the default pod network (the network that Kubernetes
	// connects every pod to unless the pod opts out). The container runtime performs its configured
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+	// network plumbing and the pod is assigned routable pod IPs. This is
```
danwinship: "routable pod IPs" is often used to mean "pod IPs that are reachable from outside the cluster". Just remove "routable"?

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+
+	// PodDefaultNetworkHost runs the pod in the host's network namespace.
+	// It is equivalent to setting hostNetwork: true; the two fields are
+	// kept in sync by API defaulting.
+	PodDefaultNetworkHost PodDefaultNetwork = "Host"
+
+	// PodDefaultNetworkNone gives the pod its own network namespace
+	// containing only the loopback interface and does not attach it to the
+	// default network. The container runtime performs no network plumbing
+	// and the pod is assigned no IP. The pod is never selected into
```
danwinship:
```suggestion
	// default network. The Pod's podIPs are left unset. The pod is never selected into
```

It's not necessarily correct to say that the runtime "performs no network plumbing". It performs no network plumbing _related to the default pod network_, but you basically already said that anyway.

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+	// Services and, by default, receives no cluster DNS configuration or
```
danwinship:
```suggestion
	// and the pod is assigned no IP. The pod is never selected into
	// EndpointSlices and, by default, receives no cluster DNS configuration or
```

Nitpicky / possibly-more-future-proof

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+	// service environment variables.
+	PodDefaultNetworkNone PodDefaultNetwork = "None"
+)
+```
+
+```go
+type PodSpec struct {
+	// ...
+
+	// defaultNetwork selects how the pod is attached to the default network.
```
danwinship:
```suggestion
	// defaultNetwork selects the pod's default network.
```

or

```suggestion
	// defaultNetwork describes the pod's default networking state.
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+	// "Pod" gives the pod its own network namespace attached to the default
+	// network, "Host" runs the pod in the host network namespace
+	// (equivalent to hostNetwork: true and kept in sync with it), and
+	// "None" gives the pod an isolated network namespace with only a
+	// loopback interface, not attached to the default network and with no
+	// automatic network plumbing.
+	// When "None" is selected, dnsPolicy defaults to "None" and
+	// enableServiceLinks defaults to false (both may be overridden), and
+	// probes, lifecycle handlers and hostPorts that require networking
+	// are forbidden.
```
danwinship:
```suggestion
	// features that require networking (such as hostPorts and network-based probes and lifecycle handlers)
	// are forbidden.
```

(the original text kind of implies that some hostPorts don't require networking)

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+- The remaining conflicts (`defaultNetwork: Pod` or `None` combined with an
+  explicit `hostNetwork: true`) are rejected by validation.
+- Setting one field from another has a known problem on updates. An old
+  client that does not know `defaultNetwork` and patches `hostNetwork: false`
+  into a workload template that has `defaultNetwork: Host` stored will not
+  see its change: defaulting sets `hostNetwork` back to `true` from the
+  stored enum. A full update (PUT) from the same client works, because the
+  client drops the unknown field and `defaultNetwork` is set again from
+  `hostNetwork`. Pods are not affected, `hostNetwork` is immutable, and the
+  problem goes away once clients are updated.
```
danwinship:
```suggestion
- Setting one field from another has a known problem on updates, but
  since `hostNetwork` and `defaultNetwork` are both immutable, this
  doesn't affect us.
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+
+Additional validation adjustments:
+
+- `dnsPolicy` and `enableServiceLinks` are **defaulted, not enforced**.
+  Isolated pods default to a configuration with no network dependencies
+  (`None`, `false`), but external integrations that attach connectivity by
+  other means (DRA drivers, VM networking) are part of the contract: users
+  may override the defaults to match the connectivity those integrations
+  provide.
+- Only `exec` probes and `exec` lifecycle handlers pass validation.
```
danwinship: (Not true; `sleep` lifecycle handlers work too.)

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+    // PodSandboxStatus.
+    NONE = 1;
+}
+
+message PodSandboxConfig {
+    // ... existing fields ...
+
+    // network_mode instructs the runtime how to provision networking for
+    // the sandbox network namespace. Host-network sandboxes continue to be
+    // requested via NamespaceOption (NamespaceMode NODE) and use POD here.
```
danwinship: `POD` for host-network is surprising, and inconsistent with the definition of `POD` above ("POD attaches the sandbox to the pod network")

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+    bool network_mode_none = 4;
+}
+```
+
+Runtime obligations for `NONE` sandboxes:
+
+- create the netns according to `NamespaceOption` as usual;
+- bring up `lo` (`127.0.0.1/8`, `::1/128`), the equivalent of
+  `ip link set lo up`;
+- no plumbing on setup and none on teardown;
```
danwinship:
```suggestion
- no default-pod-network plumbing on setup or teardown;
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+- `dnsPolicy: None` with no `dnsConfig` produces an empty `resolv.conf`. A
+  user-supplied `dnsConfig` is honored: a resolver on loopback inside the
+  pod, or nameservers reachable over driver-injected interfaces. An
+  overridden `dnsPolicy` (for example `ClusterFirst`) is honored exactly as
+  for any other pod.
+
+### Control Plane Behavior
+
+- `ShouldPodBeInEndpoints` (in `k8s.io/endpointslice/util`, shared by the
+  Endpoints and EndpointSlice controllers) returns `false` for
```
danwinship:
```suggestion
  Endpoints and EndpointSlice controllers) already returns `false` for
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+  `defaultNetwork: None` pods, so they are never published even if a Service
+  selector matches.
+- The kubelet publishes support for `defaultNetwork: None` through
+  [Node Declared Features (KEP-5328)] (GA since v1.37), so the scheduler
+  keeps isolated pods away from nodes without a supporting kubelet and
+  runtime. This KEP only registers a new feature in that existing mechanism.
```
danwinship: So does `kube-scheduler` need a feature gate? (`kep.yaml` says "no")

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+  The kubelet fail-closed admission remains the final guarantee for pods
+  that bypass the scheduler.
+
+### Interaction with Core Networking APIs
+
+With `defaultNetwork: None`, "not connected to the default network" becomes
```
danwinship:
```suggestion
With `defaultNetwork: None`, "not connected to the default pod network" becomes
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+a supported state, so this section spells out how the core networking APIs
+behave for those pods. All the core networking APIs work only with the
+default network: the network that `defaultNetwork: Pod` connects pods to,
+and that host-network pods use through the node. The rules below describe
```
danwinship:
```suggestion
behave for those pods. All the core networking APIs work only with the
host network and the default pod network. The rules below describe
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+the current behavior; this KEP does not change any of them for existing
+pods:
+
+- **`status.podIPs`** only ever contains the pod's IPs on the default
+  network (the node IPs for host-network pods). Interfaces attached to isolated pods by
```
danwinship:
```suggestion
- **`status.podIPs`** only ever contains the pod's IPs on the pod's default
  network (pod-network IPs for `defaultNetwork: Pod` or node IPs for
  `defaultNetwork: Host`). Interfaces attached to isolated pods by
```

keps/sig-network/6313-network-isolated-pods/README.md
```diff
@@ -0,0 +1,1225 @@
...
+  rejected by validation for isolated pods.
+- **Service virtual IPs.** Pods that are not attached to the default
+  network are not *required* to be able to reach Service IPs, but
+  they are *allowed* to: an implementation or integration may provide
+  reachability by other means (a connect-time eBPF proxier, a
+  driver-injected interface with the appropriate routes).
+- **Cluster DNS** follows the same pattern: not required, allowed. The
+  `dnsPolicy: None` default expresses "not required"; an explicit user
+  override opting into cluster DNS is honored and works wherever the
+  operator provides a path to the resolvers.
```
danwinship: This is about _access to_ cluster DNS, but we should also clarify that where the _contents_ of cluster DNS refer to pods, it only refers to `podIPs` and thus not to `defaultNetwork: None` pods.

danwinship (approved):
Basically just wording nitpicks at this point; the only real semantic complaint below is with the details of the CRI API.

keps/sig-network/6313-network-isolated-pods/README.md
dchen1107: NodeDeclaredFeatures is built by SIG Node to accelerate and smooth node / dataplane features from the beginning. @pravk03 is the author. The NodeDeclaredFeatures is buit out-of-tree, hence its gate must be enabled on the kube-apiserver, kube-scheduler, and kubelet. This activates the NodeDeclaredFeatures scheduler plugin to match Pod requirements against node-status fields. 

Once NDF framework is enabled individual features (onboarding through the Alpha/Beta/GA process) can publish their capabilities via node.status.declaredFeatures without needing dedicated code changes or new gates in the kube-scheduler.

dchen1107:
/lgtm from the node side, and I am excited and cannot wait to see the potential changes and cleanups this KEP might bring to K8s ecosystem. ;-)

I will leave @danwinship and other network experts on the final approval.

PR #6320, opened 2026-09-02
KEP-6030: Dynamic resize of memory-backed volumes beta update

natasha41575:
Dynamic resize of memory-backed volumes beta update

Issue: https://github.com/kubernetes/enhancements/issues/6030

### DIFF
--- a/keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
+++ b/keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
@@ -97,7 +97,7 @@ tags, and then generate with `hack/update-toc.sh`.
     - [OOM-kills](#oom-kills)
 - [Design Details](#design-details)
   - [API Changes](#api-changes)
-    - [New Status Fields and State Mapping](#new-status-fields-and-state-mapping)
+    - [State Mapping and Status Tracking](#state-mapping-and-status-tracking)
     - [API Validation and Restrictions](#api-validation-and-restrictions)
     - [Resize Restart Policy](#resize-restart-policy)
     - [Atomic Resize Principle](#atomic-resize-principle)
@@ -246,11 +246,11 @@ The mitigation for this risk is that the API server can emit a warning when it s
 
 ### API Changes
 
-#### New Status Fields and State Mapping
+#### State Mapping and Status Tracking
 
 To support dynamic resizing, the `Pod.Spec.Volumes[].EmptyDir.SizeLimit` field is made **mutable** when modified via the `/resize` subresource for existing Pods (when `medium` is `Memory`). This represents the **Desired State**.
 
-We introduce a new sub-structure under `corev1.VolumeStatus` (in `Pod.Status`) to track the actual resize state of the volume:
+This enhancement builds upon the resource state tracking model established in [In-Place Pod Vertical Scaling (KEP-1287)](/keps/sig-node/1287-in-place-update-pod-resources#resource-states), extending the lifecycle progression to memory-backed volumes:
 
 *   **Desired State**: Represented by the mutable `Pod.Spec.Volumes[].EmptyDir.SizeLimit`.
 *   **Allocated State**: Checkpointed and maintained by the Kubelet's `AllocationManager` for local resource accounting, but **not exposed** in `Pod.Status`.
@@ -258,28 +258,12 @@ We introduce a new sub-structure under `corev1.VolumeStatus` (in `Pod.Status`) t
     *   *Initial Admission*: Admitted and checkpointed locally by the Kubelet once the initial resource admission phase completes.
     *   *Admission during Resize*: Upon admitting a resize update, the `AllocationManager` updates the Kubelet's local checkpoint with the newly acknowledged capacity.
     *   *Scope*: Maintained only for memory-backed `emptyDir` volumes.
-*   **Actual State**: Represented by the new `VolumeStatus.EmptyDir.SizeLimit`.
-    *   *Definition*: Represents the actual mounted filesystem capacity reported by the mount/remount call.
-    *   *Pod Creation*: Remains unpopulated (`nil`) until Kubelet completes the initial volume mount setup and receives capacity confirmation from the mounter.
-    *   *Pod Resize*: During an active resize operation, this field retains the *old* mounted size until the Kubelet successfully actuates the `tmpfs` remount, at which point it is updated to the new actual capacity.
-    *   *Default/Fallback*: Remains unpopulated (`nil`) for all non-memory-backed volumes.
-
-The new status field `Pod.Status.ContainerStatuses[].VolumeMounts[].VolumeStatus.EmptyDir.SizeLimit` is only populated for memory-backed volumes, and remains unpopulated for other types of volumes. The `VolumeStatus` Go struct is updated as follows:
-
-```go
-type VolumeStatus struct {
-    ...
-    // emptyDir represents the status of an emptyDir volume.
-    // +optional
-    EmptyDir *EmptyDirVolumeStatus `json:"emptyDir,omitempty" protobuf:"bytes,2,opt,name=emptyDir"`
-}
-type EmptyDirVolumeStatus struct {
-    // sizeLimit represents the actual mounted capacity of the emptyDir volume.
-    // This is only populated for memory-backed emptyDir.
-    // +optional
-    SizeLimit *resource.Quantity `json:"sizeLimit,omitempty" protobuf:"bytes,1,opt,name=sizeLimit"`
-}
-```
+*   **Actuated State**: Checkpointed and maintained locally by `KuberuntimeManager` to track the volume capacity passed to the volume plugin to actuate.
+    *   *Definition*: Represents the target size that the Kubelet successfully actuated via `ResizeEphemeralVolume`.
+    *   *Scope*: Not reported in the API.
+*   **Actual State**: Represents the actual mounted filesystem capacity of the `tmpfs` volume on the node.
+    *   *Definition*: Represents the actual mounted capacity reported by the mount/remount call.
+    *   *Scope*: Unlike container resources, actual volume capacity is **not exposed** in `Pod.Status`. Users can inspect the actual mount size on the node or within the container via `df` command on the volume mount point.
 
 #### API Validation and Restrictions
 
@@ -335,7 +319,7 @@ graph TD
     KRT -->|8b. Updates Cgroups| CRI
     KRT -->|"8c. Remounts Volume (Upsize)"| VM
     VM -->|9. Executes Remount System Call| Kernel
-    KRT -->|"10. Updates Status (Volume Mount Actual State)"| API
+    KRT -->|"10. Clears PodResizeInProgress Condition"| API
 ```
 
 The following steps elaborate on this flow:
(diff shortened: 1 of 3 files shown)


natasha41575:
/assign @tallclair 
node

/assign @jsafrane 
storage

/assign @msau42 
storage and API

/assign @kannon92 
PRR

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -568,20 +595,20 @@ Below are some examples to consider, in addition to the aforementioned [maturity
...
+- [X] Unit test coverage is complete.
+- [X] E2E tests implemented in CI verify successful upsize and downsize operations in a live cluster.
 
 #### Beta
-- Gather feedback from users.
-- Metrics are defined and implemented.
-- Kubelet emits a warning event when `sizeLimit` exceeds pod-level memory limits.
-- The plan evaluates whether integration tests are required.
-- Edge cases around "shrinkage safety" are resolved (e.g., reporting distinct errors when attempting to shrink a volume below current usage).
+- [ ] Gather feedback from users.
```
kannon92, line 603: How are you planning to gather feedback for this feature?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
kannon92, line 684: This isn't quite true right.

`pkg/kubelet/eviction/eviction_manager.go:550` disables emptyDirLimitEviction for all medium: Memory volumes whenever the gate is on — not just for pods using resize:

```
if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScalingMemoryBackedVolumes) && source.EmptyDir.Medium == v1.StorageMediumMemory {
    continue
}
```

At beta default-on this silently changes behavior for every existing workload with a sized memory-backed emptyDir: they get ENOSPC on write instead of pod eviction + reschedule.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -465,24 +448,24 @@ On startup, the container runtime helps to bind the Kubelet's host path into the
 In the existing `InPlacePodVerticalScaling` feature, a best-effort safety logic ensures that container memory limits are not decreased below usage.
 
 #### Alpha
-For Alpha, shrinkage safety checks for memory-backed volumes are out of scope. If a user attempts to shrink a volume below its current usage, the `tmpfs` remount fails at the kernel level with an `EINVAL` error. This error is propagated back to the Kubelet and surfaced in the `PodResizeInProgress` condition as `mount point not mounted or bad option`. The Kubelet will periodically retry the resize operation until it succeeds, is cancelled, or the Pod terminates.
+For Alpha, shrinkage safety checks for memory-backed volumes are out of scope. If a user attempts to shrink a volume below its current usage, `/bin/mount` fails at the kernel level (e.g. on modern Linux kernels using the `fsconfig` tmpfs mount API, returning `fsconfig() failed: tmpfs: Too small a size for current use.` with exit status 32). This error is propagated back to the Kubelet and surfaced in the `PodResizeInProgress` condition. The Kubelet will periodically retry the resize operation until it succeeds, is cancelled, or the Pod terminates.
```
kannon92, line 451: Do we have any ideas how this will behave on windows? Or what version of the linux kernel where this behavior is not expected?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/kep.yaml
```diff
@@ -50,5 +50,6 @@ disable-supported: true
 
 # The following PRR answers are required at beta release
 metrics:
-  - TBD
+  - volume_requested_resizes_total
```
kannon92, line 53: These metric names are not correct. I think you'd need kubelet in the name.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -966,6 +1017,10 @@ details). For now, we leave it here.
 
 ###### How does this feature react if the API server and/or etcd is unavailable?
 
+If the API server or etcd is unavailable, users will not be able to submit new resize requests. Existing requests will be processed by the Kubelet, but the pod status will not be updated with the resize status. 
+
+Once the API server or etcd is available again, new requests will be accepted and processed normally.
+
 ###### What are other known failure modes?
```
kannon92, line 1024: AI is calling out cgroup v1 as a potentail failure case.

8. Cgroup v1 rejection is invisible until it fails. NDF discovery is gate-only (Discover just checks FeatureGates.Enabled), so on a cgroup v1 node the API server accepts the resize and the kubelet then parks it as Infeasible. With the gate default-on at beta this is a new stuck state with no pre-flight signal. Add it to the known failure modes with detection (kubelet_pod_infeasible_resizes_total) and mitigation, or make Discover require cgroup v2.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
kannon92, line 750: Aren't you disabling evictions in place of kernel OOM enforcement? Why would evictions increase?

kannon92, line 869: Should you mention NodeDeclaredFeatures?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -966,6 +1017,10 @@ details). For now, we leave it here.
 
 ###### How does this feature react if the API server and/or etcd is unavailable?
 
+If the API server or etcd is unavailable, users will not be able to submit new resize requests. Existing requests will be processed by the Kubelet, but the pod status will not be updated with the resize status.
```
kannon92, line 1020: Reading this I was thinking of static pods.

Maybe you could mention that this feature is also disabled for static pods?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: The feature was requested by Anyscale (for the IPPR Ray autoscaler integration); we are in contact them and will be engaging them for feedback.

We also have a few other customers internally that we have plans to engage with for feedback. So I anticipate that we will have enough here before code freeze in ~2 months.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -394,13 +381,15 @@ Init containers are commonly used to pre-populate a memory-backed volume (e.g.,
...
-*   The Kubelet does not block the resize but issues a **Warning Event**.
-*   If `sizeLimit` > cgroup memory limits, the volume's `ENOSPC` protection is effectively disabled, as the container is OOM-killed before the volume quota is reached. This is consistent with what pod creation permits today (see [Restricting sizeLimit to not exceed memory limits](#restricting-sizelimit-to-not-exceed-memory-limits)).
+When mounting or resizing a memory-backed volume, the Kubelet today computes the actual mounted `tmpfs` capacity as:
+$$\text{capacity} = \min(\text{NodeAllocatableMemory}, \text{PodMemoryLimit}, \text{VolumeSizeLimit})$$
+
+* If `sizeLimit` is omitted (`nil` or `0`), the volume size defaults to the total pod memory limit (or node allocatable if pod limits are unset).
+* If `sizeLimit` is specified and exceeds the total pod memory limit (e.g. 256Mi on a 200Mi pod), the mounted `tmpfs` volume capacity is capped to the total pod memory limit (200Mi). The Kubelet emits a `Warning VolumeSizeExceedsPodMemoryLimit` event during admission.
```
xing-yang, line 392: Do you need to update the OOM-kills section accordingly?

tallclair (commented):
Approved for sig-node
/assign @dchen1107 
For official approval.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Somewhat. Pods today (before this feature) already get ENOSPC on write when workloads attempt to surpass the `sizeLimit` for memory-backed emptyDir volumes. The section in this KEP that talks about this is intended to be more like background information / preexisting context.

So I don't think the eviction manager for memory volumes was ever effective in practice, since the limit was hard-enforced by the kernel. I put it behind the feature gate as a safety measure, but I think the failure mode in question does not occur under normal operations. 

That said, disabling eviction mgr for memory-backed volumes is still a change to default behavior, so I'll update this section accordingly.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: The feature is disabled entirely on windows.

For linux, versions >= 5.2 use the new fsconfig mount API (`fsconfig() failed: tmpfs: Too small a size for current use`), while older kernels (< 5.2) or older mount binaries fail with the legacy mount failure (`mount point not mounted or bad option` or `cannot remount tmpfs read-write, is write-protected`). In both cases, the kernel safely rejects the shrink below active memory without corrupting data, and Kubelet surfaces the mount error directly in PodResizeInProgress.

I added this information to the KEP.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/kep.yaml
natasha41575: Yeah all the metrics will have the `kubelet_` prefix. Fixed!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Added, thanks!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: > Aren't you disabling evictions in place of kernel OOM enforcement? 

Kernel OOM enforcement is already in place (before this feature), which means that evictions never occurred for memory-backed volumes in practice. We're disabling these evictions, but related to my other comment https://github.com/kubernetes/enhancements/pull/6320#discussion_r4057725630, I don't think this is a meaningful change

But you're right, this section is wrong, there's no reason for evictions to increase, took that out.

natasha41575: I thought this question was about whether this feature depends on other out-of-tree (i.e. non-k8s) services running in the cluster, rather than other FGs? I added NodeDeclaredFeatures here though

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: done, thanks!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: updated the oom-kills section with this information!

kannon92 (commented):
/approve
for PRR

natasha41575:
/hold

just to gate the merge

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -802,18 +842,20 @@ These goals will help you determine what you need to measure (SLIs) in the next
 question.
 -->
 
+- Resize requests should succeed (`kubelet_pod_resize_duration_milliseconds{"error=true"}` should be low)
```
xing-yang, line 845: `kubelet_pod_resize_duration_milliseconds{"error=true"}` -> 
`kubelet_pod_resize_duration_milliseconds{error="true"}`

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -238,58 +243,42 @@ While the controller can use `InPlacePodVerticalScaling` to scale up a container
...
     *   *Admission during Resize*: Upon admitting a resize update, the `AllocationManager` updates the Kubelet's local checkpoint with the newly acknowledged capacity.
     *   *Scope*: Maintained only for memory-backed `emptyDir` volumes.
-*   **Actual State**: Represented by the new `VolumeStatus.EmptyDir.SizeLimit`.
-    *   *Definition*: Represents the actual mounted filesystem capacity reported by the mount/remount call.
-    *   *Pod Creation*: Remains unpopulated (`nil`) until Kubelet completes the initial volume mount setup and receives capacity confirmation from the mounter.
-    *   *Pod Resize*: During an active resize operation, this field retains the *old* mounted size until the Kubelet successfully actuates the `tmpfs` remount, at which point it is updated to the new actual capacity.
-    *   *Default/Fallback*: Remains unpopulated (`nil`) for all non-memory-backed volumes.
-
-The new status field `Pod.Status.ContainerStatuses[].VolumeMounts[].VolumeStatus.EmptyDir.SizeLimit` is only populated for memory-backed volumes, and remains unpopulated for other types of volumes. The `VolumeStatus` Go struct is updated as follows:
```
xing-yang, line 267: This is still mentioned in the Rollback and Rollout sections.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -777,13 +817,13 @@ and operation of this feature.
...
-- [ ] Events
-  - Event Reason: 
-- [ ] API .status
-  - Condition name: 
-  - Other field: 
-- [ ] Other (treat as last resort)
-  - Details:
+- [X] Events
+  - Event Reason: `ResizeStarted` (emitted when kubelet accepts a resize request, along with the new requested size)
+  - Event Reason: `ResizeCompleted` (emitted when kubelet successfully resizes the volume)
```
xing-yang, line 822: What about ResizeError?

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
```diff
@@ -994,8 +1044,24 @@ For each of them, fill in the following information by copying the below templat
...
 -->
 
+- **Volume shrinkage below current in-use memory**:
+  - **Detection**: The `PodResizeInProgress` condition in pod status is set to `Status: True`, `Reason: Error` with a message from the mount failure (e.g., `tmpfs: Too small a size for current use.`), and a `Warning ResizeError` event is emitted on the pod.
+  - **Mitigations**: Running containers are unaffected (0 container restarts) and the volume remains mounted at its previous capacity. The workload can delete files to reduce usage, or the user/autoscaler can update the pod spec with a higher `sizeLimit` or cancel the resize.
+  - **Diagnostics**: Kubelet logs at `v=2` with `Resizing emptyDir volume` and mount error details.
+  - **Testing**: Covered by unit tests in `pkg/volume/emptydir` and node E2E tests verifying retry behavior and condition reporting on shrink failures.
+
+- **Resize requested on a node running cgroups v1**:
+  - **Detection**: The resize request is marked `Infeasible` in Pod status (`PodResizePending: Infeasible`), the metric `kubelet_pod_infeasible_resizes_total` is incremented, and a `Warning ResizeInfeasible` event is emitted on the Pod.
```
xing-yang, line 1054: Can you add `kubelet_pod_infeasible_resizes_total` to the metrics section?

natasha41575: added `kubelet_pod_infeasible_resizes_total` to all metrics sections, thanks!

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: SG, I added `ResizeError` and `ResizeInfeasible` to the list here for completeness

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: Updated this! Thanks for the catch.

keps/sig-node/6030-dynamic-resize-of-memory-backed-volumes/README.md
natasha41575: fixed, thanks!

xing-yang:
lgtm

dchen1107:
cc/ @andrewsykim etc. from OSS Ray community.

Issue #6264, opened 2026-08-05
KEP: Add configurable etcd object compression for kube-apiserver storage


FANHIDE:
### Enhancement Description

- One-line enhancement description (can be used as a release note): Storage object compression in kube-apiserver
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/pull/6321
- Discussion Link: https://docs.google.com/document/d/1x9RNaaysyO0gXHIr1y50QFbiL1x8OWnk2v3XnrdkT5Y/edit?tab=t.0#bookmark=id.y9s25v4d4dnr
- PRs by stage and milestone:
  - [ ] Alpha - v1.38
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6321
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update PR(s):

## Proposal

Introduce an Alpha, default-disabled `EtcdCompression` feature gate for kube-apiserver storage. When enabled, it compresses selected API objects before writing them to etcd and transparently decompresses them when reading.

## Motivation

Large API objects can consume substantial etcd disk capacity and network bandwidth. Compression can reduce this cost for operators that choose an explicitly configured resource scope.

## Initial scope

- Feature gate: `EtcdCompression`, Alpha, default disabled.
- kube-apiserver only.
- Explicit resource selection; no implicit compression of all resources.
- An encoded-value marker that permits an enabled API server to distinguish compressed and uncompressed existing values.
- A supported composition model with encryption-at-rest transformers.

## Required design decisions before implementation

- On-disk encoding format and framing.
- Upgrade, rollback, and control-plane version-skew behavior.
- Interaction order with encryption-at-rest and storage migration.
- Resource-selection API, validation, and observability.
- Unit, integration, and e2e coverage, plus compression performance characterization.

## Non-goals

- Changing etcd transport compression.
- Enabling compression by default.
- Automatically rewriting all existing objects.

## Proposed next step

Prepare a KEP under `keps/sig-api-machinery/` after SIG API Machinery feedback identifies maintainers and confirms the storage-format contract.

/area apiserver
/sig api-machinery

kannon92:
cc @jberkus

mengqiy:
FMU it won't necessarily require an etcd change.

>A supported composition model with encryption-at-rest transformers.

Compression needs to happen before encryption to achieve a better compression ratio.

[...]

PR #6321, opened 2026-09-02
KEP-6264: storage object compression 

mengqiy:
- initial draft for storage object compression KEP

- Issue link: https://github.com/kubernetes/enhancements/issues/6264

- Other comments:
  - the test section and PRR section will be added later after gathering initial feedback


mengqiy:
@jpbetz @deads2k @liggitt Looking for early feedback from sig apimachinery folks

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,950 @@
...
+
+CPU cost lands on both the write and read hot paths. Compression sits on every mutating request for a
+selected resource, and the worst read case is watch-cache initialisation, a full LIST decompressed
+sequentially on the request goroutine. KEP-2338 measured the analogous mistake: enabling gzip for all
+suitable API responses "caused a significant performance regression in both CPU usage (2x) and tail
+latency (2-5x) on the Kubernetes apiservers", which is why a 128 KB size floor was introduced
+([KEP-2338 README][kep-2338]). Mitigations here are level 1, a configurable floor defaulting to 1 KiB,
+pooled coder state, per-resource opt-in, and DEFLATE's asymmetry: decompression performs no match
+search, only Huffman decoding and byte copies, so on real Pods it runs 1.7-2.0x faster than
+compression. A watch-cache-initialisation benchmark is an alpha deliverable.
```
neolit123, line 384: the doc mentions benchmarks, but i personally think it would be worth to provide some benchmarks as part of the KEP (pre-alpha), to reveal the R/W overhead from this compression proposal with various algorithms.

neolit123:
> Error:      	"map[/home/prow/go/src/github.com/kubernetes/enhancements/keps/sig-api-machinery/6264-storage-object-compression/kep.yaml:[metadata file missing for KEP: /home/prow/go/src/github.com/kubernetes/enhancements/keps/sig-api-machinery/6264-storage-object-compression/kep.yaml]]" should have 0 item(s), but has 1

you'd have to include the accompanying kep.yaml file.

keps/sig-api-machinery/6264-storage-object-compression/README.md
mengqiy: Scale testing is WIP. I will share some result soon.

>various algorithms.

This KEP currently focuses deflate only due the dependency constraints. If sig apimachinery becomes open to additional dependencies, then we can add additional algorithms

hakuna-matatah:
pss

Teaming with @mengqiy  on this KEP.

Preliminary test results for pod sizes ranging from 6KB - 20KB, with compression, it shows better  results even for CPU/Mem usage on APIserver outside of etcd DB size savings against non-compressed pod specs and better APIServer SLOs.


Test was with pod # 150k created at 400/sec throughput. 

It is only applied to Pods resources, rest of resources didn't use the feature during the test.

hakuna-matatah:
Discussed in Sig-APIMachinery today  along with @mengqiy on this, next step is to run this benchmark test https://github.com/kubernetes/perf-tests/blob/master/clusterloader2/testing/write-throughput/config.yaml and post results here.

serathius:
My main worry is that adding decompression on watch read path will cause additional overhead on watch cache read path further pushing watch staleness. This is currently one of the top bottlenecks in K8s scalability, how fast we can propagate change depends on apiserver and informers ability to keep up with stream of changes from etcd. It's the apiserver that is the slow one, not etcd. I would prefer to introduce compression on etcd storage layer so the storage size limit is resolved without introducing overhead to watch consume pipeline. 

From my experiments on https://github.com/etcd-io/etcd/issues/22332 the object size itself is not a problem for write throughput etcd maintains write throughput even up to 100GB storage sizes, but it impact on compaction and defrag. Putting compression on apiserver just adds complexity and pushes the problem further, but doesn't resolve it.

With etcd being part of Kubernetes as SIG-etcd I would like to have us think holistically what is the best place to solve the problem.

mengqiy:
>My main worry is that adding decompression on watch read path will cause additional overhead on watch cache read path further pushing watch staleness. This is currently one of the top bottlenecks in K8s scalability, how fast we can propagate change depends on apiserver and informers ability to keep up with stream of changes from etcd.

For an object to propagate from etcd to apiserver watch cache, it needs to go through multiple phases:
etcd server send over wire (-) -> etcd client (apiserver) decode (-) -> decrypt (optional) (-) -> decompress (+) -> deserialize (same) -> add to watch cache
`-` means lower latency,  `+` means higher latency and `same` means same amount of work as before. Overhead of decompression should mostly be offset by savings from other steps.

There are 2 bottleneck patterns that can affect watch cache freshness:
1. Single thread processor. This KEP doesn't introduce any single thread processor.
2. Lock contention. This KEP doesn't introduce any new locks in watch cache. And there's no signal showing the lock contention becomes worse.

> I would prefer to introduce compression on etcd storage layer so the storage size limit is resolved without introducing overhead to watch consume pipeline.

A few downsides to do it in etcd:
1. Encrypted object is almost incompressible. And it's a relatively common practice that user encrypt almost all resources. To support compress-before-encrypt in etcd, etcd need to build compression and **re-build encryption feature** which has already been built in apiserver. 
2. We won't get the benefits of smaller object on wire between etcd and apiserver. The CPU/mem overhead on compression won't be compensated by saving from smaller obj on wire.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,950 @@
...
+
+Framing costs 5 bytes for `stored` and 6-8 for `deflate`, at worst 0.78% of a 1 KiB object and 0.003%
+of a 256 KiB one. Below roughly a kilobyte that overhead stops being negligible while the gain stops
+being worth having: a Lease at 485 bytes compresses only 1.42x, too little to repay the CPU spent on
+every read and write of it, and a genuinely incompressible object would simply grow. The floor keeps the
+overhead off those objects, since they are never framed.
+
+The floor is `minSize` on the matching configuration entry, 1 KiB by default, with a hard lower bound of
+64 bytes: framing already costs 8-12% at that size, and DEFLATE on inputs that small essentially always
+expands into the verbatim frame. The *ceiling* stays a compile-time constant, deliberately, since it is
```
gyuho, line 455: Can we clarify the size-limit contract here? Can compression make a write succeed that would previously have been rejected as too large, or do we enforce the limit against the original, uncompressed payload? Since etcd limits the whole request rather than just the value, how do we account for keys, framing, and encryption overhead?

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,950 @@
...
+would therefore be announced first, with the write path and the `None` algorithm kept for at least two
+releases and for whatever deprecation window the configuration type's stability level requires, and
+migration to uncompressed documented as a prerequisite before the writer goes.
+
+This KEP deprecates nothing that already exists. `--storage-compression-config` is new and supersedes
+nothing: not `--storage-media-type`, not response compression, not any etcd-side setting. The gate
+itself deprecates ordinarily, with the flag and configuration type outliving it, as other configuration
+files have outlived their gates.
+
+### Upgrade / Downgrade Strategy
```
gyuho, line 807: Can we explicitly align the migration contract with encryption-provider changes which already require rewriting all affected existing objects before removing old read support? Also, can we specify and test that compression staleness is combined with the encryption provider’s staleness, so compression cannot bypass a required re-encryption?

xigang:
Similar to how RocksDB performs compression at the storage layer, I also suggest implementing data compression in etcd. This would not only reduce the amount of data stored and transferred, but could also improve Raft throughput.

Implementing compression in kube-apiserver would introduce additional overhead and increase the latency of synchronizing data into the watch cache.

jpbetz:
> Similar to how RocksDB performs compression at the storage layer, I also suggest implementing data compression in etcd. This would not only reduce the amount of data stored and transferred, but could also improve Raft throughput.

How would you address the support for KMS? That appears to be a blocker for this approach.

tjungblu:
> And it's a relatively common practice that user encrypt almost all resources. 

Is that really the case? For us a large majority does not enable encryption at all, or only for Secret/Configmaps. Which may not be compressible anyway because the tokens or certificates inside them are not compressible either.

We've seen some pretty good results with zstd in bbolt. It makes sense, you can compress across much larger prefix ranges at the same time, paired with larger page sizes I think this would be a huge win (for etcd) regardless of KMS. 

Maybe the question is: why not both?

jpbetz:
> Maybe the question is: why not both?

Let's focus this KEP on the SIG API Machinery half of the question.

I'll leave it to SIG etcd to decide the etcd side.

keps/sig-api-machinery/6264-storage-object-compression/README.md
mengqiy: >Can compression make a write succeed that would previously have been rejected as too large

No. This is described in bullet point 1 in non-goal and some later session. 

>do we enforce the limit against the original, uncompressed payload?

Yes. For a write that we reject before due to too large, we still reject it even with compression.

xigang:
> > Similar to how RocksDB performs compression at the storage layer, I also suggest implementing data compression in etcd. This would not only reduce the amount of data stored and transferred, but could also improve Raft throughput.
> 
> How would you address the support for KMS? That appears to be a blocker for this approach.

Yes, KMS is a blocker. Since encryption happens in the kube-apiserver transformer chain, etcd only ever sees ciphertext for KMS-encrypted resources, and ciphertext doesn't compress. Compressing in etcd would require moving envelope encryption (KMS) into etcd, which changes the trust boundary.

jpbetz:
Hi @mengqiy Would you please fill out all needed PRR questions asap so I can do a 1st pass PRR review?  Thanks!

mengqiy:
@jpbetz Just pushed the PRR parts. I started to fill out before you asked. PTAL

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+compressibility. A per-resource ratio would be a genuinely finer content-dependent signal, and so none
+is exported.
+
+### Test Plan
+
+[ ] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+TODO
```
jpbetz: Required for alpha. Let's be detailed here. The test strategy will be hugely important to ensuring correctness.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+
+###### What are other known failure modes?
+
+- A malformed frame: stored bytes look framed but do not decode.
+  - Detection: `apiserver_storage_compression_format_errors_total` non-zero, and the whole LIST fails.
+  - Mitigations: none. What it means depends on the provider beneath. Under an authenticated encryption
+    provider like KMS v2 the ciphertext was verified before reaching this layer, so the fault is a bug
+    here and the fix is a code fix. Under `aescbc` or with no encryption the stored bytes are genuinely
+    damaged, and the only recovery is to restore from backup. Either way the object cannot be cleared
+    with unsafe deletion, which refuses this class of error.
```
jpbetz, line 1354: Let's makre sure we're tracking and testing the requirement that this is compatible with KEP-3926 ("unsafe delete"). It might require updating a couple sections of this KEP to thread that in.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+for selected resources and inflation on the read path, plus memory for pooled compressors and for bounded
+concurrent inflation; see [Bounding decompression](#bounding-decompression). Against that, smaller stored
+values mean fewer bytes over the wire to and from etcd, which saves both CPU and the buffers carrying
+them. Early measurements suggest that saving can offset the compression overhead, which the beta scale
+test is there to confirm.
+
+Network traffic to etcd decreases with the wire size.
+
+Admission control needs correcting separately, since it charges a LIST by stored bytes while the memory it
+occupies is the plaintext; see [API Priority and Fairness](#api-priority-and-fairness-and-size-accounting).
```
jpbetz: Is the estimator correction part of the alpha implementation? Lets add it to the alpha criteria and the test plan if so, otherwise put it as beta and leave TODOs.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+No change in counts, and no increase in size for almost every object: stored size decreases, with ratios
+in [Measured on production data](#measured-on-production-data). A small portion grows. A selected value
+above the configured floor that does not compress is stored in a verbatim frame, so it grows by five
+bytes.
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+With no configuration file, nothing changes. For a selected resource each mutating request adds one
+compression pass and each read of a framed value one inflation, which is cheaper. The worst read case is
+watch-cache initialisation, a full LIST inflated object by object while the request waits.
```
jpbetz: Let's not yet claim "nothing changes" without data. The watch-cache-initialization benchmark can inform this. I'm okay with a TODO here for alpha.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+which is what makes enable and disable terminate, and a value framed under one configuration reading
+back under any other. The restart-driven enable, disable and re-enable cycle needs an integration test.
+
+### Rollout, Upgrade and Rollback Planning
+
+###### How can a rollout or rollback fail? Can it impact already running workloads?
+
+No workload is affected. Only how kube-apiserver stores bytes changes, and etcd sees an opaque value
+either way. No interleaving can corrupt data, because a read is decided by the stored bytes rather than
+by the reading apiserver's configuration, so a mid-rollout fleet needs no coordination.
```
jpbetz, line 1168: Let's focus more on the migration use case here. It's the main operation impacted and it is impacted significantly.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+
+No. Without a configuration file stored bytes are byte-identical. Selecting a resource changes only its
+at-rest bytes, plus a one-time write amplification while existing objects migrate. Nothing visible
+through the API changes. Raw-etcd tooling is the exception: it can no longer read those values.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+The write path, yes and immediately: remove the resource's entry or give it the `None` algorithm. A
+rollback expressed through the gate must drop the flag too, since the flag without the gate fails
+startup. The read path, no, and that asymmetry is what makes rollback safe: it can never strand data.
```
jpbetz: Flipping the gate off is the standard rollback step, and here it fails apiserver startup unless the flag is removed in the same change. Please put that in the release note and flag help next to the downgrade warning. Alternatively, ignore the file with a warning when the gate is off; the case the startup failure guards against (a staged file activating at the beta default flip) only forfeits downgrade to a release that can already read frames.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+  apiserver cannot inflate. Roll the lagging apiserver forward.
+- `apiserver_storage_compression_format_errors_total` non-zero: investigate. Rolling back makes this
+  worse, since an older apiserver has fewer ways to read the same bytes.
+
+No counter can confirm a downgrade is safe. They are driven by reads, so a quiet counter means nothing
+has looked, not that no frames remain.
+
+###### Were upgrade and rollback tested? Was the upgrade->downgrade->upgrade path tested?
+
+No. The upstream implementation PR is not open, so upgrade/downgrade has been exercised end to end. The two
```
jpbetz:
```suggestion
No. The upstream implementation PR is not open, so upgrade/downgrade has not been exercised end to end. The two
```

? (I'm assuming this is what was intended)

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1286 @@
...
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `StorageObjectCompression`
+  - Components depending on the feature gate: kube-apiserver
+- [x] Other
+  - Describe the mechanism: both boxes apply, because neither suffices alone. Compressing needs the gate
+    and `--storage-compression-config` naming the resource with an algorithm other than `None`. The gate
+    alone is inert; the flag alone fails startup.
+  - Will enabling / disabling the feature require downtime of the control plane? No, but each apiserver
+    restarts, since the configuration is read once per process. On HA that is a rolling restart.
+  - Will enabling / disabling the feature require downtime or reprovisioning of a node? No.
```
jpbetz: Typicaly we don't introduce both a feature gate and a flag if one is sufficient, since they are both set as kube-apiserver flags anyway (and the only the cluster administrator can set them).  Do we need both?  If not, let's drop the feature gate and use the flag.

We have had quite a few features in the past that have progressed using only a flag in the same way.

mengqiy:
Added the test plan parts. What we can do in the upgrade/downgrade tests today needs to be understood if any.
I will try to address the remaining comment later today

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1417 @@
...
+
+###### What are other known failure modes?
+
+- A malformed frame: stored bytes look framed but do not decode.
+  - Detection: `apiserver_storage_compression_format_errors_total` non-zero, and the whole LIST fails.
+  - Mitigations: none. What it means depends on the provider beneath. Under an authenticated encryption
+    provider like KMS v2 the ciphertext was verified before reaching this layer, so the fault is a bug
+    here and the fix is a code fix. Under `aescbc` or with no encryption the stored bytes are genuinely
+    damaged, and the only recovery is to restore from backup. Either way the object cannot be cleared
+    with unsafe deletion, which refuses this class of error.
```
jpbetz: This removes the remedy that exists today for a genuinely corrupted value. Right now any decode or transform failure is classified as a corrupt object and is eligible for unsafe deletion (`corruptObjErrorInterpretingDecoder` treats every decode error as undecodable). With this KEP, a bit flip in a framed value under `aescbc` or with no encryption becomes a malformed-frame error that is excluded from unsafe deletion, so one damaged object makes the whole resource unlistable and the only recovery is an etcd restore. That is a worse operational position than today for exactly the providers that cannot detect the damage.

I think we need one of two things before alpha: decide open question 1 in favour of a checksum, so a mismatch is classified as real corruption and stays deletable; or keep malformed frames eligible for unsafe deletion when the provider beneath does not authenticate, since the operator is already opting into cluster-breaking potential by using that flag. Either way this answer should not end at "restore from backup" for a single object.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1417 @@
...
+
+Targeted at v1.38.
+
+- [ ] The flag is unset by default; writes are reachable only with a configuration file and are disabled
+  by removing it; a malformed or contradictory configuration fails startup rather than silently doing
+  nothing.
+- [ ] Open question 1, whether a frame carries a checksum, is decided. It must be settled before alpha
+  ships rather than at beta, because adding integrity afterwards requires a new algorithm id and a
+  release of read-before-write skew. If the answer is a checksum, a mismatch needs a classification
+  distinct from a format error, so that it stays eligible for the unsafe-deletion flow of KEP-3926.
```
jpbetz: Let's settle open question 1 in this PR rather than "before alpha ships". An implementable KEP should not leave the at-rest format open, and the failure-mode and rollback answers depend on it; see my comment on the malformed-frame failure mode under Troubleshooting.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1417 @@
...
+  other sentinel.
+- A frame written by this release, pinned as a literal at a fixed etcd path, so that future releases
+  keep proving it still decodes. This is the regression anchor for the skew rule.
+- A malformed configuration file, and a `minSize` outside its legal range, each fail startup.
+
+One more belongs in `test/integration/storageversionmigrator` and is a **beta** item rather than an alpha
+one, since it is what exercises the documented pre-downgrade procedure: deselect a resource, restart,
+migrate, then assert that no stored value under that resource leads with the frame discriminator and
+that the number of values scanned equals the number created, so an empty scan cannot pass. Run it in
+both directions, since the staleness rule is symmetric.
```
jpbetz: Suggest making this an alpha item. Downgrade below v1.38 is the one outage-class transition in this KEP, and a completed migration after deselection is its only mitigation, so it should be proven in the same release that first writes frames. `test/integration/storageversionmigrator` already exists, so the cost is one test. With it in place, the "Were upgrade and rollback tested?" answer can point at real coverage instead of "No".

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1417 @@
...
+##### e2e tests
+
+None. The feature is turned on by a kube-apiserver flag, which an e2e test cannot set against a running
+cluster, and it adds no API surface a client can observe: objects round-trip unchanged whether or not
+they are compressed. Everything an e2e test could assert is either asserted at the integration level
+with access to the raw etcd bytes, or invisible by design.
+
+##### Scale test
+
+Four runs, all required for alpha.
```
jpbetz: This can be beta.

keps/sig-api-machinery/6264-storage-object-compression/kep.yaml
```diff
@@ -0,0 +1,47 @@
+title: Storage Object Compression
+kep-number: 6264
+authors:
+  - "@hakuna-matatah"
+  - "@mengqiy"
+owning-sig: sig-api-machinery
+participating-sigs:
+  - sig-scalability
+status: provisional
```
jpbetz:
```suggestion
status: implementable
```

Need to fix this by EOD today (AoE) to mee KEP readiness criteria.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1417 @@
...
+
+#### Alpha
+
+Targeted at v1.38.
+
+- [ ] The flag is unset by default; writes are reachable only with a configuration file and are disabled
+  by removing it; a malformed or contradictory configuration fails startup rather than silently doing
+  nothing.
+- [ ] Open question 1, whether a frame carries a checksum, is decided. It must be settled before alpha
+  ships rather than at beta, because adding integrity afterwards requires a new algorithm id and a
```
jpbetz, line 899: +1. Let's sort this out by KEP freeze next week.

keps/sig-api-machinery/6264-storage-object-compression/README.md
mengqiy: Added

mengqiy:
I believe everything required for PRR readiness freeze has been done: https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#kep-readiness-deadline
@jpbetz please LMK if I misses anything.

keps/sig-api-machinery/6264-storage-object-compression/kep.yaml
mengqiy: Updated

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1434 @@
...
+etcd-delegated LISTs of that resource roughly in proportion to its ratio. The process-wide inflation cap
+still bounds total memory, so the consequence is unfair admission rather than unbounded growth.
+
+### Error classification and version skew
+
+Two read failures are deliberately not treated as corruption: an unsupported algorithm, and a malformed
+frame. The first was written by a newer apiserver and is perfectly readable by one; the second is more
+likely a bug in this layer than damaged bytes. Reporting either as a `CorruptObjError` would authorise
+deletion through `DeleteOptions.ignoreStoreReadErrorWithClusterBreakingPotential` and let an operator
+destroy an undamaged object, so neither is classified that way and the read fails outright instead.
```
jpbetz, line 672: This is a blocker for me. I think we can address it, but I want to make sure we fully handle it.

A plaintext CRC or gzip suffix is only checked after decompressing the data. This means a bit flip due to a cosmic ray, or whatever, can result in a malformed frame from a compression perspective. I need the guarantee that we can recover from this sort of corruption via the API (KEP-3926, basically).

I'm flexible on how we do this. But I'd like a commitment to test this case and ensure it's fully supported before we turn this feature on-by-defautl.

If it means a CRC check before inflation I'd like to know early so we understand the performance implications. I don't want have to merge a performance regression to ensure we handle this right after this goes on-by-default.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1434 @@
...
+  their stored size, but unencrypted clusters get the same benefit from the same code.
+- Per-resource opt-in from the first release, so an operator can compress ConfigMaps and custom
+  resources without touching Secrets.
+- Add no new third-party dependency.
+
+### Non-Goals
+
+- Using compression to exceed the object size limit. This is enforced rather than merely declared:
+  plaintext above the per-value ceiling is never framed, so such an object fails on write exactly as
+  it does today.
```
jpbetz, line 196: etcd's limit applies to the whole request (key, val, txn data, ...).

So I'm still concerned that there is an boundary case where were objects are storable only when compressed, and those can't be rewritten if compression is disabled.

Kubernetes is under heavy enough industry use where we've gotten issue reports of similar boundary cases.

Given that, let's add this case to the test plan and make sure we test it.

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1434 @@
...
+reachable by default, and every release in which it is reachable at all must state, in the release note
+and the flag help, that opting in forfeits downgrade to the previous minor for the selected resources
+until a migration has rewritten them uncompressed. At v1.38 that warning is doing real work, because
+nothing can teach v1.37 to inflate a frame.
+
+Any new algorithm must ship read support one release before any release may write it.
+
+#### Alpha
+
+Targeted at v1.38.
```
jpbetz, line 893: Things I'd like to see in the alpha criteria:
- A check at startup for a non-None policy when the emulated or min-compatibility version is below 1.38. We need this to know we support rollback within the compatibility window.
- @gyuho's above mentioned `stale = delegateStale || compressionStale` case (key-rotation test?)
- The SVM integration test (needs to be in place before on-by-default, I vote for alpha on this)
- A config-hash metric like the encryption config to ensure all replicas are in sync before migrating.
- @serathius's previously mentioned CPU numbers on Go 1.27

keps/sig-api-machinery/6264-storage-object-compression/README.md
```diff
@@ -0,0 +1,1434 @@
...
+  decides whether that residual is acceptable at alpha, whether per-resource opt-in and the absence of
+  any ratio or byte-count metric are jointly sufficient, and whether the startup warning for an
+  explicitly named `secrets` is adequate. Sign-off is recorded in Implementation History.
+
+Open question 2, a per-resource compression level, is a beta item. Open question 3, the discriminator
+invariant belonging in apimachinery, is carried by beta's closing of known gaps.
+
+#### Beta
+
+Targeted at v1.39.
```
jpbetz, line 920: For beta, I'd like to see:
- A carefully selected minSize, ideally based on a high scale run.
- A way to stop compressing new writes but continue to read compressed data. I don't want the "back to safety" lever to automatically trigger a tons of writes, potentially making things worse.
- A CI job with the feature enabled? We need to test this holistically. I'm open to other suggestions.

jpbetz:
PRR looks good to me. I've left some KEP feedback with my remaining technical concerns on the implementation and graduation criteria.

@deads2k @sttts do either of you have concerns?

PR #6323, opened 2026-09-03
KEP-6318: Allow in-place updates to container probes

HirazawaUi:
- One-line PR description: Add the initial KEP for allowing liveness, readiness, and startup probes on regular containers and restartable init containers to be updated in place through normal Pod UPDATE/PATCH operations.

- Issue link: https://github.com/kubernetes/enhancements/issues/6318

- Other comments:


keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,457 @@
...
+
+## Proposal
+
+When the `MutableContainerProbes` feature gate is enabled, the Pod API permits changes to `livenessProbe`, `readinessProbe`, and `startupProbe` on regular containers. The same fields may be changed on init containers whose `restartPolicy` is `Always`. All other Pod mutability rules continue to apply.
+
+Probe changes use the normal Pod `UPDATE` and `PATCH` endpoints. A request that also changes an immutable field is rejected as a whole. Successful persistence of the update does not mean that the kubelet has already run the new probe; propagation follows the usual PodSpec convergence model.
+
+The kubelet reconciles the probes in the latest PodSpec with its probe workers. It creates and removes workers as needed and updates existing workers without publishing an artificial probe failure. Results produced by an obsolete probe configuration or a previous container instance are not allowed to affect the current container.
+
+### User Stories (Optional)
```
toVersus, line 106: As I read it, the reason the entire probe configuration is made mutable in place is to avoid adding new API surface. I understand that motivation, but I'm not convinced the use case for updating the entire probe configuration is strong enough to justify it. As the `Notes/Constraints/Caveats` section itself notes, changing an individual Pod's spec is only temporary, and the durable fix is to update the template on the parent resource.

- For Story 2, if you want to tune readiness/liveness under production load, the common approach is to roll out with more relaxed thresholds and tighten them from there, or to validate them up front with a stress test. I don't think this justifies making probes mutable.
- For Story 1, the use case only requires turning a probe on and off, so it does not seem to need in-place updates of the entire probe configuration.

keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,457 @@
...
+
+#### Story 2 (Optional)
+
+As a workload owner, I want to tune readiness and startup probe periods, timeouts, and thresholds after observing the application under real load so that I can correct the health-check behavior without causing an otherwise unnecessary rollout.
+
+### Notes/Constraints/Caveats (Optional)
+
+- An update to an individual Pod is not copied back to the owning workload. A replacement Pod will use the controller's Pod template, so users that want the change to persist across replacement must update that template as well.
+- Adding or changing a liveness or startup probe can subsequently restart a container when the new probe fails. The update itself does not restart it.
+- Existing `update` or `patch` permission on Pods is sufficient. This is worth noting for `exec` probes because an update may change a command that the kubelet runs in a container.
```
toVersus, line 120: I think this widens the attack surface. If a controller or user that holds update/patch permission on Pods is compromised, an attacker can rewrite an exec probe to echo environment variables or local file and then read the secrets straight out of the Warning Event that kubelet records when a probe fails. If we do want the full probe configuration to be mutable, I think it would be better to have a dedicated subresource for it.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I don’t think the fact that a per-Pod change is temporary weakens the use case. The same is true for [In-Place Pod Resize](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources): the owner’s template is not automatically updated, but changing a running Pod is still valuable when recreating it would be disruptive. Workload-level propagation is a separate concern that can build on the Pod-level primitive.

Probe settings, like resource settings, can turn out to be wrong only under real production conditions. If only a timeout or threshold needs correction, restarting a stateful or low-replica workload is unnecessary. Story 1 also includes relaxing a probe, not only disabling it, which can be safer than removing liveness checking entirely.

The template should still be updated when the change needs to persist, but that does not remove the value of updating the current Pod in place.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: A dedicated subresource is definitely an option, and I considered it initially. It would provide finer-grained RBAC, but it would also add another API path for configuration already stored in each container’s `PodSpec`. That is why I listed it as an alternative.

I agree that the security impact needs careful consideration. However, `update` or `patch` access to Pods is already a powerful permission. For example, it allows changing a container image, which causes kubelet to restart the container with new code that can access the Pod’s credentials and volumes. Mutable exec probes expand that permission, but they do not create an entirely new security boundary.

keps/sig-node/6318-mutable-container-probes/README.md
toVersus: > The same is true for In-Place Pod Resize

In-place Pod resize has strong use cases, such as improving node resource efficiency and adjusting resource allocation at the Pod level via a higher-level controller (e.g., Vertical Pod Autoscaler). That said, the use case described here looks more like a convenience feature for debugging or tuning health checks. I'm wondering whether it's worth complicating the kubelet's probe handling for this.

> If only a timeout or threshold needs correction, restarting a stateful or low-replica workload is unnecessary.

Yeah. This might be a use case that seems plausible is preventing a stateful workload's (e.g., a critical database's) livenessProbe from failing and causing repeated restarts due to a temporary spike in node load. In that case, it should be fine if the temporary probe change gets reverted when the Pod is moved to another node via a drain, for example. But, since the role of livenessProbe is to address problems that can only be fixed by restarting the container, such as a deadlock, I also think that for a temporary performance degradation, taking the Pod out of service via readinessProbe would be the more correct approach. Anyway, it might be worth making the content of story 2 a bit more concrete (e.g., stateful workloads or jobs that can't tolerate interruption).

> Story 1 also includes relaxing a probe, not only disabling it, which can be safer than removing liveness checking entirely.

Rather than making the full probe configuration mutable, it might also be possible to limit it to threshold, timeout, and period seconds. This would remove any security concerns, and if the goal is to temporarily disable the probe, increasing threshold and period seconds should be enough.

keps/sig-node/6318-mutable-container-probes/README.md
toVersus: Regarding image, that's certainly true. I'd like to hear what others think about whether the existing precedent also justifies allowing this for probes.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I've clarified the User Stories to make the scenarios more precise.  Hopefully this addresses the wording concerns so we can move the design discussion forward.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I would also like to get broader and more accurate suggestions from others. If there is a consensus, I don't mind moving it out of the _Alternatives_ section and making it the primary solution for this KEP.

keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,459 @@
...
+- `spec.containers[*].startupProbe`
+- the same three fields on `spec.initContainers[*]` when the init container has `restartPolicy: Always`
+
+The new probe value is fully validated before PodSpec immutability is checked. Container count, names, ordering, and all fields outside the allowed probes remain subject to the current update rules. Updates that change probes on Pods with a `deletionTimestamp`, or on Pods in the `Succeeded` or `Failed` phase, are rejected.
+
+Static and mirror Pods remain managed from their source manifests. Users must change that source rather than patching the mirror Pod through the API.
+
+### Kubelet Behavior
+
+The probe manager will reconcile the desired workers every time the kubelet processes an updated Pod. A worker is still identified by Pod UID, container name, and probe type. Reconciliation adds missing workers, updates retained workers, and stops workers for probes that were removed.
```
toVersus, line 151: When kubelet observes a Pod update event, my understanding is that it enqueues a `SyncPodUpdate`, which eventually calls `SyncPod`. That path currently reaches `kl.probeManager.AddPod`, the same method used when a Pod is first added or when kubelet rebuilds probe workers after restart.
https://github.com/kubernetes/kubernetes/blob/d5cbb79be9160faa3a7d7323e3294c3781c16ae7/pkg/kubelet/kubelet.go#L2252-L2253

For this KEP, which implementation direction are you considering?

1. Change the current `kl.probeManager.AddPod` call into a more general reconciliation method, such as `kl.probeManager.SyncPod`, so the probe manager reconciles workers from the latest PodSpec for both create and update paths.
2. Keep separate probe-manager entry points, such as `AddPod` and `UpdatePod`, and call them based on `kubetypes.SyncPodType`.

keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,459 @@
...
+
+Workers use an immutable snapshot of probe configuration for each attempt. The manager invalidates attempts when their handler or timeout is replaced, when the worker is removed, or when the container ID changes. A late result from an invalid attempt is ignored even if cancellation did not stop the underlying probe in time.
+
+The behavior of individual changes is:
+
+| Change | Consecutive result count | Published status | Next execution |
```
toVersus, line 157: I wonder if the per-field worker state preservation rules are more complex than this feature needs, especially for Alpha.

For the main user stories, probe updates look like relatively low-frequency operational actions, such as correcting a bad probe, relaxing it during an incident, or removing/restoring it. In those cases, would it be simpler to treat any change to an existing probe config as a reset of the worker's internal state? Concretely, kubelet could preserve the last published probe result for the current container instance, but reset worker-local state on any probe config change: invalidate in-flight attempts, reset the consecutive success/failure count, and recalculate the next execution from the new config.

This would avoid the implementation complexity of a per-field state transition matrix, and it seems easier for users to reason about.

keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,459 @@
...
+
+Pod update audit records show changes to probe fields. The stored PodSpec shows the current configuration, although it does not by itself show whether that configuration was supplied at creation or by an update. Audit records can only provide this detail when the cluster's audit policy captures request bodies. No new per-workload metric is introduced.
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [x] API `.status`
+  - Condition name: `Ready` and `ContainersReady` for readiness changes
+  - Other field:
+- [x] Events
+  - Event Reason: existing probe failure and container restart events
```
toVersus, line 348: Beta would be fine, but what do you think about adding a kubelet event when it applies a probe addition, update, or removal? Audit logs can show that the PodSpec was updated, but a Pod event from kubelet would make it easier to understand what happened, especially when multiple users are working on the same Pod incident.

keps/sig-node/6318-mutable-container-probes/README.md
```diff
@@ -0,0 +1,459 @@
...
+
+The new probe value is fully validated before PodSpec immutability is checked. Container count, names, ordering, and all fields outside the allowed probes remain subject to the current update rules. Updates that change probes on Pods with a `deletionTimestamp`, or on Pods in the `Succeeded` or `Failed` phase, are rejected.
+
+Static and mirror Pods remain managed from their source manifests. Users must change that source rather than patching the mirror Pod through the API.
+
+### Kubelet Behavior
+
+The probe manager will reconcile the desired workers every time the kubelet processes an updated Pod. A worker is still identified by Pod UID, container name, and probe type. Reconciliation adds missing workers, updates retained workers, and stops workers for probes that were removed.
+
+Workers use an immutable snapshot of probe configuration for each attempt. The manager invalidates attempts when their handler or timeout is replaced, when the worker is removed, or when the container ID changes. A late result from an invalid attempt is ignored even if cancellation did not stop the underlying probe in time.
```
toVersus, line 153: > A late result from an invalid attempt is ignored even if cancellation did not stop the underlying probe in time.

I want to make sure I understand this part correctly. Today, `worker.stop()` only signals `stopCh`, so a probe that is already running inside `doProbe` may still complete after the probe is removed or its handler/timeout is changed.

Is the intent to leave that behavior mostly unchanged, and handle this by associating each in-flight probe attempt with the current probe configuration generation, then discarding its result if that generation has changed by the time the probe returns?

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I agree that simplifying the rules has value, but I don’t think every probe configuration change should reset the consecutive count. I would also prefer to keep the intended behavior consistent from Alpha to Beta.

For example, suppose a liveness probe has `failureThreshold=3` and has already recorded two consecutive failures. If the operator changes only `terminationGracePeriodSeconds`, resetting the count would require three new failures instead of one before reaching the threshold. That field controls how long the container has to shut down after a probe-triggered termination; it does not change whether a probe succeeds or fails.

Resetting the count would therefore make a shutdown-grace adjustment also postpone failure detection, despite the handler, timeout, and thresholds remaining unchanged. Even a single update has this effect, so low update frequency does not avoid the problem. This is why I don’t think we should apply a blanket reset to every configuration change.

WDYT?

toVersus: On `terminationGracePeriodSeconds`, you are right. It is not used to decide whether a probe succeeds or fails, so there is no reason for it to reset the count.

My concern was implementation complexity. I had assumed that if the consecutive result count could be reset, we could simply recreate the probe worker. Reading https://github.com/kubernetes/kubernetes/pull/142096, though, recreating the worker is not an option once we need to preserve the already published status.

So the worker has to be updated in place regardless of whether the count is reset. Given that we are already taking on that complexity, the additional cost of having per field rules looks fairly small. What we should be watching more closely is whether the new probe worker implementation introduces any regressions.

HirazawaUi: In addition to the test cases we added, I will also add a test job for this feature during the alpha stage, similar to `pull-kubernetes-e2e-kind`, which enables the `MutableContainerProbes` feature gate to ensure we don't break any existing behavior.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: Yes, that’s the idea. Each in-flight attempt is associated with the current execution epoch and container ID. When the probe is removed or its handler or timeout changes, kubelet attempts to cancel the probe and advances the epoch.

 If the underlying operation still completes, its result no longer matches the current epoch and is discarded before it can update the cached result, Pod status, or trigger a container restart. So a late result from the old probe configuration has no effect on the current container.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I’m considering the first direction: using a single reconciliation path for initial creation, Pod updates, and rebuilding state after a kubelet restart. I’d prefer to keep method names and call-path details out of the KEP, since those may change during implementation and make the KEP stale. The KEP should describe the required reconciliation behavior rather than the exact code structure.

keps/sig-node/6318-mutable-container-probes/README.md
HirazawaUi: I agree that Pod events are not useful for determining whether this feature is in use across workloads. Audit logs are the better source for that, since events are best-effort and short-lived. We can still emit a kubelet event when a probe is added, updated, or removed

kannon92:
As PRR approver I would really like to see some sig reviews on this PR before I pull this in for PRR cycle.

Once sigs take a look I'm happy to give a review for PRR.

kannon92:
/assign @SergeyKanzhelev @tallclair 

Based on sig-node 1.38 plan, you are taking on reviews from sig-node side.

PR #6324, opened 2026-09-03
KEP-5517: DRA Node Allocatable Resources beta

pravk03:
- One-line PR description: Promote DRA Node Allocatable Resources update to beta

- Issue link: https://github.com/kubernetes/enhancements/issues/5517

- Other comments:


pravk03:
/assign @johnbelamaric @liggitt @pohly
DRA API, PRR 

/assign @dom4ha
SIG Scheduling

/assign @tallclair @ffromani 
SIG Node

pravk03:
cc @alaypatel07

pravk03:
cc @kad

liggitt:
> /assign @johnbelamaric @liggitt @pohly DRA API, PRR

is this just updating the KEP with the API changes we made last release for beta? if so, I'm happy to just ack this

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -474,6 +483,18 @@ type NodeAllocatableOverhead struct {
...
+	// regardless of how many pods, containers, or claims reference or consume from it.
+	// It models host-level daemon processes that serve all workloads using the device
+	// on the node (such as a shared filesystem mounter process or a GPU MPS control
+	// daemon) and run on the host outside individual pod cgroups.
+	// Unlike PerPod and PerContainer, PerDevice overhead is an infrastructure
+	// reservation: it is not recorded in pod status and is not added to any pod or
+	// container cgroup.
+	// +optional
+	// +k8s:optional
+	PerDevice *resource.Quantity `json:"perDevice,omitempty" protobuf:"bytes,3,opt,name=perDevice"`
```
liggitt: This field is down within a singular device declaration, so it's a touch odd to call it perDevice (I don't think the `per` prefix makes sense for an overhead field that is used either 0 or 1 times)

A name that makes it clearer the overhead is tied to device allocation would be good (`whenAllocated`? `systemReserved`? not sure ... names are hard... maybe something connected to the system-reserved flag kubelet has to make it clear this is outside kubernetes' view)

Also make it clear in the docs this should not be used to represent overhead which will be consumed by a pod with its own requests, or stuff will be double-counted. It's *really* unfortunate that this couldn't be used to reserve overhead which would then be guaranteed to be given to a gang-scheduled infrastructure pod that uses that overhead. If the consumers for this overhead are likely to be pods, I'm not sure this is sufficient. If the consumers for this are likely to be outside kubernetes, this may be enough.

pravk03: Thanks. I like the comparison with [system-reserved](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#node-allocatable) as it serves a similar purpose, but is dynamic based on the device allocation state. I have changed it to `overhead.systemReserved` for now. 

The use case is that this overhead is consumed by a daemon process that the driver spins up to manage the device (sharing, mounting, etc.) when allocated. I have updated the documentation to make this clearer.

pravk03: Discussed this with @tallclair, @johnbelamaric, @mortent.
I am splitting this reservation use case into a separate enhancement issue - https://github.com/kubernetes/enhancements/issues/6352 (KEP draft - https://github.com/kubernetes/enhancements/pull/6353). There may be more options on how we account for this, and we can continue that discussion there. We can re-evaluate later if we want to fold it back into this KEP.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1108,6 +1116,17 @@ No new scheduler framework API is required for this. The `NodeAllocatableDRAClai
...
+#### DRA Admin Access
+
+Admin access (`adminAccess: true` on a claim request) is a privileged mode for monitoring and diagnostics, only allowed in
+namespaces labeled `resource.kubernetes.io/admin-access: "true"`. In the scheduler, an admin access allocation does not consume
+the device: the allocator can hand out a device that is already allocated to a workload, and the admin allocation does not block
+later allocations. Each such device is marked with `adminAccess: true` in the allocation result.
+
+For node allocatable resources, the workload's own claim already accounts for the device's resources, so charging the admin
+claim again would double count the node. Admin access results therefore contribute no footprint, get no entry in
+`pod.status.nodeAllocatableResourceClaimStatuses`, and do not block or get blocked by the mapped-claim sharing rule.
```
ffromani, line 1128: this fully makes sense but it made me realize how the DRAAdminAccess maps to the drivers from the other side of the API, which manage node allocatable resource. Also not sure this is the best place to discuss, as this item sits in the intersection of this KEP, the DRAAdminAccess KEP and the driver proper, but still:

I wonder how this maps to a driver handling core resources such CPU and memory. I can envision how a DRA driver can grant "admin access" to memory resources - enabling a RO view is probably the initial implementation, but what about CPUs?
How "admin access" to an exclusively allocated CPU device would look like? Like, can the claim use that CPU (but then is not exclusive anymore!)?

But even before that: we see this purely as a driver concern or can this have implications in the API? I can't think of obvious implication but I felt worthy raising the point.

pohly: In previous discussions we concluded that a driver may reject to grant admin access during NodePrepareResources if it cannot support that mode. The driver documentation then should mention that and admins shouldn't ask for admin access to devices of that driver.

It would be nicer if Kubernetes could detect misuse of the feature earlier, but we don't have the API to expose whether a driver supports the feature.

pravk03: Thanks for raising this point. Agree that this is mostly on the driver to handle. 
For the CPU driver, I filed https://github.com/kubernetes-sigs/dra-driver-cpu/issues/319 to handle it cleanly.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1885,62 +1964,102 @@ without needing to persist DRA allocations in Kubelet's local disk checkpoints (
...
+The HPA controller computes utilization as usage divided by requests. The requests come only from the pod spec: container
+requests are summed directly, and pod-level resources are used when set and no container is named in the metric. The controller
+never reads `pod.status.nodeAllocatableResourceClaimStatuses`.
+
+For a pod with a DRA footprint, the denominator is therefore too small, and the reported utilization is too high by the factor
+(spec + DRA) / spec. This always results in over-scaling. Three configurations are not affected: pod-level resources
+(pod-level value covers the DRA footprint, so the denominator is correct), `AverageValue` targets (no denominator), and custom or external metrics.
+
+The DRA-aware request computation already exists in the shared request helper behind an opt-in option,
+and the scheduler, Kubelet, and quota use it. The HPA controller passes the same option when computing requests, so the
```
johnbelamaric: are you saying you will update it to pass that option, or it already does so it will just work?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1885,62 +1964,102 @@ without needing to persist DRA allocations in Kubelet's local disk checkpoints (
...
+resources. This includes
+
+*   The autoscaler computes pod requests through the shared `component-helpers` function. We need to pass the
+    same DRA-aware option that the scheduler, Kubelet, and quota already use so the DRA footprint is included.
+*   The scheduler records the footprint in `pod.status.nodeAllocatableResourceClaimStatuses` during `PreBind`,
+    which the simulation does not run. The autoscaler already fills in the claim reservation that `PreBind`
+    would have made, so it can fill in this field the same way, and clear it when a pod is unscheduled.
 
-Integrating DRA node allocatable resources would involve ensuring this helper is called with the appropriate options to include `pod.status.nodeAllocatableResourceClaimStatuses`. The implications of this change need to be discussed.
+### Future Enhancements
```
johnbelamaric, line 2079: Is there anything we need to do with respect to #5677, which is going beta as well?

keps/sig-scheduling/5517-dra-node-allocatable-resources/kep.yaml
```diff
@@ -4,11 +4,10 @@ authors:
   - "@pravk03"
 owning-sig: sig-scheduling
 participating-sigs:
-  - sig-scheduling
```
dom4ha: It probably shouldn't be removed

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: This needs to be updated with beta, but it would be a minimal change to enable the right flag while calling PodRequests() helper. I will rephrase this part.

pravk03:
/assign @jackfrancis @towca

Could you please take a look from the sig-autoscaling side. I also have https://github.com/kubernetes/autoscaler/pull/10300 up which builds on existing DRA support in CA, but we still need some minimal changes.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: We could update `kubectl describe node` to cover overall consumption including DRA. `ResourcePoolStatusRequest ` can continue to show per-pool device availability. So we'd look at both to see remaining node allocatable capacity and unallocated DRA devices.

I do not see a good way to solve this in a single place though. For regular pods without claims, we do not have the detailed information as to which exact DRA device the capacity is coming out of, so `ResourcePoolStatusRequest` cannot accurately show what regular pods are consuming.

I updated the KEP to include this (`Node Capacity Reporting` section).

johnbelamaric: Ok, yeah. I think since these come from node allocatable, it makes sense to show up in `describe node`.

So for overhead, I don't think rpsr needs to do anything. For mapped resources, what will rpsr show? Will it be useful or misleading? Should we do something about it?

cc @nmn3m

pravk03: I would say its incomplete on its own, but still useful when combined with `kubectl describe node`. 

For example, in the CPU DRA driver case, its is useful if we want to see how much of a specific NUMA node is unallocated for a new exclusive CPU claim. But we would need to check the overall node usage too.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -336,7 +342,7 @@ cache, and schedules the pod. The user did not need to guess which resource to p
 
 The proposal here is to implement a **"Unified Accounting and Enforcement"** model across the control plane and the host for node allocatable resources requested through the standard pod Spec or through Dynamic Resource Allocation (DRA) claims. This involves:
 1.  **API Changes**: Updates to the DRA API for drivers to declare node allocatable resource implications in `Device` objects, and PodStatus to record DRA-based node allocatable resource allocations.
-2.  **Kube-Scheduler Changes**: Modifications in `NodeResourcesFit` and `DynamicResources` plugins to synchronize node resource usage tracking, delegating authoritative node-fit checks to the `DynamicResources` plugin when a pod utilizes DRA claims.
+2.  **Kube-Scheduler Changes**: The `DynamicResources` plugin performs the node-fit check for pods whose claims consume node allocatable resources, in addition to the existing check in `NodeResourcesFit`, which continues to evaluate standard requests. Resource scoring accounts for the footprint of the pod being scored on the node being considered.
```
dom4ha, line 345: IMHO the model in which the `DynamicResources` plugin performs node-fit check should not be continued as Beta grade solution. I think we need more generic approach to native resources accounting and allow different plugins contribute with their requirements. The `NodeResourcesFit` should remain the plugin that takes filtering, scoring and quota enforcement decisions.

This however requires changing the Pod API and making it DRA agnostic as well, so that other plugins could request additional resources without making a plugin specific API change (baked into Pod object and resource calculation framework).

For another alpha, I think it'd be fine to continue with the model currently proposed in this KEP, but to make it beta, I'd expect following the plugin-agnostic approach, especially on the API side, which is harder to change once promoted. Note that we have WAS+DRA integration concerns, and this effort is a good example of trying to keep the scheduling API plugin agnostic. Internal dependencies between scheduler plugins can be cleaned up later.

I'd imagine that the status allows to specify additional `Pod.Spec's` with annotation which plugin requested it (AI drafted):

```
status:
  # Extra plugin specific native resource request
  allocatedResources:
  - pluginName: "DynamicResources"
    source: "ResourceClaim/h100-gpu-claim"
    
    containerAllocations:
      trainer:
        requests:
          cpu: "2"
          memory: "8Gi"
        limits:
          memory: "8Gi"
          
    initContainerAllocations:
      logging-sidecar:
        requests:
          memory: "256Mi"
          
    overhead:
      hugepages-2Mi: "1Gi"
```

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -959,12 +965,14 @@ type NodeAllocatableOverheadResources struct {
...
+* The keys in the `nodeAllocatableResources` map must be exactly `cpu`, `memory`, `hugepages-<size>`, or `ephemeral-storage`.
+  All other names, including extended resources, are rejected.
 * Within a single resource mapping, at least one of the `mapping` or `overhead` fields must be specified.
 * If `mapping` is specified, it must use either `deviceMultiplier` or a combination of `capacityKey` and `capacityMultiplier`. These options are mutually exclusive.
 * If `capacityKey` is specified, it must be a valid qualified name and `capacityMultiplier` is required.
 * If the `overhead` field is specified, it must contain at least one non-negative value for either the `perPod` or `perContainer` overhead quantities.
 * For `PodStatus` updates, each entry in the `nodeAllocatableResourceClaimStatuses` array must reference a valid claim name and contain correctly formatted resource quantities.
+* Ephemeral storage is validated and supported for root filesystem allocations (see [Ephemeral Storage Support and Eviction](#ephemeral-storage-support-and-eviction) in Node Resource Enforcement for details).
 
 ### Kube-Scheduler Changes
```
dom4ha, line 977: This section should be extended with the Scoring phase

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1885,62 +1961,122 @@ without needing to persist DRA allocations in Kubelet's local disk checkpoints (
...
+**Current Behavior:**
+In the DRA plugin, `PreBind` calls `bindClaim` to record claim allocation and reservation in etcd before patching
+pod status. If a subsequent step fails, standard `Unreserve` removes the pod from `claim.status.reservedFor` but retains `claim.status.allocation` so retries on the same node can reuse it.
+
+If ResourceQuota admission evaluated the pod status patch after `bindClaim`, an over-quota pod would be 
+rejected but leave its claim allocated in etcd. While the pod waits in the scheduler backoff queue, those node
+devices and capacity would remain unavailable to other workloads.
+
+**Proposed Change:**
+This KEP proposes that for node allocatable claims we change the `PreBind` sequence so the pod status patch
```
dom4ha, line 2008: Since NodeResourceFit plugin already understands DRA resources (it takes them into account in scoring), it should be responsible for checking against the quota as well.

If we transition to a generic extra native resources tracking, the responsibility to patch status and check quota would be on `NodeResourceFit` plugin, not on the DRA one.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1885,62 +1961,122 @@ without needing to persist DRA allocations in Kubelet's local disk checkpoints (
...
+
+The HPA controller computes utilization as usage divided by requests. The requests come only from the pod spec: container
+requests are summed directly, and pod-level resources are used when set and no container is named in the metric. The controller
+never reads `pod.status.nodeAllocatableResourceClaimStatuses`.
+
+For a pod with a DRA footprint, the denominator is therefore too small, and the reported utilization is too high by the factor
+(spec + DRA) / spec. This always results in over-scaling. Three configurations are not affected: pod-level resources
+(pod-level value covers the DRA footprint, so the denominator is correct), `AverageValue` targets (no denominator), and custom or external metrics.
+
+The DRA-aware request computation already exists in the shared request helper behind an opt-in option,
```
dom4ha, line 2039: If we decide to make the native resource accounting generic (plugin agnostic), the components that just need to read, would not need to enable feature flags, but follow what is in the status. However, the components which does scheduling (like CA), would still need to know whether a certain feature is enabled or not.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Thanks @dom4ha. I am still thinking through this, but some initial thoughts and questions:

> The NodeResourcesFit should remain the plugin that takes filtering, scoring and quota enforcement decisions.

Scoring as proposed now stays in `NodeResourcesFit`, and it is *likely possible* (I will explore more) to move the pod's node allocatable status update (and hence quota enforcement) in PreBind from DynamicResources to NodeResourcesFit. The main challenge I see is with the resource fit check during Filter and keeping it agnostic to plugin order.

> This however requires changing the Pod API and making it DRA agnostic as well, so that other plugins could request additional resources without making a plugin specific API change 

I think the challenge here would be with modeling different types of DRA devices. We currently have 2 
1. devices that actively manage node allocatable resources like CPU/Mem drivers - modeled through `mapping` 
2. devices that manage other hardware but need auxiliary node allocatable resources - modeled through `overhead`

This distinction is needed to determine if claims can be shared across pods and for cgroup enforcement on the node. Further, for better enforcement and handoff between kubelet and the DRA driver, we may also more details on the type of allocation that the driver is doing - eg: exclusive or shared CPUs. This is discussed in [Future Enhancements section](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md#pass-allocation-details-from-driver-to-kubelet) currently and we may want to explore these as separate KEPs.

dom4ha: > The main challenge I see is with the resource fit check during Filter and keeping it agnostic to plugin order.

We are aware of it. Introducing configurable ordering might be an option, but at the beginning we could start with checking resources in both plugins (in all plugins that contribute to native resources requirements).

> I think the challenge here would be with modeling different types of DRA devices.

Is it possible to decompose it and make resource accounting separate from the DRA specific requirements that Node could infer from elsewhere?

pravk03: >  Introducing configurable ordering might be an option, but at the beginning we could start with checking resources in both plugins

Currently, `NodeResourcesFit` checks fit based on spec, DRA plugin checks based on spec + DRA. But this doesn't work if DRA plugin runs first. If i understand the ask here correctly, each plugin that contributes node allocatable, checks the footprint accumulated so far i,e `NodeResourcesFit` also checks spec + DRA (if resolved). I think that should be possible. 

> Is it possible to decompose it and make resource accounting separate from the DRA specific requirements that Node could infer from elsewhere?

Let me explore the possibilities here and follow up.

pravk03: After exploring this further and following offline discussions with @dom4ha, @macsko, and @johnbelamaric:

A per-container split would not work for DRA because a ResourceClaim can be shared across multiple containers within the same pod. We also need the distinction between `mapping` and `overhead` for claim sharing and Kubelet cgroup enforcement.

However, the `PodStatus` fields do not need to be DRA specific and the structure can represent any additional node-allocatable resources allocated outside of the pod spec. I propose renaming the existing fields to be DRA-agnostic in a way that they can be reused or extended for other use cases.

<pre>
// In PodStatus (staging/src/k8s.io/api/core/v1/types.go):
// AdditionalNodeAllocatableResources contains the status of node-allocatable resources that 
// were allocated outside of direct spec requests.
// <b>(Renamed from NodeAllocatableResourceClaimStatuses []NodeAllocatableResourceClaimStatus)</b>
AdditionalNodeAllocatableResources []AdditionalNodeAllocatableResource

// AdditionalNodeAllocatableResource describes the status of
// node-allocatable resources allocated outside of direct spec requests.
type AdditionalNodeAllocatableResource struct {
  // SourceKind is the kind of the object this contribution originates from (e.g., ResourceClaim).
  // <b>(New field)</b>
  SourceKind NodeAllocatableResourceSourceKind
  // SourceName is the name of the source object in the pod's namespace (e.g., name of the resource claim).
  // <b>(Renamed from ResourceClaimName)</b>
  SourceName string
  // Containers lists the names of all containers in this pod that reference the source.
  // <b>(Same as before)</b>
  Containers []string
  // Mapping contains fixed node-allocatable resource quantities allocated once per sourceKind/sourceName.
  // <b>(Same as before)</b>
  Mapping []NodeAllocatableMappedResources
  // Overhead contains variable node-allocatable resource overheads incurred per pod (PerPod) and
  // per referencing container (PerContainer) when using the source.
  // <b>(Same as before)</b>
  Overhead []NodeAllocatableOverheadResources
}
</pre>


`Mapping` represents a fixed capacity accounted once per `SourceKind`/`SourceName` combination. `Overhead` represents the variable capacity accounted based on each reference to `SourceKind`/`SourceName` (like perPod, perContainer). We can start with `ResourceClaim` as the only supported value for `SourceKind`, while the behavior of the rest of the fields remains unchanged from alpha2.

@liggitt @dom4ha Does this API direction look reasonable ?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Resolving this since this is not considered for 1.38 currently. We can pick this up in 1.39 as a separate KEP (https://github.com/kubernetes/enhancements/issues/6352).

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Added scoring stage details.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Updated `Resource Calculation` and the `Cluster Autoscaler` section to clarify this

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: I have updated the PreBind and ResourceQuota sections. PTAL. Based on my current investigation, it should be possible to move the pod.status patch to NodeResourcesFit, and I will look more closely into this aspect during implementation and see if there are edge cases.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1696,9 +1753,48 @@ To manage node stability during Out-Of-Memory (OOM) events, Kubelet applies DRA
...
+    evicted.
+
+The resource quantities are available through `pod.status.nodeAllocatableResourceClaimStatuses` like cpu and memory, so `ResourceQuota`
+accounts them under `requests.ephemeral-storage` and `limits.ephemeral-storage` with no additional changes.
+
+
 #### Integration with Memory QoS
 
-Memory QoS [KEP-2570](https://github.com/kubernetes/enhancements/pull/6143) is proposed for beta graduation in v1.37. This configures cgroup v2 memory knobs at both container-level
+Memory QoS [KEP-2570](https://github.com/kubernetes/enhancements/pull/6143) configures cgroup v2 memory knobs at both container-level
```
ffromani, line 1870: My only remaining major item is about the memory itself.
Hence, the following question. Let me stress this is really an open question I see possible reasons both to include that portion in the KEP scope AND to defer to a followup KEP.

This KEP wants to add provisions for all the node allocatable resources, and the CPU side is well handled, with a DRA driver being developed and serving both as early adopter and as validation for the provisions of this KEP. But this is not is the case for memory and hugepages. So, do we need or want to have some validation or some feedback for beta graduation or for GA graduation? do we want or need to defer that portion to a followup?

Would we benefit from the feedback of a DRA memory driver, or from adding memory support to an existing driver (we talked about merging the memory support in the DRA CPU driver and rename it accordingly) as beta/beta2 gate or as GA gate?

ffromani (commented):
From the node enforcement side this KEP is (still!) conservative and proposes no big changes, which is likely the right call considering the already huge scope and impact.
In this case it seems to me that, if we need changes in the enforcement area, we probably can handle them in a followup KEP.
But I want to raise the topic of the memory managed by a DRA driver: if nothing else, we can agree about deferring that work as well.
Other than this, it's almost LGTM to me. Need another and probably final pass before to label it, but looks good.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
liggitt: Removing the DRA-specific shape of the API seems reasonable, as long as we don't lose validation.

Are we expecting to accept arbitrary group/kind values in the future, or only ones the API server knows about and can validate specifically in container references?

The source bit is so close to the apiGroup/kind/name tuple, similar to PVC `spec.dataSource`, I'd consider grouping sourceKind/sourceName and adding apiGroup to make `source:{apiGroup:..., kind:..., name:...}`.

---

With new names, those are effectively new fields, so a 1.37 server would drop the data and a 1.38 server would allow it in. We could still possibly call it beta in terms of behavior we have experience with, but we couldn't make the feature default-enabled in 1.38 with new fields in a GA type. You could possibly mark the feature as `beta, default=false` in 1.38, and then have it `beta, default=true with minCompatibilityVersion=1.38`, so once all servers were up to 1.38 it could default-enable ...

see other examples of MinCompatibilityVersion, like
```
DRAFractionalCapacityRange: {
        {Version: version.MustParse("1.37"), Default: false, PreRelease: featuregate.Beta},
        {Version: version.MustParse("1.37"), Default: true, PreRelease: featuregate.Beta, MinCompatibilityVersion: version.MustParse("1.37")},
    },
```

that makes it _possible_ to have on by default in 1.38, but only for clusters that do a final step to set min-compatibility-version=1.38 after upgrading

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Agree that it would good to get this feedback from a memory dra driver usecase. I think we can call this out as a GA graduation criteria, or even beta2 criteria (with https://github.com/kubernetes/enhancements/pull/6324#discussion_r4075237861 my understanding is that we need atleast another beta before GA)

liggitt: (just a reminder that beta fields in GA types don't get the benefit of a do-over ... once they're beta, we're committed to that schema and to supporting data persisted in those fields)

pravk03: Ack. The current API fields are pretty minimal and must remain required for memory too. The feedback would give more confidence on the node cgroup enforcement parts and any new fields we may want to introduce in future KEPs (specifically in `mapping`).

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Thanks!. Updated the API to use source: {apiGroup, kind, name} (NodeAllocatableResourceSource) and updated the validation rules. Expecting only known apiGroup/kind types should be acceptable (cc @dom4ha for WAS use cases) so 1.38 validation will only allow resource.k8s.io/ResourceClaim. 

Also updated the Beta graduation criteria with the `MinCompatibilityVersion: 1.38` restriction.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -2324,17 +2604,40 @@ E2E tests will be added to `test/e2e/dra`:
...
+    `MinCompatibilityVersion` is `1.38`.
+
+#### Beta 2
+-   Feature gate `DRANodeAllocatableResources` is enabled by default (`Default: true`) without requiring `MinCompatibilityVersion` configuration.
+-   Validation and feedback from DRA driver/s managing CPU and memory.
+
+#### GA
+
+-   Validation and feedback from DRA driver/s managing CPU and memory.
+-   The feature has been enabled by default for at least two releases with no critical bug reports.
```
bart0sh, line 2633: As we're not enabling the feature in 1.38, this makes `stable: 1.40` milestone incorrect in the kep.yaml. Either this should be corrected or the milestone in the kep.yaml.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -2324,17 +2604,40 @@ E2E tests will be added to `test/e2e/dra`:
...
+-   HPA integration - pod utilization accounts for DRA node allocatable resources.
+-   Cluster Autoscaler integration - binpacking and node utilization account for DRA node allocatable
+    resources, including for pods placed during simulation and pods duplicated into node templates.
+-   `kubectl describe node` integration - per-pod rows and the summary consider both spec and DRA values.
+-   Feature gate `DRANodeAllocatableResources` graduates to Beta. Due to the `v1.PodStatus` field rename in v1.38, 
+    it defaults to `false` under the default `MinCompatibilityVersion` (N-1) and `true` when 
+    `MinCompatibilityVersion` is `1.38`.
+
+#### Beta 2
+-   Feature gate `DRANodeAllocatableResources` is enabled by default (`Default: true`) without requiring `MinCompatibilityVersion` configuration.
```
bart0sh, line 2627: Do we have any release target for Beta 2? 1.39 may be?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -2324,17 +2604,40 @@ E2E tests will be added to `test/e2e/dra`:
...
+-   Per-node scoring in `NodeResourcesFit` and `NodeResourcesBalancedAllocation`.
+-   Requeue events for capacity: pods rejected on DRA-augmented capacity are requeued on pod deletion and
+    pod scale-down, not only on claim and slice changes.
+-   HPA integration - pod utilization accounts for DRA node allocatable resources.
+-   Cluster Autoscaler integration - binpacking and node utilization account for DRA node allocatable
+    resources, including for pods placed during simulation and pods duplicated into node templates.
+-   `kubectl describe node` integration - per-pod rows and the summary consider both spec and DRA values.
+-   Feature gate `DRANodeAllocatableResources` graduates to Beta. Due to the `v1.PodStatus` field rename in v1.38, 
+    it defaults to `false` under the default `MinCompatibilityVersion` (N-1) and `true` when 
+    `MinCompatibilityVersion` is `1.38`.
```
bart0sh, line 2624: Should we add e2e/e2e_node test criteria?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -2295,6 +2567,14 @@ E2E tests will be added to `test/e2e/dra`:
     -   Fungible claims involving node allocatable resources.
 -   Verify that Kubelet enforces correct cgroup limits on running containers without kernel throttling or OOM kills, and applies correct OOM score adjustments.
 
+Beta:
+-   Verify that a pod requesting DRA node allocatable resources exceeding namespace quota is rejected at PreBind in a running cluster.
```
bart0sh, line 2571: Should we add more tests for the new functionality planned for Beta, e.g. eviction, HugePages, Memory QoS, HPA, etc?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -336,7 +342,9 @@ cache, and schedules the pod. The user did not need to guess which resource to p
...
 1.  **API Changes**: Updates to the DRA API for drivers to declare node allocatable resource implications in `Device` objects, and PodStatus to record DRA-based node allocatable resource allocations.
-2.  **Kube-Scheduler Changes**: Modifications in `NodeResourcesFit` and `DynamicResources` plugins to synchronize node resource usage tracking, delegating authoritative node-fit checks to the `DynamicResources` plugin when a pod utilizes DRA claims.
+2.  **Kube-Scheduler Changes**: Both `NodeResourcesFit` and `DynamicResources` evaluate the pod's accumulated resource 
+requirement (i.e., standard requests plus resolved DRA claim allocations) so the fit check is independent of 
+plugin ordering. Resource scoring accounts for the DRA footprint of both existing pods on the node and the candidate pod being scored.
 3.  **Kubelet Changes**: Updates in Kubelet to take into account resources allocated through DRA in the cgroup enforcement.
```
bart0sh, line 348: Should kube-controller-manager changes be mentioned here?

keps/sig-scheduling/5517-dra-node-allocatable-resources/kep.yaml
```diff
@@ -17,28 +17,35 @@ reviewers:
...
       - kube-scheduler
       - kube-apiserver
       - kubelet
+      - kube-controller-manager
 disable-supported: true
-
-metrics: []
\ No newline at end of file
+metrics:
+  - scheduler_plugin_execution_duration_seconds{plugin="DynamicResources"}
```
bart0sh, line 48: Should we add `NodeResourcesFit` plugin metrics here as well?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Yes, that is the goal since the current default off is because of API rename.

pravk03: Added

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: HugePages cgroup update is already covered. Added the rest.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Added.

keps/sig-scheduling/5517-dra-node-allocatable-resources/kep.yaml
pravk03: Yes, now that we want to the filter check to not be order dependent.
Added.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Added

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Updated GA to 1.41 in kep.yaml. Although, I am not sure what the recommendation is with min_compatibility_version.

bart0sh: 1.38, I guess. Well, unless there are incompatible changes in 1.39, but that's another story.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1320,15 +1442,6 @@ This was also explored as part of [KEP-1287](https://github.com/kubernetes/enhan
...
 
-##### Handling Pod Level Resources
-
-When `PodLevelResources` is used, the Kubelet's cgroup enforcement must reconcile explicit pod-level limits with DRA allocations. This requires two specific adjustments:
-*   **Pod-Level Cgroup Ceilings**:
-    If explicit pod-level limits are specified, they determine the overall pod budget. The Kubelet sets the pod's cgroup ceiling exactly to the specified `pod.spec.resources.limits`. 
-    It **does not add the DRA allocations** to the pod-level limit, because the DRA resources are already encompassed within this overall budget.
-*   **Container-Level Fallbacks**: 
-    If a container lacks its own limit, the pod-level limit is applied to the container's cgroup maximum value.
-
```
ffromani: I miss the reason for this removal. I guess the prescriptions of this section are now embedded elsewhere in the text? I'm just checking the pod-level resources interaction and requirement are documented and not lost in the churn

ffromani (commented):
LGTM from my side pending one inline question

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1004,46 +1075,64 @@ plugins.
...
+    and consumed by Kubelet during [pod admission](#kubelet-admission-control) and [cgroup enforcement](#cgroup-enforcement). 
+    The `DynamicResources` calls `bindClaim` to update `ResourceClaim.Status` with the allocated devices and reservation.
+    -   If `NodeResourcesFit` runs before the `DynamicResources` plugin (the default order), `NodeResourcesFit.PreBind` patches `pod.status.additionalNodeAllocatableResources` first. 
+        If quota is exceeded, `PreBind` fails and `ResourceClaim` is not reserved. If `NodeResourcesFit.PreBind` succeeds and `DynamicResources.PreBind` (or `Bind`) fails subsequently, 
+        `NodeResourcesFit.Unreserve` clears `pod.status.additionalNodeAllocatableResources` to release the quota.
+    -   If `NodeResourcesFit` runs after the `DynamicResources` plugin, `DynamicResources.PreBind` (`bindClaim`) runs first, followed by `NodeResourcesFit.PreBind` patching 
+        `pod.status.additionalNodeAllocatableResources`. If `NodeResourcesFit.PreBind` fails (e.g., on quota rejection), the scheduler runs `Unreserve`, where 
+        `DynamicResources.Unreserve` removes the claim reservation (`ReservedFor`) and reverts the claim allocation.
```
macsko: PreBind plugins can be called in parallel if they return `AllowParallel`. Are we okay with running both NodeResourcesFit and DynamicResources in parallel or there is some risk?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1876,71 +2040,102 @@ The Kubelet has its own admission check
...
-    These node-specific allocations are stored transiently in the in-memory `CycleState`.
-2.  **Score Stage (`NodeResourcesFit` Plugin)**: Nodes are scored using CPU/Memory spreading or packing algorithms. Although the allocations exist in `CycleState` at this point
-    `NodeResourcesFit` does not read them because:
-    *   `PreScore` calculates the pod's resource footprint once for the entire cycle, to be able to include DRA based allocations, the `NodeResourcesFit` plugin should read 
-    `DynamicResources`' internal state which is challenging and introduces coupling between plugins.
-3.  **PreBind Stage (`DynamicResources` Plugin)**: Only after a node is selected and reserved does the scheduler patch the **`Pod.Status`** in the API server to persist the 
-    `NodeAllocatableResourceClaimStatuses` field.
+Because the scheduler's `Filter` stage does not consider quota, a pod in a namespace that is out of quota still goes through node selection before evaluation in `PreBind` on each backoff retry. Exponential backoff bounds the retry churn, as namespace quota usage is not part of scheduler state and the pod's exact DRA footprint is only known after node selection.
```
macsko: This can have some negative performance implications, because the scheduler might send non-effective API calls, saturating the QPS that they share with other operations like pod binding. It would be particularly problematic if there are many pending pods that don't meet the quota. However, I think it's still acceptable and I don't see other good mitigations except the mentioned backoff time (which is 10 seconds at most). WDYT @dom4ha?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -976,14 +1041,20 @@ plugins.
...
 2. **Filter Stage:** This stage performs the node-level checks to determine if a pod fits on a specific node.
-   *  **NodeResourcesFit Plugin:** In the Alpha stage, this plugin would continue to do the resource fit based on standard requests.
-   *  **DynamicResources Plugin:** This plugin takes on the authoritative role for checking node allocatable resource fit if any of the
-      pod's `ResourceClaim`s request node allocatable resources.
+   *  **NodeResourcesFit Plugin:** This plugin checks the pod's demand against remaining node capacity:
+      -   If `NodeResourcesFit` runs before the `DynamicResources` plugin (the default order), the DRA values are
+          not resolved yet, so it checks resource fit based on spec requests only. The `DynamicResources` plugin
+          then does the authoritative check based on spec + DRA later in the chain.
+      -   If `NodeResourcesFit` runs after the `DynamicResources` plugin, the pod's node-specific DRA allocations
+          are already resolved in `CycleState`, so it includes them and checks fit based on spec + DRA.
```
dom4ha: I think we should clearly specify that the native resources requests are stored under plugin agnostic key.

What's also important, the approach opens a possibility to request additional resources also for yet another plugin. All mechanisms described in this KEP should scale to more plugins (QHints, PreFilter checks or quota enforcement), but it's worth to point it out clearly, as we don't want to design a solution that works only for two sepcific plugins.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1004,46 +1075,64 @@ plugins.
...
+        `pod.status.additionalNodeAllocatableResources`. If `NodeResourcesFit.PreBind` fails (e.g., on quota rejection), the scheduler runs `Unreserve`, where 
+        `DynamicResources.Unreserve` removes the claim reservation (`ReservedFor`) and reverts the claim allocation.
+
+6.  **Bind Stage:** This stage executes asynchronously after the main scheduling cycle has decided on a node. The scheduler
     listens for pod `Update` events, and transitions the pod from the "assumed" state to "bound" if the bind process
     succeeded. The resource accounting on the `NodeInfo` does not change at this point (as they were previously accounted for
     during the "Assume" step). If the bind fails, or if the Kubelet later rejects the Pod, the scheduler detects this and
     reverts the resource allocation in its cache, decrementing `nodeInfo.Requested`.
 
+**Requeueing:** a pod rejected because a node's remaining capacity cannot cover its DRA-augmented demand becomes
```
dom4ha: If we have three plugins that can request additional resources (counting ResourceFit plugin), none of them will have ability to assess the total number (even now the total is node specific).

I'm not saying it's required to write correct QHint rules, but I'm saying that these rules should be agnostic to such total.

That means that it probably should be enough to have correct QHint rules in ResourceFit plugin as QHints in other plugins probably won't bring anything new.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
dom4ha: The proposed Source LGTM. IIUC each claim will have a separate status entry.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
dom4ha: This falls into a class of cases when scheduling is performed by binding permanently fails. That's a good argument why we should consider having a quota enforcement plugin, which could block scheduling at earlier phases. Even though it sounds doable, it'd be probably a separate bigger effort.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
dom4ha: I bet it should be fine, since any plugin should be able to abort binding. Do you see a specific situation when it can be a problem?

macsko: I agree that there shouldn't be any issue

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
macsko: Scheduling queue calls QHints only for the plugin that rejected pod. If a plugin can reject the pod for some reason, it should register QHint matching that rejection, because only this plugin might be registered (we run Filters sequentially and break on rejection). This means that both NodeResourceFit and DynamicResources plugins have to register the QHints for these events to wake up the pod correctly.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -488,24 +497,33 @@ We add a new field `NodeAllocatableResourceClaimStatuses` to `PodStatus` as a wa
...
+
+  // AdditionalNodeAllocatableResources contains the status of node allocatable resources
+  // that were allocated for this pod outside of direct spec requests (e.g., through DRA claims). 
+  // This includes resources currently
   // reported in v1.Node `status.allocatable` that are not extended resources
   // (see https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#extended-resources).
   // Examples include "cpu", "memory", "ephemeral-storage", and hugepages.
   // +featureGate=DRANodeAllocatableResources
   // +optional
   // +listType=atomic
```
liggitt: right now, this field is marked as atomic and assumed to have a single writer, which is the scheduler

if we make this less DRA-specific, are we expecting multiple writers to contribute to this field, or still a single writer (probably the scheduler)?

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -488,24 +497,33 @@ We add a new field `NodeAllocatableResourceClaimStatuses` to `PodStatus` as a wa
...
+	// ResourceClaimName is tombstoned since it got replaced with Source.
+	// ResourceClaimName string `json:"resourceClaimName" protobuf:"bytes,1,opt,name=resourceClaimName"`
+
+	// Source identifies the object in the pod's namespace that this resource
+	// contribution originates from (e.g., a ResourceClaim).
 	// +required
 	// +k8s:required
-	ResourceClaimName string `json:"resourceClaimName" protobuf:"bytes,1,opt,name=resourceClaimName"`
-	// Containers lists the names of all containers in this pod that reference the claim.
+	Source NodeAllocatableResourceSource `json:"source" protobuf:"bytes,6,opt,name=source"`
```
liggitt: naming nit, can figure this out during implementation, but "...ResourceSource" sounds really funny :)

Other reference types end in `Reference`, so maybe `AdditionalNodeAllocatableReference`?

liggitt (commented):
the overall approach on the API bits lgtm

a couple questions/tweaks on the API surface, can fold those in here or during implementation

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
```diff
@@ -1876,71 +2040,102 @@ The Kubelet has its own admission check
...
-    This is because the scheduler's `Assume` stage updates `NodeInfo.Requested` with both standard Spec requests and dynamic DRA status claim allocations
-    for all previously assumed pods on the node.
-*   For the incoming pod being scored, scoring in `NodeResourcesFit` only considers CPU and Memory requests defined directly in the pod's Spec. It does 
-    not account for the incoming pod's DRA based allocations.
+For pods without pod-level resources covering the resource, quota is evaluated during `PreBind` when `NodeResourcesFit.PreBind` patches `pod.status.additionalNodeAllocatableResources`.
+*   **`ResourceQuota` Admission on `pod/status`:** The `ResourceQuota` admission plugin in `kube-apiserver` currently skips `status` subresources. This would be updated to evaluate
+    `pod/status` updates that modify `additionalNodeAllocatableResources` to a non-empty value.
+*   **On Quota Rejection:** If the status patch exceeds namespace quota, `kube-apiserver` rejects the patch. `NodeResourcesFit.PreBind` returns an error and `PreBind` aborts before
```
dom4ha: IIUC this mechanism will prevent exceeding the quota, but won't allow to release pod static resources, since the pod was already admitted.

I wouldn't say it's a blocker, as this is how quota enforcement works. All unschedulable pods already consumed the quota even though they cannot be scheduled for some reason.

It may be worth however to add a sentence about it here.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Thanks. This was a mistake, adding it back.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Yes, that's right.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: I will list this a future enhancement for now.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: The non-DRA sources are still expected to be other scheduler plugins, so I think atomic is still right for this field.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Updated here and in a few other places. 
I have also added a separate paragraph above to cover the parts of this design that are extensible to other plugins and the key principles.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Yeah, good point. Added a note about this.

keps/sig-scheduling/5517-dra-node-allocatable-resources/README.md
pravk03: Haha, yeah!. Renamed.

PR #6329, opened 2026-09-07
KEP-1645: Allow exporting to any set of clusters and report it in ServiceExport status

kahirokunn:
- One-line PR description: Allow a service to be exported to any set of clusters, let an implementation define a clusterset per service, and report those clusters in ServiceExport status.
- Issue link: https://github.com/kubernetes/enhancements/issues/1645
- Other comments:

A service may be exported to any set of other clusters, and which ones is implementation-defined. An implementation may define a clusterset per service, as the clusters that export the service or that it is exported to, and within such a clusterset namespace sameness applies only to the service's namespace. One clusterset for all services stays the default. Requiring one clusterset per cluster was ruled out because those clustersets cannot partially overlap: if cluster C shares services with A and B and also with D, all four would be one clusterset.

Each cluster's ServiceImport combines its own export and the exports that reach it, and conflicts are resolved among those exports. A cluster resolves only the names of the services it exports or that are exported to it.

Operators on an exporting cluster can use `ServiceExport.Status.Clusters` to see which clusters the service is exported to. Entries use the existing `ClusterStatus` format. The field is informational only, and an entry means the service is available for that cluster to import, whether or not the cluster has imported it.

Related discussion: https://github.com/kubernetes-sigs/mcs-api/issues/168

### DIFF
--- a/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
+++ b/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
@@ -87,6 +87,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [User Stories](#user-stories)
     - [Different ClusterIP Services Each Deployed to Separate Cluster](#different-clusterip-services-each-deployed-to-separate-cluster)
     - [Single Service Deployed to Multiple Clusters](#single-service-deployed-to-multiple-clusters)
+    - [Troubleshooting a Service Export](#troubleshooting-a-service-export)
   - [Constraints](#constraints)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
@@ -350,6 +351,12 @@ removal without action by or impact on the caller. Routing to my replicated
 service should optimize for cost metric (e.g. prioritize traffic local to zone,
 region).
 
+#### Troubleshooting a Service Export
+
+I operate a cluster that exports services to other clusters in a clusterset.
+When troubleshooting an export, I want to see which clusters the service is
+exported to.
+
 ### Constraints
 
 <!--
@@ -378,6 +385,10 @@ How will UX be reviewed and by whom?
 
 Consider including folks that also work outside the SIG or subproject.
 -->
+
+`ServiceExport.Status.Clusters` shows anyone who can read the `ServiceExport`
+which clusters the service is exported to. Implementations may leave the field
+absent (see [Exporting Services](#exporting-services)).
 
 ## Design Details
 
@@ -426,6 +437,14 @@ type ServiceExportSpec struct {
 
 // ServiceExportStatus contains the current status of an export.
 type ServiceExportStatus struct {
+        // clusters is the list of other clusters to which the service is
+        // exported. This field is informational only.
+        // +optional
+        // +patchStrategy=merge
+        // +patchMergeKey=cluster
+        // +listType=map
+        // +listMapKey=cluster
+        Clusters []ClusterStatus `json:"clusters,omitempty"`
         // +optional
         // +patchStrategy=merge
         // +patchMergeKey=type
@@ -442,6 +461,11 @@ metadata:
   name: my-svc
   namespace: my-ns
 status:
+  clusters:
+  - cluster: cluster-b
+  - cluster: cluster-c
+  - cluster: cluster-d
+  - cluster: cluster-e
   conditions:
   - type: Valid
     status: "True"
@@ -459,6 +483,14 @@ status:
     reason: TypeConflict
     message: "Conflicting type. Using \"ClusterSetIP\" from oldest service export in \"cluster-1\". 2/5 clusters disagree."
 ```
+
+Implementations may list in `ServiceExport.Status.Clusters` the other clusters
+to which the service is exported. The field is informational only and does not
+affect export, import, or routing. Entries identify clusters by cluster name.
+
+An entry means that the service is available for that cluster to import, even
+if the cluster does not have the namespace or has not imported it. An absent or
+empty list does not mean that there are no such clusters.
 
 To export a service, a `ServiceExport` should be created within the cluster and
 namespace that the service resides in, name-mapped to the service for export -
@@ -634,6 +666,8 @@ type ServiceImportStatus struct {
   // +kubebuilder:validation:Enum=Present;Absent
   // +optional
   EndpointSliceObjects EndpointSliceObjectsStatus `json:"endpointSliceObjects,omitempty"`
+  // clusters is the list of exporting clusters from which this service
+  // was derived.
   // +optional
   // +patchStrategy=merge
   // +patchMergeKey=cluster
@@ -648,7 +682,7 @@ type ServiceImportStatus struct {
   Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
-// ClusterStatus contains service configuration mapped to a specific source cluster
+// ClusterStatus contains service configuration mapped to a specific cluster
 type ClusterStatus struct {
  Cluster string `json:"cluster"`
 }
--- a/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
+++ b/keps/sig-multicluster/1645-multi-cluster-services-api/README.md
@@ -87,6 +87,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [User Stories](#user-stories)
     - [Different ClusterIP Services Each Deployed to Separate Cluster](#different-clusterip-services-each-deployed-to-separate-cluster)
     - [Single Service Deployed to Multiple Clusters](#single-service-deployed-to-multiple-clusters)
+    - [Service Exported to a Subset of Clusters](#service-exported-to-a-subset-of-clusters)
     - [Troubleshooting a Service Export](#troubleshooting-a-service-export)
   - [Constraints](#constraints)
   - [Risks and Mitigations](#risks-and-mitigations)
@@ -124,6 +125,7 @@ tags, and then generate with `hack/update-toc.sh`.
   - [Export via annotation](#export-via-annotation)
   - [Other conflict resolution algorithms](#other-conflict-resolution-algorithms)
   - [Exporting labels/annotations from the Service/ServiceExport objects](#exporting-labelsannotations-from-the-serviceserviceexport-objects)
+  - [Sharing services without a clusterset](#sharing-services-without-a-clusterset)
 - [Infrastructure Needed](#infrastructure-needed)
 <!-- /toc -->
 
@@ -290,7 +292,7 @@ nitty-gritty.
     https://github.com/kubernetes/community/blob/master/sig-multicluster/namespace-sameness-position-statement.md
 
 We propose a new CRD called `ServiceExport`, used to specify which services
-should be exposed across all clusters in the clusterset. `ServiceExports` must
+should be exposed to other clusters in the clusterset. `ServiceExports` must
 be created in each cluster that the underlying `Service` resides in. Creation of
 a `ServiceExport` in a cluster will signify that the `Service` with the same
 name and namespace as the export should be visible to other clusters in the
@@ -300,13 +302,14 @@ Another CRD called `ServiceImport` will be introduced to act as the in-cluster
 representation of a multi-cluster service in each importing cluster. This is
 analogous to the traditional `Service` type in Kubernetes. Importing clusters
 will have a corresponding `ServiceImport` for each uniquely named `Service` that
-has been exported within the clusterset, referenced by namespaced name.
+they export or that has been exported to them, referenced by namespaced name.
 `ServiceImport` resources will be managed by the MCS implementation's
 mcs-controller.
 
-If multiple clusters export a `Service` with the same namespaced name, they will
-be recognized as a single combined service. For example, if 5 clusters export
-`my-svc.my-ns`, each importing cluster will have one `ServiceImport` named
+If multiple clusters export a `Service` with the same namespaced name to the
+same cluster, they will be recognized as a single combined service in that
+cluster. For example, if 5 clusters export `my-svc.my-ns` to all other clusters
+in the clusterset, each importing cluster will have one `ServiceImport` named
 `my-svc` in the `my-ns` namespace and it will be associated with endpoints from
 all exporting clusters. Properties of the `ServiceImport` (e.g. ports, topology)
 will be derived from a merger of component `Service` properties.
@@ -351,6 +354,18 @@ removal without action by or impact on the caller. Routing to my replicated
 service should optimize for cost metric (e.g. prioritize traffic local to zone,
 region).
 
+#### Service Exported to a Subset of Clusters
+
+I have clusters A, B, C and D in one clusterset. The foo and bar teams share
+clusters A and B. The billing team has stricter security requirements and
+runs only in clusters C and D. The `invoices` service in cluster C is used by
+the `foo` and `bar` services in clusters A and B. The `ledger` service in
+cluster C is used only by the billing reports in cluster D. I want each
+billing service to be visible only to the clusters that use it, so that
+`ledger` is never visible to clusters A or B.
+
+![four clusters in one clusterset where the billing team's services are exported only to the clusters that use them](./export-to-a-subset-of-clusters.svg)
+
 #### Troubleshooting a Service Export
 
 I operate a cluster that exports services to other clusters in a clusterset.
@@ -390,6 +405,9 @@ Consider including folks that also work outside the SIG or subproject.
 which clusters the service is exported to. Implementations may leave the field
 absent (see [Exporting Services](#exporting-services)).
 
+Consumers that assume every cluster in the clusterset imports every exported
+service may look for a `ServiceImport` or endpoints that do not exist.
+
 ## Design Details
 
 <!--
@@ -404,6 +422,9 @@ Services will not be visible to other clusters in the clusterset by default.
 They must be explicitly marked for export by the user. This allows users to
 decide exactly which services should be visible outside of the local cluster.
 
+A service may be exported to all other clusters in the clusterset or only to
+some of them. Which clusters it is exported to is implementation-defined.
+
 Tooling may (and likely will, in the future) be built on top of this to simplify
 the user experience. Some initial ideas are to allow users to specify that all
 services in a given namespace or in a namespace selector or even a whole cluster
@@ -411,8 +432,8 @@ should be automatically exported by default. In that case, a `ServiceExport`
 could be automatically created for all `Services`. This tooling will be designed
 in a separate doc, and is secondary to the main API proposed here.
 
-To mark a service for export to the clusterset, a user will create a
-ServiceExport CR:
+To mark a service for export to other clusters in the clusterset, a user will
+create a ServiceExport CR:
 
 ```golang
 // ServiceExport declares that the associated service should be exported to
@@ -496,8 +517,8 @@ To export a service, a `ServiceExport` should be created within the cluster and
 namespace that the service resides in, name-mapped to the service for export -
 that is, they reference the `Service` with the same name as the export. If
 multiple clusters within the clusterset have `ServiceExports` with the same name
-and namespace, these will be considered the same service and will be combined at
-the clusterset level.
+and namespace, these will be considered the same service and will be combined in
+each importing cluster.
 
 _Note: A `Service` without a corresponding `ServiceExport` in its local cluster
 will not be exported even if other clusters are exporting a `Service` with the
@@ -562,9 +583,12 @@ includes 3 scenarios:
 
 A multi-cluster service will be imported only by clusters in which the service's
 namespace exists. All clusters containing the service's namespace will import
-the service. This means that all exporting clusters will also import the
+the service if it is exported to them. Exporting clusters always import the
 multi-cluster service. _An implementation may or may not decide to create
 missing namespaces automatically, that behavior is out of scope of this spec._
+
+A cluster's `ServiceImport` combines only the local `ServiceExport`, if any, and
+the `ServiceExport`s in other clusters that export the service to it.
 
 Because of the potential wide impact a `ServiceImport` may have within a
 cluster, non-cluster-admin users should not be allowed to create or modify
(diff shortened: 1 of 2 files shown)


keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -460,6 +470,14 @@ status:
     message: "Conflicting type. Using \"ClusterSetIP\" from oldest service export in \"cluster-1\". 2/5 clusters disagree."
 ```
 
+Implementations MAY report where service information is published in
+`ServiceExport.status.clusters`. Entries MUST identify destination clusters by
+name. While information remains available to a destination, its entry MUST NOT
+be excluded for a missing namespace or an ignored or rejected export. Consumers
+MUST NOT assume that listed destinations have accepted or imported the service.
+They MUST tolerate an absent or empty list and MUST NOT assume it means there
+are no destinations.
```
MrFreezeex: BTW we also don't have an equivalent explanation for the similar fields in `ServiceImport` it might be nice to have one (which may be the same paragraph too ~).

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -460,6 +470,14 @@ status:
     message: "Conflicting type. Using \"ClusterSetIP\" from oldest service export in \"cluster-1\". 2/5 clusters disagree."
 ```
 
+Implementations MAY report where service information is published in
+`ServiceExport.status.clusters`. Entries MUST identify destination clusters by
+name. While information remains available to a destination, its entry MUST NOT
+be excluded for a missing namespace or an ignored or rejected export. Consumers
```
MrFreezeex: I am wondering if this wouldn't be too strong? Like there could be some implementation in which they somehow knows that it would get ignored by the remote cluster and not including it in the list is fine?

It may be better in term of UX to always include it like in the current text. For instance we have a concept called "Global Namespace" (which isn't yet documented though) in Cilium where you add an annotation on Namespace to say if the namespace should be global (can be opt-in or opt-out) and in that case we would still create ServiceImports but with a condition signaling that it doesn't work in a non global namespace.

This better UX may not be applicable to every implementations though ~

MrFreezeex (commented):
Thanks! I left a few question/comment inline.

I am wondering if you should make the relevant partial-mesh changes in this PR too though. Those two are only independent if we would consider that adding this cluster list is a good thing in the full mesh case too probably 🤔

lauralorenz:
From SIG-MC call 9/8:
* Is this field intended to be policy or informational? Sounds like its informational. If so, make it super clear that this is NOT policy (as in nothing in MCS controller is expected to enforce or validate the information), maybe literally with "this is not intended to be policy" written all over the place lol
  * FYI prior art on a policy version of this was discussed in the past in a form like/integrating with AdminNetworkPolicy API
* Is the intention to list clusters that could potentially access the service if "they" "chose to" (aka endpoints are reachable, just need to make the namespace to kick off ServiceImport process), or is it to list all the clusters that currently have access in practice (~DNS access to endpoints in a given cluster). Similar to the difference of listing open ports vs active connections on a host.
  * If these (or others) are more states of cluster that we care about, we would need a more complex data structure to give metadata to each cluster
     * If we did this, is it just defining or pre-defining policy?
* Who (if anyone) is using this besides a ServiceExport-ing user? What are the abuse implications? Ex. historically there has been some concern even about exposing which clusters have which namespaces to users within the ClusterSet

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Agreed it read as too strong. I removed the must not and instead wrote what an entry means: the service is available for that cluster to import, whether or not the cluster has the namespace or has imported it. In your global namespace example the cluster would still be listed, and the condition on its ServiceImport is what says it wasn't imported. The list only describes the exporting side, not what the remote cluster does with it.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: The `Clusters` field on `ServiceImportStatus` in the KEP had no comment, so I added the one mcs-api already ships: "clusters is the list of exporting clusters from which this service was derived." Both fields use `ClusterStatus`, so the sentence saying entries are cluster names covers ServiceImport as well.

kahirokunn:
Thanks for bringing this back from the call. On the three points:

- It's informational only, not policy. The text now says the field doesn't affect export, import or routing, so nothing in the mcs-controller is expected to enforce it.
- It's the open ports view. An entry means the service is available for that cluster to import, whether or not the cluster has the namespace or has imported it. Since that's all it says, a plain list of cluster names is enough and there's no per cluster state to carry.
- The main user is whoever owns the ServiceExport, to see where their service went. Since entries don't depend on the namespace existing, the list doesn't reveal which clusters have which namespaces. What it does expose is cluster names to anyone who can read the ServiceExport, so I noted that under Risks and Mitigations. The field is optional, so where that's a concern an implementation can put it behind a flag, for example, and leave it off to avoid the exposure entirely.

kahirokunn:
@MrFreezeex I've added it to this PR, so please take a look when you have a chance.

kahirokunn:
I think this also covers much of the problem raised in kubernetes/community#6748, at least on the MCS side, since an implementation can now scope where a service is shared and `ServiceExport.Status.Clusters` shows the result.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -1051,6 +1107,11 @@ on all `ServiceExport`s for the conflicted service with a description of the con
 The conflict will be resolved by assigning precedence based on each
 `ServiceExport`'s `creationTimestamp`, from oldest to newest.**
 
+Conflicts are detected and resolved in each importing cluster among the
+`ServiceExport`s it combines. The `Conflict` condition is set on a
+`ServiceExport` if a conflict is detected in any importing cluster that
+combines it.
+
```
MrFreezeex: This kind of assumes that each cluster is independently assessing this while in theory you could have a central control plane that decide all of those that each cluster just consume without a second opinion.

You may need to edit the previous paragraph too since it talks about all ServiceExport BTW (it may remove the needs for this additional paragraph too or improve the previous one to incorporate some of the clarification you brought here)

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -350,6 +354,24 @@ removal without action by or impact on the caller. Routing to my replicated
...
+I have clusters A, B, C and D in one clusterset. The foo and bar teams share
+clusters A and B. The billing team has stricter security requirements and
+runs only in clusters C and D. The `invoices` service in cluster C is used by
+the foo and bar services in clusters A and B. The `ledger` service in cluster
+C is used only by the billing reports in cluster D. I want each billing
+service to be visible only to the clusters that use it, so that the ledger is
+never visible to clusters A or B.
+
+![four clusters in one clusterset where the billing team's services are exported only to the clusters that use them](./export-to-a-subset-of-clusters.svg)
```
MrFreezeex, line 370: In the schema you don't show that the service is also imported by its local cluster too, I am not sure if this is too much/unnecessarily complex to add there though ~

(BTW I am curious what do you use to make your diagrams 👀?)

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -299,16 +302,17 @@ Another CRD called `ServiceImport` will be introduced to act as the in-cluster
...
 `ServiceImport` resources will be managed by the MCS implementation's
 mcs-controller.
 
 If multiple clusters export a `Service` with the same namespaced name, they will
 be recognized as a single combined service. For example, if 5 clusters export
 `my-svc.my-ns`, each importing cluster will have one `ServiceImport` named
 `my-svc` in the `my-ns` namespace and it will be associated with endpoints from
-all exporting clusters. Properties of the `ServiceImport` (e.g. ports, topology)
-will be derived from a merger of component `Service` properties.
+all exporting clusters that export to that cluster. Properties of the
```
MrFreezeex: Hmm this paragraph may need slightly more edits, like the first sentence assume that every export will necessarily be combined and this precision look like coming a bit out of the blue within that paragraph 😅

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -393,6 +422,9 @@ Services will not be visible to other clusters in the clusterset by default.
 They must be explicitly marked for export by the user. This allows users to
 decide exactly which services should be visible outside of the local cluster.
 
+A service may be exported to all other clusters in the clusterset or only to
+some of them. Which clusters it is exported to is implementation-defined.
```
MrFreezeex: I am wondering if we should do anything to the term clusterset. For instance, in some of the previous meeting that I was describing that we could connect clusters in essentially a partial mesh @lauralorenz said that this look like essentially multiple clustersets. I don't know if defining the possibility of having multiple clustersets is the right abstraction though here. One other possibility could be to remote that term 🤷‍♂️.

I don't think deciding on what to do in that way  should necessarily be blocking your current PR though ~

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
MrFreezeex: This edited part lgtm since it seems to allow more flexibility for implementations in particular in a (hypothetical) more centralized implementation that would decide for all clusters what service are exported to!

MrFreezeex (commented):
Thanks, everything you added mostly lgtm! I left a few minor comments/questions inline (I also haven't double at everything in the KEP that wasn't directly edited by you but will do a quick check in a future review too 😅).

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Here's how I think about it.

The MCS API itself only needs namespace sameness between clusters that are directly connected through it. Whether the whole graph you get by following those connections is treated as a clusterset doesn't really matter to the API. ClusterSet is a simple design pattern you can follow when you want the overall multi-cluster network to be easy to manage. #2734 proposed something along these lines five years ago.

"Easy to manage" matters to two kinds of people. One builds a platform directly on the MCS API. They can scale by following the ClusterSet pattern, or build a more complex network without it, in which case trust is their own responsibility. That's no different from pushing apps to many clusters with something like ArgoCD, and I think that flexibility is exactly what makes the API useful. The other is a service that builds and offers multi-cluster services on top of MCS, like a KaaS. They can offer the simple ClusterSet shape or allow more complex networks, and there are trade-offs either way. In the complex case the system has to detect on its own which namespaces are the same across clusters. I recently tried drafting a standard API for that: kahirokunn/kubernetes-enhancements#5.

Working on that API made one thing clear to me. If you have a bunch of clusters and namespace sameness is defined across them, do you actually want MCS to connect all of them automatically? No. The scope where namespaces are the same, the scope you want connected on the network, and the scope you distribute apps or jobs to are all different things. What you really need to know isn't sameness, it's the boundary within which DNS must not conflict, and that's only between neighboring clusters that are actually connected. If A-B and B-C are connected but A-C isn't, I think A-C sameness can go either way, and C doesn't need to be able to resolve the services exported from A. Keeping sameness there is just more convenient for management.

A picture of what I mean. One network, and each dashed circle is where the names exported by one cluster resolve:

names-resolved-per-cluster


So where I land:

1. Some people don't want the ClusterSet idea to change, and I think ClusterSet is a
   good pattern anyway. Nothing in this proposal asks them to give it up: a platform
   built around a ClusterSet keeps behaving exactly as it does today.
2. Seeing this as "multiple clustersets" is fine too. What this feature really does is
   let clusters connect as a partial mesh. Every CNCF graduated service mesh supports
   partial mesh, and none of them limit meshing to a ClusterSet-like boundary. So
   ClusterSet is a pattern, and I'd rather not let it limit what service meshes can do
   with this API.

So what I'm after isn't replacing ClusterSet, it's making room next to it: keep
ClusterSet as the recommended shape, and state the namespace sameness requirement in
terms of directly connected clusters rather than the whole set. Platforms that follow
the pattern lose nothing and read the spec the same way they do now. Platforms that
need a partial mesh get a way to be conformant instead of being outside the spec.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Thanks, fixed.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Good catch, I had missed that. Fixed. The SVG is hand-written, I tried to match the style of the KEP-4322 diagrams.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Right, there are the two patterns, each cluster deciding on its own or a central control plane deciding for all of them. Fixed.

kahirokunn:
I've pushed a third commit that lets an implementation define a clusterset per service, with one clusterset for all services staying the default.

For context, this touches the question raised in #2734 back in 2021. JeremyOT had two concerns there:

> We don't want to remove ClusterSet from the spec. MCS is defined based on ClusterSets and I don't think we want to change that. What we do want to say is that it is valid to implement the MCS API outside of ClusterSets, but that not all features may work as expected and you'll need to understand what that means for your chosen implementation - e.g. does namespace sameness only apply selectively? Do all clusters receive the import?
> https://github.com/kubernetes/enhancements/pull/2734#discussion_r638979221

> I'm concerned about how this weakens assumptions around this API and puts more on the consumer to figure out how each implementation works
> https://github.com/kubernetes/enhancements/pull/2734#discussion_r639337673

The third commit takes the route JeremyOT suggested there: the clusterset stays the unit, and an implementation may define one per service, as the clusters that export it or that it is exported to, with namespace sameness applying to that service's namespace inside it. That answers his two questions directly: sameness applies within each service's clusterset, and a cluster imports a service only if it is exported to it. On the consumer side there is nothing to figure out per implementation, because `ServiceExport.status.clusters` and `ServiceImport.status.clusters` show the members. With one clusterset for all services nothing changes.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
MrFreezeex: Thanks for the explanation!

I think I definitely agree that a ClusterSet or said differently having a complete mesh is the right "default" given that's simpler and it's also even our stance in Cilium (in the Cluster Mesh setup doc we kind of assume that every clusters are connected and only mentions partial mesh in advanced use cases).

What I don't know if the term (and not the concept behind it) is still relevant given that:
- we would now allow to export a service to an arbitrary set of clusters (given some implementations specific rules/settings etc)
- the DNS domain is still clusterset.local (and it might be difficult/impossible to change that) while we could expose backends from a smaller set of clusters than what the ClusterSet would "officially" be
- the clusterset ClusterProperty is to my knowledge not used meaningfully by any MCS implementation
- We could potentially be describing essentially a mesh of clusters instead of describing a ClusterSet or a partial mesh and/or redefining ClusterSet as what you described as DNS zone in your schema

kahirokunn: Yeah, I've been going back and forth on this too. I ended up taking [JeremyOT's idea from #2734](https://github.com/kubernetes/enhancements/pull/2734#pullrequestreview-668523397): keep ClusterSet, but let an implementation define one per service. So the clusters a service is exported to, plus the ones exporting it, are its clusterset, and sameness only has to hold for that service's namespace inside it. Everything else in the doc can keep saying "clusterset", which also made the diff a lot smaller.

I think that covers most of your list. clusterset.local really is the zone of that clusterset now. Full mesh is still the default, we only rule out requiring a single clusterset per cluster. So I went with the "multiple clustersets" reading from my earlier comment and put it in the definition. Does that work for you?

On the ClusterProperty, I agree we could probably live without it. But that's KEP-2149 territory, so I'd rather leave it alone here and come back to it once this one lands.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -393,15 +430,29 @@ Services will not be visible to other clusters in the clusterset by default.
...
 decide exactly which services should be visible outside of the local cluster.
 
+A service may be exported to any set of other clusters. Which clusters it is
+exported to is implementation-defined. The first figure shows the same exports
+with one clusterset for all services and with a clusterset per service. In the
+second case, each dashed outline encloses the `ServiceExport`s and
+`ServiceImport`s of the services that share a clusterset, and the clusters it
+spans are that clusterset.
+
+![four clusters and three exported services, shown as one clusterset and as a clusterset per service](./one-clusterset-or-one-per-service.svg)
```
MrFreezeex, line 440: I think there might be a problem with this schema since besides the dashed line the ServiceExport/ServiceImport are the same in both "sub-figures"?

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -765,21 +847,34 @@ section provides an overview of the multicluster DNS specification and its
...
+A and by D both resolve in every member.
+
+![six clusters in one clusterset where the names exported by two of them resolve in every member](./names-resolved-in-a-clusterset.svg)
+
+Otherwise the clusters that resolve a name may differ from service to service.
+In the second figure, each dashed circle is the clusterset of the services
+exported by one cluster. G and L are in other clusters' clustersets as well as
+their own.
+
+![sixteen clusters with seven overlapping clustersets, each the clusterset of the services exported by one cluster](./names-resolved-per-cluster.svg)
```
MrFreezeex, line 876: nit: I am not entirely certain that those schema and some of the explanation should be here since this also roughly describe the existence of the ServiceImport and what contributes to it (properties and backends). Those are also described elsewhere so that's not really a big deal IMO.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -765,21 +847,34 @@ section provides an overview of the multicluster DNS specification and its
...
+spread across all endpoints imported by that cluster.
   * **Headless services:** Within an importing cluster, the clusterset domain
 name will have multiple `A`/`AAAA` records, each containing the address of a
 ready endpoint of the headless service. `<service>.<ns>.svc.clusterset.local`
 will resolve to the entire set or the subset of ready pod IPs, depending on the
 implementation and endpoint count.
 
+A cluster resolves only the names of the services it exports or that are
+exported to it. When every service has the same clusterset, a name resolves in
+every member that has its namespace. In the first figure, the names exported by
```
MrFreezeex, line 866: I am not sure that we condition anywhere else that the namespace should exist for the name resolution to work (I do agree that's what most/all implementation do in practice though)

MrFreezeex (commented):
Thanks Kahiro, I find amazing the latest version. I think you particularly nailed it on the ClusterSet per service term and all the amazing schemas! 🎉 

I left a few minor comments inline but LGTM overall!

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
MrFreezeex: Yeah that looks amazing, thanks a lot!

And yes no problem with the ClusterProperty point, we discussed last week in the US/EU meeting about this KEP in general (not particularly specifically about the clusterset property) and there would be quite a bit of work there to actualize it anyway :sweat_smile:.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Thanks for the review!

Without the namespace, we can’t create the ServiceImport, so the cluster can’t import the service and therefore can’t resolve its MCS DNS name. That’s why the namespace is required, and I believe this was already covered in the “Importing Services” section.

MrFreezeex: Yes that's the case in all existing implementations and indeed the section you mentioned has kind of implied between the line that a bit. I am not sure that the existing text in that section could strictly be interpreted as if you don't have a ServiceImport the DNS resolution must not work.

Your latest fixup commit seems to be resolving this perfectly though!

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Thanks for the review. Both figures are really about which clusters import, so I moved them up to `Importing Services`, which already describes the `ServiceImport`. The DNS section just has one sentence left saying only the clusters that import a service resolve its clusterset domain name.

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
kahirokunn: Thanks for the review. What I wanted the figure to show is that the same clusters and the same `ServiceExport`s and `ServiceImport`s can be grouped into clustersets in more than one way, all of them in one clusterset or split per service, and I wanted that to come across from the figure itself, not just the text. It probably did not, so I redrew the top half with the same dashed outline as the bottom one and reworded the labels. Does that read better?

MrFreezeex: Hmm what I don't necessarily get I think is if there's one clusterset in the top half, why each ServiceImport is not on every clusters given that we consider only one ClusterSet for everything? If the ClusterSet relevant to all the services are all the clusters, then there should be some ServiceImport everywhere right? I might have missed something :thinking:

kahirokunn: You're right, I had it wrong. Thanks for pointing it out. I've pushed a new version of the figure. How does it look? If it's fine I'll autosquash.

MrFreezeex: That looks perfect now, thanks!

kahirokunn: Thank you for your review!

MrFreezeex (approved):

keps/sig-multicluster/1645-multi-cluster-services-api/README.md
```diff
@@ -299,13 +305,14 @@ Another CRD called `ServiceImport` will be introduced to act as the in-cluster
 representation of a multi-cluster service in each importing cluster. This is
 analogous to the traditional `Service` type in Kubernetes. Importing clusters
 will have a corresponding `ServiceImport` for each uniquely named `Service` that
-has been exported within the clusterset, referenced by namespaced name.
+they export or that has been exported to them, referenced by namespaced name.
```
lauralorenz:
```suggestion
has been exported to them, referenced by namespaced name.
```

PR #6332, opened 2026-09-08
KEP-6007: add NUMA utilization-aware topology manager policy options

swatisehgal:
- One-line PR description: KEP-6007: NUMA utilization-aware topology manager policy options

- Issue link: https://github.com/kubernetes/enhancements/issues/6007

- Other comments: Introduces preferAllocatedNUMANodes policy option (none, most-allocated, least-allocated) and providerScoreWeights for the Topology Manager, enabling allocation-aware NUMA node selection. Builds on the policy options framework from KEP-3545.

### DIFF
--- a/keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
+++ b/keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
@@ -1,0 +1,751 @@
+# KEP-6007: NUMA Utilization-Aware Topology Manager Policy Options
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Balanced NUMA Utilization on Multi-Die Processors](#story-1-balanced-numa-utilization-on-multi-die-processors)
+    - [Story 2: Workload Consolidation for Power Efficiency](#story-2-workload-consolidation-for-power-efficiency)
+    - [Story 3: Database Spreading](#story-3-database-spreading)
+    - [Story 4: GPU Workload Prioritization](#story-4-gpu-workload-prioritization)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Topology Hint Score Plumbing](#topology-hint-score-plumbing)
+  - [Allocation-Aware Policy Options](#allocation-aware-policy-options)
+  - [Score-Aware Preferred-First Merge Optimization (Beta)](#score-aware-preferred-first-merge-optimization-beta)
+  - [Per-Resource Weights](#per-resource-weights)
+  - [Kubelet Configuration](#kubelet-configuration)
+    - [Example Configurations](#example-configurations)
+  - [Feature Gate](#feature-gate)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Alpha to Beta Graduation](#alpha-to-beta-graduation)
+    - [Beta to GA Graduation](#beta-to-ga-graduation)
+  - [Graduation Criteria of Options](#graduation-criteria-of-options)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Alternative 1: Scheduler-Based NUMA Balancing](#alternative-1-scheduler-based-numa-balancing)
+  - [Alternative 2: Pod-Level Annotations](#alternative-2-pod-level-annotations)
+  - [Alternative 3: Static NUMA Assignment](#alternative-3-static-numa-assignment)
+- [Infrastructure Needed (Optional)](#infrastructure-needed-optional)
+<!-- /toc -->
+
+## Release Signoff Checklist
+
+Items marked with (R) are required *prior to targeting to a milestone / release*.
+
+- [ ] (R) Enhancement issue in release milestone, which links to KEP dir in [kubernetes/enhancements] (not the initial KEP PR)
+- [ ] (R) KEP approvers have approved the KEP status as `implementable`
+- [ ] (R) Design details are appropriately documented
+- [ ] (R) Test plan is in place, giving consideration to SIG Architecture and SIG Testing input (including test refactors)
+  - [ ] e2e Tests for all Beta API Operations (endpoints)
+  - [ ] (R) Ensure GA e2e tests meet requirements for [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md)
+  - [ ] (R) Minimum Two Week Window for GA e2e tests to prove flake free
+- [ ] (R) Graduation criteria is in place
+  - [ ] (R) [all GA Endpoints](https://github.com/kubernetes/community/pull/1806) must be hit by [Conformance Tests](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/conformance-tests.md) within one minor version of promotion to GA
+- [ ] (R) Production readiness review completed
+- [ ] (R) Production readiness review approved
+- [ ] "Implementation History" section is up-to-date for milestone
+- [ ] User-facing documentation has been created in [kubernetes/website], for publication to [kubernetes.io]
+- [ ] Supporting documentation, e.g., additional design documents, links to mailing list discussions/SIG meetings, relevant PRs/issues, release notes
+
+[kubernetes.io]: https://kubernetes.io/
+[kubernetes/enhancements]: https://git.k8s.io/enhancements
+[kubernetes/kubernetes]: https://git.k8s.io/kubernetes
+[kubernetes/website]: https://git.k8s.io/website
+
+## Summary
+
+Extend the Topology Manager with a `preferAllocatedNUMANodes` policy option
+that controls NUMA node selection based on current allocation state. Valid
+values are `none` (default - preserves existing behavior), `most-allocated`
+(packing), and `least-allocated` (spreading).
+
+These new policy options build on the Topology Manager policy options framework
+introduced in [KEP-3545: Improved multi-numa alignment in Topology Manager](/keps/sig-node/3545-improved-multi-numa-alignment).
+Like `prefer-closest-numa-nodes` from KEP-3545, the new options modify the
+behavior of existing Topology Manager policies (`best-effort`, `restricted`,
+`single-numa-node`) without introducing new policies.
+
+## Motivation
+
+Modern multi-die CPU architectures expose multiple NUMA domains per socket
+(e.g., three compute dies per processor with Sub-NUMA Clustering enabled). For
+high-performance workloads such as packet-core network functions, each pod
+requires all of its allocated CPUs on a single NUMA node to maintain locality,
+while pod allocations must be balanced across the available NUMA nodes.
+
+The Topology Manager currently selects NUMA nodes based purely on structural
+properties: affinity mask width (narrowest) or NUMA distance (closest, via
+KEP-3545). It has no awareness of how heavily each NUMA node is already
+allocated. The current allocation algorithm packs workloads onto the first
+NUMA node with sufficient available resources (lowest-ID tiebreak), resulting
+in unbalanced NUMA utilization (for example, four pods on NUMA 0 and two on
+NUMA 1) even when equivalent capacity exists elsewhere. This has been
+identified in [kubernetes/kubernetes#125453](https://github.com/kubernetes/kubernetes/issues/125453).
+
+This creates two classes of problems:
+
+- **NUMA imbalance on multi-die processors:** On CPUs with 3+ NUMA domains per
+  socket, the lowest-ID tiebreak concentrates workloads on one compute die
+  while others sit idle, causing resource contention, turbo frequency
+  imbalance, and degraded performance for latency-sensitive applications.
+- **Fragmentation:** Workloads spread across partially-used NUMA nodes, leaving
+  no fully-free nodes available for large allocations that require full NUMA
+  locality.
+
+The closest existing option (`distribute-cpus-across-numa`) addresses a
+different use case: it spreads a single pod's CPUs across multiple NUMA nodes,
+which is effectively the opposite of the desired behavior.
+
+Operators today have no mechanism within the Topology Manager to influence NUMA
+node selection based on allocation state. This KEP addresses that gap.
+
+### Goals
+
+- Introduce a `preferAllocatedNUMANodes` policy option for the Topology Manager
+  that selects NUMA nodes based on current allocation state, with values
+  `none` (default, no preference, preserves existing behavior),
+  `most-allocated` (packing), and `least-allocated` (spreading).
+- Support per-resource weighting via `providerScoreWeights` so operators can
+  prioritize CPU, memory, or specific device types when computing utilization
+  scores. Weights are specified as a comma-separated string of `resource=weight`
+  pairs (e.g., `"cpu=0.3,memory=0.1,nvidia.com/gpu=0.6"`).
+- Maintain existing topology guarantee semantics. The new options only
+  influence NUMA node selection among equally valid candidates and do not
+  change which hints are considered preferred.
+- Ensure no user-visible behavior change when the feature is disabled.
+
+### Non-Goals
+
+- Runtime resource utilization monitoring (e.g., actual CPU usage percent).
+  This feature uses allocation state only.
+- Cross-node balancing or cluster-wide scheduling decisions. These remain the
+  scheduler's responsibility.
+- Device-specific utilization metrics (e.g., GPU memory usage). This feature
+  uses simple device count for now.
+- Automatic weight tuning or workload profiling.
+
+## Proposal
+
+Add a `Score` field to `TopologyHint` that captures per-NUMA-node allocation
+utilization. Hint providers (CPU manager, memory manager, device manager) each
+compute a score representing how heavily a NUMA node set is allocated. The
+Topology Manager then uses these scores, combined with new policy options, to
+prefer either the most-allocated or least-allocated NUMA nodes.
+
+The design has three parts: score plumbing in `TopologyHint` and hint
+providers, allocation-aware policy options, and per-resource weight support
+via `providerScoreWeights`. A score-aware merge optimization is planned for
+beta to improve performance on systems with many NUMA nodes.
+
+### User Stories
+
+#### Story 1: Balanced NUMA Utilization on Multi-Die Processors
+
+As a cluster operator running latency-sensitive network functions on processors
+with multiple compute dies (e.g., 3 or more NUMA domains per socket using
+sub-NUMA clustering), I want pod allocations to spread equally across all NUMA
+nodes so that no single compute die is overloaded while others sit idle.
+
+Today, each pod requires all of its CPUs on a single NUMA node for performance
+locality, but the Topology Manager's lowest-ID tiebreak concentrates pods onto
+NUMA node 0 until it is exhausted. On multi-die architectures this creates
+severe imbalance: one compute die runs near capacity while adjacent dies remain
+empty, leading to uneven turbo frequency behavior, thermal throttling, and
+wasted compute resources. The existing `distribute-cpus-across-numa` option
+addresses the opposite problem: it spreads a single pod's CPUs across multiple
+NUMA nodes, which breaks the single-NUMA-node locality these workloads require.
+
+With `preferAllocatedNUMANodes: "least-allocated"`, the Topology Manager would
+select the least-utilized NUMA node for each new pod, naturally balancing
+allocation across all compute dies while preserving per-pod NUMA locality.
+
+See also: kubernetes/kubernetes#125453.
+
+#### Story 2: Workload Consolidation for Power Efficiency
+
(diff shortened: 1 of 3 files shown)


swatisehgal:
/cc @ffromani @kad @pravk03 @jingczhang

PTAL, the KEP is ready for review.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+The Topology Manager computes an equal-weight average during hint merge:
+`aggregatedScore = sum(scores) / count`.
+
+Score is populated by all hint providers but does not influence hint selection
+on its own. It is only used when `preferAllocatedNUMANodes` is set to
+`most-allocated` or `least-allocated`.
+
+### Allocation-Aware Policy Options
+
+Add the `preferAllocatedNUMANodes` policy option (with `none`,
```
jingczhang: Should follow the existing kebab-case convention instead of introducing camelCase? suggest using "numa-allocation-strategy" to be explicit and to avoid overloading the "prefer".

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+### Kubelet Configuration
+
+Both new policy options are string values in the existing
+`TopologyManagerPolicyOptions` (`map[string]string`). No type change is
+needed.
+
+| Key | Value | Description |
+|-----|-------|-------------|
+| `preferAllocatedNUMANodes` | `"none"`, `"most-allocated"`, `"least-allocated"` | Controls NUMA node selection based on allocation state |
+| `providerScoreWeights` | `"resource=weight,..."` (e.g., `"cpu=0.3,nvidia.com/gpu=0.6"`) | Comma-separated resource=weight pairs for weighted score aggregation |
```
jingczhang: Same comment on keeping kebab-case instead of providerScoreWeights. Suggest using "numa-score-weights" or similar.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+### Per-Resource Weights
+
+Add `ProviderName` field to track which provider generated each hint:
+
+```go
+type TopologyHint struct {
+    NUMANodeAffinity bitmask.BitMask
+    Preferred        bool
+    Score            int64
+    ProviderName     string  // "cpu", "memory", "nvidia.com/gpu", etc.
```
jingczhang: Adding ProviderName to TopologyHint breaks abstraction and duplicates information. Each entry in providersHints is a map[string][]TopologyHint where the key is the resource name ("cpu", "memory", "nvidia.com/gpu", etc.). The Topology Manager already knows which provider/resource produced which hints from this outer structure. Shouldn't this be enough?

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+`most-allocated`, and `least-allocated` values) that makes Score a first-class
+selection criterion. When set to `none` (the default), Score is ignored and
+existing Narrowest/Closest behavior is preserved.
+
+Updated comparison priority order:
+
+1. Preferred flag (topology constraint satisfaction, unchanged)
+2. Score-based preference (if policy option is set, new)
+3. Affinity mask comparison: Narrowest or Closest (structural tiebreak, unchanged)
+
```
jingczhang, line 308: Should structural tiebreaks come before the score-based scoring? That way score only breaks ties among structurally equivalent candidates (same distance, same width), preserving the guarantees users opted into with prefer-closest-numa-nodes.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+- **CPU manager**: `Score = max(1, assignedExclusiveCPUs * 100 / allocatableCPUs)`
+- **Memory manager**: `Score = max(1, assignedBytes * 100 / allocatableBytes)`
+- **Device manager**: `Score = max(1, allocatedDeviceCount * 100 / allocatableDeviceCount)`
+
+The Topology Manager computes an equal-weight average during hint merge:
+`aggregatedScore = sum(scores) / count`.
+
+Score is populated by all hint providers but does not influence hint selection
+on its own. It is only used when `preferAllocatedNUMANodes` is set to
+`most-allocated` or `least-allocated`.
```
pravk03, line 290: With `topologyManagerScope=container`, hints are merged per container, so would we susceptible to more intra pod spreading with `least-allocated` ?.  

Not sure if this is a concern, but it might be good to document the behavior under different topologyManagerScope settings.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,751 @@
...
+  providerScoreWeights: "cpu=0.3,memory=0.1,nvidia.com/gpu=0.6"
+```
+
+This keeps `TopologyManagerPolicyOptions` as `map[string]string`, consistent
+with all existing policy options. The string is parsed at kubelet startup into
+a `map[string]float64` for internal use.
+
+The Topology Manager uses the parsed weights to compute a weighted average of
+provider scores during hint merge. Providers not listed in the weights string
+use a default weight of 1.0, making weights optional and forward-compatible
```
pravk03: nit: Should this be 0 ?. Defaulting to 1 would give an unspecified resource the highest priority (in the above example) ?

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Sure, will update.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: I had added ProviderName to TopologyHint for weighted score aggregation, where we need to look up weights[resourceName], this information is discarded in filterProvidersHints before reaching mergePermutation. But, I do understand that adding ProviderName duplicates information already in the map[string][]TopologyHint keys and breaks abstraction. 

We can explore the alternative of threading the resource names through filterProvidersHints as a parallel []string alongside the flattened [][]TopologyHint. I can update the KEP to remove this addition of ProviderName and try to preserve resource names in the merge pipeline rather than embedding in every hint.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Ah right, will update. thanks!

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Thanks for raising this point! The intent behind 1.0 was so that all providers weighted equally when no weights are explicitly specified but you spotted correctly that it breaks down as soon as weights are written as fractions of a whole (e.g. in "cpu=0.3,memory=0.1,nvidia.com/gpu=0.6") an unlisted intel.com/sriov-nic at 1.0 takes a larger share than the three configured resources combined.

I will update the KEP to have two cases:

1. if all the weights are unset:  every provider gets weight 1.
2. if some weights are set: only named resources contribute and unlisted ones get 0.

Please let me know if this addresses your concern.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Yes, under least-allocated containers of a pod will tend to land on different NUMA nodes. That's the option doing what it advertises. Container scope doesn't promise that containers of a pod share a NUMA node, the co-location we see today is incidental not by design. So least-allocated isn't removing a guarantee, it's making an existing non-guarantee show up more often.

I do agree that this needs to be captured and will update the KEP to capture this.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
jingczhang: The issue comes from the current design allows fractional weights (e.g., "cpu=0.3,memory=0.1,nvidia.com/gpu=0.6"), which means explicit weights can fall below the default of 1.0, causing unlisted resources to silently outweigh listed ones. A simpler and better fix is to align with kube-scheduler's NodeResourcesFit scoring: use integer weights with a minimum of 1 (default), e.g., "cpu=3,memory=1,nvidia.com/gpu=6". Since the default weight (1) equals the minimum, unlisted resources can never dominate explicitly configured ones.

swatisehgal: Thanks @jingczhang, I like that! This makes things much simpler and we can default to 1 for non specified resources.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
pravk03: Thanks. I bring this up because `container` scope is the default and this behavior may not be very obvious to the user. One way to avoid such spreading (and keep the behavior consistent with `none`, `most-allocated`) is to exclude the current pod's allocation in score calculation. From the use-cases I can think of, spreading containers of the same pod between numa nodes may be undesirable for workload performance, but I also understand there could be other scenario's where the same behavior may be desired. So, I am ok with your decision on this.

pravk03: Thanks for documenting this. 

Just one observation though —  since container scope is the default, users picking least-allocated might not be familiar with the internal details and unexpectedly see intra-pod spreading.  The alternative could be to ignore the current pod's allocation while computing the score, so that all containers within the pod get the same score. This would keep the same intra pod allocation behavior with `none`, `most-allocated`. 
Either approach has its trade-offs, so definitely fine with keeping it as-is.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+- Introduce a `numa-allocation-strategy` policy option for the Topology Manager
+  that selects NUMA nodes based on current allocation state, with values
+  `none` (default, no preference, preserves existing behavior),
+  `most-allocated` (packing), and `least-allocated` (spreading).
+- Support per-resource weighting via `numa-score-weights` so operators can
+  prioritize CPU, memory, or specific device types when computing utilization
+  scores. Weights are specified as a comma-separated string of `resource=weight`
+  pairs (e.g., `"cpu=3,memory=1,nvidia.com/gpu=6"`), where each weight is an
+  integer in the range [0, 100]. Resources the string does not name default to a
```
ffromani, line 136: nit: `Resources the string does not name` reads a bit clunky, maybe worth a rephrasing?

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+to existing tests to make this code solid enough prior to committing the
+changes necessary to implement this enhancement.
+
+##### Prerequisite testing updates
+
+None. Existing Topology Manager test infrastructure is sufficient.
+
+##### Unit tests
+
+- `k8s.io/kubernetes/pkg/kubelet/cm/topologymanager`: `<date>` - `<coverage>`
```
ffromani, line 586: you need to fill this before the approval

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+  `least-allocated`)
+- Feature gate on/off behavior
+
+##### Integration tests
+
+Integration tests will cover:
+- Multi-provider hint merge with scores
+- Policy behavior with feature gate enabled and disabled
+- Weight normalization end-to-end
+- Interaction with `prefer-closest-numa-nodes` from KEP-3545
```
ffromani, line 617: how do we plan to implement integration test? I mean, which k/k files will we change, or introduce?

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+Kubelet may fail to start if invalid policy option values are provided (e.g.,
+an unrecognized value for `numa-allocation-strategy`).
+Already running workloads are not affected. The feature only influences new
+pod admissions.
+
+###### What specific metrics should inform a rollback?
+
+An increase in topology admission errors or unexpected pod placement patterns
+(e.g., pods not landing on expected NUMA nodes) should prompt investigation.
```
ffromani, line 755: correct, but how can we help users identify these patterns?

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+###### Is the rollout accompanied by any deprecations and/or removals of features, APIs, fields of API types, flags, etc.?
+
+No.
+
+### Monitoring Requirements
+
+###### How can an operator determine if the feature is in use by workloads?
+
+Inspect the kubelet configuration of the nodes: check the feature gate status
+and the `numa-allocation-strategy` value in `topologyManagerPolicyOptions`.
```
ffromani, line 770: I wonder if it makes sense to introduce a metric (gauge?) which counts how many time the scoring took effect and influenced placement. I'd value the input of PRR folks here.

ffromani: another idea: publish metrics (gauges?) about per-resource current scores

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+###### How can someone using this feature know that it is working for their instance?
+
+- [ ] Other (treat as last resort)
+  - Details: Launch a pod requiring resources from a specific NUMA node on a
+    node with known asymmetric allocation. Verify via `taskset -cp 1` and
+    `numactl -H` inside the container that resources are assigned from the
+    expected NUMA node (most-allocated or least-allocated depending on the
+    configured option).
```
ffromani, line 779: do we have metrics which track the per-numa container count? would it be a worthy addition? it will help, I believe, in this case.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+This existing metric captures the time spent in Topology Manager admission.
+Any regression introduced by score computation would be visible as an increase
+in this metric.
+
+###### Are there any missing metrics that would be useful to have to improve observability of this feature?
+
+A metric tracking the score-based selection outcome (e.g., which NUMA node was
+selected and its score) would be useful for debugging and can be considered
+for beta.
```
ffromani, line 800: I think we will need _a_ metric for alpha to help users confirm the feature is operative. I'd be fine (process allowing) deferring more metrics and/or refining the metrics in beta.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+###### Will enabling / using this feature result in increasing time taken by any operations covered by existing SLIs/SLOs?
+
+No. The merge optimization (`mergePreferred()`) is expected to reduce hint
+merge time for common cases. Score computation adds O(n) overhead where n is
+the number of providers (typically 2-4), which is negligible.
+
+###### Will enabling / using this feature result in non-negligible increase of resource usage (CPU, RAM, disk, IO, ...) in any components?
+
+No. The additional Score field in `TopologyHint` adds 8 bytes per hint. Score
```
ffromani, line 835: and we have a hint count which is O(device_plugins)+N, so well under control and predictable.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+  - Testing: Unit tests cover weight validation.
+
+- Containers of a pod land on different NUMA nodes under `least-allocated`:
+  - Detection: Expected behavior under `topologyManagerScope: "container"`
+    rather than a defect. See
+    [Interaction with Topology Manager Scope](#interaction-with-topology-manager-scope).
+  - Mitigations: See
+    [Preserving Current Placement Behavior](#preserving-current-placement-behavior).
+  - Diagnostics: Compare `topologyManagerScope` in the kubelet configuration
+    against the pod's observed CPU and memory affinity.
```
ffromani, line 873: Likewise, I think we should identify _a_ metric to help here

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+
+## Implementation History
+
+- 2026-09-08: KEP created
+
+## Drawbacks
+
+Adds complexity to the Topology Manager hint comparison and merge logic.
+However, the additional complexity is gated behind opt-in policy options and
+has no effect on the default code path.
```
ffromani, line 888: the change is however well contained and well defined even at projected GA stage.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,920 @@
...
+**Cons:** Requires scheduler-kubelet protocol changes; breaks existing Topology
+Manager design; significantly larger scope.
+
+### Alternative 2: Pod-Level Annotations
+
+Let users specify packing/spreading preference per pod via annotations.
+
+**Pros:** More flexible, users opt-in per workload.
+**Cons:** Increases configuration burden; hard to enforce cluster-wide
+policies; inconsistent with the existing node-level Topology Manager model.
```
ffromani, line 906: and validations are an API side channel

ffromani (commented):
initial pass; previous reviewers already helped tightening the spec. I spotted only minor things so far

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: After reviewing the precedent from other kubelet features, I've removed the integration tests section as they are not typically required for topology manager policy options. We can follow the same pattern with comprehensive unit tests and e2e tests. I am open to discussion here, if you feel that we need intergation tests, we can work on adding those at Beta stage.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Great point. I've will add observability metrics to the KEP.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: I've will add `topology_manager_numa_score_selection_total` as a **counter** (not gauge, since it's a cumulative count) that increments whenever the score-based selection chooses a different NUMA node than the pre-existing narrowest/closest tiebreak would have chosen. This directly measures the impact of the feature.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: I've will add `topology_manager_numa_container_count` as a gauge that tracks the number of containers allocated on each NUMA node. I was thinking of deferring this to Beta if that's okay but if you feel strongly we can handle it at alpha stage itself.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,1004 @@
...
+node. `least-allocated` does not remove a guarantee; it surfaces an existing
+non-guarantee earlier and more often.
+
+Score quantization further limits how often this arises. Scores are integers in
+[1,100] computed as `assigned * 100 / allocatable`, so on a node with 128 CPUs
+per NUMA node a single-CPU container moves the score by 0.78, which truncates
+to 0. When the score does not change, the comparison falls through to the
+existing structural tiebreak and containers are placed as they are today.
+Spreading under `container` scope therefore only takes effect once a container
+is large enough relative to the NUMA node to move the score.
```
ffromani, line 398: Fine. We would need user examples about how different weights influence the placement decision when we will document this work. The examples you added below should be a nice start, we will probably just need to expand and to turn them into full user stories.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
```diff
@@ -0,0 +1,1004 @@
...
+  node it lands on, which attracts the next container to the same node. This is
+  a stronger form of co-location than today's tiebreak provides.
+- **Set `topologyManagerScope: "pod"`.** This guarantees a common affinity
+  across all containers, which is stricter than anything `container` scope
+  offers today. It is not a drop-in substitution: pod scope admits a pod only
+  if all of its containers can achieve a *common* alignment, so some pods that
+  are admitted under container scope today will be rejected. That tradeoff is
+  pre-existing KEP-693 behavior and is not introduced by this KEP. All example
+  configurations in this KEP use pod scope.
+- **Leave `numa-allocation-strategy` as `none`.** Placement is then unchanged.
```
ffromani: nit: this should be the first option as it's far the easier and guarantee the identical behavior wrt today

ffromani (commented):
LGTM from sig-node/technical content. Some editorial suggestions inside.
All my comments are addressed nicely and from my side we're ready for alpha stage. The beta progression is also nicely outlined.
Ping me if you want to formal label.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: I've expanded the effective-weight table into a new subsection (Worked Examples: How Weights Influence Placement) under Per-Resource Weights, and added the matching user story. Hopefully this addreses your review comment.

keps/sig-node/6007-numa-utilization-aware-topology-manager-policy-options/README.md
swatisehgal: Done!

swatisehgal:
> LGTM from sig-node/technical content. Some editorial suggestions inside. All my comments are addressed nicely and from my side we're ready for alpha stage. The beta progression is also nicely outlined. 

Addressed the last set of minor comments you had as well, Thanks for your review!

> Ping me if you want to formal label.

PTAL and if you are happy enough, please add the formal label.

ffromani (commented):
/lgtm

thanks for the quick followup

swatisehgal:
/assign @johnbelamaric 
for production readiness review

PR #6336, opened 2026-09-09
KEP-4939: update for beta graduation

amritansh1502:
- One-line PR description: Bump KEP-4939 (TLS support in gRPC container probes) to beta targeting v1.38.

- Issue link:https://github.com/kubernetes/enhancements/issues/4939

- Other comments:

### DIFF
--- a/keps/sig-node/4939-grpc-probe-with-tls/README.md
+++ b/keps/sig-node/4939-grpc-probe-with-tls/README.md
@@ -204,23 +204,29 @@ When `GRPCContainerProbeTLS` is **disabled**:
 
 - The `mode` field is **silently dropped** from new and updated pods by
   `dropDisabledGRPCContainerProbeTLS` during the strategy phase (PrepareForCreate /
-  PrepareForUpdate). Users do not see an error, the pod is created without the field.
-- Validation acts as a defense-in-depth safety net: if the field somehow survives
-  the drop (e.g., due to a code bug), validation rejects the pod with a `Forbidden`
-  error. In normal operation this path is never hit.
-- If an existing pod already has `mode` set (created while the gate was enabled),
-  the field is **preserved** in etcd for backward compatibility so that
-  read-modify-write cycles do not lose data.
+  PrepareForUpdate). The pod is created without the field, no error is shown.
+  Validation (`validateGRPCAction`) is not gate-aware; it only rejects
+  unsupported enum values (e.g. `"Verify"`). Gating is enforced by the drop
+  step, not validation.
+- If an object (e.g. a Pod or Deployment) already has `mode` set, the field is
+  **preserved** across further updates to that same object.
 
 ### Kubelet Probe Execution
 
-In `pkg/kubelet/prober/prober.go`, the kubelet reads the `Mode` field:
+In `pkg/kubelet/prober/prober.go`, the kubelet checks the feature gate live,
+on every probe execution, in addition to reading the `Mode` field:
 
 ```go
-useTLS := p.GRPC.Mode != nil && *p.GRPC.Mode == v1.GRPCProbeModeTLS
+useTLS := utilfeature.DefaultFeatureGate.Enabled(features.GRPCContainerProbeTLS) &&
+    p.GRPC.Mode != nil && *p.GRPC.Mode == v1.GRPCProbeModeTLS
 ```
 
-This boolean is passed to the gRPC prober.
+This boolean is passed to the gRPC prober. Feature gates are read once at
+kubelet startup, not hot-reloaded, so a gate change takes effect on the next
+kubelet restart. Once that restart happens, it applies immediately to every
+already-running pod on that node, not just newly created ones. See
+[Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy) for tested
+implications.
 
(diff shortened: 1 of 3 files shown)


amritansh1502:
- Set up a live kind cluster with custom-built kubelet/kube-apiserver from the feature branch, verified mode: TLS actually works end-to-end (probe succeeds against a TLS backend, fails correctly against a plaintext one as a negative control).
- Tested disabling the gate on a running pod: confirmed the pod is unaffected until the kubelet on its node restarts, then it immediately fails probes. Also confirmed re-enabling recovers it.
- While testing the "disable gate, then trigger a new ReplicaSet" path, found a real bug: the Deployment controller enters an infinite ReplicaSet-creation loop when a gated field like mode is preserved on the Deployment but gets stripped from every freshly-created ReplicaSet. Confirmed it's not specific to our KEP by reproducing the same thing with the unrelated, already-Beta ContainerRestartRules feature. Filed as kubernetes/kubernetes#141961


cc @SergeyKanzhelev

keps/sig-node/4939-grpc-probe-with-tls/README.md
```diff
@@ -292,8 +312,13 @@ All tests are gated by `framework.WithFeatureGate(features.GRPCContainerProbeTLS
 
 #### Beta
 
-- No major bugs reported during alpha
-- Gather feedback from users
+- `GRPCContainerProbeTLS` feature gate defaults to `true`.
+- Integration tests added covering create/drop/validation/preserve-on-update
```
kannon92, line 316: Tests should really be in alpha.

keps/sig-node/4939-grpc-probe-with-tls/README.md
```diff
@@ -292,8 +312,13 @@ All tests are gated by `framework.WithFeatureGate(features.GRPCContainerProbeTLS
...
+  behavior against a real kube-apiserver and etcd.
+- e2e tests stable with the gate on by default.
```
kannon92, line 318: What e2e tests do you mean?

Can you link to e2e tests running this feature?

keps/sig-node/4939-grpc-probe-with-tls/README.md
```diff
@@ -322,21 +352,22 @@ No data migration is needed. The feature is purely additive and opt-in.
 This feature requires the `GRPCContainerProbeTLS` feature gate on both
 kube-apiserver and kubelet.
 
-- **API server newer than kubelet:** The API server accepts `mode: TLS`,
-  but the older kubelet ignores the field and dials plaintext. TLS-only
-  servers will fail probes, identical to pre-feature behavior.
-- **Kubelet newer than API server:** The older API server drops the `mode`
-  field, so the kubelet never sees it and dials plaintext.
+- **API server newer than kubelet:** An older kubelet predating this KEP has
```
kannon92, line 355: I think this needs to be accounted for with code.
                                                                                                                                                                                                                                                                                 
Please provide and test a skew-safe mechanism before defaulting this on. Possibilities include preventing/scheduling around unsupported nodes, retaining a default-off beta period, or another mechanism that ensures an accepted mode: TLS request cannot be silently executed as       
 plaintext.

keps/sig-node/4939-grpc-probe-with-tls/README.md
kannon92: Looks like https://testgrid.k8s.io/sig-release-master-informing#kind-master-alpha-beta-features is running these tests. 

Can you update this section to show that we are running these tests in CI

keps/sig-node/4939-grpc-probe-with-tls/README.md
kannon92: Since these are integration tests, I'm fine to let this one go to Beta. You do have e2e tests running so this feature is well tested for beta imo.

Integration tests should be added before promoting this feature to beta but its fine to do this in 1.38.

kannon92 (commented):
PRR looks pretty good.

This [comment](https://github.com/kubernetes/enhancements/pull/6336/changes#r4066036205) is my main concern.

SergeyKanzhelev (approved):
/approve

keps/sig-node/4939-grpc-probe-with-tls/README.md
amritansh1502: Kubelet reports which new features it actually supports via Node.status.declaredFeatures, and the scheduler/admission machinery then keep a pod needing mode: TLS from ever running on a node that hasn't declared support so integrate with Node Declared Features now, so mode: TLS structurally can't silently degrade. or   Keep GRPCContainerProbeTLS default-off for this Beta, which option i should choose ?

kannon92: I'm mostly worried about silently degrading TLS settings.

@haircommander @SergeyKanzhelev @mrunalp @mikebrow 

What should the behavior be with probes with tls settings?

amritansh1502: this KEP specifically never sends any credentials. mode: TLS only adds transport encryption with InsecureSkipVerify: true, no client certs, no bearer tokens, no authentication material at all .

PR #6337, opened 2026-09-09
Update KEP-4858 IP/CIDR validation to GA

danwinship:
https://github.com/kubernetes/enhancements/issues/4858

Promotion is mostly a no-op. I even postponed the only real GA graduation criterion (integration test cleanup) from "GA" to "Post-GA Cleanup", since we might as well keep testing it with `SetFeatureGateEmulationVersionDuringTest` for as long as the code is reachable.

/sig network
/assign @aojea @soltysh


soltysh (approved):
/approve
the PRR section

PR #6338, opened 2026-09-09
KEP-5999: update for beta graduation

amritansh1502:
- One-line PR description: Bump KEP-5999 (HTTP/2 cleartext (h2c) container probes) to beta targeting v1.38.

- Issue link: https://github.com/kubernetes/enhancements/issues/5999

- Other comments:

### DIFF
--- a/keps/prod-readiness/sig-node/5999.yaml
+++ b/keps/prod-readiness/sig-node/5999.yaml
@@ -1,3 +1,5 @@
 kep-number: 5999
 alpha:
   approver: "@soltysh"
+beta:
+  approver: "@soltysh"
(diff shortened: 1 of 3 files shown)


amritansh1502:
cc @SergeyKanzhelev

keps/sig-node/5999-h2c-container-probes/README.md
```diff
@@ -494,7 +511,8 @@ make this easier to observe.
 
 ###### Are there any missing metrics that would be useful to have to improve observability of this feature?
 
-
+A `protocol` label on `prober_probe_total` (planned for Beta, see
```
haircommander, line 514: I am a bit concerned about this metric. I didn't realize it had pod_uid, pod and containers, which means it has very high cardinality. adding another field will increase it further. I think we should drop the identifying information and only use it to track the protocol

keps/prod-readiness/sig-node/5999.yaml
```diff
@@ -1,3 +1,5 @@
 kep-number: 5999
 alpha:
-  approver: "@soltysh"
\ No newline at end of file
+  approver: "@soltysh"
+beta:
+  approver: "TBD"
```
haircommander, line 5: choose a PRR approver, they'll rebalance as needed

amritansh1502: added

PR #6341, opened 2026-09-10
KEP-1645: Fix the cluster name in the conflict message example

kahirokunn:
Use `cluster-b` in the conflict message example so it matches the cluster names used elsewhere in the KEP.


MrFreezeex (approved):
/lgtm

PR #6343, opened 2026-09-10
KEP-2172: prepare for alpha in 1.38

Priyankasaggu11929:
- One-line PR description: 

   KEP-2172: prepare for alpha in 1.38

- Issue link: https://github.com/kubernetes/enhancements/issues/2172

- Other comments: 

  k/k code implementation PR - https://github.com/kubernetes/kubernetes/pull/142074

### DIFF
--- a/keps/prod-readiness/sig-node/2172.yaml
+++ b/keps/prod-readiness/sig-node/2172.yaml
@@ -1,0 +1,3 @@
+kep-number: 2172
+alpha:
+  approver: "@soltysh"
--- a/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
+++ b/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
@@ -171,9 +171,16 @@ considered a win, since it means they thought about it and are being explicit
 about it.
 
 Warning events could add up to a lot of noise across a cluster with many
-implicitly-root pods. These Events are throttled to limit volume (one event
-per hour per pod, unless kubelet restarts). The frequency can be reduced
-further if needed.
+implicitly-root pods. To bound this, events will be throttled to at most
+1 event/pod/node/hour. The frequency can be reduced further if needed.
+
+> Note: this throttle limits repeat events for the *same* pod, but each
+> new pod gets its own fresh hour budget. So a workload that creates pods
+> repeatedly (for example, a ReplicaSet recreating replicas) still gets
+> one event per new pod, not one per hour overall. But this is no worse
+> than the existing lifecycle events (Scheduled, Pulled, Created,
+> Started) that kubelet already emits per pod, so it doesn't add
+> proportionally more noise.
 
 ## Design Details
 
@@ -224,8 +231,10 @@ not evaluated for such pods.
 ### Events when running implicitly-root
 
 Whenever kubelet sees an implicitly-root container, it will create a kubernetes
-Event object warning the user.  These events will be logged no more often than
-once per hour per pod, unless Kubelet restarts.
+Event object warning the user. These events are throttled to at most
+1 event/pod/node/hour unless kubelet restarts (also see
+[Risks and Mitigations](#risks-and-mitigations) for the rapid pod creation
+case).
 
 Also, these events will be bypassed if the user explicitly sets `runAsUser` or
 `runAsGroup` in their pod.
--- a/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
+++ b/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
@@ -284,9 +284,9 @@ and if so it will color pods that are running as root in red.  For example,
 > zero-width, so naively embedding color codes would visibly widen output
 > for every row, not just flagged ones. Needs either a fix to the shared
 > table printer's width calculation, or a different surface (e.g. a
-> plain-text column instead of coloring existing text). The time between
-> Alpha and Beta will be used to work out feasibility and implement this if
-> possible.
+> plain-text column instead of coloring existing text). This will be
+> evaluated for feasibility and implemented as part of a separate KEP owned
+> by SIG CLI, if possible.
 
 ### Metrics
 
@@ -361,8 +361,6 @@ gate toggled on via `tempSetCurrentKubeletConfig`. Covered cases:
 - Gather feedback from users and providers
 - e2e tests are in Testgrid and linked in KEP
 - SIG-Scalability confirms the Events are a non-issue
-- Time between Alpha and Beta used to determine if kubectl `Color`
-  highlighting is possible, and implement it if so
 
 #### GA
 
--- a/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
+++ b/keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
@@ -1,0 +1,631 @@
+# KEP-2172: Warn about implicitly insecure running containers
+
+> **Note:** This proposal is based on the original draft by Tim Hockin
+> ([thockin/k8s-enhancements#2168](https://github.com/thockin/k8s-enhancements/pull/2168)).
+> It has been updated here to match the latest KEP template and current
+> proposed implementation details.
+
+<!-- toc -->
+- [Release Signoff Checklist](#release-signoff-checklist)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1 (Optional)](#story-1-optional)
+    - [Story 2 (Optional)](#story-2-optional)
+    - [Story 3](#story-3)
+  - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Condition when running implicitly-root](#condition-when-running-implicitly-root)
+  - [Events when running implicitly-root](#events-when-running-implicitly-root)
+  - [kubectl](#kubectl)
+    - [Visibility in <code>kubectl describe pod</code>](#visibility-in-kubectl-describe-pod)
+    - [Color](#color)
+  - [Metrics](#metrics)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+      - [Unit tests](#unit-tests)
+      - [Integration tests](#integration-tests)
+      - [e2e tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+    - [GA](#ga)
+    - [Deprecation](#deprecation)
+  - [Upgrade / Downgrade Strategy](#upgrade--downgrade-strategy)
+  - [Version Skew Strategy](#version-skew-strategy)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature Enablement and Rollback](#feature-enablement-and-rollback)
+  - [Rollout, Upgrade and Rollback Planning](#rollout-upgrade-and-rollback-planning)
+  - [Monitoring Requirements](#monitoring-requirements)
+  - [Dependencies](#dependencies)
+  - [Scalability](#scalability)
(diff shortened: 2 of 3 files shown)


keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,631 @@
...
+  Warning  ImplicitlyInsecureGroupID  156m  kubelet   container(s) [c] running as GID 0 without runAsGroup set
+```
+
+#### Color
+
+If possible, kubectl will detect whether it is printing to a console or not,
+and if so it will color pods that are running as root in red.  For example,
+`kubectl get pods` would highlight problematic pods.
+
+> Note: this is an open design question, as of now. kubectl's table output is
```
haircommander, line 331: I feel unsure if this should be included here instead of in a separate KEP for SIG cli

haircommander:
I wonder if there's anything we can do about 0 being defined in /etc/groups in the image.  In the SupplementalGroupsPolicy Merge we'd take the value from the image implicitly. We do have the groups set in container status from CRI, so we could report when there was 0 added by the image. Maybe just creating a new metric for when SupplementalGroupsPolicy: Merge is set would be fine too...

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,631 @@
...
+  Ready                       True
+  ContainersReady             True
+  PodScheduled                True
+  InsecureUserID              False
+  InsecureGroupID             True
+...
+Events:
+  Type     Reason                     Age   From      Message
+  ----     ------                     ----  ----      -------
+  Warning  ImplicitlyInsecureGroupID  156m  kubelet   container(s) [c] running as GID 0 without runAsGroup set
```
haircommander: a sufficiently motivated replicaset could make this pretty nois, even at an hour cadence. I wonder if there's anything to do... it'd at least be restricted to `podsPerNode*Nodes*Hour` .... maybe one event per pod is enough, along with persistent metrics and conditions?

Priyankasaggu11929: have updated the `Risks and Mitigations` section

to mention that events are throttled to 1 event/pod/node/hour and that cases like a replicaset recreating pods repeatedly is still bounded by `podsPerNode*nodes*hour`

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
Priyankasaggu11929: Updated the Note to mention that Kubectl color part will be evaluated in a separate KEP owned by SIG-CLI

also, removed it from beta graduation criteria.

Priyankasaggu11929:
> I wonder if there's anything we can do about 0 being defined in /etc/groups in the image. In the SupplementalGroupsPolicy Merge we'd take the value from the image implicitly. We do have the groups set in container status from CRI, so we could report when there was 0 added by the image. Maybe just creating a new metric for when SupplementalGroupsPolicy: Merge is set would be fine too...

added `supplementalGroupsPolicy: Merge` case across all three places - new pod status condition (InsecureGroupID), events and metrics

both for (i) case when implicit gid=0 appears in `status.containerStatuses[*].user.linux.supplementalGroups` due to the container's `/etc/group`

and (ii) explict case when it comes from setting gid=0 via either `spec.securityContext.fsGroup` or `spec.securityContext.supplementalGroups` 

(~have a working implementation in a local branch for ^. will push in a k/k PR soon~ PR: https://github.com/kubernetes/kubernetes/pull/142074)

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,686 @@
...
+The pod conditions/events/metrics update is skipped or retried along with
+the rest of the kubelet's normal pod status sync; no special handling is
+added.
+
+###### What are other known failure modes?
+
+- Older CRI runtime does not report `status.containerStatuses[].user.linux`
+  (predates stable KEP-3619 support). Observed directly on containerd
+  v1.7.33: kubelet ran without any error, but the `InsecureUserID`/
+  `InsecureGroupID` conditions stayed `False` even for pods that were
```
toVersus, line 650: Reporting `False` when we can't detect it seems misleading. For a condition, wouldn't it be better to use the [Unknown](https://pkg.go.dev/k8s.io/api/core/v1#ConditionStatus) condition status?

The metrics have the same problem. On a node where the runtime doesn't report `user.linux`, both gauges stay at `0`, which looks identical to a node with no insecure pods. Alerting on non-zero would silently miss exactly those nodes. Maybe we should expose whether detection is available, such as via a per-node gauge?

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,686 @@
...
+> plain-text column instead of coloring existing text). This will be
+> evaluated for feasibility and implemented as part of a separate KEP owned
+> by SIG CLI, if possible.
+
+### Metrics
+
+Kubelet will add two gauge metrics, both labeled by `id_type` (`uid`, `gid`,
+or `supplementalgroups`):
+
+- `kubelet_implicitly_insecure_pods`: number of pods with an implicitly-root
```
toVersus, line 346: Could we merge these into a single `kubelet_insecure_pods` metric and use a label (e.g. `declaration`) for `implicit`/`explicit`? Both count the same thing and only differ in how the ID was declared, which feels like something a label should express rather than two separate metric names. It would also let you get both the total and the breakdown from one metric.

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,686 @@
...
+> zero-width, so naively embedding color codes would visibly widen output
+> for every row, not just flagged ones. Needs either a fix to the shared
+> table printer's width calculation, or a different surface (e.g. a
+> plain-text column instead of coloring existing text). This will be
+> evaluated for feasibility and implemented as part of a separate KEP owned
+> by SIG CLI, if possible.
+
+### Metrics
+
+Kubelet will add two gauge metrics, both labeled by `id_type` (`uid`, `gid`,
```
toVersus, line 343: Do we need the gauge metrics at all? With the conditions in place, a third-party metrics exporter (e.g. kube-state-metrics) already covers "how many right now" via `kube_pod_status_condition`, and with pod/namespace labels on top.

A gauge also only reflects what exists at scrape time, so pods shorter-lived than the scrape interval (30s or 60s is typical) can come and go without ever being observed. Would a counter like `kubelet_insecure_pods_started_total` be a better fit here?

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
Priyankasaggu11929: updated the document to use the Unknown condition status - in cases when container runtime doesn't report user uid/gid or when the container itself is yet to start.

Also, added a note to use the `Node.Status.Features.SupplementalGroupsPolicy` field (from KEP-3619 to check if detectionis available.

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
Priyankasaggu11929: updated to merge the 2 metrics to a single `kubelet_insecure_pods`

thanks for the suggestion!

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
Priyankasaggu11929: the intention is to add a kubelet-native metric that works everywhere
(kube-state-metrics needs extra setup)

on the gauge metric part - that is intentional because the goal of this KEP is to give a live signal to act on (i.e. how many insecure pods exist right now and the number should go down to 0 as people fix them).
counter won't serve that purpose.

keps/prod-readiness/sig-node/2172.yaml
```diff
@@ -0,0 +1,3 @@
+kep-number: 2172
+alpha:
+  approver: "@soltysh"
```
jpbetz, line 3:
```suggestion
  approver: "@jpbetz"
```

jpbetz (commented):
I'm the PRR approver assigned to this KEP for 1.38 on the release board, so swapping the approver in the prod-readiness file. Questionnaire review to follow.

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,702 @@
...
+_For alpha, this section is encouraged: reviewers should consider these questions
+and attempt to answer them._
+
+_For beta, this section is required: reviewers must answer these questions._
+
+###### Will enabling / using this feature result in any new API calls?
+
+No new API calls. Conditions/events are set as part of the existing kubelet
+pod status sync path, which already calls the API server. Metrics involve no
+API calls; they are only scraped from kubelet's `/metrics` endpoint.
```
jpbetz, line 615: Events are new API calls.

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,702 @@
...
+
+###### Will enabling / using this feature result in any new calls to the cloud provider?
+
+No.
+
+###### Will enabling / using this feature result in increasing size or count of the existing API objects?
+
+Yes: two new `PodCondition` entries (`InsecureUserID`, `InsecureGroupID`) on
+implicitly-root pods, and `Event` objects (throttled to 1 event/pod/hour),
+both small and bounded per pod.
```
jpbetz, line 629: Pod size has impact on scale. Does this change the size of ALL pods or only insecure pods?

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,702 @@
...
+
+No behavioral change (this feature is purely observational). Enabling it
+adds new pod conditions, one of three warning event reasons, and a new
+metric; nothing about how a pod runs changes.
+
+###### Can the feature be disabled once it has been enabled (i.e. can we roll back the enablement)?
+
+Yes. Disabling stops new conditions/events/metric from being generated; it
+does not affect running workloads. Already-set conditions on existing pod
+objects are left as-is until the pod is otherwise resynced/recreated.
```
jpbetz, line 510: When the gate is turned off and kubelet restarts, does kubelet remove the two conditions on the next status sync, or do they stay on the pod until it is deleted? 

A stale `True` after rollback could easily become misleading...

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,702 @@
...
+-->
+
+### Feature Enablement and Rollback
+
+_This section must be completed when targeting alpha to a release._
+
+###### How can this feature be enabled / disabled in a live cluster?
+
+- [x] Feature gate (also fill in values in `kep.yaml`)
+  - Feature gate name: `InsecurePodWarnings`
```
jpbetz, line 496: Use a more specific name? This isn't for all imaginable pod insecurity warnings.

keps/sig-node/2172-warn-about-implicitly-insecure-containers/README.md
```diff
@@ -0,0 +1,702 @@
...
+set a specific `runAs...` or to set it to 0. Pete can also install
+admission controllers to only allow approved users to set the `runAs...`
+fields to 0.
+
+#### Story 3
+
+Usain the user will see `Warning` events with reasons
+`ImplicitlyInsecureUserID`, `ImplicitlyInsecureGroupID`, or
+`ImplicitlyInsecureUserAndGroupID` when they run `kubectl describe pod`, and
+will eventually choose to make them go away by running as non-root.
```
jpbetz, line 160: Does this use status conditions, events, or both? My understanding from below is that it's "both", in which case I recommend stating that directly in the proposal.  Something like "This enhancement adds status conditions, events and a metric".

jpbetz:
Hi @Priyankasaggu11929. I've left some review comments. I can review again once addressed. Thanks!

jpbetz:
Hi @Priyankasaggu11929, I can do another PRR pass and get this approved once the above comments are addressed.

Issue #5836, opened 2026-01-26
Scheduler Preemption for In-Place Pod Resize


natasha41575:
### Enhancement Description

- One-line enhancement description (can be used as a release note): The scheduler will watch for [in-place pod resize](https://github.com/kubernetes/enhancements/issues/1287) requests that are marked as Deferred and preempt lower-priority pods until there is enough room on the node for the resize to succeed. 
- Kubernetes Enhancement Proposal: https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
- Discussion Link: https://docs.google.com/document/d/1xnpexJ0ojI0GPsrI9V75EzdC_4szAcuBt6MUb5beA7M/edit?usp=sharing
- PRs by stage and milestone:
  - [x] Alpha - v1.37
    - [x] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/5932
    - [x] Code (`k/k`) update PR(s): https://github.com/kubernetes/kubernetes/pull/140000
    - [x] Docs (`k/website`) update PR(s): https://github.com/kubernetes/website/pull/56348
  - [ ] Beta - v1.38 [[Graduation criteria tracking issue](https://github.com/kubernetes/kubernetes/issues/140854)]
    - [ ] KEP (`k/enhancements`) update PR(s): https://github.com/kubernetes/enhancements/pull/6344
    - [ ] Code (`k/k`) update PR(s):
    - [ ] Docs (`k/website`) update(s):

_Please keep this description up to date. This will help the Enhancement Team to track the evolution of the enhancement efficiently._

natasha41575:
/assign
/sig scheduling

cc'ing the following sigs as collaborators / stakeholders:
/sig node
/sig autoscaling

cc @adrianmoisey @omerap12 @tallclair @dom4ha @sanposhiho @macsko

ffromani:
/cc
(so I can track updates, not planning to review)

lasomethingsomething:
Hello @natasha41575  👋, 1.37 Enhancements team here.

This is a reminder of the upcoming [PRR freeze](https://github.com/kubernetes/sig-release/blob/master/releases/release_phases.md#prr-freeze) on **Tuesday 9th June 2026 (AoE) / Wednesday 10th June 2026, 12:00 UTC**.

This enhancement is targeting stage `alpha` for 1.37 (correct me, if otherwise)

[...]

PR #6344, opened 2026-09-10
KEP 5836: Scheduler preemption for IPPR beta update

natasha41575:
Scheduler preemption for IPPR beta update

Enhancement: https://github.com/kubernetes/enhancements/issues/5836
Beta graduation criteria: https://github.com/kubernetes/kubernetes/issues/140854

Commits are organized to ease review.

Antigravity assisted with research and writing, but all design ideas came from humans and all AI output was double-checked.


keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+
+#### 3. Gang Resize Readiness Guard (Mitigating Controller Patch Skew)
+In Kubernetes, there is no transactional multi-pod `PATCH` API. External autoscalers patch worker pods sequentially (e.g. patching Worker 0, then Worker 1, then Worker 2). This means that the scheduler may see the first worker in a gang transition to `Deferred`, while the other workers are still in the process of being patched. This can lead to the scheduler preempting for only the first worker, rather than for the entire gang.
+
+The `PodGroup` API already defines `spec.schedulingPolicy.gang.minCount`, which represents the minimum number of member pods that must be schedulable concurrently to satisfy the gang's all-or-nothing quorum. To prevent the scheduler from prematurely executing preemption on the first patched worker while the remaining workers are still awaiting API updates, the `DeferredPodScheduling` plugin implements a readiness check during `PreFilter`:
```
dom4ha, line 710: It's not obvious to me that obeying minCount of already running PodGroup is always the right choice.
What if pods need to scale up independently?
What if a number of active pods drops below minCount?
Since we don't stop the remaining ones (AFAIK controller may decide to do so, but it does not have to), we should not block their ability to scale.

What about using `DisruptionMode` to adjust the resize behavior? That would mean that `All` would enforce all running pods to be scaled up or none (even the ones >minCount), but `Single` would in turn ignore minCount completely.

We are missing a case where the DisruptionMode defines either the same `minCount` that in scheduling criteria or smaller (if a workload can tolerate degraded mode), but since we don't have it yet, I propose to stick to the two existing ones.

/cc @wojtek-t @mm4tt

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+* **Zero-Disruption Rollback**: If even a single resizing pod cannot satisfy its deficit on its node, the entire preemption attempt fails atomically. Tentative state is cleared in memory, and **zero victims are evicted on any node**.
+
+This atomic behavior is necessary for several reasons:
+1. In synchronous distributed training, all workers must execute symmetric compute steps before synchronizing via collective operations. If only a subset of workers scales up, the entire distributed run is bottlenecked by the slowest un-resized worker, wasting the additional resources granted to the resized pods.
+2. Asymmetric resource allocations across workers in a gang can lead to communication timeouts or out-of-memory errors on the worker that could not be resized.
+3. If an autoscaler (such as the Vertical Pod Autoscaler) fails to resize a worker in-place, it may fall back to evicting the pod to recreate it with larger requests. For a `PodGroup` configured with `DisruptionMode: All`, evicting that single pod will trigger the cascaded termination of the entire gang.
```
dom4ha: What if the actual workers count is bigger than minCount? Do we expect it's sufficient to resize the minCount number of pods? How autoscaler would behave if it decides to turn down a >minCount pod?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -439,18 +439,37 @@ The primary risk with this approach is ensuring that no existing or future sched
 2. No downstream scheduling or preemption logic relies on strict pointer equality of pod objects across the cache and queue boundaries.
 3. Resource double-counting is avoided on the target node by excluding the pod's additional resources during resource fit calculations as described in the section below.
 
+Even with these considerations in mind, asynchronous event delivery and concurrent pod updates mean that the pod snapshot popped from the scheduling queue can occasionally carry resource requests that temporarily diverge from the cache's snapshot. In Alpha, `NodeResourcesFit` amortized this discrepancy at runtime by assuming the pod's resources as the maximum between the queued and cached snapshots. For Beta, we eliminate this divergence at the source by refreshing the queued pod snapshot directly from the node's cache snapshot at the start of the scheduling cycle, ensuring all plugins evaluate the exact same immutable pod instance.
```
macsko, line 442: Is it planned to take this pod snapshot from cache for all pods or only deferred ones?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -439,18 +439,37 @@ The primary risk with this approach is ensuring that no existing or future sched
...
-To achieve this cleanly and without hardcoding specific plugin names on the framework side, we adjust all irrelevant plugins to be skipped when processing `Deferred` pods.
+```go
+// ResponsibilityPlugin allows a scheduling plugin to declare whether it is
+// responsible for evaluating a given pod during its scheduling cycle.
+type ResponsibilityPlugin interface {
+    Plugin
+    // IsResponsible returns a Status indicating whether this plugin is responsible
+    // for evaluating the given pod. If it returns fwk.Skip, the framework skips all
+    // extension points for this plugin during the scheduling cycle.
+    IsResponsible(ctx context.Context, cycleState *CycleState, pod *v1.Pod) *fwk.Status
```
macsko, line 462: Does this also mean that if plugin returned Skip from IsResponsible in the scheduling cycle, this plugin won't be also called in the preemption algorithm?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -460,46 +479,57 @@ The `NodeName` plugin is adjusted to implement the `PreFilter` phase. Because th
...
+In most cases, this restricted search results in a resource fit error, naturally triggering the scheduler's preemption pathway to identify victims on the assigned node.
 
-However, if the resize fits without needing preemption (e.g., if cluster topology shifts or other workloads exit, freeing up capacity on the host), the `NodeResourcesFit` plugin returns a status of `UnschedulableAndUnresolvable` (with a message like `"pod resize fits, waiting for Kubelet actuation"`). Because the node status is `UnschedulableAndUnresolvable` rather than `Unschedulable`, the `DefaultPreemption` plugin's PostFilter phase skips victim selection on the node (as it only evaluates `Unschedulable` nodes). This ensures the scheduler skips the preemption phase, but keeps the deferred pod in the `Unschedulable` queue.
+However, if the resize fits without needing preemption (either because cluster topology shifted or because victim pods have finished terminating), the pod progresses through the `Permit` and `Bind` phases to complete the scheduling cycle. During `Bind`, because the pod is already bound to the node, the scheduler skips binding the pod and removes the pod from the scheduling queue.
```
macsko, line 488: Can we safely do this? What about Reserve or PreBind plugins? Do they need to check for deferred pods inside not to handle them incorrectly?

macsko, line 488: Why, compared to the alpha, the pod is removed from the queue in this case?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -460,46 +479,57 @@ The `NodeName` plugin is adjusted to implement the `PreFilter` phase. Because th
...
 
-Because a parked pod's capacity can be consumed by other workloads before the Kubelet actuates it, the scheduler registers `NodeResourcesFit` QHints to watch for pod upsize events on the same node. If another pod on the node scales up, the QHint moves the parked pod back to the `activeQ` for re-evaluation.
+Because the scheduler cache accounts for the pod using `max(desired, allocated, actual)`, the required capacity remains reserved in `nodeInfo.Requested`. However, race conditions can cause the Kubelet to resize a different pod than the one that triggered the eviction. When an assigned pod scales up on a node, the scheduler's `updatePod` event handler detects the scale-up and checks for any `Deferred` pods running on that same host. To prevent cluster-wide pod iterations and maintain scalability, the scheduler utilizes the pod informer indexer with a secondary index keyed by `spec.nodeName` to look up only the pods bound to that specific node, immediately re-enqueuing any lingering `Deferred` pods back into `activeQ` for re-evaluation and preemption.
```
macsko, line 490: This part about using the pod informer is a bit vague to me. How this requeueing will work? In which place the informer will be called?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
 *   **No Nominated Node Name (NNN) Assignment**: Standard scheduling failures can record a `NominatedNodeName` on the pod status to reserve space on a target candidate node. For a `Deferred` pod, no nominated node is set or updated. The pod is already bound to its host, and its resize can only be evaluated and satisfied on that specific host; nominating another node is invalid.
 
-### Scope of Interaction with Workload-Aware Preemption
+### Interaction with Workload-Aware Scheduling (WAS) and Gang Scheduling
+
+When in-place pod resizing operates in clusters running Workload-Aware Scheduling (WAS / WAP), the scheduler coordinates resizing and preemption at the group level to preserve workload invariants.
+
+#### 1. Group-Level Scheduling Queue Enqueuing
+When an assigned pod belonging to a `PodGroup` transitions to `Deferred`, the scheduler's informer event handlers identify its parent `PodGroup` (and root `CompositePodGroup` if nested) and enqueue the entire group into the scheduling queue. This ensures that multi-pod resize requests are evaluated collectively within `scheduleOnePodGroup` rather than as isolated, decoupled per-pod scheduling cycles.
```
macsko, line 700: What if we have a deferred pod and a new unschedulable pod, where both point to the same parent?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+
+#### 2. Node-Locked Placement Invariant
+In gang scheduling, member pods are already placed and bound to specific nodes. The scheduler enforces that resizing pods are evaluated strictly on their assigned hosts through two mechanisms:
+* **Placement Short-Circuiting**: The workload placement engine detects that all member pods already have `spec.nodeName` populated, bypassing expensive cluster-wide candidate placement combinatorics and restricting evaluation directly to the pods' current nodes.
```
macsko, line 704: IIUC, this should work without any adjustments, because plugins (e.g., TAS) identify the placement of assigned pods and restricts scheduling to that placement

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+
+* **Settling Window**: If `deferredCount < minCount`, the plugin calculates the elapsed time since the latest member transitioned to `Deferred`. If the time elapsed exceeds a settling window, the `DeferredPodScheduling` plugin assumes the partial resize attempt is intended and allows the pod to proceed to the next phase. The settling window is hardcoded to 5 seconds in the initial beta; we can consider making it configurable in a future release. This can be implemented by maintaining a debounced timer per `PodGroup` that reactivates the group for scheduling once the settling window has elapsed.
+
+#### 4. Workload-Level Resource Fit Checks
+During `scheduleOnePodGroup`, the scheduler evaluates resource availability for each resizing pod on its assigned node using the scheduler cache snapshot:
+* **Tentative In-Memory Reservation**: As each resizing pod passes its filter checks, the scheduler tentatively assumes the pod at its requested spec on the `NodeInfoSnapshot`.
```
macsko, line 718: Does the snapshot support assuming the already-assigned pod?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -245,9 +264,15 @@ with higher-level autoscaling controllers such as VPA's new `InPlaceOrRecreate`
 way to control scale-up behavior when a node lacks the capacity to fulfill the request. Consequently, workloads may face 
 disruptions such as being moved to a larger node, or suffering an OOM-kill because memory could not be scaled up in-place.
 
-Scheduler preemption eliminates the gap by introducing an configurable ability to free up capacity on a fully-utilized
+Scheduler preemption eliminates the gap by introducing a configurable ability to free up capacity on a fully-utilized
 node to allow the scale up to succeed in-place.
 
+In production Kubernetes environments, cluster administrators strive to maximize resource utilization and efficiency. A common strategy is to bin-pack unused capacity on not-yet-full nodes with lower-priority workloads, such as batch jobs, background data processing, or best-effort tasks.
```
wojtek-t, line 270: nit: can you please avoid that long lines by splitting them (it helps with reviews and more targetted comments).

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -341,57 +366,34 @@ From the scheduler's view, once the spec is updated, the resources are already r
...
-
-If we use our modified logic, where only the resizing pod uses `max(desired, allocated, actual)` and the remaining pods use `max(allocated, actual)`, then:
-
-1. Preemption logic first removes all candidate pods, so it sees node is using X (Pod 1's max).
-2. The Scheduler first tries to reprieve pod 2, which assumed to have max(X/2, X/2) resources.  X/2 + X = 1.5X, which is less than node allocatable 2X, so pod 2 is reprieved.
-3. The Scheduler tries to reprieve pod 3, which assumes to have max(X/2, X/2) resources. X/2 + 1.5X = 2X, which still fits within the node allocatable 2X, so pod 3 is reprieved.
-4. The Scheduler tries to reprieve pod 4, which assumes to have max(X/2, X/2) resources. X/2 + 2X = 2.5X, which is more than node allocatable 2X, so pod 4 is not reprieved.
+We evaluated whether to modify the preemption reprieve logic to calculate candidate victim pods using only their allocated resources, but we decided to retain `max(desired, allocated, actual)` uniformly across the scheduler due to several drawbacks of introducing context-dependent resource calculations:
```
wojtek-t, line 391: +1 for keeping this unified

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t: +1 to this question

Or should this instead be a per-method extension?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -439,18 +439,37 @@ The primary risk with this approach is ensuring that no existing or future sched
...
-* **Bypassing Irrelevant Constraints**: All scheduling plugins irrelevant to direct resource capacity fitting (such as node affinity, pod anti-affinity, and topology spread constraints) are bypassed.
+To bypass plugins that are irrelevant to deferred pods without hardcoding plugin lists or scattering ad-hoc bypass checks across individual plugins, we introduce the `ResponsibilityPlugin` extension point interface in the scheduling framework:
 
-To achieve this cleanly and without hardcoding specific plugin names on the framework side, we adjust all irrelevant plugins to be skipped when processing `Deferred` pods.
+```go
+// ResponsibilityPlugin allows a scheduling plugin to declare whether it is
+// responsible for evaluating a given pod during its scheduling cycle.
+type ResponsibilityPlugin interface {
+    Plugin
+    // IsResponsible returns a Status indicating whether this plugin is responsible
```
wojtek-t, line 459: nit: I'm not a fan of "IsResponsible" name.
How about: "IsRelevantForPod" instead? Or sth closer to it?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t: I think the first question we should ask is "what is the usecase and why do we even care about considering all members from a gang together".

Personally, I can only see one reason of - try to optimize the number/quality of preemptions.
If I have only one pod that requires upscale, I'm ideally looking for a individual pod running on that node that would make that schedulable.
If I have 100 pods, it's e.g. fine to kill the whole lower-priority job if it is exactly collocated with my pods.


Now - how often do we scale-up whole gang workloads vs individual pods?
I'm definitely missing data here so would benefit from seeing some, but my anecdotal knowledge is that I'm not really aware of usecases where we want to in-place resize the whole gang.

So unless I'm wrong, I would actually strongly suggest to avoid unnecessary complexity and ignore this usecase until we have a clear evidence that is really needed.

In other words - whenever I see an DeferredResize, I just process it (if I managed to observe more than one within my workload - great. But I don't actively do anything to batch them).

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+
+#### 6. Atomic Multi-Node Preemption
+When multiple pods in a `PodGroup` require preemption across different nodes, the preemption engine executes an atomic, all-or-nothing preemption cycle:
+* **All-or-Nothing Preemption Dry Run**: The preemption algorithm dry-runs victim candidate selection across all nodes assigned to the resizing gang members.
+* **Atomic Dispatch**: If and only if **all** resizing pods find sufficient victim capacity on their respective nodes, the scheduler commits the preemption cycle and dispatches eviction calls across all affected nodes simultaneously.
+* **Zero-Disruption Rollback**: If even a single resizing pod cannot satisfy its deficit on its node, the entire preemption attempt fails atomically. Tentative state is cleared in memory, and **zero victims are evicted on any node**.
+
+This atomic behavior is necessary for several reasons:
+1. In synchronous distributed training, all workers must execute symmetric compute steps before synchronizing via collective operations. If only a subset of workers scales up, the entire distributed run is bottlenecked by the slowest un-resized worker, wasting the additional resources granted to the resized pods.
```
wojtek-t, line 735: True - although all users I'm aware of run training in a "pod-per-node" mode.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+* If a candidate victim belongs to a `PodGroup` configured with `DisruptionMode: All`, evicting that pod cascades eviction across the entire victim gang across the cluster.
+* The preemption engine factors this collective disruption cost into candidate ranking, strictly preferring standalone pods or `DisruptionMode: Single` pods over disrupting entire gangs.
+* During the reprieve phase, victim groups are reprieved atomically.
+
+#### 6. Atomic Multi-Node Preemption
+When multiple pods in a `PodGroup` require preemption across different nodes, the preemption engine executes an atomic, all-or-nothing preemption cycle:
+* **All-or-Nothing Preemption Dry Run**: The preemption algorithm dry-runs victim candidate selection across all nodes assigned to the resizing gang members.
+* **Atomic Dispatch**: If and only if **all** resizing pods find sufficient victim capacity on their respective nodes, the scheduler commits the preemption cycle and dispatches eviction calls across all affected nodes simultaneously.
```
wojtek-t, line 731: This sounds nice on paper, but I'm not convinced we should be doing that until we hear very explicit feedback that something like that is needed.

It's a consequence of some of my other comments - scaling up a whole gang is very tricky to coordinate anyway and I'm not aware of real-life usecases for it.

So let's either provide a very concrete usecase (with numbers) or let's postpone it.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t, line 1173: Were those added? If so please link, otherwise please update and prioritize adding those.

wojtek-t, line 1135: nit: Can you please update to explicitly

Yes - 'Deferred' resizes will trigger premption of lower-priority pods.


[IPPR is enabled by-default, PriorityClasses are also widely used.]

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -1003,7 +1122,7 @@ well as the [existing list] of feature gates.
 -->
 
 - [x] Feature gate (also fill in values in `kep.yaml`)
-  - Feature gate name: SchedulerPreemptionForPodResize
+  - Feature gate name: InPlacePodVerticalScalingSchedulerPreemption
```
wojtek-t, line 1125: Were there really any kubelet changes as part of this KEP?

My understanding is that even before this KEP, kubelet was "retrying" the deferred resizes when some pod was deleted.

So is Kubelet relevant here?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -1071,13 +1190,20 @@ rollout. Similarly, consider large clusters and how enablement/disablement
 will rollout across nodes.
 -->
 
+Rollout: If kube-scheduler enables the feature before kube-apiserver, the API server will drop node.spec.podPreemptionPolicy. Controllers managing resizable nodes will fail to disable resize preemption, causing the scheduler to evict lower-priority pods on nodes intended to be protected from preemption.
```
wojtek-t, line 1193: Let's state explicitly that a feature effectively consists of two parts:

1. Preemption triggerred by Deferred updates - the implementation of this feature is purely in-memory inside scheduler.
2. The API to enabled/disable preemption on a per-node basis.

The former doesn't really depend on the order.
The later is where the order of rollout matters.

wojtek-t: What other failure modes can we expect especially for (1)?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -1071,13 +1190,20 @@ rollout. Similarly, consider large clusters and how enablement/disablement
...
+
+Rollback: Disabling the feature gate and restarting kube-scheduler reverts behavior immediately: deferred pods will no longer be placed in the scheduling queue or trigger preemption; running workloads are unaffected.
```
wojtek-t, line 1195: This is not a failure - so it's irrelevant for this question.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -1071,13 +1190,20 @@ rollout. Similarly, consider large clusters and how enablement/disablement
...
+
```
wojtek-t, line 1196: Please explain how it affects running workloads (it's very relevant because now running workloads can be preempted as a result of action that didn't trigger their preemption before).

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -341,57 +366,34 @@ From the scheduler's view, once the spec is updated, the resources are already r
...
-
-In specific edge cases, this leads to redundant preemption. "Victim A" (targeted by the first scheduler) and "Victim B" (targeted by the second scheduler) may both be terminated to satisfy a single resize request.
-
-Mitigations:
-*   **Idempotency:** The `Delete` API is already idempotent. If the scheduler picks the same victim upon restart, the API server simply acknowledges the request without further disruption.
-*   **Acceptable Waste:** In the Kubernetes priority model, ensuring the success of a higher-priority workload (the resizing pod) justifies the potential loss of multiple lower-priority victims during a rare control-plane failure.
-
-To address double preemption risks across scheduler restarts and long graceful termination periods, we are exploring tracking mechanisms (such as using the pod's nominated node name). We will fully evaluate these options and finalize the approach for Beta; see details in [Tracking Preemption Nominations to Avoid Double Preemption](#tracking-preemption-nominations-to-avoid-double-preemption).
-
 ##### Risk of Additional Preemption
```
sanposhiho, line 378: not only about the preemption, but isn't there a similar risk that 
1. podA is IPPR from 2 CPU to 8 CPU on nodeA. nodeA has 6 unallocated CPUs (i.e., PodA needs 2 more CPUs).
3. podA cannot be scaled up because it cannot find a victim
4. A new podB is being scheduled, but cannot go to nodeA because the resource fit plugin misunderstands podA is using 8 CPUs and nodeA doesn't have any free CPU.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -341,57 +366,34 @@ From the scheduler's view, once the spec is updated, the resources are already r
...
-4. The Scheduler tries to reprieve pod 4, which assumes to have max(X/2, X/2) resources. X/2 + 2X = 2.5X, which is more than node allocatable 2X, so pod 4 is not reprieved.
+We evaluated whether to modify the preemption reprieve logic to calculate candidate victim pods using only their allocated resources, but we decided to retain `max(desired, allocated, actual)` uniformly across the scheduler due to several drawbacks of introducing context-dependent resource calculations:
 
-This results in only 1 pod (Pod 4) being preempted.
-
-Such a modification to the `NodeResourceFit` plugin is out of scope for alpha due to its wide ranging implications across the scheduler. However, we will reconsider this decision prior to beta.
+* `NodeInfo` is precomputed before each scheduling cycle and shared across framework plugins. Conditionally altering resource formulas based on whether a pod is a preemptor, candidate, or reprieved victim breaks `NodeInfo` encapsulation and immutability. Recalculating pod footprints dynamically during reprieve loops would also degrade scheduler throughput by replacing fast pre-aggregated lookups with repeated per-pod evaluations.
```
sanposhiho: yes, we should not conditionally alter how the used resource on each node is precomputed. but, why do we need to conditionally alter? any issue/difficulty if NodeInfo follows what kubelet does always?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
sanposhiho: > Does this also mean that if plugin returned Skip from IsResponsible in the scheduling cycle, this plugin won't be also called in the preemption algorithm?

Yes, right? Any potential issue with this?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
sanposhiho: > Do they need to check for deferred pods

If they do, they will implement `ResponsibilityPlugin`. In in-tree plugins, no plugins at binding cycle will implement ResponsibilityPlugin and hence bindng cycles will be no-op for deferred pods. Custom plugins will be able to implemnt `ResponsibilityPlugin` and do whatever they need

sanposhiho: because the deferred pods are gone like normal pods that have bound

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -460,46 +479,57 @@ The `NodeName` plugin is adjusted to implement the `PreFilter` phase. Because th
...
+
+For in-place resize pods, the pod is already bound and `NominatedNodeName` is not set. To prevent double preemption for resizing pods, we modify the default preemption plugin to instead evaluate the pod's `spec.NodeName` when `NominatedNodeName` is empty. If there exists an already-terminating lower-priority pod on the node that the pod is bound to, the default preemption plugin safely suppresses duplicate preemption cycles across long victim termination grace periods and scheduler restarts.
```
sanposhiho: do we really need this? `NodeName` plugin should anyway pin this deferred pod to the node and other nodes won't be involved to the preemtion, right?

sanposhiho: and given that, do we even have this risk? both `Long Grace Periods` and `Scheduler Restarts` exist for normal pod preemption too

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -629,62 +642,32 @@ The strings under `disableResizePreemption` have the same validation and constra
...
+2. **Scheduler Inspection in Filter Phase**: When the scheduler evaluates a pod with a `Deferred` resize, `DeferredPodScheduling.Filter` checks `node.Spec.PodPreemptionPolicy.DisableResizePreemption` on the pod's assigned node. If the list is non-empty, `Filter` returns status `UnschedulableAndUnresolvable` with reason `ErrReasonNodeDisablesResizePreemption` (`"node had resize preemption disabled"`).
+3. **Skipping Preemption**: Because the node status is `UnschedulableAndUnresolvable` (rather than `Unschedulable`), the `DefaultPreemption` PostFilter plugin skips preemption victim selection on the node. The pod is placed in the `unschedulablePods` pool without evicting lower-priority workloads.
+4. **Queueing Hints & Dynamic Wakeup**: `DeferredPodScheduling` registers QueueingHints for and Node `Add` and `Update` events. If the node's `spec.podPreemptionPolicy.disableResizePreemption` field is cleared or modified to allow preemption, the QHint moves deferred pods assigned to that node back to the `activeQ` for immediate re-evaluation.
```
sanposhiho: do we need `Add`?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +692,127 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
 *   **Skipping `PodScheduled` Condition Updates**: The standard failure handler sets the pod's `PodScheduled` status condition to `False` (with reason `Unschedulable`). For a `Deferred` pod, updating this condition is skipped. For beta, we will implement emitting events to surface the results of the scheduling cycle for observability.
 *   **No Nominated Node Name (NNN) Assignment**: Standard scheduling failures can record a `NominatedNodeName` on the pod status to reserve space on a target candidate node. For a `Deferred` pod, no nominated node is set or updated. The pod is already bound to its host, and its resize can only be evaluated and satisfied on that specific host; nominating another node is invalid.
 
-### Scope of Interaction with Workload-Aware Preemption
+### Interaction with Workload-Aware Scheduling (WAS) and Gang Scheduling
```
sanposhiho, line 872: I ... am not sure if we should do this rich integration with WAS. I just feel like handling each pod individually (as non PG pods) is enough. What is the strong justification to this?

natasha41575: SGTM. I think your sentiment is shared based on @wojtek-t and @dom4ha's suggestions above, so I'll rewrite this section so that we handle pods individually, and we can reevaluate in the future if some justification comes up to change it. Thanks!

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: > My understanding is that even before this KEP, kubelet was "retrying" the deferred resizes when some pod was deleted.

This is correct, but there is still one small change to Kubelet under this FG - Kubelet no longer triggers its own critical admission plugin handler for deferred resize pods. Kubelet previously triggered preemption to make room for the resize of deferred pods that were marked critical (i.e. the `system-node-critical` or `system-cluster-critical` priority classes), but under this new FG it doesn't, and leaves preemption to the scheduler.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
macsko: > > Does this also mean that if plugin returned Skip from IsResponsible in the scheduling cycle, this plugin won't be also called in the preemption algorithm?
> 
> Yes, right? Any potential issue with this?

Probably not, but the current description is unclear

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
macsko: I think that we should force out-of-tree plugins to implement the IsResponsible method. Otherwise, we risk issues when they unexpectedly start getting deferred pods, which are bound, and act on these, particularly in Reserve or PreBind. Also, it feels that the KEP should mention somewhere that setting NNN in binding cycle should not be performed for deferred pods.

sanposhiho: If IsResponsible is not implemented, plugins will handle only unschedulable pods, which keeps the existing behavior. So, it won't happen that custom plugins unexpectedly start getting deferred pods. They will only when they implement IsResponsible and opt-in.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t: That makes sense, but I'm not seeing it in the KEP - can you please explicitly add that to the KEP and link in relevant places?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t: It's a valid concern, but was added by the IPPR KEP (not this one).

Worth pointing for completeness, but it should not block/affect this KEP at this point I think.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
wojtek-t: Right - I think I brought similar concern in https://github.com/kubernetes/enhancements/pull/6344#discussion_r4046909570

So it seems we're aligned despite commenting independently on it...

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Good catch, removed Add; only Node Update events are needed.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: I had a brief mention in the KEP but it was a bit vague. I fleshed out that section to more clearly describe the critical admission handler, and also added a note and link to the PRR questionnaire about changing default behavior.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Renamed to RelevancePlugin / IsRelevantForPod.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Yes, returning `fwk.Skip` from` IsRelevantForPod` skips the plugin across both the scheduling cycle and `PostFilter` preemption. I updated the KEP to clearly state this.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: The risk here is preempting additional victims on the same node, not on other nodes.

For normal unscheduled pods, DefaultPreemption prevents evicting extra victims during long termination grace periods or after a scheduler restart by checking pod.Status.NominatedNodeName and suppressing further preemption if any pod on that nominated node already has the DisruptionTarget condition.

For an already-bound resizing pod, status.NominatedNodeName is always empty (""). Without falling back to spec.NodeName when NominatedNodeName is empty, PodEligibleToPreemptOthers sees no nominated node and returns true, which would cause subsequent scheduling cycles to evict additional lower-priority victims on that same pinned node while the first victim is still terminating.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Only for deferred pods (since normal unscheduled pods have empty spec.NodeName and are not present in any cached NodeInfo). Clarified in the KEP.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: > If IsResponsible is not implemented, plugins will handle only unschedulable pods, which keeps the existing behavior. So, it won't happen that custom plugins unexpectedly start getting deferred pods. They will only when they implement IsResponsible and opt-in.

+1. I wrote this in the KEP as well.

>  Also, it feels that the KEP should mention somewhere that setting NNN in binding cycle should not be performed for deferred pods.

This is written under `#### Failure Handler Adjustments for Deferred Pods`

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: In Kubelet's admission loop, Kubelet evaluates `max(allocated, actual)` for all _other_ pods on the node, and `max(desired, allocated, actual)` for the _target_ pod being admitted. If the scheduler precomputed NodeInfo using `max(allocated, actual)` for all bound pods:

1. When evaluating a Deferred pod for preemption, the scheduler still needs that pod's desired delta to know how much capacity to preempt, which would require conditional logic for the target pod - this has severe performance and plugin complexity implications, which is what I am trying to avoid.
2. More importantly, Kubelet actuates Deferred resizes asynchronously as soon as capacity frees up without consulting the scheduler. If NodeInfo ignored desired for a Deferred pod whose victims just terminated (or when another pod exits), the scheduler could see that freed capacity as available and bind a new pod to the node concurrently while Kubelet actuates the resize, leading to the new pod to be rejected by Kubelet admission put into `Failed` state.

This section is saying that this KEP does not change the way scheduler uses NodeInfo, which diverges slightly from Kubelet resource evaluation, and accepting the risks it comes with.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Yes, this was added by the IPPR KEP. But it was done so intentionally to prevent this race condition:

- If scheduler sees podA desired capacity as 'free', it may schedule a new podB to the nodeA to use the space.
- Kubelet does not consult the scheduler when making resize decisions, so it can independently actuate podA's resize and fill the desired capacity.
- podB then asynchronously gets rejected by Kubelet (after scheduler placed it on nodeA), and ends up in Failed state.

In any case, I agree with @wojtek-t that the problem is not introduced by this KEP.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: As @sanposhiho noted, once the resize fits without needing preemption, the scheduler's work is complete. Letting it proceed through the cycle pops it from the scheduling queue just like a normal bound pod and avoids the scheduler queue from having to continuously track it while Kubelet actuates the resize.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Clarified in the KEP: this happens in the assigned-pod `UpdateFunc` handler (`updatePod` in `pkg/scheduler/eventhandlers.go`). When an assigned pod's requested resources increase, we query the pod informer indexer (`ByIndex(nodeNameIndex, pod.Spec.NodeName)`) to retrieve pods on that node and call `SchedulingQueue.Add` for any pod in the `Deferred` resize state.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Based on the consensus across reviews, I've simplified this section so that Deferred resizing pods belonging to a PodGroup are evaluated and preempted individually (as non-PodGroup pods). We can revisit group-coordinated resize preemption in the future if concrete use cases arise.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Left a comment in https://github.com/kubernetes/enhancements/pull/6344#discussion_r4064718376 as well - this section is updated to evaluate pods individually (not worry about PodGroups) for Deferred resizes.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Yes, they were. Added links and specific test names.

natasha41575: done.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: fixed this, thanks!

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: no longer relevant, so marking as resolved

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: no longer relevant, so marking as resolved

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: no longer relevant, so marking as resolved

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: no longer relevant, marking as resolved

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: no longer relevant, marking as resolved

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Updated this section accordingly, PTAL and let me know if you have further questions

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: ack, removed

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: added a note here in the KEP about running workloads

natasha41575:
Feedback is addressed, this is ready for another round.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/kep.yaml
```diff
@@ -22,22 +22,27 @@ approvers:
...
+  - name: InPlacePodVerticalScalingSchedulerPreemption
     components:
       - kube-apiserver
       - kube-scheduler
       - kubelet
 disable-supported: true
 
 metrics:
+  - preemption_attempts_total
+  - preemption_victims
```
macsko: nit:
```suggestion
  - scheduler_preemption_attempts_total
  - scheduler_preemption_victims
```

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -439,18 +483,69 @@ The primary risk with this approach is ensuring that no existing or future sched
...
 
+Even with these considerations in mind, asynchronous event delivery and
+concurrent pod updates mean that the pod snapshot popped from the scheduling
+queue can occasionally carry resource requests that temporarily diverge from the
+cache's snapshot. In Alpha, `NodeResourcesFit` amortized this discrepancy at
+runtime by assuming the pod's resources as the maximum between the queued and
+cached snapshots. For Beta, we eliminate this divergence at the source for
+deferred pods by refreshing the queued pod snapshot directly from the node's
+cache snapshot at the start of the scheduling cycle, ensuring all plugins
+evaluate the exact same immutable pod instance.
```
macsko: Can it happen that the pod in the cache is no longer deferred? What will happen in such case?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -460,46 +555,102 @@ The `NodeName` plugin is adjusted to implement the `PreFilter` phase. Because th
...
+However, race conditions can cause the Kubelet to resize a different pod than
+the one that triggered the eviction. When an assigned pod scales up on a node,
+the scheduler's assigned-pod `UpdateFunc` event handler (`updatePod` in
+`pkg/scheduler/eventhandlers.go`) detects the scale-up and checks for any
+`Deferred` pods running on that same host. To prevent cluster-wide pod
+iterations and maintain scalability, `updatePod` queries the pod informer
+indexer via its secondary index keyed by `spec.nodeName`
+(`ByIndex(nodeNameIndex, pod.Spec.NodeName)`) to look up only the pods bound to
+that specific node, calling `SchedulingQueue.Add` to immediately re-enqueue any
+lingering `Deferred` pods back into `activeQ` for re-evaluation and preemption.
```
macsko: What if the `Deferred` pod is already in the queue? Does `SchedulingQueue.Add` handle this correctly or needs adjustments?
What if the `Deferred` pod is in-flight (in the scheduling cycle)? Can it happen that its scheduling cycle says the pod fits, so it would remove the pod, while at the same time, this pod update event handler would skip adding the pod?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -709,18 +861,150 @@ When a deferred resize scheduling cycle fails (i.e., there is a fit error and pr
...
+
+#### Metrics
+
+`kube-scheduler` exports a significant number of metrics. We extend existing
+metrics and clarify skipped metrics to provide observability into scheduler
+behavior for in-place pod resizing.
+
+##### Extending existing metrics with a new `operation` label
+
+We update the following existing metrics to include a new `operation` label:
```
macsko: Are we allowed to add a new label to these metrics. For example, `scheduler_pending_pods` is stable, so it's forbidden to modify the labels

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -1215,9 +1639,7 @@ Focusing mostly on:
     heartbeats, leader election, etc.)
 -->
 
-For most of the logic, we are not introducing any new API calls. The Scheduler is already watching for pod updates. The node and pod states are cached in the Scheduler's memory, and is updated when the Informer receives an update. The kubelet will update the pod status when the resize is `Deferred` due to insufficient capacity; this is not a new API call. 
-
-The kubelet is already watching for node updates. However, the kubelet will now make new API calls to update the pod statuses of already-deferred pods in response to changes in the node-level preemption policy.
+For most of the logic, we are not introducing any new API calls. The Scheduler is already watching for pod updates. The node and pod states are cached in the Scheduler's memory, and updated when the Informers receive updates. The Kubelet updates the pod status when the resize is `Deferred` due to insufficient capacity, which is part of standard in-place pod resize behavior and not a new API call.
```
macsko: What about preemption API calls? Handling preemption for deferred pods adds new API calls to remove the victims

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
```diff
@@ -629,62 +773,70 @@ The strings under `disableResizePreemption` have the same validation and constra
...
+3. **Skipping Preemption**: Because the node status is
+   `UnschedulableAndUnresolvable` (rather than `Unschedulable`), the
+   `DefaultPreemption` PostFilter plugin skips preemption victim selection on
+   the node. The pod is placed in the `unschedulablePods` pool without evicting
+   lower-priority workloads.
+4. **Queueing Hints & Dynamic Wakeup**: `DeferredPodScheduling` registers
+   QueueingHints for Node `Update` events. If the node's
+   `spec.podPreemptionPolicy.disableResizePreemption` field is cleared or
+   modified to allow preemption, the QHint moves deferred pods assigned to that
+   node back to the `activeQ` for immediate re-evaluation.
```
macsko, line 806: Can it theoretically happen that the node is removed and re-added, but with a different preemption policy?

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Good catch on both points.

`SchedulingQueue.Add` assumes a newly created pod and does not remove from backoffQ. I updated the proposal to suggest we instead introduce `SchedulingQueue.AddOrActivate`, which will be a no-op if the Deferred pod is already in the active queue, move it to `activeQ` if it is in unscheduleable or backoff queue, or add it to active queue if it is not currently in the queue at all. For the race condition with updates during an in-flight scheduling cycle, I think we can leverage `inFlightEvents` to ensure the pod is requeued back to the active queue at the end of the cycle.

I added the high level idea to the KEP. I think we can discuss these details in implementation phase, but if you think the KEP needs more details, let me know and I will add them.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Ah sorry, thanks for the catch. I've worked with so many alpha metrics I somehow forgot there are stable ones....

I reworked this section to do something similar with what workload aware scheduling did; I added new resize-specific metrics to mirror the stable metrics. Now only the alpha metrics are extended with new label.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Thanks for the correction, updated this to add the API calls required to trigger preemption

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: Ah yes good catch, I changed this from "Update events" to say "Add or Update events".  It's already implemented in alpha for both Add and Update, but it's good to document it correctly in the KEP

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: With this proposal, the cache is being treated as the SoT. So if pod in the cache is no longer deferred (or not found), it means we assume the pod is not deferred (or deleted) and should not trigger the scheduling cycle. I added a note here that in such cases, the scheduling cycle will abort and remove the pod from the queue.

keps/sig-scheduling/5836-scheduler-preemption-for-ippr/README.md
natasha41575: actually I take that back, we need Add for restart cases, and for rapid remove then re-add with different policy cases

this is already implemented in alpha

natasha41575:
A gentle nudge @wojtek-t @macsko @dom4ha - this is ready for another round.

